PR #55436, opened 2025-12-15
Adopt ImmersiveSpace with preferredSurroundingsEffect for 'preferredDarkness' and ImmersiveSpace for 'mrui_activeStage' due to deprecation

Phinny01:
#### f97ba929583c3b390deae821fee3f3b53364103a
<pre>
Adopt ImmersiveSpace with preferredSurroundingsEffect for &apos;preferredDarkness&apos; and ImmersiveSpace for &apos;mrui_activeStage&apos; due to deprecation
<a href="https://rdar.apple.com/163514332">rdar://163514332</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=304209">https://bugs.webkit.org/show_bug.cgi?id=304209</a>

Reviewed by Richard Robinson.

Created WKFullScreenImmersiveSpaceManager to coordinate effect state between Objective-C and
SwiftUI. Added WKSurroundingsEffectView that applies preferredSurroundingsEffect on a SwiftUI view based on the
manager&apos;s current effect value. Added WKSurroundingsEffectWindow that bridges the effect view into
the UIKit fullscreen window via UIHostingController.

No ImmersiveSpace required - preferredSurroundingsEffect works in shared space for the
dark/semiDark effects WebKit needs.

* Source/WebKit/Modules/Internal/WebKitInternal.h:
* Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.h: Copied from Source/WebKit/Modules/Internal/WebKitInternal.h.
* Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift: Added.
(WKSurroundingsEffectManagerWrapper.currentEffect):
(WKSurroundingsEffectManagerWrapper.surroundingsEffect):
(WKSurroundingsEffectManager.wrapper):
(WKSurroundingsEffectManager.shared):
(WKSurroundingsEffectManager.currentEffect):
(WKSurroundingsEffectWindow.effectHostingController):
(WKSurroundingsEffectWindow.setupSurroundingsEffectIfNeeded):
* Source/WebKit/Platform/spi/visionos/WKSurroundingsEffectView.swift: Added.
(WKSurroundingsEffectView.body):
* Source/WebKit/UIProcess/ios/fullscreen/WKFullScreenWindowControllerIOS.mm:
(-[WKFullScreenParentWindowState initWithWindow:]):
(-[WKFullScreenWindowController didEnterVideoFullscreen]):
(-[WKFullScreenWindowController didExitVideoFullscreen]):
(-[WKFullScreenWindowController _enterFullScreen:windowScene:completionHandler:]):
(-[WKFullScreenWindowController _performSpatialFullScreenTransition:completionHandler:]):
(-[WKFullScreenWindowController toggleSceneDimming]):

Canonical link: <a href="https://commits.webkit.org/307258@main">https://commits.webkit.org/307258@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/53ad59c5553009d37ab874dda18cfc26ee748f90

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/143839 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/16512 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/8029 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/152507 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/97078 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/16997 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/16408 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/110609 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/97078 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/146802 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/13044 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/129245 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/91527 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/12529 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/10238 "Passed tests") | [✅ 🛠 wpe-libwebrtc](https://ews-build.webkit.org/#/builders/172/builds/2509 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/121971 "Passed tests") | | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/154819 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/16368 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/6907 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/118618 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/16403 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/13761 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/118973 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30498 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/14894 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/127073 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/71781 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/15989 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/5562 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/15723 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/79760 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/15935 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/15788 "Built successfully") | | | |

### DIFF
--- a/Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.h
+++ b/Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 Apple Inc. All rights reserved.
+ * Copyright (C) 2025 Apple Inc. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -23,19 +23,34 @@
  * THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-// Add project-level Objective-C header files here to be able to access them from within Swift sources.
+#pragma once
 
+#import <Foundation/Foundation.h>
 #import <wtf/Platform.h>
 
-#import "ModelTypes.h" // NOLINT
-#import "UIWindowScene+Extras.h"
-#import "WKMaterialHostingSupport.h"
-#import "WKMouseDeviceObserver.h"
-#import "WKPreferencesInternal.h"
-#import "WKScrollGeometry.h"
-#import "WKSeparatedImageView.h"
-#import "WKSurroundingsEffect.h"
-#import "WKUIDelegateInternal.h"
-#import "WKWebViewConfigurationInternal.h"
-#import "WKWebViewInternal.h"
-#import "_WKTextExtractionInternal.h"
+#if PLATFORM(VISION)
+
+#import <UIKit/UIKit.h>
+
+typedef NS_ENUM(NSInteger, WKSurroundingsEffectType) {
+    WKSurroundingsEffectTypeNone = 0,
+    WKSurroundingsEffectTypeSemiDark = 1,
+    WKSurroundingsEffectTypeDark = 2,
+    WKSurroundingsEffectTypeUltraDark = 3,
+};
+
+NS_HEADER_AUDIT_BEGIN(nullability, sendability)
+
+NS_SWIFT_UI_ACTOR
+@interface WKSurroundingsEffectManager : NSObject
++ (WKSurroundingsEffectManager *)shared;
+@property (nonatomic) WKSurroundingsEffectType currentEffect;
+@end
+
+@interface WKSurroundingsEffectWindow : UIWindow
+- (instancetype)initWithWindowScene:(UIWindowScene *)windowScene;
+- (void)setupSurroundingsEffectIfNeeded;
+@end
+
+NS_HEADER_AUDIT_END(nullability, sendability)
+#endif
--- a/Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
+++ b/Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
@@ -1,0 +1,107 @@
+// Copyright (C) 2025 Apple Inc. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+// 1. Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+// 2. Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+// THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+// BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+
+#if os(visionOS)
+
+import Observation
+@_weakLinked internal import SwiftUI
+import UIKit
+internal import WebKit_Internal
+
+@MainActor
+@Observable
+final class WKSurroundingsEffectManagerWrapper {
+    var currentEffect: WKSurroundingsEffectType = .none
+
+    var surroundingsEffect: SurroundingsEffect? {
+        switch currentEffect {
+        case .none:
+            nil
+        case .semiDark:
+            .semiDark
+        case .dark:
+            .dark
+        case .ultraDark:
+            .ultraDark
+        @unknown default:
+            nil
+        }
+    }
+
+    nonisolated init() {}
+}
+
+private let sharedWrapper = WKSurroundingsEffectManagerWrapper()
+
+extension WKSurroundingsEffectManager {
+    @nonobjc
+    var wrapper: WKSurroundingsEffectManagerWrapper {
+        sharedWrapper
+    }
+}
+
+@objc
+@implementation
+extension WKSurroundingsEffectManager {
+    static let sharedInstance = WKSurroundingsEffectManager()
+
+    class func shared() -> WKSurroundingsEffectManager {
+        sharedInstance
+    }
+
+    var currentEffect: WKSurroundingsEffectType {
+        get { sharedWrapper.currentEffect }
+        set { sharedWrapper.currentEffect = newValue }
+    }
+
+    override init() {}
+}
+
+@objc
+@implementation
+extension WKSurroundingsEffectWindow {
+    private final var effectHostingController: UIViewController?
+
+    @MainActor
+    func setupSurroundingsEffectIfNeeded() {
+        guard effectHostingController == nil, let rootViewController else {
+            return
+        }
+
+        let wrapper = WKSurroundingsEffectManager.shared().wrapper
+
+        let effectView = WKSurroundingsEffectView()
+            .environment(wrapper)
+        let hostingController = UIHostingController(rootView: effectView)
+        hostingController.view.backgroundColor = .clear
+        hostingController.view.isUserInteractionEnabled = false
+        hostingController.view.frame = rootViewController.view.bounds
+        hostingController.view.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+
+        effectHostingController = hostingController
+        rootViewController.addChild(hostingController)
+        rootViewController.view.insertSubview(hostingController.view, at: 0)
+        hostingController.didMove(toParent: rootViewController)
+    }
+}
+#endif
--- a/Source/WebKit/Platform/spi/visionos/WKSurroundingsEffectView.swift
+++ b/Source/WebKit/Platform/spi/visionos/WKSurroundingsEffectView.swift
@@ -1,0 +1,38 @@
+// Copyright (C) 2025 Apple Inc. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+// 1. Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+// 2. Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+// THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+// BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+
+#if os(visionOS)
+
+@_weakLinked internal import SwiftUI
+internal import WebKit_Internal
+
+struct WKSurroundingsEffectView: View {
+    @Environment(WKSurroundingsEffectManagerWrapper.self)
+    private var wrapper
+
+    var body: some View {
+        EmptyView()
+            .preferredSurroundingsEffect(wrapper.surroundingsEffect)
+    }
+}
+#endif
--- a/Source/WebKit/UIProcess/ios/fullscreen/WKFullScreenWindowControllerIOS.mm
+++ b/Source/WebKit/UIProcess/ios/fullscreen/WKFullScreenWindowControllerIOS.mm
@@ -76,6 +76,10 @@
 #import "UIWindowScene+Extras.h"
 #endif
 
+#if PLATFORM(VISION)
+#import "WKSurroundingsEffect.h"
+#endif
+
 #import "WebKitSwiftSoftLink.h"
 
 #if !HAVE(URL_FORMATTING)
@@ -358,7 +362,6 @@ static constexpr CGFloat kIncomingWindowFadeDuration = 0.6;
 static constexpr CGFloat kIncomingWindowFadeDelay = 0.2;
 static constexpr CGFloat kIncomingWindowZOffset = -170;
 static constexpr CGFloat kWindowTranslationDuration = 0.6;
-static constexpr NSString *kPrefersFullScreenDimmingKey = @"WebKitPrefersFullScreenDimming";
 #endif
 
 #pragma mark -
@@ -661,7 +664,7 @@ static constexpr NSString *kPrefersFullScreenDimmingKey = @"WebKitPrefersFullScr
 @property (nonatomic, readonly) CGSize sceneMinimumSize;
 @property (nonatomic, readonly) RSSSceneChromeOptions sceneChromeOptions;
 @property (nonatomic, readonly) MRUISceneResizingBehavior sceneResizingBehavior;
-@property (nonatomic, readonly) MRUIDarknessPreference preferredDarkness;
+@property (nonatomic, readonly) WKSurroundingsEffectType preferredSurroundingsEffect;
 @property (nonatomic, assign) BOOL prefersAutoDimming;
 
 @property (nonatomic, readonly) NSMapTable<MRUIPlatterOrnament *, WKMRUIPlatterOrnamentProperties *> *ornamentProperties;
@@ -681,11 +684,9 @@ static constexpr NSString *kPrefersFullScreenDimmingKey = @"WebKitPrefersFullScr
 
     _transform3D = window.transform3D;
     _windowClass = object_getClass(window);
-ALLOW_DEPRECATED_DECLARATIONS_BEGIN
-    _preferredDarkness = UIApplication.sharedApplication.mrui_activeStage.preferredDarkness;
-ALLOW_DEPRECATED_DECLARATIONS_END
 
     UIWindowScene *windowScene = window.windowScene;
+    _preferredSurroundingsEffect = [WKSurroundingsEffectManager shared].currentEffect;
 ALLOW_DEPRECATED_DECLARATIONS_BEGIN
     _sceneSize = windowScene.coordinateSpace.bounds.size;
 ALLOW_DEPRECATED_DECLARATIONS_END
@@ -1009,10 +1010,8 @@ ALLOW_DEPRECATED_DECLARATIONS_END
 - (void)didEnterVideoFullscreen
 {
 #if PLATFORM(VISION)
-    if (self.isFullScreen)
-ALLOW_DEPRECATED_DECLARATIONS_BEGIN
-        UIApplication.sharedApplication.mrui_activeStage.preferredDarkness = MRUIDarknessPreferenceUnspecified;
-ALLOW_DEPRECATED_DECLARATIONS_END
+    if (self.isFullScreen && [WKSurroundingsEffectManager shared].currentEffect != WKSurroundingsEffectTypeNone)
+        [WKSurroundingsEffectManager shared].currentEffect = WKSurroundingsEffectTypeNone;
 #endif
 }
 
@@ -1027,9 +1026,10 @@ ALLOW_DEPRECATED_DECLARATIONS_END
         if (RefPtr bestVideo = videoPresentationManager->bestVideoForElementFullscreen())
             prefersAutoDimming = bestVideo->playbackSessionModel()->prefersAutoDimming();
     }
-ALLOW_DEPRECATED_DECLARATIONS_BEGIN
-    UIApplication.sharedApplication.mrui_activeStage.preferredDarkness = prefersAutoDimming ? MRUIDarknessPreferenceDark : MRUIDarknessPreferenceUnspecified;
-ALLOW_DEPRECATED_DECLARATIONS_END
+
+    WKSurroundingsEffectType targetEffect = prefersAutoDimming ? WKSurroundingsEffectTypeDark : WKSurroundingsEffectTypeNone;
+    if ([WKSurroundingsEffectManager shared].currentEffect != targetEffect)
+        [WKSurroundingsEffectManager shared].currentEffect = targetEffect;
 #endif
 }
 
@@ -1083,7 +1083,12 @@ ALLOW_DEPRECATED_DECLARATIONS_END
     _blocksReturnToFullscreenFromPictureInPicture = manager->blocksReturnToFullscreenFromPictureInPicture();
     _originalWindowSize = [webView window].frame.size;
 
+#if PLATFORM(VISION)
+    RetainPtr<WKSurroundingsEffectWindow> surroundingsEffectWindow = adoptNS([[WKSurroundingsEffectWindow alloc] initWithWindowScene:windowScene]);
+    _window = surroundingsEffectWindow;
+#else
     _window = adoptNS([[UIWindow alloc] initWithWindowScene:windowScene]);
+#endif
     [_window setBackgroundColor:[UIColor clearColor]];
     [_window setWindowLevel:UIWindowLevelNormal - 1];
     [_window setHidden:NO];
@@ -1122,6 +1127,11 @@ ALLOW_DEPRECATED_DECLARATIONS_END
     [_rootViewController setTransitioningDelegate:self];
 
     _window.get().rootViewController = _rootViewController.get();
+
+#if PLATFORM(VISION)
+    if (WebKit::useSpatialFullScreenTransition())
+        [surroundingsEffectWindow setupSurroundingsEffectIfNeeded];
+#endif
 
     _fullscreenViewController = adoptNS([[WKFullScreenViewController alloc] initWithWebView:webView.get()]);
     [_fullscreenViewController setModalPresentationStyle:UIModalPresentationCustom];
@@ -2127,16 +2137,14 @@ ALLOW_DEPRECATED_DECLARATIONS_END
 
     inWindow.transform3D = CATransform3DTranslate(originalState.transform3D, 0, 0, kIncomingWindowZOffset);
 
-ALLOW_DEPRECATED_DECLARATIONS_BEGIN
-    MRUIStage *stage = UIApplication.sharedApplication.mrui_activeStage;
-    MRUIDarknessPreference targetDarkness = enter ? (self.prefersSceneDimming ? MRUIDarknessPreferenceDark : originalState.preferredDarkness) : originalState.preferredDarkness;
+    WKSurroundingsEffectType targetDarkness = enter ? (self.prefersSceneDimming ? WKSurroundingsEffectTypeDark : originalState.preferredSurroundingsEffect) : originalState.preferredSurroundingsEffect;
 
-    if (stage.preferredDarkness != targetDarkness) {
+    WKSurroundingsEffectType currentEffect = [WKSurroundingsEffectManager shared].currentEffect;
+    if (currentEffect != targetDarkness) {
         [UIView animateWithDuration:kDarknessAnimationDuration animations:^{
-            stage.preferredDarkness = targetDarkness;
+            [WKSurroundingsEffectManager shared].currentEffect = targetDarkness;
         } completion:nil];
     }
-ALLOW_DEPRECATED_DECLARATIONS_END
 
     [UIView animateWithDuration:kOutgoingWindowFadeDuration delay:0 options:UIViewAnimationOptionCurveEaseInOut animations:^{
         if (enter)
@@ -2225,11 +2233,9 @@ ALLOW_DEPRECATED_DECLARATIONS_END
     }
 
     if (self.isFullScreen) {
-ALLOW_DEPRECATED_DECLARATIONS_BEGIN
-        MRUIDarknessPreference target = updatedPrefersSceneDimming ? MRUIDarknessPreferenceDark : (_parentWindowState ? [_parentWindowState preferredDarkness] : MRUIDarknessPreferenceUnspecified);
-        MRUIStage *stage = UIApplication.sharedApplication.mrui_activeStage;
-        stage.preferredDarkness = target;
-ALLOW_DEPRECATED_DECLARATIONS_END
+        WKSurroundingsEffectType target = updatedPrefersSceneDimming ? WKSurroundingsEffectTypeDark : (_parentWindowState ? [_parentWindowState preferredSurroundingsEffect] : WKSurroundingsEffectTypeNone);
+        if ([WKSurroundingsEffectManager shared].currentEffect != target)
+            [WKSurroundingsEffectManager shared].currentEffect = target;
     }
 }
 
--- a/Source/WebKit/WebKit.xcodeproj/project.pbxproj
+++ b/Source/WebKit/WebKit.xcodeproj/project.pbxproj
@@ -1870,7 +1870,10 @@
 		961A72D02E17270900DD2AF9 /* WebExtensionBookmarksParameters.h in Headers */ = {isa = PBXBuildFile; fileRef = 96CC7DBF2E15EC2800233E69 /* WebExtensionBookmarksParameters.h */; };
 		96418A5E2DFA4B8100C19CDC /* JSWebExtensionAPIBookmarks.h in Headers */ = {isa = PBXBuildFile; fileRef = 96418A5D2DFA4B8100C19CDC /* JSWebExtensionAPIBookmarks.h */; };
 		96EE6CE82DFA4980000F90BD /* WebExtensionAPIBookmarks.h in Headers */ = {isa = PBXBuildFile; fileRef = 96EE6CE72DFA4975000F90BD /* WebExtensionAPIBookmarks.h */; };
+		97311DE42F0DB61F00B23BE9 /* WKSurroundingsEffectView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 97311DE32F0DB1CF00B23BE9 /* WKSurroundingsEffectView.swift */; };
 		9732D19C2ECFD0B600AC856D /* _WKCaptionStyleMenuControllerAVKitMac.mm in Sources */ = {isa = PBXBuildFile; fileRef = 97AD28872ECFA49800039097 /* _WKCaptionStyleMenuControllerAVKitMac.mm */; };
+		976544262EF3514300BD8947 /* WKSurroundingsEffect.h in Headers */ = {isa = PBXBuildFile; fileRef = 976544232EF34BDE00BD8947 /* WKSurroundingsEffect.h */; };
+		9788BFD92EF4A6940038405A /* WKSurroundingsEffect.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9765441E2EF22B0B00BD8947 /* WKSurroundingsEffect.swift */; };
 		97AD28742ECE625300039097 /* _WKCaptionStyleMenuControllerAVKit.mm in Sources */ = {isa = PBXBuildFile; fileRef = 97F26D492EC67188006096CD /* _WKCaptionStyleMenuControllerAVKit.mm */; };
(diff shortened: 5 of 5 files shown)


Source/WebKit/UIProcess/ios/fullscreen/WKFullScreenWindowControllerIOS.mm
```diff
@@ -68,6 +68,16 @@
 
 #if PLATFORM(VISION)
 #import "MRUIKitSPI.h"
+
+// Forward declarations for ImmersiveSpace API (visionOS 2.0+)
+// TODO: Remove once these are in MRUIKitSPI.h
```
jyavenard: that TODO comment is unlikely to be correct.
Those interfaces aren't SPI

We aren't building for visionOS < 2.0 anymore nor will we ever be going to. Please use the proper include for it, no need for forward declarations

Phinny01: You're right. They are SwiftUI APIs. I have corrected it now.

Source/WebKit/UIProcess/Cocoa/WebKitSwiftSoftLink.h
```diff
@@ -52,3 +52,5 @@ SOFT_LINK_CLASS_FOR_HEADER(WebKit, WKIdentityDocumentPresentmentMobileDocumentRe
 SOFT_LINK_CLASS_FOR_HEADER(WebKit, WKIdentityDocumentPresentmentRequest);
 SOFT_LINK_CLASS_FOR_HEADER(WebKit, WKIdentityDocumentPresentmentRequestAuthenticationCertificate);
 SOFT_LINK_CLASS_FOR_HEADER(WebKit, WKTextAnimationManager)
+SOFT_LINK_CLASS_FOR_HEADER(WebKit, WKFullScreenImmersiveSpaceManager);
+SOFT_LINK_CLASS_FOR_HEADER(WebKit, WKSurroundingsEffectWindow);
```
pxlcoder: I don't think your classes need to be in `WebKitSwift`. They can be in WebKit itself. You should be able to follow `WKMaterialHostingSupport` as an example.

Source/WebKit/UIProcess/ios/fullscreen/WKFullScreenImmersiveSpaceDelegate.h
```diff
@@ -0,0 +1,47 @@
...
+#if TARGET_OS_VISION
+
+#import <UIKit/UIKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface WKFullScreenImmersiveSpaceManager : NSObject
++ (instancetype)shared;
+@property (nonatomic) NSInteger currentEffect;
+- (void)setEffect:(NSInteger)effect;
```
pxlcoder: This should be a single readwrite property.

Source/WebKit/UIProcess/ios/fullscreen/WKFullScreenWindowControllerIOS.mm
```diff
@@ -121,6 +125,19 @@ static bool useSpatialFullScreenTransition()
     return [[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomVision;
 }
 
+static MRUIDarknessPreference getSurroundingsEffect()
+{
+    WKFullScreenImmersiveSpaceManager *manager = [getWKFullScreenImmersiveSpaceManagerClassSingleton() shared];
+    NSInteger effect = manager.currentEffect;
+    return (MRUIDarknessPreference)effect;
```
pxlcoder: What's the reason for the casting? Why not use the actual type in the Swift interface?

Source/WebKit/WebKit.xcodeproj/project.pbxproj
```diff
@@ -1835,6 +1834,8 @@
 		96418A5E2DFA4B8100C19CDC /* JSWebExtensionAPIBookmarks.h in Headers */ = {isa = PBXBuildFile; fileRef = 96418A5D2DFA4B8100C19CDC /* JSWebExtensionAPIBookmarks.h */; };
 		96EE6CE82DFA4980000F90BD /* WebExtensionAPIBookmarks.h in Headers */ = {isa = PBXBuildFile; fileRef = 96EE6CE72DFA4975000F90BD /* WebExtensionAPIBookmarks.h */; };
 		9732D19C2ECFD0B600AC856D /* _WKCaptionStyleMenuControllerAVKitMac.mm in Sources */ = {isa = PBXBuildFile; fileRef = 97AD28872ECFA49800039097 /* _WKCaptionStyleMenuControllerAVKitMac.mm */; };
+		976544222EF2847D00BD8947 /* WKFullScreenImmersiveSpaceDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9765441E2EF22B0B00BD8947 /* WKFullScreenImmersiveSpaceDelegate.swift */; };
+		976544262EF3514300BD8947 /* WKFullScreenImmersiveSpaceDelegate.h in Headers */ = {isa = PBXBuildFile; fileRef = 976544232EF34BDE00BD8947 /* WKFullScreenImmersiveSpaceDelegate.h */; settings = {ATTRIBUTES = (Private, ); }; };
```
pxlcoder: We certainly don't want this to be a private header because that means people outside WebKit could see and use it.

(following the pattern of `WKMaterialHostingSupport` mentioned above should address this)

Source/WebKit/UIProcess/Cocoa/WebKitSwiftSoftLink.h
Phinny01: Thanks that was helpful.

Source/WebKit/UIProcess/ios/fullscreen/WKFullScreenImmersiveSpaceDelegate.h
Phinny01: I addressed this

Source/WebKit/UIProcess/ios/fullscreen/WKFullScreenWindowControllerIOS.mm
Phinny01: You're right we don't need to

Source/WebKit/WebKit.xcodeproj/project.pbxproj
Phinny01: Right! thanks!

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
```diff
@@ -0,0 +1,127 @@
...
+//AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+//THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+//PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+//BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+//CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+//SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+//INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+//CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+//ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+//THE POSSIBILITY OF SUCH DAMAGE.
```
pxlcoder: There should be a single space after `//` on each line.

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
```diff
@@ -0,0 +1,127 @@
...
+//
```
pxlcoder: No need for the trailing `//`

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
```diff
@@ -0,0 +1,127 @@
...
+#if os(visionOS)
+
+import Combine
+@_weakLinked internal import SwiftUI
+import UIKit
+
+/// Manages surroundings effect state for fullscreen web content in visionOS.
+///
+/// This singleton coordinates the surroundings dimming effect between Objective-C/UIKit
+/// and SwiftUI layers, replacing the deprecated MRUIStage.preferredDarkness API.
```
pxlcoder: I don't think this comment is really necessary. Though if you decide to keep it, I would certainly remove the "replacing the deprecated MRUIStage.preferredDarkness API" – that detail is not relevant over time.

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
```diff
@@ -0,0 +1,127 @@
...
+@available(visionOS 1.0, *)
+public class WKFullScreenImmersiveSpaceManager: NSObject, ObservableObject {
+    /// Shared singleton instance of the manager.
+    @objc
+    public static let shared = WKFullScreenImmersiveSpaceManager()
+
+    /// The current surroundings effect level (0=none, 1=semiDark, 2=dark, 3=ultraDark).
+    @Published
+    @objc
+    public dynamic var currentEffect: Int = 0
```
pxlcoder: This should be an actual enumeration rather than an `Int`. Then the comment is also unnecessary.

Again `WKMaterialHostingSupport` could serve as inspiration (`WKHostedMaterialEffectType`).

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
```diff
@@ -0,0 +1,127 @@
...
+        switch self.currentEffect {
+        case 0:
+            return nil
+        case 1:
+            return .semiDark
+        case 2:
+            return .dark
+        case 3:
+            return .ultraDark
+        default:
```
pxlcoder: Once we define a real enumeration, this can be `@unknown default`.

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
```diff
@@ -0,0 +1,127 @@
...
+    /// The current surroundings effect level (0=none, 1=semiDark, 2=dark, 3=ultraDark).
+    @Published
+    @objc
+    public dynamic var currentEffect: Int = 0
+
+    private override init() {
+        super.init()
+    }
+
+    var swiftUIEffect: SurroundingsEffect? {
```
pxlcoder: We're in SwiftUI code, and this property isn't exposed outside of SwiftUI code, so SwiftUI in the name doesn't add value. `fileprivate` because this variable is only in used in this file.

```suggestion
    fileprivate var surroundingsEffect: SurroundingsEffect? {
```

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
```diff
@@ -0,0 +1,127 @@
...
+    public static let shared = WKFullScreenImmersiveSpaceManager()
+
+    /// The current surroundings effect level (0=none, 1=semiDark, 2=dark, 3=ultraDark).
+    @Published
+    @objc
+    public dynamic var currentEffect: Int = 0
+
+    private override init() {
+        super.init()
+    }
```
pxlcoder: Is the idea behind this to hide the initializer so clients must use `shared`? Otherwise it seems unnecessary.

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
```diff
@@ -0,0 +1,127 @@
...
+            return .ultraDark
+        default:
+            return nil
+        }
+    }
+}
+
+struct WKSurroundingsEffectView: View {
+    @ObservedObject
+    var manager = WKFullScreenImmersiveSpaceManager.shared
```
pxlcoder: Is the variable necessary? If so, I think it should be `private`.

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
```diff
@@ -0,0 +1,127 @@
...
+    private var effectHostingController: UIHostingController<WKSurroundingsEffectView>?
+    private var needsEffectSetup = true
+
+    public override init(windowScene: UIWindowScene) {
+        super.init(windowScene: windowScene)
+    }
+
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
+    }
```
pxlcoder: Are these initializers necessary? They don't do anything other than the original behavior.

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
```diff
@@ -0,0 +1,127 @@
...
+            .ignoresSafeArea()
+            .preferredSurroundingsEffect(manager.swiftUIEffect)
+            .allowsHitTesting(false)
+    }
+}
+
+/// A UIWindow subclass that applies SwiftUI surroundings effects to fullscreen web content.
+///
+/// This window bridges SwiftUI's preferredSurroundingsEffect modifier into the UIKit-based
+/// fullscreen window hierarchy, enabling surroundings dimming without using deprecated APIs.
```
pxlcoder: Also not sure this comment is necessary. 

"A UIWindow subclass that applies SwiftUI surroundings effects to fullscreen web content", is inaccurate since this class is used in other parts of the code to apply "SwiftUI surroundings effects to fullscreen web content". This code alone does not know about fullscreen.

And same comment above about mentioning the deprecated APIs.

Source/WebKit/UIProcess/ios/fullscreen/WKFullScreenImmersiveSpaceDelegate.h
```diff
@@ -0,0 +1,46 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import <Foundation/Foundation.h>
+#import <TargetConditionals.h>
+
+#if TARGET_OS_VISION
```
pxlcoder: We should use our own `PLATFORM` macro and remove the import of `TargetConditionals.h`.

Source/WebKit/UIProcess/ios/fullscreen/WKFullScreenWindowControllerIOS.mm
```diff
@@ -121,6 +125,21 @@ static bool useSpatialFullScreenTransition()
     return [[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomVision;
 }
 
+static constexpr NSInteger SurroundingsEffectLevelNone = 0;
+static constexpr NSInteger SurroundingsEffectLevelSemiDark = 1;
+static constexpr NSInteger SurroundingsEffectLevelDark = 2;
+static constexpr NSInteger SurroundingsEffectLevelUltraDark = 3;
```
pxlcoder: These should be an actual enum as mentioned in an earlier comment.

Source/WebKit/UIProcess/ios/fullscreen/WKFullScreenWindowControllerIOS.mm
```diff
@@ -121,6 +125,21 @@ static bool useSpatialFullScreenTransition()
...
+
+static NSInteger getSurroundingsEffect()
+{
+    return [WKFullScreenImmersiveSpaceManager shared].currentEffect;
+}
+
+static void setSurroundingsEffect(NSInteger effect)
+{
+    [WKFullScreenImmersiveSpaceManager shared].currentEffect = effect;
+}
```
pxlcoder: I don't think we need these methods. The callers could just inline the single line of code?

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
```diff
@@ -0,0 +1,127 @@
...
+import UIKit
+
+/// Manages surroundings effect state for fullscreen web content in visionOS.
+///
+/// This singleton coordinates the surroundings dimming effect between Objective-C/UIKit
+/// and SwiftUI layers, replacing the deprecated MRUIStage.preferredDarkness API.
+@MainActor
+@objc(WKFullScreenImmersiveSpaceManager)
+@available(visionOS 1.0, *)
+public class WKFullScreenImmersiveSpaceManager: NSObject, ObservableObject {
```
pxlcoder: I think this class could be renamed to remove the mention of "FullScreen" since its implementation is independent of fullscreen.

pxlcoder: On further thought is this class truly necessary? Could the current effect simply live on the `WKSurroundingsEffectWindow`?

pxlcoder (commented):
It's unclear why the new file is called `WKFullScreenImmersiveSpace**Delegate**`. There's no use of the delegate pattern here.

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
Phinny01: Yes I think it is necessary-  without it, the UI would be stale but you are right it should be private

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
Phinny01: Yes I think it is necessary. If we stored it directly on the window, we'd need manual KVO or notification mechanisms to make SwiftUI react to Objective-C property changes which I think would be more complex than this.

Source/WebKit/Platform/spi/visionos/WKFullScreenImmersiveSpaceDelegate.swift
Phinny01: Yes exactly

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.h
```diff
@@ -0,0 +1,55 @@
...
+#import <UIKit/UIKit.h>
+
+typedef NS_ENUM(NSInteger, WKSurroundingsEffectType) {
+    WKSurroundingsEffectTypeNone = 0,
+    WKSurroundingsEffectTypeSemiDark = 1,
+    WKSurroundingsEffectTypeDark = 2,
+    WKSurroundingsEffectTypeUltraDark = 3,
+};
+
+NS_ASSUME_NONNULL_BEGIN
```
rr-codes: this should be 

```
NS_HEADER_AUDIT_BEGIN(nullability, sendability)
```

instead (and likewise for the END one)

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.h
```diff
@@ -0,0 +1,55 @@
...
+
+@interface WKSurroundingsEffectManager : NSObject
```
rr-codes, line 45: this type itself should just be NS_SWIFT_UI_ACTOR, and then the methods/properties don’t have to explicitly say so

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+
+#if os(visionOS)
+
+import Combine
```
rr-codes: There is no need to use Combine; The Observation framework should be preferred instead

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+import UIKit
+
+@objc(WKSurroundingsEffectType)
+@available(visionOS 1.0, *)
+enum WKSurroundingsEffectType: Int {
+    case none = 0
+    case semiDark = 1
+    case dark = 2
+    case ultraDark = 3
+}
```
rr-codes: this is a duplicate definition of the other WKSurroundingsEffectType type and should be deleted

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+
+@MainActor
+@objc(WKSurroundingsEffectManager)
+@available(visionOS 1.0, *)
+class WKSurroundingsEffectManager: NSObject, ObservableObject {
```
rr-codes: this needs to just be `@objc @implementation extension WKSurroundingsEffectManager`

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+    @objc
```
rr-codes: and then no `@objc` on any of the members are needed

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+    static let shared = WKSurroundingsEffectManager()
+
+    @Published
+    @objc
+    dynamic var currentEffect: WKSurroundingsEffectType = .none
```
rr-codes: @Published and `dynamic` can be removed once Observation is used instead of Combine

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+
+    private override init() {
```
rr-codes: this is unnecessary; the init should just be marked as NS_UNAVAILABLE on the objc side

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+        super.init()
+    }
+
+    fileprivate var surroundingsEffect: SurroundingsEffect? {
+        switch self.currentEffect {
```
rr-codes: no `self.`

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+        case .none:
+            return nil
```
rr-codes: no need for any `return` keywords

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+            return .dark
+        case .ultraDark:
+            return .ultraDark
+        @unknown default:
+            return nil
+        }
+    }
+}
+
+struct WKSurroundingsEffectView: View {
```
rr-codes: this should be in its own file

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+    @ObservedObject
+    private var manager = WKSurroundingsEffectManager.shared
```
rr-codes: this should be injected from the environment

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+
+    var body: some View {
+        Color.clear
```
rr-codes: why not just `EmptyView()`?

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+            .ignoresSafeArea()
+            .preferredSurroundingsEffect(manager.surroundingsEffect)
+            .allowsHitTesting(false)
+    }
+}
+
+@objc(WKSurroundingsEffectWindow)
+@available(visionOS 1.0, *)
+class WKSurroundingsEffectWindow: UIWindow {
```
rr-codes: ditto about @objc @implementation

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,117 @@
...
+        didSet {
+            if needsEffectSetup && rootViewController != nil {
+                needsEffectSetup = false
+                setupSurroundingsEffect()
+            }
+        }
+    }
+
+    private func setupSurroundingsEffect() {
+        guard let rootVC = rootViewController else {
```
rr-codes:
```suggestion
        guard let rootViewController else {
```

rr-codes (changes requested):
(requesting changes per my above comments)

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.h
```diff
@@ -0,0 +1,56 @@
...
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#import <Foundation/Foundation.h>
+#import <wtf/PlatformHave.h>
```
rr-codes:
```suggestion
#import <wtf/Platform.h>
```

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,87 @@
...
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+
+#if os(visionOS)
+
+import Observation
```
rr-codes, line 26:
```suggestion
internal import Observation
```

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,87 @@
...
+@_weakLinked internal import SwiftUI
+import UIKit
```
rr-codes, line 28:
```suggestion
internal import UIKit
```

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,87 @@
...
+internal import WebKit_Internal
+
+@objc
+@implementation
+extension WKSurroundingsEffectManager {
+    private static let shared = WKSurroundingsEffectManager()
+
+    class func shared() -> Self {
+        shared as? Self ?? Self()
+    }
```
rr-codes:
```suggestion
    class func shared() -> WKSurroundingsEffectManager {
        shared
    }
```

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,87 @@
...
+        case .ultraDark:
+            .ultraDark
+        @unknown default:
+            nil
+        }
+    }
+}
+
+@available(visionOS 1.0, *)
+extension WKSurroundingsEffectManager: Observable {
```
rr-codes: this is wrong; an Objective-C type should not be made Observable, nor should Observable be directly used as conformance

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,87 @@
...
+extension WKSurroundingsEffectManager {
+    private static let shared = WKSurroundingsEffectManager()
+
+    class func shared() -> Self {
+        shared as? Self ?? Self()
+    }
+
+    var currentEffect: WKSurroundingsEffectType = .none
+
+    final var surroundingsEffect: SurroundingsEffect? {
```
rr-codes:
```suggestion
   @nonobjc
   var surroundingsEffect: SurroundingsEffect? {
```

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,87 @@
...
+extension WKSurroundingsEffectManager: Observable {
+}
+
+@objc
+@implementation
+extension WKSurroundingsEffectWindow {
+    private final var effectHostingController: UIHostingController<WKSurroundingsEffectView>?
+
+    private func setupSurroundingsEffectIfNeeded() {
+        guard effectHostingController == nil, let rootViewController = rootViewController else {
```
rr-codes:
```suggestion
        guard effectHostingController == nil, let rootViewController else {
```

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,110 @@
...
+            nil
+        }
+    }
+
+    nonisolated init() {}
+}
+
+private let sharedWrapper = WKSurroundingsEffectManagerWrapper()
+
+extension WKSurroundingsEffectManager {
```
rr-codes, line 56: why is this wrapper needed?

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,110 @@
...
+    class func shared() -> WKSurroundingsEffectManager {
+        sharedInstance
+    }
+
+    var currentEffect: WKSurroundingsEffectType {
+        get { sharedWrapper.currentEffect }
+        set { sharedWrapper.currentEffect = newValue }
+    }
+
+    override init() {}
```
rr-codes, line 77: why is this needed?

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,110 @@
...
+}
+
+@objc
+@implementation
+extension WKSurroundingsEffectWindow {
+    private final var effectHostingController: UIHostingController<WKSurroundingsEffectView>?
+    private final var effectManagerWrapper: WKSurroundingsEffectManagerWrapper?
```
rr-codes: why is this needed when it’s always just the singleton instance?

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,110 @@
...
+            return
+        }
+
+        let wrapper = WKSurroundingsEffectManager.shared().wrapper
+        effectManagerWrapper = wrapper
+
+        let effectView = WKSurroundingsEffectView()
+            .environment(wrapper)
+        let hostingController = UIHostingController(rootView: effectView)
+        hostingController.view.backgroundColor = UIColor.clear
```
rr-codes:
```suggestion
        hostingController.view.backgroundColor = .clear
```

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,110 @@
...
+        hostingController.view.isUserInteractionEnabled = false
+        hostingController.view.frame = rootViewController.view.bounds
+        hostingController.view.autoresizingMask = [UIView.AutoresizingMask.flexibleWidth, UIView.AutoresizingMask.flexibleHeight]
```
rr-codes:
```suggestion
        hostingController.view.autoresizingMask = [.flexibleWidth, .flexibleHeight]
```

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
```diff
@@ -0,0 +1,110 @@
...
+
+        effectHostingController = hostingController as? UIHostingController<WKSurroundingsEffectView>
```
rr-codes: why is this cast needed?

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffectView.swift
```diff
@@ -0,0 +1,40 @@
...
+struct WKSurroundingsEffectView: View {
+    @Environment(WKSurroundingsEffectManagerWrapper.self)
+    private var wrapper
+
+    var body: some View {
+        let effect = wrapper.surroundingsEffect
+
+        return EmptyView()
+            .preferredSurroundingsEffect(effect)
+    }
```
rr-codes, line 36:
```suggestion
    var body: some View {
        EmptyView()
            .preferredSurroundingsEffect(wrapper.surroundingsEffect)
    }
```

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffectView.swift
```diff
@@ -0,0 +1,40 @@
...
+internal import WebKit_Internal
+
+struct WKSurroundingsEffectView: View {
+    @Environment(WKSurroundingsEffectManagerWrapper.self)
+    private var wrapper
+
+    var body: some View {
+        let effect = wrapper.surroundingsEffect
+
+        return EmptyView()
```
rr-codes: why is this an empty view?

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
Phinny01: The wrapper provides @Observable support for the Objective-C WKSurroundingsEffectManager class. The @Observable macro only works with Swift classes, so the wrapper acts as an observable bridge, enabling reactive SwiftUI updates via .environment(wrapper).

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
Phinny01: You're right. This is unnecessary.

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffectView.swift
Phinny01: We only need a SwiftUI view to apply the .preferred​Surroundings​Effect() modifier. Empty​View is used because the view itself doesn't render any content - it exists solely to propagate the surroundings effect preference through the view hierarchy to the window.

Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift
Phinny01: This is required to override the Objective-C class's designated initializer, allowing us to create shared​Instance. The @implementation extension needs this to satisfy the class's initialization contract.

rr-codes (approved):

PR #55472, opened 2025-12-16
[GLIB] Flaky debug crashes due to assertion failure in ThreadedCompositor::renderLayerTree

carlosgcampos:
#### 6cd29fa8c4e7d7e0d7b58cdece6f00ee2614bcc3
<pre>
[GLIB] Flaky debug crashes due to assertion failure in ThreadedCompositor::renderLayerTree
<a href="https://bugs.webkit.org/show_bug.cgi?id=304204">https://bugs.webkit.org/show_bug.cgi?id=304204</a>

Since 303824@main the compositor can be in Scheduled state, but with the
render timer not active, because the only composition request was for a
rendering update and tiles were not ready yet. Other composition requests
are handled while waiting for tiles, when a request is made when
already in Scheduled state and the timer is not active, the timer is
then activated. RunLoopGLib sets the timer as inactive right before
calling the timer callback. It can haoppen that after the timer is set
inactive and before the callback is called or the state lock is acquired
by renderLayerTree the main thread is noitified that the last tile has
been rendered, which calls scheduleUpdateLocked() in a Scheduled state
with the timer inactive, activating the timer again instead of switching
to ScheduledWhileWaitingForRenderer. This causes that we end up
compositing twice in a row in the same frame. This patch uses its own
flag to track if the render timer is active, also protected by the state
lock, to ensure that when renderLayerTree acquires the lock the flags is
still active and state is Scheduled.

Canonical link: <a href="https://commits.webkit.org/304581@main">https://commits.webkit.org/304581@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/f355e3301a966c8358feff9b20bf691059ae632e

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/135818 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/8195 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/138/builds/47110 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/143531 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/87421 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/8853 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/8040 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/103798 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/71752 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/138764 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/6390 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/121735 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/84674 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/6118 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/3743 "Passed tests") | [✅ 🛠 wpe-cairo-libwebrtc](https://ews-build.webkit.org/#/builders/166/builds/4137 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/115358 "Passed tests") | | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/146281 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/7880 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/136/builds/40490 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/112154 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/7907 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/6610 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/112540 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/6015 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/118035 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/61794 "Built successfully") | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/20948 "Built successfully and passed tests") | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/7927 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/137/builds/36123 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/7662 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/71477 "Built successfully") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/7884 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/7747 "Built successfully") | | |

### DIFF
--- a/Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/ThreadedCompositor.cpp
+++ b/Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/ThreadedCompositor.cpp
@@ -119,7 +119,7 @@ void ThreadedCompositor::invalidate()
 
     {
         Locker locker { m_state.lock };
-        m_renderTimer.stop();
+        stopRenderTimer();
         m_state.didCompositeRenderingUpdateFunction = nullptr;
         m_state.state = State::Idle;
     }
@@ -143,6 +143,27 @@ void ThreadedCompositor::invalidate()
     m_surface = nullptr;
 }
 
+void ThreadedCompositor::startRenderTimer()
+{
+    ASSERT(m_state.lock.isHeld());
+    ASSERT(!m_state.isRenderTimerActive);
+    m_state.isRenderTimerActive = true;
+    m_renderTimer.startOneShot(0_s);
+}
+
+void ThreadedCompositor::stopRenderTimer()
+{
+    ASSERT(m_state.lock.isHeld());
+    m_state.isRenderTimerActive = false;
+    m_renderTimer.stop();
+}
+
+bool ThreadedCompositor::isOnlyRenderingUpdatePendingAndWaitingForTiles() const
+{
+    ASSERT(m_state.lock.isHeld());
+    return m_state.reasons.containsOnly({ CompositionReason::RenderingUpdate }) && m_state.isWaitingForTiles;
+}
+
 void ThreadedCompositor::suspend()
 {
     ASSERT(RunLoop::isMain());
@@ -151,7 +172,8 @@ void ThreadedCompositor::suspend()
     if (++m_suspendedCount > 1)
         return;
 
-    m_renderTimer.stop();
+    Locker locker { m_state.lock };
+    stopRenderTimer();
 }
 
 void ThreadedCompositor::resume()
@@ -164,8 +186,8 @@ void ThreadedCompositor::resume()
         return;
 
     Locker locker { m_state.lock };
-    if (m_state.state == State::Scheduled)
-        m_renderTimer.startOneShot(0_s);
+    if (m_state.state == State::Scheduled && !isOnlyRenderingUpdatePendingAndWaitingForTiles())
+        startRenderTimer();
 }
 
 bool ThreadedCompositor::isActive() const
@@ -329,6 +351,7 @@ void ThreadedCompositor::renderLayerTree()
     bool shouldNotifiyDidComposite = false;
     {
         Locker locker { m_state.lock };
+        m_state.isRenderTimerActive = false;
         reasons = std::exchange(m_state.reasons, { });
         if (reasons.contains(CompositionReason::RenderingUpdate)) {
             if (m_state.isWaitingForTiles) {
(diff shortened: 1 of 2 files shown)


Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/ThreadedCompositor.cpp
```diff
@@ -164,8 +186,8 @@ void ThreadedCompositor::resume()
         return;
 
     Locker locker { m_state.lock };
-    if (m_state.state == State::Scheduled)
-        m_renderTimer.startOneShot(0_s);
+    if (m_state.state == State::Scheduled && !isOnlyRenderingUpdatePendingAndWaitingForTiles())
```
nikolaszimmermann, line 189: Not sure I understand this change specifically. Can you elaborate here?

nikolaszimmermann: I guess this is a general correctness fix, but not necessary to fix the debug assertion and the compose-twice-per-frame issue?

Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/ThreadedCompositor.cpp
```diff
@@ -329,6 +351,7 @@ void ThreadedCompositor::renderLayerTree()
     bool shouldNotifiyDidComposite = false;
     {
         Locker locker { m_state.lock };
+        m_state.isRenderTimerActive = false;
```
nikolaszimmermann, line 354: Shouldn't we assert that this matches the actual state of the timer here, as well?

Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/ThreadedCompositor.cpp
carlosgcampos: While working on this I realized I was not handling this case, when resuming we always start the timer is we were in Scheduled state, but we might still be waiting for tiles.

Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/ThreadedCompositor.cpp
carlosgcampos: What do you mean? adding an assert to check m_state.isRenderTimerActive is true at this point? Or checking that the glib timer is not active? If it's the latter, we always set ready time to -1 in disaptch before calling the callback, so fo all timers in webkit isActive is always false inside the fired function.

nikolaszimmermann: I meant the latter. Ok, so you think no further assertion is necessary/helpful here.

nikolaszimmermann (approved):
Let's please the bots :-)

carlosgcampos:
> Let's please the bots :-)

Hopefully...

PR #55697, opened 2025-12-19
[WPE] Implement damage propagation in non-composited mode

Scony:
#### 816db0427223d2f98c26f8523c5c97ae3ccce0c0
<pre>
[WPE] Implement damage propagation in non-composited mode
<a href="https://bugs.webkit.org/show_bug.cgi?id=304482">https://bugs.webkit.org/show_bug.cgi?id=304482</a>

This change implements damage propagation in the non-composited page
rendering mode.

With this change, depending on the run time preferences, the damage
information is:
- not propagated at all,
- propagated to the platform only,
- propagated and used to reduce the amount of painting done to render
  target.

This change also adds a test case for the above and does a small
rename (frameDamageSinceLastUse() -&gt; renderTargetDamage()) in the
accelerated surface to better reflect the reality.

Canonical link: <a href="https://commits.webkit.org/305510@main">https://commits.webkit.org/305510@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/3d5929d1ba0e2729acd06460ce4f205b47b02d11

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/138605 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/10970 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/86 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/146718 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/91580 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/11674 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/11126 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/106049 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/77388 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/141552 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/8787 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/124237 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/86920 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/8375 "Passed tests") | [❌ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/6138 "Found 1 new API test failure: TestWebKitAPI.NowPlayingControlsTests.NowPlayingUpdatesThrottled (failure)") | [✅ 🛠 wpe-cairo-libwebrtc](https://ews-build.webkit.org/#/builders/166/builds/7009 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/117792 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/76 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/149470 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/10652 "Built successfully") | [❌ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/72 "Found 1 new test failure: http/tests/performance/performance-resource-timing-redirection-cross-origin-media.html (failure)") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/114434 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/10669 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/9016 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/114775 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/8588 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/120533 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/65546 "Built successfully") | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/21351 "Built successfully and passed tests") | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/10701 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/75 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/10436 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/74333 "Built successfully") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/10639 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/10490 "Built successfully") | | |


Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/NonCompositedFrameRenderer.cpp
```diff
@@ -71,6 +74,28 @@ NonCompositedFrameRenderer::~NonCompositedFrameRenderer()
...
+#else
+    UNUSED_PARAM(rect);
+#endif
+}
+
+#if ENABLE(DAMAGE_TRACKING)
+void NonCompositedFrameRenderer::resetFrameDamage()
+{
+    Ref webPage = m_webPage.get();
+    if (webPage->corePage()->settings().propagateDamagingInformation()) {
```
spenap:
```suggestion
    if (Ref webPage = m_webPage.get(); webPage->corePage()->settings().propagateDamagingInformation()) {
```

There is a chance that the promotion from WeakPtr to Ref fails, so it should be checked (or do `if (webPage && ...)` if you feel that's cleaner as the line is already long)

Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/DrawingAreaCoordinatedGraphics.cpp
```diff
@@ -100,19 +100,21 @@ void DrawingAreaCoordinatedGraphics::setNeedsDisplayInRect(const IntRect& rect)
...
+    IntRect dirtyRect = rect;
+    dirtyRect.intersect(m_webPage->bounds());
+    if (dirtyRect.isEmpty())
+        return;
+
 #if PLATFORM(WPE)
-    else if (m_nonCompositedFrameRenderer) {
+    if (m_nonCompositedFrameRenderer) {
         ASSERT(m_dirtyRegion.isEmpty());
+        m_nonCompositedFrameRenderer->addDirtyRect(dirtyRect);
```
carlosgcampos: I would call this setNeedsDisplayInRect for consistency

Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/NonCompositedFrameRenderer.cpp
```diff
@@ -71,6 +74,28 @@ NonCompositedFrameRenderer::~NonCompositedFrameRenderer()
...
+        m_frameDamage = webPage->corePage()->settings().unifyDamagedRegions()
+            ? std::make_optional<Damage>(webPage->bounds(), Damage::Mode::BoundingBox)
+            : std::make_optional<Damage>(webPage->bounds(), Damage::Mode::Rectangles);
```
carlosgcampos: I think this is easier to read as a single make optional and only make conditional the mode parameter

Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/NonCompositedFrameRenderer.cpp
```diff
@@ -87,16 +112,50 @@ void NonCompositedFrameRenderer::display()
...
-    // FIXME: Use render target damage not to re-paint whole rect.
-    webPage->drawRect(*graphicsContext, webPage->bounds());
+    auto rectToRepaint = webPage->bounds();
+#if ENABLE(DAMAGE_TRACKING)
+    if (webPage->corePage()->settings().useDamagingInformationForCompositing()) {
+        if (auto& renderTargetDamage = m_surface->renderTargetDamage())
+            rectToRepaint = renderTargetDamage->bounds();
+    }
+#endif
+    webPage->drawRect(*graphicsContext, rectToRepaint);
```
carlosgcampos, line 131: I'm not sure we should make this conditional. Propagating damage to the UI process depends on damage tracking, but in non-composited mode we have always rendered only the damage region, so I think we should do the same here. Can't we just use the dirty region here instead of the damage from the surface?

Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/NonCompositedFrameRenderer.cpp
Scony: `WeakRef`'s `get()` has a release assert and the assumption here is that web page `Ref` will always exist so IMO there's no need to do extra checks here.

Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/NonCompositedFrameRenderer.cpp
Scony: Well, it's true that old non-composited mode always renders only damaged region (using frame damage), however, it's possible as it creates a fresh buffer (`ShareableBitmap`) from scratch (within `Update` message) every single frame. When this `Update` arrives to UI process, it's then combined with current frame contents and it works.

With the new approach that utilizes render targets, the rendering code here is fundamentally different as it updates frames in-place instead of just preparing deltas. Therefore, it cannot just re-render the contents based on current damage. It would work only with a single render target. But as long as more than 1 render target is used, re-rendering of render target requires a history of changes (render target damage aka surface damage).

As for making this unconditional, IMO we should make it aligned with what compositor-based counterpart does i.e. having this behind  `ENABLE(DAMAGE_TRACKING)` and runtime preferences to control propagation/usage for frame composition.

carlosgcampos (approved):

PR #55860, opened 2025-12-25
[bmalloc] Replace old bmalloc with mimalloc

Constellation:
#### 6f5be90dea47478e24bea95498d2a334892ddded
<pre>
[bmalloc] Replace old bmalloc with mimalloc
<a href="https://bugs.webkit.org/show_bug.cgi?id=304685">https://bugs.webkit.org/show_bug.cgi?id=304685</a>
<a href="https://rdar.apple.com/167162535">rdar://167162535</a>

Reviewed by Marcus Plutowski.

We remove old bmalloc code. This is no longer used in Apple platforms
and we are no longer planning to support it. All Apple platforms are
moving to libpas, and we would like to ensure that bmalloc is always
enabled on all platforms in WebKit. Thus, this patch,

1. Integrating mimalloc as a backend of bmalloc for 32bit, Windows etc.
2. Always building and enabling bmalloc on all WebKit builds. bmalloc
   acts as a lower level library under WTF on all platforms.
3. USE(SYSTEM_MALLOC) should just change the backend of bmalloc to
   system malloc instead of disabling bmalloc builds.

mimalloc is well supported and maintained in many platforms, while it
lacks many features (1) JIT executable memory management, (2)
Gigacage-level large heap management, and (3) MTE. But most of these
lack are the same to the old bmalloc. Basically after this change,

1. All Apple platforms use libpas. Effectively we drop support of old bmalloc.
2. ARM64 / x64 Linux (at least JSCOnly) use libpas. This is great for fuzzer etc.
3. The rest of the platforms use mimalloc as a backend.

Tests: Tools/TestWebKitAPI/CMakeLists.txt
       Tools/TestWebKitAPI/glib/WPEPlatform/CMakeLists.txt

Canonical link: <a href="https://commits.webkit.org/305759@main">https://commits.webkit.org/305759@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/189a2d2252ab40dfe74d92f7dc4f7f13402a8aae

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/139172 "21696 style errors") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/11545 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/669 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/147299 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/92239 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/9b58be6a-0dd4-468f-a301-173c41201ad2) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/141044 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/12252 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/11696 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/106538 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/77593 "1 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/28c22503-7bec-4f52-b675-c5fd8e037bd9) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/142119 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/9264 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/124652 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/87405 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/98f68b3e-6df8-4040-8f0e-c9fee0a20826) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/8810 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/6580 "Passed tests") | [✅ 🛠 wpe-cairo-libwebrtc](https://ews-build.webkit.org/#/builders/166/builds/7592 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/131145 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/118263 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/554 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/150078 "Built successfully") | | 
| [⏳ 🛠 🧪 jsc-debug-arm64 ](https://ews-build.webkit.org/#/builders/JSC-Tests-O3-Debug-arm64-EWS "Waiting in queue, processing has not started yet") | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/11230 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/564 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/114927 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/11243 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/9496 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/115240 "Passed tests") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/9209 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/121002 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/66132 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/21491 "Built successfully and passed tests") | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/11273 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/529 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/170443 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/11008 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/74930 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/44376 "Passed tests") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/11211 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/11060 "Built successfully") | | | |


Constellation:
@Jarred-Sumner
Thanks! I think probably we would like to use v3 as it offers better APIs. And I've checked the code,

https://github.com/microsoft/mimalloc/issues/1055
https://github.com/microsoft/mimalloc/issues/541

And it seems that this is happening when we explicitly destroy `mi_heap_t`. And in our use case, we are not doing so: mi_heap_t is allocated per thread, and we are letting mi_malloc destroy it when thread dies instead of explicitly calling `mi_heap_destroy`.

Do you have an idea whether it is still happening?

Source/bmalloc/bmalloc/AvailableMemory.cpp
Achierius, line 1: Is the plan to remove this file (and others) later on? I understand why we might not want to do that right-right away, as it's more work on top of an already large patch, but I do think it would behoove us to find a way to tear out / migrate these remaining pieces in the near future so that we aren't left with a strange bmalloc-appendix layer in the codebase.

Achierius, line 152: For the remaining functions in this file -- is it possible to move them to WTF? Does anything still depend on them that doesn't depend on that library?

Source/bmalloc/bmalloc/Environment.cpp
```diff
@@ -186,6 +199,8 @@ bool Environment::computeIsSystemHeapEnabled()
 
 bool Environment::computeShouldBmallocAllocateThroughSystemHeap()
 {
+    if (!isLibpasOrMimallocBuilt())
+        return true;
```
Achierius, line 203: This seems incorrect: it would force libpas to _always_ allocate through the system heap, effectively disabling libpas (except as an API wrapper, and for the JIT heap &c)

Source/bmalloc/bmalloc/Environment.cpp
```diff
@@ -148,10 +148,23 @@ static bool isRunningOnValgrind()
...
+#else
+    return true;
+#endif
+}
+
 static bool isSystemHeapForceDisabled()
 {
+#if BUSE(SYSTEM_MALLOC)
+    return false;
+#else
```
Achierius, line 164: Why add this check? The idea with the WebKitSystemHeapOverrideEnablement envvar is to be able to force off the system malloc setting without a rebuild. When that's not set, this function will return false regardless.

Source/bmalloc/bmalloc/bmalloc.cpp
```diff
@@ -59,6 +59,24 @@ pas_primitive_heap_ref gigacageHeaps[static_cast<size_t>(Gigacage::NumberOfKinds
 };
 #endif
 
+#if BUSE(MIMALLOC)
+#if GIGACAGE_ENABLED
```
Achierius: Should this ever happen? I thought the plan was to disable gigacage when mimalloc was on.

Achierius (commented):
What changes (if any) had to be made to the mimalloc code itself for compatibility?

Source/bmalloc/bmalloc/Environment.cpp
Constellation: Is it correct? Currently `isLibpasOrMimallocBuilt` returns false only when `BUSE_SYSTEM_MALLOC` is set. And this is set only when we always want to use system malloc.

Source/bmalloc/bmalloc/Environment.cpp
Constellation: When `BUSE_SYSTEM_MALLOC` is enabled, this basically means that libpas / mimalloc are not built at all and only system malloc is available.

Source/bmalloc/bmalloc/bmalloc.cpp
Constellation: I attempted to enable it, but failed because mimalloc has strict limit for arena size (32GB etc...?), and it didn't work for our Gigacage size. Removing.

Source/bmalloc/bmalloc/AvailableMemory.cpp
Constellation: Yeah, that's good. Currently, we are just moving it to bmalloc, but eventually we should factor out what part is bmalloc related and what part is WTF. And then we should clean up.

Constellation: Yeah, I think we can just move them to WTF. Currently, this change is not doing so, but it would be great to do it subsequently.

Source/bmalloc/bmalloc/Environment.cpp
Achierius: It seems preferable to fail badly if WebKitSystemHeapOverrideEnablement=1 is set then, since otherwise consumers might silently get different behavior than expected -- maybe a RELEASE_ASSERT that either the envvar is not set, or that if it is set it =0?

Source/bmalloc/bmalloc/Environment.cpp
Achierius: I see, "USE_SYSTEM_MALLOC" is really "use _only_ system malloc". A comment where it's defined could help clarify that

Source/bmalloc/bmalloc/bmalloc.h
```diff
@@ -60,16 +60,29 @@ inline pas_primitive_heap_ref& heapForKind(Gigacage::Kind kind)
...
 {
 #if BUSE(LIBPAS)
     if (!isGigacage(kind))
         return bmalloc_try_allocate_inline(size, asPasAllocationMode(mode));
     return bmalloc_try_allocate_auxiliary_inline(&heapForKind(gigacageKind(kind)), size, asPasAllocationMode(mode));
+#elif BUSE(MIMALLOC)
+    BUNUSED(mode);
+    if (!isGigacage(kind))
+        return mi_malloc(size);
+    return mi_heap_malloc(heapForKind(gigacageKind(kind)), size);
```
Achierius: Gigacage code, should also be removed / replaced with an assert?

Source/bmalloc/bmalloc/bmalloc.h
```diff
@@ -203,9 +282,21 @@ BINLINE void* realloc(void* object, size_t newSize, CompactAllocationMode mode,
         return bmalloc_reallocate_inline(object, newSize, asPasAllocationMode(mode), pas_reallocate_free_if_successful);
     return bmalloc_reallocate_auxiliary_inline(
         object, &heapForKind(gigacageKind(kind)), newSize, asPasAllocationMode(mode), pas_reallocate_free_if_successful);
+#elif BUSE(MIMALLOC)
+    BUNUSED(mode);
+    void* memory = nullptr;
+    if (!isGigacage(kind))
+        memory = mi_realloc(object, newSize);
+    else
+        memory = mi_heap_realloc(heapForKind(gigacageKind(kind)), object, newSize);
```
Achierius: Gigacage code

Source/bmalloc/bmalloc/bmalloc.h
```diff
@@ -189,9 +262,15 @@ BINLINE void* tryRealloc(void* object, size_t newSize, CompactAllocationMode mod
     }
     return bmalloc_try_reallocate_auxiliary_inline(
         object, &heapForKind(gigacageKind(kind)), newSize, asPasAllocationMode(mode), pas_reallocate_free_if_successful);
+#elif BUSE(MIMALLOC)
+    BUNUSED(mode);
+    if (!isGigacage(kind))
+        return mi_realloc(object, newSize);
+    return mi_heap_realloc(heapForKind(gigacageKind(kind)), object, newSize);
```
Achierius: Gigacage code

Source/bmalloc/bmalloc/bmalloc.h
```diff
@@ -108,11 +136,21 @@ BINLINE void* zeroedMalloc(size_t size, CompactAllocationMode mode, HeapKind kin
     if (!isGigacage(kind))
         return bmalloc_allocate_zeroed_inline(size, asPasAllocationMode(mode));
     return bmalloc_allocate_auxiliary_zeroed_inline(&heapForKind(gigacageKind(kind)), size, asPasAllocationMode(mode));
+#elif BUSE(MIMALLOC)
+    BUNUSED(mode);
+    void* memory = nullptr;
+    if (!isGigacage(kind))
+        memory = mi_zalloc(size);
+    else
+        memory = mi_heap_zalloc(heapForKind(gigacageKind(kind)), size);
```
Achierius: Gigacage

Source/bmalloc/bmalloc/bmalloc.h
```diff
@@ -92,12 +117,15 @@ BINLINE void* tryZeroedMalloc(size_t size, CompactAllocationMode mode, HeapKind
     if (!isGigacage(kind))
         return bmalloc_try_allocate_zeroed_inline(size, asPasAllocationMode(mode));
     return bmalloc_try_allocate_auxiliary_zeroed_inline(&heapForKind(gigacageKind(kind)), size, asPasAllocationMode(mode));
+#elif BUSE(MIMALLOC)
+    BUNUSED(mode);
+    if (!isGigacage(kind))
+        return mi_zalloc(size);
+    return mi_heap_zalloc(heapForKind(gigacageKind(kind)), size);
```
Achierius: Gigacage

Source/bmalloc/bmalloc/Gigacage.cpp
```diff
@@ -145,6 +145,10 @@ void ensureGigacage()
             vmDeallocatePhysicalPages(base, totalSize);
 
             size_t nextCage = 0;
+            size_t sliceSize = vmPageSize();
+#if BUSE(MIMALLOC)
+            sliceSize = std::max(mi_arena_min_alignment(), sliceSize);
+#endif
```
Achierius: Mimalloc code not needed now that it won't support the gigacage?

Source/bmalloc/bmalloc/bmalloc.cpp
```diff
@@ -137,28 +117,22 @@ void scavengeThisThread()
...
     pas_scavenger_run_synchronously_now();
-#endif
     scavengeThisThread();
-    if (SystemHeap* systemHeap = SystemHeap::tryGetIfShouldSupplantBmalloc())
-        systemHeap->scavenge();
-    else {
-#if !BUSE(LIBPAS)
-        Scavenger::get()->scavenge();
+#elif BUSE(MIMALLOC)
+    mi_collect(/* force */ true);
```
Achierius, line 131: From my own reading of mimalloc (https://github.com/microsoft/mimalloc/blob/dev/src/heap.c `mi_heap_collect_ex`) it seems like this only affects the current thread -- it doesn't actually purge/force collection for pages owned by other threads. The only way to achieve something like what the scavenger does would be to call it on every thread, but that seems intrusive enough that maybe it's not worth doing?

Achierius (approved):
r=me

PR #55874, opened 2025-12-26
SVGPathElement.getTotalLength() should respect CSS 'd' property for display:none elements

Ahmad-S792:
#### fb740fd4160c7a929a26ef4d16302a441f43919c
<pre>
SVGPathElement.getTotalLength() should respect CSS &apos;d&apos; property for display:none elements
<a href="https://bugs.webkit.org/show_bug.cgi?id=304701">https://bugs.webkit.org/show_bug.cgi?id=304701</a>
<a href="https://rdar.apple.com/167195297">rdar://167195297</a>

Reviewed by NOBODY (OOPS!).

Per the SVG specification [1], getTotalLength() should compute the path length
based on the effective path data, which includes CSS &apos;d&apos; property values.
Currently, display:none elements fail to read the CSS &apos;d&apos; property because
pathByteStream() only checks renderer-&gt;style(), which is null for non-rendered
elements.

Fix by checking computedStyle() first before falling back to the renderer,
allowing display:none elements to access their CSS &apos;d&apos; property values.

[1] <a href="https://svgwg.org/svg2-draft/paths.html#InterfaceSVGGeometryElement">https://svgwg.org/svg2-draft/paths.html#InterfaceSVGGeometryElement</a>

* LayoutTests/imported/w3c/web-platform-tests/svg/types/scripted/SVGGeometryElement.getPointAtLength-04-expected.txt: Progression
* LayoutTests/imported/w3c/web-platform-tests/svg/types/scripted/SVGPathElement.getTotalLength-01-expected.txt: Ditto
(WebCore::SVGPathElement::pathByteStream const):
</pre>

### DIFF
--- a/Source/WebCore/svg/SVGPathElement.cpp
+++ b/Source/WebCore/svg/SVGPathElement.cpp
@@ -1,7 +1,7 @@
 /*
  * Copyright (C) 2004, 2005, 2006, 2008 Nikolas Zimmermann <zimmermann@kde.org>
  * Copyright (C) 2004, 2005, 2006, 2007 Rob Buis <buis@kde.org>
- * Copyright (C) 2018-2024 Apple Inc. All rights reserved.
+ * Copyright (C) 2018-2025 Apple Inc. All rights reserved.
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Library General Public
@@ -238,6 +238,14 @@ RenderPtr<RenderElement> SVGPathElement::createElementRenderer(RenderStyle&& sty
 const SVGPathByteStream& SVGPathElement::pathByteStream() const
 {
     if (document().settings().cssDPropertyEnabled()) {
+        // Try computed style first (works for display:none elements).
+        auto path = const_cast<SVGPathElement*>(this);
+        if (auto* style = path->computedStyle()) {
+            if (auto& pathFunction = style->d().tryPath())
+                return pathFunction->parameters.data.byteStream;
+        }
+
+        // Fallback to renderer style, if no computed style.
         if (CheckedPtr renderer = this->renderer()) {
             if (auto& pathFunction = renderer->style().d().tryPath())
                 return pathFunction->parameters.data.byteStream;
--- a/LayoutTests/imported/w3c/web-platform-tests/svg/types/scripted/SVGGeometryElement.getPointAtLength-04-expected.txt
+++ b/LayoutTests/imported/w3c/web-platform-tests/svg/types/scripted/SVGGeometryElement.getPointAtLength-04-expected.txt
@@ -1,7 +1,7 @@
 
 PASS SVGGeometryElement.getPointAtLength: 'display' and a valid path, path with display: default
 PASS SVGGeometryElement.getPointAtLength: 'display' and a valid path, path with display: none
-FAIL SVGGeometryElement.getPointAtLength: 'display' and a valid path, path with display: none and inline style The element's path is empty.
+PASS SVGGeometryElement.getPointAtLength: 'display' and a valid path, path with display: none and inline style
 PASS SVGGeometryElement.getPointAtLength: 'display' and a valid path, rect with display: default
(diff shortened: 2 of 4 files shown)


Source/WebCore/svg/SVGPathElement.cpp
```diff
@@ -238,6 +238,17 @@ RenderPtr<RenderElement> SVGPathElement::createElementRenderer(RenderStyle&& sty
 const SVGPathByteStream& SVGPathElement::pathByteStream() const
 {
     if (document().settings().cssDPropertyEnabled()) {
+        // Ensure style is computed before accessing it.
+        protectedDocument()->updateStyleIfNeeded();
```
annevk: Why would we do this even when there is a renderer?

Ahmad-S792: Without this - when locally I ran `svg` test suite (both internal + wpt) that we were crashing few test cases, kind of like getting stale styles or from detached element. So I added it to ensure we update style before accessing them.

annevk: But do we also need to update it when we have a renderer? Or only in the display:none case?

Ahmad-S792: Only in `display:none` case since we had the code with renderer without this update style without issue. Only with display:none case - where we don't have renderer and try to use `computedStyle`, it needs to be ensure that it is up-to-date, so we don't crash or get stale style.

annevk: But now you are also computing it in the other branch. Do you see the issue?

Ahmad-S792: I don't see any issue, why this should be an issue for renderer case?

annevk: Because we’re doing work that’s not needed?

nullhook: @Ahmad-S792 you should only reflow when there's no renderer

nit: also, do the fast path first


```c++
if (CheckedPtr renderer = this->renderer()) {
  if (auto& pathFunction = renderer->style().d().tryPath())
     return pathFunction->parameters.data.byteStream;
  return SVGPathByteStream::empty();
}

protectedDocument()->updateStyleIfNeeded();
if (auto* style = computedStyle()) {
   if (auto& pathFunction = style->d().tryPath())
      return pathFunction->parameters.data.byteStream;
}
```

nullhook: just curious, calling `computedStyle()` should typically reflow implicitly. why do need an explicit `updateStyleIfNeeded` here?

> we were crashing few test cases, kind of like getting stale styles or from detached element.

did you get crash/test failures without `updateStyleIfNeeded()` ?

Ahmad-S792: Yes - I got crashes without `updateStyleIfNeeded()` - hence I added it.

Source/WebCore/svg/SVGPathElement.cpp
```diff
@@ -238,6 +238,14 @@ RenderPtr<RenderElement> SVGPathElement::createElementRenderer(RenderStyle&& sty
 const SVGPathByteStream& SVGPathElement::pathByteStream() const
 {
     if (document().settings().cssDPropertyEnabled()) {
+        // Try computed style first (works for display:none elements).
```
nikolaszimmermann, line 241: In what situations are we still asking the renderer for it's style now? Shouldn't the computed style always be available? I'm confused, and maybe misunderstood something...

Ahmad-S792: @nikolaszimmermann - My understanding, when we don't have renderer display: none plus some display:cotents cases, we have to get computed style to give information - this is what the test is expecting.

Source/WebCore/svg/SVGPathElement.cpp
annevk: Let me rephrase: check the renderer first. Add a fallback for computed style.

PR #55886, opened 2025-12-27
Refactor self-alignment property resolution to manage 'auto' instead of 'normal'

fantasai:
#### fd2edbac21f5950d58d0720afc0471640061a6e7
<pre>
Refactor self-alignment property resolution to manage &apos;auto&apos; instead of &apos;normal&apos;
<a href="https://bugs.webkit.org/show_bug.cgi?id=304716">https://bugs.webkit.org/show_bug.cgi?id=304716</a>
<a href="https://rdar.apple.com/167221674">rdar://167221674</a>

Reviewed by Sammy Gill.

The way &apos;normal&apos; resolution is looped through the style system ties into our
layout code in an awkward way. Instead, loop &apos;auto&apos; resolution through the
style system, and make layout code handle &apos;normal&apos; (which is much more
tied to formatting context subtleties) directly. This creates more
straightforward logic in both systems, divides the work along a more
appropriate boundary, and also avoids doing &apos;normal&apos; resolution work when
we don&apos;t need it.

This is Part 1 of a refactoring stack for grid-lanes item sizing (bug 304715).

* Source/WebCore/layout/formattingContexts/flex/FlexLayout.cpp:
(WebCore::Layout::FlexLayout::crossSizeForFlexLines const):
(WebCore::Layout::FlexLayout::computeCrossSizeForFlexItems const):
(WebCore::Layout::FlexLayout::handleCrossAxisAlignmentForFlexItems const):
* Source/WebCore/rendering/PositionedLayoutConstraints.cpp:
(WebCore::PositionedLayoutConstraints::remainingSpaceForStaticAlignment const):
(WebCore::PositionedLayoutConstraints::resolveAlignmentValue const):
* Source/WebCore/rendering/RenderFlexibleBox.cpp:
(WebCore::RenderFlexibleBox::styleDidChange):
(WebCore::RenderFlexibleBox::overflowAlignmentForFlexItem const):
(WebCore::RenderFlexibleBox::alignmentForFlexItem const):
* Source/WebCore/rendering/RenderGrid.cpp:
(WebCore::RenderGrid::styleDidChange):
(WebCore::RenderGrid::columnAxisPositionForGridItem const):
(WebCore::RenderGrid::alignSelfForGridItem const): Deleted.
(WebCore::RenderGrid::justifySelfForGridItem const): Deleted.
* Source/WebCore/rendering/style/StyleSelfAlignmentData.h:
(WebCore::StyleSelfAlignmentData::isNormal const):
[...]

PR #55887, opened 2025-12-27
Untangle grid self-alignment code

fantasai:
#### 810cb2deb1447536eab4490c6cde79ebd227a04e
<pre>
Untangle grid self-alignment code
<a href="https://bugs.webkit.org/show_bug.cgi?id=304717">https://bugs.webkit.org/show_bug.cgi?id=304717</a>
<a href="https://rdar.apple.com/167221679">rdar://167221679</a>

Reviewed by Sammy Gill.

This patch untangles the self-alignment code for grid containers by:
- Consolidating self-alignment getters into a single method, so that each
    place where it&apos;s needed uses consistent logic and special cases are
    always handled instead of sometimes forgotten.
- Redesigning the API for checking stretch alignment so that we encapsulate
    logic that varies per formatting context within the formatting context
    instead of making RenderBox contain logical specific to its subclasses.

This is Part 2 of a refactoring stack for grid-lanes item sizing (bug 304715).

* Source/WebCore/rendering/GridLayoutFunctions.cpp:
(WebCore::GridLayoutFunctions::hasStretchableSizeInColumnAxis):
(WebCore::GridLayoutFunctions::hasStretchableSizeInRowAxis):
(WebCore::GridLayoutFunctions::hasAutoSizeInColumnAxis): Deleted.
(WebCore::GridLayoutFunctions::hasAutoSizeInRowAxis): Deleted.
(WebCore::GridLayoutFunctions::allowedToStretchGridItemAlongColumnAxis): Deleted.
(WebCore::GridLayoutFunctions::allowedToStretchGridItemAlongRowAxis): Deleted.
* Source/WebCore/rendering/GridLayoutFunctions.h:
* Source/WebCore/rendering/GridTrackSizingAlgorithm.cpp:
(WebCore::GridTrackSizingAlgorithm::updateBaselineAlignmentContext):
(WebCore::GridTrackSizingAlgorithm::baselineOffsetForGridItem const):
(WebCore::RenderBlock::willStretchItem const):
* Source/WebCore/rendering/RenderBlock.h:
* Source/WebCore/rendering/RenderBox.cpp:
(WebCore::RenderBox::isStretchingColumnFlexItem const):
(WebCore::RenderBox::sizesPreferredLogicalWidthToFitContent const):
(WebCore::RenderBox::shouldComputeLogicalHeightFromAspectRatio const):
(WebCore::RenderBox::columnFlexItemHasStretchAlignment const): Deleted.
(WebCore::RenderBox::hasStretchedLogicalHeight const): Deleted.
(WebCore::RenderBox::hasStretchedLogicalWidth const): Deleted.
* Source/WebCore/rendering/RenderBox.h:
(WebCore::RenderBox::selfAlignmentNormalBehavior const): Deleted.
* Source/WebCore/rendering/RenderBoxInlines.h:
(WebCore::RenderBox::hasStretchedLogicalHeight const):
(WebCore::RenderBox::hasStretchedLogicalWidth const):
* Source/WebCore/rendering/RenderFlexibleBox.cpp:
(WebCore::RenderFlexibleBox::willStretchItem const):
* Source/WebCore/rendering/RenderFlexibleBox.h:
* Source/WebCore/rendering/RenderGrid.cpp:
(WebCore::RenderGrid::selfAlignmentForGridItem const):
(WebCore::RenderGrid::selfAlignmentChangedToStretch const):
(WebCore::RenderGrid::selfAlignmentChangedFromStretch const):
(WebCore::RenderGrid::styleDidChange):
(WebCore::RenderGrid::canSetColumnAxisStretchRequirementForItem const):
(WebCore::RenderGrid::computeLayoutRequirementsForItemsBeforeLayout const):
(WebCore::RenderGrid::willStretchItem const):
(WebCore::RenderGrid::aspectRatioPrefersInline):
(WebCore::RenderGrid::applyStretchAlignmentToGridItemIfNeeded):
(WebCore::RenderGrid::isBaselineAlignmentForGridItem const):
(WebCore::RenderGrid::baselineGridItem const):
(WebCore::RenderGrid::columnAxisPositionForGridItem const):
(WebCore::RenderGrid::rowAxisPositionForGridItem const):
(WebCore::RenderGrid::columnAxisOffsetForGridItem const):
(WebCore::RenderGrid::rowAxisOffsetForGridItem const):
(WebCore::RenderGrid::alignSelfForGridItem const): Deleted.
(WebCore::RenderGrid::justifySelfForGridItem const): Deleted.
* Source/WebCore/rendering/RenderGrid.h:
* Source/WebCore/style/values/grid/StyleGridTrackSizingDirection.h:
(WebCore::Style::logicalAxis):
(WebCore::Style::gridTrackSizingDirection):

Canonical link: <a href="https://commits.webkit.org/305260@main">https://commits.webkit.org/305260@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/64310dd5925c30e01f1b70559b0120aaff2f6e1e

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/137927 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/10291 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/138/builds/49285 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/145994 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/90902 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/f2c6d2ed-bd8b-4d70-a499-64172ae81bd0/1d4f7979-e5fa-48ac-9969-aff6140fedb7) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/139799 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/10993 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/10432 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/105484 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/76976 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/f2c6d2ed-bd8b-4d70-a499-64172ae81bd0/ee875b77-86b2-489d-8984-5ea0af0a300a) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/140872 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/8191 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/123646 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/86335 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/f2c6d2ed-bd8b-4d70-a499-64172ae81bd0/2a7a50b9-3c59-47a1-828f-f43b3e300ae8) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/7808 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/5559 "Passed tests") | [✅ 🛠 wpe-cairo-libwebrtc](https://ews-build.webkit.org/#/builders/166/builds/6276 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/117202 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/135/builds/41813 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/148704 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/9974 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/136/builds/42372 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/113885 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/9991 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/8416 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/114215 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/29014 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/7749 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/119899 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/64698 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/10020 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/137/builds/37901 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/9750 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/73588 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/9961 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/9812 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/rendering/RenderBlock.h
+++ b/Source/WebCore/rendering/RenderBlock.h
@@ -93,6 +93,7 @@ public:
     static bool hasPercentHeightContainerMap();
     static void clearPercentHeightDescendantsFrom(RenderBox&);
 
+    virtual bool willStretchItem(const RenderBox& item, LogicalBoxAxis containingAxis, StretchingMode = StretchingMode::Normal) const;
     bool isContainingBlockAncestorFor(RenderObject&) const;
 
     void setHasMarginBeforeQuirk(bool b) { setRenderBlockHasMarginBeforeQuirk(b); }
--- a/Source/WebCore/rendering/RenderFlexibleBox.cpp
+++ b/Source/WebCore/rendering/RenderFlexibleBox.cpp
@@ -2254,6 +2254,23 @@ void RenderFlexibleBox::resetAutoMarginsAndLogicalTopInCrossAxis(RenderBox& flex
     }
 }
 
+bool RenderFlexibleBox::willStretchItem(const RenderBox& item, LogicalBoxAxis containingAxis, StretchingMode mode) const
+{
+    auto physicalAxis = mapAxisLogicalToPhysical(writingMode(), containingAxis);
+    if (isHorizontalFlow() == (BoxAxis::Horizontal == physicalAxis))
+        return false;
+
+    auto& itemStyle = item.style();
+
+    if (!itemStyle.alignSelf().resolve(&style()).isStretchy(mode == StretchingMode::Explicit ? ItemPosition::Normal : ItemPosition::Stretch))
+        return false;
+
+    bool isItemBlockAxis = (LogicalBoxAxis::Block == containingAxis) == !writingMode().isOrthogonal(itemStyle.writingMode());
+    return isItemBlockAxis
+        ? itemStyle.logicalHeight().isAuto() && !itemStyle.marginBefore().isAuto() && !itemStyle.marginAfter().isAuto()
+        : itemStyle.logicalWidth().isAuto() && !itemStyle.marginStart().isAuto() && !itemStyle.marginEnd().isAuto();
+}
+
 bool RenderFlexibleBox::needToStretchFlexItemLogicalHeight(const RenderBox& flexItem) const
 {
     // This function is a little bit magical. It relies on the fact that blocks
--- a/Source/WebCore/rendering/RenderFlexibleBox.h
+++ b/Source/WebCore/rendering/RenderFlexibleBox.h
@@ -62,6 +62,8 @@ public:
     bool hitTestChildren(const HitTestRequest&, HitTestResult&, const HitTestLocation&, const LayoutPoint& adjustedLocation, HitTestAction) override;
     void paintChildren(PaintInfo& forSelf, const LayoutPoint&, PaintInfo& forChild, bool usePrintRect) override;
 
+    bool willStretchItem(const RenderBox& item, LogicalBoxAxis containingAxis, StretchingMode = StretchingMode::Normal) const override;
+
     bool isHorizontalFlow() const;
     Direction crossAxisDirection() const;
 
--- a/Source/WebCore/rendering/RenderGrid.cpp
+++ b/Source/WebCore/rendering/RenderGrid.cpp
@@ -119,21 +119,34 @@ bool RenderGrid::isExtrinsicallySized() const
     return true;
 }
 
-StyleSelfAlignmentData RenderGrid::selfAlignmentForGridItem(Style::GridTrackSizingDirection alignmentContextType, const RenderBox& gridItem, const RenderStyle* gridStyle) const
+StyleSelfAlignmentData RenderGrid::selfAlignmentForGridItem(const RenderBox& gridItem, LogicalBoxAxis containingAxis, StretchingMode stretchingMode, const RenderStyle* gridStyle) const
 {
-    return alignmentContextType == Style::GridTrackSizingDirection::Columns ? justifySelfForGridItem(gridItem, StretchingMode::Any, gridStyle) : alignSelfForGridItem(gridItem, StretchingMode::Any, gridStyle);
+    if (isMasonry(containingAxis))
+        return { ItemPosition::Start };
+
+    if (CheckedPtr renderGrid = dynamicDowncast<RenderGrid>(gridItem); renderGrid && renderGrid->isSubgridInParentDirection(Style::gridTrackSizingDirection(containingAxis)))
+        return { ItemPosition::Stretch, OverflowAlignment::Default };
+
+    if (!gridStyle)
+        gridStyle = &style();
+    auto alignment = containingAxis == LogicalBoxAxis::Inline
+        ? gridItem.style().justifySelf().resolve(gridStyle) : gridItem.style().alignSelf().resolve(gridStyle);
+
+    if (alignment.isNormal() && StretchingMode::Explicit != stretchingMode)
+        alignment.setPosition(gridItem.isRenderReplaced() ? ItemPosition::Start : ItemPosition::Stretch);
+    return alignment;
 }
 
-bool RenderGrid::selfAlignmentChangedToStretch(Style::GridTrackSizingDirection alignmentContextType, const RenderStyle& oldStyle, const RenderStyle& newStyle, const RenderBox& gridItem) const
+bool RenderGrid::selfAlignmentChangedToStretch(LogicalBoxAxis containingAxis, const RenderStyle& oldStyle, const RenderStyle& newStyle, const RenderBox& gridItem) const
 {
-    return selfAlignmentForGridItem(alignmentContextType, gridItem, &oldStyle).position() != ItemPosition::Stretch
-        && selfAlignmentForGridItem(alignmentContextType, gridItem, &newStyle).position() == ItemPosition::Stretch;
+    return selfAlignmentForGridItem(gridItem, containingAxis, StretchingMode::Normal, &oldStyle).position() != ItemPosition::Stretch
+        && selfAlignmentForGridItem(gridItem, containingAxis, StretchingMode::Normal, &newStyle).position() == ItemPosition::Stretch;
 }
 
-bool RenderGrid::selfAlignmentChangedFromStretch(Style::GridTrackSizingDirection alignmentContextType, const RenderStyle& oldStyle, const RenderStyle& newStyle, const RenderBox& gridItem) const
+bool RenderGrid::selfAlignmentChangedFromStretch(LogicalBoxAxis containingAxis, const RenderStyle& oldStyle, const RenderStyle& newStyle, const RenderBox& gridItem) const
 {
-    return selfAlignmentForGridItem(alignmentContextType, gridItem, &oldStyle).position() == ItemPosition::Stretch
-        && selfAlignmentForGridItem(alignmentContextType, gridItem, &newStyle).position() != ItemPosition::Stretch;
+    return selfAlignmentForGridItem(gridItem, containingAxis, StretchingMode::Normal, &oldStyle).position() == ItemPosition::Stretch
+        && selfAlignmentForGridItem(gridItem, containingAxis, StretchingMode::Normal, &newStyle).position() != ItemPosition::Stretch;
 }
 
 void RenderGrid::styleDidChange(Style::Difference diff, const RenderStyle* oldStyle)
@@ -176,10 +189,10 @@ void RenderGrid::styleDidChange(Style::Difference diff, const RenderStyle* oldSt
             if (gridItem.isOutOfFlowPositioned())
                 continue;
 
-            if (selfAlignmentChangedToStretch(Style::GridTrackSizingDirection::Columns, *oldStyle, newStyle, gridItem)
-                || selfAlignmentChangedFromStretch(Style::GridTrackSizingDirection::Columns, *oldStyle, newStyle, gridItem)
-                || selfAlignmentChangedToStretch(Style::GridTrackSizingDirection::Rows, *oldStyle, newStyle, gridItem)
-                || selfAlignmentChangedFromStretch(Style::GridTrackSizingDirection::Rows, *oldStyle, newStyle, gridItem)) {
+            if (selfAlignmentChangedToStretch(LogicalBoxAxis::Block, *oldStyle, newStyle, gridItem)
+                || selfAlignmentChangedFromStretch(LogicalBoxAxis::Block, *oldStyle, newStyle, gridItem)
+                || selfAlignmentChangedToStretch(LogicalBoxAxis::Inline, *oldStyle, newStyle, gridItem)
+                || selfAlignmentChangedFromStretch(LogicalBoxAxis::Inline, *oldStyle, newStyle, gridItem)) {
                 gridItem.setNeedsLayout();
             }
         }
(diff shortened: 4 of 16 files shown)


fantasai:
Test failures are from an error in https://github.com/WebKit/WebKit/pull/55886 (which is now fixed). I'll rebase once that lands.

Source/WebCore/rendering/RenderGrid.cpp
```diff
@@ -1782,39 +1796,25 @@ static LayoutUnit computeOverflowAlignmentOffset(OverflowAlignment overflow, Lay
     return 0;
 }
 
-StyleSelfAlignmentData RenderGrid::alignSelfForGridItem(const RenderBox& gridItem, StretchingMode stretchingMode, const RenderStyle* gridStyle) const
+bool RenderGrid::canStretch(const RenderBox& gridItem, LogicalBoxAxis axis) const
```
sammygill: Do we *need* to have `canStretch` along with `willStretch`? It seems like the former is only called once in the latter. If not we might be able to get away with just putting `canStretch` as a lambda inside of `willStretch` (along with the other helpers on `GridLayoutFunctions` but those were already there). Otherwise it might be a bit confusing when determining when one should be called vs the other. We probably want a comment as a minimum to help with that

Source/WebCore/rendering/RenderFlexibleBox.cpp
```diff
@@ -2254,6 +2254,22 @@ void RenderFlexibleBox::resetAutoMarginsAndLogicalTopInCrossAxis(RenderBox& flex
...
+        return false;
+
+    auto& itemStyle = item.style();
+
+    if (!itemStyle.alignSelf().resolve(&style()).isStretchy(mode == StretchingMode::Explicit ? ItemPosition::Normal : ItemPosition::Stretch))
+        return false;
+
+    return (LogicalBoxAxis::Block == containingAxis) == !writingMode().isOrthogonal(itemStyle.writingMode())
+        ? itemStyle.logicalHeight().isAuto() && !itemStyle.marginBefore().isAuto() && !itemStyle.marginAfter().isAuto()
+        : itemStyle.logicalWidth().isAuto() && !itemStyle.marginStart().isAuto() && !itemStyle.marginEnd().isAuto();
```
sammygill: I feel like this might be slightly difficult to follow along with all of the conditionals that are going on. Could we maybe at least factor out the first condition into a local variable or something along those lines to reduce the overhead a bit?

Source/WebCore/rendering/RenderBlock.h
```diff
@@ -93,6 +93,7 @@ class RenderBlock : public RenderBox {
     static bool hasPercentHeightContainerMap();
     static void clearPercentHeightDescendantsFrom(RenderBox&);
 
+    virtual bool willStretch(const RenderBox& item, LogicalBoxAxis containingAxis, StretchingMode = StretchingMode::Normal) const;
```
sammygill: If three are only a couple of derived classes that are going to override this method then one thing you could consider is to not make it virtual and only define it on those classes. It would add a tad bit of code at the call sites by checking the type and down casting but it would be more explicit

Source/WebCore/rendering/RenderFlexibleBox.h
```diff
@@ -62,6 +62,8 @@ class RenderFlexibleBox : public RenderBlock {
     bool hitTestChildren(const HitTestRequest&, HitTestResult&, const HitTestLocation&, const LayoutPoint& adjustedLocation, HitTestAction) override;
     void paintChildren(PaintInfo& forSelf, const LayoutPoint&, PaintInfo& forChild, bool usePrintRect) override;
 
+    bool willStretch(const RenderBox& item, LogicalBoxAxis containingAxis, StretchingMode = StretchingMode::Normal) const override;
```
sammygill: Thoughts on `willStretchItem`?

Source/WebCore/rendering/RenderGrid.cpp
```diff
@@ -119,21 +119,35 @@ bool RenderGrid::isExtrinsicallySized() const
...
-StyleSelfAlignmentData RenderGrid::selfAlignmentForGridItem(Style::GridTrackSizingDirection alignmentContextType, const RenderBox& gridItem, const RenderStyle* gridStyle) const
+StyleSelfAlignmentData RenderGrid::selfAlignmentForGridItem(const RenderBox& gridItem, LogicalBoxAxis containingAxis, StretchingMode stretchingMode, const RenderStyle* gridStyle) const
 {
-    return alignmentContextType == Style::GridTrackSizingDirection::Columns ? justifySelfForGridItem(gridItem, StretchingMode::Any, gridStyle) : alignSelfForGridItem(gridItem, StretchingMode::Any, gridStyle);
+    if (isMasonry(containingAxis))
+        return { ItemPosition::Start };
+
+    CheckedPtr renderGrid = dynamicDowncast<RenderGrid>(gridItem);
+    if (renderGrid && renderGrid->isSubgridInParentDirection(Style::gridTrackSizingDirection(containingAxis)))
+        return { ItemPosition::Stretch, OverflowAlignment::Default };
```
sammygill: I think this could technically be the following since I don't think we use `renderGrid` outside the if

```suggestion
    
    if (CheckedPtr renderGrid = dynamicDowncast<RenderGrid>(gridItem); renderGrid->isSubgridInParentDirection(Style::gridTrackSizingDirection(containingAxis)))
        return { ItemPosition::Stretch, OverflowAlignment::Default };
```

Source/WebCore/rendering/RenderGrid.cpp
```diff
@@ -1782,39 +1796,25 @@ static LayoutUnit computeOverflowAlignmentOffset(OverflowAlignment overflow, Lay
...
-    if (renderGrid && renderGrid->isSubgridInParentDirection(Style::GridTrackSizingDirection::Rows))
-        return { ItemPosition::Stretch, OverflowAlignment::Default };
-
-    auto alignment = gridItem.style().alignSelf().resolve(gridStyle ? : &style());
-    if (alignment.isNormal())
-        return { stretchingMode == StretchingMode::Any ? selfAlignmentNormalBehavior(&gridItem) : ItemPosition::Normal };
-    return alignment;
+    return axis == LogicalBoxAxis::Block
+        ? GridLayoutFunctions::hasStretchableSizeInColumnAxis(gridItem, *this) && !GridLayoutFunctions::hasAutoMarginsInColumnAxis(gridItem, writingMode())
+        : GridLayoutFunctions::hasStretchableSizeInRowAxis(gridItem, *this) && !GridLayoutFunctions::hasAutoMarginsInRowAxis(gridItem, writingMode());
```
sammygill: What do you think of

```suggestion
    if (axis == LogicalBoxAxis::Block)
        returnGridLayoutFunctions::hasStretchableSizeInColumnAxis(gridItem, *this) && !GridLayoutFunctions::hasAutoMarginsInColumnAxis(gridItem, writingMode());
        return GridLayoutFunctions::hasStretchableSizeInRowAxis(gridItem, *this) && !GridLayoutFunctions::hasAutoMarginsInRowAxis(gridItem, writingMode());
```

sammygill (approved):

Source/WebCore/rendering/RenderGrid.cpp
fantasai: I don't have a strong preference. For complicated things I agree if/else often reads easier. However because we don't use `else` before `return`, I noticed the ternary has the advantage that the two return values line up visually so you can easily see the differences between them. :) Your call!

Source/WebCore/rendering/RenderGrid.cpp
fantasai: I think it would be 
```
if (CheckedPtr renderGrid = dynamicDowncast<RenderGrid>(gridItem); renderGrid && renderGrid->isSubgridInParentDirection(Style::gridTrackSizingDirection(containingAxis)))
```
right?

I didn't do it because I think it makes the conditional so long it's hard to read; to me that's more important than perfectly scoping the variable. But I can change it if you think scoping is more important.

Source/WebCore/rendering/RenderBlock.h
fantasai: That... sounds pretty awful? Why would we do that? Seems like a recipe for mistakes. And for making the call sites all way harder to read, if each one now has to store a bunch of conditional logic into a variable where right now it just has a single function call.

sammygill: I think it mostly depends on how many different classes will override this method or eventually override this method. I was looking at the code in `RenderBox::hasStretchedLogicalHeight` when I thought of this comment. 

That code would change from `return containingBlock->willStretch(*this, containingAxis, mode);` to

```
if (CheckedPtr renderGrid = dynamicDowncast<RenderGrid>(containingBlock))
   return renderGrid->willStretch(...)

if (CheckedPtr renderFlexibleBox = dynamicDowncast<RenderFlexibleBox>(containingBlock))
   return renderFlexibleBox->willStretch(...)
```
which I like because I can tell from just looking at this code that these two types will stretch. 


If a decent number of the time the `containingBlock` could be something other than `RenderFlexibleBox` or `RenderGrid` then it might result in a vtable lookup just to essentially return false, which technically could be less performant but in practice might not be a big deal (it's one of those things that would probably get optimized when looking at traces anyway).  But maybe that gets offset by the CheckedPtr! 

Up to you!

Source/WebCore/rendering/RenderGrid.cpp
sammygill: I have a mild preference towards scoping things to only when they are needed (not a big deal either way)

Source/WebCore/rendering/RenderGrid.cpp
sammygill: Sure! I didn't realize the alignment of the return values but now that you point it out it helps!

Source/WebCore/rendering/RenderBlock.h
fantasai: I think, as long as it's not a performance problem, let's keep the logic that should be encapsulated properly encapsulated. :) That's the whole point of having subclasses.

(The whole reason we have this PR at all is because we entangled so much subclass logic into superclasses, it wouldn't be possible to fix https://bugs.webkit.org/show_bug.cgi?id=304715 without lots of invasive (and error-prone!) fixup changes everywhere!)

Source/WebCore/rendering/RenderGrid.cpp
fantasai: Ok, I'll make the change.

PR #55924, opened 2025-12-29
Re-enable wasm tail-call inlining, and improve verbose inline debug logs.

justinmichaud:
#### b96c58b1c41538fd9f54f676499d836d606ee3d9
<pre>
Re-enable wasm tail-call inlining, and improve verbose inline debug logs.
<a href="https://bugs.webkit.org/show_bug.cgi?id=304766">https://bugs.webkit.org/show_bug.cgi?id=304766</a>

Reviewed by NOBODY (OOPS!).

With the new inline decision code, some tail calls could never be inlined.

We fix that and improve the debugging logs.

Next, we add two new mechanisms for skipping inlined handlers.

1) Inlined tail calls

These are marked and skipped in StackVisitor, and we add a wasm delegate
for their body.

2) Inlined tail callsites

These become regular calls, but after we commit to making the call,
we need to pretend our frame (and its handlers) are popped. We update
StackVisitor and add a delegate.

We take special care to test that we handle the case when we delegate
to our inline caller, but it has no handlers. In this case, we cannot
find a suitable handler exactly like if we have handlers, but none for
our tag.

We also add some extra tests for the future, when cross-module and
indirect calls may be inlined.
</pre>
https://github.com/WebKit/WebKit/commit/d1d344eeec9a7912c34a1b6db97317154f12ee06

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/174275 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/48208 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/41989 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/183849 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/132143 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/49500 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/47890 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/135572 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/95644 "1 flakes 4 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/177233 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/38997 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/156354 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/116050 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/37638 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/36007 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/32333 "Built successfully") | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/4864 "Passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/148061 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/35847 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/186275 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/35537 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/39468 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/40204 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/144473 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/47875 "Built successfully") | | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/144330 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/48554 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/32467 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/114087 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/39234 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/120478 "Built successfully") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/211310 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/47060 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/10811 "Passed tests") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/55066 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/46463 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/46694 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/46521 "Built successfully") | | |


Constellation:
Is this inlining handled correctly in unwinding?
Let's say, wasm has trytable, and then inside it, tail-call happens. And it is inlined.
In this case, StackVisitor's reporting correctly needs to skip reporting the upper frame of the tail-call (even if it is inlined).

justinmichaud:
No. I encountered this when fixing Ioanna's patch to handle inlined tail calls. I am working on a proper fix, so this can get landed after that fix.

Constellation (changes requested):
r- for now since it requires CallSite traversing work with inlined tailcalls.

Source/JavaScriptCore/wasm/WasmOpcodeOrigin.cpp
```diff
@@ -44,6 +44,7 @@ OpcodeOrigin::OpcodeOrigin(B3::Origin origin)
 
 B3::Origin OpcodeOrigin::asB3Origin()
 {
+    RELEASE_ASSERT_NOT_REACHED();
```
justinmichaud: This appears to be dead, so I would like to remove it. I just want to check with EWS first.

Source/JavaScriptCore/interpreter/StackVisitor.cpp
Constellation: Usually we use 
```
namespace StackVisitorInternal {
constexpr bool verbose = false;
}
```

Source/JavaScriptCore/jit/PCToCodeOriginMap.cpp
Constellation: It is too costly as there are many origins. Use ASSERT. This is not a slow path as it is used for all OMG compilations.

Source/JavaScriptCore/wasm/WasmHandlerInfo.cpp
Constellation: Ditto.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
Constellation: I think this is concerning. We should handle things in StackVisitor side and we should populate the appropriate CallSite information by recognizing tail calls, instead of having exception handler. Exception handler can suppress meany optimizations,, so this isn't fine.

justinmichaud:
Discussed with Keith, we should make this look more like dfg with real callees for inline frames, and skipping inside the wasm exception handler code

PR #55978, opened 2026-01-01
[GTK][WebRTC] Leaks local IP addresses

philn:
#### d6bfcfa91f51193a4c3c5681c28deb9c861dbc59
<pre>
[GTK][WebRTC] Leaks local IP addresses
<a href="https://bugs.webkit.org/show_bug.cgi?id=238873">https://bugs.webkit.org/show_bug.cgi?id=238873</a>

Reviewed by NOBODY (OOPS!).

Host ICE candidates handled by librice are now anonymized with mDNS addresses, thus no exposing local IP
addresses to the SDP. This is handled during candidates gathering in the NetworkProcess, local IP
addresses are registered to a mDNS service hosted in ThirdParty/. That service was imported from
Firefox, it is written in Rust and integrated in the CMake build using the Corrosion module. The
previous approach of relying on Avahi doesn&apos;t work in all cases, most notably Flatpak runtimes.
</pre>
https://github.com/WebKit/WebKit/commit/24ecfd6a2ae02351c018e885d497bf0e04b22e15

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/138196 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/10561 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/138/builds/49606 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/146274 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/91171 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/140070 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/11265 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/10715 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/105713 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/77113 "17 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/141142 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/8429 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/123898 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/86561 "Passed tests") | 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/137544 "Passed tests") | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/8023 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/5787 "Passed tests") | [✅ 🛠 wpe-cairo-libwebrtc](https://ews-build.webkit.org/#/builders/166/builds/6556 "Built successfully") | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/130160 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/117434 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/135/builds/42095 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/148984 "Built successfully") | 
| [✅ 🛠 🧪 jsc-arm64](https://ews-build.webkit.org/#/builders/12/builds/136780 "Built successfully and passed tests") | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/10243 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/136/builds/42652 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/114118 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/10260 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/8653 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/114453 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/7983 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/120182 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/64980 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/10290 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/137/builds/38121 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/169468 "Built successfully") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/10021 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/73857 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/44183 "Passed tests") | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/10230 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/10081 "Built successfully") | | |


Source/cmake/OptionsWPE.cmake
```diff
@@ -88,6 +88,7 @@ WEBKIT_OPTION_DEFAULT_PORT_VALUE(ENABLE_WK_WEB_EXTENSIONS PRIVATE ${ENABLE_EXPER
 WEBKIT_OPTION_DEFAULT_PORT_VALUE(ENABLE_WEBXR PRIVATE ${ENABLE_EXPERIMENTAL_FEATURES})
 WEBKIT_OPTION_DEFAULT_PORT_VALUE(ENABLE_WEBXR_HIT_TEST PRIVATE ${ENABLE_EXPERIMENTAL_FEATURES})
 WEBKIT_OPTION_DEFAULT_PORT_VALUE(ENABLE_WEBXR_LAYERS PRIVATE ${ENABLE_EXPERIMENTAL_FEATURES})
+WEBKIT_OPTION_DEFAULT_PORT_VALUE(ENABLE_MDNS_SERVICE PRIVATE ON)
```
mcatanzaro: This is the only one that's not alphabetized!

Source/cmake/OptionsGTK.cmake
```diff
@@ -141,6 +141,7 @@ WEBKIT_OPTION_DEFAULT_PORT_VALUE(ENABLE_WEBXR_HIT_TEST PRIVATE ${ENABLE_EXPERIME
 WEBKIT_OPTION_DEFAULT_PORT_VALUE(ENABLE_WEBXR_LAYERS PRIVATE ${ENABLE_EXPERIMENTAL_FEATURES})
 
 WEBKIT_OPTION_DEFAULT_PORT_VALUE(USE_LIBRICE PRIVATE ON)
+WEBKIT_OPTION_DEFAULT_PORT_VALUE(ENABLE_MDNS_SERVICE PRIVATE ON)
```
mcatanzaro: Alphabetize:

```suggestion
WEBKIT_OPTION_DEFAULT_PORT_VALUE(ENABLE_MDNS_SERVICE PRIVATE ON)
WEBKIT_OPTION_DEFAULT_PORT_VALUE(USE_LIBRICE PRIVATE ON)
```

Source/WebKit/Platform/MDNSService.cmake
```diff
@@ -0,0 +1,19 @@
+include(FetchContent)
```
mcatanzaro: This will fail in all real production builds. Nobody allows network access during production builds. So got to import Corrosion into ThirdParty as well, if you want to use it....

philn: Well, "Nobody" is a bit of a strong word here... 

I added Corrosion, but for offline builds it likely won't be sufficient because cargo still needs to download the deps during the build. We could vendor them as well, I'll take a look.

philn: Adds 89MB of code though ;)

I wonder if the vendored sources could be added only to the tarballs... Would that be acceptable?

mcatanzaro: I would say none of this is acceptable, but here we are. If you _don't_ vendor the dependencies, it's all but guaranteed that no distros will ever build it.

For what it's worth, I'd say the security risk of adding 700,000 lines of vendored Rust code far outweighs the risk of just doing mDNS or whatever else we need from this in C++.

mcatanzaro: Oh, I misread your question. I don't have any preference as to whether we vendor it via the git repo or via Tools/Scripts/make-dist.

philn: They would be vendored in the tarballs, isn't that what distros use for their builds?

mcatanzaro: I count 22 vendored dependencies of mdns-service-rs, plus mdns-service-rs itself, plus corrosion. So 24 total. Fedora allows unlimited vendored dependencies, but only if each one is declared in the spec file. I'll be theoretically responsible for noticing (which is unlikely) when a new dependency is added or removed. Since they are versioned, we also have to document their versions in the spec file and update them when they change.

It sure makes life much more inconvenient for packagers.

We won't have this problem with Swift because we agreed to ban use of SwiftPM.

Source/WebKit/NetworkProcess/webrtc/rice/RiceBackend.cpp
```diff
@@ -169,59 +169,61 @@ void RiceBackend::notifyIncomingData(unsigned streamId, WebCore::RTCIceProtocol
...
+        mdnsRegister->resolveAddress(address, WTF::move(completionHandler));
+        return;
+    }
+
     auto data = createResolveAddressData();
-    data->resolver = adoptGRef(g_resolver_get_default());
-    data->address = address;
     data->callback = WTF::move(completionHandler);
+    data->resolver = adoptGRef(g_resolver_get_default());
+    data->address = CStringView::unsafeFromUTF8(address.utf8().data());
```
calvaris, line 200: This is a problem. You should use the String as it was. The memory is going to go away the moment the CStringView is created because the CString from utf8() goes out of scope.

calvaris (approved):
Other than what I commented below, I would r=me the GStreamer part.

PR #56000, opened 2026-01-02
Allow more characters in element/attribute names and prefixes

sideshowbarker:
#### 3c11801b7231c6d98ed1419ed17497da12474002
<pre>
Allow more characters in element/attribute names and prefixes
<a href="https://bugs.webkit.org/show_bug.cgi?id=241419">https://bugs.webkit.org/show_bug.cgi?id=241419</a>

Reviewed by Darin Adler.

Implement relaxed name-validation rules per whatwg/dom#1079, whatwg/html#7991,
and whatwg/html#11453. This aligns DOM API name behavior with HTML parser behavior —
which has always allowed a wider range of characters in element and attribute names.

New validation functions in NameValidation.h/cpp:

- isValidElementName(): For createElement — allows [A-Za-z] start followed by
  any char except null/whitespace/&gt;//, or [:_&gt;=0x80] start with restricted continuation
- isValidAttributeName(): For setAttribute/toggleAttribute/createAttribute —
  no null, whitespace, /, &gt;, or =
- isValidNamespacePrefix(): For *NS methods — no null, whitespace, /, or &gt;
- isValidDoctypeName(): For createDocumentType — no null, whitespace, or &gt;
- isValidXMLName(): Preserves old XML Name production for createProcessingInstruction

* LayoutTests/fast/dom/dataset-expected.txt:
* LayoutTests/fast/dom/dataset-xhtml-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/dom/nodes/DOMImplementation-createDocument-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/dom/nodes/DOMImplementation-createDocument.html:
* LayoutTests/imported/w3c/web-platform-tests/dom/nodes/DOMImplementation-createDocumentType-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/dom/nodes/Document-createAttribute-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/dom/nodes/Document-createElement-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/dom/nodes/Document-createElementNS-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/dom/nodes/Document-createElementNS.html:
* LayoutTests/imported/w3c/web-platform-tests/dom/nodes/Document-createElementNS.js:
* LayoutTests/imported/w3c/web-platform-tests/dom/nodes/attributes-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/dom/nodes/name-validation-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/html/dom/elements/global-attributes/dataset-set-expected.txt:
* LayoutTests/platform/glib/http/tests/misc/acid3-expected.txt:
* LayoutTests/platform/mac/http/tests/misc/acid3-expected.txt:
* Source/WebCore/Sources.txt:
* Source/WebCore/dom/DOMImplementation.cpp:
(WebCore::DOMImplementation::createDocumentType):
(WebCore::createHTMLElementWithNameValidation):
(WebCore::Document::createElementForBindings):
(WebCore::Document::createProcessingInstruction):
(WebCore::Document::createElementNS):
(WebCore::Document::createAttribute):
(WebCore::Document::createAttributeNS):
(WebCore::isValidNameStart): Deleted.
(WebCore::isValidNamePart): Deleted.
(WebCore::isValidHTMLElementName):
(WebCore::Document::hasValidNamespaceForElements): Deleted.
(WebCore::Document::hasValidNamespaceForAttributes): Deleted.
(WebCore::isValidNameNonASCII): Deleted.
(WebCore::isValidNameASCII): Deleted.
(WebCore::isValidNameASCIIWithoutColon): Deleted.
(WebCore::Document::isValidName): Deleted.
(WebCore::Document::parseQualifiedName): Deleted.
* Source/WebCore/dom/Element.cpp:
(WebCore::Element::toggleAttribute):
(WebCore::Element::setAttribute):
(WebCore::Element::setAttributeNS):
(WebCore::Element::parseAttributeName):
* Source/WebCore/dom/Element.h:
* Source/WebCore/dom/NameValidation.cpp: Added.
(WebCore::NameValidation::isInvalidAttributeNameCharacter):
(WebCore::NameValidation::isInvalidElementNameCharacterAfterAlphaStart):
(WebCore::NameValidation::isValidElementNameContinuationCharacter):
(WebCore::NameValidation::isInvalidNamespacePrefixCharacter):
(WebCore::NameValidation::isInvalidDoctypeNameCharacter):
(WebCore::NameValidation::isValidElementName):
(WebCore::NameValidation::isValidAttributeName):
(WebCore::NameValidation::isValidNamespacePrefix):
(WebCore::NameValidation::isValidDoctypeName):
(WebCore::NameValidation::isValidASCIIXMLName):
(WebCore::NameValidation::isValidXMLNameStart):
(WebCore::NameValidation::isValidXMLNamePart):
(WebCore::NameValidation::isValidXMLName):
(WebCore::NameValidation::parseQualifiedNameToPair):
(WebCore::NameValidation::parseQualifiedElementName):
(WebCore::NameValidation::parseQualifiedAttributeName):
(WebCore::NameValidation::hasValidNamespaceForElements):
(WebCore::NameValidation::hasValidNamespaceForAttributes):
* Source/WebCore/dom/NameValidation.h: Added.
* Source/WebCore/dom/Node.cpp:
(WebCore::Node::checkSetPrefix):
* Source/WebCore/dom/ProcessingInstruction.cpp:
* Source/WebCore/editing/EditorCommand.cpp:
(WebCore::executeFormatBlock):
* Source/WebCore/svg/animation/SVGSMILElement.cpp:
(WebCore::SVGSMILElement::constructAttributeName const):
* Source/WebCore/xml/parser/XMLDocumentParserLibxml2.cpp:
(WebCore::handleNamespaceAttributes):
(WebCore::handleElementAttributes):

Canonical link: <a href="https://commits.webkit.org/308884@main">https://commits.webkit.org/308884@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/56dee36c129a3d50820cc5ec24b1be086643757d

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/148727 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/21440 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/15009 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/157412 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/102157 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/150600 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/21892 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/21318 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/114657 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/81644 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/151687 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/16859 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/133506 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/95427 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/15971 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/13814 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/4847 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/125570 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/11426 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/159747 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/2887 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/12948 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/122721 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/21242 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/17822 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/122945 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/33432 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/21250 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/133220 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/77404 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/18241 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/9982 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/20852 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/84654 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/20584 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/20731 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/20640 "Built successfully") | | |


LayoutTests/dom/html/level1/core/hc_documentinvalidcharacterexceptioncreateelement-expected.txt
```diff
@@ -1,2 +1,3 @@
-Test:	http://www.w3.org/2001/DOM-Test-Suite/level1/core/hc_documentinvalidcharacterexceptioncreateelement
```
Ahmad-S792: If they are outdated test and we have coverage with WPT, we should delete them.

sideshowbarker: > If they are outdated test and we have coverage with WPT, we should delete them.

👍 Made it so.

sideshowbarker:
EWS is now all-green here, so I reckon this is ready for review.

LayoutTests/fast/dom/DOMImplementation/createDocument-namespace-err-expected.txt
```diff
@@ -15,28 +15,28 @@ FAIL createDocument("", "", null)
 PASS createDocument(null, "<div>", null); threw INVALID_CHARACTER_ERR
 PASS createDocument(null, "0div", null); threw INVALID_CHARACTER_ERR
 PASS createDocument(null, "di v", null); threw INVALID_CHARACTER_ERR
-PASS createDocument(null, "di<v", null); threw INVALID_CHARACTER_ERR
+FAIL createDocument(null, "di<v", null)
```
annevk: For the tests that we decide to keep we should update them to make them pass with the new logic.

Source/WebCore/dom/Document.cpp
```diff
@@ -517,22 +517,42 @@ static void CallbackForContainIntrinsicSize(const Vector<Ref<ResizeObserverEntry
     }
 }
 
-// https://www.w3.org/TR/xml/#NT-NameStartChar
-// NameStartChar       ::=       ":" | [A-Z] | "_" | [a-z] | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x2FF] | [#x370-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]
-static inline bool isValidNameStart(char32_t c)
+// https://dom.spec.whatwg.org/#validate-and-extract
+// The functions here implement "relaxed" name-validation rules that align with the
+// names the HTML parser is capable of producing, and that were introduced in
+// https://github.com/whatwg/dom/pull/1079
```
annevk: I think just the initial link is sufficient here. We don't need all this context over the long term.

Source/WebCore/dom/Document.cpp
```diff
@@ -517,22 +517,42 @@ static void CallbackForContainIntrinsicSize(const Vector<Ref<ResizeObserverEntry
...
+
+// Invalid characters for attribute names: null, ASCII whitespace, '/', '>', '='
+static inline bool isInvalidAttributeNameCharacter(char32_t c)
 {
-    return c == ':' || (c >= 'A' && c <= 'Z') || c == '_' || (c >= 'a' && c <= 'z') || (c >= 0x00C0 && c <= 0x00D6)
-        || (c >= 0x00D8 && c <= 0x00F6) || (c >= 0x00F8 && c <= 0x02FF) || (c >= 0x0370 && c <= 0x037D) || (c >= 0x037F && c <= 0x1FFF)
-        || (c >= 0x200C && c <= 0x200D) || (c >= 0x2070 && c <= 0x218F) || (c >= 0x2C00 && c <= 0x2FeF) || (c >= 0x3001 && c <= 0xD7FF)
-        || (c >= 0xF900 && c <= 0xFDCF) || (c >= 0xFDF0 && c <= 0xFFFD) || (c >= 0x10000 && c <= 0xEFFFF);
+    return c == 0x00 || c == 0x09 || c == 0x0A || c == 0x0C || c == 0x0D || c == 0x20
+        || c == '/' || c == '>' || c == '=';
```
annevk: Since we use isASCIIAlpha below, should we use isASCIIWhitespace here?

Source/WebCore/dom/Document.cpp
```diff
@@ -7520,132 +7540,256 @@ void Document::updateCachedCookiesEnabled()
...
 {
-    if (!isValidNameStart(characters[0]))
+    if (characters.empty())
         return false;
 
-    for (size_t i = 1; i < characters.size(); ++i) {
-        if (!isValidNamePart(characters[i]))
-            return false;
+    size_t i = 0;
+    char32_t firstChar;
```
annevk: Can't we use 16 here? I don't see you combining surrogates and I don't believe that's needed. Though maybe this should even be CharType?

Source/WebCore/dom/Document.h
```diff
@@ -1238,15 +1238,21 @@ class Document
...
+
+    // Enum to specify whether a qualified name is for an element or attribute
+    enum class QualifiedNameType : bool { Element, Attribute };
 
     // The following breaks a qualified name into a prefix and a local name.
     // It also does a validity check, and returns an error if the qualified name is invalid.
-    static ExceptionOr<std::pair<AtomString, AtomString>> parseQualifiedName(const AtomString& qualifiedName);
-    static ExceptionOr<QualifiedName> parseQualifiedName(const AtomString& namespaceURI, const AtomString& qualifiedName);
+    static ExceptionOr<std::pair<AtomString, AtomString>> parseQualifiedName(const AtomString& qualifiedName, QualifiedNameType = QualifiedNameType::Element);
+    static ExceptionOr<QualifiedName> parseQualifiedName(const AtomString& namespaceURI, const AtomString& qualifiedName, QualifiedNameType = QualifiedNameType::Element);
```
annevk: I wonder if we should split all these statics out into their own file. They don't have much to do with Document after all.

Source/WebCore/dom/Document.h
```diff
@@ -1238,15 +1238,21 @@ class Document
...
-    // To get this right for all the XML cases, we probably have to improve this or move it
-    // and make it sensitive to the type of document.
-    static bool isValidName(const String&);
+    // https://dom.spec.whatwg.org/#validate-and-extract
+    static bool isValidElementName(const String&);
+    static bool isValidAttributeName(const String&);
+    static bool isValidNamespacePrefix(StringView);
+    static bool isValidDoctypeName(const String&);
+    static bool isValidXMLName(const String&); // XML Name production (for processing instructions)
+    static bool isValidName(const String&); // Legacy: uses element name rules
```
annevk: What does legacy mean? Is it no longer used by any specification? Why do we still need it?

annevk (commented):
This is very cool to see! Thanks for working on it.

I have some feedback and a couple nits that I left inline:

- Do other browsers ship this already?
- I would like to see the test changes in a separate PR so they can be independently reviewed. In particular the new tests added from WPT and the tests you propose to remove. We should not include changes to existing tests in that PR to account for the new logic. That will reduce the size of this PR quite a bit and more clearly indicate the progressions in WPT.

LayoutTests/fast/dom/DOMImplementation/createDocument-namespace-err-expected.txt
sideshowbarker: After looking back at them, it seems we don’t actually need any of them — they’re redundant with the WPT `name-validation.html` test. So, now removed them. (Locally — I will push shortly.)

Source/WebCore/dom/Document.cpp
sideshowbarker: OK, reduced the comment to just the spec link (and same for some other comments I added here)

Source/WebCore/dom/Document.cpp
sideshowbarker: Yes, changed to calling `isASCIIWhitespace`

Source/WebCore/dom/Document.cpp
sideshowbarker: Yes and Yes. Made it so. Thanks much — that makes those parts easier and cleaner

Source/WebCore/dom/Document.h
sideshowbarker: Split the statics out, to constitute a new `NameValidation.h`, and a `NameValidation.cpp` implementation.

Source/WebCore/dom/Document.h
sideshowbarker: s/Legacy/XML/

I had just meant, not from the DOM spec (current place where we define this stuff in general) but from the XML spec (not a current place where we define things any longer)

But anyway, we do still need it: For PIs (`createProcessingInstruction`)

annevk: But I see both isValidXMLName and isValidName. Can we then get rid of one?

sideshowbarker:
>  Do other browsers ship this already?

I should have also mentioned: Firefox does not ship it yet — but I am working on an implementation there too:

https://phabricator.services.mozilla.com/D277822

Source/WebCore/dom/messageports/WorkerMessagePortChannelProvider.h
darinadler: It’s unfortunate that this additional include is required rather than a forward declaration. Do you know why?

Source/WebCore/dom/Document.cpp
darinadler: Does this function need to exist? Can callers be changed to use `NameValidation.h` directly instead?

darinadler: Does this function need to exist? Can callers be changed to use `NameValidation.h` directly instead?

darinadler: Does this function need to exist? Can callers be changed to use `NameValidation.h` directly instead?

darinadler: Does this function need to exist? Can callers be changed to use `NameValidation.h` directly instead?

darinadler: Does this function need to exist? Can callers be changed to use `NameValidation.h` directly instead?

darinadler: Does this function need to exist? Can callers be changed to use `NameValidation.h` directly instead?

darinadler, line 7572: Why don’t we need a fast path any more? Is the new code faster than the old code was?

Source/WebCore/dom/NameValidation.h
darinadler:
```suggestion
// https://dom.spec.whatwg.org/#valid-element-name
bool isValidElementName(StringView);

// https://dom.spec.whatwg.org/#valid-attribute-name
bool isValidAttributeName(StringView);

// https://dom.spec.whatwg.org/#valid-namespace-prefix
bool isValidNamespacePrefix(StringView);

// https://dom.spec.whatwg.org/#valid-doctype-name
bool isValidDoctypeName(StringView);

// https://www.w3.org/TR/xml/#NT-Name - used by createProcessingInstruction
bool isValidXMLName(StringView);
```

darinadler: I see no reason this header needs to include `StringView.h`.
```suggestion
```

Source/WebCore/dom/NameValidation.cpp
darinadler: Please use `CharacterType` instead of `CharType`, since WebKit coding guidelines prefer words.

darinadler: I’m not sure there is performance optimization evidence that these need to be templates. We can consider just implementing `char16_t` versions.

In general I am ambivalent about how we are optimizing the code in this file. We should make sure these functions are really efficient for both 8-bit and 16-bit strings, but my guess is that we don’t need it all to be templates like this.

It’s not great that this pull request removes some optimizations but keeps some others and adds new ones without a clear foundation. We need to make sure this code is super efficient for common cases, but not sure we want all the code in template with separate 8-bit and 16-bit character code paths; we should decide by measuring performance.

darinadler: Please don’t use `Char` in new WebKit code. We prefer words, so `firstCharacter` would be better.

Source/WebCore/dom/NameValidation.h
darinadler: These functions should all take `StringView`, not `const String&`.

Source/WebCore/dom/Document.h
darinadler: I suggest not adding these functions to `Document`.

darinadler: I suggest removing this function from `Document` and changing callers to use functions from `NameValidation.h` instead.

Source/WebCore/dom/Document.cpp
darinadler: It doesn’t make sense to use `U16_NEXT` to search for colons in an `AtomString`. Since colons are ASCII characters, we certainly don’t need to interpret UTF-16 surrogate pairs. And also, we should use the `find` function to do this search, calling it twice to find both colons.

darinadler: These `toString` calls are inefficient and we shouldn’t add them. We can avoid them by making the functions they call work on `StringView`.

Source/WebCore/dom/Document.h
darinadler: I’d like to get these `parseQualifiedName` functions either out of `Document` entirely, or at least into `private` member functions. It’s super frustrating to have them here just so that `SVGSMILElement::constructAttributeName` can use them, with all other uses inside `Document.cpp`.

darinadler: This enumeration is awkward enough that I’d like to keep it as a local implementation detail and not in one of the most-included files in the project! We should consider just having `parseQualifiedAttributeName` in `Document.h` since it’s used by SVG, and the rest inside `Document.cpp` and not member functions.

darinadler (changes requested):

Source/WebCore/dom/messageports/WorkerMessagePortChannelProvider.h
sideshowbarker: > It’s unfortunate that this additional include is required rather than a forward declaration. Do you know why?

Got CI failures; for example, https://ews-build.webkit.org/#/builders/2/builds/148758

After asking about it on Slack at https://webkit.slack.com/archives/CU64U6FDW/p1767871164353619, this seemed to be the way to address it.

darinadler: I’d expect an include of `WeakPtr.h`, though.

sideshowbarker: > I’d expect an include of `WeakPtr.h`, though.

Now changed it just that

Source/WebCore/dom/Document.cpp
sideshowbarker: > Can callers be changed to use `NameValidation.h` directly instead?

Yes — changed all the callers to use `NameValidation.h` directly (and removed all the `Document::isValid*` wrapper functions).

sideshowbarker: > Why don’t we need a fast path any more? Is the new code faster than the old code was?

Short answer: Yes and yes

Longer answer:

The existing “fast path” checking optimizes for the common case of simple ASCII names — while the new checking here is _already_ fast for that case. It’s way simpler — all it’s doing is: checking for a few invalid characters, rather than checking against whole Unicode ranges. For names that start with a letter, it's essentially just checking: Is each character not whitespace, `/`, or `>`.

And now the colon-finding is switched to using (optimized) `String::find(':')`

Source/WebCore/dom/NameValidation.h
sideshowbarker: replaced with just `#include <wtf/Forward.h>`

Source/WebCore/dom/NameValidation.cpp
sideshowbarker: > Please use `CharacterType` instead of `CharType`, since WebKit coding guidelines prefer words.

✅ Made it so.

sideshowbarker: > I’m not sure there is performance optimization evidence that these need to be templates. We can consider just implementing `char16_t` versions.

OK — removed the templates and switched to `char16_t` using `upconvertedCharacters().span()`

sideshowbarker: > Please don’t use `Char` in new WebKit code. We prefer words, so `firstCharacter` would be better.

Made it so

Source/WebCore/dom/NameValidation.h
sideshowbarker: > These functions should all take `StringView`, not `const String&`.

✅ Changed all

Source/WebCore/dom/Document.h
sideshowbarker: > I suggest not adding these functions to `Document`.
> …
> I suggest removing this function from `Document` and changing callers to use functions from `NameValidation.h` instead.

Made it so: All the `isValid*` wrappers are now gone from `Document`.

Source/WebCore/dom/Document.cpp
sideshowbarker: > It doesn’t make sense to use `U16_NEXT` to search for colons in an `AtomString`. Since colons are ASCII characters, we certainly don’t need to interpret UTF-16 surrogate pairs. And also, we should use the `find` function to do this search, calling it twice to find both colons.

Made it so.

sideshowbarker: > These `toString` calls are inefficient and we shouldn’t add them. We can avoid them by making the functions they call work on `StringView`.

Made that switch.

Source/WebCore/dom/Document.h
sideshowbarker: > I’d like to get these `parseQualifiedName` functions either out of `Document` entirely, or at least into `private` member functions.

moved `parseQualifiedAttributeName` to `NameValidation`. `parseQualifiedName` remains in `Document` as file-local static.

sideshowbarker: Moved `parseQualifiedAttributeName` to `NameValidation`. Kept the `QualifiedNameType` enum in `Document.`cpp as file-local, not exposed.

Source/WebCore/dom/Document.cpp
darinadler: I don’t think we need this function, can just move this to callers.

darinadler: I don’t think we need this function, can just move this to callers.

Source/WebCore/dom/NameValidation.cpp
darinadler: It’s good that we changed the functions that take single `char16_t` to not unnecessarily be templates; thank you! I don’t think it’s good to require upconverting characters, requires allocating memory for 8-bit, which should be a fast case, so this should either work directly with `StringView`, or be changed back into a template that supports both character types. I suspect that working directly with `StringView` would be fine, but either approach is OK.

darinadler: It’s good that we changed the functions that take single `char16_t` to not unnecessarily be templates; thank you! I don’t think it’s good to require upconverting characters, requires allocating memory for 8-bit, which should be a fast case, so this should either work directly with `StringView`, or be changed back into a template that supports both character types. I suspect that working directly with `StringView` would be fine, but either approach is OK.

darinadler: It’s good that we changed the functions that take single `char16_t` to not unnecessarily be templates; thank you! I don’t think it’s good to require upconverting characters, requires allocating memory for 8-bit, which should be a fast case, so this should either work directly with `StringView`, or be changed back into a template that supports both character types. I suspect that working directly with `StringView` would be fine, but either approach is OK.

darinadler: It’s good that we changed the functions that take single `char16_t` to not unnecessarily be templates; thank you! I don’t think it’s good to require upconverting characters, requires allocating memory for 8-bit, which should be a fast case, so this should either work directly with `StringView`, or be changed back into a template that supports both character types. I suspect that working directly with `StringView` would be fine, but either approach is OK.

darinadler: Let’s not use `upconvertedCharacters`.

darinadler: Let’s not use `upconvertedCharacters`.

darinadler: Let’s not use `upconvertedCharacters`.

darinadler: Let’s not use `upconvertedCharacters`.

Source/WebCore/dom/NameValidation.h
darinadler: Should be able to use this forward declaration instead of including `ExceptionOr.h`:
```
template<typename> class ExceptionOr;
```

darinadler, line 29: I think we might also need `#include <utility>` unless `Forward.h` pulls that in.

Source/WebCore/editing/EditorCommand.cpp
darinadler: This is incorrect. Should just be:
```
#include "NameValidation.h"
```
If that doesn’t work on some platform, we have made a build system mistake. Likely the issue is that we need to add it to the Xcode project for Apple’s platforms.

darinadler: I’d like to see this use `WTF::move(tagName)` to avoid a little reference count churn, but maybe it’s OK without that, since it might require changing the other arguments to more complex syntax.

Source/WebCore/svg/animation/SVGSMILElement.cpp
darinadler: This is incorrect. Should just be:
```
#include "NameValidation.h"
```
If that doesn’t work on some platform, we have made a build system mistake. Likely the issue is that we need to add it to the Xcode project for Apple’s platforms.

darinadler (changes requested):
Great work, this is looking really good!

Source/WebCore/dom/Document.cpp
darinadler: I would like to know how we know it’s fast enough. I want to make sure we check with the relevant benchmark. For example, adding a call to `upconvertedCharacters`, which we should not do, likely adds a memory allocation that makes it quite a bit slower, so I’m not sure that we don’t need a fast path if we do that.

Removing an optimization that was originally motivated by a discovery from profiling performance on a benchmark or slow website is a slightly risky thing to do; we want to take extra care in such cases, and I think this may be one of those.

rniwa (commented):
We can't land this without doing perf A/B testing but it's most certainly a Speedometer 3 regression. There are many micro optimizations that were done in this part of the codebase for Speedometer 3 specifically.

Source/WebCore/dom/Document.cpp
```diff
@@ -7520,132 +7507,50 @@ void Document::updateCachedCookiesEnabled()
...
-
-        if (isValidNameASCII(characters))
-            return true;
-
-        return isValidNameNonASCII(characters);
-    }
-
-    auto characters = name.span16();
-
-    if (isValidNameASCII(characters))
```
rniwa, line 7553: It's an important optimization that we check the valid ASCII names first even when the string isn't 8-bit because most element names are ASCII regardless of the rest of the text in the source.

Source/WebCore/dom/NameValidation.cpp
```diff
@@ -0,0 +1,221 @@
...
+        }
+        return true;
+    }
+
+    return false;
+}
+
+bool isValidElementName(StringView name)
+{
+    return isValidElementNameImpl(name.upconvertedCharacters().span());
```
rniwa: This is most certainly a perf regression in Speedometer. We can't always convert to 16-bit character.

rniwa (changes requested):
I don't think the PR as is landable. We should make more targeted code change in the non-ASCIl character name validation.

Source/WebCore/dom/Document.cpp
sideshowbarker: Made it so.

Source/WebCore/dom/NameValidation.cpp
sideshowbarker: Went with the template approach, and dropped the allocation/upconverting

Source/WebCore/dom/NameValidation.h
sideshowbarker: Made it so: Changed to forward declaration in the header + moved `#include "ExceptionOr.h"` to `NameValidation.cpp`

sideshowbarker: ✅ Added `#include <utility>`

Source/WebCore/editing/EditorCommand.cpp
sideshowbarker: ✅ Corrected the include path, and added `NameValidation.h`+`NameValidation.cpp` to the Xcode project

sideshowbarker: ✅ Changed to use `WTF::move(tagName)`

Source/WebCore/svg/animation/SVGSMILElement.cpp
sideshowbarker: Updated Xcode project

Source/WebCore/dom/Document.cpp
sideshowbarker: OK, added (back) the ASCII fast paths.

Source/WebCore/dom/Document.cpp
sideshowbarker: OK, added (back) the ASCII fast paths.

Source/WebCore/dom/NameValidation.cpp
sideshowbarker: I’ve re-run Speedometer after implementing the ASCII fast paths, and I’ll post comment with the new numbers.

Source/WebCore/dom/Document.cpp
darinadler: Seems to me that a more elegant way of doing this is to overload for both argument types in `NameValidation.h`.

darinadler: This function that returns a `pair` is never used except in the implementation of the other overloaded `parseQualifiedName` function. I suggest merging the two into one function; there’s probably some historical reason they were separate.

darinadler: If `NameValidation::isValidElementName` was overloaded with a version that takes a `const QualifiedName&`, then we wouldn’t need this here. I think that might be a little nicer. Not sure, either way is OK with me.

darinadler: Thank you for moving this in here. But I think we want to share the code with `parseQualifiedAttributeName` rather than having two copies of the logic. So what we should do is put a `parseQualifiedElementName` in `NameValidation.h` and use both of those functions here in `Document.cpp`. In `NameValidation.cpp` the element and attribute versions can share code.

We won’t need versions that return `pair<AtomString, AtomString>`, though.

darinadler: This should call the `parseQualifiedAttributeName` function from `NameValidation.h` unless there’s some reason it can’t.

Source/WebCore/dom/Element.cpp
darinadler: I think we should get rid of this function and have the call sites here in `Element.cpp` and in `XMLDocumentParserLibxml2.cpp` both call the function in `NameValidation.h` directly.

I also think that `Document::hasValidNamespaceForAttributes` and `Document::hasValidNamespaceForElements` should move into `NameValidation.h` and `NameValidation.cpp`.

It’s not urgent to do this in this pull request, but I think that will be a better long term understandable and maintainable factoring for this.

Source/WebCore/dom/NameValidation.cpp
darinadler: All the code in this file still says `CharType`, not `CharacterType`.

darinadler: Is it intentional that this returns `true` for the empty string?

darinadler: Is it intentional that this returns `true` for the empty string?

darinadler: I don’t think doing this check up front is helpful. We don’t need to optimize the empty string case. I think we should put the empty string check at the end of `isValidXMLNameNonASCII` instead.

darinadler: How did you decide to leave this one as a template? I would suggest using `StringView` instead for this.

darinadler: How did you decide to leave this one as a template? I would suggest using `StringView` instead for this one.

darinadler: I don’t understand this function’s strategy. How does it make things faster to do an ASCII-only check here? The full check seems like it would be just as efficient. Also, what is checked here seems an arbitrary list of characters, not matching an ASCII subset of the rules about what is valid in a namespace prefix, and in fact it seems a little more expensive than the correct check would be.

darinadler: I don’t understand this function’s strategy. How does it make things faster to do an ASCII-only check here? The full check seems like it would be just as efficient. Also, what is checked here seems an arbitrary list of characters, unrelated to an ASCII-only subset of rules about what is valid in a doctype, and in fact it seems a little more expensive than the correct check would be.

Also, is it intentional that this returns `true` for an empty string?

darinadler (changes requested):

Source/WebCore/dom/NameValidation.cpp
sideshowbarker: > All the code in this file still says CharType, not CharacterType.

Sorry about that — all changed now (locally; will push again after going through all the review comments)

sideshowbarker: > How did you decide to leave this one as a template? I would suggest using `StringView` instead for this.

OK, I’ve now switched all of them to just using `StringView` directly instead.

(I had used templates because I’d done that previously — and I had been focused on the more-immediate problem of fixing the performance regression, and it wasn’t clear to me that using `StringView` directly was gonna make any difference for that. But, with them switched now, it’s slightly less code, and it’s certainly easier to read. So yeah, in hindsight, I can see that’s how I should have had it from the get-go…)

sideshowbarker: > ll the code in this file still says `CharType`, not `CharacterType`.

I’ve changed it all now locally, and will make sure it remains so this time before I push again

sideshowbarker: > Is it intentional that this returns `true` for the empty string?

No, that was not intentional — so, thanks for catching it. Added an `isEmpty()` check there.

sideshowbarker: > > All the code in this file still says CharType, not CharacterType.
> 
> Sorry about that — all changed now (locally; will push again after going through all the review comments)

All of course moot now, since I switched to using `StringView` directly. But I promise I did finally get around to changing it locally! (And going forward, I will make an effort to remember to follow the naming policy.)

Source/WebCore/dom/Document.cpp
sideshowbarker: Thanks, yeah, beautiful — added that overload and updated the calling code

sideshowbarker: Thanks for catching that — yes, merged the two into one that returns `QualifiedName` directly

sideshowbarker: Made it so. Or I think I did, at least. Lemme know if I muffed anything.

I did have to keep one around for `SVGSMILElement` — which looks up the namespace URI separately after parsing.

sideshowbarker: It’s now doing that.

Source/WebCore/dom/Element.cpp
sideshowbarker: > It’s not urgent to do this in this pull request, but I think that will be a better long term understandable and maintainable factoring for this.

OK, I went ahead and did it — rather than putting it on the back burner, for a follow-up PR — because I’d rather try to get it done here, while I’ve still got your attention, and can hopefully get more of your time for review to get this over the line.

Source/WebCore/dom/NameValidation.cpp
sideshowbarker: OK, moved. (Not in exactly the same way — because I subsequently refactored everything, in response to other review comments — but I think functionally the same?)

sideshowbarker: Changed to removed both ASCII fast paths:

  1. `isValidNamespacePrefix` — now just uses the blacklist check with `isInvalidNamespacePrefixCharacter`
  2. `isValidDoctypeName` — now just uses the blacklist check with `isInvalidDoctypeNameCharacter`

Both also now handle empty strings at the end (via `hasCharacters`) rather than up front.

sideshowbarker:
> Apart from the XML validation which does actually need to look at code points, I think all other validation can operate on char16_t and basic string enumeration. (Because we only need to look at code points if we perform checks on code points that are surrogates or higher than U+FFFF which we only do for XML.) This means we don't need isValidASCIIElementName and isValidASCIIAttributeName.

Yep — so I will make a further update to change the handling to that

sideshowbarker:
> Apart from the XML validation which does actually need to look at code points, I think all other validation can operate on char16_t and basic string enumeration. (Because we only need to look at code points if we perform checks on code points that are surrogates or higher than U+FFFF which we only do for XML.) This means we don't need isValidASCIIElementName and isValidASCIIAttributeName.

Finally got back around to making those changes.

Source/WebCore/dom/NameValidation.cpp
```diff
@@ -0,0 +1,269 @@
...
+}
+
+static constexpr bool isInvalidElementNameCharacterAfterAlphaStart(char16_t c)
+{
+    return !c || isASCIIWhitespace(c) || c == '/' || c == '>';
+}
+
+static constexpr bool isValidElementNameContinuationCharacter(char16_t c)
+{
+    return isASCIIAlpha(c) || isASCIIDigit(c) || c == '-' || c == '.' || c == ':' || c == '_' || c >= 0x80;
```
darinadler:
```suggestion
    return isASCIIAlphanumeric(c) || c == '-' || c == '.' || c == ':' || c == '_' || !isASCII(c);
```

Source/WebCore/dom/NameValidation.cpp
```diff
@@ -0,0 +1,269 @@
...
+// https://dom.spec.whatwg.org/#valid-attribute-name
+bool isValidAttributeName(StringView name)
+{
+    if (name.isEmpty())
+        return false;
+
+    for (unsigned i = 0; i < name.length(); ++i) {
+        if (isInvalidAttributeNameCharacter(name[i]))
+            return false;
+    }
```
darinadler:
```suggestion
    for (auto codeUnit : name.codeUnits()) {
        if (isInvalidAttributeNameCharacter(codeUnit))
            return false;
    }
```

Source/WebCore/dom/NameValidation.cpp
```diff
@@ -0,0 +1,269 @@
...
+// https://dom.spec.whatwg.org/#valid-namespace-prefix
+bool isValidNamespacePrefix(StringView prefix)
+{
+    if (prefix.isEmpty())
+        return false;
+
+    for (unsigned i = 0; i < prefix.length(); ++i) {
+        if (isInvalidNamespacePrefixCharacter(prefix[i]))
+            return false;
+    }
```
darinadler:
```suggestion
    for (auto codeUnit : prefix.codeUnits()) {
        if (isInvalidNamespacePrefixCharacter(codeUnit))
            return false;
    }
```

Source/WebCore/dom/NameValidation.cpp
```diff
@@ -0,0 +1,269 @@
...
+    return true;
+}
+
+// https://dom.spec.whatwg.org/#valid-doctype-name
+bool isValidDoctypeName(StringView name)
+{
+    for (unsigned i = 0; i < name.length(); ++i) {
+        if (isInvalidDoctypeNameCharacter(name[i]))
+            return false;
+    }
```
darinadler:
```suggestion
    for (auto codeUnit : name.codeUnits()) {
        if (isInvalidDoctypeNameCharacter(codeUnit))
            return false;
    }
```

Source/WebCore/dom/NameValidation.cpp
```diff
@@ -0,0 +1,269 @@
...
+}
+
+// https://www.w3.org/TR/xml/#NT-NameStartChar
+// NameStartChar ::= ":" | [A-Z] | "_" | [a-z] | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x2FF] | [#x370-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]
+static inline bool isValidXMLNameStart(char32_t c)
+{
+    return c == ':' || (c >= 'A' && c <= 'Z') || c == '_' || (c >= 'a' && c <= 'z') || (c >= 0x00C0 && c <= 0x00D6)
+        || (c >= 0x00D8 && c <= 0x00F6) || (c >= 0x00F8 && c <= 0x02FF) || (c >= 0x0370 && c <= 0x037D) || (c >= 0x037F && c <= 0x1FFF)
+        || (c >= 0x200C && c <= 0x200D) || (c >= 0x2070 && c <= 0x218F) || (c >= 0x2C00 && c <= 0x2FEF) || (c >= 0x3001 && c <= 0xD7FF)
+        || (c >= 0xF900 && c <= 0xFDCF) || (c >= 0xFDF0 && c <= 0xFFFD) || (c >= 0x10000 && c <= 0xEFFFF);
```
darinadler:
```suggestion
    return c == ':' || isASCIIAlpha(c) || c == '_' || (c >= 0x00C0 && c <= 0x00D6)
        || (c >= 0x00D8 && c <= 0x00F6) || (c >= 0x00F8 && c <= 0x02FF) || (c >= 0x0370 && c <= 0x037D) || (c >= 0x037F && c <= 0x1FFF)
        || (c >= 0x200C && c <= 0x200D) || (c >= 0x2070 && c <= 0x218F) || (c >= 0x2C00 && c <= 0x2FEF) || (c >= 0x3001 && c <= 0xD7FF)
        || (c >= 0xF900 && c <= 0xFDCF) || (c >= 0xFDF0 && c <= 0xFFFD) || (c >= 0x10000 && c <= 0xEFFFF);
```

Source/WebCore/dom/NameValidation.cpp
```diff
@@ -0,0 +1,269 @@
...
+}
+
+// https://www.w3.org/TR/xml/#NT-NameChar
+// NameChar ::= NameStartChar | "-" | "." | [0-9] | #xB7 | [#x0300-#x036F] | [#x203F-#x2040]
+static inline bool isValidXMLNamePart(char32_t c)
+{
+    return isValidXMLNameStart(c) || c == '-' || c == '.' || (c >= '0' && c <= '9') || c == 0x00B7
```
darinadler:
```suggestion
    return isValidXMLNameStart(c) || c == '-' || c == '.' || isASCIIDigit(c) || c == 0x00B7
```

darinadler (approved):
I suspect some of the functions in `NameValidation.h` could be written to handle common cases significantly more efficiently. But this is only valuable if they are bottlenecks and have a measurable effect on performance. We need to ensure none of this change slows down our key benchmarks, but otherwise this looks good to me.

PR #56024, opened 2026-01-02
[Grid] Slow performance in accumulateIntrinsicSizesForTrack causes sluggish behavior on some content.

sammygill:
#### 425f9293f8d4f3cd2423eabd9dd099fe0dce82b2
<pre>
[Grid] Slow performance in accumulateIntrinsicSizesForTrack causes sluggish behavior on some content.
<a href="https://bugs.webkit.org/show_bug.cgi?id=304869">https://bugs.webkit.org/show_bug.cgi?id=304869</a>
<a href="https://rdar.apple.com/problem/167456605">rdar://problem/167456605</a>

Reviewed by Elika Etemad and Alan Baradlay.

Almost all of the time is spent inside of GridTrackSizingAlgorithm::accumulateIntrinsicSizesForTrack,
so in this patch I reworked this method from being a recursive
implementation that iterates over all of the tracks to being an
iterative one that just iterates over all of the grid items. To properly
explain and understand why this is an improvement I will first explain
what the code currently does. To keep things simple we will assume that
we are sizing the columns but the same exact logic applies to the rows.

In the case where the content does not have any subgrids this is
actually fairly straightforward. GridTrackSizingAlgorithm::resolveIntrinsicTrackSizes
will iterate over each of the intrinsically sized tracks and provide a
GridIterator via GridIterator iterator(grid, GridTrackSizingDirection::Columns, 0)
to accumulateIntrinsicSizesForTrack. That method will then use the
iterator to get each of the items in column 0 by calling iterator.nextGridItem().

The interesting bit is actually how the iterator handles looping over all
of the items in a track. When looping over the columns, the iterator
stores m_direction as Columns to find the track that we will loop over.
This loop is done by keeping track of a separate index in the opposite
dimension called the varyingTrackIndex which gets incremented to loop
over all of the tracks. So to get all of the items in column 0,
varyingTrackIndex will correspond to the index of the current row we are
looking at and will get incremented until we go through all the rows.
This means that the iterator may return the same item multiple times if
it ends up spanning multiple rows. accumulateIntrinsicTrackSizes handles
this quirk by keeping track of a separate items set and skips the item
if it is already present in the set.

Each time we get a new item from the iterator, we will then either call
sizeTrackToFitNonSpanningItem on that item or put it into one of two
other vectors that will get sorted and used later in
resolveIntrinsicTrackSizes.

However, things get slightly more complex when we start throwing
subgrids into the mix. This is because subgrids may add in an extra
layer of implicit margin to items in the track through their gaps and
any other margin, border, and padding specified on them and can be
accumulated from any nested subgrids. The code currently accomplishes
this by implementing the logic in a recursive manner and stores the
values in a LayoutUnit that is taken in as a reference to the function.
This value gets modified and reset as we progress through the call
stack.

The performance problem stems from the fact that if we see a subgrid
that spans multiple rows, we will recurse into this function for each
row that it spans. This is because we still run the logic to correctly
compute the implicit margins *before* we realize that this is a grid
item we have already seen and continue.

While there may have been a couple of different ways to fix this bug, I
opted to slightly rework this function to instead loop over the grid
items once instead of recursing into each subgrid multiple times. This
is based off the observation that when we are sizing the columns, we
should not need to care about the rows that the items span since that
should not have any bearing on the sizes that are contributed for the
columns.

In the case where the item is not a subgrid, the logic is basically the
same as before, but the only difference is that we need to figure out
the track that gets passed into sizeTrackToFitNonSpanningItem rather than the subgrid.
than using the one that was passed in from resolveIntrinsicTrackSizes.

When we encounter a subgrid, the core logic is mostly the same, but we
need to keep track of the state on our own stack rather than relying on
the call stack to hold the state during recursion. To do this, I created
a new struct that is used locally within this function called
SizingData.

On my M4 Max MBP, the above test case goes from ~0.20 runs/s to ~50
runs/s. Other engines still show considerably higher runs/s, but this is
definitely a noticeable improvement and a good starting point to build
on.

Canonical link: <a href="https://commits.webkit.org/305121@main">https://commits.webkit.org/305121@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/b45bbcab1794529fe1c2067c84e206a009b53147

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/137553 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/9914 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/138/builds/48840 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/145310 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/90526 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/f2c6d2ed-bd8b-4d70-a499-64172ae81bd0/4df7c5b1-0b43-4a06-808d-168c72e8ed28) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/139425 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/10617 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/10044 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/105198 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/76814 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/140498 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/7879 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/123275 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/86051 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/f2c6d2ed-bd8b-4d70-a499-64172ae81bd0/7066f158-6164-42a7-adc9-ee6a5c999e17) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/7515 "Passed tests") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/5262 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 wpe-cairo-libwebrtc](https://ews-build.webkit.org/#/builders/166/builds/5893 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/116885 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/135/builds/41431 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/148074 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/9596 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/136/builds/41985 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/113586 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/9614 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/8090 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/113922 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/28917 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/7435 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/119514 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/64252 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/9645 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/137/builds/37567 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/9376 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/73210 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/9585 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/9437 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/rendering/GridTrackSizingAlgorithm.cpp
+++ b/Source/WebCore/rendering/GridTrackSizingAlgorithm.cpp
@@ -1742,33 +1742,51 @@ bool GridTrackSizingAlgorithm::shouldExcludeGridItemForMasonryTrackSizing(const 
     return shouldExcludeGridItemForMasonryTrackSizing;
 }
 
-void GridTrackSizingAlgorithm::accumulateIntrinsicSizesForTrack(GridTrack& track, unsigned trackIndex, GridIterator& iterator, Vector<GridItemWithSpan>& itemsSortedByIncreasingSpan, Vector<GridItemWithSpan>& itemsCrossingFlexibleTracks, SingleThreadWeakHashSet<RenderBox>& itemsSet, LayoutUnit currentAccumulatedMbp, GridLayoutState& gridLayoutState)
+void GridTrackSizingAlgorithm::aggregateGridItemsForIntrinsicSizing(Vector<GridItemWithSpan>& itemsSortedByIncreasingSpan, Vector<GridItemWithSpan>& itemsCrossingFlexibleTracks, GridLayoutState& gridLayoutState)
 {
-    while (CheckedPtr gridItem = iterator.nextGridItem()) {
-        bool isNewEntry = itemsSet.add(*gridItem).isNewEntry;
-        GridSpan span = m_renderGrid->gridSpanForGridItem(*gridItem, m_direction);
+    auto& tracks = this->tracks(m_direction);
+    struct SizingData {
+        CheckedRef<RenderBox> gridItem;
+        Vector<LayoutUnit> accumulatedSubgridMarginBorderPadding;
+    };
 
-        if (CheckedPtr inner = dynamicDowncast<RenderGrid>(*gridItem); inner && inner->isSubgridInParentDirection(iterator.direction())) {
-            // Contribute the mbp of wrapper to the first and last tracks that we span.
-            GridSpan subgridSpan = downcast<RenderGrid>(inner->parent())->gridSpanForGridItem(*inner, iterator.direction());
-            auto accumulatedMbpWithSubgrid = currentAccumulatedMbp + computeSubgridMarginBorderPadding(m_renderGrid, m_direction, track, trackIndex, span, inner.get());
-            track.setBaseSize(std::max(track.baseSize(), accumulatedMbpWithSubgrid + extraMarginFromSubgridAncestorGutters(*gridItem, span, trackIndex, iterator.direction()).value_or(0_lu)));
+    Vector<SizingData> sizingDataStack;
+    sizingDataStack.append({ *m_renderGrid->firstInFlowChildBox(), Vector<LayoutUnit>(tracks.size()) });
 
-            GridIterator subgridIterator = GridIterator::createForSubgrid(*inner, iterator, subgridSpan);
+    while (!sizingDataStack.isEmpty()) {
+        auto sizingData = sizingDataStack.takeLast();
+        if (auto* nextGridItem = sizingData.gridItem->nextInFlowSiblingBox())
+            sizingDataStack.append({ *nextGridItem, sizingData.accumulatedSubgridMarginBorderPadding });
 
-            accumulateIntrinsicSizesForTrack(track, trackIndex, subgridIterator, itemsSortedByIncreasingSpan, itemsCrossingFlexibleTracks, itemsSet, accumulatedMbpWithSubgrid, gridLayoutState);
+        CheckedRef gridItem = sizingData.gridItem;
+        if (m_renderGrid->currentGrid().orderIterator().shouldSkipChild(gridItem.get()))
+            continue;
+
+        auto gridItemSpan = m_renderGrid->gridSpanForGridItem(gridItem, m_direction);
+        bool spansContentSizedTracks = std::ranges::any_of(m_contentSizedTracksIndex, [startLine = gridItemSpan.startLine(), endLine = gridItemSpan.endLine()](auto trackIndex) {
+            return trackIndex >= startLine && trackIndex < endLine;
+        });
+        if (!spansContentSizedTracks)
+            continue;
+
+        if (CheckedPtr subgrid = dynamicDowncast<RenderGrid>(sizingData.gridItem); subgrid && subgrid->isSubgrid(GridLayoutFunctions::flowAwareDirectionForGridItem(*m_renderGrid, *subgrid, m_direction))) {
+            for (auto trackIndex : gridItemSpan) {
+                sizingData.accumulatedSubgridMarginBorderPadding[trackIndex] += computeSubgridMarginBorderPadding(m_renderGrid, m_direction, tracks[trackIndex], trackIndex, gridItemSpan, subgrid.get());
+                auto& track = tracks[trackIndex];
+                track->setBaseSize(std::max(track->baseSize(), sizingData.accumulatedSubgridMarginBorderPadding[trackIndex] + extraMarginFromSubgridAncestorGutters(*subgrid, gridItemSpan, trackIndex, m_direction).value_or(0_lu)));
+            }
+            if (auto* firstChildBox = subgrid->firstChildBox())
+                sizingDataStack.append({ *firstChildBox, sizingData.accumulatedSubgridMarginBorderPadding });
             continue;
         }
 
-        if (!isNewEntry)
-            continue;
-
-        if (spanningItemCrossesFlexibleSizedTracks(span))
-            itemsCrossingFlexibleTracks.append(GridItemWithSpan(*gridItem, span));
-        else if (span.integerSpan() == 1)
-            sizeTrackToFitNonSpanningItem(span, *gridItem, track, gridLayoutState);
-        else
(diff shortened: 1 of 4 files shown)


alanbaradlay (approved):

Source/WebCore/rendering/GridTrackSizingAlgorithm.cpp
```diff
@@ -1742,33 +1742,51 @@ bool GridTrackSizingAlgorithm::shouldExcludeGridItemForMasonryTrackSizing(const
     return shouldExcludeGridItemForMasonryTrackSizing;
 }
 
-void GridTrackSizingAlgorithm::accumulateIntrinsicSizesForTrack(GridTrack& track, unsigned trackIndex, GridIterator& iterator, Vector<GridItemWithSpan>& itemsSortedByIncreasingSpan, Vector<GridItemWithSpan>& itemsCrossingFlexibleTracks, SingleThreadWeakHashSet<RenderBox>& itemsSet, LayoutUnit currentAccumulatedMbp, GridLayoutState& gridLayoutState)
+void GridTrackSizingAlgorithm::aggreateGridItemsForIntrinsicSizing(Vector<GridItemWithSpan>& itemsSortedByIncreasingSpan, Vector<GridItemWithSpan>& itemsCrossingFlexibleTracks, GridLayoutState& gridLayoutState)
```
fantasai: Typo?

Source/WebCore/rendering/GridTrackSizingAlgorithm.cpp
```diff
@@ -1742,33 +1742,51 @@ bool GridTrackSizingAlgorithm::shouldExcludeGridItemForMasonryTrackSizing(const
...
 
-        if (spanningItemCrossesFlexibleSizedTracks(span))
-            itemsCrossingFlexibleTracks.append(GridItemWithSpan(*gridItem, span));
-        else if (span.integerSpan() == 1)
-            sizeTrackToFitNonSpanningItem(span, *gridItem, track, gridLayoutState);
-        else
-            itemsSortedByIncreasingSpan.append(GridItemWithSpan(*gridItem, span));
+        if (CheckedPtr subgrid = dynamicDowncast<RenderGrid>(sizingData.gridItem); subgrid && subgrid->isSubgrid(GridLayoutFunctions::flowAwareDirectionForGridItem(*m_renderGrid, *subgrid, m_direction))) {
+            for (auto trackIndex : gridItemSpan) {
+                sizingData.accumulatedSubgridMarginBorderPadding[trackIndex] += computeSubgridMarginBorderPadding(m_renderGrid, m_direction, tracks[trackIndex], trackIndex, gridItemSpan, subgrid.get());
```
fantasai, line 1774: Would it make sense to pull this out of the for loop, since it only applies to the first/last tracks?

fantasai (approved):

Source/WebCore/rendering/GridTrackSizingAlgorithm.cpp
sammygill: Yes!

Source/WebCore/rendering/GridTrackSizingAlgorithm.cpp
sammygill: Yeah I think so... I tried to do that locally and the test cases weren't very happy.... so I'm gonna do it in an additional patch since I plan to do some follow ups if you don't mind

PR #56085, opened 2026-01-05
Expose CopyOnWrite as part of SharedMemory object mappings

gavin-apple:
#### 649a89a5f4628f38549bf2a18969873a5fe86eaf
<pre>
Expose CopyOnWrite as part of SharedMemory object mappings
<a href="https://bugs.webkit.org/show_bug.cgi?id=304934">https://bugs.webkit.org/show_bug.cgi?id=304934</a>
<a href="https://rdar.apple.com/167545782">rdar://167545782</a>

Reviewed by Simon Fraser.

Expose the ability to map using CopyOnWrite as part of our SharedMemory
implementation. This is then adopted for ShareableBitmap.

Note: This change supports Darwin platforms only. The CopyOnWrite behavior
for Windows &amp; Unix differ when it comes to writes to the mapping, by the
originator, after the receiver requests a private/copy on write
mapping. On Darwin platforms, these changes are caught by CoW and are not
visible in the receiver. On Unix &amp; Windows, these writes are visible by
the receiver.

* Source/WebCore/platform/SharedMemory.h:
* Source/WebCore/platform/android/SharedMemoryAndroid.cpp:
(WebCore::SharedMemory::map):
* Source/WebCore/platform/cocoa/SharedMemoryCocoa.mm:
(WebCore::SharedMemory::map):
* Source/WebCore/platform/graphics/ShareableBitmap.cpp:
(WebCore::ShareableBitmap::create):
(WebCore::ShareableBitmap::createReadOnly):
* Source/WebCore/platform/graphics/ShareableBitmap.h:
* Source/WebCore/platform/unix/SharedMemoryUnix.cpp:
(WebCore::SharedMemory::map):
* Source/WebCore/platform/win/SharedMemoryWin.cpp:
(WebCore::SharedMemory::map):
* Source/WebKit/WebProcess/GPU/graphics/ImageBufferShareableBitmapBackend.cpp:
(WebKit::ImageBufferShareableBitmapBackend::create):
* Tools/TestWebKitAPI/CMakeLists.txt:
* Tools/TestWebKitAPI/Tests/WebCore/ShareableBitmap.cpp: Added.
(TestWebKitAPI::fillTestPattern):
(TestWebKitAPI::expectTestPattern):
(TestWebKitAPI::TEST(ShareableBitmap, ensureCOWBothMapsRWSenderWrite)):
(TestWebKitAPI::TEST(ShareableBitmap, ensureCOWBothMapsRWSenderWriteReceiverWrite)):
(TestWebKitAPI::TEST(ShareableBitmap, ensureCOWBothMapsROSenderWrite)):
* Tools/TestWebKitAPI/Tests/WebCore/SharedMemoryTests.cpp:
(TestWebKitAPI::TEST_P):

Canonical link: <a href="https://commits.webkit.org/305762@main">https://commits.webkit.org/305762@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/3bd4e41e8b885ebfd46a708764ba8fdede4a6469

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/139206 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/11580 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/703 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/147333 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/92273 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/141078 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/12288 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/11730 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/106572 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/77607 "3 flakes 1 failures") | [loading 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/145ee3a6-9a5b-43b0-9aad-eb992279a689) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/142153 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/9304 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/124692 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/87432 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/7c0f7ae7-aa65-4534-a9ed-d59044aec316) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/8855 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/6609 "Passed tests") | [✅ 🛠 wpe-cairo-libwebrtc](https://ews-build.webkit.org/#/builders/166/builds/7627 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/118310 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/588 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/150112 "Built successfully") | | 
| [⏳ 🛠 🧪 jsc-debug-arm64 ](https://ews-build.webkit.org/#/builders/JSC-Tests-O3-Debug-arm64-EWS "Waiting in queue, processing has not started yet") | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/11265 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/606 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/114955 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/11277 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/9536 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/115268 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/29313 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/9309 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/121035 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/66181 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/11307 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/567 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/11042 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/74978 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/11245 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/11095 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/platform/SharedMemory.h
+++ b/Source/WebCore/platform/SharedMemory.h
@@ -113,11 +113,12 @@ class SharedMemory : public ThreadSafeRefCounted<SharedMemory> {
 public:
     using Handle = SharedMemoryHandle;
     using Protection = SharedMemoryProtection;
+    enum class CopyOnWrite : bool { No, Yes };
 
     // FIXME: Change these factory functions to return Ref<SharedMemory> and crash on failure.
     WEBCORE_EXPORT static RefPtr<SharedMemory> allocate(size_t);
     WEBCORE_EXPORT static RefPtr<SharedMemory> copyBuffer(const WebCore::FragmentedSharedBuffer&);
-    WEBCORE_EXPORT static RefPtr<SharedMemory> map(Handle&&, Protection);
+    WEBCORE_EXPORT static RefPtr<SharedMemory> map(Handle&&, Protection, CopyOnWrite = CopyOnWrite::No);
 #if USE(UNIX_DOMAIN_SOCKETS)
     WEBCORE_EXPORT static RefPtr<SharedMemory> wrapMap(void*, size_t, int fileDescriptor);
 #elif OS(DARWIN)
--- a/Source/WebCore/platform/graphics/ShareableBitmap.cpp
+++ b/Source/WebCore/platform/graphics/ShareableBitmap.cpp
@@ -136,9 +136,9 @@ RefPtr<ShareableBitmap> ShareableBitmap::createFromImageDraw(NativeImage& image,
     return bitmap;
 }
 
-RefPtr<ShareableBitmap> ShareableBitmap::create(Handle&& handle, SharedMemory::Protection protection)
+RefPtr<ShareableBitmap> ShareableBitmap::create(Handle&& handle, SharedMemory::Protection protection, SharedMemory::CopyOnWrite copyOnWrite)
 {
-    auto sharedMemory = SharedMemory::map(WTF::move(handle.m_handle), protection);
+    auto sharedMemory = SharedMemory::map(WTF::move(handle.m_handle), protection, copyOnWrite);
     if (!sharedMemory)
         return nullptr;
 
@@ -150,7 +150,7 @@ std::optional<Ref<ShareableBitmap>> ShareableBitmap::createReadOnly(std::optiona
     if (!handle)
         return std::nullopt;
 
-    auto sharedMemory = SharedMemory::map(WTF::move(handle->m_handle), SharedMemory::Protection::ReadOnly);
+    auto sharedMemory = SharedMemory::map(WTF::move(handle->m_handle), SharedMemory::Protection::ReadOnly, defaultCopyOnWrite);
     if (!sharedMemory)
         return std::nullopt;
 
--- a/Source/WebCore/platform/graphics/ShareableBitmap.h
+++ b/Source/WebCore/platform/graphics/ShareableBitmap.h
@@ -47,6 +47,13 @@ class GraphicsContext;
 class Image;
 class NativeImage;
 
+#if OS(DARWIN)
+inline constexpr auto defaultCopyOnWrite = SharedMemory::CopyOnWrite::Yes;
+#else
+// FIXME: https://bugs.webkit.org/show_bug.cgi?id=305633
+inline constexpr auto defaultCopyOnWrite = SharedMemory::CopyOnWrite::No;
+#endif
+
 class ShareableBitmapConfiguration {
 public:
     ShareableBitmapConfiguration() = default;
@@ -151,7 +158,7 @@ public:
     WEBCORE_EXPORT static RefPtr<ShareableBitmap> createFromImageDraw(NativeImage&, const DestinationColorSpace&, const IntSize& destinationSize, const IntSize& sourceSize);
 
     // Create a shareable bitmap from a handle.
-    WEBCORE_EXPORT static RefPtr<ShareableBitmap> create(Handle&&, SharedMemory::Protection = SharedMemory::Protection::ReadWrite);
+    WEBCORE_EXPORT static RefPtr<ShareableBitmap> create(Handle&&, SharedMemory::Protection = SharedMemory::Protection::ReadWrite, SharedMemory::CopyOnWrite = defaultCopyOnWrite);
 
     // Create a shareable bitmap from a ReadOnly handle.
     WEBCORE_EXPORT static std::optional<Ref<ShareableBitmap>> createReadOnly(std::optional<Handle>&&);
--- a/Source/WebKit/WebProcess/GPU/graphics/ImageBufferShareableBitmapBackend.cpp
+++ b/Source/WebKit/WebProcess/GPU/graphics/ImageBufferShareableBitmapBackend.cpp
@@ -92,7 +92,7 @@ std::unique_ptr<ImageBufferShareableBitmapBackend> ImageBufferShareableBitmapBac
 
 std::unique_ptr<ImageBufferShareableBitmapBackend> ImageBufferShareableBitmapBackend::create(const Parameters& parameters, ShareableBitmap::Handle handle)
 {
-    auto bitmap = ShareableBitmap::create(WTF::move(handle));
+    auto bitmap = ShareableBitmap::create(WTF::move(handle), SharedMemory::Protection::ReadWrite, SharedMemory::CopyOnWrite::No);
     if (!bitmap)
         return nullptr;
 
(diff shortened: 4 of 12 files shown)


nikolaszimmermann:
An enum class CopyOnWrite { No, Yes } is more appropriate to use than a bool in the constructor. Otherwise, the callsites that pass true at the moment are hard to understand without looking up the meaning of the ", true" boolean.

smfr (approved):

Source/WebCore/platform/graphics/ShareableBitmap.cpp
shallawa: `RemoteRenderingBackend::cacheNativeImage()` calls this function in GPUProcess. If the WebContentProcess changes the pixels shared buffer, should not GPUProcess copy it on write?

gavin-apple: You're right! I've updated the PR to enforce COW mapping through all ShareableBitmap create paths that take a ShareableBitmapHandle. This should resolve the cacheNativeImage issue. 

There's also some follow-up work todo to enforce on the send side - when the sender is a privileged process.

Source/WebCore/platform/graphics/ShareableBitmap.h
```diff
@@ -47,6 +47,12 @@ class GraphicsContext;
 class Image;
 class NativeImage;
 
+#if OS(DARWIN)
+inline constexpr auto defaultCopyOnWrite = SharedMemory::CopyOnWrite::Yes;
+#else
+inline constexpr auto defaultCopyOnWrite = SharedMemory::CopyOnWrite::No;
```
gavin-apple, line 54: I'm looking for guidance on what to do for non-darwin platforms.

Right now, it seems like the system provided copy-on-write behavior differs on Linux/Windows from what we get on XNU. On XNU, if we map a page as copy=true, when an edit happens on the sender or receiver side the copy is triggered and those modifications aren't visible to the other party. On Linux/Windows, it appears to only occur when the receiver mapping the entity makes a modification. This is ~supported by man 2 mmap:

> Updates to the mapping are not visible to other processes mapping the same file, and are not carried through to the underlying file. It is unspecified whether changes made to the file after the mmap() call are visible in the mapped region.

The current PR just implements the behavior on Darwin and ASSERTs that we never pass CopyOnWrite::Yes on other platforms. I'm currently confirming that there's no other (kernel assisted) way to achieve the same behavior on Linux/Windows. If there isn't, would we prefer to keep this for Darwin only? Or implement a more explicit copy implementation (with presumably extra memory overhead) for other platforms?

Source/WebCore/platform/SharedMemory.h
shallawa, line 116: I do not have a strong opinion here but maybe using `Protection` in an `OptionSet` and adding `CopyOnWrite` as an option will make the code a little bit cleaner:
```
enum class SharedMemoryProtection : uint8_t { 
    Read  = 1 << 0, 
    Write = 1 << 1,
    Copy  = 1 << 2
};
```
I see `mach_vm_map()` spilt the `CopyOnWrite` from the `vmProtection` as you do in this patch. So your approach is fine also.

Source/WebKit/WebProcess/GPU/graphics/ImageBufferShareableBitmapBackend.cpp
shallawa, line 95: This can be
```
{ SharedMemory::Protection::Read, SharedMemory::Protection::Write }
```

Source/WebCore/platform/graphics/ShareableBitmap.h
shallawa: I think this is fine since the other platforms do not support the `CopyOnWrite` in the way we expect for fixing this bug. I think filling bugs for Windows and Linux and adding `FIXME:` comment here referencing these bugs should be okay.

mcatanzaro: Well it's presumably not possible to ever fix it, so I prefer the proposal to assert that `CopyOnWrite::Yes` is never used.

PR #56122, opened 2026-01-06
PlatformSpeechSynthesizer::initializeVoiceList() should not block on macOS

minorninth:
#### 92ee2ee16c0c555965ce4f7c969d185ac029c3a0
<pre>
PlatformSpeechSynthesizer::initializeVoiceList() should not block on macOS
<a href="https://bugs.webkit.org/show_bug.cgi?id=304944">https://bugs.webkit.org/show_bug.cgi?id=304944</a>
<a href="https://rdar.apple.com/164447645">rdar://164447645</a>

Reviewed by Darin Adler and Tyler Wilcock.

Uses a new asynchronous variant of the internal method to fetch the voice list.

* LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list-expected.txt: Added.
* LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list.html: Added.
* LayoutTests/fast/speechsynthesis/voices-non-mock.html:
* Source/WebCore/Configurations/AllowedSPI-legacy.toml:
* Source/WebCore/PAL/pal/spi/cocoa/AXSpeechManagerSPI.h:
* Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm:
(WebCore::PlatformSpeechSynthesizer::initializeVoiceList):
* Source/WebCore/platform/mock/PlatformSpeechSynthesizerMock.cpp:
(WebCore::PlatformSpeechSynthesizerMock::initializeVoiceList):
* Source/WebCore/platform/mock/PlatformSpeechSynthesizerMock.h:
(WebCore::PlatformSpeechSynthesizerMock::setInitialVoiceListToEmpty):
(WebCore::Internals::simulateSpeechSynthesizerVoiceListChange):
(WebCore::Internals::setInitialVoiceListToEmpty):
* Source/WebCore/testing/Internals.h:
* Source/WebCore/testing/Internals.idl:

Canonical link: <a href="https://commits.webkit.org/305553@main">https://commits.webkit.org/305553@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/e14ad6b5efd967a5f6c3f11e49d4dc033dcb8251

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/138702 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/11068 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/184 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/146816 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/91677 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/140575 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/11772 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/11222 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/106134 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/77446 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/141649 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/8872 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/124307 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/87004 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/8460 "Passed tests") | [❌ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/6214 "Found 1 new API test failure: TestWebKitAPI.NowPlayingControlsTests.NowPlayingUpdatesThrottled (failure)") | [✅ 🛠 wpe-cairo-libwebrtc](https://ews-build.webkit.org/#/builders/166/builds/7114 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/117881 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/148 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/149572 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/10750 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/149 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/114525 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/10768 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/9094 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/114859 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/29202 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/8707 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/120605 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/65645 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/10798 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/149 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/10533 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/74439 "Built successfully") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/10736 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/10587 "Built successfully") | | |

### DIFF
--- a/LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list.html
+++ b/LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list.html
@@ -1,0 +1,36 @@
+<!DOCTYPE html>
+<html>
+<head>
+<script src="../../resources/js-test.js"></script>
+</head>
+<body>
+<script>
+    description("This tests that the voice list can initially be empty and populated after voiceschanged event.");
+
+    window.jsTestIsAsync = true;
+    if (window.testRunner)
+        testRunner.waitUntilDone();
+
+    if (window.internals) {
+        internals.enableMockSpeechSynthesizer();
+        internals.setInitialVoiceListToEmpty();
+
+        initialVoices = speechSynthesis.getVoices();
+        shouldBe("initialVoices.length", "0");
+
+        speechSynthesis.onvoiceschanged = () => {
+            voices = speechSynthesis.getVoices();
+            shouldBeTrue("voices.length > 0");
+            finishJSTest();
+        };
+
+        setTimeout(() => {
+            internals.simulateSpeechSynthesizerVoiceListChange();
+        }, 0);
+    } else {
+        debug("Test requires window.internals to run.");
+        finishJSTest();
+    }
+</script>
+</body>
+</html>
--- a/Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm
+++ b/Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm
@@ -38,6 +38,7 @@
 
 #import <pal/spi/cocoa/AXSpeechManagerSPI.h>
 #import <wtf/BlockObjCExceptions.h>
+#import <wtf/MainThread.h>
 #import <wtf/RetainPtr.h>
 
 #import <pal/cocoa/AVFoundationSoftLink.h>
@@ -293,13 +294,31 @@ void PlatformSpeechSynthesizer::initializeVoiceList()
         return;
 
     BEGIN_BLOCK_OBJC_EXCEPTIONS
-    // SpeechSynthesis replaces on-device compact with higher quality compact voices. These
-    // are not available to WebKit so we're losing these default voices for WebSpeech.
-    // Only show built-in voices when requesting through WebKit to reduce fingerprinting surface area.
-    for (AVSpeechSynthesisVoice *voice in [PAL::getAVSpeechSynthesisVoiceClassSingleton() speechVoicesIncludingSuperCompact]) {
-        if (voice.isSystemVoice)
-            m_voiceList.append(PlatformSpeechSynthesisVoice::create(voice.identifier, voice.name, voice.language, true, true));
+
+    Class avSpeechSynthesisVoiceClass = PAL::getAVSpeechSynthesisVoiceClassSingleton();
+    auto appendVoices = [protectedThis = Ref { *this }](NSArray<AVSpeechSynthesisVoice *> *voices) {
+        for (AVSpeechSynthesisVoice *voice in voices) {
+            if (voice.isSystemVoice)
+                protectedThis->m_voiceList.append(PlatformSpeechSynthesisVoice::create(voice.identifier, voice.name, voice.language, /* localService */ true, /* isDefault */ true));
+        }
+    };
+
+    // Support older OS versions that don't have the asynchronous version het.
+    // Remove this once 26.3 is the minimum OS version supported by Safari
+    if (![avSpeechSynthesisVoiceClass respondsToSelector:@selector(speechVoicesIncludingSuperCompactWithCompletionHandler:)]) {
+        appendVoices([avSpeechSynthesisVoiceClass speechVoicesIncludingSuperCompact]);
+        return;
     }
+
+    [avSpeechSynthesisVoiceClass speechVoicesIncludingSuperCompactWithCompletionHandler:^(NSArray<AVSpeechSynthesisVoice *> *voices) {
+        callOnMainThread([this, protectedThis = Ref { *this }, voices = RetainPtr { voices }, appendVoices]() {
+            BEGIN_BLOCK_OBJC_EXCEPTIONS
+            appendVoices(voices.get());
+            m_speechSynthesizerClient.voicesDidChange();
+            END_BLOCK_OBJC_EXCEPTIONS
+        });
+    }];
+
     END_BLOCK_OBJC_EXCEPTIONS
 }
 
--- a/LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list-expected.txt
+++ b/LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list-expected.txt
@@ -1,0 +1,11 @@
+This tests that the voice list can initially be empty and populated after voiceschanged event.
+
+On success, you will see a series of "PASS" messages, followed by "TEST COMPLETE".
+
+
+PASS initialVoices.length is 0
+PASS voices.length > 0 is true
+PASS successfullyParsed is true
+
+TEST COMPLETE
+
--- a/LayoutTests/fast/speechsynthesis/voices-non-mock.html
+++ b/LayoutTests/fast/speechsynthesis/voices-non-mock.html
@@ -8,24 +8,43 @@
 <script>
     description("This tests that we can get synthesizer voices.");
 
+    window.jsTestIsAsync = true;
+    if (window.testRunner)
+        testRunner.waitUntilDone();
+
     var speech = window.speechSynthesis;
-    var list = speech.getVoices();
+    var timeoutID;
 
-    var foundDefaultVoice = false;
-    var foundEnglishVoice = false;
-    var voiceCount = list.length;
+    function runTest() {
+        clearTimeout(timeoutID);
+        var list = speech.getVoices();
 
-    for (let k = 0; k < list.length; k++) {
-        let voice = list[k];
-        if (voice.lang == "en-US")
-            foundEnglishVoice = true;
-        if (voice.default)
-            foundDefaultVoice = true;
+        foundDefaultVoice = false;
+        foundEnglishVoice = false;
+        voiceCount = list.length;
+
+        for (let k = 0; k < list.length; k++) {
+            let voice = list[k];
+            if (voice.lang == "en-US")
+                foundEnglishVoice = true;
+            if (voice.default)
+                foundDefaultVoice = true;
+        }
+
+        shouldBeTrue(`voiceCount >= ${window.internals.minimumExpectedVoiceCount}`);
+        shouldBeTrue("foundEnglishVoice");
+        shouldBeTrue("foundDefaultVoice");
+
+        finishJSTest();
     }
 
-    shouldBeTrue(`voiceCount >= ${window.internals.minimumExpectedVoiceCount}`);
-    shouldBeTrue("foundEnglishVoice");
-    shouldBeTrue("foundDefaultVoice");
+    speech.onvoiceschanged = runTest;
+
+    var list = speech.getVoices();
+    if (list.length > 0)
+        runTest();
+    else
+        timeoutID = setTimeout(runTest, 1000);
 
 </script>
 
--- a/Source/WebCore/Configurations/AllowedSPI-legacy.toml
+++ b/Source/WebCore/Configurations/AllowedSPI-legacy.toml
@@ -115,6 +115,13 @@ selectors = [
 requires = ["USE_APPLE_INTERNAL_SDK"]
 
 [[legacy]]
+# These are needed to only show built-in voices to reduce fingerprinting surface area.
+selectors = [
+    { name = "speechVoicesIncludingSuperCompact", class = "?" },
+    { name = "speechVoicesIncludingSuperCompactWithCompletionHandler:", class = "?" },
+]
+
+[[legacy]]
 classes = [
     "AVAssetCollection",
     "AVContentKeyReportGroup",
@@ -411,7 +418,6 @@ selectors = [
     { name = "sharedTextEffectsWindowForWindowScene:", class = "?" },
     { name = "signature", class = "RSABSSATokenReady" },
     { name = "sizeValue", class = "?" },
-    { name = "speechVoicesIncludingSuperCompact", class = "?" },
     { name = "startRoutingVideoToPictureInPicturePlayerLayerView", class = "?" },
     { name = "startingColumn", class = "?" },
     { name = "startingRow", class = "?" },
--- a/Source/WebCore/PAL/pal/spi/cocoa/AXSpeechManagerSPI.h
+++ b/Source/WebCore/PAL/pal/spi/cocoa/AXSpeechManagerSPI.h
@@ -41,9 +41,12 @@ DECLARE_SYSTEM_HEADER
 
 #include <AVFoundation/AVFoundation.h>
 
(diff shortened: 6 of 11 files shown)


Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm
```diff
@@ -296,10 +296,22 @@ - (void)speechSynthesizer:(AVSpeechSynthesizer *)synthesizer willSpeakRangeOfSpe
     // SpeechSynthesis replaces on-device compact with higher quality compact voices. These
     // are not available to WebKit so we're losing these default voices for WebSpeech.
     // Only show built-in voices when requesting through WebKit to reduce fingerprinting surface area.
+
+#if PLATFORM(MAC)
+    [PAL::getAVSpeechSynthesisVoiceClassSingleton() speechVoicesIncludingSuperCompactWithCompletionHandler:^(NSArray<AVSpeechSynthesisVoice *> *voices) {
+        for (AVSpeechSynthesisVoice *voice in voices) {
```
twilco, line 300: Should this be a RetainPtr? Maybe we can let the safer-cpp tell us if it's necessary (but I'd assume so)

Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm
```diff
@@ -296,10 +296,22 @@ - (void)speechSynthesizer:(AVSpeechSynthesizer *)synthesizer willSpeakRangeOfSpe
...
+            if (voice.isSystemVoice)
+                m_voiceList.append(PlatformSpeechSynthesisVoice::create(voice.identifier, voice.name, voice.language, true, true));
```
twilco: Could be nice to annotate these bools with comments, e.g. `/* localService */ true, /* isDefault */ true`

Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm
```diff
@@ -296,10 +296,22 @@ - (void)speechSynthesizer:(AVSpeechSynthesizer *)synthesizer willSpeakRangeOfSpe
     // SpeechSynthesis replaces on-device compact with higher quality compact voices. These
     // are not available to WebKit so we're losing these default voices for WebSpeech.
     // Only show built-in voices when requesting through WebKit to reduce fingerprinting surface area.
+
+#if PLATFORM(MAC)
+    [PAL::getAVSpeechSynthesisVoiceClassSingleton() speechVoicesIncludingSuperCompactWithCompletionHandler:^(NSArray<AVSpeechSynthesisVoice *> *voices) {
```
smfr: Are we sure the callback is called on the main thread?

smfr: Also, won't making this async cause the first JS call to retrieve the voice list to return an empty response? This is sync from the web process to the UI process, so I don't see how making this async won't break things. Is there a test that shows this?

Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm
minorninth: I ended up adding a RetainPtr on voices when calling on the main thread so that should address it.

Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm
minorninth: Good catch, switched to dispatch to the main thread.

Yes, the first JS call will return an empty array. Thankfully that's what Chromium has done for years, so I think the chance of breaking anything is low. I just confirmed with a test page, Chromium also returns 0 initially and fires onvoiceschanged a moment later, so Safari's implementation will match.

It'd be worth considering pre-initializing this list on startup, but I think that'd be a more risky change, as that would mean that any instability in the system could affect all pages rather than the relatively small subset that use speech synthesis.

darinadler (approved):

Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm
smfr: It's good to know that our behavior will match chrome now. Should we test that "initially empty" followed by  `onvoiceschanged` behavior in a test?

minorninth: Done. I added a new mock-based test, and also updated another test that was covering the live behavior to handle this case.

Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm
```diff
@@ -296,10 +297,27 @@ - (void)speechSynthesizer:(AVSpeechSynthesizer *)synthesizer willSpeakRangeOfSpe
...
+        }];
+        return;
+    }
+#endif
+
     for (AVSpeechSynthesisVoice *voice in [PAL::getAVSpeechSynthesisVoiceClassSingleton() speechVoicesIncludingSuperCompact]) {
         if (voice.isSystemVoice)
-            m_voiceList.append(PlatformSpeechSynthesisVoice::create(voice.identifier, voice.name, voice.language, true, true));
+            m_voiceList.append(PlatformSpeechSynthesisVoice::create(voice.identifier, voice.name, voice.language, /* localService */ true, /* isDefault */ true));
     }
```
darinadler: Can we refactor so the loop is shared rather than written twice? We could have a lambda that takes an `NSArray<AVSpeechSynthesisVoice *> *` and call it in two places.

minorninth: Good idea

Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm
darinadler, line 322: Realized that if we need this "block Objective-C exceptions" behavior then we might need to do it inside the block passed to `callOnMainThread`.

minorninth: Oh, good catch. Definitely safer to add it there.

LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list.html
```diff
@@ -0,0 +1,40 @@
...
+        shouldBe("initialVoices.length", "0");
+
+        speechSynthesis.onvoiceschanged = () => {
+            voices = speechSynthesis.getVoices();
+            shouldBeTrue("voices.length > 0");
+            finishJSTest();
+        };
+
+        setTimeout(() => {
+            if (window.internals)
```
twilco: No need to check window.internals here, we already did so in the enclosing if

LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list.html
```diff
@@ -0,0 +1,40 @@
...
+<html>
+<head>
+<script src="../../resources/js-test-pre.js"></script>
+</head>
+<body>
+<script>
+    description("This tests that the voice list can initially be empty and populated after voiceschanged event.");
+
+    if (window.internals)
+        window.internals.enableMockSpeechSynthesizer();
```
twilco: Any reason not to do this in the main window.internals if-block?

twilco (approved):

LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list.html
```diff
@@ -0,0 +1,40 @@
+<!DOCTYPE html>
+<html>
+<head>
+<script src="../../resources/js-test-pre.js"></script>
```
darinadler: New tests should use just `js-test.js`, not `js-test-pre.js` and `js-test-post.js`.
```suggestion
<script src="../../resources/js-test.js"></script>
```

LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list.html
darinadler: As long as we are nitpicking, once we have checked the existence of `window.internals`, it’s fine to just write `internals`.
```suggestion
        internals.enableMockSpeechSynthesizer();
```

LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list.html
```diff
@@ -0,0 +1,40 @@
...
+
+    if (window.internals)
+        window.internals.enableMockSpeechSynthesizer();
+
+    window.jsTestIsAsync = true;
+    if (window.testRunner)
+        testRunner.waitUntilDone();
+
+    if (window.internals) {
+        window.internals.setInitialVoiceListToEmpty();
```
darinadler:
```suggestion
        internals.setInitialVoiceListToEmpty();
```

LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list.html
```diff
@@ -0,0 +1,40 @@
...
+
+        speechSynthesis.onvoiceschanged = () => {
+            voices = speechSynthesis.getVoices();
+            shouldBeTrue("voices.length > 0");
+            finishJSTest();
+        };
+
+        setTimeout(() => {
+            if (window.internals)
+                window.internals.simulateSpeechSynthesizerVoiceListChange();
```
darinadler:
```suggestion
            internals.simulatespeechSynthesizerVoiceListChange();
```

LayoutTests/fast/speechsynthesis/speech-synthesis-empty-initial-voice-list.html
```diff
@@ -0,0 +1,40 @@
...
+        }, 0);
+    } else {
+        debug("Test requires window.internals to run.");
+        finishJSTest();
+    }
+</script>
+<script src="../../resources/js-test-post.js"></script>
```
darinadler:
```suggestion
```

Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm
darinadler, line 298: Should have a blank line before this since there’s a blank line before `END_BLOCK_OBJC_EXCEPTIONS` below.

darinadler (approved):

darinadler (approved):

Source/WebCore/platform/cocoa/PlatformSpeechSynthesizerCocoa.mm
darinadler: I have one last thought about how to future-proof this a little more, not urgent to do this.

Long term we will want to just call `speechVoicesIncludingSuperCompactWithCompletionHandler:` and get rid of the old code calling `speechVoicesIncludingSuperCompact`. So we could reverse the way this code is written to make it read that way:
```Objective-C++
    // Support older OS versions that do not yet have the asynchronous method.
    if (![speechVoiceClass respondsToSelector:@selector(speechVoicesIncludingSuperCompactWithCompletionHandler:)]) {
        appendVoices([speechVoiceClass speechVoicesIncludingSuperCompact]);
        return;
    }

    [speechVoiceClass speechVoicesIncludingSuperCompactWithCompletionHandler:^(NSArray<AVSpeechSynthesisVoice *> *voices) {
        callOnMainThread([this, protectedThis = Ref { *this }, voices = RetainPtr { voices }, appendVoices]() {
            BEGIN_BLOCK_OBJC_EXCEPTIONS
            appendVoices(voices.get());
            m_speechSynthesizerClient.voicesDidChange();
            END_BLOCK_OBJC_EXCEPTIONS
        });
    }];
```

minorninth: May as well do it now while we're working on it. I added a comment with the minimum required OS version too so it will be more clear when we can consider removing it.

PR #56346, opened 2026-01-09
[MSE] Fix sample DTS to prevent deletion of preexisting samples when PTS doesn't conflict

eocanha:
#### 52d9299455a3bc8ad61035abe8771fc1b2399854
<pre>
[MSE] Fix sample DTS to prevent deletion of preexisting samples when PTS doesn&apos;t conflict
<a href="https://bugs.webkit.org/show_bug.cgi?id=305208">https://bugs.webkit.org/show_bug.cgi?id=305208</a>

Reviewed by NOBODY (OOPS!).

Let&apos;s consider a preexisting samples A, B, C, and a new sample N with the
following timestamps:

       Decoding time  Presentation time
       -------------  -----------------
Sample A: 827.826164000  827.826164000
Sample B: 827.826164100  827.826164100
Sample C: 827.826164200  827.826164200
Sample N: 827.826125000  827.909542000

These disparities in DTS vs. PTS offsets between samples A and B exist
because both samples come from completely different videos (eg: because
of ad insertion).

In a timeline:

DTS ···NABC·······
PTS ····ABC···N···

Judging by PTS, the samples A, B, C should remain and theoretically don&apos;t
need to be erased. However, considering DTS, the samples A, B, C have to
be erased, because the sample N would reach the video decoder earlier (by
decoding ordering) and change the state of the decoder, making it
unsuitable for samples A, B, C.

This commit tries to fix the situation by changing sample N (the new
sample) DTS to a value slightly higher than sample C DTS (that is, DTS plus
epsilon, a small value), like this:

       Decoding time  Presentation time
       -------------  -----------------
Sample A:  827.826164000  827.826164000
Sample B:  827.826164100  827.826164100
Sample C:  827.826164200  827.826164200
Sample N: *827.826164201* 827.909542000

DTS ····ABCN······
PTS ····ABC···N···

This fix avoids the deletion of samples A, B, C and the potential
problems associated with that.

A layout test has been added to show the problem and prove that the fix
solves it.

* LayoutTests/media/media-source/media-source-samples-out-of-order-erase-last-previous-frames-expected.txt: Added.
* LayoutTests/media/media-source/media-source-samples-out-of-order-erase-last-previous-frames.html: Added.
* Source/WebCore/platform/graphics/SourceBufferPrivate.cpp:
(WebCore::SourceBufferPrivate::processMediaSample): If a new sync sample is being added and there are next samples in decode order with PTS lower than the one of the new sample, fix the DTS of the new sample by assigning it to have a safeDecodeTime (the DTS of the highest (by DTS order) preexisting sample, plus epsilon), but only if that DTS wouldn&apos;t be higher than the DTS of the next sample that should come after the new sample being added (estimated as sample.decodeTime() + 2*duration, and keeping a safety margin).
</pre>

### DIFF
--- a/LayoutTests/media/media-source/media-source-samples-out-of-order-erase-last-previous-frames-expected.txt
+++ b/LayoutTests/media/media-source/media-source-samples-out-of-order-erase-last-previous-frames-expected.txt
@@ -1,0 +1,23 @@
+
+RUN(video.src = URL.createObjectURL(source))
+EVENT(sourceopen)
+RUN(sourceBuffer = source.addSourceBuffer("video/mock; codecs=mock"))
+First segment has normal, monotonically increasing samples with decode timestamp equal to presentation timestamp.
+RUN(sourceBuffer.appendBuffer(mediaSegment))
+EVENT(updateend)
+EXPECTED (bufferedSamples.length == '4') OK
+{PTS({0/10 = 0.000000}), DTS({0/10 = 0.000000}), duration({10/10 = 1.000000}), flags(1), generation(0)}
+{PTS({10/10 = 1.000000}), DTS({10/10 = 1.000000}), duration({10/10 = 1.000000}), flags(0), generation(0)}
+{PTS({20/10 = 2.000000}), DTS({20/10 = 2.000000}), duration({10/10 = 1.000000}), flags(0), generation(0)}
+{PTS({30/10 = 3.000000}), DTS({30/10 = 3.000000}), duration({10/10 = 1.000000}), flags(0), generation(0)}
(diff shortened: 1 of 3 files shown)


eocanha:
Some of the detected test failures are caused by the fact that edit list support is disabled in WebKitGTK, and this is creating small offsets that trigger unnecessary "out-of-order frame detected" events.

I've also tested the patch [in downstream wpe-2.38](https://github.com/WebPlatformForEmbedded/WPEWebKit/tree/eocanha/eocanha-debug-160) (with edit lists support), running the layout test there, and the mediasource-config-change-mp4-v-framerate.html test at least passes, but video decoding errors appear. This is enough evidence to consider the fix as problematic.

I'm withdrawing this PR by now.

PR #56662, opened 2026-01-15
dNR: fails to match separator `^` to end of URL in `urlFilter`

dotproto:
#### 708061a0763a677c5fc4de2ec77969225f26c964
<pre>
dNR: fails to match separator `^` to end of URL in `urlFilter`
<a href="https://bugs.webkit.org/show_bug.cgi?id=297838">https://bugs.webkit.org/show_bug.cgi?id=297838</a>

Reviewed by NOBODY (OOPS!).

Previously, the separator character (`^`) only matched when an allowed character
was present; it did not match the end of the URL as expected. This change
adjusts the regular expression used internally to either match an allowed
character or the end of the URL.

Test: Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm

* Source/WebKit/UIProcess/Extensions/Cocoa/_WKWebExtensionDeclarativeNetRequestRule.mm:
(-[_WKWebExtensionDeclarativeNetRequestRule
_regexURLFilterForChromeURLFilter:]): Updated internal RegExp used to match the
separator character.
* Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm:
(TestWebKitAPI::TEST(WKWebExtensionAPIDeclarativeNetRequest,
RuleConversionWithURLFilterAndRequestDomains)): Added a new test case to check
separator character handling.
(TestWebKitAPI::TEST(WKWebExtensionAPIDeclarativeNetRequest,
URLFilterSpecialCharacters)): Updated existing test case to match the new RegExp
pattern.
</pre>
https://github.com/WebKit/WebKit/commit/7947544d74b88e63a1bc28bdbcf815b835d5f11d

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/141259 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/13643 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/2957 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/149833 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/94356 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/143132 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/14354 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/13796 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/108524 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/78565 "2 flakes 8 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/144210 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/11072 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/126415 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/89429 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/10647 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/8256 "Passed tests") | [⏳ 🛠 wpe-libwebrtc ](https://ews-build.webkit.org/#/builders/WPE-LibWebRTC-Build-EWS "Waiting in queue, processing has not started yet") | 
| | [❌ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/119908 "Found 1 new API test failure: TestWebKitAPI.WKWebExtensionAPIDeclarativeNetRequest.URLFilterSpecialCharacterRuleList (failure)") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/2391 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/152227 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/13329 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/2835 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/116622 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/13345 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/11638 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/116962 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/13008 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/123067 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/68503 "The change is no longer eligible for processing.") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/13372 "Built successfully") | [❌ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/2447 "5 flakes 2 failures") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/13111 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/77078 "Built successfully") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/13310 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/13155 "Built successfully") | | |

### DIFF
--- a/Source/WebKit/UIProcess/Extensions/Cocoa/_WKWebExtensionDeclarativeNetRequestRule.mm
+++ b/Source/WebKit/UIProcess/Extensions/Cocoa/_WKWebExtensionDeclarativeNetRequestRule.mm
@@ -1163,10 +1163,10 @@ static BOOL isArrayOfRequestMethodsValid(NSArray<NSString *> *requestMethods)
     // Documentation: https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/
 
     // Supported special charcters:
-    // '*' : Wildcard: Matches any number of characters.
-    // '|' : Left/right anchor: If used at either end of the pattern, specifies the beginning/end of the url respectively.
+    // '*'  : Wildcard: Matches any number of characters.
+    // '|'  : Left/right anchor: If used at either end of the pattern, specifies the beginning/end of the url respectively.
     // '||' : Domain name anchor: If used at the beginning of the pattern, specifies the start of a (sub-)domain of the URL.
-    // '^' : Separator character: This matches anything except a letter, a digit or one of the following: _ - . %.
+    // '^'  : Separator character: This matches anything except a letter, a digit or one of the following: _ - . %. This also match the end of the URL.
 
     // Therefore urlFilter is composed of the following parts: (optional Left/Domain name anchor) + pattern + (optional Right anchor).
     // All other regex special charaters are escaped in the pattern.
@@ -1183,10 +1183,22 @@ static BOOL isArrayOfRequestMethodsValid(NSArray<NSString *> *requestMethods)
     if (hasEndAnchor)
         chromeURLFilter = [chromeURLFilter substringToIndex:chromeURLFilter.length - 1];
 
+    BOOL hasEndSeparator = [chromeURLFilter hasSuffix:@"^"];
+    if (hasEndSeparator)
+        chromeURLFilter = [chromeURLFilter substringToIndex:chromeURLFilter.length - 1];
+
     NSString *regexFilter = escapeCharactersInString(chromeURLFilter, @"?+[(){}$|\\.");
 
     regexFilter = [regexFilter stringByReplacingOccurrencesOfString:@"*" withString:@".*"];
+
     regexFilter = [regexFilter stringByReplacingOccurrencesOfString:@"^" withString:@"[^a-zA-Z0-9_.%-]"];
+
+    if (hasEndSeparator) {
+        if (hasEndAnchor)
+            regexFilter = [regexFilter stringByAppendingString:@"([^a-zA-Z0-9_.%-])?"];
+        else
+            regexFilter = [regexFilter stringByAppendingString:@"([^a-zA-Z0-9_.%-].*)?$"];
+    }
 
     if (hasDomainNameAnchor)
         regexFilter = [@"^[^:]+://+([^:/]+\\.)?" stringByAppendingString:regexFilter];
--- a/Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm
+++ b/Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm
@@ -29,9 +29,11 @@
 
 #import "HTTPServer.h"
 #import "TestNavigationDelegate.h"
+#import "TestResourceLoadDelegate.h"
 #import "WebExtensionUtilities.h"
 #import <WebKit/WKPreferencesPrivate.h>
 #import <WebKit/_WKFeature.h>
+#import <WebKit/_WKResourceLoadInfo.h>
 #import <WebKit/_WKWebExtensionDeclarativeNetRequestRule.h>
 #import <WebKit/_WKWebExtensionDeclarativeNetRequestTranslator.h>
 
@@ -2882,8 +2884,8 @@ TEST(WKWebExtensionAPIDeclarativeNetRequest, RuleConversionWithURLFilterAndReque
                 @"resource-type": @[ @"script" ],
             },
 #if ENABLE(DNR_ON_RULE_MATCHED_DEBUG)
-        @"_identifier": @1,
-        @"_rulesetIdentifier": @"Test Ruleset",
+            @"_identifier": @1,
+            @"_rulesetIdentifier": @"Test Ruleset",
 #endif
         };
 
@@ -2906,6 +2908,9 @@ TEST(WKWebExtensionAPIDeclarativeNetRequest, RuleConversionWithURLFilterAndReque
     testPattern(@"com", @"://www.*", @"://www\\..*com");
     testPattern(@"com", @"://www./foo/bar", @"://www\\..*com/foo/bar");
     testPattern(@"com", @"://www.*/foo/bar", @"://www\\..*com/foo/bar");
+
+    // URL filter contains a separator character
+    testPattern(@"apple.com", @"||apple.com/*.css^", @"^[^:]+://+([^:/]+\\.)?apple\\.com/.*\\.css([^a-zA-Z0-9_.%-].*)?$");
 
     // Concatenating the request domain and URL filter
     testPattern(@"com", @"/foo/bar", @"^[^:]+://+([^:/]+\\.)?com.*/foo/bar");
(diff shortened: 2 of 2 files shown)


Source/WebKit/UIProcess/Extensions/Cocoa/_WKWebExtensionDeclarativeNetRequestRule.mm
```diff
@@ -1186,7 +1186,7 @@ - (NSString *)_regexURLFilterForChromeURLFilter:(NSString *)chromeURLFilter
     NSString *regexFilter = escapeCharactersInString(chromeURLFilter, @"?+[(){}$|\\.");
 
     regexFilter = [regexFilter stringByReplacingOccurrencesOfString:@"*" withString:@".*"];
-    regexFilter = [regexFilter stringByReplacingOccurrencesOfString:@"^" withString:@"[^a-zA-Z0-9_.%-]"];
+    regexFilter = [regexFilter stringByReplacingOccurrencesOfString:@"^" withString:@"(?:[^a-zA-Z0-9_.%-]|$)"];
```
xeenon: Unfortunately, WebKit does not support `|` in the regex for content blocker rules. I think the right fix here would be to check for "^" at the suffix on the input, and then simply append "$" at the end of the output. It is only important at the end of the expression, so that should work.

xeenon (changes requested):

Source/WebKit/UIProcess/Extensions/Cocoa/_WKWebExtensionDeclarativeNetRequestRule.mm
xeenon: More details about what is supported [here](https://webkit.org/blog/3476/content-blockers-first-look/#:~:text=It%20is%20possible%20to%20use%20the%20beginning%20of%20line%20(“^”)%20and%20end%20of%20line%20(“$”)%20marker%20but%20they%20are%20restricted%20to%20be%20the%20first%20and%20last%20character%20of%20the%20expression.%20For%20example%2C%20a%20pattern%20like%20“^bar$”%20is%20perfectly%20valid%2C%20while%20“(foo)?^bar$”%20causes%20a%20syntax%20error.).

dotproto: Addressed in https://github.com/WebKit/WebKit/pull/56662/commits/fd7fe1e44c0699a2592fcf82b995ea0d939db13a.

Source/WebKit/UIProcess/Extensions/Cocoa/_WKWebExtensionDeclarativeNetRequestRule.mm
```diff
@@ -1183,10 +1183,15 @@ - (NSString *)_regexURLFilterForChromeURLFilter:(NSString *)chromeURLFilter
...
 
+    BOOL hasEndSeparator = [chromeURLFilter hasSuffix:@"^"];
+
     NSString *regexFilter = escapeCharactersInString(chromeURLFilter, @"?+[(){}$|\\.");
 
     regexFilter = [regexFilter stringByReplacingOccurrencesOfString:@"*" withString:@".*"];
+
     regexFilter = [regexFilter stringByReplacingOccurrencesOfString:@"^" withString:@"[^a-zA-Z0-9_.%-]"];
+    if (hasEndSeparator)
+        regexFilter = [regexFilter stringByAppendingString:@"?"];
```
xeenon: I think this is what is needed:

```objc
BOOL hasEndSeparator = [chromeURLFilter hasSuffix:@"^"];
if (hasEndSeparator)
    chromeURLFilter = [chromeURLFilter substringToIndex:chromeURLFilter.length - 1];

NSString *regexFilter = escapeCharactersInString(chromeURLFilter, @"?+[(){}$|\\.");
regexFilter = [regexFilter stringByReplacingOccurrencesOfString:@"*" withString:@".*"];
regexFilter = [regexFilter stringByReplacingOccurrencesOfString:@"^" withString:@"[^a-zA-Z0-9_.%-]"];

if (hasEndSeparator)
    regexFilter = [regexFilter stringByAppendingString:@"([^a-zA-Z0-9_.%-].*)?$"];
```

Importantly the `.*` and `$` additions.

Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm
```diff
@@ -3491,6 +3496,142 @@ static void runRedirectRule(bool useEnhancedSecurity)
...
+        @"        try { await fetch(`${base}/123`) } catch(e) { }",
+        @"        try { await fetch(`${base}/example?123`) } catch(e) { }",
+        @"        try { await fetch(`${base}/example?1234`) } catch(e) { }",
+        @"        browser.test.sendMessage('Requests Complete');",
+        @"    }",
+        @"    fetchSubresources();",
+        @"</script>",
+    ]);
+
+
```
xeenon, line 3604:
```suggestion

```

Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm
```diff
@@ -3491,6 +3496,142 @@ static void runRedirectRule(bool useEnhancedSecurity)
...
+    });
+
+    Vector<String> expectedBlockedUrls {
+        "/1.txt"_s,
+        "/2.txt"_s,
+        "/2.txt?"_s,
+        "/2.txt?foo"_s,
+        "/example?123"_s,
+    };
+    Vector<String> expectedAllowedUrls {
```
xeenon:
```suggestion
    };

    Vector<String> expectedAllowedUrls {
```

Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm
```diff
@@ -3491,6 +3496,142 @@ static void runRedirectRule(bool useEnhancedSecurity)
...
+        "/"_s,
+        "/101.txt"_s,
```
xeenon: Consider adding these tests.

```suggestion
        "/101.txt"_s,
        "/2.txtfoo"_s,
        "/2.txt%20"_s,
```

xeenon (changes requested):

Source/WebKit/UIProcess/Extensions/Cocoa/_WKWebExtensionDeclarativeNetRequestRule.mm
dotproto: Is the use of `.*` here meant to cover arbitrary contents after the `^` in the urlPattern? 

My understanding was that `^` matches a single character at a given position in the match string, but that character is optional at the end of a URL. Unfortunately, Chrome's docs are ambiguous:

> '^' : Separator character: This matches anything except a letter, a digit, or one of the following: `_`, `-`, `.`, or `%`. This also match the end of the URL.

It's not clear to me if "end of the URL" here means that the final character is optional or that anything after this character is optional. ~In my testing it appears that Firefox uses the former and Chrome seems to use the latter.~

EDIT: I was mistaken. Chrome and Firefox both use the latter reading, which seems consistent with @xeenon's intent in suggestion `.*`.

dotproto: I've updated this logic as suggested and further itterated on it in https://github.com/WebKit/WebKit/commit/c46127bdb78f7bf9881ee5bad03894e3cd89cc9a.

Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm
dotproto: Fixed in https://github.com/WebKit/WebKit/commit/c46127bdb78f7bf9881ee5bad03894e3cd89cc9a

Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm
dotproto: I've added these and many other tests in https://github.com/WebKit/WebKit/commit/c46127bdb78f7bf9881ee5bad03894e3cd89cc9a

Source/WebKit/UIProcess/Extensions/Cocoa/_WKWebExtensionDeclarativeNetRequestRule.mm
```diff
@@ -1163,10 +1163,10 @@ - (NSString *)_regexURLFilterForChromeURLFilter:(NSString *)chromeURLFilter
     // Documentation: https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/
 
     // Supported special charcters:
-    // '*' : Wildcard: Matches any number of characters.
-    // '|' : Left/right anchor: If used at either end of the pattern, specifies the beginning/end of the url respectively.
+    // '*'  : Wildcard: Matches any number of characters.
+    // '|'  : Left/right anchor: If used at either end of the pattern, specifies the beginning/end of the url respectively.
     // '||' : Domain name anchor: If used at the beginning of the pattern, specifies the start of a (sub-)domain of the URL.
-    // '^' : Separator character: This matches anything except a letter, a digit or one of the following: _ - . %.
+    // '^'  : Separator character: This matches anything except a letter, a digit or one of the following: _ - . %.
```
xeenon: Update to match Chrome docs.
```suggestion
    // '^'  : Separator character: This matches anything except a letter, a digit or one of the following: _ - . %.  This also match the end of the URL.
```

Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm
```diff
@@ -3491,6 +3496,202 @@ static void runRedirectRule(bool useEnhancedSecurity)
...
+        std::optional<bool> wasAllowed;
+    };
+
+    auto *rules = @(R"JSON(
+        [
+            { "id" : 1, "priority": 1, "action" : { "type" : "block" }, "condition" : { "urlFilter" : "||localhost:*/path/file.txt^" } },
+            { "id" : 2, "priority": 1, "action" : { "type" : "block" }, "condition" : { "urlFilter" : "example*^123|" } },
+            { "id" : 3, "priority": 1, "action" : { "type" : "block" }, "condition" : { "urlFilter" : "example*456|" } },
+            { "id" : 4, "priority": 1, "action" : { "type" : "block" }, "condition" : { "urlFilter" : "separatorEnd^|" } },
+            { "id" : 5, "priority": 1, "action" : { "type" : "block" }, "condition" : { "urlFilter" : "endSeparator|^" } }
```
xeenon: Doing `|^` should likely be an error, since `|` is only allowed and special at the start or end. This `|` will just be escaped in the regex and literally match a `|` in the URL. I'd drop or rework this test case.

Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm
```diff
@@ -3491,6 +3496,202 @@ static void runRedirectRule(bool useEnhancedSecurity)
...
+        { "/example/path/456?"_s, true, std::nullopt },
+        { "/example456"_s, false, std::nullopt },
+        // urlFilter with terminal separator charactor and right anchor (^|)
+        { "/separatorEnd"_s, false, std::nullopt },
+        { "/separatorEnd?"_s, false, std::nullopt },
+        { "/separatorEnd?allowed"_s, true, std::nullopt },
+        // urlFilter with terminal right anchor and separator charactor (|^)
+        { "/endSeparator"_s, true, std::nullopt },
+        { "/endSeparator?"_s, true, std::nullopt },
+        { "/endSeparator?allowed"_s, true, std::nullopt },
```
xeenon: Ditto.

xeenon (changes requested):

Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm
dotproto: By "`|^` should likely … error," do you mean that we should throw when parsing rule 5? If so, I agree that it's likely a mistake by the rule author, but I don't think we should error because it's still a valid rule declaration. Also, this rule is parsed without error by the other browsers.

While iterating on tests related to `|^` rule, I observed inconsistent behavior in how all browsers are handling percent encoding and urlFilter matching. I've included the URL filters and results observations here. The full source of the test extension can be found [in this gist](https://gist.github.com/dotproto/df3a92c1e8075ccfddd824c152064658).

```json
  { "id" : 1, "priority": 1, "action" : { "type" : "block" }, "condition" : { "urlFilter" : "/literal|^" } },
  { "id" : 2, "priority": 1, "action" : { "type" : "block" }, "condition" : { "urlFilter" : "/encoded%7C^" } }
```

| Request | Safari 26.2 | Firefox 141 | Chrome 144 | https://github.com/WebKit/WebKit/commit/c46127bdb78f7bf9881ee5bad03894e3cd89cc9a | 
| --- | --- | --- | --- | --- | 
| `fetch('/literal%7C')` | ✅ allowed | ✅ allowed | ✅ allowed | ✅ allowed  |
| `fetch('/literal%7C+')` | ✅ allowed | ✅ allowed | ✅ allowed | ✅ allowed  |
| `fetch('/literal%7C%2B')` | ✅ allowed | ✅ allowed | ✅ allowed | ✅ allowed  |
| `fetch('/literal\|')` | ✅ allowed | ❌ blocked | ✅ allowed | ❌ blocked |
| `fetch('/literal\|+')` | ❌ blocked | ❌ blocked | ✅ allowed | ❌ blocked |
| `fetch('/literal\|%2B')` | ✅ allowed | ✅ allowed | ✅ allowed | ✅ allowed  |
| `fetch('/encoded%7C')` | ✅ allowed | ❌ blocked | ❌ blocked | ❌ blocked |
| `fetch('/encoded%7C+')` | ❌ blocked | ❌ blocked | ❌ blocked | ❌ blocked |
| `fetch('/encoded%7C%2B')` | ✅ allowed | ✅ allowed | ✅ allowed | ✅ allowed  |
| `fetch('/encoded\|')` | ✅ allowed | ✅ allowed | ❌ blocked | ✅ allowed  |
| `fetch('/encoded\|+')` | ✅ allowed | ✅ allowed | ❌ blocked | ✅ allowed  |
| `fetch('/encoded\|%2B')` | ✅ allowed | ✅ allowed | ✅ allowed | ✅ allowed  |

The last column here shows the most recent commit in this PR at the this comment was written. I chose to use `+` in these tests because it is another character that should match `^` and it *may* be percent encoded the HTTP request. Actually, that appears to be the key point of difference this Firefox/this PR and Chrome; based on what's shown in DevTools, it appears that Chrome always percent encodes `+` while the other browsers do not.

As of https://github.com/WebKit/WebKit/commit/c46127bdb78f7bf9881ee5bad03894e3cd89cc9a (yesterday's revision), the `URLFilterSpecialCharacterRuleList` test I added only behaves as expected when test case URLs are percent encoded. As such, I'll update this PR shortly to cover the equivalent of the `encoded` urlFilter (rule 1) and the `encoded%7C` request URL.

I'm open to continuing to iterate on this test and urlFilter handling more generally, but I think it may be better to do so in a separate PR as I think we've addressed the specific issue identified in [bug 297838](https://bugs.webkit.org/show_bug.cgi?id=297838).

xeenon: This is great! Yes, I agree any other changes here would be better in a new PR or raised as issues for discussion in the WECG.

Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm
```diff
@@ -3491,6 +3496,204 @@ static void runRedirectRule(bool useEnhancedSecurity)
...
+
+    [mainRequestScriptLines addObjectsFromArray:@[
+        @"        browser.test.sendMessage('Requests Complete');",
+        @"    }",
+        @"    fetchSubresources();",
+        @"</script>",
+    ]];
+
+    auto mainRequestBody = Util::constructScript(mainRequestScriptLines);
+    NSLog(@"mainRequestBody %@", mainRequestBody);
```
xeenon:
```suggestion
```

xeenon (approved):

Tools/TestWebKitAPI/Tests/WebKitCocoa/WKWebExtensionAPIDeclarativeNetRequest.mm
dotproto: Oops, just fixed this in https://github.com/WebKit/WebKit/pull/56662/commits/7947544d74b88e63a1bc28bdbcf815b835d5f11d.

PR #56682, opened 2026-01-16
[Site Isolation] Block setting src of cross-origin frames/iframes to javascript: urls

atar13:
#### 65450223b6e9342cf5acaad151ed32cf8cc72b44
<pre>
[Site Isolation] Block setting src of cross-origin frames/iframes to javascript: urls
<a href="https://bugs.webkit.org/show_bug.cgi?id=305615">https://bugs.webkit.org/show_bug.cgi?id=305615</a>
<a href="https://rdar.apple.com/168267776">rdar://168267776</a>

Reviewed by Sihui Liu.

This patch modifies HTMLFrameElementBase::canLoadURL
to indicate that a javascript: url cannot be loaded
on a cross-origin frame which is in a different process.

This fixes several tests in LayoutTests/http/tests/security/javascriptURL/
which attempt to set the src attribute of cross-origin frames/iframes
to javascript: urls (via several different methods).
The tests expect setting the src attribute of a cross-origin frame/iframe
to a javascript: url to be blocked and report a security error.
However, previously, with site isolation, no security error
was reported.

This patch blocks and prints a security error when checking if a
javascript: URL can be loaded in a RemoteFrame (happens in
HTMLFrameElementBase::canLoadURL).
With site isolation off, HTMLFrameElementBase::canLoadURL
calls ScriptController::canAccessFromCurrentOrigin
to check if origins matched. If they didn&apos;t, a SecurityError
was printed.

See step 4 of Section 7.4.2.3.2 of the HTML navigation spec
which describes the special case of navigating to javascript
urls.
<a href="https://html.spec.whatwg.org/#the-javascript">https://html.spec.whatwg.org/#the-javascript</a>:-url-special-case

        If initiatorOrigin is not same origin-domain with
        targetNavigable&apos;s active document&apos;s origin, then return.

Previously, when site isolation was enabled,
contentDocument was null and canLoadURL didn&apos;t immediately
return false and print a SecurityError. Now, whenever the
target frame is detected to be a RemoteFrame, we print a
SecurityError since the frame is of a different origin and
in a different process.

This patch also updates ScriptController::canAccessFromCurrentOrigin
to work on both RemoteFrames and LocalFrames.

Note:
HTMLFrameElementBase::canLoadURL is called by
HTMLFrameElementBase::openURL which is called
by HTMLFrameElementBase::setLocation which is called
by the following test cases which attempt
to set the src attribute in various ways.

No new tests, but this patch fixes the following tests with
site isolation enabled:
LayoutTests/http/tests/security/javascriptURL/javascriptURL-execution-context-frame-src-getAttribute-value.html
LayoutTests/http/tests/security/javascriptURL/javascriptURL-execution-context-frame-src-htmldom.html
LayoutTests/http/tests/security/javascriptURL/javascriptURL-execution-context-frame-src-setAttribute.html
LayoutTests/http/tests/security/javascriptURL/javascriptURL-execution-context-frame-src-setAttributeNS.html
LayoutTests/http/tests/security/javascriptURL/javascriptURL-execution-context-frame-src-setAttributeNode.html
LayoutTests/http/tests/security/javascriptURL/javascriptURL-execution-context-frame-src-setAttributeNodeNS.html
LayoutTests/http/tests/security/javascriptURL/javascriptURL-execution-context-iframe-src-getAttribute-value.html
LayoutTests/http/tests/security/javascriptURL/javascriptURL-execution-context-iframe-src-htmldom.html
LayoutTests/http/tests/security/javascriptURL/javascriptURL-execution-context-iframe-src-setAttribute.html
LayoutTests/http/tests/security/javascriptURL/javascriptURL-execution-context-iframe-src-setAttributeNS.html
LayoutTests/http/tests/security/javascriptURL/javascriptURL-execution-context-iframe-src-setAttributeNode.html
LayoutTests/http/tests/security/javascriptURL/javascriptURL-execution-context-iframe-src-setAttributeNodeNS.html

* Source/WebCore/bindings/js/JSDOMBindingSecurity.cpp:
(WebCore::canAccessTargetOrigin):
(WebCore::canAccessFrame):
(WebCore::canAccessDocument):
(WebCore::BindingSecurity::shouldAllowAccessToFrame):
* Source/WebCore/bindings/js/JSDOMBindingSecurity.h:
* Source/WebCore/bindings/js/ScriptController.cpp:
(WebCore::ScriptController::canAccessFromCurrentOrigin):
* Source/WebCore/bindings/js/ScriptController.h:
* Source/WebCore/html/HTMLFrameElementBase.cpp:
(WebCore::HTMLFrameElementBase::canLoadURL const):

Canonical link: <a href="https://commits.webkit.org/308353@main">https://commits.webkit.org/308353@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/4cf652721ba0b4f6cd00ae31b009791ca23088a3

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/147095 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/19776 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/13366 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/155777 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/100509 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/148969 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/20234 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/19676 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/113355 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/80870 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/150057 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/15577 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/132148 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94113 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/308e6f0d-2f49-4934-bdd9-da089bd14c77) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/14783 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/12565 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/3219 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/124370 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/10075 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/158108 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/1239 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/11515 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/121379 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/19577 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/16421 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/121580 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/19586 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/131832 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/75545 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/22697 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/17125 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/8642 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/19192 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/82947 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/18922 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/19073 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/18981 "Built successfully") | | | |


Source/WebCore/html/HTMLFrameElementBase.cpp
```diff
@@ -80,6 +82,15 @@ bool HTMLFrameElementBase::canLoadURL(const URL& completeURL) const
         RefPtr contentDocument = this->contentDocument();
         if (contentDocument && !ScriptController::canAccessFromCurrentOrigin(contentDocument->protectedFrame().get(), protectedDocument().get()))
             return false;
+        else if (RefPtr remoteFrame = dynamicDowncast<RemoteFrame>(contentFrame())) {
+            Ref targetFrameOrigin = remoteFrame->frameDocumentSecurityOriginOrOpaque();
+            Ref accessingDocumentOrigin = document().securityOrigin();
+
+            if (!accessingDocumentOrigin->isSameOriginDomain(targetFrameOrigin)) {
+                document().addConsoleMessage(MessageSource::JS, MessageLevel::Error, remoteFrame->window().crossDomainAccessErrorMessage(*document().window(), IncludeTargetOrigin::Yes));
+                return false;
```
atar13: Does it make more sense to fold this into `ScriptController::canAccessFromCurrentOrigin` or make a similar function for `RemoteFrame`s in `ScriptController`?

Source/WebCore/html/HTMLFrameElementBase.cpp
```diff
@@ -80,6 +81,13 @@ bool HTMLFrameElementBase::canLoadURL(const URL& completeURL) const
         RefPtr contentDocument = this->contentDocument();
         if (contentDocument && !ScriptController::canAccessFromCurrentOrigin(contentDocument->protectedFrame().get(), protectedDocument().get()))
             return false;
+        else if (RefPtr remoteFrame = dynamicDowncast<RemoteFrame>(contentFrame())) {
```
szewai: The direction seems right, but I think the right fix might be updating `ScriptController::canAccessFromCurrentOrigin` to take a `Frame` instead of `LocalFrame` (you can get the remote frame's document security origin with `Frame::frameDocumentSecurityOrigin()`)

atar13: For the RemoteFrame case, should we still check that origins match? Or should we always block access to remote frames as they are in a different process?

szewai: I think we should check for origin. "different process" is technically an implementation choice.
If spec says we need to support cross-origin access, then we should put the frames in the same process (or have some way to support the function across processes), not "because we put them in different processes, so we cannot support this"

Source/WebCore/bindings/js/JSDOMBindingSecurity.cpp
```diff
@@ -72,28 +72,22 @@ static void reportErrorAccessingRemoteFrame(JSC::JSGlobalObject* lexicalGlobalOb
...
     case ThrowSecurityError: {
         VM& vm = lexicalGlobalObject->vm();
         auto scope = DECLARE_THROW_SCOPE(vm);
-        throwSecurityError(*lexicalGlobalObject, scope, targetDocument->window()->crossDomainAccessErrorMessage(active, IncludeTargetOrigin::No));
+        throwSecurityError(*lexicalGlobalObject, scope, targetDOMWindow->crossDomainAccessErrorMessage(active, IncludeTargetOrigin::No));
         break;
     }
     case LogSecurityError:
-        printErrorMessageForFrame(targetDocument->frame(), targetDocument->window()->crossDomainAccessErrorMessage(active, IncludeTargetOrigin::Yes));
+        active.printErrorMessage(targetDOMWindow->crossDomainAccessErrorMessage(active, IncludeTargetOrigin::Yes));
```
atar13, line 90: I am not 100% sure what to do here. 

The previous code calls:
`printErrorMessageForFrame` 
which runs:
`frame->document()->window()->printErrorMessage(message);`

However, `RemoteFrame` doesn't have an accessible `document()` so `frame->document()->window()->printErrorMessage(message)` wouldn't work.

Instead, I was thinking of calling `targetDOMWindow.printErrorMessage` where `targetDOMWindow` could be a `RemoteDOMWindow` (`targetWindow` comes from `targetFrame->window()` in `canAccessFrame` where `targetFrame` is a `RemoteFrame`), however when I tried that the expected security error didn't get printed out to the console (run-webkit-tests didn't pick it up).

For now, I am printing from the context of the "active" document but the previous implementation printed from the window of the `targetDocument->frame()`.

Should I figure out why printing from a `RemoteDOMWindow` doesn't work? It seems wrong to change the source of the prints.

Source/WebCore/html/HTMLFrameElementBase.cpp
atar13: Updated `ScriptController::canAccessFromCurrentOrigin` to work on `Frame`s in https://github.com/WebKit/WebKit/commit/3db2a8f81a322896ecf02e5c88a347409be82a83

Source/WebCore/bindings/js/JSDOMBindingSecurity.cpp
szewai: Checking the implementation, it seems the cause is `DOMWindow::console()` returns null for remote window.
I feel like using the console of active window is the right behavior, as from developer's perspective it might be useful to see the log at where the code that causes the violation is running.
@achristensen07 What do you think?

achristensen07: I think switching to active is an improvement.

szewai (approved):

PR #56811, opened 2026-01-19
[visionOS] WebAVPlayerController leaks when playing a video in Safari one window and entering tab overview in another window

ddkilzer:
#### 30ae065f0e43e2a0f52667ffc1520bb74cec19c7
<pre>
[visionOS] WebAVPlayerController leaks when playing a video in Safari one window and entering tab overview in another window
&lt;<a href="https://bugs.webkit.org/show_bug.cgi?id=305752">https://bugs.webkit.org/show_bug.cgi?id=305752</a>&gt;
&lt;<a href="https://rdar.apple.com/121550335">rdar://121550335</a>&gt;

Reviewed by Geoffrey Garen.

Tests: Tools/TestWebKitAPI/Tests/WebCore/cocoa/WebAVPlayerControllerLeakTests.mm

Fix the following bugs:

1. Fix leak of WebAVPlayerController object by adding
   -[WebAVPlayerControllerForwarder dealloc] method that clears the
   _playerController instance variable.  When using RetainPtr&lt;&gt; as an
   ivar, the ObjC runtime does not auto-generate -.cxx_construct or
   -.cxx_destruct methods, which caused the leak.
2. Fix addition of the _playerController instance variable to dynamic
   WebAVPlayerControllerForwarder_AVKitCompatible class by calling
   class_addIvar() before objc_registerClassPair().
3. Simplify WebAVPlayerControllerForwarder by replacing RetainPtr with a
   direct instance variable for _playerController, implementing
   getter/setter methods for _playerController that work for both the
   original and dynamic classes by computing the ivar location at
   runtime, then using a getter/setter in all other instance methods
   instead of direct instance variable access.  Using direct instance
   variable access would read from/write to the wrong location in the
   WebAVPlayerControllerForwarder_AVKitCompatible object since
   _playerController was in a different location.  This removes the need
   for custom methods in the dynamic class.

* Source/WebCore/platform/ios/WebAVPlayerController.h:
(WebCore::createWebAVPlayerController):
(WebCore::webAVPlayerControllerClassSingleton):
- Move into WebCore namespace.
- Export webAVPlayerControllerClassSingleton() for testing.
* Source/WebCore/platform/ios/WebAVPlayerController.mm:
(-[WebAVPlayerControllerForwarder playerController]):
(-[WebAVPlayerControllerForwarder setPlayerController:]):
- Add @property definition to document playerController instance
  variable with getter and setter.
- Change _playerController ivar from RetainPtr&lt;&gt; to raw pointer for
  simplicity and avoid pointer math in getter and setter.
- Implement getter and setter so that they work for both
  WebAVPlayerControllerForwarder and the dynamically constructed class.
(-[WebAVPlayerControllerForwarder init]):
- Update to use setter instead of direct ivar access.
(-[WebAVPlayerControllerForwarder dealloc]): Add.
- Use setter to clear _playerController ivar.
(-[WebAVPlayerControllerForwarder respondsToSelector:]):
- Update to use getter.
(-[WebAVPlayerControllerForwarder forwardingTargetForSelector:]):
- Update to use getter.
(-[WebAVPlayerControllerForwarder _forwardingTargetForKeyPath:]):
- Update to use getter.
- Use RetainPtr&lt;&gt; for local variables and optimize with autorelease()
  and WTF::move().
(createWebAVPlayerControllerForwarderClassSingleton):
- Call class_addIvar() before objc_registerClassPair() to make sure
  _playerController ivar is created properly.
- Ignore &quot;.&quot; methods when overriding AVPlayerController methods.
(WebCore::createWebAVPlayerController):
(WebCore::webAVPlayerControllerClassSingleton):
- Move into WebCore namespace.

* Tools/TestWebKitAPI/SourcesCocoa.txt:
- Add Tests/WebCore/cocoa/WebAVPlayerControllerLeakTests.mm.
- Add Tests/WebCore/cocoa/WebAVPlayerControllerLeakTests.mm.
* Tools/TestWebKitAPI/Tests/WebCore/cocoa/WebAVPlayerControllerLeakTests.mm: Add.
(TestWebKitAPI::TEST(WebAVPlayerControllerLeak, NormalForwarderClassReleasesIvar)): Add.
(TestWebKitAPI::TEST(WebAVPlayerControllerLeak, DynamicForwarderClassReleasesIvar)): Add.
- Add tests to verify correct instance object behavior.
* Tools/TestWebKitAPI/Tests/WebKitCocoa/EnhancedSecurity.mm:
* Tools/TestWebKitAPI/Tests/WebKitCocoa/IPCTestingAPI.mm:
* Tools/TestWebKitAPI/Tests/WebKitCocoa/PasteWebArchive.mm:
* Tools/TestWebKitAPI/Tests/WebKitCocoa/ScreenTime.mm:
* Tools/TestWebKitAPI/Tests/WebKitCocoa/TextExtractionTests.mm:
- Add missing headers after UnifiedSources reshuffle.

Canonical link: <a href="https://commits.webkit.org/306145@main">https://commits.webkit.org/306145@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/d1af943dace9d695df8e1c30736f7dea09c41bef

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/140458 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/12840 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/1971 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/148787 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/93545 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b73e7b18-30df-423a-8291-c9cc8a50769c) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/142331 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/13552 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/12994 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/107680 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/93545 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [loading 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/d7767468-19e0-4be7-b02f-110b3722535b) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/143409 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/10435 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/125733 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/88580 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/d8003248-b646-41f8-9843-1d09e4ea016a) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/10051 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/7610 "Passed tests") | [✅ 🛠 wpe-cairo-libwebrtc](https://ews-build.webkit.org/#/builders/166/builds/8891 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/119290 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/1752 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/151412 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/12528 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/1825 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/115983 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/12543 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/10747 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/116321 "Passed tests") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/11579 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/122236 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/67551 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/21678 "Built successfully and passed tests") | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/12570 "Built successfully") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/1706 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/12310 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/76270 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/12508 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/12354 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/platform/ios/WebAVPlayerController.mm
+++ b/Source/WebCore/platform/ios/WebAVPlayerController.mm
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2014-2018 Apple Inc. All rights reserved.
+ * Copyright (C) 2014-2026 Apple Inc. All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -61,10 +61,45 @@ static double WebAVPlayerControllerLiveStreamMinimumTargetDuration = 1.0; // Min
 static double WebAVPlayerControllerLiveStreamSeekableTimeRangeMinimumDuration = 30.0;
 
 @interface WebAVPlayerControllerForwarder : NSObject
+@property (readwrite, strong) WebAVPlayerController *playerController;
 @end
 
 @implementation WebAVPlayerControllerForwarder {
-    RetainPtr<WebAVPlayerController> _playerController;
+    SUPPRESS_UNRETAINED_MEMBER WebAVPlayerController *_playerController;
+}
+
+// IMPORTANT: The -init and -dealloc methods below are copied to the dynamically created
+// WebAVPlayerControllerForwarder_AVKitCompatible class, which inherits from AVPlayerController.
+// When these methods call [super init] or [super dealloc], the compiler generates calls to
+// objc_msgSendSuper2, which stores a reference to the COMPILE-TIME class (WebAVPlayerControllerForwarder)
+// in the __objc_superrefs section. At runtime, objc_msgSendSuper2 looks up that class's superclass
+// (NSObject), NOT the dynamic class's superclass (AVPlayerController).
+//
+// This means:
+// - [super init] in the copied -init calls NSObject's -init, skipping AVPlayerController's -init
+// - [super dealloc] in the copied -dealloc calls NSObject's -dealloc, skipping AVPlayerController's -dealloc
+//
+// This is intentional: we use AVPlayerController as a superclass only to satisfy Swift's type system
+// at runtime, but we never actually initialize or use AVPlayerController's functionality.
+
+// Use object_getInstanceVariable/object_setInstanceVariable so these methods work for both
+// this class and the dynamically created WebAVPlayerControllerForwarder_AVKitCompatible subclass
+// (which has _playerController at a different offset due to inheriting from AVPlayerController).
+
+- (WebAVPlayerController *)playerController
+{
+    void *value = nullptr;
+    object_getInstanceVariable(self, "_playerController", &value);
+    return [[(__bridge WebAVPlayerController *)value retain] autorelease];
+}
+
+- (void)setPlayerController:(WebAVPlayerController *)playerController
+{
+    void *oldValue = nullptr;
+    object_getInstanceVariable(self, "_playerController", &oldValue);
+    [playerController retain];
+    object_setInstanceVariable(self, "_playerController", (__bridge void *)playerController);
+    [(__bridge id)oldValue release];
 }
 
 - (instancetype)init
@@ -73,19 +108,26 @@ static double WebAVPlayerControllerLiveStreamSeekableTimeRangeMinimumDuration = 
     if (!self)
         return nil;
 
-    _playerController = adoptNS([[WebAVPlayerController alloc] init]);
+    self.playerController = adoptNS([[WebAVPlayerController alloc] init]).get();
 
     return self;
 }
 
+- (void)dealloc
+{
+    self.playerController = nil;
+    [super dealloc];
+}
+
 - (BOOL)respondsToSelector:(SEL)aSelector
 {
-    return [_playerController respondsToSelector:aSelector];
+    return [self.playerController respondsToSelector:aSelector];
 }
 
 - (id)forwardingTargetForSelector:(SEL)aSelector
 {
-    return _playerController.get();
+    UNUSED_PARAM(aSelector);
+    return self.playerController;
 }
 
 - (id)valueForKey:(NSString *)key
@@ -122,21 +164,22 @@ static double WebAVPlayerControllerLiveStreamSeekableTimeRangeMinimumDuration = 
 {
     // Only use the proxy object for key paths identifying undefined properties on WebAVPlayerController.
 
-    NSObject *target = [_playerController playerControllerProxy];
+    RetainPtr<WebAVPlayerController> playerController = self.playerController;
+    RetainPtr<NSObject> target = [playerController playerControllerProxy];
 
     NSString *propertyNameFromKeyPath = [[keyPath componentsSeparatedByString:@"."] firstObject];
     if (!propertyNameFromKeyPath.length)
-        return target;
+        return target.autorelease();
 
-    auto properties = class_copyPropertyListSpan([_playerController class]);
+    auto properties = class_copyPropertyListSpan([playerController class]);
     for (auto& property : properties.span()) {
         NSString *propertyName = [NSString stringWithUTF8String:property_getName(property)];
         if ([propertyNameFromKeyPath isEqualToString:propertyName]) {
-            target = _playerController.get();
+            target = WTF::move(playerController);
             break;
         }
     }
-    return target;
+    return target.autorelease();
 }
 
 @end
@@ -150,15 +193,26 @@ static Class createWebAVPlayerControllerForwarderClassSingleton()
     Class superClass = getAVPlayerControllerClassSingleton();
     Class implClass = [WebAVPlayerControllerForwarder class];
     Class newClass = objc_allocateClassPair(superClass, "WebAVPlayerControllerForwarder_AVKitCompatible", 0);
+
+    // Add ivar BEFORE registering the class pair (required by ObjC runtime).
+    class_addIvar(newClass, "_playerController", sizeof(WebAVPlayerController *), log2(sizeof(WebAVPlayerController *)), @encode(WebAVPlayerController *));
+
     objc_registerClassPair(newClass);
 
-    // Remove all of AVPlayerController's methods.
+    // Override all of AVPlayerController's methods with unknownMethodImp
+    // (except internal '.' methods like .cxx_destruct).
     auto methods = class_copyMethodListSpan(superClass);
     IMP unknownMethodImp = class_getMethodImplementation(superClass, NSSelectorFromString(@"_web_unknownMethod"));
-    for (auto& method : methods.span())
-        class_addMethod(newClass, method_getName(method), unknownMethodImp, method_getTypeEncoding(method));
+    for (auto& method : methods.span()) {
+        SEL selector = method_getName(method);
+        if ([NSStringFromSelector(selector) hasPrefix:@"."])
+            continue;
+        class_addMethod(newClass, selector, unknownMethodImp, method_getTypeEncoding(method));
+    }
 
     // Copy methods from WebAVPlayerControllerForwarder.
+    // All methods use object_getInstanceVariable/object_setInstanceVariable for the _playerController
+    // ivar, so they work regardless of ivar offset differences between class hierarchies.
     methods = class_copyMethodListSpan(implClass);
     for (auto& method : methods.span()) {
         SEL selector = method_getName(method);
@@ -167,10 +221,10 @@ static Class createWebAVPlayerControllerForwarderClassSingleton()
         class_replaceMethod(newClass, selector, method_getImplementation(method), method_getTypeEncoding(method));
     }
 
-    class_addIvar(newClass, "_playerController", sizeof(RetainPtr<WebAVPlayerController>), log2(sizeof(RetainPtr<WebAVPlayerController>)), @encode(RetainPtr<WebAVPlayerController>));
-
     return newClass;
 }
+
+namespace WebCore {
 
 RetainPtr<WebAVPlayerController> createWebAVPlayerController()
 {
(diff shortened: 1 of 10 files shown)


ddkilzer:
> > ```diff
> >  - (WebAVPlayerController *)playerController
> >  {
> > -    void *value = nullptr;
> > -    object_getInstanceVariable(self, "_playerController", &value);
> > -    return [[(__bridge WebAVPlayerController *)value retain] autorelease];
> > +    Ivar ivar = class_getInstanceVariable([self class], "_playerController");
> > +    uintptr_t ivarAddress = reinterpret_cast<uintptr_t>(self) + ivar_getOffset(ivar);
> > +    auto& selfPlayerController = *reinterpret_cast<RetainPtr<WebAVPlayerController>*>(ivarAddress);
> > +    return [[selfPlayerController.get() retain] autorelease];
> >  }
> >  
> > -- (void)setPlayerController:(WebAVPlayerController *)controller
> > +- (void)setPlayerController:(WebAVPlayerController *)playerController
> >  {
> > -    void *oldValue = nullptr;
> > -    object_getInstanceVariable(self, "_playerController", &oldValue);
> > -    if (oldValue != (__bridge void *)controller) {
> > -        [(__bridge id)oldValue release];
> > -        object_setInstanceVariable(self, "_playerController", (__bridge void *)[controller retain]);
> > -    }
> > +    Ivar ivar = class_getInstanceVariable([self class], "_playerController");
> > +    uintptr_t ivarAddress = reinterpret_cast<uintptr_t>(self) + ivar_getOffset(ivar);
> > +    auto& selfPlayerController = *reinterpret_cast<RetainPtr<WebAVPlayerController>*>(ivarAddress);
> > +    if (selfPlayerController != playerController)
> > +        selfPlayerController = playerController;
> >  }
> > ```
> 
> I agree the diff you're posting would look worse. _However_, it looks like it being a raw pointer and this extra complexity with `object_getInstanceVariable` is a regression from your PR since on main it is a RetainPtr:
> 
> ```
> @implementation WebAVPlayerControllerForwarder {
>     RetainPtr<WebAVPlayerController> _playerController;
> }
> ```
> 
> So maybe there is a way to fix the bug you're fixing without the added complexity?

I guess I didn't go into enough detail in my commit log message to explain this.

Let me try to explain what's happening before and after this PR in more detail.

### Before this PR

The `WebAVPlayerControllerForwarder` class works as expected with a `RetainPtr<>` instance variable because the compiler generates everything, including `-.cxx_construct` and `-.cxx_destruct`, for it at compile time.

However, the `WebAVPlayerControllerForwarder` class is not actually used anywhere because it's just used as the basis for creating the dynamic `WebAVPlayerControllerForwarder_AVKitCompatible` class at runtime (wherein the method implementations from `WebAVPlayerControllerForwarder` are copied to the dynamically created `WebAVPlayerControllerForwarder_AVKitCompatible` class).

The current issues with the dynamically created `WebAVPlayerControllerForwarder_AVKitCompatible` class are as follows:

1. The `_playerController` instance variable is never added to the `WebAVPlayerControllerForwarder_AVKitCompatible` class at runtime because `class_addIvar()` is called _after_ `objc_registerClassPair()`.

2. The method implementations from `WebAVPlayerControllerForwarder` that are copied to `WebAVPlayerControllerForwarder_AVKitCompatible` are all compiled to assume that `RetainPtr<WebAVPlayerController> _playerController` instance variable is stored at offset +8 in the class.  That's wrong for the dynamic class since it has a superclass defined (that's apparently 720 bytes), so these methods were reading from/writing to the wrong place in memory!

   Even if the `_playerController` instance variable _were_ added to the `WebAVPlayerControllerForwarder_AVKitCompatible` class at runtime, it would be located _after_ the `AVPlayerController` superclass of `WebAVPlayerControllerForwarder_AVKitCompatible`.

   This behavior didn't have any bad side-effects in practice since the `AVPlayerController` superclass was effectively neutered by assigning an unknown implementation methods (see `@"_web_unknownMethod"`) to every instance method it had, so no memory technically got corrupted, although it was certainly not doing what you'd expect.

3. Because the `-.cxx_construct` and `-.cxx_destruct` methods were not copied over from `WebAVPlayerControllerForwarder` to `WebAVPlayerControllerForwarder_AVKitCompatible`, that meant the constructor and destructor for the `RetainPtr<>` instance variable (that was never created) also never ran.

   This is why the leak occurred!  When `WebAVPlayerControllerForwarder_AVKitCompatible` was destroyed, the memory where the `RetainPtr<>` was read from/written to was just deallocated without the destructor running and decrementing the ref count for `_playerController`.

### After this PR

- The `_playerController` instance variable is added to the `WebAVPlayerControllerForwarder_AVKitCompatible` class at runtime because `class_addIvar()` is correctly called _before_ `objc_registerClassPair()`.
- A `@property` is declared for the `_playerController` instance variable.
- Custom getter/setter methods for the `playerController` property are implemented that work correctly for both the `WebAVPlayerControllerForwarder` and `WebAVPlayerControllerForwarder_AVKitCompatible` classes because they dynamically query the ObjC runtime for the instance variable location.
- The instance methods in the `WebAVPlayerControllerForwarder` class are all updated to use property syntax (and thus the custom getter/setter methods) so that all the methods will work correctly when copied over to `WebAVPlayerControllerForwarder_AVKitCompatible`.
- The constructor and destructor for `RetainPtr<WebAVPlayerController> _playerController` don't need to be called since we're using a plain Objective-C object pointer to store the instance variable.  (We could call the constructor/destructor if we really thought using `RetainPtr<>` was was the best option, though it would require the same pointer math and then invoking each manually.)

Source/WebCore/platform/ios/WebAVPlayerController.mm
```diff
@@ -61,10 +61,32 @@
...
+    void *value = nullptr;
+    object_getInstanceVariable(self, "_playerController", &value);
+    return [[(__bridge WebAVPlayerController *)value retain] autorelease];
+}
+
+- (void)setPlayerController:(WebAVPlayerController *)playerController
+{
+    void *oldValue = nullptr;
+    object_getInstanceVariable(self, "_playerController", &oldValue);
+    if (oldValue != (__bridge void *)playerController) {
```
geoffreygaren: If you do the retain before the release, you can avoid the branch.

geoffreygaren (approved):

Source/WebCore/platform/ios/WebAVPlayerController.mm
ddkilzer: I've assumed it was better to avoid refcount churn than branching in ObjC code (since ObjC code is rarely on the hot path).

I'll make the change.

geoffreygaren: It's a minor point, but I prefer avoiding the branch because it eliminates the need to think about a special case.

As far as performance is concerned, you're right that refcount churn is more costly when it happens, but my mental model is that no-op assignment is rare.

PR #56834, opened 2026-01-19
Fix Site Isolation compatibility with Enhanced Security

ronan-apple:
#### 4895f35302d070075756b77cc3e24ce2644cd53b
<pre>
Fix Site Isolation compatibility with Enhanced Security
<a href="https://bugs.webkit.org/show_bug.cgi?id=305328">https://bugs.webkit.org/show_bug.cgi?id=305328</a>
<a href="https://rdar.apple.com/164474301">rdar://164474301</a>

Reviewed by Alex Christensen and Sihui Liu.

With the introduction of Enhanced Security, it became possible to hit a
case where we wanted to swap Enhanced Security states for a site that
already existed within the BrowsingContextGroup. This caused us to hit
an assert where we were trying to add a process for a site that already
existed in the process map.

We should not do this - instead, if we already have an existing process
for a site in the current BrowsingContextGroup, we should remain with
whatever the Enhanced Security state is for that process.

This change allows re-enabling various tests in EnhancedSecurityPolicies
with site isolation enabled, which previously would fail.

An additional test has been added that confirms the expected behaviour
with a WKWebView navigation to the same site as the current main frame,
ensuring this does not change Enhanced Security state.

Test: Tools/TestWebKitAPI/Tests/WebKitCocoa/EnhancedSecurityPolicies.mm

* Tools/TestWebKitAPI/Tests/WebKitCocoa/EnhancedSecurityPolicies.mm:
(testAlertWithEnhancedSecurity):
(runMultiHopThenBack):
(runMultiHopThenBackJavascript):
(runWebViewSameSiteMainFrameNavigation):
(runHttpRedirectsHttpsWithExplicitNavigationToMeaningfulSite):

Canonical link: <a href="https://commits.webkit.org/306717@main">https://commits.webkit.org/306717@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/813046e20bd32146c9287952595cd5514e6dbdd0

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/142071 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/14467 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/4607 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/150686 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/95247 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/d0156bb6-86df-41d7-884e-95d7488135d4) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/143938 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/15185 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/14620 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/109207 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/78937 "Passed tests") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/767da01e-b9c4-4827-a0e5-460491fe2e1a) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/145020 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/11743 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/127175 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/90104 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/41465b28-f95c-4f2c-ae7b-ac4c1c97c2c9) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/11279 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/8940 "Passed tests") | [✅ 🛠 wpe-libwebrtc](https://ews-build.webkit.org/#/builders/172/builds/737 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/120612 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/3476 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/153054 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/14146 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/4111 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/117280 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/14168 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/12342 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/117600 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/29988 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/13652 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/124288 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/69846 "The change is no longer eligible for processing.") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/14195 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/3382 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/13927 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/77911 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/14131 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/13972 "Built successfully") | | | |

### DIFF
--- a/Source/WebKit/UIProcess/WebPageProxy.cpp
+++ b/Source/WebKit/UIProcess/WebPageProxy.cpp
@@ -5163,14 +5163,6 @@ void WebPageProxy::receivedNavigationActionPolicyDecision(WebProcessProxy& proce
     m_isLockdownModeExplicitlySet = (websitePolicies && websitePolicies->isLockdownModeExplicitlySet()) || m_configuration->isLockdownModeExplicitlySet();
     auto lockdownMode = (websitePolicies ? websitePolicies->lockdownModeEnabled() : shouldEnableLockdownMode()) ? WebProcessProxy::LockdownMode::Enabled : WebProcessProxy::LockdownMode::Disabled;
 
-    if (frame.isMainFrame() && preferences->enhancedSecurityHeuristicsEnabled())
-        internals().enhancedSecurityTracker.trackNavigation(navigation, hasOpenedPage());
-
-    auto enhancedSecurity = currentEnhancedSecurityState(websitePolicies.get());
-
-    if (preferences->enhancedSecurityHeuristicsEnabled())
-        protect(this->websiteDataStore())->trackEnhancedSecurityForDomain(RegistrableDomain { navigation.currentRequest().url() }, enhancedSecurity);
-
     Ref browsingContextGroup = m_browsingContextGroup;
     bool usesSameWebsiteDataStore = websiteDataStore.ptr() == &this->websiteDataStore();
     bool mainFrameSiteChanges = !m_mainFrame || Site { m_mainFrame->url() } != Site { navigation.currentRequest().url() };
@@ -5178,6 +5170,17 @@ void WebPageProxy::receivedNavigationActionPolicyDecision(WebProcessProxy& proce
         browsingContextGroup = *targetBackForwardItem->browsingContextGroup();
     else if (processSwapRequestedByClient == ProcessSwapRequestedByClient::Yes || !usesSameWebsiteDataStore || (navigation.isRequestFromClientOrUserInput() && !navigation.isFromLoadData() && mainFrameSiteChanges))
         browsingContextGroup = BrowsingContextGroup::create();
+
+    if (frame.isMainFrame() && preferences->enhancedSecurityHeuristicsEnabled())
+        internals().enhancedSecurityTracker.trackNavigation(navigation, hasOpenedPage());
+
+    auto enhancedSecurity = currentEnhancedSecurityState(websitePolicies.get());
+
+    if (RefPtr process = browsingContextGroup->processForSite(Site { navigation.currentRequest().url() }))
+        enhancedSecurity = process->process().enhancedSecurity();
+
+    if (preferences->enhancedSecurityHeuristicsEnabled())
+        protect(this->websiteDataStore())->trackEnhancedSecurityForDomain(RegistrableDomain { navigation.currentRequest().url() }, enhancedSecurity);
 
     Site site { navigation.currentRequest().url() };
     Site mainFrameSite = frame.isMainFrame() ? site : Site { URL(protect(pageLoadState())->activeURL()) };
--- a/Tools/TestWebKitAPI/Tests/WebKitCocoa/EnhancedSecurityPolicies.mm
+++ b/Tools/TestWebKitAPI/Tests/WebKitCocoa/EnhancedSecurityPolicies.mm
@@ -163,6 +163,9 @@ static void testAlertWithEnhancedSecurity(RetainPtr<TestUIDelegate> uiDelegate, 
 
     EXPECT_WK_STREQ(result[0], message);
     EXPECT_EQ([result[1] boolValue], enhancedSecurityEnabled);
+
+    if ([result[1] boolValue] != enhancedSecurityEnabled)
+        NSLog(@"Enhanced Security state mismatch for alert: %s (expected: %s, actual: %s)", message.utf8().data() , enhancedSecurityEnabled ? "Enabled" : "Disabled", [result[1] boolValue] ? "Enabled" : "Disabled");
 }
 
 static RetainPtr<TestWKWebView> enhancedSecurityTestConfiguration(
@@ -563,8 +566,8 @@ static void runOpenerThenSelfNavigation(bool useSiteIsolation)
         { "navigated"_s, ExpectedEnhancedSecurity::Enabled },
     });
 }
-// FIXME: rdar://164474301 (Fix SiteIsolation compatibility with EnhancedSecurity feature)
-TEST_WITHOUT_SITE_ISOLATION(OpenerThenSelfNavigation)
+
+TEST_WITH_AND_WITHOUT_SITE_ISOLATION(OpenerThenSelfNavigation)
 
 static void runHttpOpeningHttpsNoOpener(bool useSiteIsolation)
 {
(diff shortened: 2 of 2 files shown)


Source/WebKit/UIProcess/BrowsingContextGroup.h
```diff
@@ -91,7 +91,7 @@ class BrowsingContextGroup : public RefCountedAndCanMakeWeakPtr<BrowsingContextG
     HashSet<WebCore::Site> m_sharedProcessSites;
     WeakHashSet<WebPageProxy> m_pagesInSharedProcess;
 
-    HashMap<WebCore::Site, WeakPtr<FrameProcess>> m_processMap;
+    HashMap<std::pair<WebCore::Site, EnhancedSecurity>, WeakPtr<FrameProcess>> m_processMap;
```
szewai: Can you add a new class (e.g. `SiteKey`) for the key instead of using pair? It'll be more clear what each member means
(You need to implement `hash()` to make it usable as key (see `class Site`))

Source/WebKit/UIProcess/BrowsingContextGroup.cpp
```diff
@@ -51,7 +51,7 @@ void BrowsingContextGroup::sharedProcessForSite(WebsiteDataStore& websiteDataSto
 {
     if (!preferences.siteIsolationEnabled() || !preferences.siteIsolationSharedProcessEnabled())
         return completionHandler(nullptr);
-    if (site.isEmpty() || m_processMap.contains(site))
+    if (site.isEmpty() || m_processMap.contains(std::make_pair(site, enhancedSecurity)))
         return completionHandler(nullptr);
     if (!m_sharedProcessSites.contains(site)) {
```
szewai, line 56: A `BrowsingContextGroup` only has one shared process (to be used for multiple sites) now, and it seems it might not work as expected if there are sites in it with different EnhanceSecurity values.

Source/WebKit/UIProcess/BrowsingContextGroup.h
ronan-apple: Ah yes, this is better, thank you! I've named it `EnhancedSecuritySiteKey` just to be more explicit in the name, I hope that suits?

I didn't adjust the BrowsingContextGroup interfaces (e.g. `processForSite` still takes separate Site and EnhancedSecurity instead of a `SiteKey`, which I think may be clearer because things like the `sharedProcessForSite` would perhaps be more confusing to wrap those parameters up in a `SiteKey`) - but if you think otherwise happy to make changes here!

Source/WebKit/UIProcess/BrowsingContextGroup.cpp
ronan-apple: So I took a bit of a look at this - we've sort of adopted the same approach as the `LockdownMode` parameter for this, where the protection is done in `WebProcessCache::takeSharedProcess`, which has these checks:

```
    if (it->value->process().lockdownMode() != lockdownMode) {
        WEBPROCESSCACHE_RELEASE_LOG("takeSharedProcess: cannot take process, lockdown mode not identical", it->value->process().processID());
        return nullptr;
    }

    if (it->value->process().enhancedSecurity() != enhancedSecurity) {
        WEBPROCESSCACHE_RELEASE_LOG("takeSharedProcess: cannot take process, enhanced security not identical", it->value->process().processID());
        return nullptr;
    }
```

So I think if the shared process isn't compatible we won't use it. (I think the way this done is nice though and will handle the case, e.g., where Enhanced Security is _always_ enabled - via API for example - and so the shared process should have Enhanced Security enabled and we'd be fine to use the shared process in that case).

Source/WebKit/UIProcess/EnhancedSecuritySiteKey.h
```diff
@@ -0,0 +1,65 @@
...
+
+#pragma once
+
+#include "EnhancedSecurity.h"
+#include <WebCore/Site.h>
+#include <wtf/HashTraits.h>
+
+namespace WebKit {
+
+class EnhancedSecuritySiteKey {
```
szewai: Can we just call it `SiteKey`? I think we might extend it to include more properties in the future besides `EnhancedSecurity`

szewai (commented):
The PR looks good

achristensen07 (approved):

szewai (approved):

PR #56939, opened 2026-01-21
([Site Isolation] Web Inspector: support cross-origin iframes in the Worker domain)

AnujPanta1:
#### 047f3dc902f24c8e1cb0cc9e67afbe22d7e314f6
<pre>
([Site Isolation] Web Inspector: support cross-origin iframes in the Worker domain)
<a href="https://bugs.webkit.org/show_bug.cgi?id=308488">https://bugs.webkit.org/show_bug.cgi?id=308488</a>
<a href="https://rdar.apple.com/143596746">rdar://143596746</a>

Reviewed by Qianlang Chen.

Introducing FrameWorkerAgent to enable worker discovery and inspection in
cross-origin iframes under site isolation. When the iframe spawns a worker,
InspectorInstrumentation resolves the page-level InstrumentingAgents which
has no connected PageWorkerAgent in that process.

FrameWorkerAgent registers in the shared persistentWorkerAgent slot on the
frame&apos;s InstrumentingAgents at frontend connect time, matching the existing
PageWorkerAgent lifecycle. This works because page and frame WorkerAgents
never coexist on the same InstrumentingAgents instance, and FrameWorkerAgent
inherits from InspectorWorkerAgent. The PageChannel threading model is
inherited from InspectorWorkerAgent — no changes needed.

Under site isolation, PageWorkerAgent early-returns in
connectToAllWorkerInspectorProxies since every frame (including main) has
its own FrameWorkerAgent. FrameWorkerAgent filters
WorkerInspectorProxy::proxiesForPage to only connect workers whose
scriptExecutionContext Document belongs to the inspected frame. This
prevents double-connection when multiple same-site frames share a WebProcess.

instrumentingAgents(ScriptExecutionContext&amp;) now prefers the frame&apos;s
InstrumentingAgents when a document has an associated frame. Since the
frame&apos;s agents fall back to the page&apos;s agents for unset slots, the existing
workerStarted, workerTerminated, and shouldWaitForDebuggerOnStart Impl
functions work without any routing changes

Adding &quot;frame&quot; to Worker.json targetTypes is sufficient for the frontend —
WorkerManager.initializeTarget already gates on hasDomain(&quot;Worker&quot;), which
now returns true for FrameTargets.

TargetManager.targetDestroyed cleans up child WorkerTargets when a
FrameTarget is destroyed, since Worker.workerTerminated messages from the
backend get dropped due to a race in willDestroyFrame.

Tests: http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html
       http/tests/site-isolation/inspector/worker/worker-resources/iframe-with-worker.html

* LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe-expected.txt: Added.
* LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html: Added.
* LayoutTests/inspector/worker/resources/cross-origin-worker.js: Added.
* LayoutTests/inspector/worker/resources/iframe-with-worker.html: Added.
* Source/JavaScriptCore/inspector/protocol/Worker.json:
* Source/WebCore/Sources.txt:
* Source/WebCore/inspector/FrameInspectorController.cpp:
(WebCore::FrameInspectorController::createLazyAgents):
(WebCore::InspectorInstrumentation::instrumentingAgents):
* Source/WebCore/inspector/agents/frame/FrameWorkerAgent.cpp: Copied from Source/WebCore/inspector/agents/page/PageWorkerAgent.cpp.
(WebCore::FrameWorkerAgent::FrameWorkerAgent):
(WebCore::FrameWorkerAgent::didCreateFrontendAndBackend):
(WebCore::FrameWorkerAgent::willDestroyFrontendAndBackend):
(WebCore::FrameWorkerAgent::connectToAllWorkerInspectorProxies):
* Source/WebCore/inspector/agents/frame/FrameWorkerAgent.h: Copied from Source/WebCore/inspector/agents/page/PageWorkerAgent.cpp.
* Source/WebCore/inspector/agents/page/PageWorkerAgent.cpp:
(WebCore::PageWorkerAgent::connectToAllWorkerInspectorProxies):
* Source/WebInspectorUI/UserInterface/Controllers/TargetManager.js:
(WI.TargetManager.prototype.targetDestroyed):

Canonical link: <a href="https://commits.webkit.org/311312@main">https://commits.webkit.org/311312@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/dc22ee63574feb732a3a7bd2cf802ef39a6d03b6

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/156612 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/29947 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/23131 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/165435 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/110693 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/158483 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/30084 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/29951 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/121302 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/110693 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/159570 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/23539 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/140643 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/101970 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/22595 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/20776 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/13207 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/148662 "Built successfully and passed tests") | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/132274 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/18472 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/167918 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/17447 "Built successfully and passed tests") | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/12038 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/20089 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/129417 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/29550 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/24854 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/129527 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35083 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/29473 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/140266 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/87274 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/24353 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/17069 "Passed tests") | [  ~~🛠 jsc-armv7~~](https://ews-build.webkit.org/#/builders/35/builds/188573 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/29181 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/93145 "Built successfully") | [  ~~🧪 jsc-armv7-tests~~](https://ews-build.webkit.org/#/builders/35/builds/188573 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/28706 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/28936 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/28831 "Built successfully") | | | |

### DIFF
--- a/LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html
+++ b/LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html
@@ -1,0 +1,107 @@
+<!-- webkit-test-runner [ SiteIsolationEnabled=true ] -->
+<script src="../../../inspector/resources/inspector-test.js"></script>
+<script>
+    let iframe;
+
+    function addCrossOriginIFrame() {
+        let crossOriginHost = location.hostname === "localhost" ? "127.0.0.1" : "localhost";
+        iframe = document.createElement("iframe");
+        iframe.src = `http://${crossOriginHost}:8000/site-isolation/inspector/worker/resources/iframe-with-worker.html`;
+        document.body.appendChild(iframe);
+    }
+
+    function removeCrossOriginIFrame() {
+        iframe.remove();
+    }
+
+    function test() {
+        let suite = InspectorTest.createAsyncSuite("SiteIsolation.Worker.CrossOriginIFrame");
+
+        let parentFrameTarget;
+        let workerTarget;
+
+        suite.addTestCase({
+            name: "SiteIsolation.Worker.CrossOrigin.Setup",
+            description: "Adding a cross-origin iframe with a worker should create both frame and worker targets.",
+            async test() {
+                // When a cross-origin iframe loads under site isolation, the target
+                // lifecycle is: a non-provisional FrameTarget is created, then a
+                // provisional FrameTarget is created, then the non-provisional one is
+                // destroyed and the provisional one commits. The worker connects to
+                // whichever FrameTarget is current at the time Worker.enable() runs.
+                //
+                // Rather than trying to match a specific FrameTarget from the events
+                // array (which can see multiple Frame targets due to the provisional
+                // swap), use the worker's parentTarget as the source of truth.
+                let workerAddedPromise = new Promise((resolve) => {
+                    WI.targetManager.addEventListener(WI.TargetManager.Event.TargetAdded, (event) => {
+                        if (event.data.target.type === WI.TargetType.Worker)
+                            resolve(event.data.target);
+                    });
+                });
+
+                InspectorTest.evaluateInPage("addCrossOriginIFrame()");
+
+                workerTarget = await workerAddedPromise;
+                parentFrameTarget = workerTarget.parentTarget;
+
+                InspectorTest.expectNotNull(parentFrameTarget, "Frame target should be created.");
+                InspectorTest.expectEqual(parentFrameTarget.type, WI.TargetType.Frame, "Parent target should have Frame type.");
+                InspectorTest.expectThat(parentFrameTarget.hasDomain("Worker"), "Parent frame target should have the Worker domain.");
+
+                InspectorTest.expectNotNull(workerTarget, "Worker target should be created.");
+                InspectorTest.expectEqual(workerTarget.type, WI.TargetType.Worker, "Worker target should have Worker type.");
+                InspectorTest.expectEqual(workerTarget.parentTarget, parentFrameTarget, "Worker's parent target should be the frame target.");
+                InspectorTest.log(`Worker target name: ${workerTarget.displayName}`);
+            }
+        });
+
+        suite.addTestCase({
+            name: "SiteIsolation.Worker.CrossOrigin.EvaluateInWorker",
+            description: "Should be able to evaluate JavaScript in the worker spawned by the cross-origin iframe.",
+            async test() {
+                InspectorTest.assert(workerTarget, "Worker target should exist from setup.");
+
+                let response = await workerTarget.RuntimeAgent.evaluate.invoke({
+                    expression: "passphrase",
+                    objectGroup: "test",
+                    returnByValue: true,
+                });
+                InspectorTest.expectThat(!response.wasThrown, "Evaluation should not throw.");
+                InspectorTest.log("Resulting passphrase value: " + response.result.value);
+            }
+        });
+
+        suite.addTestCase({
+            name: "SiteIsolation.Worker.CrossOrigin.RemoveIframe",
+            description: "Removing the cross-origin iframe should clean up both worker and frame targets.",
+            async test() {
+                let removedTargets = [];
+                let allRemovedPromise = new Promise((resolve) => {
+                    WI.targetManager.addEventListener(WI.TargetManager.Event.TargetRemoved, (event) => {
+                        removedTargets.push(event.data.target);
+                        if (removedTargets.length >= 2)
+                            resolve();
+                    });
+                });
+
+                InspectorTest.evaluateInPage("removeCrossOriginIFrame()");
+
+                await allRemovedPromise;
+
+                let removedWorker = removedTargets.find((t) => t.type === WI.TargetType.Worker);
+                let removedFrame = removedTargets.find((t) => t.type === WI.TargetType.Frame);
+
+                InspectorTest.expectNotNull(removedWorker, "Worker target should be removed.");
+                InspectorTest.expectNotNull(removedFrame, "Frame target should be removed.");
+            }
+        });
+
+        suite.runTestCasesAndFinish();
+    }
+</script>
+
+<body onload="runTest()">
+    <p>Test that workers spawned in cross-origin iframes are properly surfaced through the frame target's Worker domain
+        under site isolation.</p>
+</body>
--- a/LayoutTests/platform/mac-site-isolation/TestExpectations
+++ b/LayoutTests/platform/mac-site-isolation/TestExpectations
@@ -255,6 +255,7 @@ http/tests/site-isolation/inspector/debugger/pause-in-cross-origin-iframe.html [
 http/tests/site-isolation/inspector/debugger/scriptParsed-frame-target.html [ Pass ]
 http/tests/site-isolation/inspector/debugger/setBreakpoint-cross-origin-iframe.html [ Pass ]
 http/tests/site-isolation/inspector/page/override-setting-cross-origin-iframe.html [ Pass Failure ]
+http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html [ Pass ]
 http/tests/ssl/applepay/page-cache-active-apple-pay-session.html [ Failure ]
 http/tests/ssl/applepay/page-cache-inactive-apple-pay-session.html [ Failure ]
 http/tests/ssl/media-stream/get-user-media-secure-connection.html [ Failure ]
--- a/Source/WebCore/inspector/InspectorInstrumentation.cpp
+++ b/Source/WebCore/inspector/InspectorInstrumentation.cpp
@@ -36,6 +36,7 @@
 #include "ComputedEffectTiming.h"
 #include "ContextDestructionObserverInlines.h"
 #include "DOMWrapperWorld.h"
+#include "Document.h"
 #include "DocumentLoader.h"
 #include "Event.h"
 #include "EventTargetInlines.h"
@@ -1383,8 +1384,11 @@ InstrumentingAgents& InspectorInstrumentation::instrumentingAgents(Page& page)
 InstrumentingAgents* InspectorInstrumentation::instrumentingAgents(ScriptExecutionContext& context)
 {
     // Using RefPtr makes us hit the m_inRemovedLastRefFunction assert.
-    if (WeakPtr document = dynamicDowncast<Document>(context))
+    if (WeakPtr document = dynamicDowncast<Document>(context)) {
+        if (auto* frame = document->frame())
+            return &instrumentingAgents(*frame);
         return instrumentingAgents(document->page());
+    }
     if (auto* workerOrWorkletGlobal = dynamicDowncast<WorkerOrWorkletGlobalScope>(context))
         return &instrumentingAgents(*workerOrWorkletGlobal);
     return nullptr;
--- a/Source/WebCore/inspector/agents/frame/FrameWorkerAgent.cpp
+++ b/Source/WebCore/inspector/agents/frame/FrameWorkerAgent.cpp
@@ -1,0 +1,86 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+ * THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "FrameWorkerAgent.h"
+
+#include "Document.h"
+#include "InstrumentingAgents.h"
+#include "LocalFrame.h"
+#include "Page.h"
+#include "WorkerInspectorProxy.h"
+#include <wtf/TZoneMallocInlines.h>
+
+namespace WebCore {
+
+using namespace Inspector;
+
+WTF_MAKE_TZONE_ALLOCATED_IMPL(FrameWorkerAgent);
+
+FrameWorkerAgent::FrameWorkerAgent(FrameAgentContext& context)
+    : InspectorWorkerAgent(context)
+    , m_inspectedFrame(context.inspectedFrame)
+{
+}
+
+FrameWorkerAgent::~FrameWorkerAgent() = default;
+
+void FrameWorkerAgent::didCreateFrontendAndBackend()
+{
+    Ref { m_instrumentingAgents.get() }->setPersistentWorkerAgent(this);
+}
+
+void FrameWorkerAgent::willDestroyFrontendAndBackend(DisconnectReason reason)
+{
+    Ref { m_instrumentingAgents.get() }->setPersistentWorkerAgent(nullptr);
+
+    InspectorWorkerAgent::willDestroyFrontendAndBackend(reason);
+}
+
+void FrameWorkerAgent::connectToAllWorkerInspectorProxies()
+{
+    RefPtr frame = m_inspectedFrame.get();
+    if (!frame)
+        return;
+
+    RefPtr page = frame->page();
+    if (!page || !page->identifier())
+        return;
+
+    // FIXME: https://bugs.webkit.org/show_bug.cgi?id=312381 SI WorkerInspectorProxy should support proxiesForFrame()
+    for (Ref proxy : WorkerInspectorProxy::proxiesForPage(*page->identifier())) {
+        RefPtr context = proxy->scriptExecutionContext();
+        if (!context)
+            continue;
+
+        RefPtr document = dynamicDowncast<Document>(context.get());
+        if (!document || document->frame() != frame.get())
+            continue;
+
+        connectToWorkerInspectorProxy(proxy);
+    }
+}
+
+} // namespace WebCore
--- a/Source/WebCore/inspector/agents/page/PageWorkerAgent.cpp
+++ b/Source/WebCore/inspector/agents/page/PageWorkerAgent.cpp
@@ -26,8 +26,8 @@
 #include "config.h"
 #include "PageWorkerAgent.h"
 
-#include "Document.h"
 #include "Page.h"
+#include "Settings.h"
 #include "WorkerInspectorProxy.h"
 
 namespace WebCore {
@@ -46,6 +46,12 @@ PageWorkerAgent::~PageWorkerAgent() = default;
 
 void PageWorkerAgent::connectToAllWorkerInspectorProxies()
 {
+    // Under site isolation, every frame (including the main frame) has its own
+    // FrameInspectorController with a FrameWorkerAgent that handles workers for
+    // that frame. PageWorkerAgent does not need to connect to any workers.
+    if (m_page->settings().siteIsolationEnabled())
+        return;
+
     for (Ref proxy : WorkerInspectorProxy::proxiesForPage(*m_page->identifier()))
         connectToWorkerInspectorProxy(proxy);
 }
--- a/Source/WebInspectorUI/UserInterface/Controllers/TargetManager.js
+++ b/Source/WebInspectorUI/UserInterface/Controllers/TargetManager.js
@@ -159,6 +159,14 @@ WI.TargetManager = class TargetManager extends WI.Object
         if (!target)
             return;
 
+        // FIXME: https://bugs.webkit.org/show_bug.cgi?id=312386 SI Move cleanup of children workers to backend
+        if (target.type === WI.TargetType.Frame) {
+            for (let workerTarget of this.workerTargets) {
+                if (workerTarget.parentTarget === target)
+                    WI.workerManager.workerTerminated(workerTarget.identifier);
+            }
+        }
+
         this._checkAndHandlePageTargetTermination(target);
         this.removeTarget(target);
     }
--- a/LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe-expected.txt
+++ b/LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe-expected.txt
@@ -1,0 +1,21 @@
+Test that workers spawned in cross-origin iframes are properly surfaced through the frame target's Worker domain under site isolation.
+
+
+== Running test suite: SiteIsolation.Worker.CrossOriginIFrame
+-- Running test case: SiteIsolation.Worker.CrossOrigin.Setup
+PASS: Frame target should be created.
+PASS: Parent target should have Frame type.
+PASS: Parent frame target should have the Worker domain.
+PASS: Worker target should be created.
+PASS: Worker target should have Worker type.
+PASS: Worker's parent target should be the frame target.
+Worker target name: CrossOriginWorker
+
+-- Running test case: SiteIsolation.Worker.CrossOrigin.EvaluateInWorker
+PASS: Evaluation should not throw.
+Resulting passphrase value: worker-in-cross-origin-iframe
+
+-- Running test case: SiteIsolation.Worker.CrossOrigin.RemoveIframe
+PASS: Worker target should be removed.
+PASS: Frame target should be removed.
+
--- a/LayoutTests/inspector/worker/resources/cross-origin-worker.js
+++ b/LayoutTests/inspector/worker/resources/cross-origin-worker.js
@@ -1,0 +1,5 @@
+var passphrase = "worker-in-cross-origin-iframe";
+
+self.addEventListener("message", (event) => {
+    self.postMessage(passphrase);
+});
--- a/LayoutTests/inspector/worker/resources/iframe-with-worker.html
+++ b/LayoutTests/inspector/worker/resources/iframe-with-worker.html
@@ -1,0 +1,4 @@
+<script>
+var worker = new Worker("cross-origin-worker.js", {name: "CrossOriginWorker"});
+worker.postMessage("ping");
+</script>
--- a/Source/JavaScriptCore/inspector/protocol/Worker.json
+++ b/Source/JavaScriptCore/inspector/protocol/Worker.json
@@ -1,7 +1,7 @@
 {
     "domain": "Worker",
     "debuggableTypes": ["web-page"],
-    "targetTypes": ["page", "worker"],
+    "targetTypes": ["frame", "page", "worker"],
     "commands": [
         {
             "name": "enable",
--- a/Source/WebCore/Sources.txt
+++ b/Source/WebCore/Sources.txt
@@ -1918,6 +1918,7 @@ inspector/agents/WebDebuggerAgent.cpp
 inspector/agents/WebHeapAgent.cpp
 inspector/agents/frame/FrameConsoleAgent.cpp
 inspector/agents/frame/FrameDebuggerAgent.cpp
+inspector/agents/frame/FrameWorkerAgent.cpp
 inspector/agents/page/PageAuditAgent.cpp
 inspector/agents/page/PageCanvasAgent.cpp
 inspector/agents/page/PageConsoleAgent.cpp
--- a/Source/WebCore/inspector/FrameInspectorController.cpp
+++ b/Source/WebCore/inspector/FrameInspectorController.cpp
@@ -37,6 +37,7 @@
 #include "FrameDebuggerAgent.h"
 #include "FrameInlines.h"
 #include "FrameRuntimeAgent.h"
+#include "FrameWorkerAgent.h"
 #include "InspectorInstrumentation.h"
 #include "InspectorWebAgentBase.h"
 #include "InstrumentingAgents.h"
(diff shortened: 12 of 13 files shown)


Source/WebCore/inspector/InspectorBackendClient.h
```diff
@@ -75,6 +75,10 @@ class InspectorBackendClient {
     using DeveloperPreference = InspectorBackendClientDeveloperPreference;
     virtual void setDeveloperPreferenceOverride(DeveloperPreference, std::optional<bool>) { }
 
+    // Worker inspector target lifecycle notifications
+    virtual void workerInspectorTargetCreated(class WorkerInspectorProxy&) { }
+    virtual void workerInspectorTargetDestroyed(class WorkerInspectorProxy&) { }
```
burg: Nit: not sure why `class` used here in parameters


the method naming is a bit misaligned here, the InspectorInstrumentation is notified about worker lifecycle but it doesn't know about targets. I would simplify to having the external callback be named `workerStarted` to match previously.

Source/WebCore/inspector/InspectorInstrumentation.cpp
```diff
@@ -1046,16 +1049,32 @@ bool InspectorInstrumentation::shouldWaitForDebuggerOnStartImpl(InstrumentingAge
...
+    if (!page)
+        return nullptr;
+    return page->inspectorController().inspectorBackendClient();
+}
+
 void InspectorInstrumentation::workerStartedImpl(InstrumentingAgents& instrumentingAgents, WorkerInspectorProxy& proxy)
 {
-    if (auto* workerAgent = instrumentingAgents.persistentWorkerAgent())
-        workerAgent->workerStarted(proxy);
+    if (auto* client = inspectorBackendClientForWorker(proxy))
```
burg: Prefer early return so we can drop the else { } and its indent.

Source/WebCore/inspector/InspectorInstrumentation.cpp
```diff
@@ -1046,16 +1049,32 @@ bool InspectorInstrumentation::shouldWaitForDebuggerOnStartImpl(InstrumentingAge
...
+    }
 }
 
 void InspectorInstrumentation::workerTerminatedImpl(InstrumentingAgents& instrumentingAgents, WorkerInspectorProxy& proxy)
 {
-    if (auto* workerAgent = instrumentingAgents.persistentWorkerAgent())
-        workerAgent->workerTerminated(proxy);
+    if (auto* client = inspectorBackendClientForWorker(proxy))
+        client->workerInspectorTargetDestroyed(proxy);
+    else {
```
burg: Ditto.

Source/WebCore/inspector/InspectorInstrumentation.cpp
burg: Hmm, this gets a little tricky with compatibility in mind. @the-chenergy 

If we intend to support older (shipping) frontends-and I think we should-, such clients will expect "Worker" domain to exist on Page targets, and to send commands/events through it.

It might be easier to have these instrumentations co-existing for the time being. If we need a legacy WorkerAgent due to connecting to an older frontend, it can be created lazily.

Source/WebCore/inspector/agents/page/PageWorkerAgent.cpp
```diff
@@ -44,6 +45,14 @@ PageWorkerAgent::PageWorkerAgent(PageAgentContext& context)
 
 PageWorkerAgent::~PageWorkerAgent() = default;
 
+Inspector::Protocol::ErrorStringOr<void> PageWorkerAgent::enable()
+{
+    if (m_page->inspectorController().inspectorBackendClient())
```
burg: Due to the earlier comment above, we may need to drop this exclusivity check.

Source/WebCore/inspector/agents/page/PageWorkerAgent.h
```diff
@@ -41,6 +41,9 @@ class PageWorkerAgent final : public InspectorWorkerAgent {
 
     ~PageWorkerAgent();
 
+    // InspectorWorkerAgent overrides
+    Inspector::Protocol::ErrorStringOr<void> enable() override;
```
burg: Nit: `Inspector::CommandResult<void>`

Source/WebCore/workers/WorkerInspectorProxy.cpp
```diff
@@ -146,11 +146,28 @@ auto WorkerInspectorProxy::pageOrWorkerGlobalScopeIdentifier(ScriptExecutionCont
     return context.identifier();
 }
 
+static Page* findTopLevelPage(ScriptExecutionContext& context)
+{
+    if (RefPtr document = dynamicDowncast<Document>(context))
+        return document->page();
+
+    if (RefPtr workerScope = dynamicDowncast<WorkerGlobalScope>(context)) {
+        auto proxies = WorkerInspectorProxy::proxiesForWorkerGlobalScope(workerScope->identifier());
```
burg: I don't remember offhand why there are multiple proxies per workerScope. Is this to handle cases like shared workers? Might be worth a comment.

Source/WebInspectorUI/UserInterface/Controllers/TargetManager.js
```diff
@@ -187,10 +187,10 @@ WI.TargetManager = class TargetManager extends WI.Object
         case InspectorBackend.Enum.Target.TargetInfoType.Page:
             return new WI.PageTarget(parentTarget, targetId, WI.UIString("Page"), connection, {isProvisional, isPaused});
         case InspectorBackend.Enum.Target.TargetInfoType.Worker:
-            return new WI.WorkerTarget(parentTarget, targetId, WI.UIString("Worker"), connection, {isPaused});
+            return new WI.WorkerTarget(parentTarget, targetId, /* url */ "", WI.UIString("Worker"), connection, {isPaused});
```
burg: What is the URL supposed to be? main resource URL?

Source/WebKit/UIProcess/Inspector/WebPageInspectorController.h
```diff
@@ -73,7 +73,9 @@ class WebPageInspectorController {
 
     void createWebPageInspectorTarget(const String& targetId, Inspector::InspectorTargetType);
     void createWebFrameInspectorTarget(WebFrameProxy&, const String& targetId);
+    void createWorkerInspectorTarget(IPC::Connection&, const String& targetId, const String& url, const String& name);
```
burg: Nit: URL?

Source/WebKit/UIProcess/Inspector/WebWorkerInspectorTargetProxy.cpp
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright (C) 2025 Apple Inc. All rights reserved.
```
burg: Nit: 2026

Source/WebKit/WebProcess/WebPage/WebPage.cpp
```diff
@@ -4091,6 +4093,44 @@ void WebPage::sendMessageToTargetBackend(const String& message)
     ensureInspectorTarget()->sendMessageToTargetBackend(message);
 }
 
+void WebPage::connectWorkerInspector(const String& targetId, Inspector::FrontendChannel::ConnectionType connectionType)
```
burg: Nit: should probably mention `inspectorBackend` or `inspectorFrontend`

Source/WebKit/WebProcess/WebPage/WebPage.cpp
```diff
@@ -4091,6 +4093,44 @@ void WebPage::sendMessageToTargetBackend(const String& message)
...
+{
+    auto it = m_workerInspectorTargets.find(targetId);
+    if (it != m_workerInspectorTargets.end())
```
burg: Nit: prefer early-return if there's nothing to do, rather than the inverse.

Source/WebKit/WebProcess/WebPage/WebPage.cpp
```diff
@@ -4091,6 +4093,44 @@ void WebPage::sendMessageToTargetBackend(const String& message)
...
+}
+
+void WebPage::disconnectWorkerInspector(const String& targetId)
+{
+    auto it = m_workerInspectorTargets.find(targetId);
+    if (it != m_workerInspectorTargets.end())
+        it->value->disconnect();
+}
+
+void WebPage::sendMessageToWorkerInspectorTarget(const String& targetId, const String& message)
```
burg: Uhh, shouldn't this stuff be (already) in InspectorTargetAgent? I'd expect WebPageInspectorController to be in charge adding/removing from TargetAgent.

Source/WebKit/WebProcess/WebPage/WebPage.cpp
```diff
@@ -4091,6 +4093,44 @@ void WebPage::sendMessageToTargetBackend(const String& message)
...
+{
+    auto it = m_workerInspectorTargets.find(targetId);
+    if (it != m_workerInspectorTargets.end())
+        it->value->sendMessageToTargetBackend(message);
+}
+
+void WebPage::workerInspectorTargetCreated(WebCore::WorkerInspectorProxy& proxy)
+{
+    String targetId = makeString("page-"_s, identifier().toUInt64(), '-', proxy.identifier());
```
burg: This is wrong, it should be "worker-XX" right?

burg: Let's use a helper method somewhere rather than duplicating the string construction.

burg (commented):
Great progress.

What's the lifecycle like here if a new cross-origin iframe in a new web process is created after the page is already being inspected?

Source/WebCore/inspector/InspectorInstrumentation.cpp
the-chenergy: Right, I think for the Worker domain specifically, it might be easier if our compatibility solution is to still have a legacy worker agent in the page. (In contrast to many other domains like Console and Runtime, where it's fine for the page to not have a console agent and instead route messages to its main frame instead, and the routing part can be done by function calls and no IPC. Having to route Worker messages might be more difficult as it involves IPC back to the UI process where the handling of workers will live)

burg: Just for posterity, we decided to keep both pre- and post-SI agent implementations around for the time being. The legacy page-based agents will be retained to support WebKitLegacy and to support older frontends which must connect with Site Isolation disabled.

LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html
```diff
@@ -0,0 +1,105 @@
...
+                    if (event.data.target.type === WI.TargetType.Worker)
+                        resolve();
+                });
+            });
+
+            InspectorTest.evaluateInPage("addCrossOriginIFrame()");
+
+            // Wait for the worker target to be added (with timeout).
+            let timeoutPromise = new Promise((_, reject) => setTimeout(() => reject(new Error("Timed out waiting for worker target")), 10000));
+            await Promise.race([workerAddedPromise, timeoutPromise]);
```
the-chenergy: I think just awaiting the `workerAddedPromise` is sufficient? The test case itself has a built-in timeout of ten seconds

LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html
```diff
@@ -0,0 +1,105 @@
...
+
+            crossOriginTarget = addedTargets.find((t) => t instanceof WI.FrameTarget);
+            workerTarget = addedTargets.find((t) => t.type === WI.TargetType.Worker);
```
the-chenergy:
```suggestion
            crossOriginTarget = addedTargets.find((t) => t.type === WI.TargetType.Frame);
            workerTarget = addedTargets.find((t) => t.type === WI.TargetType.Worker);
```
Or the other way around for `workerTarget`, for a more consistent look

Source/WebCore/inspector/InspectorInstrumentation.cpp
```diff
@@ -79,6 +80,7 @@
 #include "WorkerInspectorController.h"
 #include "WorkerOrWorkletGlobalScope.h"
 #include "agents/frame/FrameRuntimeAgent.h"
+#include "agents/frame/FrameWorkerAgent.h"
```
the-chenergy: I think these can just include the file name without the relative paths

Source/WebCore/inspector/InstrumentingAgents.h
```diff
@@ -92,6 +93,7 @@ class WebHeapAgent;
 #define DEFINE_INSPECTOR_AGENT_Timeline(macro, Getter, Setter) DEFINE_INSPECTOR_AGENT(macro, InspectorTimelineAgent, TimelineAgent, Getter, Setter)
 #define DEFINE_INSPECTOR_AGENT_Timeline_Page(macro, Getter, Setter) DEFINE_INSPECTOR_AGENT(macro, PageTimelineAgent, PageTimelineAgent, Getter, Setter)
 #define DEFINE_INSPECTOR_AGENT_Worker(macro, Getter, Setter) DEFINE_INSPECTOR_AGENT(macro, InspectorWorkerAgent, WorkerAgent, Getter, Setter)
+#define DEFINE_INSPECTOR_AGENT_Worker_Frame(macro, Getter, Setter) DEFINE_INSPECTOR_AGENT(macro, FrameWorkerAgent, FrameWorkerAgent, Getter, Setter)
```
the-chenergy: Is this (the field `m_frameWorkerAgent` on `InstrumentingAgents`) actually necessary? If we could just use the existing worker agent (defined the line above), we can drop the extra argument from the two functions in `InspectorInstrumentation` and get the agent is a simpler way, similar to how I did it with the console agent where `instrumentingAgents.enabledConsoleAgent()` just returns either frame or page console agent as appropriate

Source/WebCore/inspector/InspectorInstrumentation.cpp
```diff
@@ -1047,21 +1049,46 @@ void InspectorInstrumentation::didDispatchDOMStorageEventImpl(InstrumentingAgent
...
 {
+    if (RefPtr document = dynamicDowncast<Document>(proxy.scriptExecutionContext())) {
+        if (RefPtr frame = document->frame()) {
+            if (CheckedPtr frameWorkerAgent = frame->inspectorController().instrumentingAgents().persistentFrameWorkerAgent()) {
+                frameWorkerAgent->workerStarted(proxy);
+                return;
+            }
+        }
+    }
+
```
the-chenergy: This is what [my comment above](https://github.com/WebKit/WebKit/pull/56939/changes#r2915332930) was referring to. Can we get away with not creating a frame-specific worker agent field in `InstrumentingAgents`, so we can basically not touch this function and `shouldWaitForDebuggerOnStartImpl`?

LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html
```diff
@@ -0,0 +1,105 @@
...
+        name: "SiteIsolation.Worker.CrossOrigin.Setup",
+        description: "Adding a cross-origin iframe with a worker should create both frame and worker targets.",
+        async test() {
+            // Collect all TargetAdded events. The frame target fires first, then the
+            // worker target fires after Worker.enable() is sent during initializeTarget.
+            let addedTargets = [];
+            let workerAddedPromise = new Promise((resolve) => {
+                WI.targetManager.addEventListener(WI.TargetManager.Event.TargetAdded, (event) => {
+                    addedTargets.push(event.data.target);
+                    if (event.data.target.type === WI.TargetType.Worker)
```
the-chenergy: The test below for removing the targets looks like we allow the frame and worker targets to be removed in either order, but here we're being more strict. Perhaps let's make them look more consistent one way or the other?
```suggestion
                    if (addedTargets.length >= 2)
```

LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html
```diff
@@ -0,0 +1,105 @@
...
+        async test() {
+            InspectorTest.assert(workerTarget, "Worker target should exist from setup.");
+
+            let response = await workerTarget.RuntimeAgent.evaluate.invoke({
+                expression: "passphrase",
+                objectGroup: "test",
+                returnByValue: true,
+            });
+            InspectorTest.expectThat(!response.wasThrown, "Evaluation should not throw.");
+            InspectorTest.expectEqual(response.result.value, "worker-in-cross-origin-iframe", "Worker should have the expected passphrase value.");
```
the-chenergy:
```suggestion
            InspectorTest.log("Resulting passphrase value: " + response.result.value);
```
So we don't need to hard code the passphrase string twice (here and in the worker JS)

Remember to rerun the test to collect new expectations

LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html
```diff
@@ -0,0 +1,105 @@
...
+                });
+            });
+
+            InspectorTest.evaluateInPage("addCrossOriginIFrame()");
+
+            // Wait for the worker target to be added (with timeout).
+            let timeoutPromise = new Promise((_, reject) => setTimeout(() => reject(new Error("Timed out waiting for worker target")), 10000));
+            await Promise.race([workerAddedPromise, timeoutPromise]);
+
+            crossOriginTarget = addedTargets.find((t) => t instanceof WI.FrameTarget);
```
burg: Nit: `isCrossOriginTarget`, `isWorkerTarget`

LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html
burg: I don't think this is necessary, but if it is, let's extract it to a helper method.

Source/WebCore/inspector/InspectorInstrumentation.cpp
burg: Yes, we shouldn't have added that.

Source/WebCore/inspector/InstrumentingAgents.h
burg: Since they don't need to coexist to preserve functionality, I think it's simpler and preferable to use one slot.

burg (commented):
Have you considered the approach we discussed in the past, where WorkerAgent is only instantiated for legacy/non-SI, and for SI the instrumentations feed worker lifecycle into Target system?

LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html
```diff
@@ -0,0 +1,107 @@
...
+                    WI.targetManager.addEventListener(WI.TargetManager.Event.TargetAdded, (event) => {
+                        if (event.data.target.type === WI.TargetType.Worker)
+                            resolve(event.data.target);
+                    });
+                });
+
+                InspectorTest.evaluateInPage("addCrossOriginIFrame()");
+
+                workerTarget = await workerAddedPromise;
+                crossOriginTarget = workerTarget.parentTarget;
```
the-chenergy: I think we can rename `crossOriginTarget` to `ownerFrameTarget` or `parentFrameTarget` to be more clear. (The worker target is technically cross-origin too!)

LayoutTests/http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html
```diff
@@ -0,0 +1,107 @@
...
+        });
+
+        suite.runTestCasesAndFinish();
+    }
+</script>
+
+<body onload="runTest()">
+    <p>Test that workers spawned in cross-origin iframes are properly surfaced through the frame target's Worker domain
+        under site isolation.</p>
+</body>
```
the-chenergy: Add newline at EOF

Source/WebCore/inspector/agents/page/PageWorkerAgent.cpp
```diff
@@ -46,8 +48,19 @@ PageWorkerAgent::~PageWorkerAgent() = default;
...
+    for (Ref proxy : WorkerInspectorProxy::proxiesForPage(*m_page->identifier())) {
+        // Under site isolation, workers in non-main frames are handled by their
+        // frame's FrameWorkerAgent. Skip them here to avoid wrong parent target.
+        if (m_page->settings().siteIsolationEnabled()) {
+            RefPtr context = proxy->scriptExecutionContext();
+            if (RefPtr document = dynamicDowncast<Document>(context.get())) {
+                if (RefPtr frame = document->frame(); frame && !frame->isMainFrame())
+                    continue;
+            }
+        }
```
the-chenergy: Hmm, I again was hoping that the main frame wouldn't get a special treatment of still being handled by the page target and its Worker agent instead. (I remember bringing this up but I can't find my past comment 😓)

Under SI, the main frame also still get a `FrameInspectorController` (and thus also a `FrameWorkerAgent`, alongside other agents like console) allocated to it. Shouldn't `FrameWorkerAgent` be able to handle the main frame just fine? Relying on `PageWorkerAgent` for the main frame kind of works against the goal of making frame targets independent from the page target as much as possible

I also see that in `InspectorInstrumentation` [(here)](https://github.com/WebKit/WebKit/pull/56939/changes#diff-37f2d688795c0e79bc8fbcbc6756b9358b68c405655a021f53a837f242937a07R1388), the main frame indeed did not get a special treatment (we're not returning the page's agent instead). So is there a subtle reason why this was needed that I'm missing?

the-chenergy: Or maybe I'm misunderstanding this code—this `PageWorkerAgent::connectToAllWorkerInspectorProxies` function is for the case of handling of the page target specifically, so we're actually skipping the __non-main__ frames so that they don't get handled by this `PageWorkerAgent` (and by their `FrameWorkerAgent` instead). And in practice this won't be used under site isolation because we'd be calling into `FrameWorkerAgent:: connectToAllWorkerInspectorProxies` instead. Was that why? In that case, perhaps we can just early-return this function altogether if SI is on; is my understand right?

LayoutTests/platform/mac-site-isolation/TestExpectations
```diff
@@ -252,9 +254,16 @@ http/tests/security/contentSecurityPolicy/embed-redirect-allowed2.html [ Failure
 http/tests/security/contentSecurityPolicy/frame-src-cross-origin-load.html [ Failure ]
 http/tests/security/contentSecurityPolicy/upgrade-insecure-requests/proper-nested-upgrades.html [ Failure ]
 http/tests/security/cross-origin-local-storage.html [ Failure ]
+http/tests/security/file-system-access-via-dataTransfer.html [ Failure ]
+http/tests/security/window-name-after-cross-origin-aux-frame-navigation.html [ Failure ]
+http/tests/security/window-name-after-cross-origin-sub-frame-navigation.html [ Failure ]
+http/tests/security/xss-DENIED-getSVGDocument-iframe.html [ Failure ]
+http/tests/security/xss-DENIED-getSVGDocument-object.html [ Failure ]
+http/tests/security/xss-DENIED-window-open-parent.html [ Failure ]
```
the-chenergy: Did you mean to add these?

Source/WebCore/inspector/agents/frame/FrameWorkerAgent.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+    RefPtr frame = m_inspectedFrame.get();
+    if (!frame)
+        return;
+
+    RefPtr page = frame->page();
+    if (!page || !page->identifier())
+        return;
+
+    for (Ref proxy : WorkerInspectorProxy::proxiesForPage(*page->identifier())) {
```
the-chenergy, line 73: Here it looks like the `WorkerInspectorProxy` is still relying on the page to work, right? Do you think eventually as we move to frame targets, that class should be working with frames instead (so it would become `proxiesForFrame` instead)? That might be worth pointing out as a direction of future work

Source/WebInspectorUI/UserInterface/Controllers/TargetManager.js
```diff
@@ -159,6 +159,16 @@ WI.TargetManager = class TargetManager extends WI.Object
...
 
+        // When a FrameTarget is destroyed, clean up any child WorkerTargets
+        // that were parented to it. The backend does not send separate
+        // targetDestroyed messages for these workers.
+        if (target.type === WI.TargetType.Frame) {
+            for (let workerTarget of this.workerTargets) {
+                if (workerTarget.parentTarget === target)
+                    WI.workerManager.workerTerminated(workerTarget.identifier);
+            }
+        }
```
the-chenergy: Why don't we make the backend in fact send the `targetDestroyed` when the child workers terminate?

LayoutTests/platform/mac-site-isolation/TestExpectations
```diff
@@ -256,6 +256,7 @@ http/tests/site-isolation/inspector/debugger/pause-in-cross-origin-iframe.html [
 http/tests/site-isolation/inspector/debugger/scriptParsed-frame-target.html [ Pass ]
 http/tests/site-isolation/inspector/debugger/setBreakpoint-cross-origin-iframe.html [ Pass ]
 http/tests/site-isolation/inspector/page/override-setting-cross-origin-iframe.html [ Pass Failure ]
+http/tests/site-isolation/inspector/worker/worker-in-cross-origin-iframe.html [ Pass ]
```
the-chenergy, line 258: Is this actually needed?

I think we only explicitly include a test in a `TestExpectations` file as `[ Pass ]` when we've disabled it somewhere else. Our tests skipped in the [mac-wk2 expectations here](https://github.com/WebKit/WebKit/blob/main/LayoutTests/platform/mac-wk2/TestExpectations#L2411-L2418) don't include this one, so maybe we don't need to include this change here

Source/WebCore/inspector/agents/frame/FrameWorkerAgent.cpp
```diff
@@ -0,0 +1,87 @@
...
+    RefPtr frame = m_inspectedFrame.get();
+    if (!frame)
+        return;
+
+    RefPtr page = frame->page();
+    if (!page || !page->identifier())
+        return;
+
+    // FIXME: WorkerInspectorProxy should support proxiesForFrame() so that
+    // frame-level agents do not need to filter by page and then by frame.
```
the-chenergy: Can you file a Bugzilla for this `FIXME` and include a link prior to landing?

Source/WebInspectorUI/UserInterface/Controllers/TargetManager.js
```diff
@@ -159,6 +159,16 @@ WI.TargetManager = class TargetManager extends WI.Object
         if (!target)
             return;
 
+        // FIXME: Worker.workerTerminated messages from the backend get dropped
+        // because the frame target is removed from m_targets before the WebProcess
+        // disconnect completes. Clean up child WorkerTargets on the frontend for now.
```
the-chenergy: Can you file a Bugzilla for this FIXME and include a link prior to landing?

the-chenergy (approved):

Source/WebCore/inspector/agents/frame/FrameWorkerAgent.cpp
AnujPanta1: yes, will file both fixes

the-chenergy:
Remember to file and include a Bugzilla to the commit message

Also let's remove the parentheses around the commit message title

```diff
-([Site Isolation] Web Inspector: support cross-origin iframes in the Worker domain)
+[Site Isolation] Web Inspector: support cross-origin iframes in the Worker domain
+https://webkit.org/b/??????
 rdar://143596746
```

rniwa:
> https://bugs.webkit.org/show_bug.cgi?id=308488
> ([Site Isolation] Web Inspector: support cross-origin iframes in the Worker domain)
> [rdar://143596746](https://rdar.apple.com/143596746)

Bug title should appear before bug URL.

PR #57252, opened 2026-01-26
[ImageLoader] Refactor updateFromElement() with early returns and cleaner structure

hjanuschka:
### Related PRs
- PR 1: https://github.com/WebKit/WebKit/pull/57251 
- PR 2: https://github.com/WebKit/WebKit/pull/57252
- PR 3: https://github.com/WebKit/WebKit/pull/57253

#### f93a23d146509aa942534c9dc4bf1a1dac43d97e
<pre>
[ImageLoader] Refactor updateFromElement() with early returns and cleaner structure
<a href="https://bugs.webkit.org/show_bug.cgi?id=190031">https://bugs.webkit.org/show_bug.cgi?id=190031</a>

Reviewed by Brent Fulgham.

Refactor ImageLoader::updateFromElement() to use early returns and reduce
nesting. This prepares for adding microtask-based loading in a follow-up.

* Source/WebCore/loader/ImageLoader.cpp:
(WebCore::ImageLoader::updateFromElement):

Canonical link: <a href="https://commits.webkit.org/307625@main">https://commits.webkit.org/307625@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/4c9f5c735b5bcda6c716428194b3538cbb52a230

| Misc | iOS, visionOS, tvOS & watchOS  | macOS  | Linux |  Windows |
| ----- | ---------------------- | ------- |  ----- |  --------- |
| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/144922 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/17602 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/9324 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/153593 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/98557 "Built successfully") 
[...]

PR #57251, opened 2026-01-26
[CSP] Add CodePosition struct for preserving source location across async boundaries

hjanuschka:
### Related PRs
- PR 1: https://github.com/WebKit/WebKit/pull/57251 
- PR 2: https://github.com/WebKit/WebKit/pull/57252
- PR 3: https://github.com/WebKit/WebKit/pull/57253

#### 4102647ece0e60863a737081dcee11f37255cbe0
<pre>
[CSP] Add CodePosition struct for preserving source location across async boundaries
<a href="https://bugs.webkit.org/show_bug.cgi?id=190031">https://bugs.webkit.org/show_bug.cgi?id=190031</a>

Reviewed by Brent Fulgham.

This adds a CodePosition struct to ContentSecurityPolicy for capturing
JavaScript source location before async boundaries like microtasks.

* Source/WebCore/page/csp/ContentSecurityPolicy.h:
* Source/WebCore/page/csp/ContentSecurityPolicy.cpp:
(WebCore::ContentSecurityPolicy::getCurrentCodePosition):

Canonical link: <a href="https://commits.webkit.org/307627@main">https://commits.webkit.org/307627@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/7fd2b1aa59596d999fddeb1d037c593050c9eef5

| Misc | iOS, visionOS, tvOS & watchOS  | macOS  | Linux |  Windows |
| ----- | ---------------------- | ------- |  ----- |  --------- |
| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/144932 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/17612 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/9334 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/153603 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/98567 "Built successfully") 
[...]

PR #57253, opened 2026-01-26
[ImageLoader] Defer image loading to microtask per HTML spec

hjanuschka:
https://bugs.webkit.org/show_bug.cgi?id=276586

Reviewed by NOBODY (OOPS!).

**PR 3 of 3** for implementing microtask-based image loading per HTML spec.

Per HTML spec, image loading should be deferred to a microtask to allow
batching of multiple attribute changes. When JavaScript sets img.src
followed by img.srcset, only the final srcset should be loaded.

This change queues a microtask when updateFromElement() is called for
HTMLImageElement, allowing subsequent attribute changes to be batched.
The CodePosition is captured before the microtask to preserve CSP
source location for violation reporting.

### Related PRs
- PR 1: https://github.com/WebKit/WebKit/pull/57251 (merged)
- PR 2: https://github.com/WebKit/WebKit/pull/57252 (merged)
- PR 3: https://github.com/WebKit/WebKit/pull/57253 (this PR)

* Source/WebCore/loader/ImageLoader.h:
* Source/WebCore/loader/ImageLoader.cpp:
(WebCore::ImageLoader::updateFromElement):
(WebCore::ImageLoader::doUpdateFromElement):
https://github.com/WebKit/WebKit/commit/8d36e77d679dfc4c41d3ebb70436d11a7f4c5323

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/158085 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/31422 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/24615 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/166913 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/112168 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/159956 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/31559 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/31424 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/122421 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85941 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/161043 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/24708 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/141983 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/103090 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/23764 "Passed tests") | [❌ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/22084 "1 api test failed or timed out") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/14686 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/133448 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/19775 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/169403 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/14757 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/21398 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/130607 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/31168 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/26149 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/130722 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/31106 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/141569 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/89002 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/25444 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/18375 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/30658 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/96191 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/30179 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/30409 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/30306 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/loader/ImageLoader.cpp
+++ b/Source/WebCore/loader/ImageLoader.cpp
@@ -29,6 +29,7 @@
 #include "Chrome.h"
 #include "ChromeClient.h"
 #include "ContainerNodeInlines.h"
+#include "ContentSecurityPolicy.h"
 #include "CookieJar.h"
 #include "CrossOriginAccessControl.h"
 #include "DocumentLoader.h"
@@ -266,11 +267,20 @@ void ImageLoader::updateFromElement(RelevantMutation relevantMutation)
             // It is possible that attributes are bulk-set via Element::parserSetAttributes. In that case, it is possible that attribute vectors are already configured,
             // but corresponding attributeChanged is not called yet. This causes inconsistency in HTMLImageElement. Eventually, we will get the consistent state, but
             // if "src" attributeChanged is not called yet, imageURL can be invalid and it does not work well for ResourceRequest.
-            // In this case, we should behave same as attr.isNull().
             imageURL = imageElement->currentURL();
             if (imageURL.isNull()) {
-                didUpdateCachedImage(relevantMutation, WTF::move(newImage));
-                return;
+                // For srcset/picture and lazy-loading, keep behavior equivalent to missing/invalid src while we wait for consistent image source state.
+                if (imageElement->usesSrcsetOrPicture() || imageElement->isLazyLoadable()) {
+                    didUpdateCachedImage(relevantMutation, WTF::move(newImage));
+                    return;
+                }
+
+                // For plain src loads, fall back to attribute URL so we still issue the request.
+                imageURL = document->completeURL(attr);
+                if (imageURL.isNull()) {
+                    didUpdateCachedImage(relevantMutation, WTF::move(newImage));
+                    return;
+                }
             }
         } else
             imageURL = document->completeURL(attr);
@@ -281,6 +291,79 @@ void ImageLoader::updateFromElement(RelevantMutation relevantMutation)
 
     auto request = createPotentialAccessControlRequest(WTF::move(resourceRequest), WTF::move(options), document, crossOriginAttribute);
     request.setInitiator(element);
+    m_imageComplete = false;
+
+    // Bypass microtask deferral for non-HTML image elements, shadow tree resources,
+    // image documents, blob/data URLs, and images already available in the memory cache,
+    // as deferral is only needed to batch attribute changes on HTML image elements.
+    if (!imageElement || loadingForElementInShadowTree || imageElement->document().isImageDocument() || request.resourceRequest().url().protocolIsBlob() || request.resourceRequest().url().protocolIsData() || (!imageElement->usesSrcsetOrPicture() && canReuseFromListOfAvailableImages(request, document))) {
+        updateFromElementWithLatestAttributeValues(relevantMutation, WTF::move(request));
+        return;
+    }
+
+    if (m_microtaskQueued)
+        return;
+
+    // Capture source location now so CSP violation reports inside the microtask
+    // can attribute the violation to the script that set the image attributes.
+    auto codePosition = ContentSecurityPolicy::currentCodePosition();
+
+    m_microtaskQueued = true;
+    // https://html.spec.whatwg.org/C#update-the-image-data step 8
+    Ref protectedThis { *this };
+    document->eventLoop().queueMicrotask(document->vm(), [imageLoader = WTF::move(protectedThis), relevantMutation, movedRequest = WTF::move(request), codePosition = WTF::move(codePosition)] mutable {
+        imageLoader->m_microtaskQueued = false;
+        Ref document = imageLoader->document();
+        if (codePosition)
+            document->setCodePosition(*codePosition);
+        imageLoader->updateFromElementWithLatestAttributeValues(relevantMutation, WTF::move(movedRequest));
+    });
+}
+
+void ImageLoader::updateFromElementWithLatestAttributeValues(RelevantMutation relevantMutation, CachedResourceRequest&& request)
+{
+    Ref element = this->element();
+    Ref document = element->document();
+    auto clearCodePosition = makeScopeExit([&] {
+        document->setCodePosition(std::nullopt);
+    });
+
+    CachedResourceHandle<CachedImage> newImage;
+
+    RefPtr imageElement = dynamicDowncast<HTMLImageElement>(element);
+    if (imageElement) {
+        URL imageURL = imageElement->currentURL();
+        if (imageURL.isNull()) {
+            if (imageElement->usesSrcsetOrPicture() || imageElement->isLazyLoadable()) {
+                didUpdateCachedImage(relevantMutation, WTF::move(newImage));
+                return;
+            }
+            imageURL = request.resourceRequest().url();
+            if (imageURL.isNull()) {
+                didUpdateCachedImage(relevantMutation, WTF::move(newImage));
+                return;
+            }
+        }
+
+        auto loadingForElementInShadowTree = element->isInUserAgentShadowTree() || m_elementIsUserAgentShadowRootResource;
+        ResourceLoaderOptions options = CachedResourceLoader::defaultCachedResourceOptions();
+        options.contentSecurityPolicyImposition = loadingForElementInShadowTree ? ContentSecurityPolicyImposition::SkipPolicyCheck : ContentSecurityPolicyImposition::DoPolicyCheck;
+        options.shouldEnableContentExtensionsCheck = loadingForElementInShadowTree ? ShouldEnableContentExtensionsCheck::No : ShouldEnableContentExtensionsCheck::Yes;
+        options.loadedFromPluginElement = LoadedFromPluginElement::No;
+        options.sameOriginDataURLFlag = SameOriginDataURLFlag::Set;
+        options.serviceWorkersMode = ServiceWorkersMode::All;
+        options.referrerPolicy = imageElement->referrerPolicy();
+        options.fetchPriority = imageElement->fetchPriority();
+        if (imageElement->usesSrcsetOrPicture())
+            options.initiator = Initiator::Imageset;
+
+        auto crossOriginAttribute = element->attributeWithoutSynchronization(HTMLNames::crossoriginAttr);
+        ResourceRequest resourceRequest(WTF::move(imageURL));
+        resourceRequest.setInspectorInitiatorNodeIdentifier(InspectorInstrumentation::identifierForNode(element));
+
+        request = createPotentialAccessControlRequest(WTF::move(resourceRequest), WTF::move(options), document, crossOriginAttribute);
+        request.setInitiator(element);
+    }
 
     if (m_loadManually) {
         Ref cachedResourceLoader = document->cachedResourceLoader();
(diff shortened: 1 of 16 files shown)


brentfulgham:
Could you please update this PR after merging PR 57251 and PR 57252? That will probably also help the Early Warning System (ews) failures you are getting with this patch.

LayoutTests/platform/mac-wk2/TestExpectations
```diff
@@ -1390,8 +1388,6 @@ webkit.org/b/246357 imported/w3c/web-platform-tests/feature-policy/reporting/pay
 imported/w3c/web-platform-tests/clear-site-data/navigation-insecure.html [ Skip ]
 imported/w3c/web-platform-tests/clear-site-data/resource.html [ Skip ]
```
Ahmad-S792, line 1334: This seems unrelated to the change.

Source/WebCore/page/csp/ContentSecurityPolicy.h
```diff
@@ -57,22 +58,6 @@ class OrdinalNumber;
 
 namespace WebCore {
 
-// CodePosition captures source file location for CSP violation reporting.
-// This is used to preserve source location across async boundaries (e.g., microtasks).
-struct CodePosition {
```
Ahmad-S792, line 62: Isn't purpose of PR1 was to do this change but now you are undoing it here? So PR1 was not needed?

hjanuschka: In PR3, the CodePosition struct itself is unchanged from PR1. 
getCurrentCodePosition is also unchanged from PR1.

What changed in PR3 is placement and reuse:
- CodePosition was moved out of ContentSecurityPolicy.h into Source/WebCore/page/csp/CodePosition.h
- ContentSecurityPolicy.h now includes that header
- This lets Document and ImageLoader share the same type without pulling in the full CSP header and creating awkward dependencies

so in theo. PR1 is wrong, but it only came to surface when trying to make EWS green on PR3 :/ sorry - splitting is painfull

LayoutTests/platform/mac-wk2/TestExpectations
hjanuschka: required to make EWS green, should i undo it and we keep it as EWS-fault?

Source/WebCore/page/csp/ContentSecurityPolicy.h
hjanuschka: @Ahmad-S792 ping

Source/WebCore/loader/ImageLoader.cpp
```diff
@@ -224,17 +226,14 @@ void ImageLoader::updateFromElement(RelevantMutation relevantMutation)
...
-
-    // Fire an error event if the URL contains only whitespace.
-    if (StringView(attr).containsOnly<isASCIIWhitespace<char16_t>>()) {
-        m_failedLoadURL = attr;
-        m_hasPendingErrorEvent = true;
-        loadEventSender().dispatchEventSoon(*this, eventNames().errorEvent);
+    // Do not load any image if the 'src' attribute is missing or if it is
+    // an empty string.
+    if (attr.isNull() || StringView(attr).containsOnly<isASCIIWhitespace<UChar>>()) {
+        if (!attr.isNull()) {
```
rniwa: I don't think this is an improvement over the existing code. Please revert.

Source/WebCore/loader/ImageLoader.cpp
```diff
@@ -281,6 +289,78 @@ void ImageLoader::updateFromElement(RelevantMutation relevantMutation)
 
     auto request = createPotentialAccessControlRequest(WTF::move(resourceRequest), WTF::move(options), document, crossOriginAttribute);
     request.setInitiator(element);
+    m_imageComplete = false;
+
+    // https://html.spec.whatwg.org/C#update-the-image-data step 7.4
+    // If the image is already in the cache, load it synchronously.
```
rniwa: Please don't add these "what" comments that simply repeats the semantics of the code.
Instead, explain why we're doing this. So why is it that we load synchronously load here
Nothing in step 7.4 of https://html.spec.whatwg.org/multipage/images.html#update-the-image-data
indicates this is what we should do

Source/WebCore/loader/ImageLoader.cpp
```diff
@@ -281,6 +289,78 @@ void ImageLoader::updateFromElement(RelevantMutation relevantMutation)
...
+    if (!imageElement || loadingForElementInShadowTree || imageElement->document().isImageDocument() || request.resourceRequest().url().protocolIsBlob() || (!imageElement->usesSrcsetOrPicture() && canReuseFromListOfAvailableImages(request, document))) {
+        doUpdateFromElement(relevantMutation, WTF::move(request));
+        return;
+    }
+
+    // A microtask is already queued; let it handle the loading.
```
rniwa: This repeats the code. Please remove.

Source/WebCore/loader/ImageLoader.cpp
```diff
@@ -281,6 +289,78 @@ void ImageLoader::updateFromElement(RelevantMutation relevantMutation)
...
+    if (m_microtaskQueued)
+        return;
+
+    // Capture script source location before queueing microtask, so CSP reports and console messages preserve script attribution.
+    if (auto codePosition = ContentSecurityPolicy::getCurrentCodePosition())
+        document->setCodePosition(*codePosition);
```
rniwa: How does this make sense? There could be other micro tasks that end up reporting other CSP issues or the console messages.

Source/WebCore/loader/ImageLoader.cpp
```diff
@@ -281,6 +289,78 @@ void ImageLoader::updateFromElement(RelevantMutation relevantMutation)
...
+{
+    Ref element = this->element();
+    Ref document = element->document();
+    auto clearCodePosition = makeScopeExit([&] {
+        document->setCodePosition(std::nullopt);
+    });
+
+    CachedResourceHandle<CachedImage> newImage;
+
+    // Rebuild the request from latest attributes since they can change between queueing and running the microtask.
```
rniwa: Instead of adding a comment like this, please give the function a descriptive name such as `updateFromElementWithLatestAttributeValues`.

Source/WebCore/loader/ImageLoader.cpp
```diff
@@ -481,7 +568,7 @@ void ImageLoader::notifyFinished(CachedResource& resource, const NetworkLoadMetr
 
         if (hasPendingDecodePromises())
             rejectDecodePromises("Access control error."_s);
-
```
rniwa, line 485: Please don't make these unrelated whitespace changes.

Source/WebCore/page/csp/ContentSecurityPolicy.cpp
```diff
@@ -951,12 +951,47 @@ void ContentSecurityPolicy::reportViolation(const String& effectiveViolatedDirec
 
         info.documentURI = shouldReportProtocolOnly(document->url()) ? document->url().protocol().toString() : document->url().strippedForUseAsReferrer().string;
 
-        auto stack = createScriptCallStack(JSExecState::currentState(), 2);
-        auto* callFrame = stack->firstNonNativeCallFrame();
-        if (callFrame && callFrame->lineNumber()) {
-            info.sourceFile = createURLForReporting(URL { callFrame->preRedirectURL().isEmpty() ? callFrame->sourceURL() : callFrame->preRedirectURL() }, effectiveViolatedDirective, usesReportTo);
-            info.lineNumber = callFrame->lineNumber();
-            info.columnNumber = callFrame->columnNumber();
+        std::optional<CodePosition> javaScriptCodePosition = getCurrentCodePosition();
```
rniwa: While this was added [307627@main](https://commits.webkit.org/307627@main), this function name is incorrect: `getCurrentCodePosition()`.
Only functions with an out argument should be prefixed with `get`.
This function should be called `currentCodePosition` instead:
https://webkit.org/code-style-guidelines/#names-setter-getter

Source/WebCore/page/csp/ContentSecurityPolicy.cpp
```diff
@@ -951,12 +951,47 @@ void ContentSecurityPolicy::reportViolation(const String& effectiveViolatedDirec
...
+        std::optional<CodePosition> parserCodePosition;
+
+        String parserURL;
+        unsigned parserLine = 0;
+        unsigned parserColumn = 0;
+        document->getParserLocation(parserURL, parserLine, parserColumn);
```
rniwa, line 958: Whereas this function name with "get" prefix is appropriate: https://webkit.org/code-style-guidelines/#names-setter-getter

rniwa (changes requested):
I think there are enough stylistic issues here that I don't think we want to merge this as is.

Source/WebCore/page/csp/ContentSecurityPolicy.h
Ahmad-S792: No worries. I am not reviewer so I will wait for someone else to give official review.

Source/WebCore/loader/ImageLoader.cpp
hjanuschka: moved the code position

Source/WebCore/loader/ImageLoader.cpp
rniwa: Why is blob URL special? I don't think this is right.

rniwa, line 333: It seems like this can be declared inside `if`?

rniwa, line 407: This variable is never used.

rniwa, line 404: What is this change about?

rniwa: We should just directly access ¡m_microtaskQueued1 instead of adding a helper function.

rniwa: I also don't understand why an element inside a shadow tree is special. Maybe UA shadow tree but it doesn't make much sense to anything special for author defined shadow roots.

hjanuschka: This is to keep the HEIC fallback request options in sync with the primary image request path. In the fallback path, shouldEnableContentExtensionsCheck was previously left at the default, which could diverge from the main request behavior. I’m explicitly setting it so fallback loading uses the same policy/content-extension behavior as the primary request (including UA-shadow-tree cases).

hjanuschka: this is only intended for UA shadow tree resources, not author-defined shadow roots.  The condition uses element->isInUserAgentShadowTree() || m_elementIsUserAgentShadowRootResource, so author shadow roots are not included. I kept the microtask-bypass logic aligned with the existing UA-shadow-tree policy path.  I can rename the local variable/comment to make “UA shadow tree only” explicit if that helps readability

PR #55393, opened 2025-12-15
Swift WebBackForwardList (off by default)

adetaylor:
<pre>
Swift WebBackForwardList (off by default)
https://bugs.webkit.org/show_bug.cgi?id=305483
rdar://159589815

Reviewed by NOBODY (OOPS!).

This introduces a Swift version of the existing BackForwardList CoreIPC message
handler.

RATIONALE:

WebKit wishes to use Swift to provide much stronger memory safety in its most
sensitive attack surface: CoreIPC message handlers. This PR prototypes how
we can convert a single, simple, message handler to Swift. In itself, converting
this message handler provides negligible security benefit (in fact, rewrites
usually introduce extra bugs) but having a message handler in Swift paves the
way for future work - in particular, it ensures that any future new subsystems
can be written in Swift rather than C++.

BUILDING:

By default this change has no effect. If the LocalOverride option:
  ENABLE_BACK_FORWARD_LIST_SWIFT=ENABLE_BACK_FORWARD_LIST_SWIFT
is given, then a Swift version of the WebBackForwardList will be used. It is
intended to be functionally the same as the C++ version. A recent SDK will
need to be used.

CODE STRUCTURE:

This code is intended to be almost line-for-line the same as the C++ version,
to ease review, which means it is MORE COMPLEX than a typical Swift version.
For instance, an idiomatic Swift version would use an enum to ensure that
the different data members of the BackForwardList cannot become out of sync.

Swift types can't yet inherit from C++ types so when we're in Swift mode,
we still retain a C++ class to inherit from API::Object.

This PR necessarily includes various #ifdefs throughout the codebase to deal
with the Swift vs C++ version. When/if the Swift version becomes the sole
version, they can be removed. We considered allowing Ref<SwiftType> to
minimize the number of places such ifdefs were required, but consensus
was that this just makes things more confusing long-term, and it's better
to stick with ifdefs for now.

FUTURE WORK:

* If this becomes the default or sole implementation, we may choose to
  restructure it to be more idiomatic Swift (moving some runtime
  branches to be compile-time enforced, as noted above)
* The wrappers for WTF logging and IPC
  MESSAGE_CHECKs are temporary - radars note the tasks to build better
  versions.
* Site isolation may change the C++ version of this code, and those changes
  will need to be reflected in the Swift version.
* This code still produces some compiler warnings, which will be resolved
  by adding annotations to further WTF and WebKit C++ types (e.g. WeakRef).
  These will be raised separately.
[...]

PR #57287, opened 2026-01-26
Swift WebBackForwardList (off by default) - WTF extras

adetaylor:
#### 5ee6967eaa817f7f55158feaf55b168342f08aac
<pre>
Swift WebBackForwardList (off by default) - WTF extras
<a href="https://bugs.webkit.org/show_bug.cgi?id=306871">https://bugs.webkit.org/show_bug.cgi?id=306871</a>
<a href="https://rdar.apple.com/169532895">rdar://169532895</a>

Reviewed by Richard Robinson.

We&apos;re introducing a Swift version of the Back Forward List. For the general
design and rationale, see
<a href="https://github.com/WebKit/WebKit/pull/55393">https://github.com/WebKit/WebKit/pull/55393</a>
It will be landed as a series of separate commits in order to ease code review.

In this commit we introduce some WTF/Swift extensions. Specifically,
* Bidirectional constructors for WTF::String &lt;-&gt; Swift String
* An ability to iterate WTF::Vector&lt;WTF::Ref&lt;T&gt;&gt;

In the future these should become a formal part of WTF, a task represented by
<a href="https://rdar.apple.com/164119356">rdar://164119356</a> (there&apos;s similar code in WebGPU which should be merged
across equivalently) - this likely needs to wait for more comprehensive
enablement of Swift projects on a wider selection of targets, instead of just
off-by-default features.

We are adding these types and functions separately in order to ease code review
of subsequent stages. It&apos;s not useful until later commits also land.
It should however be harmless meanwhile.

Canonical link: <a href="https://commits.webkit.org/306853@main">https://commits.webkit.org/306853@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/89281be111428e147e4102cd6ecbf2a5a75e2288

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/142559 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/15022 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/5412 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/151223 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/95748 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/fdef483b-ae51-452f-8ad0-387ab3761e01) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/15686 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/15109 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/109635 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/95748 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/0727f1da-516d-4765-a173-e3680df2b470) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/145508 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/12110 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/127579 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/90544 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b580e4a6-1a62-42b5-8583-044afefdee45) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/11620 "Passed tests") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/9291 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 wpe-libwebrtc](https://ews-build.webkit.org/#/builders/172/builds/1227 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/120980 "Passed tests") | | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/153542 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/14655 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/4691 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/117657 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/14689 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/12754 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/117993 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30089 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/14013 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/124866 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/70346 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/14697 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/3822 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/14434 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/78399 "Built successfully") | | | 
| | [  ~~🛠 watch~~](https://ews-build.webkit.org/#/builders/163/builds/14642 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/14495 "Built successfully") | | | |

### DIFF
--- a/Source/WebKit/UIProcess/StdlibExtras.swift
+++ b/Source/WebKit/UIProcess/StdlibExtras.swift
@@ -1,0 +1,124 @@
+// Copyright (C) 2025-2026 Apple Inc. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+// 1. Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+// 2. Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+// THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+// BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+
+#if compiler(>=6.2)
+
+// FIXME (rdar://164119356): Move parts of StdLibExtras.swift into WTF
+// (those parts which are not specific to WebKit-level types) - and enable
+// irrespective of BACK_FORWARD_LIST_SWIFT
+#if ENABLE_BACK_FORWARD_LIST_SWIFT
+
+internal import WebKit_Internal
+internal import wtf
+
+/// Conform any WTF::Ref<T> to this protocol to get useful extensions.
+protocol CxxRef {
+    associatedtype Pointee // you only need to specify this in your conformance
+    init(_ object: Pointee)
+    func copyRef() -> Self
+}
+
+/// Conform any WTF::Vector<WTF::Ref<T>> to this protocol to get useful extensions and iterators.
+protocol CxxRefVector {
+    associatedtype Element: CxxRef // you only need to specify this in your conformance
+    init()
+    mutating func append(consuming: Element)
+    mutating func reserveCapacity(_ newCapacity: Int)
+    func size() -> Int
+    // swift-format-ignore: AlwaysUseLowerCamelCase, NoLeadingUnderscores
+    func __atUnsafe(_ index: Int) -> UnsafePointer<Element>
+}
+
+extension WTF.String: LosslessStringConvertible {
+    /// Construct a `WTF.String` from a `Swift.String`.
+    public init(_ string: Swift.String) {
+        // rdar://162517354 prevents us from simply writing
+        // self = WTF.String.fromUTF8(swiftString.utf8CString.span);
+        // Safety - we are guaranteed to get a valid buffer from the Swift
+        // string for the duration that we're using it to construct the WTF::String.
+        // The WTF::String will take a copy.
+        let span = string.utf8CString.span
+        self = unsafe span.withUnsafeBufferPointer { ptr in
+            // Warning here is rdar://163018821
+            // swift-format-ignore: NeverForceUnwrap
+            let cppspan = unsafe SpanConstChar(ptr.baseAddress!, span.count)
+            return unsafe WTF.String.fromUTF8(cppspan)
+        }
+    }
+
+    /// Return a `Swift.String` from this `WTF.String`.
+    public var description: Swift.String {
+        // We could possibly make this quicker by treating a C++ span as
+        // a Sequence. For now, we want to avoid unsafe as much as possible.
+        String(utf8(WTF.LenientConversion).toStdString())
+    }
+}
+
+extension WTF.String: ExpressibleByStringLiteral {
+    /// Construct a `WTF.String` from a string literal.
+    public init(stringLiteral: Swift.String) {
+        self.init(stringLiteral)
+    }
+}
+
+extension CxxRefVector {
+    init(array: [Element.Pointee]) {
+        var vec = Self()
+        vec.reserveCapacity(array.count)
+        for item in array {
+            vec.append(consuming: Element(item))
+        }
+        self = vec
+    }
+}
+
+// Iterator for WTF::Vectors of Ref types.
+// rdar://169297366 when fixed will conform WTF::Vector directly to Sequence.
+// We can't do that manually since this would require C++ interop types to be public
+struct CxxRefVectorIterator<Vec: CxxRefVector>: Sequence, IteratorProtocol {
+    typealias Element = Vec.Element
+    var vec: Vec
+    var pos: Int
+
+    init(vec: Vec) {
+        self.vec = vec
+        self.pos = 0
+    }
+
+    mutating func next() -> Vec.Element? {
+        if pos >= vec.size() {
+            return nil
+        }
+        // Safety: we'll make a copy of the referent
+        // before the vector goes out of scope. It's guaranteed
+        // to have a valid lifetime, be initialized, and be
+        // within the vector bounds.
+        let item = unsafe vec.__atUnsafe(pos)
+        pos += 1
+        return unsafe item.pointee.copyRef()
+    }
+}
+
+#endif // ENABLE_BACK_FORWARD_LIST_SWIFT
+
+#endif // compiler(>=6.2)
--- a/Source/WebKit/Shared/API/APIArray.swift
+++ b/Source/WebKit/Shared/API/APIArray.swift
@@ -1,0 +1,45 @@
+// Copyright (C) 2025-2026 Apple Inc. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions
+// are met:
+// 1. Redistributions of source code must retain the above copyright
+//    notice, this list of conditions and the following disclaimer.
+// 2. Redistributions in binary form must reproduce the above copyright
+//    notice, this list of conditions and the following disclaimer in the
+//    documentation and/or other materials provided with the distribution.
+//
+// THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+// AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+// THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+// PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+// BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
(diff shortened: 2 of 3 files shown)


Source/WebKit/UIProcess/StdlibExtras.swift
```diff
@@ -0,0 +1,127 @@
...
+// FIXME (rdar://164119356): Move parts of StdLibExtras.swift into WTF
+// (those parts which are not specific to WebKit-level types) - and enable
+// irrespective of BACK_FORWARD_LIST_SWIFT
+#if ENABLE_BACK_FORWARD_LIST_SWIFT
+
+internal import WebKit_Internal
+internal import wtf
+
+// Conform any WTF::Ref<T> to this protocol to get useful extensions
+protocol CxxRef {
```
adetaylor, line 35: @rr-codes has put extensive comments about this over on https://github.com/WebKit/WebKit/pull/55393 which will need to be acted upon here before this lands

adetaylor: @rr-codes I've made the string changes you proposed (thanks!) Please can you review, especially my safety comments?

I don't believe we can avoid the protocols and manual conformances of `WTF::Vector<WTF::Ref<T>>` and friends, because `SWIFT_CONFORMS_TO_PROTOCOL` doesn't allow us to specify an associated type. It might be that your work to diagnose why `WTF::Vector<T>` isn't automatically a sequence, plus the very recent work in  https://github.com/swiftlang/swift/pull/86421, allows us to discard lots of this in future.

Finally - I haven't put this into its own `WTFString+Extras.swift` file because I want to wait for rdar://164119356 to allow us to put that code _into_ WTF itself!

Source/WebKit/UIProcess/StdlibExtras.swift
```diff
@@ -0,0 +1,142 @@
...
+    associatedtype Element: CxxRef // you only need to specify this in your conformance
+    init()
+    mutating func append(consuming: Element)
+    mutating func reserveCapacity(_ newCapacity: Int)
+    func size() -> Int
+    // swift-format-ignore: AlwaysUseLowerCamelCase, NoLeadingUnderscores
+    func __atUnsafe(_ index: Int) -> UnsafePointer<Element>
+}
+
+extension String {
```
rr-codes: I think all this string stuff can just be

```
extension WTF.String : LosslessStringConvertible {
    public init(_ string: Swift.String) {
        let ns = string as NSString
        let cf: CFString = ns as CFString
        self = WTF.String(cf)
    }

    public var description: Swift.String {
        unsafe createNSString().get()
    }
}

extension WTF.String : ExpressibleByStringLiteral {
    public init(stringLiteral: Swift.String) {
        self.init(stringLiteral)
    }
}

extension WTF.String : ExpressibleByNilLiteral {
    public init(nilLiteral: ()) {
        self = WTF.nullString().pointee
    }
}
```

and no need for anything else

Source/WebKit/UIProcess/StdlibExtras.swift
```diff
@@ -0,0 +1,142 @@
...
+        vec.reserveCapacity(array.count)
+        for item in array {
+            vec.append(consuming: Element(item))
+        }
+        self = vec
+    }
+}
+
+// Iterator for WTF::Vectors of Ref types
+// We do not attempt to conform the Vector itself to Sequence since this would
```
rr-codes: wdym? `WTF::Vector` should automatically be a Swift Sequence since it has begin and end and all that

Source/WebKit/UIProcess/StdlibExtras.swift
```diff
@@ -0,0 +1,142 @@
...
+extension WTF.String: ExpressibleByNilLiteral {
+    public init(nilLiteral: ()) {
+        // Safety: nullString returns a reference to some static data
+        // which is guaranteed to be interpretable as a WTF::String, and as it's
+        // static data, has no lifetime concerns.
+        self = unsafe WTF.nullString().pointee
+    }
+}
+
+extension API.Array {
```
rr-codes: this shouldn’t really be in this file since it’s not really an StdLibExtra it’s just an extension on a WebKit type?

adetaylor: OK, I'll move this out. Do you feel it's necessary or desirable to have an `APIArray.swift` (on the assumption that Swift will eventually be ubiquitous across all WebKit types) or shall I keep that in `WebBackForwardList.swift`?

Source/WebKit/UIProcess/StdlibExtras.swift
adetaylor: Yeah I must admit I wrote that comment before I realised that `WTF::Vector` was supposed to do that. I'll see if I can get `WTF::Vector`'s `Sequence`ness to work (which as discussed elsewhere might be due to a lack of `cplusplus20` somewhere in a modulemap). If not, I'll cite a radar number here before landing this. SGTY?

Source/WebKit/UIProcess/StdlibExtras.swift
adetaylor: Ah, right, I hadn't realised I could use those to convert from a `WTF::String` to a `String` but yeah makes sense - I'll remove the `String` `init`

Source/WebKit/UIProcess/StdlibExtras.swift
```diff
@@ -0,0 +1,142 @@
...
+}
+
+extension WTF.String: ExpressibleByStringLiteral {
+    public init(stringLiteral: Swift.String) {
+        self.init(stringLiteral)
+    }
+}
+
+extension WTF.String: ExpressibleByNilLiteral {
+    public init(nilLiteral: ()) {
```
adetaylor: I discussed with Egor and he's concerned that being able to do `WTF.String(nil)` might get confusing when there are more likely to be `Optional<WTF.String>` types floating around in WebKit. @rr-codes can you explain your rationale for wanting this?

Source/WebKit/UIProcess/StdlibExtras.swift
```diff
@@ -0,0 +1,142 @@
...
+        let cf: CFString = ns as CFString
+        self = WTF.String(cf)
+    }
+
+    public var description: Swift.String {
+        // Safety: `createNSString()` returns a
+        // RetainPtr which is guaranteed to keep
+        // the NSString alive while we construct a Swift string from its
+        // raw pointer.
+        let ns = createNSString()
```
adetaylor: In retrospect we can't do this because it won't work on non-Apple platforms. I'm going back to my original round-tripping functions via spans of characters.

Source/WebKit/UIProcess/StdlibExtras.swift
adetaylor: OK, squashed together a bit - but with the old I'm,plementations instead of those reliant on `NSString`. PTAL

Source/WebKit/UIProcess/StdlibExtras.swift
adetaylor: Moved out.

Source/WebKit/UIProcess/StdlibExtras.swift
adetaylor: Removed it pending this discussion.

Source/WebKit/UIProcess/StdlibExtras.swift
adetaylor: Went back to the old implementation.

Source/WebKit/UIProcess/StdlibExtras.swift
rr-codes: >I discussed with Egor and he's concerned that being able to do WTF.String(nil) might get confusing when there are more likely to be Optional<WTF.String> types floating around in WebKit. @rr-codes can you explain your rationale for wanting this?

It’s because in WebKit, there aren’t really any uses of `std::optional<WTF::String>`, rather; `WTF::String` has its own concept of “is null” and is represented by `WTF::nullString()` (I am personally not a fan of this in the first place but it is the way it is) and so having it be expressible by nil literal would make it consistent with this behavior, since having an optional WTF.String is a bit weird because then there is two “is null” things to check

Source/WebKit/UIProcess/StdlibExtras.swift
```diff
@@ -0,0 +1,122 @@
...
+            // Warning here is rdar://163018821
+            // swift-format-ignore: NeverForceUnwrap
+            let cppspan = unsafe SpanConstChar(ptr.baseAddress!, span.count)
+            return unsafe WTF.String.fromUTF8(cppspan)
+        }
+    }
+
+    /// Return a `Swift.String` from this `WTF.String`.
+    public var description: Swift.String {
+        String(utf8(WTF.LenientConversion).toStdString())
```
rr-codes, line 73: is there a way to avoid having to convert to an std string and rather just copy the bytes to a new Swift.String?

Source/WebKit/UIProcess/StdlibExtras.swift
adetaylor: I've updated the comment to reflect the current situation. Egor and I spent quite a bit of time on it today. I can't get `WTF::Vector` to confirm to `Sequence` even with various modulemap changes. I'll continue poking at it tomorrow but it should not block this PR, please +1 if this is the only remaining objection (which of course it might not be!)

Source/WebKit/UIProcess/StdlibExtras.swift
adetaylor: Yes, I believe we could do this by turning a C++ span into a Sequence somehow. But only if we use `unsafe` and, for now, our [guidance](https://github.com/WebKit/WebKit/wiki/Safer-Swift-Guidelines) is to avoid that all costs. I've added a comment saying this, in case future folks want to choose a different balance of speed vs safety.

Source/WebKit/UIProcess/StdlibExtras.swift
```diff
@@ -0,0 +1,124 @@
...
+internal import wtf
+
+// Conform any WTF::Ref<T> to this protocol to get useful extensions
+protocol CxxRef {
+    associatedtype Pointee // you only need to specify this in your conformance
+    init(_ object: Pointee)
+    func copyRef() -> Self
+}
+
+// Conform any WTF::Vector<WTF::Ref<T>> to this protocol to get useful extensions and iterators
```
rr-codes: extremely minor and optional suggestion but these could be `///` instead of `//` and then they’ll look nicer in Xcode because they’ll be doc comments

rr-codes (approved):

Source/WebKit/UIProcess/StdlibExtras.swift
adetaylor: Will do!

adetaylor: Done

PR #57291, opened 2026-01-26
Swift WebBackForwardList (off by default) - Swift utilities

adetaylor:
#### c4cab0f5e07be8625bbbcf8775ddf7ad02fa024c
<pre>
Swift WebBackForwardList (off by default) - Swift utilities
<a href="https://bugs.webkit.org/show_bug.cgi?id=306874">https://bugs.webkit.org/show_bug.cgi?id=306874</a>
<a href="https://rdar.apple.com/169533295">rdar://169533295</a>

Reviewed by Richard Robinson.

We&apos;re introducing a Swift version of the Back Forward List. For the general
design and rationale, see
<a href="https://github.com/WebKit/WebKit/pull/55393">https://github.com/WebKit/WebKit/pull/55393</a>
It will be landed as a series of separate commits in order to ease code review.

In this commit, we introduce a series of C++ types and functions to call
from Swift. Specifically,
* Swift can&apos;t access templated C++ types so we declare concrete types using
  &apos;using&apos;
* We introduce a few functions that rely on C++ preprocessor macros.
* A few other utility functions which are workarounds for missing functionality
  on the Swift side.

A note on naming: it&apos;s messy that this file is called &apos;*Utilities.h&apos;;
if it contained just types it would be better to call it &apos;*Types.h&apos; but as it&apos;s
a mixture this seemed like the least bad name.

We are landing this commit separately in order to ease code review
of subsequent stages. In itself it&apos;s not useful and has no effect
until subsequent commits land.

Canonical link: <a href="https://commits.webkit.org/306864@main">https://commits.webkit.org/306864@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/9f1cef4ddaa6e614f1c6558aaea3bcb10fb6ea1a

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/142575 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/15039 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/5471 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/151252 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/95764 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/15702 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/15125 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/109660 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/95764 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/145524 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/12120 "Passed tests") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/127620 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/90569 "Passed tests") | | [loading 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/ef357fe7-8034-406f-8514-8b9752a278b1) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/11630 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/9303 "Passed tests") | [✅ 🛠 wpe-libwebrtc](https://ews-build.webkit.org/#/builders/172/builds/1244 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/120990 "Passed tests") | | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/153558 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/14671 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/4733 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/117671 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/14705 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/12767 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/118006 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30098 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/14029 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/124908 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/70362 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/14713 "Built successfully") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/3863 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/14450 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/78415 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/14658 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/14511 "Built successfully") | | | |

### DIFF
--- a/Source/WebKit/UIProcess/WebBackForwardList.cpp
+++ b/Source/WebKit/UIProcess/WebBackForwardList.cpp
@@ -34,6 +34,7 @@
 #include "WebBackForwardCache.h"
 #include "WebBackForwardListCounts.h"
 #include "WebBackForwardListFrameItem.h"
+#include "WebBackForwardListSwiftUtilities.h"
 #include "WebFrameProxy.h"
 #include "WebInspectorUtilities.h"
 #include "WebPageProxy.h"
@@ -856,3 +857,33 @@ String WebBackForwardListWrapper::loggingString()
 #endif // ENABLE(BACK_FORWARD_LIST_SWIFT)
 
 } // namespace WebKit
+
+#if ENABLE(BACK_FORWARD_LIST_SWIFT)
+
+WebCore::BackForwardFrameItemIdentifier generateBackForwardFrameItemIdentifier()
+{
+    return WebCore::BackForwardFrameItemIdentifier::generate();
+}
+
+WebCore::BackForwardItemIdentifier generateBackForwardItemIdentifier()
+{
+    return WebCore::BackForwardItemIdentifier::generate();
+}
+
+// rdar://168139823 is the task of doing a productionized version of WebKit Swift logging
+void doLog(const char* WTF_NONNULL msg)
+{
+    LOG(BackForward, "%s", msg);
+}
+
+void doLoadingReleaseLog(const char* WTF_NONNULL msg)
+{
+    RELEASE_LOG(Loading, "%s", msg);
+}
+// rdar://168139740 is the task of doing a productionized Swift MESSAGE_CHECK
+void messageCheckFailed(Ref<WebKit::WebProcessProxy> process)
+{
+    MESSAGE_CHECK_BASE(false, process->connection());
+}
+
+#endif // ENABLE(BACK_FORWARD_LIST_SWIFT)
--- a/Source/WebKit/UIProcess/WebBackForwardListSwiftUtilities.h
+++ b/Source/WebKit/UIProcess/WebBackForwardListSwiftUtilities.h
@@ -1,0 +1,70 @@
+/*
+ * Copyright (C) 2025-2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+ * THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include "Logging.h"
+#include "WebBackForwardListItem.h"
+#include "WebBackForwardListMessages.h"
+#include "WebProcessProxy.h"
+#include <cstdint>
+#include <wtf/Function.h>
+#include <wtf/Markable.h>
+#include <wtf/RefCountable.h>
+#include <wtf/RefCounted.h>
+#include <wtf/SwiftBridging.h>
+#include <wtf/Vector.h>
+#include <wtf/WeakPtr.h>
+
+#if ENABLE(BACK_FORWARD_LIST_SWIFT)
+
+// Workaround for rdar://162358154
+using SpanConstChar = std::span<const char>;
+
+// These can't be inline due to rdar://162531519
+void doLog(const char* WTF_NONNULL msg); // rdar://168139823
+void doLoadingReleaseLog(const char* WTF_NONNULL msg); // rdar://168139823
+void messageCheckFailed(Ref<WebKit::WebProcessProxy>); // rdar://168139740
+
+// Workaround for rdar://162357139
+template<typename T>
+inline bool contentsMatch(const T& lhs, const T& rhs)
+{
+    return lhs == rhs;
+}
+
+// Workaround for rdar://162193891
+WebCore::BackForwardFrameItemIdentifier generateBackForwardFrameItemIdentifier();
+WebCore::BackForwardItemIdentifier generateBackForwardItemIdentifier();
+
+// Workaround for rdar://129159672
+inline void setOptionalUInt32Value(std::optional<uint32_t>& optional, uint32_t value)
+{
+    optional = value;
+}
+
+using WebBackForwardListItemFilter = WTF::RefCountable<WTF::Function<bool (WebKit::WebBackForwardListItem&)>>;
+
+#endif // ENABLE(BACK_FORWARD_LIST_SWIFT)
--- a/Source/WebKit/Shared/API/APIArray.h
+++ b/Source/WebKit/Shared/API/APIArray.h
@@ -92,6 +92,8 @@ private:
     Vector<RefPtr<Object>> m_elements;
 } SWIFT_SHARED_REFERENCE(refArray, derefArray);
 
+using RefAPIArray = Ref<Array>;
+
 } // namespace API
 
 inline void refArray(API::Array* WTF_NONNULL obj)
--- a/Source/WebKit/Shared/API/APIObject.h
+++ b/Source/WebKit/Shared/API/APIObject.h
@@ -318,6 +318,9 @@ inline API::Object* Object::unwrap(void* object)
 }
 #endif
 
+using VectorRefPtrAPIObject = Vector<RefPtr<Object>>;
+using RefPtrAPIObject = RefPtr<Object>;
+
 } // namespace API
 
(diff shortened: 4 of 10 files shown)


Source/WebKit/UIProcess/WebBackForwardListSwiftUtilities.h
```diff
@@ -0,0 +1,111 @@
...
+#include <wtf/WeakPtr.h>
+
+#if ENABLE(BACK_FORWARD_LIST_SWIFT)
+namespace WebKit {
+class WebBackForwardListAPIImpl;
+};
+#endif
+
+// Workaround for rdar://162358154
+using VectorAPIObject = Vector<RefPtr<API::Object>>;
```
rr-codes: I think it makes more sense for all these `using` declarations to go in the header of their associated type, instead of having all these unrelated types jammed together in this one file (for example, 

```
using RefFrameState = Ref<WebKit::FrameState>
```

makes more sense in FrameState.h, along with all the other related types.

adetaylor: Interesting, I hadn't thought of that.
I've started a chat to find out when rdar://162358154 is likely to be fixed, because if it will be soon, I think we should keep this workaround in one place so we can revert it easily. Otherwise I agree with you.

adetaylor: Done!

Source/WebKit/UIProcess/WebBackForwardList.cpp
```diff
@@ -790,3 +791,28 @@ String WebBackForwardList::loggingString()
...
+{
+    return WebCore::BackForwardFrameItemIdentifier::generate();
+}
+
+WebCore::BackForwardItemIdentifier generateBackForwardItemIdentifier()
+{
+    return WebCore::BackForwardItemIdentifier::generate();
+}
+
+void doLog(const char* WTF_NONNULL msg)
```
rr-codes, line 874: I definitely don’t think these are needed; the standard `Logging` API should be able to be used instead, since the macro ultimately just uses `os_log` on Apple platforms, so the functionality is by definition the same. And then on other platforms it can be bridged to whatever mechanisms they use.

Source/WebKit/UIProcess/WebBackForwardListSwiftUtilities.h
```diff
@@ -0,0 +1,90 @@
...
+// Workaround for rdar://162358154
+using SpanConstChar = std::span<const char>;
+
+// These can't be inline due to rdar://162531519
+void doLog(const char* WTF_NONNULL msg); // rdar://168139823
+void doLoadingReleaseLog(const char* WTF_NONNULL msg); // rdar://168139823
+void messageCheckFailed(Ref<WebKit::WebProcessProxy>); // rdar://168139740
+
+// Workaround for rdar://85881664
+inline API::Object* WTF_NONNULL toAPIObject(WebKit::WebBackForwardListItem* WTF_NONNULL item)
```
rr-codes: can this go into `WebBackForwardListItem.h` instead? I’m not really a fan personally of having “Utilities” files rather than just putting the functions near the types they belong to (just a personal opinion)

Source/WebKit/UIProcess/WebBackForwardListSwiftUtilities.h
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// Workaround for rdar://162519380
+inline RefPtr<WebKit::WebProcessProxy> downcastToWebProcessProxy(WebKit::AuxiliaryProcessProxy* WTF_NONNULL app)
+{
+    return downcast<WebKit::WebProcessProxy>(app);
+}
+
+// Workaround for rdar://162193891
+WebCore::BackForwardFrameItemIdentifier generateBackForwardFrameItemIdentifier();
```
rr-codes, line 59: hm, are you sure this is actually a compiler bug and not a webkit bug? it seems likely that it could be the latter

Source/WebKit/UIProcess/WebBackForwardListSwiftUtilities.h
```diff
@@ -0,0 +1,90 @@
...
+void messageCheckFailed(Ref<WebKit::WebProcessProxy>); // rdar://168139740
+
+// Workaround for rdar://85881664
+inline API::Object* WTF_NONNULL toAPIObject(WebKit::WebBackForwardListItem* WTF_NONNULL item)
+{
+    return item;
+}
+
+// Workaround for rdar://130765784
+inline bool identitiesMatch(const WebKit::WebBackForwardListItem* WTF_NULLABLE lhs, const WebKit::WebBackForwardListItem* WTF_NULLABLE rhs)
```
rr-codes: in StdLibExtras.swift there is

```
func === (_ lhs: WGPUTexture, _ rhs: WGPUTexture) -> Bool {
```

which seems to functionally do the same thing, but seems a slightly nicer approach. Can you do that here instead of these identitiesMatch and contentsMatch functions?

Source/WebKit/UIProcess/WebBackForwardListSwiftUtilities.h
adetaylor: Done

Source/WebKit/UIProcess/WebBackForwardListSwiftUtilities.h
adetaylor: Done

Source/WebKit/UIProcess/WebBackForwardListSwiftUtilities.h
adetaylor: I don't know what to say about this! If you look at the radar I put a ton of diagnosis on it, and I don't know how to work around it in WebKit. If you have any Ideas, let me know.

I've checked that it still applies with the latest toolchain I'm using.

Source/WebKit/UIProcess/WebBackForwardList.cpp
adetaylor: As we've discussed in various other review comments, it's not trivial to replace these. At the very least (even for the release log) this will require different logging objects for Apple platforms vs Android vs journald vs other. And that's just for release logging; it gets even more complex for `LOG` where we have to cache logs to emit during tests, etc. and be aware of different log channels.

I consider it a non-trivial project to design how to expose this nicely in Swift and I want to do it separately, tracked as rdar://168139823. I think it would be a mistake to rush it in this initial Swift change. I think we'll probably end up using the OS log hook even in the WCP, and that's definitely something which needs to be taken very carefully.

I remain convinced that for this initial change, we should use small simple placeholder logging functions like these, while we design the proper replacement.

rr-codes: can you add a comment with a link to a radar to address this in the future?

Source/WebKit/UIProcess/WebBackForwardListSwiftUtilities.h
```diff
@@ -25,18 +25,42 @@
...
-};
+// These can't be inline due to rdar://162531519
+void doLog(const char* WTF_NONNULL msg); // rdar://168139823
+void doLoadingReleaseLog(const char* WTF_NONNULL msg); // rdar://168139823
+void messageCheckFailed(Ref<WebKit::WebProcessProxy>); // rdar://168139740
 
-} // namespace WebCore
+// Workaround for rdar://162357139
+template<typename T>
+inline bool contentsMatch(const T& lhs, const T& rhs)
```
rr-codes, line 53: I think at one point I suggested doing the same thing that Geoff did and override `func ===` in Swift to make it a bit nicer? I think that should be possible here too?

rr-codes (approved):

Source/WebKit/UIProcess/WebBackForwardListSwiftUtilities.h
adetaylor: You did, for a different function called `identitiesMatch`, and I did!

That _would_ also be possible here too (with `func ==` presumably) - but not for templated types. We'd need three or four different Swift functions instead of one C++ function. I'm fine either way - WDYT?

Source/WebKit/UIProcess/WebBackForwardList.cpp
adetaylor: WIll do! It's linked from the header but I'll add it here too

Source/WebKit/UIProcess/WebBackForwardListSwiftUtilities.h
adetaylor: (Discussed with Richard and we'll stick with this, hopefully temporary until that swift bug is fixed)

PR #57363, opened 2026-01-27
[MSE] Fix sample DTS to prevent deletion of preexisting samples when PTS doesn't conflict

eocanha:
#### 13a26d1f906516826613902bd3813f867553644b
<pre>
[MSE] Fix sample DTS to prevent deletion of preexisting samples when PTS doesn&apos;t conflict
<a href="https://bugs.webkit.org/show_bug.cgi?id=305208">https://bugs.webkit.org/show_bug.cgi?id=305208</a>

Reviewed by Alicia Boya Garcia.

Let&apos;s consider a preexisting samples A, B, C, and a new sample N with the
following timestamps:

       Decoding time  Presentation time
       -------------  -----------------
Sample A: 827.826164000  827.826164000
Sample B: 827.826164100  827.826164100
Sample C: 827.826164200  827.826164200
Sample N: 827.826125000  827.909542000

These disparities in DTS vs. PTS offsets between samples A and B exist
because both samples come from completely different videos (eg: because
of ad insertion).

In a timeline:

DTS ···NABC·······
PTS ····ABC···N···

Judging by PTS, the samples A, B, C should remain and theoretically don&apos;t
need to be erased. However, considering DTS, the samples A, B, C have to
be erased, because the sample N would reach the video decoder earlier (by
decoding ordering) and change the state of the decoder, making it
unsuitable for samples A, B, C.

This commit tries to fix the situation by changing sample N (the new
sample) DTS to a value slightly higher than sample C DTS (that is, DTS plus
epsilon, a small value), like this:

       Decoding time  Presentation time
       -------------  -----------------
Sample A:  827.826164000  827.826164000
Sample B:  827.826164100  827.826164100
Sample C:  827.826164200  827.826164200
Sample N: *827.826164201* 827.909542000

DTS ····ABCN······
PTS ····ABC···N···

This fix avoids the deletion of samples A, B, C and the potential
problems associated with that.

A layout test has been added to show the problem and prove that the fix
solves it.

* LayoutTests/media/media-source/media-source-samples-out-of-order-erase-last-previous-frames-expected.txt: Added.
* LayoutTests/media/media-source/media-source-samples-out-of-order-erase-last-previous-frames.html: Added.
* Source/WebCore/platform/graphics/SourceBufferPrivate.cpp:
(WebCore::SourceBufferPrivate::processMediaSample): If a new sync sample is being added and there are next samples in decode order with PTS lower than the one of the new sample, fix the DTS of the new sample by assigning it to have a safeDecodeTime (the DTS of the highest (by DTS order) preexisting sample, plus epsilon), but only if that DTS wouldn&apos;t be higher than the DTS of the next sample that should come after the new sample being added (estimated as sample.decodeTime() + 2*duration, and keeping a safety margin).

Canonical link: <a href="https://commits.webkit.org/307359@main">https://commits.webkit.org/307359@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/b8ec24e9bdb706a4f081110b2a8fa5c0513d4ec5

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/143994 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/16673 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/8226 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/152664 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/97233 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/145869 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/17155 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/16566 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/110725 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/79596 "Exiting early after 60 failures. 15479 tests run. 60 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/146957 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/13149 "Passed tests") | [❌ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/129387 "Found 2 new API test failures: TestWebKitAPI.WKWebExtensionDataRecord.GetDataRecordsForMultipleContexts, TestWebKitAPI.MediaLoading.LockdownModeHLS (failure)") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/91644 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/12616 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/10347 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/110 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/122086 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/6032 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/154976 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/16525 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/7087 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/118737 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/16561 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/13887 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/119091 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30557 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/14998 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/127243 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/71946 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/16147 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/5691 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/15881 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/79926 "Built successfully") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/16092 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/15947 "Built successfully") | | |

### DIFF
--- a/LayoutTests/media/media-source/media-source-samples-out-of-order-erase-last-previous-frames.html
+++ b/LayoutTests/media/media-source/media-source-samples-out-of-order-erase-last-previous-frames.html
@@ -1,0 +1,225 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <title>media-source-samples-out-of-order-erase-last-previous-frames</title>
+    <script src="mock-media-source.js"></script>
+    <script src="../video-test.js"></script>
+    <script>
+    var source;
+    var sourceBuffer;
+    var initSegment;
+    var mediaSegment;
+
+    if (window.internals)
+        internals.initializeMockMediaSource();
+
+    window.addEventListener('load', async event => {
+        findMediaElement();
+
+        source = new MediaSource();
+        run('video.src = URL.createObjectURL(source)');
+        await waitFor(source, 'sourceopen');
+
+        run('sourceBuffer = source.addSourceBuffer("video/mock; codecs=mock")');
+
+        // Test 1.
+
+        consoleWrite("Test 1: First segment has normal, monotonically increasing samples with decode timestamp equal ");
+        consoleWrite("to presentation timestamp.");
+
+        mediaSegment = concatenateSamples([
+            makeAInit(4, [makeATrack(0, 'mock', TRACK_KIND.VIDEO)]),
+            // Syntax: makeASample(presentationTime, decodeTime, duration, timeScale, trackID, flags, generation)
+            makeASample( 0,  0, 10, 10, 0, SAMPLE_FLAG.SYNC, 0),
+            makeASample(10, 10, 10, 10, 0, SAMPLE_FLAG.NONE, 0),
+            makeASample(20, 20, 10, 10, 0, SAMPLE_FLAG.NONE, 0),
+            makeASample(30, 30, 10, 10, 0, SAMPLE_FLAG.NONE, 0),
+        ]);
+        run('sourceBuffer.appendBuffer(mediaSegment)');
+        await waitFor(sourceBuffer, 'updateend');
+
+        bufferedSamples = await internals.bufferedSamplesForTrackId(sourceBuffer, 0);
+        testExpected("bufferedSamples.length", 4);
+        bufferedSamples.forEach(consoleWrite);
+
+        consoleWrite("Test 1: Second, segment with overlap in decoding order but no overlap in presentation order shouldn't ");
+        consoleWrite("cause any deletion if decoding order is corrected to avoid conflicts.");
+
+        mediaSegment = concatenateSamples([
+            makeAInit(1, [makeATrack(0, 'mock', TRACK_KIND.VIDEO)]),
+            makeASample(40, 9, 50, 10, 0, SAMPLE_FLAG.SYNC, 1),
+        ]);
+        run('sourceBuffer.appendBuffer(mediaSegment)');
+        await waitFor(sourceBuffer, 'updateend');
+
+        bufferedSamples = await internals.bufferedSamplesForTrackId(sourceBuffer, 0);
+        testExpected("bufferedSamples.length", 5);
+        bufferedSamples.forEach(consoleWrite);
+
+        run('sourceBuffer.remove(0, Infinity)');
+        await waitFor(sourceBuffer, 'updateend');
+
+        // Test 2.
+
+        consoleWrite("Test 2: First segment has normal, monotonically increasing samples with decode timestamp equal ");
+        consoleWrite("to presentation timestamp, and a last sync sample.");
+
+        mediaSegment = concatenateSamples([
+            makeAInit(4, [makeATrack(0, 'mock', TRACK_KIND.VIDEO)]),
+            makeASample(  0,   0, 10, 10, 0, SAMPLE_FLAG.SYNC, 0),
+            makeASample( 10,  10, 10, 10, 0, SAMPLE_FLAG.NONE, 0),
+            makeASample( 20,  20, 10, 10, 0, SAMPLE_FLAG.NONE, 0),
+            makeASample( 30,  30, 10, 10, 0, SAMPLE_FLAG.NONE, 0),
+            makeASample(100, 100, 10, 10, 0, SAMPLE_FLAG.SYNC, 0)
+        ]);
+        run('sourceBuffer.appendBuffer(mediaSegment)');
+        await waitFor(sourceBuffer, 'updateend');
+
+        bufferedSamples = await internals.bufferedSamplesForTrackId(sourceBuffer, 0);
+        testExpected("bufferedSamples.length", 5);
+        bufferedSamples.forEach(consoleWrite);
+
+        consoleWrite("Test 2: Second, segment with overlap in decoding order but no overlap in presentation order shouldn't ");
+        consoleWrite("cause any deletion if decoding order is corrected to avoid conflicts.");
+
+        mediaSegment = concatenateSamples([
+            makeAInit(1, [makeATrack(0, 'mock', TRACK_KIND.VIDEO)]),
+            makeASample(40, 9, 50, 10, 0, SAMPLE_FLAG.SYNC, 1),
+        ]);
+        run('sourceBuffer.appendBuffer(mediaSegment)');
+        await waitFor(sourceBuffer, 'updateend');
+
+        bufferedSamples = await internals.bufferedSamplesForTrackId(sourceBuffer, 0);
+        testExpected("bufferedSamples.length", 6);
+        bufferedSamples.forEach(consoleWrite);
+
+        run('sourceBuffer.remove(0, Infinity)');
+        await waitFor(sourceBuffer, 'updateend');
+
+        // Test 3.
+
+        consoleWrite("Test 3: First segment has normal, monotonically increasing samples with decode timestamp equal ");
+        consoleWrite("to presentation timestamp.");
+
+        mediaSegment = concatenateSamples([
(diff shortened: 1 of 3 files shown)


Source/WebCore/platform/graphics/SourceBufferPrivate.cpp
ntrrgc: `- contiguousFrameTolerance - contiguousFrameTolerance` is equivalent to `- 2 * contiguousFrameTolerance`.

ntrrgc: The variable is called `lastSampleToErase`, but this is a bit confusing because this entire code section is trying to avoid it being deleted. I understand that the real meaning is that it is the last sample that would be erased if we don't do something to avoid it, but this is not immediately obvious.

`lastSampleBeforeSync`, maybe? Or any better name you can think of.

ntrrgc: `nextSyncSample->first` is the decode key of `nextSyncSample`, so -- assuming I got this right -- what we're trying to do here is to get the sample immediately before the `nextSyncSample` forward iterator. We don't need to perform an extra search for it or use reverse iterators, we can replace all of this with:

```cpp
if (/* ... */ && nextSyncSample != trackBuffer.samples().decodeOrder().begin()) {
    auto lastSampleBeforeSyncRef = std::prev(nextSyncSample)->second;
```

The intention is clearer this way and it allows us to get rid of a few if statements.

ntrrgc: I can imagine a case where the duration of the sample is such that this code doesn't trigger and we get the problem -- e.g. consider the same appends as in the first layout tests but with all durations being the same.

However, given the risk of regressions I'm not going to ask you to venture into that before this becomes a problem in practice.

LayoutTests/media/media-source/media-source-samples-out-of-order-erase-last-previous-frames.html
ntrrgc, line 218: All these tests use samples of different duration for the second append. I would recommend also testing the more realistic case where the durations don't change, like in the example described in your commit message. It would be a bit disappointing if the patch (or later changes to WebKit) end up only handling cases different from the original bug.

ntrrgc (approved):
Comments aside, looks good to me.

As an aside: I'm increasingly questioning whether the MSE algorithms should be relying on DTS as meaningful anywhere but within the context of a single GOP, precisely because of problems like this that arise when mixing content from multiple sources.

Source/WebCore/platform/graphics/SourceBufferPrivate.cpp
eocanha: I know, ~~but MediaTime only allows a limited set of operations, and multiplication isn't one of them~~.

Oh! MediaTimes [actually have multiplication](https://github.com/WebKit/WebKit/blob/main/Source/WTF/wtf/MediaTime.h#L77)! Ok. I'll use that instead. Thanks for pointing it out.

eocanha: > it is the last sample that would be erased if we don't do something to avoid it

Yes, that was the meaning, and I agree about the name being a poor one. `lastSampleBeforeSync` (plus a comment explaining why that sample is important, because it would have been the last one to be deleted if we don't do anything to avoid it) is a good idea. Thanks for proposing it.

eocanha: wait, wait... can we iterate backwards on direct iterators? :open_mouth:
Looking at the C++ reference, that can only be done if [the iterator is bidirectional](https://en.cppreference.com/w/cpp/iterator/prev.html). I don't know if the iterator created by `decodeOrder() `meets that requirement (I'll check it). If it does and `prev()` works as expected, I'll use your proposed solution, since it's clearly simpler and better. Thanks!

eocanha: I wanted to be as restrictive as possible with the conditions that trigger the fix, in order to minimize the chances of creating unexpected problems/regressions. The current conditions apply to the original use case that triggered the unwanted deletion, so better let's be on the safe side. If anybody complains about these unwanted deletions in the future, we can analyze theis specific case and relax the requirements that trigger the fix, if you think it's a good idea.

LayoutTests/media/media-source/media-source-samples-out-of-order-erase-last-previous-frames.html
eocanha: I'll try to come up with a layout test that uses the same duration as well.

Source/WebCore/platform/graphics/SourceBufferPrivate.cpp
eocanha: While implementing this I ran into some tricky realizations: `std::prev()` is Undefined Behaviour when applied on `begin()` (because you can't have anything semantically equivalent to `rend()` in a forward iterator; that's why you added the extra check in the `if`). But also, what if `nextSyncSample` is `end()`? We need to make sure that there's at least one item in the collection to iterate back to. That's why I ended up adding an extra `trackBuffer.samples().decodeOrder().size() > 0` condition to the `if`.

eocanha:
The failing api-mac tests seem unrelated to MSE and the win-tests are exiting early even without the patch, so I'm considering the patch as is ready to land. Merging.

PR #57493, opened 2026-01-29
Reserve capacity for DocumentSharedObjectPool to improve perf of parserSetAttributes

NathanSolomon733:
#### e25d72b12cd9aea37de3dd55b87620b9c33ed13c
<pre>
Reserve capacity for DocumentSharedObjectPool to improve perf of parserSetAttributes
<a href="https://bugs.webkit.org/show_bug.cgi?id=306548">https://bugs.webkit.org/show_bug.cgi?id=306548</a>
<a href="https://rdar.apple.com/169193738">rdar://169193738</a>

Reviewed by Ryosuke Niwa.

We spend ~12% of time in WebCore::DocumentSharedObjectPool::cachedShareableElementDataWithAttributes
during parserSetAttributes which adds attributes to a HashSet&lt;Ref, ShareableElementDataHash&gt;.
After a closer look at the disassembly of cachedShareableElementDataWithAttributes(), we spend ~38%
of the time of this function doing quadratic probing.

Track peak element count (rounded up to nearest power of 2) per RegistrableDomain.
On subsequent visits to the same domain, reserve the set&apos;s capacity upfront with
its previously known peak count.

After reserving initial capacity, profiling shows that
DocumentSharedObjectPool::cachedShareableElementDataWithAttributes now shows
fewer quadratic probing operations.

This also reduces hash set resize events that were occurring.

This shows a small perf improvement.

(WebCore::Document::setParsing):
* Source/WebCore/dom/DocumentSharedObjectPool.cpp:
(WebCore::peakSizeInPast):
(WebCore::DocumentSharedObjectPool::DocumentSharedObjectPool):
(WebCore::DocumentSharedObjectPool::~DocumentSharedObjectPool):
* Source/WebCore/dom/DocumentSharedObjectPool.h:

Canonical link: <a href="https://commits.webkit.org/307651@main">https://commits.webkit.org/307651@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/7dd61335030ddcea9e36d6ebce0467279acaf09a

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/144667 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/17346 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/8950 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/153338 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/98302 "Built successfully") | [❌ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/65b87199-3788-40f5-9e42-d2c52eb685be) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/146542 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/17829 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/17240 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/111257 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/79774 "Passed tests") | [loading 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b0b927f1-8f96-4b8d-a018-6d7693386b13) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/147630 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/13618 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/129894 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/92152 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/12994 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/10749 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/783 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/122505 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/6604 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/155650 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/17198 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/7684 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/119274 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/17237 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/14381 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/119591 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30743 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/15408 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/127846 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/72740 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/16820 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/6201 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/16556 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/80599 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/16765 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/16620 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/dom/DocumentSharedObjectPool.cpp
+++ b/Source/WebCore/dom/DocumentSharedObjectPool.cpp
@@ -36,6 +36,26 @@ namespace WebCore {
 
 WTF_MAKE_TZONE_ALLOCATED_IMPL(DocumentSharedObjectPool);
 
+static HashMap<RegistrableDomain, size_t>& peakSizeInPast()
+{
+    static MainThreadNeverDestroyed<HashMap<RegistrableDomain, size_t>> map;
+    return map;
+}
+
+DocumentSharedObjectPool::DocumentSharedObjectPool(RegistrableDomain&& domain)
+    : m_domain(WTF::move(domain))
+{
+    m_shareableElementDataCache.reserveInitialCapacity(peakSizeInPast().get(m_domain));
+}
+
+DocumentSharedObjectPool::~DocumentSharedObjectPool()
+{
+    size_t currentSizeRoundedUp = roundUpToPowerOfTwo(m_shareableElementDataCache.size());
+    auto result = peakSizeInPast().add(m_domain, currentSizeRoundedUp);
+    if (!result.isNewEntry && currentSizeRoundedUp > result.iterator->value)
+        result.iterator->value = currentSizeRoundedUp;
+}
+
 struct DocumentSharedObjectPool::ShareableElementDataHash {
     static unsigned hash(const Ref<ShareableElementData>& data)
     {
--- a/Source/WebCore/dom/DocumentSharedObjectPool.h
+++ b/Source/WebCore/dom/DocumentSharedObjectPool.h
@@ -26,6 +26,7 @@
 
 #pragma once
 
+#include "RegistrableDomain.h"
 #include <memory>
 #include <wtf/HashSet.h>
 #include <wtf/RefPtr.h>
@@ -40,12 +41,15 @@ class ShareableElementData;
 class DocumentSharedObjectPool {
     WTF_MAKE_TZONE_ALLOCATED(DocumentSharedObjectPool);
 public:
+    explicit DocumentSharedObjectPool(RegistrableDomain&&);
+    ~DocumentSharedObjectPool();
     Ref<ShareableElementData> cachedShareableElementDataWithAttributes(std::span<const Attribute>);
 
 private:
     struct ShareableElementDataHash;
     using ShareableElementDataCache = HashSet<Ref<ShareableElementData>, ShareableElementDataHash>;
     ShareableElementDataCache m_shareableElementDataCache;
+    RegistrableDomain m_domain;
 };
 
 } // namespace WebCore
--- a/Source/WebCore/dom/Document.cpp
+++ b/Source/WebCore/dom/Document.cpp
@@ -4386,7 +4386,7 @@ void Document::setParsing(bool b)
     m_bParsing = b;
 
     if (m_bParsing && !m_sharedObjectPool)
-        m_sharedObjectPool = makeUnique<DocumentSharedObjectPool>();
+        m_sharedObjectPool = makeUnique<DocumentSharedObjectPool>(RegistrableDomain { securityOrigin().data() });
 
     if (!m_bParsing && view() && !view()->needsLayout())
         protect(view())->fireLayoutRelatedMilestonesIfNeeded();


Source/WebCore/dom/DocumentSharedObjectPool.cpp
rniwa: How did you come to pick 1024 as the appropriate initial capacity?
It seems like a document with just a couple of elements, that will be an overkill.

Did you run memory tests to make sure this won't regress memory usage?

NathanSolomon733: 1024 was chosen as the first initial capacity that would eliminate all resize events in SP3. Logging on real world sites shows that the size of the hash set rarely exceeds even 256. This reserving of initial capacity is really for improving perf in the extreme cases, which is more common in SP3. 
When I tested this change in our memory benchmarks though, they all came back neutral with no regressions. So, while this change will not help the common use cases, the overhead of reserving initial capacity is not noticeable in our memory benchmarks, which I assume to include the normal case and extreme cases.

We still get a very small perf improvement with reserving to 512, so I am going to amend this patch to use 512 instead.

rniwa: Can we maybe remember/cache how many slots are used in the past per domain and use that?

NathanSolomon733: That could work, but I'm not sure where we would store that structure. The DocumentSharedObjectPool is stored per Document, and each navigation constructs a new Document. We'd have to pass in the previous usage amount to the DocumentSharedObjectPool on construction, and that data structure would have to persist across different WebContent processes being created/destroyed.
Would we have to store that in the UI process and pass it in via IPC during page construction, and also send out the final usage via IPC once parsing completes?

rniwa: oh, I was thinking that maybe we can just remember that in-process in WebContent. Even with PSON or site isolation, we're not gonna swap processes between the same site navigations so remembering that we've reached a high peak will be suffice.

NathanSolomon733: Oh I see, you're referring to a per-process static cache that is keyed by domain, and the value can be the max size reached. I can implement that.

rniwa: yeah, that way, if we've reached a high peak in a previous document, we'd avoid the cost in the next document. Seems like a pretty decent optimization strategy.

NathanSolomon733: I am A/B testing that change now. I was going to wait until I get back the results to amend this PR, but if preferred I can just push it early.

NathanSolomon733: This has been delayed because I got inconclusive results in A/B testing if the latest version introduces other regressions. I got some (likely false/inconsistent) regressions on some subtests, so I am re A/B testing and have been waiting for those re-tests to complete.
Full details are in Radar.

NathanSolomon733: New re-run results show this is still a small progression. The regressions I was seeing were not reproducible, so I still think this patch is a small improvement.

rniwa: I'd call this `peakCapacityInPast` or `maxPeakCapacity` instead of `capacityHints`, which seems rather vague.

rniwa: Why not just use get? Surely, nobody should be adding 0 as a value to this map.

rniwa: Hm... I think we're better off remembering the current `capacity`, not `size` since `size` may not be a multiple of two.

rniwa: This will do two hash map lookups, which is inefficient. We should use `add` instead like this:
```cpp
auto result = capacityHints().add(m_domain, currentCapacity);
if (!result.isNewEntry && currentCapacity > result.iterator->value)
    result.iterator->value = currentCapacity;
```

Source/WebCore/dom/DocumentSharedObjectPool.h
rniwa, line 46: We should probably remember the peak capacity in this function instead of using the size at the time of destruction.

Source/WebCore/dom/DocumentSharedObjectPool.cpp
NathanSolomon733: Changing this now. I did not realize that get() returns a default value of 0 for size_t values if there's no entry in the map.

Source/WebCore/dom/DocumentSharedObjectPool.h
NathanSolomon733: I could add a member variable like m_peakCapacity and update it, but since there are no operations that remove entries from m_shareableElementDataCache, the capacity at destruction time should equal the peak capacity. I don't see how the capacity could be larger at any earlier point in the pool's lifetime.

rniwa: ah, ok.

Although on the other hand... if we never evict an entry, checking how many of the entries are relevant at the end of time may tell us whether we should be doing removal or not.

Also... should we consider keeping the cache across same-origin document navigations? It seems plausible that multiple documents from the same site will have a similar set of attributes, etc...

rniwa (approved):

Source/WebCore/dom/DocumentSharedObjectPool.cpp
```diff
@@ -36,6 +36,26 @@ namespace WebCore {
...
+    : m_domain(WTF::move(domain))
+{
+    m_shareableElementDataCache.reserveInitialCapacity(peakCapacityInPast().get(m_domain));
+}
+
+DocumentSharedObjectPool::~DocumentSharedObjectPool()
+{
+    size_t currentCapacity = m_shareableElementDataCache.capacity();
+    auto result = peakCapacityInPast().add(m_domain, currentCapacity);
+    if (!result.isNewEntry && currentCapacity > result.iterator->value)
```
rniwa: Actually, this condition means that this cached capacity will only ever increase.
That's not great if we navigate from a large page to a small page.
So maybe we do indeed want to use `size()` but just round it up to the powers of two.
Basically use `Malloc::nextCapacity(m_shareableElementDataCache.size())`.

NathanSolomon733: Agreed, I found out from the layout test results it's a pathological increase to use capacity and then pass it in to reserveInitialCapacity.
capacity() returns buckets allocated,  but reserveInitialCapacity(N) interprets N as elements to store and determines the bucket count needed to not exceed the load factor. To maintain <= 75% load factor, it must allocate more than N buckets for N elements. Internally, reserveInitialCapacity(N) calls shouldExpand(N, N). Since N/N = 100% load, which exceeds the 75% threshold, it doubles the table size to 2N buckets.
This creates a loop where reserveInitialCapacity(8) makes 16 buckets, reserve(16) makes 32 buckets, etc.

NathanSolomon733: Instead of Malloc::nextCapacity, I need to use roundUpToPowerOfTwo(m_shareableElementDataCache.size()), which is what HashTable.h uses internally for sizing.

 Malloc is a template parameter inside Vector/HashTable, so it's not accessible outside those classes. I could call FastMalloc::nextCapacity, but that returns 1.5x the input. roundUpToPowerOfTwo rounds up to the nearest power of 2 which is what we want.

PR #57506, opened 2026-01-29
webkitexpectationspy: a package for test expectations that supports plugins for different test harnesses

Smackteo:
#### 575bd26bf14f2880c4ec145dc952fc4f89766a82
<pre>
webkitexpectationspy: a package for test expectations that supports plugins for different test harnesses
<a href="https://rdar.apple.com/169210427">rdar://169210427</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=306564">https://bugs.webkit.org/show_bug.cgi?id=306564</a>

Reviewed by Jonathan Bedard.

* Tools/Scripts/libraries/webkitexpectationspy/MANIFEST.in: Added.
* Tools/Scripts/libraries/webkitexpectationspy/README.md: Added.
* Tools/Scripts/libraries/webkitexpectationspy/run-tests: Added.
* Tools/Scripts/libraries/webkitexpectationspy/setup.py: Added.
(readme):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/__init__.py: Added.
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/configuration.py: Added.
(ConfigurationCategory):
(get_token_category):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py: Added.
(Expectation):
Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/filesystem.py: Added.
(FilesystemProtocol):

* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/linter.py: Added.
(LintWarning):
(LintWarning.__init__):
(LintWarning.__str__):
(LintWarning.__repr__):
(LineFix):
(LineFix.__init__):
(ExpectationsLinter):
(ExpectationsLinter.__init__):
(ExpectationsLinter.lint):
(ExpectationsLinter.get_fixes):
(ExpectationsLinter.apply_fixes):
(ExpectationsLinter._parse_all_lines):
(ExpectationsLinter._check_category_ordering):
(ExpectationsLinter._check_alphabetical_ordering):
(ExpectationsLinter._check_combination_collapse):
(ExpectationsLinter._check_universal_skip):
(ExpectationsLinter._add_reorder_fix):
(ExpectationsLinter._add_reorder_fix.sort_key):
(ExpectationsLinter._add_collapse_fix):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/manager.py: Added.
(ExpectationsManager):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/mocks/__init__.py: Added.
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/model.py: Added.
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/parser.py: Added.
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/plugins/__init__.py: Added.
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/plugins/api_tests.py: Added.
(APITestPlugin):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/plugins/base.py: Added.
(FormatPlugin):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/plugins/layout_tests.py: Added.
(LayoutTestPlugin):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/py.typed: Added.
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/tests/__init__.py: Added.
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/tests/test_api_plugin.py: Added.
(APITestPluginTest):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/tests/test_layout_plugin.py: Added.
(LayoutTestPluginTest):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/tests/test_linter.py: Added.
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/tests/test_manager.py: Added.
(MockFilesystem):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/tests/test_model.py: Added.
(ExpectationsModelTest):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/tests/test_parser.py: Added.
(ExpectationParserTest):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/tests/test_version_specifier.py: Added.
(VersionSpecifierParseTest):
* Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/version_specifier.py: Added.
(VersionSpecifier):

Canonical link: <a href="https://commits.webkit.org/312308@main">https://commits.webkit.org/312308@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/9eaf1b37dc012e67c6dcf06ce4de7d2071790924

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/159501 "Passed style check") | [  ~~🛠 ios~~](https://ews-build.webkit.org/#/builders/159/builds/32968 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 mac~~](https://ews-build.webkit.org/#/builders/168/builds/26046 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [  ~~🛠 wpe~~](https://ews-build.webkit.org/#/builders/5/builds/168339 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/113881 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/a2a3c500-0a6a-40f2-af41-7f9beff9fc30) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/161370 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/33002 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/32921 "Built successfully") | [  ~~🧪 wpe-wk2~~](https://ews-build.webkit.org/#/builders/5/builds/168339 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/113881 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/a2ee3c30-57e6-4b1f-9ad6-f64a265a1bb7) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/162458 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/162/builds/25842 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/143269 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-wpe~~](https://ews-build.webkit.org/#/builders/5/builds/168339 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/c3eba083-da66-407d-947a-fa6b1d2b90b0) 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/158821 "Passed tests") | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/24895 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/23350 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/16109 "Built successfully") | | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/134586 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-wk1~~](https://ews-build.webkit.org/#/builders/167/builds/21041 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 gtk~~](https://ews-build.webkit.org/#/builders/2/builds/170830 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | | [  ~~🧪 mac-wk2~~](https://ews-build.webkit.org/#/builders/169/builds/22672 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/2/builds/170830 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [  ~~🛠 vision~~](https://ews-build.webkit.org/#/builders/153/builds/32623 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/27426 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-gtk~~](https://ews-build.webkit.org/#/builders/2/builds/170830 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35671 "Built successfully and passed tests") | [  ~~🛠 vision-sim~~](https://ews-build.webkit.org/#/builders/160/builds/32575 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-wk2-stress~~](https://ews-build.webkit.org/#/builders/8/builds/142834 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/90714 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [  ~~🧪 vision-wk2~~](https://ews-build.webkit.org/#/builders/160/builds/32575 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/19648 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [  ~~🛠 tv~~](https://ews-build.webkit.org/#/builders/158/builds/32141 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [  ~~🛠 tv-sim~~](https://ews-build.webkit.org/#/builders/157/builds/31600 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [  ~~🛠 watch~~](https://ews-build.webkit.org/#/builders/163/builds/31872 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [  ~~🛠 watch-sim~~](https://ews-build.webkit.org/#/builders/152/builds/31789 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | |

### DIFF
--- a/Tools/Scripts/libraries/webkitexpectationspy/run-tests
+++ b/Tools/Scripts/libraries/webkitexpectationspy/run-tests
@@ -1,0 +1,28 @@
+#!/usr/bin/env python3
+
+# Copyright (C) 2024-2026 Apple Inc. All rights reserved.
+# See LICENSE file for license terms.
+
+import os
+import logging
+import platform
+import sys
+import webkitexpectationspy
+
+from webkitcorepy import AutoInstall, log, __path__ as wcp_paths
+from webkitcorepy.testing import PythonTestRunner
+
+
+libraries = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+AutoInstall.set_directory(os.path.join(libraries, 'autoinstalled', 'python-{}-{}'.format(sys.version_info[0], platform.machine())))
+
+if __name__ == '__main__':
+    logging.basicConfig(level=logging.WARNING)
+
+    sys.exit(PythonTestRunner(
+        description='Run unit tests for the webkitexpectationspy module.',
+        modules={
+            os.path.dirname(__file__): ['webkitexpectationspy'],
+            os.path.dirname(wcp_paths[0]): ['webkitcorepy'],
+        }, loggers=[logging.getLogger(), log],
+    ).main(*sys.argv[1:]))
--- a/Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/__init__.py
+++ b/Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/__init__.py
@@ -1,0 +1,57 @@
+# Copyright (C) 2024-2026 Apple Inc. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1.  Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2.  Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+# ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""webkitexpectationspy - Test expectations management for WebKit.
+
+A system for parsing, querying, and linting test expectation files.
+Supports API tests and layout tests with suite-specific plugins.
+
+Example usage:
+    from webkitexpectationspy import ExpectationsManager, ResultStatus
+    from webkitexpectationspy.suites import APITestSuite
+
+    manager = ExpectationsManager(suite=APITestSuite())
+    manager.load_file('TestExpectations')
+
+    exp = manager.get_expectation('TestWebKitAPI.WTF.StringTest')
+    if exp.skip:
+        print('Test is skipped')
+    if ResultStatus.FAIL in exp.expected:
+        print('Test expected to fail')
+"""
+
+__version__ = '3.0.0'
+
+from webkitexpectationspy.expectations import Expectation, ResultStatus
+from webkitexpectationspy.modifiers import ModifiersBase, LayoutTestModifiers
+from webkitexpectationspy.suites.layout_tests import LayoutTestStatus
+from webkitexpectationspy.manager import ExpectationsManager
+
+__all__ = [
+    '__version__',
+    'Expectation',
+    'ExpectationsManager',
+    'ResultStatus',
+    'LayoutTestStatus',
+    'ModifiersBase',
+    'LayoutTestModifiers',
+]
--- a/Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/configuration.py
+++ b/Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/configuration.py
@@ -1,0 +1,227 @@
+# Copyright (C) 2024-2026 Apple Inc. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1.  Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2.  Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+# ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""Configuration categories, typed enums, and ConfigurationSpecifier for test expectations."""
+
+import enum
+import re
+from dataclasses import dataclass, field
+from typing import FrozenSet, Optional, Set, List
+
+
+class ConfigurationCategory(enum.Enum):
+    """Configuration category — declaration order defines the canonical ordering."""
+    PLATFORM = 'platform'
+    VERSION = 'version'
+    STYLE = 'style'
+    HARDWARE = 'hardware'
+    ARCHITECTURE = 'architecture'
+    FLAVOR = 'flavor'
+
+
+_CATEGORY_ORDER = {cat: i for i, cat in enumerate(ConfigurationCategory)}
+
+
+class Platform(enum.Enum):
+    MAC = 'mac'
+    IOS = 'ios'
+    WATCHOS = 'watchos'
+    TVOS = 'tvos'
+    VISIONOS = 'visionos'
+    LINUX = 'linux'
+    WIN = 'win'
+    GTK = 'gtk'
+    WPE = 'wpe'
+
+
+class BuildType(enum.Enum):
+    DEBUG = 'debug'
+    RELEASE = 'release'
+    PRODUCTION = 'production'
+    ASAN = 'asan'
+    GUARDMALLOC = 'guardmalloc'
+
+
+class Architecture(enum.Enum):
+    ARM64 = 'arm64'
+    X86_64 = 'x86_64'
+    X86 = 'x86'
+    ARM64_32 = 'arm64_32'
+    ARMV7K = 'armv7k'
+
+
+class Hardware(enum.Enum):
+    SIMULATOR = 'simulator'
+    DEVICE = 'device'
+    IPHONE = 'iphone'
+    IPAD = 'ipad'
+    VISION = 'vision'
+    WATCH = 'watch'
+    TV = 'tv'
+
+
+_PLATFORM_BY_VALUE = {p.value: p for p in Platform}
+_BUILD_TYPE_BY_VALUE = {b.value: b for b in BuildType}
+_ARCHITECTURE_BY_VALUE = {a.value: a for a in Architecture}
+_HARDWARE_BY_VALUE = {h.value: h for h in Hardware}
+
+PLATFORM_TOKENS = frozenset(p.value for p in Platform)
+STYLE_TOKENS = frozenset(b.value for b in BuildType)
+HARDWARE_TOKENS = frozenset(h.value for h in Hardware)
+ARCHITECTURE_TOKENS = frozenset(a.value for a in Architecture)
+
+
+# Future-proofing escape hatch: any lowercase token ending in "os" (e.g. a
+# future platform like "xros") is recognized as a platform without requiring
+# the Platform enum to be amended. The enum remains the source of truth for
+# enumerated iteration and linter token ordering; this regex only broadens
+# classification.
+_PLATFORM_SHAPE_RE = re.compile(r'^[a-z]+os$')
+
+CATEGORY_TOKENS = {
+    ConfigurationCategory.PLATFORM: PLATFORM_TOKENS,
+    ConfigurationCategory.STYLE: STYLE_TOKENS,
+    ConfigurationCategory.HARDWARE: HARDWARE_TOKENS,
+    ConfigurationCategory.ARCHITECTURE: ARCHITECTURE_TOKENS,
+}
+
+CATEGORY_NAMES = {
+    ConfigurationCategory.PLATFORM: 'Platform',
+    ConfigurationCategory.VERSION: 'Version',
+    ConfigurationCategory.STYLE: 'Style',
+    ConfigurationCategory.HARDWARE: 'Hardware',
+    ConfigurationCategory.ARCHITECTURE: 'Architecture',
+    ConfigurationCategory.FLAVOR: 'Flavor',
+}
+
+
+@dataclass(frozen=True)
+class ConfigurationSpecifier:
+    platforms: FrozenSet[Platform] = field(default_factory=frozenset)
+    architectures: FrozenSet[Architecture] = field(default_factory=frozenset)
+    build_types: FrozenSet[BuildType] = field(default_factory=frozenset)
+    hardware: FrozenSet[Hardware] = field(default_factory=frozenset)
+    flavors: FrozenSet[str] = field(default_factory=frozenset)
+    version_specifier: Optional[object] = None
+
+    @classmethod
+    def from_tokens(cls, tokens: Set[str], version_specifiers=None):
+        platforms = frozenset(
+            _PLATFORM_BY_VALUE[t] for t in tokens if t in _PLATFORM_BY_VALUE
+        )
+        build_types = frozenset(
+            _BUILD_TYPE_BY_VALUE[t] for t in tokens if t in _BUILD_TYPE_BY_VALUE
+        )
+        architectures = frozenset(
+            _ARCHITECTURE_BY_VALUE[t] for t in tokens if t in _ARCHITECTURE_BY_VALUE
+        )
+        hardware_set = frozenset(
+            _HARDWARE_BY_VALUE[t] for t in tokens if t in _HARDWARE_BY_VALUE
+        )
+        known = PLATFORM_TOKENS | STYLE_TOKENS | HARDWARE_TOKENS | ARCHITECTURE_TOKENS
+        flavors = frozenset(t for t in tokens if t not in known)
+
+        vs = tuple(version_specifiers) if version_specifiers else None
+
+        return cls(
+            platforms=platforms,
+            architectures=architectures,
+            build_types=build_types,
+            hardware=hardware_set,
+            flavors=flavors,
+            version_specifier=vs,
+        )
+
+    def matches(self, current_tokens: Set[str],
+                current_version: Optional[str] = None,
+                version_order: Optional[List[str]] = None) -> bool:
+        if self.platforms:
+            if not any(p.value in current_tokens for p in self.platforms):
+                return False
+        if self.build_types:
+            if not any(b.value in current_tokens for b in self.build_types):
+                return False
+        if self.architectures:
+            if not any(a.value in current_tokens for a in self.architectures):
+                return False
+        if self.hardware:
+            if not any(h.value in current_tokens for h in self.hardware):
+                return False
+        if self.flavors:
+            if not self.flavors.issubset(current_tokens):
+                return False
+        if self.version_specifier:
+            if not current_version or not version_order:
+                return False
+            specs = self.version_specifier if isinstance(self.version_specifier, tuple) else (self.version_specifier,)
+            for spec in specs:
+                if not spec.matches(current_version, version_order):
+                    return False
+        return True
+
+    def to_tokens(self) -> Set[str]:
+        tokens = set()
+        for p in self.platforms:
+            tokens.add(p.value)
+        for b in self.build_types:
+            tokens.add(b.value)
+        for a in self.architectures:
+            tokens.add(a.value)
+        for h in self.hardware:
+            tokens.add(h.value)
+        tokens.update(self.flavors)
+        return tokens
+
+
+def matches_platform(token):
+    token_lower = token.lower()
+    if token_lower in PLATFORM_TOKENS:
+        return True
+    if _PLATFORM_SHAPE_RE.match(token_lower):
+        return True
+    return False
+
+
+def get_token_category(token, version_tokens=None):
+    token_lower = token.lower()
+
+    if token_lower in PLATFORM_TOKENS:
+        return ConfigurationCategory.PLATFORM
+    if _PLATFORM_SHAPE_RE.match(token_lower):
+        return ConfigurationCategory.PLATFORM
+    if token_lower in STYLE_TOKENS:
+        return ConfigurationCategory.STYLE
+    if token_lower in HARDWARE_TOKENS:
+        return ConfigurationCategory.HARDWARE
+    if token_lower in ARCHITECTURE_TOKENS:
+        return ConfigurationCategory.ARCHITECTURE
+
+    if token.endswith('+') or token.endswith('-') or ('-' in token and not token.endswith('-')):
+        return ConfigurationCategory.VERSION
+
+    if version_tokens and token_lower in version_tokens:
+        return ConfigurationCategory.VERSION
+
+    if token_lower.isidentifier() or re.match(r'^[a-z][a-z0-9_-]*$', token_lower):
+        return ConfigurationCategory.FLAVOR
+
+    return None
--- a/Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectation_file.py
+++ b/Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectation_file.py
@@ -1,0 +1,74 @@
+# Copyright (C) 2024-2026 Apple Inc. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1.  Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2.  Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+# ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""Line-based expectation file format (TestExpectations).
+
+This is the primary format used by WebKit layout tests and API tests.
+Each non-blank, non-comment line describes one expectation.
+"""
+
+from typing import List
+
+from webkitexpectationspy.expectations import Expectation
+from webkitexpectationspy.parser import ExpectationParser, ParseWarning
+
+
+class ExpectationFile:
+    def __init__(self, filename, raw_content, parsed_expectations, parse_warnings):
+        self._filename = filename
+        self._raw_content = raw_content
+        self._expectations = parsed_expectations
+        self._warnings = parse_warnings
+
+    @classmethod
+    def parse(cls, filename, content, suite=None):
+        parser = ExpectationParser(suite=suite)
+        results = parser.parse(filename, content)
+
+        expectations = []
+        warnings = []
+        for exp, line_warnings in results:
+            warnings.extend(line_warnings)
+            if exp is not None:
+                expectations.append(exp)
+
+        return cls(filename, content, expectations, warnings)
+
+    def serialize(self) -> str:
+        lines = []
+        for exp in self._expectations:
+            lines.append(repr(exp))
+        return '\n'.join(lines)
+
+    def expectations(self) -> List[Expectation]:
+        return self._expectations
+
+    def warnings(self) -> List[ParseWarning]:
+        return self._warnings
+
+    @property
+    def filename(self) -> str:
+        return self._filename
+
+    @property
+    def raw_content(self) -> str:
+        return self._raw_content
--- a/Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
+++ b/Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
@@ -1,0 +1,164 @@
+# Copyright (C) 2024-2026 Apple Inc. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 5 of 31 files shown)


Tools/Scripts/libraries/webkitexpectationspy/run-tests
```diff
@@ -0,0 +1,24 @@
...
+
+# Copyright (C) 2024-2025 Apple Inc. All rights reserved.
+# See LICENSE file for license terms.
+
+import os
+import sys
+import unittest
+
+# Add parent directory to path for development
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
```
JonWBedard: `webkitcorepy` has a Python test runner that we use, let's pick that up here.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/version_specifier.py
```diff
@@ -0,0 +1,132 @@
+# Copyright (C) 2024-2025 Apple Inc. All rights reserved.
```
JonWBedard: Check all the Copyrights here, should be 2026.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/parser.py
```diff
@@ -0,0 +1,289 @@
...
+
+    def __repr__(self):
+        return 'ParseWarning({!r}, {}, {!r})'.format(
+            self.filename, self.line_number, self.error)
+
+
+class ExpectationParser:
+    """Parser for test expectation files with plugin support."""
+
+    WEBKIT_BUG_PREFIX = 'webkit.org/b/'
```
JonWBedard, line 59: If we pull in `waebkitbugspy`, we could use any arbitrary bug tracker.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/filesystem.py
```diff
@@ -0,0 +1,73 @@
...
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""Filesystem abstraction for test expectations."""
+
+import os
+from typing import Protocol, runtime_checkable
+
+
+@runtime_checkable
+class FilesystemProtocol(Protocol):
```
JonWBedard: This is a bit of a surprising abstraction to me...do we actually need it? I know `webkitpy` loves it's `port`/`host` abstraction, but we've moved away from that.

Smackteo: Alternative would be using PathTestCase with real files, but that's slower and more complex for testing pure parsing logic

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/mocks/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (C) 2024-2026 Apple Inc. All rights reserved.
+# See LICENSE file for license terms.
+
+"""Mock objects for testing."""
```
gsnedders: Why would we need any mocks for this?

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/plugins/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""Format plugins for webkitexpectationspy."""
+
+from webkitexpectationspy.plugins.base import FormatPlugin
+from webkitexpectationspy.plugins.api_tests import APITestPlugin
+from webkitexpectationspy.plugins.layout_tests import LayoutTestPlugin
+
+__all__ = ['FormatPlugin', 'APITestPlugin', 'LayoutTestPlugin']
```
gsnedders: I know a lot of our existing code does this, but I'd really like to get away from importing everything upfront from `__init__.py`, because it has a real significant performance cost for code we don't use.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+CRASH = 2
+TIMEOUT = 3
+SKIP = 4
+SLOW = 5
+
+# Additional expectations for layout tests
+IMAGE_ONLY_FAILURE = 6
+TEXT = 7
+AUDIO = 8
+LEAK = 9
```
gsnedders: These should be an `enum.Enum`.

That said, we almost certainly should actually split results and modifiers apart, in a way that the existing TestExpectations code does not — and modifiers (`SKIP`, `SLOW`… though layout tests also support `REBASELINE`, `WONTFIX`, and `DUMPJSCONSOLELOGINSTDERR`) should probably be an `enum.Flag` (though we'll want to use `auto()` as their value so they combine correctly).

At a broad level, I'd suggest starting off by making sure we have a good data model to represent results (and this is a kinda tricky problem to start with!), and then build the representation of expectations on top of that.

Arguably, results should be `enum.Flag` too so if you have multiple results for a single test you could return `Result.PASS | Result.FAIL`.

(Also: Where do we actually treat them as integers? The results database? Anywhere else?)

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+
+# Expectation constants (values 0-3 are commonly used by test runners)
+PASS = 0
+FAIL = 1
+CRASH = 2
+TIMEOUT = 3
+SKIP = 4
+SLOW = 5
+
+# Additional expectations for layout tests
```
gsnedders: Why aren't these in the plugin? Why are these here? Why are they leaking into an arguably abstract model?

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+
+# String to constant mappings (base set)
+BASE_EXPECTATION_MAP = {
+    'pass': PASS,
+    'fail': FAIL,
+    'failure': FAIL,
+    'crash': CRASH,
+    'timeout': TIMEOUT,
+    'skip': SKIP,
+}
```
gsnedders: With an `Enum`, I'd be tempted to just add `_missing_` very similar to that in the docs: https://docs.python.org/3/library/enum.html#enum.Enum._missing_

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+    FAIL: 'Fail',
+    CRASH: 'Crash',
+    TIMEOUT: 'Timeout',
+    SKIP: 'Skip',
+    SLOW: 'Slow',
+    IMAGE_ONLY_FAILURE: 'ImageOnlyFailure',
+    TEXT: 'Text',
+    AUDIO: 'Audio',
+    LEAK: 'Leak',
+}
```
gsnedders: With an `Enum`, I'd be tempted to do something like:

```python
def __str__(self):
    if self.name == "IMAGE_ONLY_FAILURE":
        return "ImageOnlyFailure"
    return self.name.title()
```

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+class Expectation:
+    """Represents expectations for a single test or pattern."""
+
+    def __init__(self, test_pattern, expectations, configurations=None,
+                 slow_timeout=None, bug_ids=None, filename=None, line_number=None,
+                 version_specifiers=None):
```
gsnedders: Plz add type signatures here.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+from webkitexpectationspy.manager import ExpectationsManager
+from webkitexpectationspy.filesystem import FilesystemProtocol, DefaultFilesystem
+
+# Plugin base class
+from webkitexpectationspy.plugins.base import FormatPlugin
+
+# Built-in plugins
+from webkitexpectationspy.plugins.api_tests import APITestPlugin
+from webkitexpectationspy.plugins.layout_tests import LayoutTestPlugin
+
```
gsnedders: Ditto, let's not import everything and have to pay the cost of everything for every import of this library.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/configuration.py
```diff
@@ -0,0 +1,106 @@
...
+
+    Tokens in expectation files should appear in this order within
+    configuration blocks.
+    """
+    PLATFORM = 1      # mac, ios, linux, win, gtk, wpe
+    VERSION = 2       # Sonoma, iOS17, with +/-/range modifiers
+    STYLE = 3         # debug, release, asan, guardmalloc
+    HARDWARE = 4      # simulator, device, iphone, ipad
+    ARCHITECTURE = 5  # arm64, x86_64, x86, arm64_32, armv7k
+    FLAVOR = 6        # freeform: wk1, wk2, siteisolation, etc.
```
gsnedders: Why is this an IntEnum? These ints have no semantic value as far as I can tell (the ordering matters, but enums have order anyway, and we don't need `ConfigurationCategory.PLATFORM + 1` to work).

I would've expected we would have a `Configuration` dataclass, with these as properties in order, and this would then make a lot of the below expressible within the type system.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/filesystem.py
gsnedders: We should either use real files or `pyfakefs`.

Let's really, really, really not have our own abstraction around the filesystem.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+        self.bug_ids = tuple(bug_ids) if bug_ids else ()
+        self.filename = filename
+        self.line_number = line_number
+        self._is_wildcard = test_pattern.endswith('*')
+        self._prefix = test_pattern[:-1] if self._is_wildcard else None
+
+    @property
+    def is_wildcard(self):
+        """Returns True if this expectation uses a wildcard pattern."""
+        return self._is_wildcard
```
gsnedders: Do we actually care about this? (And at least historically Alexey was opposed: https://bugs.webkit.org/show_bug.cgi?id=258399)

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def is_skip(self):
+        """Returns True if test should be skipped."""
+        return SKIP in self.expectations
```
gsnedders:
```suggestion
    @property
    def skip(self):
        """Returns True if test should be skipped."""
        return SKIP in self.expectations
```

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def is_slow(self):
+        """Returns True if test is marked slow."""
+        return self.slow_timeout is not None or SLOW in self.expectations
```
gsnedders: Same as `skip` above, but also: in our abstract representation, we shouldn't have multiple ways to mark something as slow.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+        if self.slow_timeout is not None:
+            return self.slow_timeout
+        if SLOW in self.expectations:
+            return default_timeout * 5
+        return default_timeout
+
+    def is_flaky(self):
+        """Returns True if test can have multiple outcomes."""
+        actual_expectations = self.expectations - {SKIP, SLOW}
+        return len(actual_expectations) > 1
```
gsnedders: Again, this should probably be a property called `flaky`, and if you split results and modifiers up this becomes much cleaner.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/configuration.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+# Static tokens by category (alphabetically ordered within each)
+PLATFORM_TOKENS = frozenset({'gtk', 'ios', 'linux', 'mac', 'win', 'wpe'})
+STYLE_TOKENS = frozenset({'asan', 'debug', 'guardmalloc', 'release'})
+HARDWARE_TOKENS = frozenset({'device', 'ipad', 'iphone', 'simulator'})
+ARCHITECTURE_TOKENS = frozenset({'arm64', 'arm64_32', 'armv7k', 'x86', 'x86_64'})
```
gsnedders: I'm not exactly keen on duplicating these.

And while the source code has them ordered, they're not actually (because sets have no order).

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+    SKIP: 'Skip',
+    SLOW: 'Slow',
+    IMAGE_ONLY_FAILURE: 'ImageOnlyFailure',
+    TEXT: 'Text',
+    AUDIO: 'Audio',
+    LEAK: 'Leak',
+}
+
+
+class Expectation:
```
gsnedders, line 69: This should probably be a dataclass?

gsnedders, line 69: Having only an object called `Expectation` in `expectations.py` seems a little odd — there's nothing really plural here.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+            default_timeout: The default timeout to use if not slow
+
+        Returns:
+            The timeout in seconds.
+        """
+        if self.slow_timeout is not None:
+            return self.slow_timeout
+        if SLOW in self.expectations:
+            return default_timeout * 5
+        return default_timeout
```
gsnedders: This feels pretty weird, though maybe is just modelling what we already have. Using either a specific value or a multiplier just seems… odd.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+
+        # Check version specifiers
+        if self.version_specifiers:
+            if not current_version or not version_order:
+                return False
+            for spec in self.version_specifiers:
+                if not spec.matches(current_version, version_order):
+                    return False
+
+        return True
```
gsnedders: The fact we have all these arguments here strongly suggests to me that this is in the wrong place? We probably want a `Configuration` object to be able to tell us whether a given configuration matches it.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+    def __eq__(self, other):
+        if not isinstance(other, Expectation):
+            return False
+        return (self.test_pattern == other.test_pattern and
+                self.expectations == other.expectations and
+                self.configurations == other.configurations and
+                self.version_specifiers == other.version_specifiers)
+
+    def __hash__(self):
+        return hash((self.test_pattern, self.expectations, self.configurations))
```
gsnedders: All of these further suggest we should be using a frozen dataclass, fwiw.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
```diff
@@ -0,0 +1,194 @@
...
+        """Check if an actual result matches expectations.
+
+        Args:
+            actual_status: The actual test result constant
+
+        Returns:
+            True if the result was expected.
+        """
+        if self.is_skip() and actual_status == SKIP:
+            return True
```
gsnedders: This is redundant — the check below already covers this.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/linter.py
```diff
@@ -0,0 +1,367 @@
...
+        self.fixable = fixable
+        self.fix_description = fix_description
+
+    def __str__(self):
+        prefix = '[fixable] ' if self.fixable else ''
+        return '{}:{}: {}{}'.format(self.filename, self.line_number, prefix, self.message)
+
+    def __repr__(self):
+        return 'LintWarning({!r}, {}, {!r})'.format(
+            self.filename, self.line_number, self.message)
```
gsnedders, line 61: This can easily be a dataclass.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/linter.py
```diff
@@ -0,0 +1,367 @@
...
+
+
+class LineFix:
+    """Represents a fix to apply to a specific line."""
+
+    def __init__(self, line_number, new_content=None, delete=False):
+        self.line_number = line_number
+        self.new_content = new_content
+        self.delete = delete
```
gsnedders: This can easily be a dataclass.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/linter.py
```diff
@@ -0,0 +1,367 @@
...
+
+
+class ExpectationsLinter:
```
gsnedders: This feels a bit big. I'd kinda suggest splitting it up so that each rule/fixer is its own class.

gsnedders (changes requested):
Thanks for working on this! Having a unified expectations library is a great goal.

I have some concerns about the data model that I think are worth discussing before we land this — fundamentally this library is all about how we represent expectations, and it's important that we get the right data model, given changing data models can be painful.

Both webkitpy currently and what's currently in this PR conflate three semantically distinct concepts:

1. Result statuses: what actually happened (PASS, FAIL, CRASH, TIMEOUT, SKIP, IMAGE, TEXT...)
2. Expected statuses: what outcomes are acceptable (a (non-strict) subset of results; currently SKIP isn't allowed here)
3. Execution modifiers: directives about how to run tests (SKIP, SLOW, WONTFIX, REBASELINE)

Currently these all live in `expectations: frozenset` with integer constants. SKIP is particularly interesting since it's both a result status and a modifier, but the conflation shows up elsewhere too — SLOW and REBASELINE are modifiers, not expected outcomes.

I'd suggest we think through:
- Separate types for results vs modifiers (per-suite, since layout tests have IMAGE/TEXT/AUDIO that API tests don't)
- How this integrates with resultsdb's severity-ordered status values
- Whether enum.Flag vs integer constants vs something else is the right representation

As a minor naming thing: I'd lean toward "suites" rather than "plugins" for the test-suite-specific pieces — they're not really plugging into anything, they're just different test suites with different status/modifier sets, and this doesn't have any ability to provide something novel.

We should have some discussion about this on a call soon, and figure out what we want to do. I can maybe share some more concrete idea of what I think we should be doing before then.

Beyond the above, I've left a bunch of comments below: I haven't reviewed everything (far from it, honestly), but they hopefully give a flavour of how I'm thinking about this.

And finally, we should think about how we're going to migrate webkitpy to use this code — the API surface needs to be such that we can migrate webkitpy's existing code to this.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/suites/api_tests.py
JonWBedard: This variable shouldn't be coming from outside this file. The design of storing suite expectations inside the library is a bit questionable in the first place. If we're going to stick with in, we should have a single "api_tests" and "layout_tests" file, and the rest of the library should be generic.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/suites/base.py
JonWBedard: You should be able to have a complete generic version. What you have defined for API tests is probably pretty close to the generic version (pass, fail, crash, timeout).

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/suites/layout_tests.py
JonWBedard: Why are these functions not part of a class and specific to layout tests?

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/tests/__init__.py
JonWBedard: Annoyingly, we need the full license in every file.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/__init__.py
JonWBedard: Since this is a new library, do we actually need anything that's backward-compatible? Certainly "no" for API tests, and I think we'd prefer a full re-write of existing layout tests. Even if we don't go the full re-write route, backward-compatible aliases sounds like something that should be included in a future change dedicated to layout tests, if at all.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/configuration.py
JonWBedard: These type of comments aren't useful and should be removed.

JonWBedard, line 44: Do we _need_ to define the set of platforms? This is missing watchOS, tvOS and visionOS. But also, forcing a platform definition here seems bound to cause future pain.

JonWBedard, line 61: Need to include "Production" here as well, which is subtly different than "Release"

JonWBedard, line 72: Lots more here. `vision`, `watch`, `tv`, for example.

JonWBedard: Lots of these functions aren't served well by docstrings, on re-upload, let's audit these.

JonWBedard, line 129: I wonder if "platform", in particular, should be served by a `*os` regex-token, along with specifying the platforms that aren't explicit.

JonWBedard: How is the `version_specifier` determined? Is that OS name? version number? Both?

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectation_file.py
JonWBedard: Need better variable names here.

JonWBedard: I don't understand why we have a generic `ExpectationFile` class.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
JonWBedard: This module should be very de-coupled from the results database. If anything, this module would be _used_ by the results database in an ideal world.

JonWBedard: Instead of typing this to the resultsdatabase, could you use define these as the status values in the `enums`? The whole reason for this was sorting inside the results database itself, the specific values really should never be visible to clients.

JonWBedard: You either need the order defined or you have the numerical enums, you don't need both.

JonWBedard: Is "rebaseline" really a valid expectation? I don't understand how it could be.

JonWBedard: Many comments here are not helpful, should clean them up.

JonWBedard: Backward compatible with what, exactly?

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
JonWBedard: I don't think we should support wildcards in V1. It's an idea we should litigate in the future, but for now, it produce extra complexity without any benefit.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/expectations.py
JonWBedard: We should probably define "greater than" and "less than" as well. "greater than" would mean "this expectation includes everything this other expectation includes, and more", while "less than" would be the inverse.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/json_format.py
JonWBedard: Let's drop this if it's not fully implemented.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/resultsdb.py
JonWBedard, line 37: I saw this repeated in a few other places. Re-using these integer codes is fine, but remember what they're actually useful for: sorting! We shouldn't be re-doing sorting if we have this code in place.

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/version_specifier.py
JonWBedard, line 44: Does this also support version numbers?

I'm increasingly of the belief that version numbers are probably more useful, and easier to parse. We could have a "version name to number" conversion map, or better yet, callers could just hand these classes one.

Smackteo: Yes both forms work. VersionSpecifier.parse() accepts tokens like Sonoma+ or 14.0+

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/suites/api_tests.py
```diff
@@ -0,0 +1,82 @@
...
+
+class APITestSuite(TestSuiteFormat):
+    """Suite format for API test expectations."""
+
+    @property
+    def name(self) -> str:
+        return 'api-tests'
+
+    @property
+    def expectation_map(self) -> Dict:
```
JonWBedard: Do these even differ from the base right now?

Tools/Scripts/libraries/webkitexpectationspy/run-tests
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
```
JonWBedard, line 1: Can we add a `lint` command that accepts a file?

Tools/Scripts/libraries/webkitexpectationspy/webkitexpectationspy/version_specifier.py
```diff
@@ -0,0 +1,169 @@
...
+
+class VersionSpecifier:
+    """Represents a version constraint with optional range/modifier.
+
+    Supports:
+        - EXACT: 'Sonoma' or '14.0' (this version only)
+        - AND_LATER: 'Sonoma+' or '14.0+' (this and later versions)
+        - AND_EARLIER: 'Sonoma-' or '14.0-' (this and earlier versions)
+        - RANGE: 'Ventura-Sequoia' or '13.0-15.0' (inclusive range)
+    """
```
JonWBedard, line 52: You should have `Claude` go through all these docstrings and remove ones which can be reconstructed with a trivial reading of the code.

JonWBedard (approved):

PR #57646, opened 2026-02-01
[Skia] System fallback font selection shouldn't take style into account

fujii:
#### ce075d61e972d89d1dc3b22ae60164276d6c0a8f
<pre>
[Skia] System fallback font selection shouldn&apos;t take style into account
<a href="https://bugs.webkit.org/show_bug.cgi?id=295962">https://bugs.webkit.org/show_bug.cgi?id=295962</a>

WebKit has WebCore::SystemFallbackFontCache class which caches system fallback
fonts. The cache key is a tuple of a string, locale and emoji policy. But, the
weight, width and slant are&apos;t involved into the cache key. However, Skia
backend looked for a system fallback font with the font style. This resulted in
mixed fonts for a text.

* LayoutTests/platform/glib/fast/forms/visual-hebrew-text-field-expected.txt:
* LayoutTests/platform/glib/fast/text/in-rendered-text-rtl-expected.txt:
* LayoutTests/platform/glib/fast/text/international/bidi-LDB-2-CSS-expected.txt:
* LayoutTests/platform/glib/fast/text/international/bidi-LDB-2-HTML-expected.txt:
* LayoutTests/platform/glib/fast/text/international/bidi-LDB-2-formatting-characters-expected.txt:
* LayoutTests/platform/glib/fast/text/international/bidi-european-terminators-expected.txt:
* LayoutTests/platform/glib/fast/text/international/bidi-ignored-for-first-child-inline-expected.txt:
* LayoutTests/platform/glib/fast/text/international/rtl-caret-expected.txt:
* LayoutTests/platform/glib/fast/text/midword-break-after-breakable-char-expected.txt:
* LayoutTests/platform/gtk/fast/forms/select-writing-direction-natural-expected.txt:
* LayoutTests/platform/wpe/fast/forms/select-writing-direction-natural-expected.txt:
* Source/WebCore/platform/graphics/skia/FontCacheSkia.cpp:
(WebCore::FontCache::systemFallbackForCharacterCluster):
* Source/WebCore/platform/graphics/skia/SkiaSystemFallbackFontCache.cpp:
(WebCore::FontSetCache::FontSetCache):
(WebCore::FontSetCache::create):
(WebCore::FontSetCache::fontForCharacterCluster):
(WebCore::FontSetCache::FontSet::create):
(WebCore::SkiaSystemFallbackFontCache::fontForCharacterCluster):
(WebCore::fontconfigStyle):
(WebCore::FontSetCacheKey::FontSetCacheKey): Deleted.
(WebCore::FontSetCacheKey::isHashTableDeletedValue const): Deleted.
(WebCore::add): Deleted.
(WebCore::FontSetCacheKeyHash::hash): Deleted.
(WebCore::FontSetCacheKeyHash::equal): Deleted.
* Source/WebCore/platform/graphics/skia/SkiaSystemFallbackFontCache.h:

Canonical link: <a href="https://commits.webkit.org/307565@main">https://commits.webkit.org/307565@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/a06afc71981875e433d9313e521b64e591cfe2f4

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/144519 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/17198 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/8760 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/153189 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/98154 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/146394 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/17680 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/17092 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/111130 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/79747 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/147482 "Passed tests") | [⏳ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/iOS-26-Simulator-WK2-Tests-EWS "Waiting to run tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/129779 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/92041 "Passed tests") | 
| | [⏳ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/iOS-26-Simulator-WPT-WK2-Tests-EWS "Waiting to run tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/10657 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/635 "Built successfully") | 
| | [⏳ 🧪 api-ios](https://ews-build.webkit.org/#/builders/API-Tests-iOS-Simulator-EWS "Waiting to run tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/6469 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/155502 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/17050 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/7525 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/119133 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/17088 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/14273 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/119487 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/15308 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/127682 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/72591 "Built successfully") | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/22338 "Built successfully and passed tests") | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/16672 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/6085 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/16408 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/80451 "Built successfully") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/16617 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/16472 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/platform/graphics/skia/FontCacheSkia.cpp
+++ b/Source/WebCore/platform/graphics/skia/FontCacheSkia.cpp
@@ -138,9 +138,9 @@ RefPtr<Font> FontCache::systemFallbackForCharacterCluster(const FontDescription&
     Vector<const char*, 1> bcp47;
     if (isEmoji)
         bcp47.append("und-Zsye");
-    auto typeface = fontManager().matchFamilyStyleCharacter(nullptr, skiaFontStyle(description), bcp47.mutableSpan().data(), bcp47.size(), baseCharacter);
+    auto typeface = fontManager().matchFamilyStyleCharacter(nullptr, { }, bcp47.mutableSpan().data(), bcp47.size(), baseCharacter);
 #else
-    auto typeface = m_skiaSystemFallbackFontCache.fontForCharacterCluster(skiaFontStyle(description), isEmoji ? "und-Zsye"_s : description.computedLocale(), stringView);
+    auto typeface = m_skiaSystemFallbackFontCache.fontForCharacterCluster(isEmoji ? "und-Zsye"_s : description.computedLocale(), stringView);
 #endif
     if (!typeface)
         return nullptr;
--- a/Source/WebCore/platform/graphics/skia/SkiaSystemFallbackFontCache.cpp
+++ b/Source/WebCore/platform/graphics/skia/SkiaSystemFallbackFontCache.cpp
@@ -36,69 +36,6 @@
 
 namespace WebCore {
 
-static void fontconfigStyle(const SkFontStyle& style, int& weight, int& width, int& slant)
-{
-    struct MapRanges {
-        SkScalar oldValue;
-        SkScalar newValue;
-    };
-
-    IGNORE_CLANG_WARNINGS_BEGIN("unsafe-buffer-usage")
-    auto mapRanges = [](SkScalar value, MapRanges const ranges[], int rangesCount) -> SkScalar {
-        if (value < ranges[0].oldValue)
-            return ranges[0].newValue;
-
-        for (int i = 0; i < rangesCount - 1; ++i) {
-            if (value < ranges[i + 1].oldValue)
-                return ranges[i].newValue + ((value - ranges[i].oldValue) * (ranges[i + 1].newValue - ranges[i].newValue) / (ranges[i + 1].oldValue - ranges[i].oldValue));
-        }
-        return ranges[rangesCount - 1].newValue;
-    };
-    IGNORE_CLANG_WARNINGS_END
-
-    static constexpr MapRanges weightRanges[] = {
-        { SkFontStyle::kThin_Weight,       FC_WEIGHT_THIN },
-        { SkFontStyle::kExtraLight_Weight, FC_WEIGHT_EXTRALIGHT },
-        { SkFontStyle::kLight_Weight,      FC_WEIGHT_LIGHT },
-        { 350,                             FC_WEIGHT_DEMILIGHT },
-        { 380,                             FC_WEIGHT_BOOK },
-        { SkFontStyle::kNormal_Weight,     FC_WEIGHT_REGULAR },
-        { SkFontStyle::kMedium_Weight,     FC_WEIGHT_MEDIUM },
-        { SkFontStyle::kSemiBold_Weight,   FC_WEIGHT_DEMIBOLD },
-        { SkFontStyle::kBold_Weight,       FC_WEIGHT_BOLD },
-        { SkFontStyle::kExtraBold_Weight,  FC_WEIGHT_EXTRABOLD },
-        { SkFontStyle::kBlack_Weight,      FC_WEIGHT_BLACK },
-        { SkFontStyle::kExtraBlack_Weight, FC_WEIGHT_EXTRABLACK },
-    };
-    weight = mapRanges(style.weight(), weightRanges, std::size(weightRanges));
-
-    static constexpr MapRanges widthRanges[] = {
-        { SkFontStyle::kUltraCondensed_Width, FC_WIDTH_ULTRACONDENSED },
-        { SkFontStyle::kExtraCondensed_Width, FC_WIDTH_EXTRACONDENSED },
-        { SkFontStyle::kCondensed_Width,      FC_WIDTH_CONDENSED },
-        { SkFontStyle::kSemiCondensed_Width,  FC_WIDTH_SEMICONDENSED },
-        { SkFontStyle::kNormal_Width,         FC_WIDTH_NORMAL },
-        { SkFontStyle::kSemiExpanded_Width,   FC_WIDTH_SEMIEXPANDED },
-        { SkFontStyle::kExpanded_Width,       FC_WIDTH_EXPANDED },
-        { SkFontStyle::kExtraExpanded_Width,  FC_WIDTH_EXTRAEXPANDED },
-        { SkFontStyle::kUltraExpanded_Width,  FC_WIDTH_ULTRAEXPANDED },
-    };
-    width = mapRanges(style.width(), widthRanges, std::size(widthRanges));
-
-    slant = FC_SLANT_ROMAN;
-    switch (style.slant()) {
-    case SkFontStyle::kUpright_Slant:
-        slant = FC_SLANT_ROMAN;
-        break;
-    case SkFontStyle::kItalic_Slant:
-        slant = FC_SLANT_ITALIC;
-        break;
-    case SkFontStyle::kOblique_Slant:
-        slant = FC_SLANT_OBLIQUE;
-        break;
-    }
-}
-
 static String filePathFromPattern(FcPattern* pattern)
 {
     FcChar8* filename = nullptr;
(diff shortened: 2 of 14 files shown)


Source/WebCore/platform/graphics/skia/FontCacheSkia.cpp
```diff
@@ -140,7 +140,7 @@ RefPtr<Font> FontCache::systemFallbackForCharacterCluster(const FontDescription&
         bcp47.append("und-Zsye");
     auto typeface = fontManager().matchFamilyStyleCharacter(nullptr, skiaFontStyle(description), bcp47.mutableSpan().data(), bcp47.size(), baseCharacter);
 #else
-    auto typeface = m_skiaSystemFallbackFontCache.fontForCharacterCluster(skiaFontStyle(description), isEmoji ? "und-Zsye"_s : description.computedLocale(), stringView);
+    auto typeface = m_skiaSystemFallbackFontCache.fontForCharacterCluster(isEmoji ? "und-Zsye"_s : description.computedLocale(), stringView);
```
carlosgcampos, line 143: I think we are expected to match the style here, like when not using the cache.

fujii: What do you mean? What's your suggestion? Should I change WebCore::SystemFallbackFontCache class to include the font style to the cache key?

carlosgcampos: I don't know to be honest, but I think FontCache::systemFallbackForCharacterCluster() should behave the same way with and without the cache. The same cache in chromium uses the style for the key, but then they only apply the slant to the pattern, I don't know why.

aperezdc: Could it be so the default fallback system font can be used also for italic/slanted even if the font itself does not have an italic variant (by means of using synthetic slant/italics)?

fujii: I don't see any problems for italic and bold with my patch.

fujii: > I think FontCache::systemFallbackForCharacterCluster() should behave the same way with and without the cache.

I don't understand what does this mean?
`FontCache::systemFallbackForCharacterCluster()` has no way to bypass `m_skiaSystemFallbackFontCache.fontForCharacterCluster()`.

> The same cache in chromium uses the style for the key, but then they only apply the slant to the pattern, I don't know why.()

Which file?

carlosgcampos: > > I think FontCache::systemFallbackForCharacterCluster() should behave the same way with and without the cache.
> 
> I don't understand what does this mean? `FontCache::systemFallbackForCharacterCluster()` has no way to bypass `m_skiaSystemFallbackFontCache.fontForCharacterCluster()`.
> 

Same behavior as before we introduced the cache, which is what we still do for android and windows.

> > The same cache in chromium uses the style for the key, but then they only apply the slant to the pattern, I don't know why.()
> 
> Which file?

https://github.com/chromium/chromium/blob/ba081a425e1b6fd5cc0050cb45371a64d68b34f4/ui/gfx/font_fallback_linux.cc#L106

fujii: Got it. If we want to do same in WebKit, we have to add `bool isSlant;` for `CharacterFallbackMapKey` only for `USE(SKIA)`.
https://github.com/WebKit/WebKit/blob/21c8e1ba5450de52e65a221932ea5c611c9575e2/Source/WebCore/platform/graphics/SystemFallbackFontCache.h#L45-L50

In my opinion, we shouldn't add a platform `#if` without knowing an actual problem. I think we should take the same approach with Safari until we know an actual problem.

Source/WebCore/platform/graphics/skia/SkiaSystemFallbackFontCache.cpp
```diff
@@ -169,25 +69,19 @@ class FontSetCache {
...
+    const FontSetCache::Font* fontForCharacterCluster(const String& locale, FcCharSet* charSet)
     {
-        int weight, width, slant;
-        fontconfigStyle(style, weight, width, slant);
-
-        auto& fontSet = m_cache.ensure(FontSetCacheKey(weight, width, slant), [&] {
-            return FontSet::create(locale, weight, width, slant);
-        }).iterator->value;
-        if (!fontSet)
+        m_fontSet = FontSet::create(locale);
```
carlosgcampos: We are now always creating this, which makes this cache useless, no? Since the the FontSet is now always the same for the given locale, maybe we could pass the locale to the constructor and create the FontSet from there. We could also add a create() that creates the FontSet and passes it to the constructor returning nullptr when it fails, this way we never have a FontSetCache with a null FontSet.

fujii: Fixed. Thank you very much.

Source/WebCore/platform/graphics/skia/SkiaSystemFallbackFontCache.cpp
```diff
@@ -298,11 +197,11 @@ sk_sp<SkTypeface> SkiaSystemFallbackFontCache::fontForCharacterCluster(const SkF
...
     }
 
-    auto& fontSetCache = m_cache.ensure(locale.isNull() ? emptyString() : locale, [] {
-        return makeUnique<FontSetCache>();
+    auto& fontSetCache = m_cache.ensure(locale.isNull() ? emptyString() : locale, [&] {
+        return FontSetCache::create(locale);
     }).iterator->value;
 
-    const auto* font = fontSetCache->fontForCharacterCluster(style, locale, charSet);
+    const auto* font = fontSetCache->fontForCharacterCluster(charSet);
```
carlosgcampos, line 207: Now we should check fontSetCache is not nullptr, since FontSetCache::create(locale) can return nullptr

fujii: Good catch. Fixed. Thank you.

carlosgcampos (approved):

PR #57659, opened 2026-02-01
Use separate types for platform and DOM representations of encrypted media

weinig:
#### 17d6f3825a0b9a767e05a64b9fb0ed7dc2d7964d
<pre>
Use separate types for platform and DOM representations of encrypted media
<a href="https://bugs.webkit.org/show_bug.cgi?id=306729">https://bugs.webkit.org/show_bug.cgi?id=306729</a>

Reviewed by NOBODY (OOPS!).

As was recently done with MediaCapabilities, this creates distinct types
for the platform and DOM representations of encrypted media support with
toPlatform/fromPlatform functions to bridge between the two.

As with MediaCapabilities, this has a few benefits:
  - It allows the DOM versions to be next to their IDL files
  - It allows the platform versions to contain additional values
    not exposed by the DOM (important, because IDL dictionaries
    must match exactly now, allowing us to remove some more uses
    of `LegacyNativeDictionaryRequiredInterfaceNullability`).
  - Creates a clear layering distinction.

There are few remaining conceptual layering violations due to CDMPrivate.cpp
implementing algorithms defined by the encrypted media spec. Ideally, this
should be moved out of platform.

For this change, no renames were necessary, as the platform layer already
had a distinct prefix, CDM.

* Source/WebCore/Modules/encryptedmedia/CDM.cpp:
* Source/WebCore/Modules/encryptedmedia/CDM.h:
* Source/WebCore/Modules/encryptedmedia/MediaKeyEncryptionScheme.h:
* Source/WebCore/Modules/encryptedmedia/MediaKeyEncryptionScheme.idl:
* Source/WebCore/Modules/encryptedmedia/MediaKeyMessageEventInit.idl:
* Source/WebCore/Modules/encryptedmedia/MediaKeyMessageType.h:
* Source/WebCore/Modules/encryptedmedia/MediaKeyMessageType.idl:
* Source/WebCore/Modules/encryptedmedia/MediaKeySession.cpp:
* Source/WebCore/Modules/encryptedmedia/MediaKeySession.idl:
* Source/WebCore/Modules/encryptedmedia/MediaKeySessionType.h:
* Source/WebCore/Modules/encryptedmedia/MediaKeySessionType.idl:
* Source/WebCore/Modules/encryptedmedia/MediaKeyStatus.h:
* Source/WebCore/Modules/encryptedmedia/MediaKeyStatusMap.idl:
* Source/WebCore/Modules/encryptedmedia/MediaKeySystemAccess.cpp:
* Source/WebCore/Modules/encryptedmedia/MediaKeySystemAccess.idl:
* Source/WebCore/Modules/encryptedmedia/MediaKeySystemConfiguration.h:
* Source/WebCore/Modules/encryptedmedia/MediaKeySystemMediaCapability.h:
* Source/WebCore/Modules/encryptedmedia/MediaKeySystemMediaCapability.idl:
* Source/WebCore/Modules/encryptedmedia/MediaKeys.cpp:
* Source/WebCore/Modules/encryptedmedia/MediaKeys.h:
* Source/WebCore/Modules/encryptedmedia/MediaKeys.idl:
* Source/WebCore/Modules/encryptedmedia/MediaKeysRequirement.h:
* Source/WebCore/Modules/encryptedmedia/MediaKeysRequirement.idl:
* Source/WebCore/Modules/encryptedmedia/MediaKeysRestrictions.h: Removed.
* Source/WebCore/Modules/encryptedmedia/NavigatorEME.cpp:
* Source/WebCore/platform/encryptedmedia/CDMKeySystemConfiguration.h:
* Source/WebCore/platform/encryptedmedia/CDMLogging.cpp:
* Source/WebCore/platform/encryptedmedia/CDMMediaCapability.h:
* Source/WebCore/platform/encryptedmedia/CDMPrivate.cpp:
* Source/WebCore/platform/encryptedmedia/clearkey/CDMClearKey.cpp:
* Source/WebCore/platform/graphics/avfoundation/CDMFairPlayStreaming.cpp:
* Source/WebCore/platform/graphics/avfoundation/objc/CDMInstanceFairPlayStreamingAVFObjC.mm:
* Source/WebCore/platform/graphics/gstreamer/eme/CDMThunder.cpp:
* Source/WebCore/platform/graphics/gstreamer/eme/CDMThunder.h:
* Source/WebCore/testing/MockCDMFactory.cpp:
* Source/WebCore/testing/MockCDMFactory.h:
</pre>
https://github.com/WebKit/WebKit/commit/17d6f3825a0b9a767e05a64b9fb0ed7dc2d7964d

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/141811 "2 style errors") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/14197 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/3769 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/150402 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/94936 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/143678 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/14907 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/14356 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/108975 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/78809 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/144760 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/11525 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/126929 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/89871 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/11076 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/8717 "Passed tests") | [✅ 🛠 wpe-libwebrtc](https://ews-build.webkit.org/#/builders/172/builds/471 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/120384 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/2948 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/152793 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/13886 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/3411 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/117066 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/13901 "Built successfully") | [❌ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/12116 "Found 1 new test failure: imported/w3c/web-platform-tests/html/semantics/forms/the-input-element/range-restore-oninput-onchange-event.https.html (failure)") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/117388 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/13436 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/123635 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/69544 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/13924 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/2916 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/13663 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/77649 "Built successfully") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/13866 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/13710 "Built successfully") | | |


cdumez (approved):

jernoble:
I don't understand why having a WebCore enum _be_ a platform enum is a layering violation. The layering all goes downwards from WebCore to platform. What's more, this change introduces giant sections of new copy code to something that is, in effect, a static cast, introducing complexity for what appears to be extremely little value.

jernoble:
> It's not _(jer: a layering violation)_ (and I don't think I say it is, but if I do, that's incorrect).

Yes, you only said that separating the enums makes the layering clear, but you don't say it's a layering violation, except in the one case where a platform-level object is referring to a WebCore-level enum in CDMThunder. And it can do so without causing a compile time error because those two enums are just aliases of one another and you can freely cast between them. What I'm gathering from "making the layering more clear" is that you're trying to enforce the layering by making the using of a WebCore-level enum a compile error inside platform. I don't see the benefit, however, that splitting up then enum definitions would bring since platform/ code can still include the new enum's header. 

If more of platform/ were in PAL/ then it _would_ be a compile-time error for classes like CDMThunder to refer to headers inside WebCore because those headers wouldn't exist in the build directory yet. And in that case, that would _actually_ make the layering clear and enforceable at runtime, but still wouldn't require separate enum definitions to exist in both halves of the layering divide with `toPlatform/fromPlatform` to convert between them.

> By isolating them, we don't have to worry about how this will affect other layers.

Except we do, because someone will have to hand-edit the `toPlatform/fromPlatform` code to copy the enum values between the two types. It doesn't eliminate the worry, it just makes it more messy. Additionally, there's no concrete benefit to doing so in this patch; there's only a hypothetical future benefit in a future patch. So why is it critical to do this now, rather than in the future?

Even in the case where it makes sense to copy between the enum types, why were those casts implemented as giant switch statements? Maybe the compiler at a high enough level of optimization will be able to complete elide the switches, but that's not guaranteed. Surely it would be possible to use `EnumTraits` and constexpr checks to validate that values of each enum are the same (and that the enums are the same length) without doing those checks at runtime.

PR #57685, opened 2026-02-02
Web Inspector: Adding color contrast information within Color Picker

AnujPanta1:
#### 70fe1a50bf5af47e75e73943e95489c7d94c69c5
<pre>
Web Inspector: Adding color contrast information within Color Picker
<a href="https://bugs.webkit.org/show_bug.cgi?id=260101">https://bugs.webkit.org/show_bug.cgi?id=260101</a>
<a href="https://rdar.apple.com/113887185">rdar://113887185</a>

Reviewed by Devin Rousso (OOPS\!).

Add contrast ratio threshold lines to the color picker overlay when editing text color
properties. Two lines indicate AA and AAA WCAG compliance thresholds, helping developers
pick accessible colors without leaving the picker.

All contrast calculations happen in the frontend using the WCAG 2.0 relative luminance
formula (<a href="https://www.w3.org/TR/WCAG20/#relativeluminancedef)">https://www.w3.org/TR/WCAG20/#relativeluminancedef)</a> and contrast ratio formula
(<a href="https://www.w3.org/TR/WCAG20/#contrast-ratiodef).">https://www.w3.org/TR/WCAG20/#contrast-ratiodef).</a> This avoids backend round-trips and
provides instant feedback as users drag within the color square. The implementation uses
binary search to find brightness values that achieve target luminance at each saturation
level, rendering results as SVG polylines.

Semi-transparent foreground colors are blended over the background before calculating
contrast, ensuring accurate ratios for colors with alpha &lt; 1.

WCAG defines relaxed thresholds for &quot;large text&quot; (≥18pt or ≥14pt bold). Normal text requires
4.5:1 for AA and 7:1 for AAA. Large text only requires 3:1 for AA and 4.5:1 for AAA. The
implementation reads computed font-size and font-weight to detect large text and adjusts
the threshold lines accordingly.

A contrast info section below the picker displays the current ratio and
compliance badge (AAA/AA/Fail) alongside the background color swatch.

* Source/WebInspectorUI/UserInterface/Models/Color.js:
(WI.Color.prototype.contrastComplianceForRatio):
(WI.Color.prototype.relativeLuminance):
(WI.Color.prototype.contrastRatio):
(WI.Color.prototype.contrastCompliance):
(WI.Color.prototype.blendOverBackground):
(WI.Color):
* Source/WebInspectorUI/UserInterface/Views/ColorPicker.css:
(.color-picker &gt; .contrast-info + .variable-color-swatches):
(.color-picker &gt; .variable-color-swatches &gt; h2):
(.color-picker &gt; .contrast-info):
(.color-picker &gt; .contrast-info &gt; .contrast-label):
(.color-picker &gt; .contrast-info &gt; .contrast-ratio):
(.color-picker &gt; .contrast-info &gt; .compliance-badge):
(.color-picker &gt; .contrast-info &gt; .compliance-badge.contrast-aaa):
(@media (prefers-color-scheme: dark) .color-picker &gt; .contrast-info &gt; .compliance-badge.contrast-aaa):
(.color-picker &gt; .contrast-info &gt; .compliance-badge.contrast-aa):
(@media (prefers-color-scheme: dark) .color-picker &gt; .contrast-info &gt; .compliance-badge.contrast-aa):
(.color-picker &gt; .contrast-info &gt; .compliance-badge.contrast-fail):
(@media (prefers-color-scheme: dark) .color-picker &gt; .contrast-info &gt; .compliance-badge.contrast-fail):
(.color-picker &gt; .contrast-info &gt; .contrast-separator):
(.color-picker &gt; .contrast-info &gt; .inline-swatch):
* Source/WebInspectorUI/UserInterface/Views/ColorPicker.js:
(WI.ColorPicker.prototype.async colorInputsWrapperElement):
(WI.ColorPicker.prototype.set opacity):
(WI.ColorPicker.prototype.set color):
(WI.ColorPicker.prototype._updateColor):
(WI.ColorPicker.prototype._createContrastInfoSection):
(WI.ColorPicker.prototype._updateContrastInfo):
(WI.ColorPicker):
* Source/WebInspectorUI/UserInterface/Views/ColorSquare.css:
(.color-square &gt; .contrast-lines-svg):
(.color-square &gt; .contrast-lines-svg &gt; .contrast-line):
(.color-square &gt; .contrast-lines-svg &gt; .contrast-line.contrast-aa-threshold):
(.color-square &gt; .contrast-lines-svg &gt; .contrast-line.contrast-aaa-threshold):
(.color-square &gt; .contrast-label):
(.color-square &gt; .contrast-label.contrast-aa-label):
(.color-square &gt; .contrast-label.contrast-aaa-label):
(@media (prefers-color-scheme: dark) .color-square &gt; .contrast-label):
* Source/WebInspectorUI/UserInterface/Views/ColorSquare.js:
(WI.ColorSquare):
(WI.ColorSquare.prototype.get contrastBackgroundColor):
(WI.ColorSquare.prototype.set contrastBackgroundColor):
(WI.ColorSquare.prototype.get opacity):
(WI.ColorSquare.prototype.set opacity):
(WI.ColorSquare.prototype.get isLargeText):
(WI.ColorSquare.prototype.set isLargeText):
(WI.ColorSquare.prototype._updateBaseColor):
(WI.ColorSquare.prototype._drawSRGBOutline):
(WI.ColorSquare.prototype._drawContrastLines):
(WI.ColorSquare.prototype._calculateContrastLinePoints):
(WI.ColorSquare.prototype._findBrightnessForLuminance):
(WI.ColorSquare.prototype._updatePolylinePoints):
(WI.ColorSquare.prototype._updateContrastLabel):
* Source/WebInspectorUI/UserInterface/Views/InlineSwatch.js:
* Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js:
(WI.SpreadsheetStyleProperty.prototype.inlineSwatchGetContrastInfo):

Canonical link: <a href="https://commits.webkit.org/308318@main">https://commits.webkit.org/308318@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/7fbf3c41e18259886e18acb083c8db745decffe8

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/147164 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/19845 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/13435 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/155846 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/100578 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [❌ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/f27667b2-c5fa-4002-a699-2c0da325ee9a) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/20303 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/19746 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/113410 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/100578 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [❌ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/f1c1c46e-cfff-4709-a191-9a7f496bbc8a) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/150126 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/15628 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/132193 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94171 "Passed tests") | | [❌ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b8fddd38-b306-475b-8237-1b4df7a4414b) 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/14820 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/12605 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/3288 "Built successfully") | | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/124410 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/10127 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/158177 "Built successfully") | | 
| | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/1308 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/11567 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/121437 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/19646 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/16475 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/121640 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/31148 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/19655 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/131884 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/75614 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/17176 "Passed tests") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/8682 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/19262 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/83016 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/18992 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/19142 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/19050 "Built successfully") | | | |

### DIFF
--- a/Source/WebInspectorUI/UserInterface/Models/Color.js
+++ b/Source/WebInspectorUI/UserInterface/Models/Color.js
@@ -536,6 +536,19 @@ WI.Color = class Color
         return Number.constrain(Math.round(value), 0, 255);
     }
 
+    static contrastComplianceForRatio(ratio, {isLargeText} = {})
+    {
+        let aaaThreshold = isLargeText ? WI.Color.ContrastThreshold.AA : WI.Color.ContrastThreshold.AAA;
+        if (ratio >= aaaThreshold)
+            return WI.Color.ContrastCompliance.AAA;
+
+        let aaThreshold = isLargeText ? WI.Color.ContrastThreshold.AALargeText : WI.Color.ContrastThreshold.AA;
+        if (ratio >= aaThreshold)
+            return WI.Color.ContrastCompliance.AA;
+
+        return WI.Color.ContrastCompliance.Fail;
+    }
+
     // Public
 
     nextFormat(format)
@@ -919,6 +932,53 @@ WI.Color = class Color
             hex = "0" + hex;
         return hex;
     }
+
+    // https://www.w3.org/TR/WCAG20/#relativeluminancedef
+    relativeLuminance()
+    {
+        let [r, g, b] = WI.Color._toLinearLight(this.normalizedRGB);
+        return (0.2126729 * r) + (0.7151522 * g) + (0.0721750 * b);
+    }
+
+    contrastRatio(other)
+    {
+        console.assert(other instanceof WI.Color, other);
+
+        let l1 = this.relativeLuminance();
+        let l2 = other.relativeLuminance();
+
+        if (l1 < l2)
+            [l1, l2] = [l2, l1];
+
+        return (l1 + 0.05) / (l2 + 0.05);
+    }
+
+    contrastCompliance(other, {isLargeText} = {})
+    {
+        let ratio = this.contrastRatio(other);
+        return WI.Color.contrastComplianceForRatio(ratio, {isLargeText});
+    }
+
+    blendOverBackground(backgroundColor)
+    {
+        console.assert(backgroundColor instanceof WI.Color, backgroundColor);
+
+        if (this.alpha === 1)
+            return this.copy();
+
+        let foregroundRGB = this.normalizedRGB;
+        let backgroundRGB = backgroundColor.normalizedRGB;
+        let alpha = this.alpha;
+
+        let blendedRGB = [
+            (foregroundRGB[0] * alpha) + (backgroundRGB[0] * (1 - alpha)),
+            (foregroundRGB[1] * alpha) + (backgroundRGB[1] * (1 - alpha)),
+            (foregroundRGB[2] * alpha) + (backgroundRGB[2] * (1 - alpha)),
+        ];
+
+        return new WI.Color(WI.Color.Format.ColorFunction, blendedRGB, this.gamut);
+    }
+
 };
 
 WI.Color.Format = {
@@ -948,6 +1008,18 @@ WI.Color.FunctionNames = new Set([
     "color",
     "color-mix",
 ]);
+
+WI.Color.ContrastThreshold = {
+    AAA: 7,
+    AA: 4.5,
+    AALargeText: 3,
+};
+
+WI.Color.ContrastCompliance = {
+    AAA: "AAA",
+    AA: "AA",
+    Fail: "Fail",
+};
 
 WI.Color.Keywords = {
     "aliceblue": [240, 248, 255, 1],
--- a/Source/WebInspectorUI/UserInterface/Views/ColorPicker.css
+++ b/Source/WebInspectorUI/UserInterface/Views/ColorPicker.css
@@ -113,6 +113,12 @@
     padding: 8px 0 0;
 }
 
+.color-picker > .contrast-info + .variable-color-swatches {
+    margin-top: 8px;
+    padding-top: 8px;
+    border-top: 1px solid var(--border-color);
+}
+
 .color-picker > .variable-color-swatches > ul{
     list-style: none;
     display: flex;
@@ -137,4 +143,72 @@
 .color-picker > .variable-color-swatches > h2 {
     margin: 0;
     font-size: 1.15em;
+}
+
+.color-picker > .contrast-info {
+    display: flex;
+    align-items: center;
+    gap: 6px;
+    padding: 8px 0 0;
+    margin-top: 8px;
+    border-top: 1px solid var(--border-color);
+    font-size: 11px;
+}
+
+.color-picker > .contrast-info > .contrast-label {
+    color: var(--text-color-secondary);
+}
+
+.color-picker > .contrast-info > .contrast-ratio {
+    font-variant-numeric: tabular-nums;
+}
+
+.color-picker > .contrast-info > .compliance-badge {
+    padding: 2px 6px;
+    border-radius: 3px;
+    font-size: 10px;
+    font-weight: bold;
+    text-transform: uppercase;
+}
+
+.color-picker > .contrast-info > .compliance-badge.contrast-aaa {
+    background-color: hsl(120, 50%, 35%);
+    color: white;
+}
+
+@media (prefers-color-scheme: dark) {
+    .color-picker > .contrast-info > .compliance-badge.contrast-aaa {
+        background-color: hsl(120, 50%, 45%);
+    }
+}
+
+.color-picker > .contrast-info > .compliance-badge.contrast-aa {
+    background-color: hsl(90, 50%, 40%);
+    color: white;
+}
+
+@media (prefers-color-scheme: dark) {
+    .color-picker > .contrast-info > .compliance-badge.contrast-aa {
+        background-color: hsl(90, 50%, 50%);
+    }
+}
+
+.color-picker > .contrast-info > .compliance-badge.contrast-fail {
+    background-color: hsl(0, 60%, 50%);
+    color: white;
+}
+
+@media (prefers-color-scheme: dark) {
+    .color-picker > .contrast-info > .compliance-badge.contrast-fail {
+        background-color: hsl(0, 60%, 60%);
+    }
+}
+
+.color-picker > .contrast-info > .contrast-separator {
+    color: var(--text-color-tertiary);
+    margin: 0 2px;
+}
+
+.color-picker > .contrast-info > .inline-swatch {
+    --inline-swatch-margin-right-override: 0;
 }
--- a/Source/WebInspectorUI/UserInterface/Views/ColorPicker.js
+++ b/Source/WebInspectorUI/UserInterface/Views/ColorPicker.js
@@ -25,13 +25,18 @@
 
 WI.ColorPicker = class ColorPicker extends WI.Object
 {
-    constructor({preventChangingColorFormats, colorVariables} = {})
+    constructor({preventChangingColorFormats, colorVariables, contrastInfo} = {})
     {
         super();
 
         this._preventChangingColorFormats = !!preventChangingColorFormats;
+        this._contrastInfo = contrastInfo || null;
 
         this._colorSquare = new WI.ColorSquare(this, 200);
+        if (this._contrastInfo?.backgroundColor) {
+            this._colorSquare.contrastBackgroundColor = this._contrastInfo.backgroundColor;
+            this._colorSquare.isLargeText = !!this._contrastInfo.isLargeText;
+        }
 
         this._hueSlider = new WI.Slider;
         this._hueSlider.delegate = this;
@@ -92,6 +97,10 @@ WI.ColorPicker = class ColorPicker extends WI.Object
             colorInputsWrapperElement.appendChild(pickColorElement);
         }
 
+        this._contrastInfoElement = null;
+        if (this._contrastInfo?.backgroundColor)
+            this._createContrastInfoSection();
+
         this._opacity = 0;
         this._opacityPattern = "url(Images/Checkers.svg)";
 
@@ -103,7 +112,7 @@ WI.ColorPicker = class ColorPicker extends WI.Object
 
             let swatchesTitle = variableColorSwatchesContainer.appendChild(document.createElement("h2"));
             swatchesTitle.textContent = WI.UIString("Variables", "Variables @ Color Picker", "Title of swatches section in Color Picker");
-            
+
             let variableColorSwatchesListElement = variableColorSwatchesContainer.appendChild(document.createElement("ul"));
             let sortedColorVariables = WI.ColorPicker.sortColorVariables(colorVariables);
 
@@ -236,6 +245,7 @@ WI.ColorPicker = class ColorPicker extends WI.Object
             return;
 
         this._opacity = opacity;
+        this._colorSquare.opacity = opacity;
         this._updateColor();
     }
 
@@ -253,6 +263,7 @@ WI.ColorPicker = class ColorPicker extends WI.Object
         this._color = color;
 
         this._colorSquare.tintedColor = this._color;
+        this._colorSquare.opacity = this._color.alpha;
 
         this._hueSlider.value = this._color.hsl[0] / 360;
 
@@ -261,6 +272,7 @@ WI.ColorPicker = class ColorPicker extends WI.Object
 
         this._showColorComponentInputs();
         this._updateColorGamut();
+        this._updateContrastInfo();
 
         this._dontUpdateColor = false;
     }
@@ -318,6 +330,7 @@ WI.ColorPicker = class ColorPicker extends WI.Object
         this.dispatchEventToListeners(WI.ColorPicker.Event.ColorChanged, {color: this._color});
 
         this._updateColorGamut();
+        this._updateContrastInfo();
     }
 
     _updateOpacitySlider()
@@ -492,6 +505,57 @@ WI.ColorPicker = class ColorPicker extends WI.Object
                 this._updateColorGamut();
                 this.dispatchEventToListeners(WI.ColorPicker.Event.ColorChanged, {color: this._color});
             });
+        }
+    }
+
+    _createContrastInfoSection()
+    {
+        this._contrastInfoElement = this._element.appendChild(document.createElement("div"));
+        this._contrastInfoElement.classList.add("contrast-info");
+
+        let labelElement = this._contrastInfoElement.appendChild(document.createElement("span"));
+        labelElement.classList.add("contrast-label");
+        labelElement.textContent = WI.UIString("Contrast", "Contrast @ Color Picker", "Label for contrast ratio section in Color Picker");
+
+        this._contrastRatioElement = this._contrastInfoElement.appendChild(document.createElement("span"));
+        this._contrastRatioElement.classList.add("contrast-ratio");
+
+        this._complianceBadgeElement = this._contrastInfoElement.appendChild(document.createElement("span"));
+        this._complianceBadgeElement.classList.add("compliance-badge");
+
+        let separatorElement = this._contrastInfoElement.appendChild(document.createElement("span"));
+        separatorElement.classList.add("contrast-separator");
+        separatorElement.textContent = WI.UIString("vs", "vs @ Color Picker Contrast", "Separator between foreground and background colors in contrast info");
+
+        let backgroundSwatch = new WI.InlineSwatch(WI.InlineSwatch.Type.Color, this._contrastInfo.backgroundColor, {readOnly: true, tooltip: WI.UIString("Background Color")});
+        this._contrastInfoElement.appendChild(backgroundSwatch.element);
+    }
+
+    _updateContrastInfo()
+    {
+        if (!this._contrastInfo?.backgroundColor || !this._contrastInfoElement)
+            return;
+
+        let effectiveForeground = this._color.blendOverBackground(this._contrastInfo.backgroundColor);
+
+        let ratio = effectiveForeground.contrastRatio(this._contrastInfo.backgroundColor);
+        let isLargeText = !!this._contrastInfo.isLargeText;
+        let compliance = WI.Color.contrastComplianceForRatio(ratio, {isLargeText});
+
+        this._contrastRatioElement.textContent = ratio.toFixed(2) + ":1";
+
+        this._complianceBadgeElement.textContent = compliance;
+        this._complianceBadgeElement.className = "compliance-badge";
+        switch (compliance) {
+        case WI.Color.ContrastCompliance.AAA:
+            this._complianceBadgeElement.classList.add("contrast-aaa");
+            break;
+        case WI.Color.ContrastCompliance.AA:
+            this._complianceBadgeElement.classList.add("contrast-aa");
+            break;
+        case WI.Color.ContrastCompliance.Fail:
+            this._complianceBadgeElement.classList.add("contrast-fail");
+            break;
         }
     }
 };
--- a/Source/WebInspectorUI/UserInterface/Views/ColorSquare.css
+++ b/Source/WebInspectorUI/UserInterface/Views/ColorSquare.css
@@ -99,3 +99,56 @@
         stroke-opacity: var(--stroke-opacity) / 2;
     }
 }
+
+.color-square > .contrast-lines-svg {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    pointer-events: none;
+    z-index: 1;
+}
+
+.color-square > .contrast-lines-svg > .contrast-line {
+    fill: none;
+    stroke-width: 1px;
+    stroke-linecap: round;
+    stroke-linejoin: round;
+}
+
+.color-square > .contrast-lines-svg > .contrast-line.contrast-aa-threshold {
+    stroke: white;
+    stroke-opacity: 0.8;
+}
+
+.color-square > .contrast-lines-svg > .contrast-line.contrast-aaa-threshold {
+    stroke: white;
+    stroke-opacity: 0.6;
+}
+
+.color-square > .contrast-label {
+    position: absolute;
+    padding: 1px 3px;
+    font-size: 9px;
+    font-weight: bold;
+    border-radius: 2px;
+    pointer-events: none;
+    z-index: 2;
+    color: white;
+    text-shadow: 0 0 2px black;
+}
+
+.color-square > .contrast-label.contrast-aa-label {
+    opacity: 0.9;
+}
+
+.color-square > .contrast-label.contrast-aaa-label {
+    opacity: 0.7;
+}
+
+@media (prefers-color-scheme: dark) {
+    .color-square > .contrast-label {
+        text-shadow: 0 0 2px white;
+    }
+}
--- a/Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
+++ b/Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
@@ -34,6 +34,7 @@ WI.ColorSquare = class ColorSquare
         this._y = 0;
         this._gamut = null;
         this._crosshairPosition = null;
+        this._contrastBackgroundColor = null;
 
         this._element = document.createElement("div");
         this._element.className = "color-square";
@@ -46,8 +47,16 @@ WI.ColorSquare = class ColorSquare
         lightnessGradientElement.className = "lightness-gradient fill";
 
         this._srgbLabelElement = null;
-        this._svgElement = null;
-        this._polylineElement = null;
+        this._sRGBSVGElement = null;
+        this._sRGBPolylineElement = null;
+
+        this._contrastSVGElement = null;
+        this._contrastAAPolylineElement = null;
+        this._contrastAAAPolylineElement = null;
+        this._contrastAALabelElement = null;
+        this._contrastAAALabelElement = null;
+        this._opacity = 1;
+        this._isLargeText = false;
 
         this._element.addEventListener("mousedown", this);
         this._element.addEventListener("keydown", this._handleKeyDown.bind(this));
@@ -82,6 +91,46 @@ WI.ColorSquare = class ColorSquare
     {
         this._hue = hue;
         this._updateBaseColor();
+    }
+
+    get contrastBackgroundColor()
+    {
+        return this._contrastBackgroundColor;
+    }
+
+    set contrastBackgroundColor(color)
+    {
+        console.assert(!color || color instanceof WI.Color, color);
+        this._contrastBackgroundColor = color || null;
+        this._drawContrastLines();
+    }
+
+    get opacity()
+    {
+        return this._opacity;
+    }
+
+    set opacity(value)
+    {
+        if (this._opacity === value)
+            return;
+
+        this._opacity = value;
+        this._drawContrastLines();
+    }
+
+    get isLargeText()
+    {
+        return this._isLargeText;
+    }
+
+    set isLargeText(value)
+    {
+        if (this._isLargeText === value)
+            return;
+
+        this._isLargeText = value;
+        this._drawContrastLines();
     }
 
     get tintedColor()
@@ -237,6 +286,9 @@ WI.ColorSquare = class ColorSquare
 
         if (this._gamut === WI.Color.Gamut.DisplayP3)
             this._drawSRGBOutline();
+
+        if (this._contrastBackgroundColor)
+            this._drawContrastLines();
     }
 
     _updateCrosshairBackground()
@@ -246,18 +298,17 @@ WI.ColorSquare = class ColorSquare
 
     _drawSRGBOutline()
     {
-        if (!this._svgElement) {
+        if (!this._sRGBSVGElement) {
             this._srgbLabelElement = this._element.appendChild(document.createElement("span"));
             this._srgbLabelElement.className = "srgb-label";
             this._srgbLabelElement.textContent = WI.unlocalizedString("sRGB");
             this._srgbLabelElement.title = WI.UIString("Edge of sRGB color space", "Label for a guide within the color picker");
 
-            const svgNamespace = "http://www.w3.org/2000/svg";
-            this._svgElement = this._element.appendChild(document.createElementNS(svgNamespace, "svg"));
-            this._svgElement.classList.add("svg-root");
+            this._sRGBSVGElement = this._element.appendChild(createSVGElement("svg"));
+            this._sRGBSVGElement.classList.add("svg-root");
 
-            this._polylineElement = this._svgElement.appendChild(document.createElementNS(svgNamespace, "polyline"));
-            this._polylineElement.classList.add("srgb-edge");
+            this._sRGBPolylineElement = this._sRGBSVGElement.appendChild(createSVGElement("polyline"));
+            this._sRGBPolylineElement.classList.add("srgb-edge");
         }
 
         let points = [];
@@ -293,12 +344,171 @@ WI.ColorSquare = class ColorSquare
 
         this._srgbLabelElement.style.right = `${this._dimension - points.lastValue.x}px`;
 
-        this._polylineElement.points.clear();
+        this._sRGBPolylineElement.points.clear();
         for (let {x, y} of points) {
-            let svgPoint = this._svgElement.createSVGPoint();
+            let svgPoint = this._sRGBSVGElement.createSVGPoint();
             svgPoint.x = x;
             svgPoint.y = y;
-            this._polylineElement.points.appendItem(svgPoint);
+            this._sRGBPolylineElement.points.appendItem(svgPoint);
         }
+    }
+
+    _drawContrastLines()
+    {
+        if (!this._contrastBackgroundColor) {
+            if (this._contrastSVGElement) {
+                this._contrastSVGElement.hidden = true;
+                this._contrastAALabelElement.hidden = true;
+                this._contrastAAALabelElement.hidden = true;
+            }
+            return;
+        }
+
+        if (!this._contrastSVGElement) {
+            this._contrastSVGElement = this._element.appendChild(createSVGElement("svg"));
+            this._contrastSVGElement.classList.add("contrast-lines-svg");
+
+            this._contrastAAPolylineElement = this._contrastSVGElement.appendChild(createSVGElement("polyline"));
+            this._contrastAAPolylineElement.classList.add("contrast-line", "contrast-aa-threshold");
+
+            this._contrastAAAPolylineElement = this._contrastSVGElement.appendChild(createSVGElement("polyline"));
+            this._contrastAAAPolylineElement.classList.add("contrast-line", "contrast-aaa-threshold");
+
+            this._contrastAALabelElement = this._element.appendChild(document.createElement("span"));
+            this._contrastAALabelElement.className = "contrast-label contrast-aa-label";
+            this._contrastAALabelElement.textContent = "AA";
+
+            this._contrastAAALabelElement = this._element.appendChild(document.createElement("span"));
+            this._contrastAAALabelElement.className = "contrast-label contrast-aaa-label";
+            this._contrastAAALabelElement.textContent = "AAA";
+        }
+
+        this._contrastSVGElement.hidden = false;
+
+        let aaThreshold = this._isLargeText ? WI.Color.ContrastThreshold.AALargeText : WI.Color.ContrastThreshold.AA;
+        let aaaThreshold = this._isLargeText ? WI.Color.ContrastThreshold.AA : WI.Color.ContrastThreshold.AAA;
+
+        this._contrastAALabelElement.title = this._isLargeText
+            ? WI.UIString("WCAG AA minimum contrast for large text (3:1)", "WCAG AA minimum contrast for large text (3:1) @ Tooltip for AA contrast line in color picker", "Tooltip for AA contrast line in color picker for large text")
+            : WI.UIString("WCAG AA minimum contrast (4.5:1)", "WCAG AA minimum contrast (4.5:1) @ Tooltip for AA contrast line in color picker", "Tooltip for AA contrast line in color picker");
+
+        this._contrastAAALabelElement.title = this._isLargeText
+            ? WI.UIString("WCAG AAA enhanced contrast for large text (4.5:1)", "WCAG AAA enhanced contrast for large text (4.5:1) @ Tooltip for AAA contrast line in color picker", "Tooltip for AAA contrast line in color picker for large text")
+            : WI.UIString("WCAG AAA enhanced contrast (7:1)", "WCAG AAA enhanced contrast (7:1) @ Tooltip for AAA contrast line in color picker", "Tooltip for AAA contrast line in color picker");
+
+        let backgroundLuminance = this._contrastBackgroundColor.relativeLuminance();
+
+        let aaPoints = this._calculateContrastLinePoints(aaThreshold, backgroundLuminance);
+        this._updatePolylinePoints(this._contrastAAPolylineElement, aaPoints);
+        this._updateContrastLabel(this._contrastAALabelElement, aaPoints, backgroundLuminance);
+
+        let aaaPoints = this._calculateContrastLinePoints(aaaThreshold, backgroundLuminance);
+        this._updatePolylinePoints(this._contrastAAAPolylineElement, aaaPoints);
+        this._updateContrastLabel(this._contrastAAALabelElement, aaaPoints, backgroundLuminance);
+    }
+
+    _calculateContrastLinePoints(targetRatio, backgroundLuminance)
+    {
+        if (backgroundLuminance < 0.5)
+            targetRatio = 1 / targetRatio;
+
+        let targetLuminance = ((backgroundLuminance + 0.05) * targetRatio) - 0.05;
+
+        if (targetLuminance < 0 || targetLuminance > 1)
+            return [];
+
+        let points = [];
+        for (let x = 0; x <= this._dimension; ++x) {
+            let saturation = (x / this._dimension) * 100;
+            let y = this._findBrightnessForLuminance(saturation, targetLuminance);
+
+            if (y <= 0 || y >= this._dimension)
+                break;
+
+            points.push({x, y});
+        }
+
+        return points;
+    }
+
+    _findBrightnessForLuminance(saturation, targetLuminance)
+    {
+        let low = 0;
+        let high = 100;
+        const epsilon = 0.001;
+        const maxIterations = 20;
+
+        for (let i = 0; i < maxIterations; ++i) {
+            let mid = (low + high) / 2;
+            let rgb = WI.Color.hsv2rgb(this._hue, saturation, mid);
+            let color = new WI.Color(WI.Color.Format.ColorFunction, rgb.concat(this._opacity));
+
+            let effectiveColor = color;
+            if (this._opacity < 1 && this._contrastBackgroundColor)
+                effectiveColor = color.blendOverBackground(this._contrastBackgroundColor);
+
+            let luminance = effectiveColor.relativeLuminance();
+
+            if (Math.abs(luminance - targetLuminance) < epsilon)
+                return (1 - (mid / 100)) * this._dimension;
+
+            if (luminance < targetLuminance)
+                low = mid;
+            else
+                high = mid;
+        }
+
+        let mid = (low + high) / 2;
+        if (mid >= 99 || mid <= 1)
+            return Infinity;
+
+        return (1 - (mid / 100)) * this._dimension;
+    }
+
+    _updatePolylinePoints(polylineElement, points)
+    {
+        polylineElement.points.clear();
+
+        if (points.length === 0) {
+            polylineElement.hidden = true;
+            return;
+        }
+
+        polylineElement.hidden = false;
+
+        for (let {x, y} of points) {
+            let svgPoint = this._contrastSVGElement.createSVGPoint();
+            svgPoint.x = x;
+            svgPoint.y = y;
+            polylineElement.points.appendItem(svgPoint);
+        }
+    }
+
+    _updateContrastLabel(labelElement, points, backgroundLuminance)
+    {
+        if (points.length === 0) {
+            labelElement.hidden = true;
+            return;
+        }
+
+        labelElement.hidden = false;
+
+        let lastPoint = points[points.length - 1];
+        let firstPoint = points[0];
+
+        let useLastPoint = lastPoint.x >= this._dimension - 10;
+        let point = useLastPoint ? lastPoint : firstPoint;
+
+        if (useLastPoint) {
+            labelElement.style.right = "2px";
+            labelElement.style.removeProperty("left");
+        } else {
+            labelElement.style.left = `${point.x + 4}px`;
+            labelElement.style.removeProperty("right");
+        }
+
+        let labelHeight = 16;
+        let yOffset = backgroundLuminance < 0.5 ? -labelHeight : 4;
+        labelElement.style.top = `${Math.max(0, Math.min(this._dimension - labelHeight, point.y + yOffset))}px`;
     }
 };
--- a/Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js
+++ b/Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js
@@ -490,6 +490,35 @@ WI.SpreadsheetStyleProperty = class SpreadsheetStyleProperty extends WI.Object
         return this._property.ownerStyle.nodeStyles.computedStyle.variablesForType(WI.CSSStyleDeclaration.VariablesGroupType.Colors);
     }
 
+    inlineSwatchGetContrastInfo(inlineSwatch)
+    {
+        let propertyName = this._property.name.toLowerCase();
+        let isTextColorProperty = propertyName === "color" || propertyName === "-webkit-text-fill-color" || propertyName === "-webkit-text-stroke-color";
+
+        if (!isTextColorProperty)
+            return null;
+
+        let computedStyle = this._property.ownerStyle.nodeStyles?.computedStyle;
+        if (!computedStyle)
+            return null;
+
+        let backgroundColor = WI.Color.fromString(computedStyle.propertyForName("background-color")?.value ?? "");
+        if (!backgroundColor)
+            return null;
+
+        if (backgroundColor.alpha < 1)
+            backgroundColor = backgroundColor.blendOverBackground(WI.Color.fromString("white"));
+
+        let fontSizeInPt = parseFloat(computedStyle.propertyForName("font-size")?.value) * 0.75;
+        let isLargeText = fontSizeInPt >= 18;
+        if (!isLargeText && fontSizeInPt >= 14) {
+            let fontWeight = parseInt(computedStyle.propertyForName("font-weight")?.value);
+            isLargeText = fontWeight >= 700;
+        }
+
+        return {backgroundColor, isLargeText};
+    }
+
     // Private
 
     _toggle()
--- a/Source/WebInspectorUI/Localizations/en.lproj/localizedStrings.js
+++ b/Source/WebInspectorUI/Localizations/en.lproj/localizedStrings.js
@@ -443,6 +443,8 @@ localizedStrings["Connection Close Frame"] = "Connection Close Frame";
 localizedStrings["Connection Closed"] = "Connection Closed";
 localizedStrings["Connection ID"] = "Connection ID";
 localizedStrings["Connection:"] = "Connection:";
+/* Label for contrast ratio section in Color Picker */
+localizedStrings["Contrast @ Color Picker"] = "Contrast";
 localizedStrings["Console"] = "Console";
 localizedStrings["Console Evaluation"] = "Console Evaluation";
 localizedStrings["Console Evaluation %d"] = "Console Evaluation %d";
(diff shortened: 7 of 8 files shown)


burg:
This looks really cool and I can't wait to dig into how it works.

High level design feedback:
- How does it interact with other color picker sub-tools? is there a way to hide or show the contrast info? If I'm trying to make sure my site is compliant, this would be a wonderful addition. But if I was actually trying to pick a specific color, I'd want to minimize the visual bias of putting red and green blocks next to a color palette. beyond this PR, would be awesome to have a tool palette/dropdown so that the user's preference can be adjusted and remembered for next time.
- The intent is that this uses WCAG calculations that factor in text size. But in general, is this intended to be used to adjust text contrast only, or color contrast in general? Does the formula differ per color space?
- In some cases the actual displayed color on the screen (and therefore effective contrast) will not match the computed color because of partial opacity+image on the background. What if anything can we do about this? I guess if this is about text contrast only, it's not as important?
- The green contrast curves for AA and AAA look jagged, is there a way to stroke them with more antialiasing and/or a less contrasting color?
- As a new web developer, I wouldn't be familiar with what AA and AAA refer to. Can you find a place to add a help link that takes the user to Web Inspector documentation? It could be a (?) for the color picker in general.
- In the second green curve screenshot, the text is overlaid on the curve line. This makes it ambiguous as to which of the three graph segments are actually AA/AA/Fail. When the text is on one side, I think it is clear enough.

Other feedback:
- When we've agreed on the design, please take a bug for updating Web Inspector Reference.
- Commit feedback: Missing space after '.'. Overall, I wanted actual links to the content that is referenced so that I could double-check what it says. Can you add a WCAG spec permalink?

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -882,6 +882,75 @@ WI.Color = class Color
             hex = "0" + hex;
         return hex;
     }
+
+    relativeLuminance()
+    {
+        let [r, g, b] = this.normalizedRGB;
+
+        const toLinear = (c) => c <= 0.04045 ? c / 12.92 : Math.pow((c + 0.055) / 1.055, 2.4);
```
dcrousso: we already have this same code below that w can use here instead
```js
let [r, g, b] = WI.Color._toLinearLight(this.normalizedRGB);
return (0.2126729 * r) + (0.7151522 * g) + (0.0721750 * b);
```

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -882,6 +882,75 @@ WI.Color = class Color
...
+
+        return 0.2126 * R + 0.7152 * G + 0.0722 * B;
+    }
+
+    contrastRatio(otherColor)
+    {
+        console.assert(otherColor instanceof WI.Color);
+
+        let L1 = this.relativeLuminance();
+        let L2 = otherColor.relativeLuminance();
```
dcrousso: Style: these should be named `l1` and `l2` as capitals are reserved for `class`

Source/WebInspectorUI/UserInterface/Models/Color.js
dcrousso: also, if this code is from a spec or website then please link it here so that there's some clarity as to where these magic numbers come from (e.g. i found the expanded forms of the decimal values in `WI.Color.srgbToDisplayP3`)

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -882,6 +882,75 @@ WI.Color = class Color
...
+        let R = toLinear(r);
+        let G = toLinear(g);
+        let B = toLinear(b);
+
+        return 0.2126 * R + 0.7152 * G + 0.0722 * B;
+    }
+
+    contrastRatio(otherColor)
+    {
+        console.assert(otherColor instanceof WI.Color);
```
dcrousso: NIT: please add `otherColor` as a second argument in case this fails so we can see what value as passed in instead of just that the assertion failed
```js
console.assert(otherColor instanceof WI.Color, otherColor);
```

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -882,6 +882,75 @@ WI.Color = class Color
...
+        const toLinear = (c) => c <= 0.04045 ? c / 12.92 : Math.pow((c + 0.055) / 1.055, 2.4);
+
+        let R = toLinear(r);
+        let G = toLinear(g);
+        let B = toLinear(b);
+
+        return 0.2126 * R + 0.7152 * G + 0.0722 * B;
+    }
+
+    contrastRatio(otherColor)
```
dcrousso: NIT: i think we can just call this `other` since it's obvious from the next line that it's expecting a `WI.Color` (and it's not like you'd be able to compare a `WI.Color` with anything else anyways)

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -882,6 +882,75 @@ WI.Color = class Color
...
+        let blendedRGB = [
+            fgRGB[0] * alpha + bgRGB[0] * (1 - alpha),
+            fgRGB[1] * alpha + bgRGB[1] * (1 - alpha),
+            fgRGB[2] * alpha + bgRGB[2] * (1 - alpha),
+        ];
+
+        return new WI.Color(WI.Color.Format.RGB, blendedRGB.map(c => c * 255).concat(1));
+    }
+
+    static contrastComplianceForRatio(ratio, isLargeText = false)
```
dcrousso: Style: please move this up to the end of the `// Static` section (right above the `// Public`)

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -882,6 +882,75 @@ WI.Color = class Color
...
+
+    contrastCompliance(otherColor, isLargeText = false)
+    {
+        let ratio = this.contrastRatio(otherColor);
+        return WI.Color.contrastComplianceForRatio(ratio, isLargeText);
+    }
+
+    blendOverBackground(backgroundColor)
+    {
+        console.assert(backgroundColor instanceof WI.Color);
```
dcrousso: ditto about adding a second argument

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -882,6 +882,75 @@ WI.Color = class Color
...
+            fgRGB[2] * alpha + bgRGB[2] * (1 - alpha),
+        ];
+
+        return new WI.Color(WI.Color.Format.RGB, blendedRGB.map(c => c * 255).concat(1));
+    }
+
+    static contrastComplianceForRatio(ratio, isLargeText = false)
+    {
+        if (isLargeText) {
+            if (ratio >= WI.Color.ContrastThreshold.AA)
```
dcrousso: oops did you mean `WI.Color.ContrastThreshold.AAA`?

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -882,6 +882,75 @@ WI.Color = class Color
...
+
+    blendOverBackground(backgroundColor)
+    {
+        console.assert(backgroundColor instanceof WI.Color);
+
+        if (this.alpha === 1)
+            return this.copy();
+
+        let fgRGB = this.normalizedRGB;
+        let bgRGB = backgroundColor.normalizedRGB;
```
dcrousso: Style: please dont abbreviate names as it's harder for those unfamiliar to understand what they mean
```js
let foregroundRGB = this.normalizedRGB;
let backgroundRGB = backgroundColor.normalizedRGB;
```

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -882,6 +882,75 @@ WI.Color = class Color
...
+        let alpha = this.alpha;
+
+        let blendedRGB = [
+            fgRGB[0] * alpha + bgRGB[0] * (1 - alpha),
```
dcrousso: Style: please add `(` `)` when switching between operators so that it's explicitly clear what happens first
```js
(fgRGB[0] * alpha) + (bgRGB[0] * (1 - alpha)),
```

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -882,6 +882,75 @@ WI.Color = class Color
...
+            fgRGB[1] * alpha + bgRGB[1] * (1 - alpha),
+            fgRGB[2] * alpha + bgRGB[2] * (1 - alpha),
+        ];
+
+        return new WI.Color(WI.Color.Format.RGB, blendedRGB.map(c => c * 255).concat(1));
```
dcrousso: Style: please give arguments a more real name where able and if it adds more clarity
Style: we always add `(` `)` even for single argument arrow functions
```js
return new WI.Color(WI.Color.Format.RGB, blendedRGB.map((component) => component * 255));
```
note that the `.concat(1)` should be unnecessary since `WI.Color` will assume an `alpha` of `1` if only given three arguments

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -882,6 +882,75 @@ WI.Color = class Color
...
+    }
+
+    static contrastComplianceForRatio(ratio, isLargeText = false)
+    {
+        if (isLargeText) {
```
dcrousso: rather than duplicating the entire decision tree, can we just modify the one `if`?
```js
if (ratio >= WI.Color.ContrastThreshold.AAA)
    return WI.Color.ContrastCompliance.AAA;

let aaThreshold = isLargeText ? WI.Color.ContrastThreshold.AALargeText : WI.Color.ContrastThreshold.AA;
if (ratio >= aaThreshold)
    return WI.Color.ContrastCompliance.AA;

return WI.Color.ContrastCompliance.Fail;
```

Source/WebInspectorUI/UserInterface/Models/Color.js
dcrousso: Style: we much prefer putting optional arguments into a single `options = {}` at the end so that it's clear from callsites what it means (since it's a keyed value in an object) and allows for more arguments in the future
```js
static contrastComplianceForRatio(ratio, {isLargeText} = {})
```

Source/WebInspectorUI/UserInterface/Views/ColorPicker.css
```diff
@@ -136,4 +142,55 @@
...
+    border-top: 1px solid var(--border-color);
+    font-size: 11px;
+}
+
+.color-picker > .contrast-info > .contrast-label {
+    color: var(--text-color-secondary);
+}
+
+.color-picker > .contrast-info > .contrast-ratio {
+    font-weight: 600;
```
dcrousso: i feel like this draws too much attention for developers that are just trying to change a color

the big red "FAIL" badge should probably be enough visual indicator so i'd probably leave this as regular text

Source/WebInspectorUI/UserInterface/Views/ColorPicker.js
```diff
@@ -25,13 +25,18 @@
 
 WI.ColorPicker = class ColorPicker extends WI.Object
 {
-    constructor({preventChangingColorFormats, colorVariables} = {})
+    constructor({preventChangingColorFormats, colorVariables, contrastInfo} = {})
     {
         super();
 
         this._preventChangingColorFormats = !!preventChangingColorFormats;
+        this._contrastInfo = contrastInfo;
```
dcrousso: Style: this should have a `|| null` if it's optional

Source/WebInspectorUI/UserInterface/Views/ColorPicker.js
```diff
@@ -25,13 +25,18 @@
...
 
         this._colorSquare = new WI.ColorSquare(this, 200);
+        if (this._contrastInfo?.backgroundColor) {
+            this._colorSquare.contrastBackgroundColor = this._contrastInfo.backgroundColor;
+            this._colorSquare.isLargeText = this._contrastInfo.isLargeText || false;
```
dcrousso: NIT: normally we use `!!` instead of `|| false` as a fallback value for booleans

Source/WebInspectorUI/UserInterface/Views/ColorPicker.js
```diff
@@ -182,6 +191,7 @@ WI.ColorPicker = class ColorPicker extends WI.Object
             return;
 
         this._opacity = opacity;
+        this._colorSquare.opacity = opacity;
```
dcrousso: hmm im not sure we wanna do this as if it's a super low value then it could be extremely hard to see what's even in the `WI.ColorSquare`

plus, the `WI.Slider` for opacity already shows what the current color looks like with opacity considered

im not necessarily against this change, but i think it's separate from the purpose of this PR (i.e. adding color contrast info)

ditto below

Source/WebInspectorUI/UserInterface/Views/ColorPicker.js
```diff
@@ -401,6 +414,61 @@ WI.ColorPicker = class ColorPicker extends WI.Object
...
+        this._contrastRatioElement.classList.add("contrast-ratio");
+
+        this._complianceBadgeElement = this._contrastInfoElement.appendChild(document.createElement("span"));
+        this._complianceBadgeElement.classList.add("compliance-badge");
+
+        let separatorElement = this._contrastInfoElement.appendChild(document.createElement("span"));
+        separatorElement.classList.add("contrast-separator");
+        separatorElement.textContent = WI.UIString("vs", "vs @ Color Picker Contrast", "Separator between foreground and background colors in contrast info");
+
+        let backgroundSwatch = new WI.InlineSwatch(WI.InlineSwatch.Type.Color, this._contrastInfo.backgroundColor, {readOnly: true});
```
dcrousso: should we maybe also add something like `tooltip: WI.UIString("Background Color")`?

Source/WebInspectorUI/UserInterface/Views/ColorPicker.js
```diff
@@ -401,6 +414,61 @@ WI.ColorPicker = class ColorPicker extends WI.Object
...
+    }
+
+    _updateContrastInfo()
+    {
+        if (!this._contrastInfo?.backgroundColor || !this._contrastInfoElement)
+            return;
+
+        let effectiveForeground = this._color.alpha < 1
+            ? this._color.blendOverBackground(this._contrastInfo.backgroundColor)
+            : this._color;
```
dcrousso: Style: we dont normally split ternaries onto multiple lines like this
```js
let effectiveForeground = this._color.blendOverBackground(this._contrastInfo.backgroundColor);
```
note that doesn't even need to check `this._color.alpha` since `WI.Color.prototype.blendOverBackground` will do that for you :)

Source/WebInspectorUI/UserInterface/Views/ColorPicker.js
```diff
@@ -401,6 +414,61 @@ WI.ColorPicker = class ColorPicker extends WI.Object
...
+        let isLargeText = this._contrastInfo.isLargeText || false;
+        let compliance = WI.Color.contrastComplianceForRatio(ratio, isLargeText);
+
+        this._contrastRatioElement.textContent = ratio.toFixed(2) + ":1";
+
+        this._complianceBadgeElement.textContent = compliance;
+        this._complianceBadgeElement.className = "compliance-badge";
+        switch (compliance) {
+        case WI.Color.ContrastCompliance.AAA:
+            this._complianceBadgeElement.classList.add("aaa");
```
dcrousso: i think `contrast-aaa` would be clearer

ditto below (especially `contrast-fail`)

Source/WebInspectorUI/UserInterface/Views/ColorPicker.js
```diff
@@ -401,6 +414,61 @@ WI.ColorPicker = class ColorPicker extends WI.Object
...
+    {
+        if (!this._contrastInfo?.backgroundColor || !this._contrastInfoElement)
+            return;
+
+        let effectiveForeground = this._color.alpha < 1
+            ? this._color.blendOverBackground(this._contrastInfo.backgroundColor)
+            : this._color;
+
+        let ratio = effectiveForeground.contrastRatio(this._contrastInfo.backgroundColor);
+        let isLargeText = this._contrastInfo.isLargeText || false;
```
dcrousso: ditto about using `!!`

Source/WebInspectorUI/UserInterface/Views/ColorSquare.css
```diff
@@ -99,3 +99,53 @@
...
+
+.color-square > .contrast-label.aa-label {
+    color: hsl(90, 80%, 70%);
+}
+
+.color-square > .contrast-label.aaa-label {
+    color: hsl(120, 70%, 60%);
+}
+
+@media (prefers-color-scheme: light) {
```
dcrousso: our usual assumption is that things start out in light mode and then change into dark mode, so please flip this to `prefers-color-scheme: dark` and adjust the CSS accordingly

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -84,6 +93,46 @@ WI.ColorSquare = class ColorSquare
...
     }
 
+    get contrastBackgroundColor()
+    {
+        return this._contrastBackgroundColor;
+    }
+
+    set contrastBackgroundColor(color)
+    {
+        console.assert(!color || color instanceof WI.Color);
```
dcrousso: ditto about a second argument

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -84,6 +93,46 @@ WI.ColorSquare = class ColorSquare
...
+        this._contrastBackgroundColor = color;
```
dcrousso: since this is optional please add a `|| null`

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -301,4 +353,157 @@ WI.ColorSquare = class ColorSquare
...
     }
+
+    _drawContrastLines()
+    {
+        if (!this._contrastBackgroundColor)
+            return;
+
+        if (!this._contrastSvgElement) {
+            const svgNamespace = "http://www.w3.org/2000/svg";
+            this._contrastSvgElement = this._element.appendChild(document.createElementNS(svgNamespace, "svg"));
```
dcrousso: you can use `createSVGElement` instead of this :)

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -301,4 +353,157 @@ WI.ColorSquare = class ColorSquare
...
+
+        this._aaLabelElement.title = this._isLargeText
+            ? WI.UIString("WCAG AA minimum contrast for large text (3:1)", "WCAG AA minimum contrast for large text (3:1) @ Tooltip for AA contrast line in color picker", "Tooltip for AA contrast line in color picker for large text")
+            : WI.UIString("WCAG AA minimum contrast (4.5:1)", "WCAG AA minimum contrast (4.5:1) @ Tooltip for AA contrast line in color picker", "Tooltip for AA contrast line in color picker");
+
+        this._aaaLabelElement.title = this._isLargeText
+            ? WI.UIString("WCAG AAA enhanced contrast for large text (4.5:1)", "WCAG AAA enhanced contrast for large text (4.5:1) @ Tooltip for AAA contrast line in color picker", "Tooltip for AAA contrast line in color picker for large text")
+            : WI.UIString("WCAG AAA enhanced contrast (7:1)", "WCAG AAA enhanced contrast (7:1) @ Tooltip for AAA contrast line in color picker", "Tooltip for AAA contrast line in color picker");
+
+        let bgLuminance = this._contrastBackgroundColor.relativeLuminance();
```
dcrousso: ditto about not abbreviating variable names

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -301,4 +353,157 @@ WI.ColorSquare = class ColorSquare
...
+        let aaaPoints = this._calculateContrastLinePoints(aaaThreshold, bgLuminance, isDarkBackground);
+        this._updatePolylinePoints(this._aaaPolylineElement, aaaPoints);
+        this._updateContrastLabel(this._aaaLabelElement, aaaPoints, isDarkBackground);
+    }
+
+    _calculateContrastLinePoints(targetRatio, bgLuminance, isDarkBackground)
+    {
+        let targetLuminance;
+        if (isDarkBackground)
+            targetLuminance = targetRatio * (bgLuminance + 0.05) - 0.05;
```
dcrousso: ditto about adding `(` `)` to make the operator ordering explicit

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -301,4 +353,157 @@ WI.ColorSquare = class ColorSquare
...
+        else
+            targetLuminance = (bgLuminance + 0.05) / targetRatio - 0.05;
+
+        if (targetLuminance < 0 || targetLuminance > 1)
+            return [];
+
+        let points = [];
+        for (let x = 0; x <= this._dimension; x++) {
```
dcrousso: NIT: `++x`

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -301,4 +353,157 @@ WI.ColorSquare = class ColorSquare
...
+    }
+
+    _findBrightnessForLuminance(saturation, targetLuminance)
+    {
+        let low = 0;
+        let high = 100;
+        const epsilon = 0.001;
+        const maxIterations = 20;
+
+        for (let i = 0; i < maxIterations; i++) {
```
dcrousso: NIT: `++i`

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -301,4 +353,157 @@ WI.ColorSquare = class ColorSquare
...
+            svgPoint.x = x;
+            svgPoint.y = y;
+            polylineElement.points.appendItem(svgPoint);
+        }
+    }
+
+    _updateContrastLabel(labelElement, points, isDarkBackground)
+    {
+        if (points.length === 0) {
+            labelElement.style.display = "none";
```
dcrousso: can we use `labelElement.hidden = true;` instead?

ditto with `labelElement.hidden = false;` below

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -301,4 +353,157 @@ WI.ColorSquare = class ColorSquare
...
+
+        return (1 - (mid / 100)) * this._dimension;
+    }
+
+    _updatePolylinePoints(polylineElement, points)
+    {
+        polylineElement.points.clear();
+
+        if (points.length === 0) {
+            polylineElement.style.display = "none";
```
dcrousso: can we use `polylineElement.hidden = true;` instead?

ditto with `polylineElement.hidden = false;` below

Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js
```diff
@@ -490,6 +490,107 @@ WI.SpreadsheetStyleProperty = class SpreadsheetStyleProperty extends WI.Object
         return this._property.ownerStyle.nodeStyles.computedStyle.variablesForType(WI.CSSStyleDeclaration.VariablesGroupType.Colors);
     }
 
+    inlineSwatchGetContrastInfo(inlineSwatch)
+    {
+        let propertyName = this._property.name.toLowerCase();
+        let isTextColorProperty = propertyName === "color"
+            || propertyName === "-webkit-text-fill-color"
+            || propertyName === "-webkit-text-stroke-color";
```
dcrousso: Style: ditto about not splitting this onto multiple lines

Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js
```diff
@@ -490,6 +490,107 @@ WI.SpreadsheetStyleProperty = class SpreadsheetStyleProperty extends WI.Object
...
+        let backgroundColorValue = computedStyle.propertyForName("background-color")?.value;
+        if (!backgroundColorValue)
+            return null;
+
+        let backgroundColor = WI.Color.fromString(backgroundColorValue);
+        if (!backgroundColor)
+            return null;
+
+        if (backgroundColor.alpha < 1) {
+            let whiteBackground = WI.Color.fromString("white");
```
dcrousso: NIT: id inline this

Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js
```diff
@@ -490,6 +490,107 @@ WI.SpreadsheetStyleProperty = class SpreadsheetStyleProperty extends WI.Object
...
+    {
+        let computedStyle = nodeStyles.computedStyle;
+        if (!computedStyle)
+            return false;
+
+        let fontSizeValue = computedStyle.propertyForName("font-size")?.value;
+        if (!fontSizeValue)
+            return false;
+
+        let fontSizeMatch = fontSizeValue.match(/^([\d.]+)(px|pt|em|rem)?$/);
```
dcrousso: this seems pretty fragile.  is there any other way we could have the backend compute this instead?

Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js
```diff
@@ -490,6 +490,107 @@ WI.SpreadsheetStyleProperty = class SpreadsheetStyleProperty extends WI.Object
...
+        switch (unit) {
+        case "pt":
+            fontSizeInPt = fontSize;
+            break;
+        case "px":
+            fontSizeInPt = fontSize * 0.75;
+            break;
+        case "em":
+        case "rem":
+            fontSizeInPt = fontSize * 16 * 0.75;
```
dcrousso: what if the user has changed the default font size to something else?

Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js
```diff
@@ -490,6 +490,107 @@ WI.SpreadsheetStyleProperty = class SpreadsheetStyleProperty extends WI.Object
...
+        if (fontSizeInPt >= 18)
+            return true;
+
+        if (fontSizeInPt >= 14) {
+            let fontWeight = computedStyle.propertyForName("font-weight")?.value;
+            if (fontWeight) {
+                let weight = parseInt(fontWeight, 10);
+                if (!isNaN(weight) && weight >= 700)
+                    return true;
+                if (fontWeight === "bold" || fontWeight === "bolder")
```
dcrousso: NIT: i bet these are more common than the number value so id check these first

Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js
dcrousso: actually yeah isn't the computed style always in `px`?

Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js
dcrousso: similar to the above, isn't the computed value always a number?

Source/WebInspectorUI/UserInterface/Views/ColorSquare.css
```diff
@@ -99,3 +99,53 @@
...
+
+.color-square > .contrast-lines-svg > .contrast-line {
+    fill: none;
+    stroke-width: 1px;
+    stroke-linecap: round;
+    stroke-linejoin: round;
+}
+
+.color-square > .contrast-lines-svg > .contrast-line.aa-threshold {
+    stroke: hsl(90, 70%, 50%);
```
dcrousso: is there a reason we're using a green color for this instead of just a white (or even more ideally something that contrasts with the selected color)?

Source/WebInspectorUI/UserInterface/Views/ColorSquare.css
```diff
@@ -99,3 +99,53 @@
...
+}
+
+.color-square > .contrast-lines-svg > .contrast-line.aaa-threshold {
+    stroke: hsl(120, 60%, 40%);
```
dcrousso: ditto

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -882,6 +882,75 @@ WI.Color = class Color
             hex = "0" + hex;
         return hex;
     }
+
+    relativeLuminance()
+    {
+        let [r, g, b] = this.normalizedRGB;
```
dcrousso: i think this can also return values >1 for P3 colors

does all the math below handle that too?

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -49,6 +50,14 @@ WI.ColorSquare = class ColorSquare
         this._svgElement = null;
         this._polylineElement = null;
```
dcrousso: we should rename this to `_sRGBPolylineElement` (and maybe rename `_sRGBLabelElement` as well) for clarity

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -49,6 +50,14 @@ WI.ColorSquare = class ColorSquare
...
 
+        this._contrastSvgElement = null;
+        this._aaPolylineElement = null;
```
dcrousso: id similarly call these `_contrastAAPolylineElement` and `_contrastAALabelElement` and etc.

dcrousso (commented):
really awesome work!  mostly style and NIT with a few major comments

also +1 to everything @burg mentioned with a few specific responses

> - How does it interact with other color picker sub-tools? is there a way to hide or show the contrast info? If I'm trying to make sure my site is compliant, this would be a wonderful addition. But if I was actually trying to pick a specific color, I'd want to minimize the visual bias of putting red and green blocks next to a color palette. beyond this PR, would be awesome to have a tool palette/dropdown so that the user's preference can be adjusted and remembered for next time.

while i agree that it would be nice to be able to hide this, since it's a single line i think adding a clickable header to control whether the section is expanded would result in more space being consumed (i.e. why have a section to hide one "row" when you can just show the one "row" in the same amount of space?)

i do agree that a checkbox or something that would allow developers to at least turn off the lines might be nice, but at the same time this is probably something that developers should _always_ be aware of (at least when dealing with text/background colors) so maybe it's a nice forcing function for it to always be there?

and while we're at it, maybe the contrast lines and compliance should also be shown when editing `background-color` too?

> - The green contrast curves for AA and AAA look jagged, is there a way to stroke them with more antialiasing and/or a less contrasting color?

FYI this is also the case for the sRGB limit line when editing P3 colors

> - In the second green curve screenshot, the text is overlaid on the curve line. This makes it ambiguous as to which of the three graph segments are actually AA/AA/Fail. When the text is on one side, I think it is clear enough.

yeah we definitely dont ever want the text to overlap with the line

along these lines (pun intended), is it possible for the AA and AAA lines to be super close together?  if so, what does this UI look like in that case?

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -293,12 +344,164 @@ WI.ColorSquare = class ColorSquare
...
+            this._sRGBPolylineElement.points.appendItem(svgPoint);
         }
     }
+
+    _drawContrastLines()
+    {
+        if (!this._contrastBackgroundColor)
+            return;
+
+        if (!this._contrastSvgElement) {
```
burg: Nit: _contrastSVGElement?

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -919,6 +931,53 @@ WI.Color = class Color
             hex = "0" + hex;
         return hex;
     }
+
+    // https://www.w3.org/TR/WCAG20/#relativeluminancedef
+    relativeLuminance()
```
burg, line 937: Shouldn't we have some basic test coverage for new Color.js utilities?

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -919,6 +931,53 @@ WI.Color = class Color
...
+
+        let foregroundRGB = this.normalizedRGB;
+        let backgroundRGB = backgroundColor.normalizedRGB;
+        let alpha = this.alpha;
+
+        let blendedRGB = [
+            (foregroundRGB[0] * alpha) + (backgroundRGB[0] * (1 - alpha)),
+            (foregroundRGB[1] * alpha) + (backgroundRGB[1] * (1 - alpha)),
+            (foregroundRGB[2] * alpha) + (backgroundRGB[2] * (1 - alpha)),
+        ];
```
dcrousso, line 977: NIT: i realize it's slightly cleaner to have the `map` below but it'd be nice to avoid the extra array since we create the source here
```js
let blendedRGB = [
    ((foregroundRGB[0] * alpha) + (backgroundRGB[0] * (1 - alpha))) * 255,
    ((foregroundRGB[1] * alpha) + (backgroundRGB[1] * (1 - alpha))) * 255,
    ((foregroundRGB[2] * alpha) + (backgroundRGB[2] * (1 - alpha))) * 255,
];
```

Source/WebInspectorUI/UserInterface/Models/Color.js
```diff
@@ -919,6 +931,53 @@ WI.Color = class Color
...
+
+        return new WI.Color(WI.Color.Format.RGB, blendedRGB.map((component) => component * 255));
```
dcrousso: should we try to propagate at least the `gamut` from `this`?
```js
return new WI.Color(WI.Color.Format.ColorFunction, blendedRGB, this.gamut);
```
note that im using `color()` as i dont think `rgb()` would suffice if we're blending two P3 colors (but i could be wrong)

Source/WebInspectorUI/UserInterface/Views/ColorPicker.css
```diff
@@ -137,4 +143,54 @@
...
+
+.color-picker > .contrast-info > .contrast-ratio {
+    font-variant-numeric: tabular-nums;
+}
+
+.color-picker > .contrast-info > .compliance-badge {
+    padding: 2px 6px;
+    border-radius: 3px;
+    font-size: 10px;
+    font-weight: 600;
```
dcrousso: why not just use `bold`?

Source/WebInspectorUI/UserInterface/Views/ColorPicker.css
```diff
@@ -137,4 +143,54 @@
...
+    text-transform: uppercase;
+}
+
+.color-picker > .contrast-info > .compliance-badge.contrast-aaa {
+    background-color: hsl(120, 50%, 35%);
```
dcrousso, line 175: does this color look good in dark mode?

Source/WebInspectorUI/UserInterface/Views/ColorPicker.css
```diff
@@ -137,4 +143,54 @@
...
+    color: white;
+}
+
+.color-picker > .contrast-info > .compliance-badge.contrast-aa {
+    background-color: hsl(90, 50%, 40%);
```
dcrousso, line 186: ditto about dark mode

Source/WebInspectorUI/UserInterface/Views/ColorPicker.css
```diff
@@ -137,4 +143,54 @@
...
+    color: white;
+}
+
+.color-picker > .contrast-info > .compliance-badge.contrast-fail {
+    background-color: hsl(0, 60%, 50%);
```
dcrousso, line 197: ditto about dark mode

Source/WebInspectorUI/UserInterface/Views/ColorPicker.js
```diff
@@ -494,6 +507,59 @@ WI.ColorPicker = class ColorPicker extends WI.Object
...
+        this._complianceBadgeElement.classList.add("compliance-badge");
+
+        let separatorElement = this._contrastInfoElement.appendChild(document.createElement("span"));
+        separatorElement.classList.add("contrast-separator");
+        separatorElement.textContent = WI.UIString("vs", "vs @ Color Picker Contrast", "Separator between foreground and background colors in contrast info");
+
+        let backgroundSwatch = new WI.InlineSwatch(WI.InlineSwatch.Type.Color, this._contrastInfo.backgroundColor, {readOnly: true, tooltip: WI.UIString("Background Color")});
+        this._contrastInfoElement.appendChild(backgroundSwatch.element);
+
+        this._element.appendChild(this._contrastInfoElement);
```
dcrousso: why not do this on the first line?
```js
this._contrastInfoElement = this._element.appendChild(document.createElement("div"));
```

Source/WebInspectorUI/UserInterface/Views/ColorPicker.js
```diff
@@ -494,6 +507,59 @@ WI.ColorPicker = class ColorPicker extends WI.Object
...
+        if (!this._contrastInfo?.backgroundColor || !this._contrastInfoElement)
+            return;
+
+        let effectiveForeground = this._color.blendOverBackground(this._contrastInfo.backgroundColor);
+
+        let ratio = effectiveForeground.contrastRatio(this._contrastInfo.backgroundColor);
+        let isLargeText = !!this._contrastInfo.isLargeText;
+        let compliance = WI.Color.contrastComplianceForRatio(ratio, {isLargeText});
+
+        this._contrastRatioElement.textContent = ratio.toFixed(2) + ":1";
```
dcrousso, line 545: should this be localized?

Source/WebInspectorUI/UserInterface/Views/ColorSquare.css
```diff
@@ -99,3 +99,56 @@
...
+.color-square > .contrast-lines-svg > .contrast-line.contrast-aaa-threshold {
+    stroke: white;
+    stroke-opacity: 0.6;
+}
+
+.color-square > .contrast-label {
+    position: absolute;
+    padding: 1px 3px;
+    font-size: 9px;
+    font-weight: 600;
```
dcrousso: ditto about just using `bold`

Source/WebInspectorUI/UserInterface/Views/ColorSquare.css
```diff
@@ -99,3 +99,56 @@
...
+    border-radius: 2px;
+    pointer-events: none;
+    z-index: 2;
+    color: white;
+    text-shadow: 0 0 2px rgba(0, 0, 0, 0.5);
```
dcrousso: hmm i dont think we really use `text-shadow` or `drop-shadow` anywhere else in Web Inspector

is there a reason we have to have it here?

if the goal is just to help make the text legible, id suggest having it be a solid color instead of slightly transparent

Source/WebInspectorUI/UserInterface/Views/ColorSquare.css
```diff
@@ -99,3 +99,56 @@
...
+    opacity: 0.9;
+}
+
+.color-square > .contrast-label.contrast-aaa-label {
+    opacity: 0.7;
+}
+
+@media (prefers-color-scheme: dark) {
+    .color-square > .contrast-label {
+        text-shadow: 0 0 2px rgba(255, 255, 255, 0.8), 0 0 4px rgba(0, 0, 0, 0.3);
```
dcrousso: can you please attach some screen shots showing this?

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -46,8 +47,16 @@ WI.ColorSquare = class ColorSquare
         lightnessGradientElement.className = "lightness-gradient fill";
 
         this._srgbLabelElement = null;
-        this._svgElement = null;
-        this._polylineElement = null;
+        this._sRGBSvgElement = null;
```
dcrousso: NIT: i think we normally capitalize initialisms even when they're smushed into other stuff like this
```js
this._srgbSVGElement = null;
```

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -46,8 +47,16 @@ WI.ColorSquare = class ColorSquare
...
+        this._sRGBPolylineElement = null;
+
+        this._contrastSvgElement = null;
```
dcrousso: ditto
```js
this._contrastSVGElement = null;
```

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -293,12 +344,164 @@ WI.ColorSquare = class ColorSquare
...
+
+        if (targetLuminance < 0 || targetLuminance > 1)
+            return [];
+
+        let points = [];
+        for (let x = 0; x <= this._dimension; ++x) {
+            let saturation = (x / this._dimension) * 100;
+            let y = this._findBrightnessForLuminance(saturation, targetLuminance);
+
+            if (y === null || y <= 0 || y >= this._dimension)
```
dcrousso: see below about using `isNaN(y)` or not even needing anything specific if it's `Infinity` since that's handled by the last check

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -293,12 +344,164 @@ WI.ColorSquare = class ColorSquare
...
+
+            if (luminance < targetLuminance)
+                low = mid;
+            else
+                high = mid;
+        }
+
+        let mid = (low + high) / 2;
+        if (mid >= 99 || mid <= 1)
+            return null;
```
dcrousso: let's use `NaN` or `Infinity` so that it keeps the return type consistent

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -293,12 +344,164 @@ WI.ColorSquare = class ColorSquare
...
+
+        if (useLastPoint) {
+            labelElement.style.right = "2px";
+            labelElement.style.removeProperty("left");
+        } else {
+            labelElement.style.left = `${point.x + 4}px`;
+            labelElement.style.removeProperty("right");
+        }
+
+        let yOffset = isDarkBackground ? -16 : 4;
```
dcrousso: how were `-16` and `4` computed?

Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js
```diff
@@ -490,6 +490,84 @@ WI.SpreadsheetStyleProperty = class SpreadsheetStyleProperty extends WI.Object
...
+
+        let isLargeText = this._isLargeText(nodeStyles);
+
+        return {
+            backgroundColor,
+            isLargeText,
+        };
+    }
+
+    _getEffectiveBackgroundColor(nodeStyles)
```
dcrousso: if you inline this above i think that could save a few extra checks
```js
let computedStyle = this._property.ownerStyle.nodeStyles.computedStyle;

let backgroundColor = WI.Color.fromString(computedStyle.propertyForName("background-color")?.value ?? "");
if (!backgroundColor)
    return null;

if (backgroundColor.alpha < 1)
    backgroundColor = backgroundColor.blendOverBackground(WI.Color.fromString("white"));

if (backgroundColor.alpha === 0)
    backgroundColor = WI.Color.fromString("white");

let fontSizeInPt = parseFloat(computedStyle.propertyForName("font-size")?.value) * 0.75;
let isLargeText = fontSizeInPt >= 18;
if (!isLargeText && fontSizeInPt >= 14) {
    let fontWeight = parseInt(computedStyle.propertyForName("font-weight")?.value);
    isLargeText = fontWeight >= 700;
}

return {backgroundColor, isLargeText};
```

Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js
```diff
@@ -490,6 +490,84 @@ WI.SpreadsheetStyleProperty = class SpreadsheetStyleProperty extends WI.Object
...
+        if (backgroundColor.alpha < 1)
+            backgroundColor = backgroundColor.blendOverBackground(WI.Color.fromString("white"));
+
+        if (backgroundColor.alpha === 0)
+            return WI.Color.fromString("white");
+
+        return backgroundColor;
+    }
+
+    _isLargeText(nodeStyles)
```
dcrousso: see above

dcrousso (commented):
nice work!  i think just a few more things and we're good!

Source/WebInspectorUI/UserInterface/Models/Color.js
AnujPanta1: Inlined (without * 255 since we switched to ColorFunction)

Source/WebInspectorUI/UserInterface/Views/ColorPicker.js
AnujPanta1: I left as is because it's ratio notation, but I can make it into a localized string

Source/WebInspectorUI/UserInterface/Views/SpreadsheetStyleProperty.js
AnujPanta1: inlined, and cleaned the two methods

AnujPanta1:
changes & screen shots incoming in morning, thank you both for reviews !

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
AnujPanta1: There were done via/ trial during development

Source/WebInspectorUI/UserInterface/Views/ColorSquare.css
AnujPanta1: removed this

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
dcrousso: it'd be nice to tie this into some other value if possible (i.e. using some relevant existing CSS variable)

or at least have a comment explaining how you got these results

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -293,12 +344,164 @@ WI.ColorSquare = class ColorSquare
...
+        this._updateContrastLabel(this._contrastAAALabelElement, aaaPoints, isDarkBackground);
+    }
+
+    _calculateContrastLinePoints(targetRatio, backgroundLuminance, isDarkBackground)
+    {
+        let targetLuminance;
+        if (isDarkBackground)
+            targetLuminance = (targetRatio * (backgroundLuminance + 0.05)) - 0.05;
+        else
+            targetLuminance = ((backgroundLuminance + 0.05) / targetRatio) - 0.05;
```
dcrousso: since these are so similar i think you can simplify it
```js
if (!isDarkBackground)
    targetRatio = 1 / targetRatio;

let targetLuminance = ((backgroundLuminance + 0.05) * targetRatio) - 0.05;
```

Source/WebInspectorUI/UserInterface/Views/ColorSquare.js
```diff
@@ -293,12 +344,164 @@ WI.ColorSquare = class ColorSquare
...
+        this._contrastAALabelElement.title = this._isLargeText
+            ? WI.UIString("WCAG AA minimum contrast for large text (3:1)", "WCAG AA minimum contrast for large text (3:1) @ Tooltip for AA contrast line in color picker", "Tooltip for AA contrast line in color picker for large text")
+            : WI.UIString("WCAG AA minimum contrast (4.5:1)", "WCAG AA minimum contrast (4.5:1) @ Tooltip for AA contrast line in color picker", "Tooltip for AA contrast line in color picker");
+
+        this._contrastAAALabelElement.title = this._isLargeText
+            ? WI.UIString("WCAG AAA enhanced contrast for large text (4.5:1)", "WCAG AAA enhanced contrast for large text (4.5:1) @ Tooltip for AAA contrast line in color picker", "Tooltip for AAA contrast line in color picker for large text")
+            : WI.UIString("WCAG AAA enhanced contrast (7:1)", "WCAG AAA enhanced contrast (7:1) @ Tooltip for AAA contrast line in color picker", "Tooltip for AAA contrast line in color picker");
+
+        let backgroundLuminance = this._contrastBackgroundColor.relativeLuminance();
+        let isDarkBackground = backgroundLuminance < 0.5;
```
dcrousso: i feel like we should inline this in `_calculateContrastLinePoints` instead of having it be a separate variable (using the comment below)
```js
if (backgroundLuminance < 0.5)
    targetRatio = 1 / targetRatio;
```

Source/WebInspectorUI/UserInterface/Models/Color.js
dcrousso: @AnujPanta1 did you maybe forget to `git add` before updating the PR? 😅

dcrousso (approved):
awesome work!  please upload some final screenshots of light and dark mode with as many scenarios as you can so that if in the future we have to come back to this PR we can see what it looked like at the time :)

AnujPanta1:
image
image
image
image
image

AnujPanta1:
Yes you're right I didn't actually add some changes apologizes (found a commit sitting on top of my branch) , I'll implement suggested fixes and push up I left some screenshots of how things look in light/dark mode

AnujPanta1:
Follow ups include, adding tests for relativeLuminance, and adding background support

PR #57692, opened 2026-02-02
REGRESSION(266411@main): [GLib] `mousemove` events are not throttle to one per rendering update

obyknovenius:
#### 548162624138f2bbd8079ae8474feb8e504937f8
<pre>
REGRESSION(266411@main): [GLib] `mousemove` events are not throttle to one per rendering update
<a href="https://bugs.webkit.org/show_bug.cgi?id=306766">https://bugs.webkit.org/show_bug.cgi?id=306766</a>

Reviewed by Wenson Hsieh and Darin Adler.

Starting from 266411@main, it is required for a `DrawingArea` subclass
to override `scheduleRenderingUpdate()` and return `true` in order to
enable throttling of `mousemove` events to one per rendering update.
This method is currently only implemented by `RemoteLayerTreeDrawingArea`
which is used by Apple ports. The main purpose of it is to schedule next
rendering update to process deferred `mousemove` events. But it is
possible to schedule next rendering update by calling
`RenderingUpdateScheduler::scheduleRenderingUpdate()` which is driven by
the display refresh monitor.

This patch also implements `EventSenderProxy::waitForPendingMouseEvents()`
on GTK, WPE, and Windows by calling a new private
`WKPageDoAfterProcessingAllPendingMouseEvents()` function, which is
required to pass layout tests that use asynchronous mouse events.

Because `EventSenderProxy::waitForPendingMouseEvents()` internally calls
`m_testController-&gt;runUntil()`, `TestController::platformRunUntil()` now
explicitly iterates the main context instead of running and stopping the
main run loop to avoid completely stopping the TestController after
processing pending mouse events, which would cause tests to time out.

* Source/WebKit/UIProcess/API/C/WKPage.cpp:
(WKPageDoAfterProcessingAllPendingMouseEvents):
* Source/WebKit/UIProcess/API/C/WKPagePrivate.h:
(WebKit::WebPage::mouseEvent):
* Tools/WebKitTestRunner/gtk/EventSenderProxyGtk.cpp:
(WTR::doAfterProcessingAllPendingMouseEventsCallback):
(WTR::EventSenderProxy::waitForPendingMouseEvents):
* Tools/WebKitTestRunner/gtk/TestControllerGtk.cpp:
(WTR::TestController::notifyDone):
(WTR::TestController::platformRunUntil):
* Tools/WebKitTestRunner/libwpe/EventSenderProxyLibWPE.cpp:
(WTR::doAfterProcessingAllPendingMouseEventsCallback):
(WTR::EventSenderProxy::waitForPendingMouseEvents):
* Tools/WebKitTestRunner/win/EventSenderProxyWin.cpp:
(WTR::doAfterProcessingAllPendingMouseEventsCallback):
(WTR::EventSenderProxy::waitForPendingMouseEvents):
* Tools/WebKitTestRunner/wpe/TestControllerWPE.cpp:
(WTR::TestController::notifyDone):
(WTR::TestController::platformRunUntil):

Canonical link: <a href="https://commits.webkit.org/307370@main">https://commits.webkit.org/307370@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/bde4b2500c52dbc98136386c00b4a24d1c948a2d

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/144136 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/16815 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/8370 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/152806 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/97375 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/146011 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/17297 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/16709 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/110831 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/79653 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/147099 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/13250 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/129493 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/91749 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/12700 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/10438 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/252 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/122183 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/6146 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/155118 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/16667 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/7201 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/118847 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/16704 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/13991 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/119205 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30574 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/15091 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/127357 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/72088 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/16289 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/5795 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/16023 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/80068 "Built successfully") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/16234 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/16089 "Built successfully") | | |

### DIFF
--- a/Source/WebKit/WebProcess/WebPage/WebPage.cpp
+++ b/Source/WebKit/WebProcess/WebPage/WebPage.cpp
@@ -3690,7 +3690,7 @@ void WebPage::mouseEvent(FrameIdentifier frameID, const WebMouseEvent& mouseEven
 
     flushDeferredDidReceiveMouseEvent();
 
-    if (shouldDeferDidReceiveEvent && drawingArea->scheduleRenderingUpdate()) {
+    if (shouldDeferDidReceiveEvent) {
         // For mousemove events where the user is only hovering (not clicking and dragging),
         // we defer sending the DidReceiveEvent() IPC message until the end of the rendering
         // update to throttle the rate of these events to the rendering update frequency.
@@ -3698,6 +3698,7 @@ void WebPage::mouseEvent(FrameIdentifier frameID, const WebMouseEvent& mouseEven
         // coalesces mousemove events until the DidReceiveEvent() message is received after
         // the rendering update.
         m_deferredDidReceiveMouseEvent = { { mouseEvent.type(), handled } };
+        protect(corePage())->scheduleRenderingUpdate({ });
         return;
     }
 
--- a/Tools/WebKitTestRunner/gtk/EventSenderProxyGtk.cpp
+++ b/Tools/WebKitTestRunner/gtk/EventSenderProxyGtk.cpp
@@ -37,6 +37,7 @@
 #include "StringFunctions.h"
 #include "TestController.h"
 #include <WebKit/GtkVersioning.h>
+#include <WebKit/WKPagePrivate.h>
 #include <gdk/gdk.h>
 #include <gdk/gdkkeysyms.h>
 #include <webkit/WebKitWebViewBaseInternal.h>
@@ -431,8 +432,30 @@ void EventSenderProxy::setTouchModifier(WKEventModifiers, bool)
 }
 #endif // ENABLE(TOUCH_EVENTS)
 
+struct DoAfterProcessingAllPendingMouseEventsCallbackContext {
+    bool done { false };
+    bool timedOut { false };
+};
+
+static void doAfterProcessingAllPendingMouseEventsCallback(void* userData)
+{
+    auto* context = static_cast<DoAfterProcessingAllPendingMouseEventsCallbackContext*>(userData);
+    if (context->timedOut) {
+        delete context;
+        return;
+    }
+    context->done = true;
+}
+
 void EventSenderProxy::waitForPendingMouseEvents()
 {
+    auto* context = new DoAfterProcessingAllPendingMouseEventsCallbackContext;
+    WKPageDoAfterProcessingAllPendingMouseEvents(m_testController->mainWebView()->page(), context, doAfterProcessingAllPendingMouseEventsCallback);
+    m_testController->runUntil(context->done, 100_ms);
+    if (context->done)
+        delete context;
+    else
+        context->timedOut = true;
 }
 
 } // namespace WTR
--- a/Tools/WebKitTestRunner/gtk/TestControllerGtk.cpp
+++ b/Tools/WebKitTestRunner/gtk/TestControllerGtk.cpp
@@ -49,7 +49,6 @@ namespace WTR {
 
 void TestController::notifyDone()
 {
-    RunLoop::mainSingleton().stop();
 }
 
 void TestController::platformInitialize(const Options&)
@@ -68,7 +67,6 @@ void TestController::platformRunUntil(bool& done, WTF::Seconds timeout)
         TimeoutTimer(WTF::Seconds timeout, bool& timedOut)
             : m_timer(RunLoop::mainSingleton(), "TestController::TimeoutTimer"_s, [&timedOut] {
                 timedOut = true;
-                RunLoop::mainSingleton().stop();
             })
         {
             m_timer.setPriority(G_PRIORITY_DEFAULT_IDLE);
@@ -79,8 +77,9 @@ void TestController::platformRunUntil(bool& done, WTF::Seconds timeout)
         RunLoop::Timer m_timer;
     } timeoutTimer(timeout, timedOut);
 
+    auto* mainContext = g_main_context_default();
     while (!done && !timedOut)
-        RunLoop::mainSingleton().run();
+        g_main_context_iteration(mainContext, TRUE);
 }
 
 static char* getEnvironmentVariableAsUTF8String(const char* variableName)
--- a/Tools/WebKitTestRunner/libwpe/EventSenderProxyLibWPE.cpp
+++ b/Tools/WebKitTestRunner/libwpe/EventSenderProxyLibWPE.cpp
@@ -26,8 +26,10 @@
 #include "config.h"
 #include "EventSenderProxy.h"
 
+#include "PlatformWebView.h"
 #include "TestController.h"
 #include <WebCore/NotImplemented.h>
+#include <WebKit/WKPagePrivate.h>
 #include <wtf/MonotonicTime.h>
 
 #if USE(LIBWPE)
@@ -194,8 +196,30 @@ void EventSenderProxy::cancelTouchPoint(int index)
 
 #endif // ENABLE(TOUCH_EVENTS)
 
+struct DoAfterProcessingAllPendingMouseEventsCallbackContext {
+    bool done { false };
+    bool timedOut { false };
+};
+
+static void doAfterProcessingAllPendingMouseEventsCallback(void* userData)
+{
+    auto* context = static_cast<DoAfterProcessingAllPendingMouseEventsCallbackContext*>(userData);
+    if (context->timedOut) {
+        delete context;
+        return;
+    }
+    context->done = true;
+}
+
 void EventSenderProxy::waitForPendingMouseEvents()
 {
+    auto* context = new DoAfterProcessingAllPendingMouseEventsCallbackContext;
+    WKPageDoAfterProcessingAllPendingMouseEvents(m_testController->mainWebView()->page(), context, doAfterProcessingAllPendingMouseEventsCallback);
+    m_testController->runUntil(context->done, 100_ms);
+    if (context->done)
+        delete context;
+    else
+        context->timedOut = true;
 }
 
 } // namespace WTR
--- a/Tools/WebKitTestRunner/win/EventSenderProxyWin.cpp
+++ b/Tools/WebKitTestRunner/win/EventSenderProxyWin.cpp
@@ -30,6 +30,7 @@
 #include "PlatformWebView.h"
 #include "TestController.h"
 #include <WebCore/NotImplemented.h>
+#include <WebKit/WKPagePrivate.h>
 
 namespace WTR {
 
@@ -367,8 +368,30 @@ void EventSenderProxy::cancelTouchPoint(int)
 }
 #endif // ENABLE(TOUCH_EVENTS)
 
+struct DoAfterProcessingAllPendingMouseEventsCallbackContext {
+    bool done { false };
+    bool timedOut { false };
+};
+
+static void doAfterProcessingAllPendingMouseEventsCallback(void* userData)
+{
+    auto* context = static_cast<DoAfterProcessingAllPendingMouseEventsCallbackContext*>(userData);
+    if (context->timedOut) {
+        delete context;
+        return;
+    }
+    context->done = true;
+}
+
 void EventSenderProxy::waitForPendingMouseEvents()
 {
+    auto* context = new DoAfterProcessingAllPendingMouseEventsCallbackContext;
+    WKPageDoAfterProcessingAllPendingMouseEvents(m_testController->mainWebView()->page(), context, doAfterProcessingAllPendingMouseEventsCallback);
+    m_testController->runUntil(context->done, 100_ms);
+    if (context->done)
+        delete context;
+    else
+        context->timedOut = true;
 }
 
 } // namespace WTR
--- a/Source/WebKit/UIProcess/API/C/WKPage.cpp
+++ b/Source/WebKit/UIProcess/API/C/WKPage.cpp
@@ -3527,3 +3527,10 @@ void WKPageClearBackForwardCache(WKPageRef page)
     RefPtr protectedPage = toProtectedImpl(page);
     protect(protectedPage->backForwardCache())->removeEntriesForPage(*protectedPage);
 }
+
+void WKPageDoAfterProcessingAllPendingMouseEvents(WKPageRef page, void* context, WKPageDoAfterProcessingAllPendingMouseEventsFunction completionHandler)
+{
+    toProtectedImpl(page)->doAfterProcessingAllPendingMouseEvents([context, completionHandler] {
+        completionHandler(context);
+    });
+}
--- a/Source/WebKit/UIProcess/API/C/WKPagePrivate.h
+++ b/Source/WebKit/UIProcess/API/C/WKPagePrivate.h
@@ -237,6 +237,9 @@ WK_EXPORT void WKPageDisplayAndTrackRepaintsForTesting(WKPageRef page, void* con
 typedef void (*WKPageFindStringForTestingFunction)(bool found, void* functionContext);
 WK_EXPORT void WKPageFindStringForTesting(WKPageRef page, void* context, WKStringRef string, WKFindOptions options, unsigned maxMatchCount, WKPageFindStringForTestingFunction completionHandler);
 
+typedef void (*WKPageDoAfterProcessingAllPendingMouseEventsFunction)(void* functionContext);
+WK_EXPORT void WKPageDoAfterProcessingAllPendingMouseEvents(WKPageRef page, void* context, WKPageDoAfterProcessingAllPendingMouseEventsFunction function);
+
 #ifdef __cplusplus
 }
 #endif
--- a/Tools/WebKitTestRunner/wpe/TestControllerWPE.cpp
+++ b/Tools/WebKitTestRunner/wpe/TestControllerWPE.cpp
@@ -43,7 +43,6 @@ namespace WTR {
 
 void TestController::notifyDone()
 {
-    RunLoop::mainSingleton().stop();
 }
 
 void TestController::setHidden(bool)
@@ -68,9 +67,8 @@ void TestController::platformRunUntil(bool& done, WTF::Seconds timeout)
     class TimeoutTimer {
     public:
         TimeoutTimer(WTF::Seconds timeout, bool& timedOut)
-            : m_timer(RunLoop::mainSingleton(), "TestController::TimeoutTimer::Timer"_s, [&timedOut] {
+            : m_timer(RunLoop::mainSingleton(), "TestController::TimeoutTimer"_s, [&timedOut] {
                 timedOut = true;
-                RunLoop::mainSingleton().stop();
             })
         {
             m_timer.setPriority(G_PRIORITY_DEFAULT_IDLE);
@@ -81,8 +79,9 @@ void TestController::platformRunUntil(bool& done, WTF::Seconds timeout)
         RunLoop::Timer m_timer;
     } timeoutTimer(timeout, timedOut);
 
+    auto* mainContext = g_main_context_default();
     while (!done && !timedOut)
-        RunLoop::mainSingleton().run();
+        g_main_context_iteration(mainContext, TRUE);
 }
 
 void TestController::platformDidCommitLoadForFrame(WKPageRef, WKFrameRef)


Source/WebKit/UIProcess/API/C/gtk/WKView.cpp
```diff
@@ -73,3 +73,10 @@ GVariant* WKViewContentsOfUserInterfaceItem(WKViewRef viewRef, const char* userI
 {
     return webkitWebViewBaseContentsOfUserInterfaceItem(toImpl(viewRef), userInterfaceItem);
 }
+
+void WKViewDoAfterProcessingAllPendingMouseEvents(WKViewRef viewRef, void* context, WKViewDoAfterProcessingAllPendingMouseEventsFunction function)
+{
+    webkitWebViewBaseDoAfterProcessingAllPendingMouseEvents(toImpl(viewRef), [context, function]() {
+        function(context);
+    });
+}
```
carlosgcampos: This would be simpler if you add WKPage api instead, since in the end what we want is calling WebPageProxy::doAfterProcessingAllPendingMouseEvents(). So, if you add WKPageDoAfterProcessingAllPendingMouseEvents() we don't need all the changes in both GTK and WPE views for something that is actually cross-platform.

Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/DrawingAreaCoordinatedGraphicsGLib.cpp
```diff
@@ -704,4 +705,28 @@ bool DrawingAreaCoordinatedGraphics::shouldUseTiledBackingForFrameView(const Web
...
+{
+    if (m_isRenderingUpdateScheduled)
+        return true;
+
+    m_isRenderingUpdateScheduled = true;
+
+    m_scheduleRenderingTimer.startOneShot(m_preferredRenderingUpdateInterval);
+
+    return true;
+}
```
carlosgcampos: I'm not sure this is right, there might be a rendering update already scheduled without calling this method. If this regressed in 266411@main and we didn't implement this, why do we need to implement ir now? What happens if we just return true from this and wait for the RenderingUpdateScheduler to schedule the next rendering update?

Tools/WebKitTestRunner/gtk/TestControllerGtk.cpp
```diff
@@ -49,7 +49,6 @@ namespace WTR {
 
 void TestController::notifyDone()
 {
-    RunLoop::mainSingleton().stop();
```
carlosgcampos, line 52: Don't we need to do anything here?

Source/WebKit/UIProcess/API/C/gtk/WKView.cpp
obyknovenius: You're right about it. I've considered this approach but decided to go with `WKView` implementation because of the following reasons:

1. This is how it's implemented in Apple ports: https://github.com/WebKit/WebKit/blob/ade17394d9f35fee6901de14e429172e39b20022/Tools/WebKitTestRunner/mac/EventSenderProxy.mm#L962-L968
Note that they use `targetView` which is a view returned by previous event hit test. We don't do that for now, but I thought we may need this in the future. In GTK we could use `gtk_widget_contains` or something similar.

3. I don't know where it would be a good place to add such function. It must be defined in a private header. The best candidate is [`Source/WebKit/UIProcess/API/C/WKPagePrivate.h`](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/UIProcess/API/C/WKPagePrivate.h), but it will require approval from Apple which I'm not sure they will agree with, since they implemented it via `WKView` already. There's no common linux API for `WKPage`. So, I'd have to add it to [`Source/WebKit/UIProcess/API/C/wpe/WKPagePrivateWPE.h`](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/UIProcess/API/C/wpe/WKPagePrivateWPE.h), create `Source/WebKit/UIProcess/API/C/gtk/WKPagePrivateGTK.h` and still duplicate code. It's not much different from what we have now.

4. We may need to call this API from API tests where we don't use common C API like `WKView` or `WKPage`, so I thought it's a good idea to be able call it using GLib API. Look at this mac test as an example [`Tools/TestWebKitAPI/Tests/mac/MouseEventTests.mm`](https://github.com/WebKit/WebKit/blob/main/Tools/TestWebKitAPI/Tests/mac/MouseEventTests.mm).

In general, adding testing only API on linux is a bit messy right now and done different ways using different APIs. It would be nice to have something similar to Apple's [`Source/WebKit/UIProcess/API/Cocoa/WKWebViewPrivateForTesting.h`](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/UIProcess/API/Cocoa/WKWebViewPrivateForTesting.h).

Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/DrawingAreaCoordinatedGraphicsGLib.cpp
obyknovenius: > I'm not sure this is right, there might be a rendering update already scheduled without calling this method.

I think you mean [`DrawingArea::triggerRenderingUpdate()`](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/WebProcess/WebPage/DrawingArea.h#L138-L139) which should "cause a rendering update to happen as soon as possible" where `DrawinArea::scheduleRenderingUpdate()` simply calls it after `m_preferredRenderingUpdateInterval`. `DrawingAreaCoordinatedGraphics::triggerRenderingUpdate()` calls either `m_layerTreeHost->scheduleRenderingUpdate()` or `scheduleDisplay()`. Both methods are not reentrant. So, I think it's safe.

> If this regressed in 266411@main and we didn't implement this, why do we need to implement ir now? What happens if we just return true from this and wait for the RenderingUpdateScheduler to schedule the next rendering update?

I'm not sure how it worked before 266411@main, but now if we don't do anything in `DrawingAreaCoordinatedGraphics::scheduleRenderingUpdate()` and just return `true` move events will not be delivered at all. This is because in this case move events are not handled immediately but being deferred to the next rendering update. See https://github.com/WebKit/WebKit/blob/main/Source/WebKit/WebProcess/WebPage/WebPage.cpp#L3693-L3702. This is very similar to what Apple does: https://github.com/WebKit/WebKit/blob/main/Source/WebKit/WebProcess/WebPage/RemoteLayerTree/RemoteLayerTreeDrawingArea.mm#L557-L580

Tools/WebKitTestRunner/gtk/TestControllerGtk.cpp
obyknovenius: I don't think so. It seems like the run loop is controlled solely by `TestController::runUntil()`. This method is no-op on Apple ports. Keeping it may cause flaky timeouts.

Source/WebKit/UIProcess/API/C/gtk/WKView.cpp
carlosgcampos: > You're right about it. I've considered this approach but decided to go with `WKView` implementation because of the following reasons:
> 
> 1. This is how it's implemented in Apple ports: 
>      
>        
>          [WebKit/Tools/WebKitTestRunner/mac/EventSenderProxy.mm](https://github.com/WebKit/WebKit/blob/ade17394d9f35fee6901de14e429172e39b20022/Tools/WebKitTestRunner/mac/EventSenderProxy.mm#L962-L968)
>        
>        
>            Lines 962 to 968
>          in
>          [ade1739](/WebKit/WebKit/commit/ade17394d9f35fee6901de14e429172e39b20022)
>        
>      
>      
>        
>    
>            
>              
>               if (RetainPtr targetView = std::exchange(m_targetView, nullptr)) { 
>            
>    
>            
>              
>                   __block bool doneProcessingMouseEvents = false; 
>            
>    
>            
>              
>                   [checked_objc_cast<WKWebView>(targetView.get()) _doAfterProcessingAllPendingMouseEvents:^{ 
>            
>    
>            
>              
>                       doneProcessingMouseEvents = true; 
>            
>    
>            
>              
>                   }]; 
>            
>    
>            
>              
>                   m_testController->runUntil(doneProcessingMouseEvents, 100_ms); 
>            
>    
>            
>              
>               } 
>            
>        
>      
>    
>    
>    Note that they use `targetView` which is a view returned by previous event hit test. We don't do that for now, but I thought we may need this in the future. In GTK we could use `gtk_widget_contains` or something similar.

The equivalent to this is not adding C API for WKView, but adding glib API. But I still think that for this particular case it's a lot simpler to add WKPage API.

> 2. I don't know where it would be a good place to add such function. It must be defined in a private header. The best candidate is [`Source/WebKit/UIProcess/API/C/WKPagePrivate.h`](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/UIProcess/API/C/WKPagePrivate.h), but it will require approval from Apple which I'm not sure they will agree with, since they implemented it via `WKView` already. There's no common linux API for `WKPage`. So, I'd have to add it to [`Source/WebKit/UIProcess/API/C/wpe/WKPagePrivateWPE.h`](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/UIProcess/API/C/wpe/WKPagePrivateWPE.h), create `Source/WebKit/UIProcess/API/C/gtk/WKPagePrivateGTK.h` and still duplicate code. It's not much different from what we have now.

I don't see the problem of adding WKPage api to WKPagePrivate.h not all API there is used by all the ports.

> 3. We may need to call this API from API tests where we don't use common C API like `WKView` or `WKPage`, so I thought it's a good idea to be able call it using GLib API. Look at this mac test as an example [`Tools/TestWebKitAPI/Tests/mac/MouseEventTests.mm`](https://github.com/WebKit/WebKit/blob/main/Tools/TestWebKitAPI/Tests/mac/MouseEventTests.mm).> 
> In general, adding testing only API on linux is a bit messy right now and done different ways using different APIs. It would be nice to have something similar to Apple's [`Source/WebKit/UIProcess/API/Cocoa/WKWebViewPrivateForTesting.h`](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/UIProcess/API/Cocoa/WKWebViewPrivateForTesting.h).

But you are adding WKView API anyway. If we ever need to use that from API tests we can add private glib api for that.

Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/DrawingAreaCoordinatedGraphicsGLib.cpp
carlosgcampos: What I don't understand is why we use a timer and wait one frame time to trigger the rendering update, we don't know at what point of the frame we are, triggerRenderingUpdate is supposed to be called by the RenderingUpdateScheduler controlled by the display refresh monitor.

Source/WebKit/UIProcess/API/C/gtk/WKView.cpp
obyknovenius: I'm convinced now. I replaced `WKViewDoAfterProcessingAllPendingMouseEvents` with `WKPageDoAfterProcessingAllPendingMouseEvents()`. The patch is much simpler now.

Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/DrawingAreaCoordinatedGraphicsGLib.cpp
obyknovenius: I replaced `DrawingArea::scheduleRenderingUpdate()` with `protect(corePage())->scheduleRenderingUpdate(RenderingUpdateStep::EventTiming)`. It unifies the flow for all ports.

Source/WebKit/WebProcess/WebPage/WebPage.cpp
```diff
@@ -3690,14 +3690,15 @@ void WebPage::mouseEvent(FrameIdentifier frameID, const WebMouseEvent& mouseEven
...
-    if (shouldDeferDidReceiveEvent && drawingArea->scheduleRenderingUpdate()) {
+    if (shouldDeferDidReceiveEvent) {
         // For mousemove events where the user is only hovering (not clicking and dragging),
         // we defer sending the DidReceiveEvent() IPC message until the end of the rendering
         // update to throttle the rate of these events to the rendering update frequency.
         // This logic works in tandem with the mouse event queue in the UI process, which
         // coalesces mousemove events until the DidReceiveEvent() message is received after
         // the rendering update.
         m_deferredDidReceiveMouseEvent = { { mouseEvent.type(), handled } };
+        protect(corePage())->scheduleRenderingUpdate(RenderingUpdateStep::EventTiming);
```
whsieh: I think `EventTiming` is about something really specific (related to how event timing is surfaced through the Performance API) — is there a reason we need to specify this here?

whsieh (approved):

Source/WebKit/WebProcess/WebPage/WebPage.cpp
obyknovenius: Right. I will remove it.

Tools/WebKitTestRunner/libwpe/EventSenderProxyLibWPE.cpp
```diff
@@ -196,6 +198,11 @@ void EventSenderProxy::cancelTouchPoint(int index)
 
 void EventSenderProxy::waitForPendingMouseEvents()
 {
+    bool doneProcessingMouseEvents = false;
+    WKPageDoAfterProcessingAllPendingMouseEvents(m_testController->mainWebView()->page(), &doneProcessingMouseEvents, [](void* context) {
+        *static_cast<bool*>(context) = true;
+    });
+    m_testController->runUntil(doneProcessingMouseEvents, 100_ms);
```
darinadler: I think there’s a problem here with the timeout that may even lead to a security vulnerability. If this times out after 100ms and then this function returns _before_ `WKPageDoAfterProcessingAllPendingMouseEvents` calls the callback, and then the callback is called later, the function will write the `true` onto the stack, possibly corrupting memory. I’m not sure how to write a correct, safe version, but this one seems wrong.

Tools/WebKitTestRunner/win/EventSenderProxyWin.cpp
```diff
@@ -369,6 +370,11 @@ void EventSenderProxy::cancelTouchPoint(int)
 
 void EventSenderProxy::waitForPendingMouseEvents()
 {
+    bool doneProcessingMouseEvents = false;
+    WKPageDoAfterProcessingAllPendingMouseEvents(m_testController->mainWebView()->page(), &doneProcessingMouseEvents, [](void* context) {
+        *static_cast<bool*>(context) = true;
+    });
+    m_testController->runUntil(doneProcessingMouseEvents, 100_ms);
```
darinadler: I think there’s a problem here with the timeout that may even lead to a security vulnerability. If this times out after 100ms and then this function returns _before_ `WKPageDoAfterProcessingAllPendingMouseEvents` calls the callback, and then the callback is called later, the function will write the `true` onto the stack, possibly corrupting memory. I’m not sure how to write a correct, safe version, but this one seems wrong.

Tools/WebKitTestRunner/gtk/EventSenderProxyGtk.cpp
```diff
@@ -433,6 +434,11 @@ void EventSenderProxy::setTouchModifier(WKEventModifiers, bool)
 
 void EventSenderProxy::waitForPendingMouseEvents()
 {
+    bool doneProcessingMouseEvents = false;
+    WKPageDoAfterProcessingAllPendingMouseEvents(m_testController->mainWebView()->page(), &doneProcessingMouseEvents, [](void* context) {
+        *static_cast<bool*>(context) = true;
+    });
+    m_testController->runUntil(doneProcessingMouseEvents, 100_ms);
```
darinadler: I think there’s a problem here with the timeout that may even lead to a security vulnerability. If this times out after 100ms and then this function returns _before_ `WKPageDoAfterProcessingAllPendingMouseEvents` calls the callback, and then the callback is called later, the function will write the `true` onto the stack, possibly corrupting memory. I’m not sure how to write a correct, safe version, but this one seems wrong.

darinadler (changes requested):

Tools/WebKitTestRunner/libwpe/EventSenderProxyLibWPE.cpp
obyknovenius: We now allocate a context object on the heap and delete it either on a callback or after the timeout has reached.

Tools/WebKitTestRunner/gtk/EventSenderProxyGtk.cpp
obyknovenius: We now allocate a context object on the heap and delete it either on a callback or after the timeout has reached.

Source/WebKit/WebProcess/WebPage/WebPage.cpp
darinadler, line 3701: The old code used to look at the return value from this function. Is it truly best to set `m_deferredDidReceiveMouseEvent` and return early even if `scheduleRenderingUpdate` returns `false`? I’m assuming the answer is yes and we have appropriate test coverage.

Tools/WebKitTestRunner/libwpe/EventSenderProxyLibWPE.cpp
darinadler: We didn’t use `Box` so we shouldn’t add this include.
```suggestion
```

Tools/WebKitTestRunner/win/EventSenderProxyWin.cpp
darinadler: We didn’t use `RefCountable` so we shouldn’t add this include.
```suggestion
```

darinadler (approved):

Source/WebKit/WebProcess/WebPage/WebPage.cpp
obyknovenius: `Page::scheduleRenderingUpdateInternal()` schedules rendering update either via `DrawingArea::scheduleRenderingUpdate()` if it returns `true` or via `RenderingUpdateScheduler::scheduleRenderingUpdate()` otherwise.

aperezdc:
Backported into `webkitglib/2.50` as commit 753555622f9c997cbb4b1d6d0cefb9f7259c0fe5

PR #57697, opened 2026-02-02
[WPE] Web Inspector: implement platformSave

spenap:
#### 657ee259022b8d9de156d21aaf0ef08ca3e0ae30
<pre>
[WPE] Web Inspector: implement platformSave
<a href="https://bugs.webkit.org/show_bug.cgi?id=306768">https://bugs.webkit.org/show_bug.cgi?id=306768</a>

Reviewed by NOBODY (OOPS!).

Extracted a common GIO-based implementation for WebKitGTK and WPE WebKit
into WebInspectorUIProxyGLib.

For WebKitGTK, the logic keeps using file chooser dialog to get an
appropriate destination filename.

For WPE, this uses G_USER_DIRECTORY_DOWNLOAD and falls back to the
user&apos;s home directory if undefined, saving the file there with the first
8 elements of the contents hash to avoid accidental overwrites.

* Source/WebKit/SourcesGTK.txt:
* Source/WebKit/SourcesWPE.txt:
* Source/WebKit/UIProcess/Inspector/glib/WebInspectorUIProxyGLib.cpp: Added.
(WebKit::fileReplaceContentsCallback):
(WebKit::platformSaveDataToFile):
* Source/WebKit/UIProcess/Inspector/glib/WebInspectorUIProxyGLib.h:
* Source/WebKit/UIProcess/Inspector/gtk/WebInspectorUIProxyGtk.cpp:
(WebKit::WebInspectorUIProxy::platformSave):
(WebKit::fileReplaceContentsCallback): Deleted.
* Source/WebKit/UIProcess/Inspector/wpe/WebInspectorUIProxyWPE.cpp:
(WebKit::computeContentHash):
(WebKit::WebInspectorUIProxy::platformSave):
* Source/WebKit/WebProcess/Inspector/wpe/WebInspectorUIWPE.cpp:
(WebKit::WebInspectorUI::canSave):
</pre>

### DIFF
--- a/Source/WebKit/UIProcess/Inspector/glib/WebInspectorUIProxyGLib.cpp
+++ b/Source/WebKit/UIProcess/Inspector/glib/WebInspectorUIProxyGLib.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 Igalia S.L.
+ * Copyright (C) 2026 Igalia S.L.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -24,50 +24,44 @@
  */
 
 #include "config.h"
-#include "WebInspectorUI.h"
-
-#if ENABLE(WPE_PLATFORM)
+#include "WebInspectorUIProxy.h"
 
 #include <gio/gio.h>
-#include <wtf/glib/GResources.h>
+#include <wtf/glib/GRefPtr.h>
 #include <wtf/glib/GUniquePtr.h>
+#include <wtf/text/Base64.h>
 
 namespace WebKit {
 
-void WebInspectorUI::didEstablishConnection()
+static void fileReplaceContentsCallback(GObject* sourceObject, GAsyncResult* result, gpointer)
 {
-    WTF::registerInspectorResourceIfNeeded();
+    GFile* file = G_FILE(sourceObject);
+    g_file_replace_contents_finish(file, result, nullptr, nullptr);
 }
 
-bool WebInspectorUI::canSave(InspectorFrontendClient::SaveMode saveMode)
+void WebInspectorUIProxy::platformSaveDataToFile(GRefPtr<GFile>&& file, const String& content, bool base64Encoded)
 {
-    switch (saveMode) {
-    case InspectorFrontendClient::SaveMode::SingleFile:
-        return true;
+    GUniqueOutPtr<GError> error;
 
-    case InspectorFrontendClient::SaveMode::FileVariants:
-        return false;
-    }
+    Vector<uint8_t> dataVector;
+    CString dataString;
+    if (base64Encoded) {
+        auto decodedData = base64Decode(content);
+        if (!decodedData)
+            return;
+        decodedData->shrinkToFit();
+        dataVector = WTF::move(*decodedData);
+    } else
+        dataString = content.utf8();
 
-    ASSERT_NOT_REACHED();
-    return false;
-}
+    const char* data = !dataString.isNull() ? dataString.data() : reinterpret_cast<const char*>(dataVector.span().data());
+    size_t dataLength = !dataString.isNull() ? dataString.length() : dataVector.size();
 
-bool WebInspectorUI::canLoad()
-{
-    return false;
-}
+    g_file_replace_contents_async(file.get(), data, dataLength, nullptr, false,
+        G_FILE_CREATE_NONE, nullptr, fileReplaceContentsCallback, protect(inspectorPage()).get());
 
-bool WebInspectorUI::canPickColorFromScreen()
-{
-    return false;
-}
-
-String WebInspectorUI::localizedStringsURL() const
-{
-    return "resource:///org/webkit/inspector/Localizations/en.lproj/localizedStrings.js"_s;
+    if (error)
+        g_warning("Failed to save inspector data: %s", error->message);
 }
 
 } // namespace WebKit
-
-#endif // ENABLE(WPE_PLATFORM)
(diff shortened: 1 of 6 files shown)


Source/WebKit/UIProcess/Inspector/glib/WebInspectorUIProxyGLib.cpp
```diff
@@ -0,0 +1,67 @@
...
+static void fileReplaceContentsCallback(GObject* sourceObject, GAsyncResult* result, gpointer)
+{
+    GFile* file = G_FILE(sourceObject);
+    g_file_replace_contents_finish(file, result, nullptr, nullptr);
+}
+
+void WebInspectorUIProxy::platformSaveDataToFile(GRefPtr<GFile>&& file, const String& content, bool base64Encoded)
+{
+    GUniqueOutPtr<GError> error;
+
```
aperezdc, line 45: The `error` variable is never assigned, so it makes no effect. Remove it from here.

I suppose the intention was to check whether the save operation was completed successfully. If that is the case, move it into `fileReplaceContentsCallback()` above, pass it to `g_file_replace_contents_finish()` and print the error message there.

Source/WebKit/UIProcess/Inspector/glib/WebInspectorUIProxyGLib.cpp
```diff
@@ -0,0 +1,67 @@
...
+        auto decodedData = base64Decode(content);
+        if (!decodedData)
+            return;
+        decodedData->shrinkToFit();
+        dataVector = WTF::move(*decodedData);
+    } else
+        dataString = content.utf8();
+
+    const char* data = !dataString.isNull() ? dataString.data() : reinterpret_cast<const char*>(dataVector.span().data());
+    size_t dataLength = !dataString.isNull() ? dataString.length() : dataVector.size();
```
aperezdc, line 58: We do not know what's the lifetime of `content` because we do not take its ownership. That means it may be freed (or `dataString` may be freed) while the async save operation is ongoing.

I recommend using `WEBKIT_DEFINE_ASYNC_DATA_STRUCT` to declare a helper type that holds either `content` or the Base64-decoded, then pass an instance as the userdata to keep the data being saved alive during async processing, and free it in the callback.

Source/WebKit/UIProcess/Inspector/glib/WebInspectorUIProxyGLib.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+    g_file_replace_contents_async(file.get(), data, dataLength, nullptr, false,
+        G_FILE_CREATE_NONE, nullptr, fileReplaceContentsCallback, protect(inspectorPage()).get());
```
aperezdc, line 61: There is no need to pass the `inspectorPage()` pointer to the completion callback. Better pass an async data struct as suggested above.

Source/WebKit/UIProcess/Inspector/glib/WebInspectorUIProxyGLib.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+    if (error)
+        g_warning("Failed to save inspector data: %s", error->message);
```
aperezdc, line 64: Remove.

Source/WebKit/UIProcess/Inspector/glib/WebInspectorUIProxyGLib.cpp
aperezdc: (I know that the existing code already has this issue... which means it worked by chance. Let's improve it while at it!)

Source/WebKit/UIProcess/Inspector/wpe/WebInspectorUIProxyWPE.cpp
```diff
@@ -230,9 +230,25 @@ void WebInspectorUIProxy::platformShowCertificate(const WebCore::CertificateInfo
...
+    URL url { saveDatas[0].url };
+    auto* filename = url.path().substring(1).utf8().data();
+
+    const gchar* downloadsDir = g_get_user_special_dir(G_USER_DIRECTORY_DOWNLOAD);
+    if (!downloadsDir) {
+        // If we don't have XDG user dirs info, set just to HOME.
+        downloadsDir = g_get_home_dir();
+    }
+
+    GRefPtr<GFile> file = adoptGRef(g_file_new_build_filename(downloadsDir, filename, nullptr));
```
aperezdc, line 250: This should probably check whether the file already exists, and if that's the case try adding a numeric suffix to it.

Another alternative is to always add a few characters (around 12, like Git does for abbreviated commit identifiers) of the SHA256 checksum of the saved data contents; and in this way if a file is replaced, it means the contents were the same already on disk and it's not too bad to use the same file name.

aperezdc (changes requested):
I think this still needs some work, please take a look at the comments.

Source/WebKit/UIProcess/Inspector/wpe/WebInspectorUIProxyWPE.cpp
aperezdc: Or, to keep this patch smaller and focused, you may want to add the handling for existing files in a follow-up patch.

Source/WebKit/UIProcess/Inspector/glib/WebInspectorUIProxyGLib.cpp
spenap: I think that's why the code was calling protect, but I agree with you that creating a helper type is cleaner and explicit, and future proof vs refactoring. Thanks! I'll make the change.

PR #57739, opened 2026-02-03
[GFC] Use StyleSelfAlignmentData machinery for alignment.

sammygill:
#### ca88a69658d65166ef83b19e593252e139c5fc53
<pre>
[GFC] Use StyleSelfAlignmentData machinery for alignment.
<a href="https://bugs.webkit.org/show_bug.cgi?id=306817">https://bugs.webkit.org/show_bug.cgi?id=306817</a>
<a href="https://rdar.apple.com/169487049">rdar://169487049</a>

Reviewed by Elika Etemad.

StyleSelfAlignmentData provides a helper in the form of adjustmentFromStartEdge
that helps with the offset needed for alignment. Instead of enumerating the
various alignment values and duplicating the exact same logic that is
found here, we can just hook into it instead by passing in the
appropriate information.

* Source/WebCore/layout/formattingContexts/grid/GridFormattingContext.cpp:
(WebCore::Layout::GridFormattingContext::constructPlacedGridItems const):
We need to pass in the alignment subject&apos;s writing mode to the API so we
store this on the PlacedGridItem when we construct it so that later we
can access it during alignment.

* Source/WebCore/layout/formattingContexts/grid/GridFormattingContext.h:
(WebCore::Layout::GridFormattingContext::writingMode const):
It also needs the alignment container&apos;s writing mode so add a function
on the formatting context that can be accessed at the same time as the
alignment subject&apos;s.

* Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp:
(WebCore::Layout::GridLayout::layout):
(WebCore::Layout::GridLayout::performInlineAxisSelfAlignment):
(WebCore::Layout::GridLayout::performBlockAxisSelfAlignment):
Add in the border box sizes of the items, the formatting context writing
mode, and the computed gap value since we will need this for alignment.
Besides that, all we need to do is compute the remaining space for
alignment by taking the grid item&apos;s margin box and subtracting it from
the grid area&apos;s size in the appropriate dimension. Then, we should have
all of the information needed to call StyleSelfAlignmentData::adjustmentFromStartEdge.

Canonical link: <a href="https://commits.webkit.org/307365@main">https://commits.webkit.org/307365@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/0b3233c9bc4e7bfe9c523e049f68abbefa506f50

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/144109 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/16788 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/8342 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/152779 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/97348 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/bafeb31e-266d-4127-8d1c-22f970bc11cc) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/145984 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/17270 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/16682 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/110813 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/79638 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/147072 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/13229 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/129473 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/91731 "Passed tests") | | [❌ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/c58fd034-e85d-4ce1-bd96-9c36f6572695) 
| | [⏳ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/iOS-26-Simulator-WPT-WK2-Tests-EWS "Waiting to run tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/10422 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/225 "Built successfully") | | 
| | [⏳ 🧪 api-ios](https://ews-build.webkit.org/#/builders/API-Tests-iOS-Simulator-EWS "Waiting to run tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/6126 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/155091 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/16640 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/7181 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/118827 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/16676 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/13972 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/119185 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30566 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/15070 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/127337 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/72061 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/16262 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/5775 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/15996 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/80041 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/16207 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/16062 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/layout/formattingContexts/grid/GridFormattingContext.cpp
+++ b/Source/WebCore/layout/formattingContexts/grid/GridFormattingContext.cpp
@@ -223,13 +223,10 @@ PlacedGridItems GridFormattingContext::constructPlacedGridItems(const GridAreas&
 {
     PlacedGridItems placedGridItems;
     placedGridItems.reserveInitialCapacity(gridAreas.size());
-    CheckedRef gridContainerStyle = root().style();
+    CheckedRef formattingContextStyle = root().style();
     for (auto [ unplacedGridItem, gridAreaLines ] : gridAreas) {
 
         CheckedRef gridItemStyle = unplacedGridItem.m_layoutBox->style();
-
-        auto usedJustifySelf = gridItemStyle->justifySelf().resolve(gridContainerStyle.ptr());
-        auto usedAlignSelf = gridItemStyle->alignSelf().resolve(gridContainerStyle.ptr());
 
         ComputedSizes inlineAxisSizes {
             gridItemStyle->width(),
@@ -248,9 +245,8 @@ PlacedGridItems GridFormattingContext::constructPlacedGridItems(const GridAreas&
         };
 
         auto& boxGeometry = geometryForGridItem(unplacedGridItem.m_layoutBox);
-        placedGridItems.constructAndAppend(unplacedGridItem, gridAreaLines, inlineAxisSizes, blockAxisSizes,
-            boxGeometry.horizontalBorderAndPadding(), boxGeometry.verticalBorderAndPadding(), usedJustifySelf,
-            usedAlignSelf, gridItemStyle->usedZoomForLength());
+        placedGridItems.constructAndAppend(unplacedGridItem, gridAreaLines, inlineAxisSizes, blockAxisSizes, boxGeometry.horizontalBorderAndPadding(), boxGeometry.verticalBorderAndPadding(),
+            gridItemStyle->justifySelf().resolve(formattingContextStyle.ptr()), gridItemStyle->alignSelf().resolve(formattingContextStyle.ptr()), gridItemStyle->writingMode(), gridItemStyle->usedZoomForLength());
     }
     return placedGridItems;
 }
--- a/Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
+++ b/Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
@@ -231,78 +231,64 @@ std::pair<UsedTrackSizes, GridItemRects> GridLayout::layout(UnplacedGridItems& u
     // https://drafts.csswg.org/css-grid-1/#alignment
     // After a grid container’s grid tracks have been sized, and the dimensions of all grid items
     // are finalized, grid items can be aligned within their grid areas.
-    auto inlineAxisPositions = performInlineAxisSelfAlignment(placedGridItems, usedInlineMargins, gridAreaSizes.inlineSizes);
-    auto blockAxisPositions = performBlockAxisSelfAlignment(placedGridItems, usedBlockMargins, gridAreaSizes.blockSizes);
+    auto inlineAxisPositions = performInlineAxisSelfAlignment(placedGridItems, usedInlineMargins, usedInlineSizes, gridAreaSizes.inlineSizes);
+    auto blockAxisPositions = performBlockAxisSelfAlignment(placedGridItems, usedBlockMargins, usedBlockSizes, gridAreaSizes.blockSizes);
 
     auto gridItemRects = computeGridItemRects(placedGridItems, inlineAxisPositions, blockAxisPositions, usedInlineSizes, usedBlockSizes, usedInlineMargins, usedBlockMargins);
 
     return { usedTrackSizes, gridItemRects };
 }
 
-BorderBoxPositions GridLayout::performInlineAxisSelfAlignment(const PlacedGridItems& placedGridItems, const Vector<UsedMargins>& inlineMargins, const Vector<LayoutUnit>& gridAreaInlineSizes)
+BorderBoxPositions GridLayout::performInlineAxisSelfAlignment(const PlacedGridItems& placedGridItems, const Vector<UsedMargins>& inlineMargins, const UsedInlineSizes& borderBoxSizes,
+    const Vector<LayoutUnit>& gridAreasInlineSizeList)
 {
-    UNUSED_PARAM(gridAreaInlineSizes);
     BorderBoxPositions borderBoxPositions;
     borderBoxPositions.reserveInitialCapacity(placedGridItems.size());
 
-    auto computeMarginBoxPosition = [](const PlacedGridItem& placedGridItem) -> LayoutUnit {
-        switch (placedGridItem.inlineAxisAlignment().position()) {
-        case ItemPosition::FlexStart:
-        case ItemPosition::SelfStart:
-        case ItemPosition::Start:
-            return { };
+    auto& formattingContextWritingMode = formattingContext().writingMode();
+    for (size_t gridItemIndex = 0; gridItemIndex < placedGridItems.size(); ++gridItemIndex) {
+        auto& gridItem = placedGridItems[gridItemIndex];
 
+        auto& [marginStart, marginEnd] = inlineMargins[gridItemIndex];
+        auto marginBoxSize = marginStart + borderBoxSizes[gridItemIndex] + marginEnd;
+        auto remainingSpace = gridAreasInlineSizeList[gridItemIndex] - marginBoxSize;
+
+        // Normal behavior:
         // https://www.w3.org/TR/css-align-3/#justify-grid
         // Sizes as either stretch (typical non-replaced elements) or start (typical replaced elements);
         // see Grid Item Sizing in [CSS-GRID-1]. The resulting box is then start-aligned.
         //
         // Stretching should be handled by GridLayout::layoutGridItems.
-        case ItemPosition::Normal:
-            return { };
-        default:
-            ASSERT_NOT_IMPLEMENTED_YET();
-            return { };
-        }
-    };
+        auto marginBoxPosition = StyleSelfAlignmentData::adjustmentFromStartEdge(remainingSpace, gridItem.inlineAxisAlignment().position(), LogicalBoxAxis::Inline, formattingContextWritingMode, gridItem.writingMode());
 
-    for (size_t gridItemIndex = 0; gridItemIndex < placedGridItems.size(); ++gridItemIndex) {
-        auto& gridItem = placedGridItems[gridItemIndex];
-        auto marginBoxPosition = computeMarginBoxPosition(gridItem);
         borderBoxPositions.append(marginBoxPosition + inlineMargins[gridItemIndex].marginStart);
     }
 
     return borderBoxPositions;
 }
 
-BorderBoxPositions GridLayout::performBlockAxisSelfAlignment(const PlacedGridItems& placedGridItems, const Vector<UsedMargins>& blockMargins, const Vector<LayoutUnit>& gridAreaBlockSizes)
+BorderBoxPositions GridLayout::performBlockAxisSelfAlignment(const PlacedGridItems& placedGridItems, const Vector<UsedMargins>& blockMargins, const UsedBlockSizes& borderBoxSizes,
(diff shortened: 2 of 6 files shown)


Source/WebCore/layout/formattingContexts/grid/GridFormattingContext.cpp
```diff
@@ -250,7 +250,8 @@ PlacedGridItems GridFormattingContext::constructPlacedGridItems(const GridAreas&
             gridItemStyle->marginBottom()
         };
 
-        placedGridItems.constructAndAppend(unplacedGridItem, gridAreaLines, inlineAxisSizes, blockAxisSizes, usedJustifySelf(), usedAlignSelf(), gridItemStyle->usedZoomForLength());
+        placedGridItems.constructAndAppend(unplacedGridItem, gridAreaLines, inlineAxisSizes, blockAxisSizes, usedJustifySelf(), usedAlignSelf(),
```
fantasai: Would it make sense for the constructor to handle the usedJustify/AlignSelf() extraction so you don't have to do pass it around?

Also you shouldn't need an anonymous function here, just call `gridItemStyle.alignSelf().resolve(gridContainerStyle)`.

Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
```diff
@@ -235,81 +235,70 @@ std::pair<UsedTrackSizes, GridItemRects> GridLayout::layout(const GridFormatting
...
 
     return { usedTrackSizes, gridItemRects };
 }
 
-BorderBoxPositions GridLayout::performInlineAxisSelfAlignment(const PlacedGridItems& placedGridItems, const Vector<UsedMargins>& inlineMargins, const Vector<LayoutUnit>& gridAreaInlineSizes)
+BorderBoxPositions GridLayout::performInlineAxisSelfAlignment(const PlacedGridItems& placedGridItems, const Vector<UsedMargins>& inlineMargins, const UsedInlineSizes& borderBoxSizes,
+    const Vector<LayoutUnit>& gridAreaInlineSizes, const Style::GapGutter& computedColumnGutter, const WritingMode& formattingContextWritingMode)
 {
-    UNUSED_PARAM(gridAreaInlineSizes);
+    auto usedColumnGutter = GridLayoutUtils::computeGapValue(computedColumnGutter);
```
fantasai: Still getting used to LFC stuff... the formatting context gets torn down after we're done with layout, right? In that case, maybe cache the gutter on the formatting context so you only have to compute it once and you don't have to pass it around everywhere as an independent parameter?

Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
```diff
@@ -235,81 +235,70 @@ std::pair<UsedTrackSizes, GridItemRects> GridLayout::layout(const GridFormatting
...
+    auto blockAxisPositions = performBlockAxisSelfAlignment(placedGridItems, usedBlockMargins, usedBlockSizes, gridAreaSizes.blockSizes, rowGap, formattingContextWritingMode);
 
     auto gridItemRects = computeGridItemRects(placedGridItems, inlineAxisPositions, blockAxisPositions, usedInlineSizes, usedBlockSizes, usedInlineMargins, usedBlockMargins);
 
     return { usedTrackSizes, gridItemRects };
 }
 
-BorderBoxPositions GridLayout::performInlineAxisSelfAlignment(const PlacedGridItems& placedGridItems, const Vector<UsedMargins>& inlineMargins, const Vector<LayoutUnit>& gridAreaInlineSizes)
+BorderBoxPositions GridLayout::performInlineAxisSelfAlignment(const PlacedGridItems& placedGridItems, const Vector<UsedMargins>& inlineMargins, const UsedInlineSizes& borderBoxSizes,
+    const Vector<LayoutUnit>& gridAreaInlineSizes, const Style::GapGutter& computedColumnGutter, const WritingMode& formattingContextWritingMode)
```
fantasai: Why pass the WritingMode? Can't we just get it off the formattingContext() right here?

Source/WebCore/layout/formattingContexts/grid/GridFormattingContext.cpp
sammygill: Yeah that is an option. I want to keep the constructor as simple as possible I suppose which is why I just have it assigning the passed in arguments

sammygill: Removed the lambdas!

Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
sammygill: Yeah the integration code creates the FormattingContext which creates GridLayout. Avoiding recomputing the gap was something I planned on doing just haven't gotten around to doing it yet....

Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
sammygill: They are declared as static files in the header (which I suppose they don't need to be declared that way)

Source/WebCore/layout/formattingContexts/grid/GridFormattingContext.cpp
fantasai: I think keeping the parameters simpler is probably better for code readability... and also the more you can encapsulate the object's own logic within the object, the easier it is to follow what other methods are doing _to_ the object.

In this case, pulling data out of the gridItem and caching it in the placedGridItem seems like a pretty reasonable thing for a PlacedGridItem constructor to do, imho.

Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
fantasai: Not sure what that means, but if we have access to a bit of data already available, passing it in as a parameter seems like unnecessary complexity. The only reason to do that would be if it's expensive to compute and we're doing it a lot; but afaict that's not a problem here.

Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
```diff
@@ -235,81 +235,70 @@ std::pair<UsedTrackSizes, GridItemRects> GridLayout::layout(const GridFormatting
...
 
-    auto computeMarginBoxPosition = [](const PlacedGridItem& placedGridItem) -> LayoutUnit {
-        switch (placedGridItem.inlineAxisAlignment().position()) {
-        case ItemPosition::FlexStart:
-        case ItemPosition::SelfStart:
-        case ItemPosition::Start:
-            return { };
+    for (size_t gridItemIndex = 0; gridItemIndex < placedGridItems.size(); ++gridItemIndex) {
+        auto& gridItem = placedGridItems[gridItemIndex];
+        auto gridAreaSize = GridLayoutUtils::gridAreaDimensionSize(gridItem.columnStartLine(), gridItem.columnEndLine(), gridAreaInlineSizes, usedColumnGutter);
```
yulunwu2: we have

    auto gridAreaSizes = computeGridAreaSizes(placedGridItems, columnGap, rowGap, usedTrackSizes);

on line 227, do we need to recompute this?

yulunwu2 (commented):
overall lgtm, question about the need to computing grid sizes again(?)

Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
sammygill: Changed to calling `formattingContext().writingMode()`!

FWIW generally what I was trying to do is create a collection of static/free functions that just take in whatever they need for their computation as arguments as much as possible. IMO it makes it a bit easier to reason about what a function needs in order to do its job instead of just randomly picking things off of random and potentially complex structures.

Not a big deal here since we were already doing it one level above but that's just generally the approach I've been working with! Happy to have more in depth discussions though! :D

Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
sammygill: Nope! should be fixed now

fantasai (approved):
Let’s follow up on some of the parameter style questions next week, but looks good enough to land. :)

PR #58106, opened 2026-02-07
Replace `RenderMenuList` type checks with `usesMenuList()`

nt1m:
#### 3d5917256979e97640e626bebabbd1088362b23f
<pre>
Replace `RenderMenuList` type checks with `usesMenuList()`
<a href="https://bugs.webkit.org/show_bug.cgi?id=307226">https://bugs.webkit.org/show_bug.cgi?id=307226</a>
<a href="https://rdar.apple.com/169860596">rdar://169860596</a>

Reviewed by Darin Adler.

We want to eventually remove `RenderMenuList`, replace the renderer type checks with the equivalent `usesMenuList()` DOM method.

No behavior change.

* Source/WebCore/html/HTMLSelectElement.cpp:
(WebCore::HTMLSelectElement::platformHandleKeydownEvent):
(WebCore::HTMLSelectElement::menuListDefaultEventHandler):
(WebCore::HTMLSelectElement::defaultEventHandler):
(WebCore::HTMLSelectElement::showPopup):
(WebCore::HTMLSelectElement::hostWindow const): Windows-only method for PopupMenuClient, which is only used for menulists.

Canonical link: <a href="https://commits.webkit.org/307038@main">https://commits.webkit.org/307038@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/3ce36f61bfd46bd90fe059cbf4f2ef2ad73d71ff

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/143182 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/15657 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/6651 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/151858 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/96403 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/cf059407-c2cd-4955-81c4-7b63825abdf6) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/145049 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/16317 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/15738 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/110111 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/79262 "Passed tests") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/6f79f023-b7ba-4657-8828-ff72699dee31) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/146131 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/12548 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/128131 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/91022 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/680dab9c-8b9e-49ee-9d50-07021ef1092d) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/12031 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/9744 "Passed tests") | [✅ 🛠 wpe-libwebrtc](https://ews-build.webkit.org/#/builders/172/builds/1855 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/121448 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/4628 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/154169 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/15661 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/5614 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/118130 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/15666 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/13232 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/118470 "Passed tests") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/14396 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/125815 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/71052 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/22076 "Built successfully and passed tests") | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/15326 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/4456 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/15060 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/79045 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/15271 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/15122 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/html/HTMLSelectElement.cpp
+++ b/Source/WebCore/html/HTMLSelectElement.cpp
@@ -1245,7 +1245,7 @@ bool HTMLSelectElement::platformHandleKeydownEvent(KeyboardEvent* event)
             // Calling focus() may cause us to lose our renderer. Return true so
             // that our caller doesn't process the event further, but don't set
             // the event as handled.
-            if (!is<RenderMenuList>(renderer()))
+            if (!renderer() || !usesMenuList())
                 return true;
 
             // Save the selection so it can be compared to the new selection
@@ -1267,7 +1267,7 @@ bool HTMLSelectElement::platformHandleKeydownEvent(KeyboardEvent* event)
 void HTMLSelectElement::menuListDefaultEventHandler(Event& event)
 {
     ASSERT(renderer());
-    ASSERT(renderer()->isRenderMenuList());
+    ASSERT(usesMenuList());
 
     auto& eventNames = WebCore::eventNames();
     if (event.type() == eventNames.keydownEvent) {
@@ -1343,7 +1343,7 @@ void HTMLSelectElement::menuListDefaultEventHandler(Event& event)
                 protect(document())->updateStyleIfNeeded();
 
                 // Calling focus() may remove the renderer or change the renderer type.
-                if (!is<RenderMenuList>(renderer()))
+                if (!renderer() || !usesMenuList())
                     return;
 
                 // Save the selection so it can be compared to the new selection
@@ -1360,7 +1360,7 @@ void HTMLSelectElement::menuListDefaultEventHandler(Event& event)
                 protect(document())->updateStyleIfNeeded();
 
                 // Calling focus() may remove the renderer or change the renderer type.
-                if (!is<RenderMenuList>(renderer()))
+                if (!renderer() || !usesMenuList())
                     return;
 
                 // Save the selection so it can be compared to the new selection
@@ -1387,7 +1387,7 @@ void HTMLSelectElement::menuListDefaultEventHandler(Event& event)
 #if !PLATFORM(IOS_FAMILY)
         protect(document())->updateStyleIfNeeded();
 
-        if (is<RenderMenuList>(renderer())) {
+        if (renderer() && usesMenuList()) {
             ASSERT(!m_popupIsVisible);
             // Save the selection so it can be compared to the new
             // selection when we call onChange during selectOption,
@@ -1667,7 +1667,7 @@ void HTMLSelectElement::defaultEventHandler(Event& event)
         return;
     }
 
-    if (is<RenderMenuList>(renderer))
+    if (usesMenuList())
         menuListDefaultEventHandler(event);
     else 
         listBoxDefaultEventHandler(event);
@@ -1777,8 +1777,8 @@ void HTMLSelectElement::showPopup()
     if (m_popupIsVisible)
         return;
 
-    CheckedPtr renderer = dynamicDowncast<RenderMenuList>(this->renderer());
-    if (!renderer)
+    CheckedPtr renderer = this->renderer();
+    if (!renderer || !usesMenuList())
         return;
 
     RefPtr frame = document().frame();
@@ -2042,8 +2042,8 @@ FontSelector* HTMLSelectElement::fontSelector() const
 
 HostWindow* HTMLSelectElement::hostWindow() const
 {
-    if (CheckedPtr renderer = dynamicDowncast<RenderMenuList>(this->renderer()))
-        return renderer->hostWindow();
+    if (renderer() && usesMenuList())
+        return renderer()->hostWindow();
     return nullptr;
 }
 #endif


Source/WebCore/html/HTMLSelectElement.cpp
```diff
@@ -1343,7 +1343,7 @@ void HTMLSelectElement::menuListDefaultEventHandler(Event& event)
                 protect(document())->updateStyleIfNeeded();
 
                 // Calling focus() may remove the renderer or change the renderer type.
-                if (!is<RenderMenuList>(renderer()))
+                if (!renderer())
                     return;
```
darinadler, line 1347: The part of this comment that says “or change the renderer type” no longer makes sense, so at least we need to fix the comment. Even if it’s no longer the “renderer type” is there some reason to check that the element did not change from a menu to a list halfway through this handler? Maybe we need to add a check that `usesMenuList()` is still true?

Source/WebCore/html/HTMLSelectElement.cpp
```diff
@@ -1360,7 +1360,7 @@ void HTMLSelectElement::menuListDefaultEventHandler(Event& event)
...
```
darinadler, line 1364: The part of this comment that says “or change the renderer type” no longer makes sense, so at least we need to fix the comment. Even if it’s no longer the “renderer type” is there some reason to check that the element did not change from a menu to a list halfway through this handler? Maybe we need to add a check that `usesMenuList()` is still true?

Source/WebCore/html/HTMLSelectElement.cpp
```diff
@@ -1387,7 +1387,7 @@ void HTMLSelectElement::menuListDefaultEventHandler(Event& event)
 #if !PLATFORM(IOS_FAMILY)
         protect(document())->updateStyleIfNeeded();
 
-        if (is<RenderMenuList>(renderer())) {
+        if (renderer()) {
```
darinadler: The old code was not just checking for a null renderer, but also if it’s still a menu list. Maybe we need to add a check that `usesMenuList()` is still true? Or maybe that was not an important case to handle specially?

Source/WebCore/html/HTMLSelectElement.cpp
```diff
@@ -1245,7 +1245,7 @@ bool HTMLSelectElement::platformHandleKeydownEvent(KeyboardEvent* event)
             // Calling focus() may cause us to lose our renderer. Return true so
             // that our caller doesn't process the event further, but don't set
             // the event as handled.
-            if (!is<RenderMenuList>(renderer()))
+            if (!renderer())
                 return true;
```
darinadler, line 1249: The old code was not just checking for a null renderer, but also if it’s still a menu list. Maybe we need to add a check that `usesMenuList()` is still true? Or maybe that was not an important case to handle specially?

darinadler (changes requested):
Good to prepare to get rid of the special renderer class. But I don’t see enough rationale for whether we need to handle cases where the type of select element changes from a menu to a list box partway in surprising places, or where we handled the menu and list box cases differently before. A more conservative approach would be to include more calls to `usesMenuList()` to more exactly match the old behavior. Or if we know those checks weren’t really important, then we can just remove them.

I feel like the refactor to not rely on the renderer class to indicate the type and the removal of unnecessary checks are two separate tasks, and maybe could be done independently rather than doing a mix of both at the same time?

Source/WebCore/html/HTMLSelectElement.cpp
```diff
@@ -2042,9 +2042,7 @@ FontSelector* HTMLSelectElement::fontSelector() const
 
 HostWindow* HTMLSelectElement::hostWindow() const
 {
-    if (CheckedPtr renderer = dynamicDowncast<RenderMenuList>(this->renderer()))
-        return renderer->hostWindow();
-    return nullptr;
+    return renderer() && renderer()->hostWindow();
```
annevk: This is wrong and the Windows bot shows it.

nt1m: Fixed!

Source/WebCore/html/HTMLSelectElement.cpp
```diff
@@ -2042,8 +2042,8 @@ FontSelector* HTMLSelectElement::fontSelector() const
 
 HostWindow* HTMLSelectElement::hostWindow() const
 {
-    if (CheckedPtr renderer = dynamicDowncast<RenderMenuList>(this->renderer()))
-        return renderer->hostWindow();
+    if (renderer())
```
annevk: Now this is missing the `usesMenuList()` check.

nt1m: I don't think it matters particularly in this context, but I can re-add it for the sake of keeping the behavior the same.

darinadler (approved):

PR #58169, opened 2026-02-09
[WebDriver][BiDi] browsingContext.create should preserve user context metadata and emit stable top-level contextCreated events

juliopiubello:
#### b0d012c08130243439cfe6d73661e61959ecb2b1
<pre>
[WebDriver][BiDi] browsingContext.create should preserve user context metadata and emit stable top-level contextCreated events
<a href="https://bugs.webkit.org/show_bug.cgi?id=286790.">https://bugs.webkit.org/show_bug.cgi?id=286790.</a>

Reviewed by NOBODY (OOPS!).

This change completes the user-context foundation around
`browsingContext.create`. The create command now derives the effective user
context from either `userContext` or `referenceContext`, creates the new
top-level context with that user context, and reports matching `userContext`
and `clientWindow` metadata in browsing-context tree results and lifecycle
events.

It also emits the top-level `browsingContext.contextCreated` event exactly
once and before the create command returns, and normalizes nullable event
fields in the outgoing protocol payload. With create flow and event metadata
aligned, stale expectations for now-passing `create` and `context_created`
tests can be removed.

Test: Tools/Scripts/run-webdriver-tests --gtk imported/w3c/webdriver/tests/bidi/browsing_context/create imported/w3c/webdriver/tests/bidi/
browsing_context/context_created

* Source/WebKit/UIProcess/Automation/BidiBrowserAgent.cpp:
(WebKit::BidiBrowserAgent::didCreatePage):
(WebKit::BidiBrowserAgent::willClosePage):
(WebKit::BidiBrowserAgent::hasUserContext): Added.
(WebKit::BidiBrowserAgent::setUserContextIDForPage): Added.
(WebKit::BidiBrowserAgent::getUserContextIDForPage): Added.
Track per-page user-context overrides so page creation, teardown, and event
reporting agree on the effective user context.

* Source/WebKit/UIProcess/Automation/BidiBrowserAgent.h:
Add helpers and storage for per-page user-context overrides.

* Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp:
(WebKit::BidiBrowsingContextAgent::create):
(WebKit::BidiBrowsingContextAgent::getNavigableInfo):
Derive the effective user context for `browsingContext.create`, validate it,
and use page metadata when building browsing-context tree results.

* Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.h:
Pass the owning page through `getNavigableInfo`.

* Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp:
(WebKit::WebAutomationSession::createBrowsingContext):
(WebKit::WebAutomationSession::createBrowsingContextForUserContext): Added.
(WebKit::WebAutomationSession::createBrowsingContextInternal): Added.
(WebKit::WebAutomationSession::didCreatePage):
(WebKit::WebAutomationSession::emitContextCreatedEvent):
(WebKit::WebAutomationSession::getClientWindowAndUserContext): Added.
(WebKit::WebAutomationSession::getUserContextIDForPage): Added.
(WebKit::WebAutomationSession::hasUserContextID): Added.
Create pages in the requested user context, preserve pending user-context
assignments during page creation, and emit top-level `contextCreated` events
once with stable ordering.

* Source/WebKit/UIProcess/Automation/WebAutomationSession.h:
Add helpers and tracking for user-context-aware page creation and one-shot
top-level `contextCreated` emission.

(WebKit::normalizeBrowsingContextInfoEvent): Added.
Normalize internal `&quot;null&quot;` placeholders and empty children payloads into the
JSON shape expected by BiDi events.

Remove stale expectations for `browsing_context/create` and
`browsing_context/context_created` tests that now pass, and keep only the
remaining expected failures
</pre>
https://github.com/WebKit/WebKit/commit/54b2def7a4eec92958ba90fba571e1e38d0430f5

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/150584 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/23342 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/16910 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/159306 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/104018 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/23773 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/23494 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/116210 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/82550 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/153544 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/18319 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/135092 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/96938 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/17418 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/15367 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/7154 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/127032 "Passed tests") | | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/161780 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/4900 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/14569 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/124208 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/23144 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/19414 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/124406 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/23132 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/134811 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/79521 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/19495 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/11571 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/22746 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/86544 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/22458 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/22610 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/22512 "Built successfully") | | |

### DIFF
--- a/Source/WebKit/UIProcess/Automation/BidiBrowserAgent.cpp
+++ b/Source/WebKit/UIProcess/Automation/BidiBrowserAgent.cpp
@@ -56,13 +56,19 @@ String toUserContextIDProtocolString(const PAL::SessionID& sessionID)
     return builder.toString();
 }
 
+} // namespace
+
+const String& defaultClientWindowID()
+{
+    static NeverDestroyed<const String> clientWindowID(MAKE_STATIC_STRING_IMPL("unknown-window"));
+    return clientWindowID;
+}
+
 const String& defaultUserContextID()
 {
     static NeverDestroyed<const String> contextID(MAKE_STATIC_STRING_IMPL("default"));
     return contextID;
 }
-
-} // namespace
 
 struct BidiBrowserAgent::BidiUserContextDeletionRecord {
     WTF_MAKE_TZONE_ALLOCATED_INLINE(BidiUserContextDeletionRecord);
@@ -90,7 +96,7 @@ BidiBrowserAgent::~BidiBrowserAgent() = default;
 
 void BidiBrowserAgent::didCreatePage(WebPageProxy& page)
 {
-    String userContextID = toUserContextIDProtocolString(page.sessionID());
+    String userContextID = getUserContextIDForPage(page);
     auto it = m_userContextsPendingDeletion.find(userContextID);
     if (it == m_userContextsPendingDeletion.end())
         return;
@@ -101,7 +107,10 @@ void BidiBrowserAgent::didCreatePage(WebPageProxy& page)
 
 void BidiBrowserAgent::willClosePage(const WebPageProxy& page)
 {
-    String userContextID = toUserContextIDProtocolString(page.sessionID());
+    String userContextID = getUserContextIDForPage(page);
+
+    m_pageUserContextOverrides.remove(page.identifier());
+
     auto it = m_userContextsPendingDeletion.find(userContextID);
     if (it == m_userContextsPendingDeletion.end())
         return;
@@ -112,6 +121,39 @@ void BidiBrowserAgent::willClosePage(const WebPageProxy& page)
 
     it->value->callback({ });
     m_userContextsPendingDeletion.remove(it);
+}
+
+bool BidiBrowserAgent::hasUserContext(const String& userContextID) const
+{
+    if (userContextID == defaultUserContextID())
+        return true;
+
+    return m_userContexts.contains(userContextID);
+}
+
+void BidiBrowserAgent::setUserContextIDForPage(const WebPageProxy& page, const String& userContextID)
+{
+    if (userContextID.isEmpty() || userContextID == defaultUserContextID()) {
+        m_pageUserContextOverrides.remove(page.identifier());
+        return;
+    }
+
+    m_pageUserContextOverrides.set(page.identifier(), userContextID);
+}
+
+String BidiBrowserAgent::getUserContextIDForPage(const WebPageProxy& page) const
+{
+    if (auto it = m_pageUserContextOverrides.find(page.identifier()); it != m_pageUserContextOverrides.end())
+        return it->value;
+
+    String userContextID = toUserContextIDProtocolString(page.sessionID());
+
+    // During removeUserContext teardown, contexts are moved from m_userContexts to
+    // m_userContextsPendingDeletion until their pages are closed.
+    if (m_userContexts.contains(userContextID) || m_userContextsPendingDeletion.contains(userContextID))
+        return userContextID;
+
+    return defaultUserContextID();
 }
 
 // MARK: Inspector::BidiBrowserDispatcherHandler methods.
@@ -162,16 +204,20 @@ void BidiBrowserAgent::removeUserContext(const String& userContextID, CommandCal
 
     auto it = m_userContexts.find(userContextID);
     // https://www.w3.org/TR/webdriver-bidi/#command-browser-removeUserContext step 4.
-    ASYNC_FAIL_WITH_PREDEFINED_ERROR_AND_DETAILS_IF(it == m_userContexts.end(), InvalidParameter, "no such user context"_s);
+    ASYNC_FAIL_WITH_PREDEFINED_ERROR_AND_DETAILS_IF(it == m_userContexts.end(), NoSuchUserContext, "no such user context"_s);
 
     auto userContext = m_userContexts.take(it);
-    size_t pageCount = userContext->allPages().size();
+    auto pages = userContext->allPages();
+    size_t pageCount = pages.size();
     if (!pageCount) {
         callback({ });
         return;
     }
 
     m_userContextsPendingDeletion.set(userContextID, makeUnique<BidiUserContextDeletionRecord>(WTF::move(userContext), pageCount, WTF::move(callback)));
+
+    for (Ref page : pages)
+        page->closePage();
 }
 
 #if !USE(GLIB)
--- a/Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
+++ b/Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
@@ -30,6 +30,7 @@
 #if ENABLE(WEBDRIVER_BIDI)
 
 #include "AutomationProtocolObjects.h"
+#include "BidiBrowserAgent.h"
 #include "FrameTreeNodeData.h"
 #include "Logging.h"
 #include "PageLoadState.h"
@@ -124,11 +125,30 @@ void BidiBrowsingContextAgent::create(Inspector::Protocol::BidiBrowsingContext::
     RefPtr session = m_session.get();
     ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(!session, InternalError);
 
-    // FIXME: implement `referenceContext` option.
     // FIXME: implement `background` option.
-    // FIXME: implement `userContext` option.
 
-    session->createBrowsingContext(browsingContextPresentationFromCreateType(createType), [callback = WTF::move(callback)](CommandResultOf<BrowsingContext, Inspector::Protocol::Automation::BrowsingContextPresentation>&& result) {
+    String userContextID = optionalUserContext;
+
+    if (!optionalReferenceContext.isEmpty()) {
+        auto handles = session->extractBrowsingContextHandles(optionalReferenceContext);
+        ASYNC_FAIL_IF_UNEXPECTED_RESULT(handles);
+
+        auto [pageHandle, frameHandle] = handles.value();
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(!frameHandle.isEmpty(), InvalidParameter);
+
+        RefPtr referencePage = session->webPageProxyForHandle(pageHandle);
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(!referencePage, FrameNotFound);
+
+        if (userContextID.isEmpty())
+            userContextID = session->getUserContextIDForPage(*referencePage);
+    }
+
+    if (userContextID.isEmpty())
+        userContextID = defaultUserContextID();
+
+    ASYNC_FAIL_WITH_PREDEFINED_ERROR_AND_DETAILS_IF(!session->hasUserContextID(userContextID), NoSuchUserContext, "no such user context"_s);
+
+    session->createBrowsingContextForUserContext(browsingContextPresentationFromCreateType(createType), userContextID, [callback = WTF::move(callback)](CommandResultOf<BrowsingContext, Inspector::Protocol::Automation::BrowsingContextPresentation>&& result) {
         if (!result) {
             callback(makeUnexpected(result.error()));
             return;
@@ -153,17 +173,20 @@ Inspector::Protocol::BidiBrowsingContext::BrowsingContext BidiBrowsingContextAge
     return session->handleForWebFrameID(frameID);
 }
 
-Ref<Inspector::Protocol::BidiBrowsingContext::Info> BidiBrowsingContextAgent::getNavigableInfo(const WebKit::FrameTreeNodeData& tree, std::optional<uint64_t> maxDepth, IncludeParentID includeParentID)
+Ref<Inspector::Protocol::BidiBrowsingContext::Info> BidiBrowsingContextAgent::getNavigableInfo(const WebKit::FrameTreeNodeData& tree, const WebPageProxy& page, std::optional<uint64_t> maxDepth, IncludeParentID includeParentID)
 {
     // https://w3c.github.io/webdriver-bidi/#get-the-navigable-info
 
-    // FIXME: Properly support different user contexts, which will likely map to different WebAutomationSessions.
-    // https://bugs.webkit.org/show_bug.cgi?id=288104
+    RefPtr session = m_session.get();
+    auto [clientWindow, userContext] = session
+        ? session->getClientWindowAndUserContext(page)
+        : std::pair<String, String>(defaultClientWindowID(), defaultUserContextID());
+
     auto info = Inspector::Protocol::BidiBrowsingContext::Info::create()
         .setContext(getBrowsingContextID(tree.info.frameID))
         .setUrl(tree.info.request.url().string())
-        .setClientWindow("placeholder_window"_s)
-        .setUserContext("default"_s)
+        .setClientWindow(clientWindow)
+        .setUserContext(userContext)
         .setChildrenIsNull()
         .setOriginalOpenerIsNull()
         .release();
@@ -186,7 +209,7 @@ Ref<Inspector::Protocol::BidiBrowsingContext::Info> BidiBrowsingContextAgent::ge
     auto childrenInfo = JSON::ArrayOf<Inspector::Protocol::BidiBrowsingContext::Info>::create();
     auto newDepth = maxDepth ? std::optional<uint64_t>(*maxDepth - 1) : std::nullopt;
     for (auto& child : tree.children)
-        childrenInfo->addItem(getNavigableInfo(child, newDepth, IncludeParentID::No));
+        childrenInfo->addItem(getNavigableInfo(child, page, newDepth, IncludeParentID::No));
 
     info->setChildren(WTF::move(childrenInfo));
     return info;
@@ -204,7 +227,7 @@ void BidiBrowsingContextAgent::getNextTree(Vector<Ref<WebPageProxy>>&& pagesToPr
     Ref webPageProxy = pagesToProcess.takeLast();
     webPageProxy->getAllFrameTrees([this, pagesToProcess = WTF::move(pagesToProcess), resultsObject = WTF::move(resultsObject), callback = WTF::move(callback), maxDepth, protectedPage = Ref { webPageProxy }](Vector<WebKit::FrameTreeNodeData>&& trees) mutable {
         for (auto& tree : trees) {
-            auto infoTree = getNavigableInfo(tree, maxDepth, IncludeParentID::Yes);
+            auto infoTree = getNavigableInfo(tree, protectedPage.get(), maxDepth, IncludeParentID::Yes);
             resultsObject->addItem(WTF::move(infoTree));
         }
         getNextTree(WTF::move(pagesToProcess), WTF::move(resultsObject), maxDepth, WTF::move(callback));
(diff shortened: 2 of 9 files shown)


dcrousso (approved):
nice!

dcrousso:
oh also please fix the incorrect indentation in the PR description and commit message before merging :)

Source/WebKit/UIProcess/Automation/BidiBrowserAgent.cpp
```diff
@@ -114,6 +117,41 @@ void BidiBrowserAgent::willClosePage(const WebPageProxy& page)
...
+{
+    if (auto it = m_pageUserContextOverrides.find(page.identifier()); it != m_pageUserContextOverrides.end())
+        return it->value;
+
+    String sessionIDString = toUserContextIDProtocolString(page.sessionID());
+
+    // During removeUserContext teardown, contexts are moved from m_userContexts to
+    // m_userContextsPendingDeletion until their pages are closed.
+    bool isKnownUserContext = m_userContexts.contains(sessionIDString)
+        || m_userContextsPendingDeletion.contains(sessionIDString);
```
dcrousso: id just inline this since the variable is only used once

Source/WebKit/UIProcess/Automation/BidiBrowserAgent.cpp
```diff
@@ -114,6 +117,41 @@ void BidiBrowserAgent::willClosePage(const WebPageProxy& page)
...
+
+    m_pageUserContextOverrides.set(page.identifier(), userContextID);
+}
+
+String BidiBrowserAgent::getUserContextIDForPage(const WebPageProxy& page) const
+{
+    if (auto it = m_pageUserContextOverrides.find(page.identifier()); it != m_pageUserContextOverrides.end())
+        return it->value;
+
+    String sessionIDString = toUserContextIDProtocolString(page.sessionID());
```
dcrousso: other callers of this give the variable that holds the result the name `userContextID`

should we use that here too for consistency?

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -124,11 +124,30 @@ void BidiBrowsingContextAgent::create(Inspector::Protocol::BidiBrowsingContext::
...
+
+        RefPtr referencePage = session->webPageProxyForHandle(pageHandle);
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(!referencePage, FrameNotFound);
+
+        if (userContextID.isEmpty())
+            userContextID = session->getUserContextIDForPage(*referencePage);
+    }
+
+    if (userContextID.isEmpty())
+        userContextID = "default"_s;
```
dcrousso: please use `defaultUserContextID()` instead (which i think will require making it usable outside of `Source/WebKit/UIProcess/Automation/BidiBrowserAgent.cpp` but i think it's worth it for ease-of-use)

as a followup it'd be nice to clean up all uses of `"default"_s` in other WebDriver code too :)

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -153,17 +172,20 @@ Inspector::Protocol::BidiBrowsingContext::BrowsingContext BidiBrowsingContextAge
...
+Ref<Inspector::Protocol::BidiBrowsingContext::Info> BidiBrowsingContextAgent::getNavigableInfo(const WebKit::FrameTreeNodeData& tree, const WebPageProxy& page, std::optional<uint64_t> maxDepth, IncludeParentID includeParentID)
 {
     // https://w3c.github.io/webdriver-bidi/#get-the-navigable-info
 
-    // FIXME: Properly support different user contexts, which will likely map to different WebAutomationSessions.
-    // https://bugs.webkit.org/show_bug.cgi?id=288104
+    RefPtr session = m_session.get();
+    auto [clientWindow, userContext] = session
+        ? session->getClientWindowAndUserContext(page)
+        : std::pair<String, String>("unknown-window"_s, "default"_s);
```
dcrousso: ditto above for `defaultClientWindowID()` and `defaultUserContextID()`

Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp
```diff
@@ -460,6 +460,18 @@ static Inspector::Protocol::Automation::BrowsingContextPresentation NODELETE toP
 }
 
 void WebAutomationSession::createBrowsingContext(std::optional<Inspector::Protocol::Automation::BrowsingContextPresentation>&& presentationHint, CommandCallbackOf<String, Inspector::Protocol::Automation::BrowsingContextPresentation>&& callback)
+{
+    createBrowsingContextInternal(WTF::move(presentationHint), emptyString(), WTF::move(callback));
```
dcrousso, line 467: rather than pass `emptyString()` can we use `defaultUserContextID()` so that we can remove some of the `!userContextIDForBidi.isEmpty()` checks inside `WebAutomationSession::createBrowsingContextInternal` since i think most of that will already be handled inside `BidiBrowserAgent::setUserContextIDForPage` (specifically that it'll ignore `defaultUserContextID()`)?

Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp
```diff
@@ -1097,23 +1132,33 @@ void WebAutomationSession::fragmentNavigatedForFrame(const WebFrameProxy& frame,
...
-            for (auto& tree : trees)
-                recursivelyEmitContextCreatedEvent(tree, std::nullopt);
-        });
-    }
+    if (!m_pagesWithContextCreatedEventEmitted.add(page.identifier()).isNewEntry)
+        return;
+
+    auto contextHandle = handleForWebPageProxy(page);
+
+    String originalOpenerHandle = "null"_s;
```
dcrousso: is there a `defaultOpenerHandle()`?

Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp
```diff
@@ -1097,23 +1132,33 @@ void WebAutomationSession::fragmentNavigatedForFrame(const WebFrameProxy& frame,
...
+    if (RefPtr openerPage = getOpenerPage(page))
+        originalOpenerHandle = handleForWebPageProxy(*openerPage);
+
+    String url = page.currentURL();
+    if (url.isEmpty())
+        url = "about:blank"_s;
```
dcrousso: can we use `aboutBlankURL().string()`?

Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp
```diff
@@ -1097,23 +1132,33 @@ void WebAutomationSession::fragmentNavigatedForFrame(const WebFrameProxy& frame,
...
+
+    auto [clientWindow, userContext] = getClientWindowAndUserContext(page);
+
+    m_bidiProcessor->browsingContextDomainNotifier().contextCreated(contextHandle, url, originalOpenerHandle, "null"_s, JSON::ArrayOf<Inspector::Protocol::BidiBrowsingContext::Info>::create(), clientWindow, userContext);
```
dcrousso: NIT: can we create a variable like `const auto parentBrowsingContext = "null"_s;` so that it's clear what it's for?

also, i think since it's the only thing marked optional you could maybe omit it?  this is usually why we keep optional parameters at the end of the command arguments as it's a lot easier to reason about that way

Source/WebKit/UIProcess/Automation/WebDriverBidiProcessor.cpp
```diff
@@ -154,6 +154,32 @@ static String toBidiErrorCode(int errorCode, const String& inspectorInternalMsg)
     }
 }
 
+static void normalizeBrowsingContextInfoEvent(JSON::Object& message)
```
dcrousso: this does not seem ideal.  it effectively breaks the contract established by the protocol JSON (e.g. `children` is specified as an array so why would we want it to be `null`?).  unless this is something absolutely necessary (and id want a _really_ good reason why) then id suggest modifying the protocol to allow for this and having the handler deal with the repercussions (e.g. having `children` be `any` and doing additional type checking/casting)

Source/WebKit/UIProcess/Automation/BidiBrowserAgent.cpp
juliopiubello: done

Source/WebKit/UIProcess/Automation/BidiBrowserAgent.cpp
juliopiubello: renamed it to userContextID for consistency with the rest of the callers

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
juliopiubello: promoted defaultUserContextID() out of BidiBrowserAgent.cpp and used it here

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
juliopiubello: switched the fallback to the shared helpers as well

Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp
juliopiubello: Agreed, but I only made that change after fixing the old FIFO association. It now uses defaultUserContextID() with request-scoped tracking instead of the global pending queue.

Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp
juliopiubello: I reworked it and now we don't need to pass raw sentinel strings through, the event payload is built directly with nullable

Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp
juliopiubello: this is gone now, replaced with the source-built event payloads

Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp
juliopiubello: also not needed anymore

Source/WebKit/UIProcess/Automation/WebDriverBidiProcessor.cpp
juliopiubello: Agreed. I removed the transport-layer rewrite and now construct the nullable event fields at the source in WebAutomationSession. The issue was in how we emitted the event, not in the protocol contract.

Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp
```diff
@@ -469,11 +485,42 @@ void WebAutomationSession::createBrowsingContext(std::optional<Inspector::Protoc
...
+    RELEASE_ASSERT(addResult.isNewEntry);
+#endif
+
+    m_client->requestNewPageWithOptions(*this, static_cast<API::AutomationSessionBrowsingContextOptions>(options), [protectedThis = Ref { *this }, userContextIDForBidi = userContextIDForBidi.isolatedCopy(),
+#if ENABLE(WEBDRIVER_BIDI)
+        requestIdentifier,
+#endif
+        callback = WTF::move(callback)](WebPageProxy* page) {
+#if ENABLE(WEBDRIVER_BIDI)
+        bool removedPendingRequest = protectedThis->m_pendingCreateBrowsingContextRequestIdentifiers.remove(requestIdentifier);
```
dcrousso, line 500: this indentation is really weird :/

i wonder if for the sake of clarity we should just have `requestNewPageWithOptions` always require a `requestIdentifier` but just have it be `UNUSED_PARAM`

Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp
```diff
@@ -995,6 +1042,52 @@ static String navigationIDToProtocolString(std::optional<WebCore::NavigationIden
...
+        object.setValue("children"_s, JSON::Value::null());
+        return;
+    }
+
+    object.setArray("children"_s, children.releaseNonNull());
+}
+
+static void sendBrowsingContextInfoEvent(WebDriverBidiProcessor& bidiProcessor, ASCIILiteral method, const String& context, const String& url, const String& originalOpener, std::optional<String>&& parent, RefPtr<JSON::ArrayOf<Inspector::Protocol::BidiBrowsingContext::Info>>&& children, const String& clientWindow, const String& userContext)
+{
+    auto params = JSON::Object::create();
```
dcrousso, line 1076: again the fact that we're manually crafting this message object entirely defeats the purpose of having a protocol JSON

does the `"nullable": true` not actually do anything?  id much rather modify the protocol C++ generator to use a `std::variant<..., JSON::Value::Type::Null>` or something along those lines so that we still have the rest of the guarantees provided by the system (i.e. if not `null` then it's the expected type)

juliopiubello: sorry for the back and forth with this one. 

I was worried on keeping this PR out of the shared codegen work. The generator still does not carry nullable on event parameters, should I go ahead and add to this PR the change there? 

For what i got the event parameters only carry optional, so the generated event notifier cannot express the full tri-state the protocol allows.

Or else what I can do is not change the shared inspecteor to model nullable event and make the nullable fields inside of the object use the generated null-aware setters, and keep the outer BiDi event envelope still wrapping manually. Recovering the type guarantees but not solving the nullable event parameters generically

Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp
juliopiubello: will do!

Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp
juliopiubello: > I was worried on keeping this PR out of the shared codegen work. The generator still does not carry nullable on event parameters, should I go ahead and add to this PR the change there?

should I go ahead and implement the rework on shared coden to add nullable event parameter support*

dcrousso: id suggest adding nullable support to the protocol C++ generator separately before this PR as that will help keep things clean

id probably have something like the following in `Source/WTF/wtf/JSONValues.h`
```diff
-static Ref<Value> null();
+class Null : public Value {
+private:
+    friend class Value;
+
+protected:
+    Null() = default;
+};
+
+static Ref<Null> null();
```
to introduce a new concrete type that can only be created via `JSON::Value::null()` and then have this for the protocol C++
```cpp
using Nullable<T> = std::variant<T, JSON::Null>;
```
so that callers of protocol events can provide either the `T` or a `JSON::Value::null()`

juliopiubello: perfect, I will move this PR to draft, then push a PR for it and then come back here later, thanks for the help!

while we are at it, since the field is both optional and nullable the full c++ type will probably look like  std::optional<Nullable<T>> ?

dcrousso: yeah i think `std::optional<Nullable<T>>` is what we want

also, i'd strongly suggest using this time to reorder the parameters of all commands and events such that optional items are last (with all nullable items being before that) so that it's super clear what arguments can be omitted (i.e. all required arguments then all nullable arguments then all optional arguments)

dcrousso: hopefully we haven't shipped any of those as that would require some compatibility code shenanigans in the frontend invoker/receiver 🥲

juliopiubello: got it :)

burg: Yeah, I recommend fixing the codegen in this case. Feel free to copy the protocol definitions in question that are not generated correctly, and use them as a test case in inspector code generator tests.

PR #58237, opened 2026-02-09
[appearance: base] Create shadow trees for base slider controls

shallawa:
#### 5c7d9687b3f3301a590ecf4a5d3606fdc7018bd8
<pre>
[appearance: base] Create shadow trees for base slider controls
<a href="https://bugs.webkit.org/show_bug.cgi?id=305895">https://bugs.webkit.org/show_bug.cgi?id=305895</a>
<a href="https://rdar.apple.com/168599962">rdar://168599962</a>

Reviewed by NOBODY (OOPS!).

Support appearance: base for the &lt;meter&gt; element. Add pseudo elements for the
track and fill. Create a shadow tree elements for the these pseudo elements.
These new shadow tree elements have to coexist with the shadow tree elements of
the appearance: none case. Only a shadow sub-tree should be shown at any time.
Switching between them will happen in HTMLMeterElement::didRecalcStyle().

Also we need to add three pseudo classes for the meter gauge regions. The right
pseudo class will be chosen in SelectorChecker::checkOne() based on the value
attribute of the &lt;meter&gt; element.

Tests: fast/forms/appearance-base/appearance-base-meter-dynamic.html
       fast/forms/appearance-base/appearance-base-meter-initial.html
       fast/forms/appearance-base/appearance-base-meter-value-dynamic.html

* LayoutTests/fast/dom/HTMLMeterElement/meter-element-markup-expected.txt:
* LayoutTests/fast/forms/appearance-base/appearance-base-meter-dynamic-expected.html: Added.
* LayoutTests/fast/forms/appearance-base/appearance-base-meter-dynamic.html: Added.
* LayoutTests/fast/forms/appearance-base/appearance-base-meter-initial-expected.html: Added.
* LayoutTests/fast/forms/appearance-base/appearance-base-meter-initial.html: Added.
* LayoutTests/fast/forms/appearance-base/appearance-base-meter-value-dynamic-expected.html: Added.
* LayoutTests/fast/forms/appearance-base/appearance-base-meter-value-dynamic.html: Added.
* LayoutTests/platform/glib/fast/dom/HTMLMeterElement/meter-boundary-values-expected.txt:
* LayoutTests/platform/glib/fast/dom/HTMLMeterElement/meter-element-expected.txt:
* LayoutTests/platform/glib/fast/dom/HTMLMeterElement/meter-element-repaint-on-update-value-expected.txt:
* LayoutTests/platform/glib/fast/dom/HTMLMeterElement/meter-optimums-expected.txt:
* LayoutTests/platform/glib/fast/dom/HTMLMeterElement/meter-styles-changing-pseudo-expected.txt:
* LayoutTests/platform/glib/fast/dom/HTMLMeterElement/meter-styles-expected.txt:
* LayoutTests/platform/glib/fast/forms/appearance-base/appearance-base-meter-dynamic-expected.html: Added.
* LayoutTests/platform/glib/fast/forms/appearance-base/appearance-base-meter-initial-expected.html: Added.
* LayoutTests/platform/glib/fast/forms/appearance-base/appearance-base-meter-value-dynamic-expected.html: Added.
* LayoutTests/platform/ios/fast/dom/HTMLMeterElement/meter-element-repaint-on-update-value-expected.txt:
* LayoutTests/platform/ios/fast/dom/HTMLMeterElement/meter-styles-changing-pseudo-expected.txt:
* LayoutTests/platform/ios/fast/dom/HTMLMeterElement/meter-styles-expected.txt:
* LayoutTests/platform/mac-sequoia-wk2/fast/forms/appearance-base/appearance-base-meter-dynamic-expected.html: Added.
* LayoutTests/platform/mac-sequoia-wk2/fast/forms/appearance-base/appearance-base-meter-initial-expected.html: Added.
* LayoutTests/platform/mac-sequoia-wk2/fast/forms/appearance-base/appearance-base-meter-value-dynamic-expected.html: Added.
* LayoutTests/platform/mac/fast/dom/HTMLMeterElement/meter-element-repaint-on-update-value-expected.txt:
* LayoutTests/platform/mac/fast/dom/HTMLMeterElement/meter-styles-changing-pseudo-expected.txt:
* LayoutTests/platform/win/fast/dom/HTMLMeterElement/meter-element-expected.txt:
* LayoutTests/platform/win/fast/dom/HTMLMeterElement/meter-element-repaint-on-update-value-expected.txt:
* LayoutTests/platform/win/fast/dom/HTMLMeterElement/meter-styles-changing-pseudo-expected.txt:
* LayoutTests/platform/win/fast/forms/appearance-base/appearance-base-meter-dynamic-expected.html: Added.
* LayoutTests/platform/win/fast/forms/appearance-base/appearance-base-meter-initial-expected.html: Added.
* LayoutTests/platform/win/fast/forms/appearance-base/appearance-base-meter-value-dynamic-expected.html: Added.
* Source/JavaScriptCore/inspector/protocol/CSS.json:
* Source/WebCore/css/CSSPseudoSelectors.json:
* Source/WebCore/css/SelectorChecker.cpp:
(WebCore::SelectorChecker::checkOne const):
* Source/WebCore/css/SelectorCheckerTestFunctions.h:
(WebCore::matchesEvenLessGoodPseudoClass):
(WebCore::matchesOptimumPseudoClass):
(WebCore::matchesSuboptimumPseudoClass):
(WebCore::SelectorCompiler::JSC_DEFINE_NOEXCEPT_JIT_OPERATION):
(WebCore::SelectorCompiler::addPseudoClassType):
* Source/WebCore/html/HTMLMeterElement.cpp:
(WebCore::HTMLMeterElement::attributeChanged):
(WebCore::HTMLMeterElement::didChangeElementValue):
(WebCore::HTMLMeterElement::appendShadowTreeForAutoAppearance):
(WebCore::HTMLMeterElement::appendShadowTreeForBaseAppearance):
(WebCore::HTMLMeterElement::didAddUserAgentShadowRoot):
(WebCore::HTMLMeterElement::didElementStateChange): Deleted.
* Source/WebCore/html/HTMLMeterElement.h:
* Source/WebInspectorUI/UserInterface/Controllers/CSSManager.js:
(WI.CSSManager.displayNameForPseudoId):
</pre>
https://github.com/WebKit/WebKit/commit/5c7d9687b3f3301a590ecf4a5d3606fdc7018bd8

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/145983 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/18672 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/10821 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/154662 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/99519 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/9115c428-a28e-4037-b26e-efc0477d5beb) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/147858 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/19152 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/18562 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/112292 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/80388 "1 flakes 1 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/899ac72f-461a-44f4-8285-4cceeec670db) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/148946 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/14674 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/131139 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/93192 "Passed tests") | | [loading 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/7395c1ab-207b-47df-aa52-cf912c265675) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/13956 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/11711 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/2107 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/137965 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/123512 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/8073 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/156973 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/6787 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/194 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/9314 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/120303 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/18506 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/15471 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/120638 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/18539 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/129489 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/74217 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/16344 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/7430 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/177287 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/18125 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/81893 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/45507 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/17861 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/18041 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/17921 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/css/CSSPseudoSelectors.json
+++ b/Source/WebCore/css/CSSPseudoSelectors.json
@@ -119,6 +119,7 @@
             "comment": "For scrollbar styling.",
             "status": "non-standard"
         },
+        "even-less-good": {},
         "first-child": {},
         "first-of-type": {},
         "focus": {},
@@ -197,6 +198,7 @@
         "open": {
             "settings-flag": "openPseudoClassEnabled"
         },
+        "optimum": {},
         "optional": {},
         "out-of-range": {},
         "past": {
@@ -237,6 +239,7 @@
             "comment": "For scrollbar styling.",
             "status": "non-standard"
         },
+        "suboptimum": {},
         "target": {},
         "user-invalid": {},
         "user-valid": {},
@@ -536,6 +539,10 @@
             "aliases": [
                 "-webkit-file-upload-button"
             ],
+            "user-agent-part": true
+        },
+        "fill": {
+            "settings-flag": "cssAppearanceBaseEnabled",
             "user-agent-part": true
         },
         "first-letter": {
--- a/Source/WebCore/html/HTMLMeterElement.cpp
+++ b/Source/WebCore/html/HTMLMeterElement.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2010 Nokia Corporation and/or its subsidiary(-ies).
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Library General Public
@@ -22,18 +23,12 @@
 #include "HTMLMeterElement.h"
 
 #include "Attribute.h"
-#include "ContainerNodeInlines.h"
-#include "ElementInlines.h"
-#include "ElementIterator.h"
 #include "HTMLDivElement.h"
-#include "HTMLFormElement.h"
 #include "HTMLNames.h"
 #include "HTMLParserIdioms.h"
 #include "HTMLStyleElement.h"
 #include "NodeName.h"
-#include "Page.h"
 #include "RenderMeter.h"
-#include "RenderStyle+GettersInlines.h"
 #include "RenderTheme.h"
 #include "ScriptDisallowedScope.h"
 #include "ShadowRoot.h"
@@ -84,12 +79,13 @@ void HTMLMeterElement::attributeChanged(const QualifiedName& name, const AtomStr
     case AttributeNames::lowAttr:
     case AttributeNames::highAttr:
     case AttributeNames::optimumAttr:
-        didElementStateChange();
+        didChangeElementValue();
         break;
     default:
-        HTMLElement::attributeChanged(name, oldValue, newValue, attributeModificationReason);
         break;
     }
+
+    HTMLElement::attributeChanged(name, oldValue, newValue, attributeModificationReason);
 }
 
 double HTMLMeterElement::min() const
@@ -189,14 +185,21 @@ static void setValueClass(HTMLElement& element, HTMLMeterElement::GaugeRegion ga
     ASSERT_NOT_REACHED();
 }
 
-void HTMLMeterElement::didElementStateChange()
+void HTMLMeterElement::didChangeElementValue()
 {
-    Ref valueElement = *m_valueElement;
-    valueElement->setInlineStyleProperty(CSSPropertyInlineSize, valueRatio() * 100, CSSUnitType::CSS_PERCENTAGE);
-    setValueClass(valueElement, gaugeRegion());
+    RefPtr shadowRoot = userAgentShadowRoot();
+    if (!shadowRoot)
+        return;
 
-    if (CheckedPtr renderer = renderMeter())
-        renderer->updateFromElement();
+    if (RefPtr valueElement = m_valueElement) {
+        valueElement->setInlineStyleProperty(CSSPropertyInlineSize, valueRatio() * 100, CSSUnitType::CSS_PERCENTAGE);
+        setValueClass(*valueElement, gaugeRegion());
+    }
+
+    if (RefPtr fillElement = m_fillElement) {
+        fillElement->setInlineStyleProperty(CSSPropertyTransform, makeString("translate(-"_s, (1 - valueRatio()) * 100, "%, 0)"_s));
+        fillElement->invalidateStyleInternal();
+    }
 }
 
 RenderMeter* HTMLMeterElement::renderMeter() const
@@ -204,39 +207,61 @@ RenderMeter* HTMLMeterElement::renderMeter() const
     return dynamicDowncast<RenderMeter>(renderer());
 }
 
-void HTMLMeterElement::didAddUserAgentShadowRoot(ShadowRoot& root)
+void HTMLMeterElement::appendShadowTreeForAutoAppearance(ShadowRoot& root)
 {
-    ASSERT(!m_valueElement);
-
     static MainThreadNeverDestroyed<const String> shadowStyle(StringImpl::createWithoutCopying(meterElementShadowUserAgentStyleSheet));
 
     Ref document = this->document();
-    Ref style = HTMLStyleElement::create(document);
-    ScriptDisallowedScope::EventAllowedScope styleScope { style };
-    style->setTextContent(String { shadowStyle });
+    Ref styleElement = HTMLStyleElement::create(document);
+    ScriptDisallowedScope::EventAllowedScope styleScope { styleElement };
+    styleElement->setTextContent(String { shadowStyle });
     ScriptDisallowedScope::EventAllowedScope rootScope { root };
-    root.appendChild(WTF::move(style));
+    root.appendChild(WTF::move(styleElement));
 
     // Pseudos are set to allow author styling.
-    Ref inner = HTMLDivElement::create(document);
-    ScriptDisallowedScope::EventAllowedScope innerScope { inner };
-    inner->setIdAttribute("inner"_s);
-    inner->setUserAgentPart(UserAgentParts::webkitMeterInnerElement());
-    root.appendChild(inner);
+    Ref innerElement = HTMLDivElement::create(document);
+    ScriptDisallowedScope::EventAllowedScope innerScope { innerElement };
+    innerElement->setIdAttribute("inner"_s);
+    innerElement->setUserAgentPart(UserAgentParts::webkitMeterInnerElement());
+    innerElement->setInlineStyleProperty(CSSPropertyDisplay, "-internal-auto-base(inline-block, none) !important"_s);
+    root.appendChild(innerElement);
 
-    Ref bar = HTMLDivElement::create(document);
-    ScriptDisallowedScope::EventAllowedScope barScope { bar };
-    bar->setIdAttribute("bar"_s);
-    bar->setUserAgentPart(UserAgentParts::webkitMeterBar());
-    inner->appendChild(bar);
+    Ref barElement = HTMLDivElement::create(document);
+    ScriptDisallowedScope::EventAllowedScope barScope { barElement };
+    barElement->setIdAttribute("bar"_s);
+    barElement->setUserAgentPart(UserAgentParts::webkitMeterBar());
+    innerElement->appendChild(barElement);
 
     Ref valueElement = HTMLDivElement::create(document);
     ScriptDisallowedScope::EventAllowedScope valueElementScope { valueElement };
     valueElement->setIdAttribute("value"_s);
-    bar->appendChild(valueElement);
-    m_valueElement = WTF::move(valueElement);
-
-    didElementStateChange();
+    barElement->appendChild(valueElement);
+    m_valueElement = valueElement;
 }
 
-} // namespace
+void HTMLMeterElement::appendShadowTreeForBaseAppearance(ShadowRoot& root)
+{
+    Ref document = this->document();
+    Ref trackElement = HTMLDivElement::create(document);
+    ScriptDisallowedScope::EventAllowedScope trackScope { trackElement };
+    trackElement->setUserAgentPart(UserAgentParts::track());
+    trackElement->setInlineStyleProperty(CSSPropertyAppearance, "inherit"_s);
+    trackElement->setInlineStyleProperty(CSSPropertyDisplay, "-internal-auto-base(none, inline-block) !important"_s);
+    root.appendChild(trackElement);
+
+    Ref fillElement = HTMLDivElement::create(document);
+    ScriptDisallowedScope::EventAllowedScope fillScope { fillElement };
+    fillElement->setUserAgentPart(UserAgentParts::fill());
+    trackElement->appendChild(fillElement);
+    m_fillElement = fillElement;
+}
+
+void HTMLMeterElement::didAddUserAgentShadowRoot(ShadowRoot& root)
+{
+    appendShadowTreeForAutoAppearance(root);
+    if (document().settings().cssAppearanceBaseEnabled())
+        appendShadowTreeForBaseAppearance(root);
+    didChangeElementValue();
+}
+
+} // namespace WebCore
--- a/LayoutTests/fast/dom/HTMLMeterElement/meter-element-markup-expected.txt
+++ b/LayoutTests/fast/dom/HTMLMeterElement/meter-element-markup-expected.txt
@@ -18,6 +18,12 @@ Both meter elements should have a nested shadow box with a width specified:
 |           id="value"
 |           style="inline-size: 70%;"
 |           useragentpart="-webkit-meter-optimum-value"
+|     <div>
+|       style="appearance: inherit;"
+|       useragentpart="track"
+|       <div>
+|         style="transform: translate(-30%, 0px);"
+|         useragentpart="fill"
 | "\n    "
 | <meter>
 |   high="6"
@@ -40,6 +46,12 @@ Both meter elements should have a nested shadow box with a width specified:
 |           id="value"
 |           style="inline-size: 100%;"
 |           useragentpart="-webkit-meter-suboptimum-value"
+|     <div>
+|       style="appearance: inherit;"
+|       useragentpart="track"
+|       <div>
+|         style="transform: translate(0%, 0px);"
+|         useragentpart="fill"
 | "\n    "
 | <meter>
 |   high="6"
@@ -62,4 +74,10 @@ Both meter elements should have a nested shadow box with a width specified:
 |           id="value"
 |           style="inline-size: 100%;"
 |           useragentpart="-webkit-meter-even-less-good-value"
+|     <div>
+|       style="appearance: inherit;"
+|       useragentpart="track"
+|       <div>
+|         style="transform: translate(0%, 0px);"
+|         useragentpart="fill"
 | "\n  "
--- a/LayoutTests/fast/forms/appearance-base/appearance-base-meter-dynamic-expected.html
+++ b/LayoutTests/fast/forms/appearance-base/appearance-base-meter-dynamic-expected.html
@@ -1,0 +1,126 @@
+<style>
+    .auto {
+        font: 11px system-ui;
+        position: relative;
+        top: 50%;
+        transform: translateY(-50%);
+        font: 11px system-ui;
+        width: calc(100% - 1px);
+        height: calc(100% - 2px);
+        border-radius: 999em;
+        overflow: hidden;
+        background: rgb(230, 230, 230);
+    }
+
+    .auto > .fill {
+        height: 100%; 
+        width: 100%; 
+        block-size: 100%;
+        box-sizing: border-box;
+        border-radius: 999em;
+        overflow: hidden;
+    }
+
+    .none {
+        font: 11px system-ui;
+        width: 100%;
+        height: 100%; 
+        overflow: hidden;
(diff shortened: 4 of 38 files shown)


Source/WebCore/html/HTMLMeterElement.cpp
```diff
@@ -190,25 +195,66 @@ static void setValueClass(HTMLElement& element, HTMLMeterElement::GaugeRegion ga
...
+        if (child->userAgentPart() == userAgentPart)
+            return WTF::move(child);
+    }
 
-    if (CheckedPtr renderer = renderMeter())
-        renderer->updateFromElement();
+    return nullptr;
+}
+
+void HTMLMeterElement::didElementAppearanceChange(const RenderStyle& style)
```
rniwa: I'd call `didChangeElementAppearance` or `didUpdateElementAppearance`.

Source/WebCore/html/HTMLMeterElement.cpp
```diff
@@ -190,25 +195,66 @@ static void setValueClass(HTMLElement& element, HTMLMeterElement::GaugeRegion ga
     }
 }
 
-void HTMLMeterElement::didElementStateChange()
+static RefPtr<Element> elementByUserAgentPart(ShadowRoot& shadowRoot, const AtomString& userAgentPart)
 {
-    Ref valueElement = *m_valueElement;
-    valueElement->setInlineStyleProperty(CSSPropertyInlineSize, valueRatio() * 100, CSSUnitType::CSS_PERCENTAGE);
-    setValueClass(valueElement, gaugeRegion());
+    for (Ref child : descendantsOfType<Element>(shadowRoot)) {
```
rniwa: This is rather inefficient. Can we just keep a `WeakPtr` to `fill` instead?

Source/WebCore/html/HTMLMeterElement.cpp
```diff
@@ -190,25 +195,66 @@ static void setValueClass(HTMLElement& element, HTMLMeterElement::GaugeRegion ga
...
+    if (!inner)
+        return;
+
+    RefPtr track = dynamicDowncast<HTMLDivElement>(elementByUserAgentPart(*shadowRoot, UserAgentParts::track()));
+    if (!track)
+        return;
+
+    if (style.appearance() == StyleAppearance::Base) {
+        inner->setInlineStyleProperty(CSSPropertyDisplay, CSSValueNone);
+        track->setInlineStyleProperty(CSSPropertyDisplay, CSSValueBlock);
```
rniwa: Shouldn't this be an `inline-block`?

Source/WebCore/html/HTMLMeterElement.cpp
```diff
@@ -190,25 +195,66 @@ static void setValueClass(HTMLElement& element, HTMLMeterElement::GaugeRegion ga
...
+    } else {
+        inner->setInlineStyleProperty(CSSPropertyDisplay, CSSValueBlock);
+        track->setInlineStyleProperty(CSSPropertyDisplay, CSSValueNone);
+    }
+}
+
+void HTMLMeterElement::didElementValueChange()
```
rniwa: Ditto. `didChangeElementValue` or `didUpdateElementValue`.

Source/WebCore/html/HTMLMeterElement.cpp
```diff
@@ -190,25 +195,66 @@ static void setValueClass(HTMLElement& element, HTMLMeterElement::GaugeRegion ga
...
+        setValueClass(*value, gaugeRegion());
+    }
+
+    if (RefPtr fill = dynamicDowncast<HTMLDivElement>(elementByUserAgentPart(*shadowRoot, UserAgentParts::fill()))) {
+        fill->setInlineStyleProperty(CSSPropertyTransform, makeString("translate(-"_s, (1 - valueRatio()) * 100, "%, 0)"_s));
+        fill->invalidateStyleInternal();
+    }
+
+    if (CheckedPtr renderer = this->renderer())
+        renderer->repaint();
```
rniwa: This is rather unusual isn't it? Why is it desirable / okay to synchronously trigger `repaint` right here?

Source/WebCore/html/HTMLMeterElement.cpp
```diff
@@ -190,25 +195,63 @@ static void setValueClass(HTMLElement& element, HTMLMeterElement::GaugeRegion ga
...
+    RefPtr inner = dynamicDowncast<HTMLDivElement>(shadowRoot->getElementById(String("inner"_s)));
+    if (!inner)
+        return;
+
+    RefPtr track = dynamicDowncast<HTMLDivElement>(elementByUserAgentPart(*shadowRoot, UserAgentParts::track()));
+    if (!track)
+        return;
+
+    if (style.appearance() == StyleAppearance::Base) {
+        inner->setInlineStyleProperty(CSSPropertyDisplay, CSSValueNone);
```
rniwa: I don't think we're supposed to mutate DOM tree like this during `createElementRenderer`.

Source/WebCore/rendering/style/RenderStyleConstants.h
```diff
@@ -64,6 +64,8 @@ enum class PseudoElementType : uint8_t {
     SpellingError,
     TargetText,
     Checkmark,
+    Track,
+    Fill,
```
nt1m: You shouldn't need this and all the boilerplate related to it if you're using `"user-agent-part": true`

Source/WebCore/html/HTMLMeterElement.cpp
```diff
@@ -235,9 +281,28 @@ void HTMLMeterElement::didAddUserAgentShadowRoot(ShadowRoot& root)
...
+    track->setUserAgentPart(UserAgentParts::track());
+
+    track->appendChild(WTF::move(fill));
+    root.appendChild(WTF::move(track));
+}
+
+void HTMLMeterElement::didAddUserAgentShadowRoot(ShadowRoot& root)
+{
+    appendShadowTreeForAutoAppearance(root);
+    appendShadowTreeForBaseAppearance(root);
```
anttijk: Does Chrome deal with this in similar way? Are we going to have two shadow trees in all HTML elements that support base appearance? 

If this needs optimizing we could probably create the shadow trees (maybe for the whole document) after figuring out there are elements actually using base appearance and then trigger another style recall to style them.

Source/WebCore/html/HTMLMeterElement.cpp
```diff
@@ -84,12 +87,50 @@ void HTMLMeterElement::attributeChanged(const QualifiedName& name, const AtomStr
...
+    if (!track)
+        return;
+
+    if (style->appearance() == StyleAppearance::Base) {
+        inner->setInlineStyleProperty(CSSPropertyDisplay, CSSValueNone);
+        track->setInlineStyleProperty(CSSPropertyDisplay, CSSValueInlineBlock);
+    } else {
+        inner->setInlineStyleProperty(CSSPropertyDisplay, CSSValueInlineBlock);
+        track->setInlineStyleProperty(CSSPropertyDisplay, CSSValueNone);
+    }
```
anttijk: This seems wrong way to do this. I think `StyleAdjuster::adjust` could check if the element is a shadow tree root for base or auto, check the appearance property parent style (which is the same as host style for shadow tree roots) and adjust the `display` property accordingly.

Source/WebCore/html/HTMLMeterElement.cpp
shallawa: As far as I know, Chrome has not implemented appearance: base for in-page form controls.

My understanding is we should not change the shadow DOM tree when the style changes.

So if we know that all of the form controls has `appearance: none;` and we create shadow DOM trees for `none` only, the JS can change the style to be `appearance: base;` at any point. What should happen in this case?

rniwa: Yeah, we need some kind of container-query-like mechanism which can trigger a DOM mutation in a shadow tree based on the value of `appearance` property if we wanted to mutate DOM tree on demand like that.

An alternative approach is to add a bespoke mechanism to toggle `display` property based on `appearance`.

rniwa: Yet another approach we had considered is to update these UA shadow trees at the beginning of style recalc like SVG use elements but that won't quite work because we need the used value for `appearance` to figure out which appearance shadow tree would take so it needs to happen at the end of a style recalc instead.

nt1m: https://commits.webkit.org/307303@main should have fixed `display: -internal-auto-base(...)`, so we should consider leveraging that here.

Source/WebCore/html/HTMLMeterElement.cpp
shallawa: This function was deleted.

Source/WebCore/html/HTMLMeterElement.cpp
shallawa: Done. This was replaced by `"-internal-auto-base(none, inline-block) !important"`

Source/WebCore/html/HTMLMeterElement.cpp
shallawa: Done.

Source/WebCore/html/HTMLMeterElement.cpp
shallawa: I am not sure. But this `repaint()` is not new. It was added in https://commits.webkit.org/84444@main.

Source/WebCore/html/HTMLMeterElement.cpp
shallawa: Yes this is right. And this code was reverted from the PR.

Source/WebCore/rendering/style/RenderStyleConstants.h
shallawa: Done. The pseudo elements definitions were removed from this PR.

Source/WebCore/html/HTMLMeterElement.cpp
shallawa: Yes using `-internal-auto-base(...)` in setting the `display` property fixes the issue and I do not have to mutate the render tree or the CSS styles when the `appearance` changes.

Source/WebCore/html/HTMLMeterElement.cpp
shallawa: This code was reverted from the PR.

Source/WebCore/html/HTMLMeterElement.cpp
shallawa: This code was removed and `WeakPtrs` are used for the `fill` and the `value` elements.

Source/WebCore/css/CSSPseudoSelectors.json
```diff
@@ -119,6 +119,7 @@
             "comment": "For scrollbar styling.",
             "status": "non-standard"
         },
+        "even-less-good": {},
```
nt1m, line 122: I'd rather separate these behind a separate change

Source/WebCore/html/HTMLMeterElement.cpp
```diff
@@ -189,54 +185,83 @@ static void setValueClass(HTMLElement& element, HTMLMeterElement::GaugeRegion ga
...
-    Ref bar = HTMLDivElement::create(document);
-    ScriptDisallowedScope::EventAllowedScope barScope { bar };
-    bar->setIdAttribute("bar"_s);
-    bar->setUserAgentPart(UserAgentParts::webkitMeterBar());
-    inner->appendChild(bar);
+    Ref innerElement = HTMLDivElement::create(document);
+    ScriptDisallowedScope::EventAllowedScope innerScope { innerElement };
+    innerElement->setIdAttribute("inner"_s);
+    innerElement->setUserAgentPart(UserAgentParts::webkitMeterInnerElement());
+    innerElement->setInlineStyleProperty(CSSPropertyDisplay, "-internal-auto-base(inline-block, none) !important"_s);
```
nt1m, line 226: Not sure why !important matters here, !important only affects cascade, so it won't do anything unless something else overrides it.

nt1m: `appearance: inherit` is missing from this one so I don't know how this works.

Also, while using inline styles work, it'll likely end up duplicated once you address multiple controls.

Source/WebCore/css/CSSPseudoSelectors.json
shallawa: I created this PR https://github.com/WebKit/WebKit/pull/59526 to address adding the pseudo classes of the <meter> element.

Source/WebCore/html/HTMLMeterElement.cpp
shallawa: > `appearance: inherit` is missing from this one so I don't know how this works.

It is in the `meterElementShadow.css` file

```
div#inner {
    appearance: inherit;
    box-sizing: inherit;
    height: 100%;
    width: 100%;
}
```

PR #58240, opened 2026-02-09
[Site Isolation] Web Inspector: support cross-origin iframes in Runtime domain

AnujPanta1:
#### 1a86de0bf48e776d1b8ad1bbc4a27cf1f6eab79b
<pre>
[Site Isolation] Web Inspector: support cross-origin iframes in Runtime domain
<a href="https://bugs.webkit.org/show_bug.cgi?id=307370">https://bugs.webkit.org/show_bug.cgi?id=307370</a>
<a href="https://rdar.apple.com/143598571">rdar://143598571</a>

Reviewed by BJ Burg.

Introducing FrameRuntimeAgent to enable JavaScript evaluation in cross-origin iframes under site
isolation. When cross-origin iframes run in separate WebProcesses, they need their own Runtime
agents to handle protocol commands independently since PageRuntimeAgent can only operate within
a single process.

Each FrameInspectorController creates its own InjectedScriptManager instance rather than sharing
with the parent PageInspectorController. Even when frames happen to share a process, isolating
the InjectedScriptManager per frame ensures independent runtime state — objectId namespaces,
saved result indices, and execution context IDs — so the frontend can treat each frame target
as a fully independent target. This also prevents entanglement when tearing down one frame&apos;s
inspector state while another remains live.

InspectorInstrumentation routing was updated to direct didClearWindowObjectInWorld notifications
to the appropriate agent based on frame type. Subframes with FrameInspectorControllers route to
their FrameRuntimeAgent for context updates and return early, completely skipping PageRuntimeAgent.
Main frames continue using PageRuntimeAgent for backwards compatibility. The routing preserves
InspectorPageAgent calls for bootstrap script injection in both code paths.

On the frontend, FrameTarget stores its main execution context directly via the inherited
_executionContext field on WI.Target rather than going through
ExecutionContextList.pageExecutionContext.
Navigation-reset logic lives in FrameTarget.addExecutionContext, keeping ExecutionContextList
free of frame-target-specific concerns. RuntimeObserver dispatches context creation events to
FrameTarget.addExecutionContext for frame-type targets instead of routing through NetworkManager.

FrameRuntimeAgent stores the FrameIdentifier directly and converts to a protocol string on demand
in frameIdForProtocol rather than caching the string at construction time.

* Source/JavaScriptCore/inspector/protocol/Runtime.json:
* Source/WebCore/Sources.txt:
* Source/WebCore/inspector/FrameInspectorController.cpp:
(WebCore::FrameInspectorController::FrameInspectorController):
(WebCore::FrameInspectorController::createRuntimeAgent):
(WebCore::FrameInspectorController::createLazyAgents):
(WebCore::FrameInspectorController::debugger):
* Source/WebCore/inspector/FrameInspectorController.h:
(WebCore::InspectorInstrumentation::didClearWindowObjectInWorldImpl):
* Source/WebCore/inspector/InstrumentingAgents.h:
* Source/WebCore/inspector/agents/frame/FrameRuntimeAgent.cpp: Added.
(WebCore::FrameRuntimeAgent::FrameRuntimeAgent):
(WebCore::m_frameIdentifier):
(WebCore::FrameRuntimeAgent::enable):
(WebCore::FrameRuntimeAgent::disable):
(WebCore::FrameRuntimeAgent::callFunctionOn):
(WebCore::FrameRuntimeAgent::injectedScriptForEval):
(WebCore::FrameRuntimeAgent::muteConsole):
(WebCore::FrameRuntimeAgent::unmuteConsole):
(WebCore::FrameRuntimeAgent::frameIdForProtocol const):
(WebCore::toProtocol):
(WebCore::FrameRuntimeAgent::reportExecutionContextCreation):
(WebCore::FrameRuntimeAgent::notifyContextCreated):
(WebCore::FrameRuntimeAgent::didClearWindowObjectInWorld):
* Source/WebCore/inspector/agents/frame/FrameRuntimeAgent.h: Added.
(WI.NetworkManager.prototype.executionContextCreated):
* Source/WebInspectorUI/UserInterface/Controllers/RuntimeManager.js:
(WI.RuntimeManager.prototype.initializeTarget):
* Source/WebInspectorUI/UserInterface/Models/ExecutionContext.js:
(WI.ExecutionContext):
* Source/WebInspectorUI/UserInterface/Protocol/FrameTarget.js:
(WI.FrameTarget.prototype.get executionContextList):
(WI.FrameTarget.prototype.get executionContext):
(WI.FrameTarget.prototype.addExecutionContext):
(WI.FrameTarget.prototype.clearExecutionContexts):
(WI.FrameTarget):
* Source/WebInspectorUI/UserInterface/Protocol/RuntimeObserver.js:
(WI.RuntimeObserver.prototype.executionContextCreated):
(WI.RuntimeObserver):
* Source/WebInspectorUI/UserInterface/Views/QuickConsole.js:
(WI.QuickConsole):
(WI.QuickConsole.prototype.closed):
(WI.QuickConsole.prototype._updateActiveExecutionContextDisplay):
(WI.QuickConsole.prototype._populateActiveExecutionContextNavigationItemContextMenu):
(WI.QuickConsole.prototype._handleFrameTargetExecutionContextAdded):
(WI.QuickConsole.prototype._frameTargetsWithExecutionContext):

Canonical link: <a href="https://commits.webkit.org/308589@main">https://commits.webkit.org/308589@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/138fa40b29ee0f73854ffb0f0b1c04416216c099

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/147912 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/20599 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/14194 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/156594 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/101327 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [loading 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/91ad8f4a-c4ba-4035-82b5-f9ea75058885) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/149786 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/21054 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/20503 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/114030 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/101327 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/ff8fe26a-f106-47a1-bc43-acd0a9773cad) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/150874 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/16274 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/132853 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94794 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/15420 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/13216 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/4034 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/139881 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/125026 "Passed tests") | [  ~~🧪 mac-wk1~~](https://ews-build.webkit.org/#/builders/167/builds/10750 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/158929 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/8701 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/2064 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/12244 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/122065 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/20397 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/17147 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/122268 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/31336 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/20406 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/132551 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/76549 "Built successfully") | | 
| | [  ~~🧪 vision-wk2~~](https://ews-build.webkit.org/#/builders/164/builds/17757 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/9313 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/179334 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/20014 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/83774 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/45936 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/19744 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/19892 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/19800 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/inspector/FrameInspectorController.cpp
+++ b/Source/WebCore/inspector/FrameInspectorController.cpp
@@ -34,6 +34,7 @@
 #include "DocumentPage.h"
 #include "FrameConsoleAgent.h"
 #include "FrameInlines.h"
+#include "FrameRuntimeAgent.h"
 #include "InspectorInstrumentation.h"
 #include "InspectorWebAgentBase.h"
 #include "InstrumentingAgents.h"
@@ -44,6 +45,7 @@
 #include "Settings.h"
 #include "WebInjectedScriptHost.h"
 #include "WebInjectedScriptManager.h"
+#include <JavaScriptCore/Debugger.h>
 #include <JavaScriptCore/InspectorAgentBase.h>
 #include <JavaScriptCore/InspectorBackendDispatcher.h>
 #include <JavaScriptCore/InspectorFrontendRouter.h>
@@ -63,7 +65,7 @@ WTF_MAKE_TZONE_ALLOCATED_IMPL(FrameInspectorController);
 FrameInspectorController::FrameInspectorController(LocalFrame& frame, PageInspectorController& parentPageController)
     : m_frame(frame)
     , m_instrumentingAgents(InstrumentingAgents::create(*this, parentPageController.instrumentingAgents()))
-    , m_injectedScriptManager(parentPageController.injectedScriptManager())
+    , m_injectedScriptManager(WebInjectedScriptManager::create(*this, WebInjectedScriptHost::create()))
     , m_frontendRouter(FrontendRouter::create())
     , m_backendDispatcher(BackendDispatcher::create(m_frontendRouter.copyRef(), &parentPageController.backendDispatcher()))
     , m_executionStopwatch(Stopwatch::create())
@@ -124,12 +126,36 @@ void FrameInspectorController::createConsoleAgent()
     m_didCreateConsoleAgent = true;
 }
 
+void FrameInspectorController::createRuntimeAgent()
+{
+    if (m_didCreateRuntimeAgent)
+        return;
+
+    RefPtr frame = m_frame.get();
+    if (!frame)
+        return;
+
+    auto context = frameAgentContext();
+    m_agents.append(makeUniqueRef<FrameRuntimeAgent>(context));
+    m_didCreateRuntimeAgent = true;
+}
+
 void FrameInspectorController::createLazyAgents()
 {
     if (m_didCreateLazyAgents)
         return;
 
     m_didCreateLazyAgents = true;
+
+    RefPtr frame = m_frame.get();
+    if (!frame || !protect(frame->settings())->siteIsolationEnabled())
+        return;
+
+    // Create debugger before agents that depend on it.
+    // FIXME: <https://webkit.org/b/298909> Add FrameDebuggerAgent to actively use this debugger.
+    m_debugger = makeUnique<JSC::Debugger>(vm());
+
+    createRuntimeAgent();
 }
 
 void FrameInspectorController::connectFrontend(Inspector::FrontendChannel& frontendChannel, bool isAutomaticInspection, bool immediatelyPause)
@@ -227,8 +253,8 @@ Stopwatch& FrameInspectorController::executionStopwatch() const
 
 JSC::Debugger* FrameInspectorController::debugger()
 {
-    // FIXME <https://webkit.org/b/298909> Add Debugger support for frame targets.
-    return nullptr;
+    // FIXME: <https://webkit.org/b/298909> Add full Debugger domain support for frame targets.
+    return m_debugger.get();
 }
 
 JSC::VM& FrameInspectorController::vm()
--- a/Source/WebCore/inspector/InspectorInstrumentation.cpp
+++ b/Source/WebCore/inspector/InspectorInstrumentation.cpp
@@ -78,6 +78,7 @@
 #include "WebSocketFrame.h"
 #include "WorkerInspectorController.h"
 #include "WorkerOrWorkletGlobalScope.h"
+#include "agents/frame/FrameRuntimeAgent.h"
 #include <JavaScriptCore/ConsoleMessage.h>
 #include <JavaScriptCore/ConsoleTypes.h>
 #include <JavaScriptCore/InspectorDebuggerAgent.h>
@@ -107,6 +108,13 @@ void InspectorInstrumentation::didClearWindowObjectInWorldImpl(InstrumentingAgen
 {
     if (auto* pageDebuggerAgent = instrumentingAgents.enabledPageDebuggerAgent())
         pageDebuggerAgent->didClearWindowObjectInWorld(frame, world);
+
+    if (auto* frameRuntimeAgent = frame.inspectorController().instrumentingAgents().enabledFrameRuntimeAgent()) {
+        frameRuntimeAgent->didClearWindowObjectInWorld(world);
+        if (CheckedPtr pageAgent = instrumentingAgents.enabledPageAgent())
+            pageAgent->didClearWindowObjectInWorld(frame, world);
+        return;
+    }
 
     if (auto* pageRuntimeAgent = instrumentingAgents.enabledPageRuntimeAgent())
         pageRuntimeAgent->didClearWindowObjectInWorld(frame, world);
--- a/Source/WebCore/inspector/agents/frame/FrameRuntimeAgent.cpp
+++ b/Source/WebCore/inspector/agents/frame/FrameRuntimeAgent.cpp
@@ -1,0 +1,270 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL APPLE INC. OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ * OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "FrameRuntimeAgent.h"
+
+#include "DOMWrapperWorld.h"
+#include "Document.h"
+#include "DocumentLoader.h"
+#include "DocumentPage.h"
+#include "FrameConsoleClient.h"
+#include "FrameLoader.h"
+#include "InstrumentingAgents.h"
+#include "JSDOMWindowCustom.h"
+#include "JSExecState.h"
+#include "LocalDOMWindow.h"
+#include "LocalFrame.h"
+#include "LocalFrameInlines.h"
+#include "Page.h"
+#include "ScriptController.h"
+#include "SecurityOrigin.h"
+#include "UserGestureEmulationScope.h"
+#include "WindowProxy.h"
+#include <JavaScriptCore/InjectedScript.h>
+#include <JavaScriptCore/InjectedScriptManager.h>
+#include <JavaScriptCore/InspectorProtocolObjects.h>
+#include <wtf/CheckedRef.h>
+#include <wtf/TZoneMallocInlines.h>
+#include <wtf/text/MakeString.h>
+
+namespace WebCore {
+
+using namespace Inspector;
+
+WTF_MAKE_TZONE_ALLOCATED_IMPL(FrameRuntimeAgent);
+
+FrameRuntimeAgent::FrameRuntimeAgent(FrameAgentContext& context)
+    : InspectorRuntimeAgent(context)
+    , m_frontendDispatcher(makeUniqueRef<RuntimeFrontendDispatcher>(context.frontendRouter))
+    , m_backendDispatcher(RuntimeBackendDispatcher::create(Ref { context.backendDispatcher }, this))
+    , m_instrumentingAgents(context.instrumentingAgents)
+    , m_inspectedFrame(context.inspectedFrame)
+    , m_frameIdentifier(context.inspectedFrame->frameID())
+{
+}
+
+FrameRuntimeAgent::~FrameRuntimeAgent() = default;
+
+CommandResult<void> FrameRuntimeAgent::enable()
+{
+    Ref agents = m_instrumentingAgents.get();
+    if (agents->enabledFrameRuntimeAgent() == this)
+        return { };
+
+    auto result = InspectorRuntimeAgent::enable();
+    if (!result)
+        return result;
+
+    reportExecutionContextCreation();
+    agents->setEnabledFrameRuntimeAgent(this);
+
+    return result;
+}
+
+CommandResult<void> FrameRuntimeAgent::disable()
+{
+    Ref { m_instrumentingAgents.get() }->setEnabledFrameRuntimeAgent(nullptr);
+
+    return InspectorRuntimeAgent::disable();
+}
+
+CommandResultOf<Ref<Inspector::Protocol::Runtime::RemoteObject>, std::optional<bool> /* wasThrown */, std::optional<int> /* savedResultIndex */> FrameRuntimeAgent::evaluate(const String& expression, const String& objectGroup, std::optional<bool>&& includeCommandLineAPI, std::optional<bool>&& doNotPauseOnExceptionsAndMuteConsole, std::optional<Inspector::Protocol::Runtime::ExecutionContextId>&& executionContextId, std::optional<bool>&& returnByValue, std::optional<bool>&& generatePreview, std::optional<bool>&& saveResult, std::optional<bool>&& emulateUserGesture)
+{
+    Inspector::Protocol::ErrorString errorString;
+
+    auto injectedScript = injectedScriptForEval(errorString, WTF::move(executionContextId));
+    if (injectedScript.hasNoValue())
+        return makeUnexpected(errorString);
+
+    RefPtr frame = m_inspectedFrame.get();
+    RefPtr page = frame ? frame->page() : nullptr;
+    if (!page)
+        return makeUnexpected("Frame or page not available"_s);
+
+    UserGestureEmulationScope userGestureScope(*page, emulateUserGesture.value_or(false), dynamicDowncast<Document>(executionContext(injectedScript.globalObject())));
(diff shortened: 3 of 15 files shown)


Source/WebCore/platform/mediastream/mac/ScreenCaptureKitCaptureSource.mm
AnujPanta1, line 1: wait these are random fixes I had locally, I will remove in the next round of reviews !

Source/WebCore/platform/mediastream/mac/ScreenCaptureKitCaptureSource.mm
```diff
@@ -415,12 +418,11 @@ - (void)outputVideoEffectDidStopForStream:(SCStream *)stream
     }
```
AnujPanta1, line 415: ditto will remove this, unrelated

Source/WebCore/inspector/agents/frame/FrameRuntimeAgent.cpp
```diff
@@ -0,0 +1,245 @@
...
+}
+
+void FrameRuntimeAgent::unmuteConsole()
+{
+    FrameConsoleClient::unmute();
+}
+
+String FrameRuntimeAgent::frameIdForProtocol() const
+{
+    return m_frameId;
```
burg: Once IdentifierRegistry is a thing, the code would be like:
```
return InspectorBackend::shared().identifierRegistry().frameId(m_inspectedFrame->identifier())
```

Source/WebCore/inspector/agents/frame/FrameRuntimeAgent.h
```diff
@@ -0,0 +1,76 @@
...
+    WTF_MAKE_NONCOPYABLE(FrameRuntimeAgent);
+    WTF_MAKE_TZONE_ALLOCATED(FrameRuntimeAgent);
+public:
+    FrameRuntimeAgent(FrameAgentContext&);
+    ~FrameRuntimeAgent();
+
+    // RuntimeBackendDispatcherHandler
+    Inspector::Protocol::ErrorStringOr<void> enable() override;
+    Inspector::Protocol::ErrorStringOr<void> disable() override;
+    Inspector::Protocol::ErrorStringOr<std::tuple<Ref<Inspector::Protocol::Runtime::RemoteObject>, std::optional<bool> /* wasThrown */, std::optional<int> /* savedResultIndex */>> evaluate(const String& expression, const String& objectGroup, std::optional<bool>&& includeCommandLineAPI, std::optional<bool>&& doNotPauseOnExceptionsAndMuteConsole, std::optional<Inspector::Protocol::Runtime::ExecutionContextId>&&, std::optional<bool>&& returnByValue, std::optional<bool>&& generatePreview, std::optional<bool>&& saveResult, std::optional<bool>&& emulateUserGesture) override;
```
burg: Please use CommandResult<T> or CommandResultOf<T, ...> for the complicated return types.

burg (commented):
This looks good overall. I will re-read in more detail once I look at the other agent PRs.

Is there really a need to instantiate Debugger if we don't do anything with it? I'd leave that part out or explain why the stub need to exist before it does anything.

How do we test this change?

Source/WebCore/inspector/agents/frame/FrameRuntimeAgent.cpp
```diff
@@ -0,0 +1,247 @@
...
+    if (!frame)
+        return;
+
+    if (!frame->script().canExecuteScripts(ReasonForCallingCanExecuteScripts::NotAboutToExecuteScript))
+        return;
+
+    // Don't report contexts if we're still on about:blank - wait for didClearWindowObjectInWorld after navigation
+    RefPtr document = frame->document();
+    if (document && document->url().isAboutBlank())
+        return;
```
the-chenergy, line 200: Is it possible for `document` to be null here, and are we OK with executing the code below if it's null?
```suggestion
    // Don't report contexts if we're still on about:blank - wait for didClearWindowObjectInWorld after navigation
    RefPtr document = frame->document();
    if (!document || document->url().isAboutBlank())
        return;
```

Source/WebCore/inspector/FrameInspectorController.cpp
```diff
@@ -44,6 +44,7 @@
 #include "Settings.h"
 #include "WebInjectedScriptHost.h"
 #include "WebInjectedScriptManager.h"
+#include "agents/frame/FrameRuntimeAgent.h"
```
the-chenergy:
```suggestion
#include "FrameRuntimeAgent.h"
```
And then alphabetize the includes

Source/WebCore/inspector/InspectorInstrumentation.cpp
```diff
@@ -108,6 +109,18 @@ void InspectorInstrumentation::didClearWindowObjectInWorldImpl(InstrumentingAgen
...
+            // Route to FrameRuntimeAgent instead of PageRuntimeAgent, but still call PageAgent for bootstrap scripts
+            if (auto* pageAgent = instrumentingAgents.enabledPageAgent())
+                pageAgent->didClearWindowObjectInWorld(frame, world);
+            return;
+        }
+    }
+
+    // Main frame or frames without FrameRuntimeAgent use PageRuntimeAgent
     if (auto* pageRuntimeAgent = instrumentingAgents.enabledPageRuntimeAgent())
         pageRuntimeAgent->didClearWindowObjectInWorld(frame, world);
```
the-chenergy: Will using the frame runtime agent also for the main frame be problematic? I think our pattern for backward compatibility is to use whether or not certain agents exist in the frame target to decide which agent to use. So, as long as we only create `FrameRuntimeAgent` in `FrameInspectorController` if SI is on, we should always count on using those if they're present rather than still use the page's agent

```suggestion
    if (auto* frameRuntimeAgent = frame.inspectorController().instrumentingAgents().enabledFrameRuntimeAgent()) {
        frameRuntimeAgent->didClearWindowObjectInWorld(world);
        // Route to FrameRuntimeAgent instead of PageRuntimeAgent, but still call PageAgent for bootstrap scripts
        if (auto* pageAgent = instrumentingAgents.enabledPageAgent())
            pageAgent->didClearWindowObjectInWorld(frame, world);
        return;
    } else {
	    // Frames without FrameRuntimeAgent use PageRuntimeAgent
		if (auto* pageRuntimeAgent = instrumentingAgents.enabledPageRuntimeAgent())
			pageRuntimeAgent->didClearWindowObjectInWorld(frame, world);
	}
```

AnujPanta1:
@the-chenergy thank you, I applied all suggestions

Source/WebCore/inspector/InspectorInstrumentation.cpp
```diff
@@ -108,6 +109,13 @@ void InspectorInstrumentation::didClearWindowObjectInWorldImpl(InstrumentingAgen
...
 
+    if (auto* frameRuntimeAgent = frame.inspectorController().instrumentingAgents().enabledFrameRuntimeAgent()) {
+        frameRuntimeAgent->didClearWindowObjectInWorld(world);
+        if (auto* pageAgent = instrumentingAgents.enabledPageAgent())
+            pageAgent->didClearWindowObjectInWorld(frame, world);
+        return;
+    }
+
     if (auto* pageRuntimeAgent = instrumentingAgents.enabledPageRuntimeAgent())
         pageRuntimeAgent->didClearWindowObjectInWorld(frame, world);
```
the-chenergy, line 120: Do we still have to do this if frame targets and their runtime agents are used? I think it'd be nice if SI is on, we completely use the new code path, as if the entire domain from the page target is deprecated. That separation should make the eventual removal of the page target easier

Source/WebInspectorUI/UserInterface/Models/ExecutionContextList.js
```diff
@@ -48,13 +48,21 @@ WI.ExecutionContextList = class ExecutionContextList
...
+            }
         }
 
         this._contexts.push(context);
 
-        if (context.type === WI.ExecutionContext.Type.Normal && context.target.type === WI.TargetType.Page) {
+        // Set pageExecutionContext for both Page and Frame targets
+        if (context.type === WI.ExecutionContext.Type.Normal && (context.target.type === WI.TargetType.Page || context.target.type === WI.TargetType.Frame)) {
             console.assert(!this._pageExecutionContext);
             this._pageExecutionContext = context;
```
the-chenergy: In the spirit of separating the frame target from soon-to-be legacy stuff: do we really need to manage a `pageExecutionContext` for a frame? How is frame's `pageExecutionContext` actually used? 😅 

Same comment applies to the event `WI.FrameTarget.Event.PageExecutionContextChanged`

Source/WebCore/inspector/InspectorInstrumentation.cpp
burg: I think this would prevent using PageRuntimeAgent as a fallback for FrameRuntimeAgent that coexists at the same time- it won't have the data from instrumentation. Is there actually any case where Runtime can't handle frame target commands after this PR?

Source/WebCore/inspector/agents/frame/FrameRuntimeAgent.cpp
```diff
@@ -0,0 +1,247 @@
...
+
+WTF_MAKE_TZONE_ALLOCATED_IMPL(FrameRuntimeAgent);
+
+FrameRuntimeAgent::FrameRuntimeAgent(FrameAgentContext& context)
+    : InspectorRuntimeAgent(context)
+    , m_frontendDispatcher(makeUniqueRef<Inspector::RuntimeFrontendDispatcher>(context.frontendRouter))
+    , m_backendDispatcher(Inspector::RuntimeBackendDispatcher::create(Ref { context.backendDispatcher }, this))
+    , m_instrumentingAgents(context.instrumentingAgents)
+    , m_inspectedFrame(context.inspectedFrame)
+    , m_frameId(makeString("frame-"_s, context.inspectedFrame->frameID().toUInt64()))
```
burg: Is it possible to keep the actual FrameIdentifier as the member variable and convert to string when needed?

Source/WebCore/inspector/agents/frame/FrameRuntimeAgent.cpp
```diff
@@ -0,0 +1,247 @@
...
+    Ref agents = m_instrumentingAgents.get();
+    if (agents->enabledFrameRuntimeAgent() == this)
+        return { };
+
+    auto result = InspectorRuntimeAgent::enable();
+    if (!result)
+        return result;
+
+    reportExecutionContextCreation();
+
```
burg: Nit: extra newline

Source/WebCore/inspector/agents/frame/FrameRuntimeAgent.cpp
```diff
@@ -0,0 +1,247 @@
...
+    agents->setEnabledFrameRuntimeAgent(this);
+
+    return result;
+}
+
+Inspector::CommandResult<void> FrameRuntimeAgent::disable()
```
burg: (applies throughout) In the .cpp file, this should not need Inspector:: namespacing. add a `using Inspector` inside the `namespace {}`.

Source/WebInspectorUI/UserInterface/Protocol/FrameTarget.js
```diff
@@ -29,6 +29,38 @@ WI.FrameTarget = class FrameTarget extends WI.Target
...
+    // Public
+
+    get executionContextList()
+    {
+        return this._executionContextList;
+    }
+
+    get executionContext()
+    {
+        return this._executionContextList.pageExecutionContext;
```
burg: So are we actually using the new context? I agree we should not keep pageExecutionContext tracking in this target unless there's a need.

Source/WebCore/inspector/agents/frame/FrameRuntimeAgent.cpp
```diff
@@ -0,0 +1,247 @@
+/*
+ * Copyright (C) 2025 Apple Inc. All rights reserved.
```
burg: Nit: 2026

Source/WebInspectorUI/UserInterface/Views/QuickConsole.js
```diff
@@ -317,6 +326,37 @@ WI.QuickConsole = class QuickConsole extends WI.View
                     addExecutionContextsForWorker(workerTarget, 1);
             }
         }
+
+        // Add FrameTarget execution contexts (for site-isolated frames)
+        let frameTargets = WI.targetManager.targets.filter((target) => target instanceof WI.FrameTarget && target.executionContext);
```
burg: Nit: Redundant computation of frameTargets

burg (approved):

Source/WebCore/inspector/FrameInspectorController.cpp
```diff
@@ -124,12 +126,36 @@ void FrameInspectorController::createConsoleAgent()
...
 void FrameInspectorController::createLazyAgents()
 {
     if (m_didCreateLazyAgents)
         return;
 
     m_didCreateLazyAgents = true;
+
+    RefPtr frame = m_frame.get();
+    if (!frame || !protect(frame->settings())->siteIsolationEnabled())
+        return;
```
the-chenergy, line 152: I'm curious about why this was still needed—shouldn't this only be called only when SI is on, as only with SI on we expose the `FrameInspectorTarget` to the frontend in `WebPageInspectorController`, and thus only then we might call `FrameInspectorController::connectFrontend`? I was hoping we could even add an `ASSERT(siteIsolationEnabled())` here

PR #58276, opened 2026-02-10
Implement overflow-clip-margin parsing.

surajt6:
#### 1e0a89e6c89b21d3451df681f148d659bd70a51f
<pre>
Implement overflow-clip-margin parsing.

<a href="https://bugs.webkit.org/show_bug.cgi?id=308225">https://bugs.webkit.org/show_bug.cgi?id=308225</a>

Reviewed by Sam Weinig.

Parse overflow-clip-margin CSS property value behind a feature flag.

Added additional tests that use calc().

* LayoutTests/imported/w3c/web-platform-tests/css/css-cascade/all-prop-initial-xml-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/css/css-cascade/all-prop-revert-layer-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/css/css-overflow/overflow-no-interpolation-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/css/css-overflow/parsing/overflow-clip-margin-computed-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/css/css-overflow/parsing/overflow-clip-margin-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/css/css-overflow/parsing/overflow-clip-margin.html:
* LayoutTests/imported/w3c/web-platform-tests/css/css-typed-om/the-stylepropertymap/properties/overflow-clip-margin-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/css/css-typed-om/the-stylepropertymap/properties/overflow-clip-margin.html:
* LayoutTests/platform/glib/imported/w3c/web-platform-tests/css/css-cascade/all-prop-initial-xml-expected.txt:
* Source/WebCore/Sources.txt:
* Source/WebCore/css/CSSPrimitiveValueMappings.h:
* Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp: Added.
(WebCore::CSSPropertyParserHelpers::consumeOverflowClipMargin):
* Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.h: Added.
* Source/WebCore/css/parser/CSSPropertyParserCustom.h:
* Source/WebCore/rendering/style/RenderStyleConstants.cpp:
(WebCore::operator&lt;&lt;):
* Source/WebCore/rendering/style/RenderStyleConstants.h:
* Source/WebCore/style/StyleBuilderCustom.h:
(WebCore::Style::forwardInheritedValue):
* Source/WebCore/style/computed/data/StyleNonInheritedRareData.cpp:
(WebCore::Style::NonInheritedRareData::NonInheritedRareData):
(WebCore::Style::NonInheritedRareData::operator== const):
(WebCore::Style::NonInheritedRareData::dumpDifferences const):
* Source/WebCore/style/computed/data/StyleNonInheritedRareData.h:
* Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp: Added.
(WebCore::Style::CSSValueConversion&lt;OverflowClipMargin&gt;::operator):
* Source/WebCore/style/values/overflow/StyleOverflowClipMargin.h: Added.
(WebCore::Style::OverflowClipMargin::OverflowClipMargin):
(WebCore::Style::OverflowClipMargin::switchOn const):

Canonical link: <a href="https://commits.webkit.org/307956@main">https://commits.webkit.org/307956@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/3f38288d7b3d3ad85ddd95c97a42ba5598247dd8

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/145712 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/18394 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/10341 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/154384 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/99348 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/147587 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/18879 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/18287 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/112064 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/80286 "1 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/148675 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/14433 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/130905 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/92970 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/13751 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/11516 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/1831 "Built successfully") | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/137716 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/123302 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/7747 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/156697 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/6521 "Built successfully and passed tests") | [⏳ 🛠 ios-safer-cpp ](https://ews-build.webkit.org/#/builders/Apple-iOS-26-Safer-CPP-Checks-EWS "Waiting in queue, processing has not started yet") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/8900 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/120075 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/18244 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/15225 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/120418 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30929 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/18290 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/129061 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/73985 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/16144 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/7161 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/177015 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/17866 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/81645 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/45460 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/17602 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/17824 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/17665 "Built successfully") | | |


Ahmad-S792:
@surajt6 - Congratulations on your first PR on WebKit, please don't hesitate to ask any question and you can join WebKit OpenSource Slack channel for any help or input as well.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.h
```diff
@@ -0,0 +1,120 @@
...
+        }
+        RELEASE_ASSERT_NOT_REACHED();
+    }
+
+    bool operator==(const OverflowClipMargin&) const = default;
+
+private:
+    OverflowClipMarginType m_type;
+    Length m_length;
+    OverflowClipReferenceBox m_clipBox;
```
weinig: I realize this is still a draft, so perhaps you were going to change this, but this should use something like the following for the underlying type:

`Variant<OverflowClipReferenceBox, Length, SpaceSeparatedTuple<OverflowClipReferenceBox, Length>>`

That way, only the valid state is initialized.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.h
```diff
@@ -0,0 +1,120 @@
...
+
+#include <WebCore/CSSValueKeywords.h>
+#include <WebCore/RenderStyleConstants.h>
+#include <WebCore/StylePrimitiveNumericTypes.h>
+#include <WebCore/StyleValueTypes.h>
+
+namespace WebCore {
+namespace Style {
+
+struct OverflowClipMargin {
```
weinig, line 38: Please add a spec link and the grammar here as a comment.

```suggestion
// <'overflow-clip-margin'> = <visual-box> || <length [0,∞]>
// https://drafts.csswg.org/css-overflow/#propdef-overflow-clip-margin
struct OverflowClipMargin {
```

Source/WebCore/rendering/style/RenderStyleConstants.h
```diff
@@ -234,6 +234,18 @@ enum class Overflow : uint8_t {
     PagedY
 };
 
+enum class OverflowClipReferenceBox : uint8_t {
+    BorderBox,
+    ContentBox,
+    PaddingBox
+};
+
+enum class OverflowClipMarginType : uint8_t {
```
weinig: You shouldn't need this if you use a Variant, but even if you decide not to, this can be internal and private to OverflowClipMargin.

Source/WebCore/rendering/style/RenderStyleConstants.h
```diff
@@ -1208,6 +1220,8 @@ WTF::TextStream& operator<<(WTF::TextStream&, OutlineStyle);
 WTF::TextStream& operator<<(WTF::TextStream&, WebCore::Overflow);
 WTF::TextStream& operator<<(WTF::TextStream&, OverflowAlignment);
 WTF::TextStream& operator<<(WTF::TextStream&, OverflowWrap);
+WTF::TextStream& operator<<(WTF::TextStream&, OverflowClipReferenceBox);
```
weinig: These aren't needed for types that implemented in CSSPrimitiveValueMappings.h‎ using the macros. (I haven't yet removed all the ones I could here, but this is a good reminder to me to do so).

Source/WebCore/rendering/style/RenderStyleConstants.h
```diff
@@ -234,6 +234,18 @@ enum class Overflow : uint8_t {
     PagedY
 };
 
+enum class OverflowClipReferenceBox : uint8_t {
```
weinig: I would consider calling this VisualBox, and adding a spec link to https://drafts.csswg.org/css-box-4/#typedef-visual-box.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "StyleBuilderChecking.h"
+
+namespace WebCore {
+namespace Style {
+
+using namespace CSS::Literals;
+
+auto CSSValueConversion<OverflowClipMargin>::operator()(BuilderState& state, const CSSValue& value) -> OverflowClipMargin
+{
+    if (RefPtr primitiveValue = requiredDowncast<CSSPrimitiveValue>(state, value)) {
```
weinig: This should be a dynamic_downcast. `requiredDowncast` will tell the builder the conversion failed if the cast isn't successful. That means the CSSValueList code below will never be successfully run.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
```diff
@@ -0,0 +1,83 @@
...
+        if (primitiveValue->isValueID()) {
+            auto overflowClipReferenceBox = fromCSSValue<OverflowClipReferenceBox>(*primitiveValue);
+            switch (overflowClipReferenceBox) {
+            case OverflowClipReferenceBox::PaddingBox:
+                return { 0_css_px };
+            case OverflowClipReferenceBox::BorderBox:
+            case OverflowClipReferenceBox::ContentBox:
+                return { overflowClipReferenceBox };
+            };
+        }
```
weinig: After this, you should add: 

```suggestion
        }

        state.setCurrentPropertyInvalidAtComputedValueTime();
        return 0_css_px;
```

This will make sure that any other types of primitive values force an invalid style resolution.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    if (auto list = requiredListDowncast<CSSValueList, CSSValue, 2>(state, value)) {
```
weinig: This can be written a bit more succinctly using an early return and a more specific subtype.

```suggestion
    auto list = requiredListDowncast<CSSValueList, CSSPrimitiveValue, 2>(state, value);
    if (!list)
        return 0_css_px;
```

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
```diff
@@ -0,0 +1,83 @@
...
+        Ref item0 = list->item(0);
+        Ref item1 = list->item(1);
+
+        RefPtr value0 = requiredDowncast<CSSPrimitiveValue>(state, item0);
+        RefPtr value1 = requiredDowncast<CSSPrimitiveValue>(state, item1);
+
+        if (value0->isValueID() && value1->isLength()) {
```
weinig: In cases like these, we allow the values to be in any order. So this could be written as:

```c++
std::optional<OverflowClipReferenceBox> referenceBox;
std::optional<OverflowClipMargin::Length> length;

for (Ref item : *list) {
    if (item->isValueID()) {
        if (referenceBox) {
            state.setCurrentPropertyInvalidAtComputedValueTime();
            return 0_css_px;
        }
        referenceBox = toStyleFromCSSValue<OverflowClipReferenceBox>(state, item);
    } else if (item->isLength()) {
        if (length) {
            state.setCurrentPropertyInvalidAtComputedValueTime();
            return 0_css_px;
        }
        length = toStyleFromCSSValue<OverflowClipMargin::Length>(state, item);
    } else {
        state.setCurrentPropertyInvalidAtComputedValueTime();
        return 0_css_px;
    }
}

return { *referenceBox, *length };
```

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
```diff
@@ -0,0 +1,83 @@
...
+{
+    if (RefPtr primitiveValue = requiredDowncast<CSSPrimitiveValue>(state, value)) {
+        if (primitiveValue->isLength())
+            return { toStyleFromCSSValue<OverflowClipMargin::Length>(state, *primitiveValue) };
+
+        if (primitiveValue->isValueID()) {
+            auto overflowClipReferenceBox = fromCSSValue<OverflowClipReferenceBox>(*primitiveValue);
+            switch (overflowClipReferenceBox) {
+            case OverflowClipReferenceBox::PaddingBox:
+                return { 0_css_px };
```
weinig: You probably don't want to do this. Even if the behavior will end up being that it should act like a length of 0px is set, the resolved value (that is what the value returned from getComputedStyle is called) will still need to be `padding-box`.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.h
surajt6: `Variant`'s a lot better than what I wrote. I was mostly following `StyleAspectRatio.h` and `StyleAspectRatio.cpp` for my implementation. I'll change it. It makes the rest of the Class a lot cleaner.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.h
surajt6: Sure, I'll add it.

Source/WebCore/rendering/style/RenderStyleConstants.h
surajt6: I'll use variant and remove this.

Source/WebCore/rendering/style/RenderStyleConstants.h
surajt6: Great suggestion! I'll add VisualBox and update it here.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
surajt6: Makes sense. I'll update it.

Source/WebCore/rendering/style/RenderStyleConstants.h
surajt6: Sure, I'll remove it.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
surajt6: Sure, I'll update it. I want to confirm my understanding of what this actually does. Essentially the `setCurrentPropertyInvalidAtComputedValueTime` will make sure that the `overflow-clip-margin`'s computed value is set to `0px` if it's invalid. 

Take for example `StyleBorderRadius.cpp`. Why is `setCurrentPropertyInvalidAtComputedValueTime` not called here - https://searchfox.org/wubkat/source/Source/WebCore/style/values/borders/StyleBorderRadius.cpp#71 ?

And for `StyleWebKitTextStrokeWidth.cpp`, it's only called once in line 58 but not in line 42 -
https://searchfox.org/wubkat/source/Source/WebCore/style/values/non-standard/StyleWebKitTextStrokeWidth.cpp#38. Why is that the case?

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
surajt6: That's a lot cleaner. I'll update it.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
surajt6: Sure I'll update it. I'm curious as to when the value order is reversed. The auto generated parser makes sure that the CSSValueList always has referenceBox as first value. In what situation would the CSSValueList have length as the first value?

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
surajt6: Yup that makes a lot of sense. I realized later that this code has nothing to do with CSS value canonicalization.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
weinig: The setCurrentPropertyInvalidAtComputedValueTime() stuff is largely "belt and suspenders" precautions against a specific weird edge case where rather than the CSSValue coming from the CSSParser, it comes from the CSS typedom. This can happen by using things like the `StylePropertyMap` interface to set CSS values. There are plans to fix this in a more comprehensive way, but for now, being conservative in CSSValueConversion like this works.

> Sure, I'll update it. I want to confirm my understanding of what this actually does. Essentially the setCurrentPropertyInvalidAtComputedValueTime will make sure that the overflow-clip-margin's computed value is set to 0px if it's invalid.

Not quite. When setCurrentPropertyInvalidAtComputedValueTime() is used, it sets a bit in the style builder that tells it that no matter what value ends up being set, it should be unset. If you look in Style::Builder::applyProperty(), at the end of the function, there is logic that checks isCurrentPropertyInvalidAtComputedValueTime() and applies the unset.

We usually return the default value for the property when we call setCurrentPropertyInvalidAtComputedValueTime() just for simplicity. It doesn't really matter what is returned in those cases.

> Take for example StyleBorderRadius.cpp. Why is setCurrentPropertyInvalidAtComputedValueTime not called here - https://searchfox.org/wubkat/source/Source/WebCore/style/values/borders/StyleBorderRadius.cpp#71 ?

I think that check for `if (!value.isPair())` is probably unnecessary and wrong, since the requiredPairDowncast will perform the same check.

> And for StyleWebKitTextStrokeWidth.cpp, it's only called once in line 58 but not in line 42 -
https://searchfox.org/wubkat/source/Source/WebCore/style/values/non-standard/StyleWebKitTextStrokeWidth.cpp#38. Why is that the case?

It's not necessary to call it for line 42 because that is what the  requiredDowncast<CSSPrimitiveValue>() above it did.  `requiredDowncast` is a `dynamicDowncast` that in addition to checking the type, calls setCurrentPropertyInvalidAtComputedValueTime() if the type check fails.

We need to do it on line 58 because we are narrowing our check beyond just CSSPrimitiveValue, and we don't have a specific "required" function for that case.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
weinig: I'd have to dive deeper to find out if there is a way to get it to happen, but the short answer is that the CSS typedom (e.g. StylePropertyMap) can also be the source of the CSSValue passed to CSSValueConversion. For a case like this, there would have to be another property had a similar (but with the reverse default ordering) grammar, and allowed `<length> <visual-box>`. If such a property existed, one could do something like:

```
      var element = document.getElementById("element");
      element.attributeStyleMap.set('overflow-clip-margin', CSSStyleValue.parse('made-up-property', '30px border-box'))
```

As a result of this other source, we choose to be conservative in the CSSValueConversion codepath.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
surajt6: Ah, I get it now. Thanks for the clear explanation.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
surajt6: Got it.

surajt6:
@weinig I have updated the PR with a new commit and now this PR only includes the CSS property parsing logic to keep it small. The parsing is also behind a feature flag. Style computation and layout/rendering will for overflow-clip-margin will be implemented in future PRs. I also went ahead and updated some css-overflow layout tests related to the overflow-clip-margin property. I've tried to incorporate all of the feedback in earlier revisions. Thanks again for taking the time to answer my questions and reviewing this PR. Cheers!

Source/WebCore/css/CSSProperties.json
```diff
@@ -7309,6 +7309,25 @@
                 "url": "https://drafts.csswg.org/css-overflow-3/#propdef-overflow-block"
             }
         },
+        "overflow-clip-margin": {
+            "animation-type": "not animatable",
```
weinig, line 7313: Please add the following here:

            "animation-type-comment": "FIXME: Current spec animation type is 'see prose'",

Source/WebCore/css/CSSProperties.json
```diff
@@ -7309,6 +7309,25 @@
...
+            "inherited": false,
+            "initial": "0px",
+            "values": ["border-box", "content-box", "padding-box"],
```
weinig: Our convention is put each of these elements in the array on their own line.

Source/WebCore/css/CSSProperties.json
```diff
@@ -7309,6 +7309,25 @@
...
+            "codegen-properties": {
+                "computed-style-storage-path": ["m_nonInheritedData", "rareData"],
+                "computed-style-storage-kind": "reference",
+                "computed-style-type": "Style::OverflowClipMargin",
+                "parser-grammar-unused": "<visual-box> || <length [0,inf]>",
+                "parser-function": "consumeOverflowClipMargin",
+                "parser-grammar-unused-reason": "Needs support for transforming parsed input to minimal form and default values in '||' groups.",
```
weinig, line 7328: Please put the "parser-grammar-unused-reason" immediately following the "parser-grammar-unused".

Source/WebCore/css/CSSProperties.json
```diff
@@ -14343,6 +14362,14 @@
                 "url": "https://drafts.fxtf.org/css-masking-1/#typedef-geometry-box"
             }
         },
+        "<visual-box>": {
+          "grammar": "<box>",
```
weinig: I would put the full grammar here. `<box>` isn't even in a spec anymore, so others should probably be using `<visual-box>` going forward.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.h
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    template<typename... F> decltype(auto) switchOn(F&&... f) const
+    {
+        return WTF::switchOn(m_value, std::forward<F>(f)...);
+    }
+
+    bool operator==(const OverflowClipMargin&) const = default;
+private:
+
```
weinig, line 52: Please remove this newline.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.h
```diff
@@ -0,0 +1,67 @@
...
+        : m_value { length }
+    {
+    }
+
+    template<typename... F> decltype(auto) switchOn(F&&... f) const
+    {
+        return WTF::switchOn(m_value, std::forward<F>(f)...);
+    }
+
+    bool operator==(const OverflowClipMargin&) const = default;
```
weinig, line 51: Please add a newline under this.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.h
```diff
@@ -0,0 +1,67 @@
...
+#include <WebCore/StyleValueTypes.h>
+
+namespace WebCore {
+namespace Style {
+
+// <'overflow-clip-margin'> = <visual-box> || <length [0,∞]>
+// https://drafts.csswg.org/css-overflow/#overflow-clip-margin
+struct OverflowClipMargin {
+    using Length = Style::Length<CSS::Nonnegative, float>;
+    using VisualBox = CSSBoxType;
```
weinig: CSSBoxType is really too broad for this. I would stick with adding a new VisualBox type.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.h
```diff
@@ -0,0 +1,67 @@
...
+
+    OverflowClipMargin(CSS::ValueLiteral<CSS::LengthUnit::Px> length)
```
weinig, line 41: You are going to need more constructors. If you want to leave that for the next change, that is fine.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.cpp
```diff
@@ -0,0 +1,41 @@
...
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "StyleOverflowClipMargin.h"
+
+namespace WebCore {
+namespace Style {
+
+using namespace CSS::Literals;
```
weinig: I would move this to be in the body of CSSValueConversion<OverflowClipMargin>::operator() to constrain it.

weinig (approved):

Source/WebCore/css/parser/CSSPropertyParserCustom.h
```diff
@@ -72,6 +72,7 @@
 #include "CSSPropertyParserConsumer+Masking.h"
 #include "CSSPropertyParserConsumer+Motion.h"
 #include "CSSPropertyParserConsumer+NumberDefinitions.h"
+#include "CSSPropertyParserConsumer+Overflow.h"
```
weinig, line 75: Why is this needed?

Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+#include "config.h"
+#include "CSSPropertyParserConsumer+Overflow.h"
+
+#include "CSSParserTokenRange.h"
+#include "CSSPrimitiveValue.h"
+#include "CSSPropertyParserConsumer+CSSPrimitiveValueResolver.h"
+#include "CSSPropertyParserConsumer+Primitives.h"
+#include "CSSPropertyParserConsumer+Shapes.h"
```
weinig: Is this include needed?

Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include "CSSPropertyParserConsumer+URL.h"
```
weinig: Is this include needed?

Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp
```diff
@@ -0,0 +1,90 @@
...
+        length = consumeLength(range, state);
+        return !!length;
+    };
+
+    for (size_t i = 0; i < 2 && !range.atEnd(); ++i) {
+        if (tryConsumeVisualBox(range) || tryConsumeLength(range, state))
+            continue;
+        break;
+    }
+
```
weinig, line 65: You need to check that at least one of the values got parsed here.

```
if (!visualBox && !length)
    return nullptr;
```

Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp
```diff
@@ -0,0 +1,90 @@
...
+    CSSValueListBuilder list;
+    // Default value is padding-box
+    if (visualBox && !isValueID(visualBox, CSSValueID::CSSValuePaddingBox))
+        list.append(visualBox.releaseNonNull());
+
+    // Default value is 0px
+    if (length && !length->isZero().value_or(true))
```
weinig: Why is `value_or(true)` right here?

weinig (changes requested):

Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp
weinig: (Note, isZero() will return nullopt when it can't be determined if the value is zero, like when calc() is used).

Source/WebCore/css/CSSProperties.json
surajt6: Sure, will do.

Source/WebCore/css/CSSProperties.json
surajt6: I'll update it.

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.h
surajt6: Sure.

Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
```diff
@@ -1299,6 +1299,20 @@ CSSMathDepthEnabled:
       default: true
     WebCore:
       default: true
+      
+CSSOverflowClipMarginEnabled:
+  type: bool
+  status: testable
+  category: css
+  humanReadableName: "CSS Overflow Clip Margin"
+  humanReadableDescription: "Enable CSS Overflow Clip Margin"
```
nt1m:
```suggestion
  humanReadableName: "CSS overflow-clip-margin"
  humanReadableDescription: "Enable CSS overflow-clip-margin"
```

small nit to match the style of the others

Source/WebCore/style/values/overflow/StyleOverflowClipMargin.h
surajt6: I'll leave that for the next change.

Source/WebCore/css/parser/CSSPropertyParserCustom.h
surajt6: Its needed because `consumeOverflowClipMargin` is used in the auto generated `CSSPropertyParsing.cpp` as follows -

```
    case CSSPropertyID::CSSPropertyOverflowAnchor:
        return consumeIdent(range, isKeywordValidForOverflowAnchor);
    case CSSPropertyID::CSSPropertyOverflowClipMargin:
        return consumeOverflowClipMargin(range, state);
```

`CSSPropertyParsing.cpp` includes `CSSPropertyParserCustom.h`. If I don't include `CSSPropertyParserConsumer+Overflow.h`, the build fails with a compilation error.

Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp
surajt6: These are not needed. I'll remove them.

Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp
surajt6: Nope. I'll remove them.

Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp
surajt6: This is the code that's below the the `for` loop
```
CSSValueListBuilder list;
// Default value is padding-box
if (visualBox && !isValueID(visualBox, CSSValueID::CSSValuePaddingBox))
    list.append(visualBox.releaseNonNull());

// Default value is 0px
if (length && !length->isZero().value_or(true))
    list.append(length.releaseNonNull());

if (list.isEmpty())
    return { CSSPrimitiveValue::create(0, CSSUnitType::CSS_PX) };
```

Isn't the `list.isEmpty()` sufficient to check that at least one of the values got parsed? Or is there a specific reason we want return a `nullptr` when both of them empty?

Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
surajt6: Sure, I'll update the text.

Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp
surajt6: I initially wrote `value_or(true)` because if length is empty then we would not want to add it to the list. But now I realize that it won't work when `calc()` is used. It should be `value_or(false)`. I'll change it.

LayoutTests/imported/w3c/web-platform-tests/css/css-overflow/parsing/overflow-clip-margin.html
```diff
@@ -27,6 +27,17 @@
...
+test_valid_value("overflow-clip-margin", "calc(100px - 50px)", "calc(50px)");
+test_valid_value("overflow-clip-margin", "calc(100px - 100px)", "calc(0px)");
+test_valid_value("overflow-clip-margin", "calc(0.5em - 100px)");
+test_invalid_value("overflow-clip-margin", "calc(100% - 0.5em)");
+test_invalid_value("overflow-clip-margin", "calc(100% - 50px)");
+
+test_valid_value("overflow-clip-margin", 'padding-box calc(0.5em - 100px)', 'calc(0.5em - 100px)');
+test_valid_value("overflow-clip-margin", 'padding-box calc(100px - 100px)', 'calc(0px)');
+test_valid_value("overflow-clip-margin", 'border-box calc(0.5em - 100px)', 'border-box calc(0.5em - 100px)');
+test_invalid_value("overflow-clip-margin", 'border-box calc(0.5em - 100%)');
```
surajt6, line 39: These tests mimic the output of Firefox Dev Edition. I added the tests under the `LayoutTests/imported` directory. Please let me know if I need to add them in a new file in a different location.

Source/WebCore/rendering/style/RenderStyleConstants.h
```diff
@@ -1232,5 +1238,6 @@ WTF::TextStream& operator<<(WTF::TextStream&, MaskType);
 WTF::TextStream& operator<<(WTF::TextStream&, ShapeRendering);
 WTF::TextStream& operator<<(WTF::TextStream&, TextAnchor);
 WTF::TextStream& operator<<(WTF::TextStream&, VectorEffect);
+WTF::TextStream& operator<<(WTF::TextStream&, VisualBox);
```
surajt6, line 1241: This is needed. When I don't add it, the build fails with the following error - 

```
/Volumes/wk/WebKit/Source/WebCore/css/values/CSSValueAggregates.h:1798:54: error: invalid operands to binary expression
      ('TextStream' and 'const WebCore::VisualBox')
 1798 |     WTF::switchOn(value, [&](const auto& value) { ts << value; });
```

surajt6:
I apologize for the silly build failures in EWS. The build was passing on my Mac. I did a clean and rebuild to try to reproduce the error but couldn't. The build failures seem simple to fix after reading through the error logs. In addition, even though the Mac build passed on EWS, there are a few layout test failures. I'm working on fixing the layout tests and the build failures on iOS, visionOS, tvOS & watchOS platforms. I will push another revision with the fixes. @weinig - your feedback is still welcome while I work through the EWS errors.

LayoutTests/imported/w3c/web-platform-tests/css/css-cascade/all-prop-revert-layer-expected.txt
```diff
@@ -232,6 +232,7 @@ PASS outline-style
 PASS outline-width
 PASS overflow-anchor
 PASS overflow-block
+FAIL overflow-clip-margin assert_not_equals: Should get a different computed value. got disallowed value "0px"
```
surajt6, line 235: This will pass once I add support for overflow-clip-margin computed values.

LayoutTests/imported/w3c/web-platform-tests/css/css-typed-om/the-stylepropertymap/properties/overflow-clip-margin-expected.txt
```diff
@@ -1,9 +1,38 @@
...
+PASS Can set 'overflow-clip-margin' to CSS-wide keywords: initial
+PASS Can set 'overflow-clip-margin' to CSS-wide keywords: inherit
+PASS Can set 'overflow-clip-margin' to CSS-wide keywords: unset
+PASS Can set 'overflow-clip-margin' to CSS-wide keywords: revert
+PASS Can set 'overflow-clip-margin' to var() references:  var(--A)
+FAIL Can set 'overflow-clip-margin' to the 'border-box' keyword: border-box assert_equals: expected "CSSKeywordValue" but got "CSSUnitValue"
+FAIL Can set 'overflow-clip-margin' to the 'content-box' keyword: content-box assert_equals: expected "CSSKeywordValue" but got "CSSUnitValue"
+FAIL Can set 'overflow-clip-margin' to the 'padding-box' keyword: padding-box assert_equals: expected "CSSKeywordValue" but got "CSSUnitValue"
+PASS Can set 'overflow-clip-margin' to a length: 0px
+FAIL Can set 'overflow-clip-margin' to a length: -3.14em Invalid values
```
surajt6, line 11: Setting to `overflow-clip-margin` to -3.14em results in `Type Error: Invalid Values` as it only accepts non-negative values. This behavior is consistent with Chrome and Firefox as well. `testsuite.js` needs to be update to treat this test as a PASS. I added a FIXME in the test case file to fix this issue.

Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp
surajt6: I realized later that we need this. I added it in the new revision.

Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp
```diff
@@ -0,0 +1,88 @@
...
+        if (visualBox)
+            return false;
+        visualBox = CSSPropertyParsing::consumeVisualBox(range);
+        return !!visualBox;
+    };
+
+    RefPtr<CSSPrimitiveValue> length;
+    auto tryConsumeLength = [&length](CSSParserTokenRange& range, CSS::PropertyParserState& state) -> bool {
+        auto consumeLength = [](CSSParserTokenRange& range, CSS::PropertyParserState& state) -> RefPtr<CSSPrimitiveValue> {
+            return CSSPrimitiveValueResolver<CSS::Length<CSS::Range { 0, CSS::Range::infinity } >>::consumeAndResolve(range, state, { .unitlessZeroLength = UnitlessZeroQuirk::Allow });
```
weinig: This can be simplified down to

```
            return CSSPrimitiveValueResolver<CSS::Length<CSS::Nonnegative>>::consumeAndResolve(range, state);
```

`unitlessZeroLength = UnitlessZeroQuirk::Allow` is the default.

Source/WebCore/rendering/style/RenderStyleConstants.h
weinig: If you add a mapping in CSSPrimitiveValueMappings.h it won't be. That will probably be something you need to add in the next PR though. So feel free to keep this and then remove it later.

weinig (approved):

Source/WebCore/css/parser/CSSPropertyParserConsumer+Overflow.cpp
surajt6: I will update this in the next PR as I would prefer not to push a new revision with just this minor change.

PR #58370, opened 2026-02-11
Focus cannot advance from one remote frame to a sibling remote frame by pressing 'tab'.

megangardner:
#### 4f9f29fa8bbfd43ef01a66205b59b6e1dd6caad6
<pre>
Focus cannot advance from one remote frame to a sibling remote frame by pressing &apos;tab&apos;.
<a href="https://bugs.webkit.org/show_bug.cgi?id=307510">https://bugs.webkit.org/show_bug.cgi?id=307510</a>
<a href="https://rdar.apple.com/170108802">rdar://170108802</a>

Reviewed by Ryosuke Niwa.

When we are trying to advance focus and reach the end of a frame
and don&apos;t find anything to focus, but we know we have a parent
frame, we need to continue searching in the scope of that parent
frame. I think there is more work to be done for frames that have
no focusable elements, but this is still forward progress on remote
frame focus.

Test: Tools/TestWebKitAPI/Tests/WebKitCocoa/FocusWebView.mm

(WebCore::ChromeClient::frameOwnerElementForFrameID const):
* Source/WebCore/page/FocusController.cpp:
(WebCore::FocusNavigationScope::owner const):
(WebCore::FocusController::findFocusableElementContinuingFromFrame):
(WebCore::FocusController::findFocusableElementAcrossFocusScope):
(WebCore::FocusController::findFocusableElementWithinScope):
* Source/WebCore/page/FocusController.h:
* Source/WebCore/page/Frame.cpp:
(WebCore::Frame::parent const):
* Source/WebCore/page/Frame.h:
* Source/WebCore/page/RemoteFrameClient.h:
(WebKit::WebFrameProxy::findFocusableElementContinuingFromFrame):
* Source/WebKit/UIProcess/WebFrameProxy.messages.in:
(WebKit::WebChromeClient::frameOwnerElementForFrameID const):
* Source/WebKit/WebProcess/WebCoreSupport/WebRemoteFrameClient.cpp:
(WebKit::WebRemoteFrameClient::findFocusableElementContinuingFromFrame):
* Source/WebKit/WebProcess/WebCoreSupport/WebRemoteFrameClient.h:
* Source/WebKit/WebProcess/WebPage/WebFrame.cpp:
(WebKit::WebFrame::findFocusableElementContinuingFromFrame):
* Source/WebKit/WebProcess/WebPage/WebFrame.h:
* Source/WebKit/WebProcess/WebPage/WebFrame.messages.in:
* Tools/TestWebKitAPI/Tests/WebKitCocoa/FocusWebView.mm:
(TestWebKitAPI::MultipleFrames)):

Canonical link: <a href="https://commits.webkit.org/307928@main">https://commits.webkit.org/307928@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/beecbad8e3ce5d32dd79f52ca155a21f5109294e

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/145961 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/18651 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/10789 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/154640 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/99505 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/147836 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/19133 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/18540 "Built successfully") | [  ~~🧪 wpe-wk2~~](https://ews-build.webkit.org/#/builders/34/builds/112276 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/99505 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/5daa921c-ce55-45d3-b732-15b66ec150b3) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/148924 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/14658 "Passed tests") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/131125 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/93178 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/93f37955-9d93-4aed-bea9-fd549f84a0cb) 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/13945 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/11701 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/2085 "Built successfully") | | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/123498 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-wk1~~](https://ews-build.webkit.org/#/builders/167/builds/8056 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/156951 "Built successfully") | | 
| | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/180 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [❌ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/9296 "Exiting early after 10 failures. 10 tests run. 1 failures") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/1/builds/120289 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/18489 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/15456 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/120627 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30913 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/18523 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/129471 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/74202 "Built successfully") | | 
| | [  ~~🧪 vision-wk2~~](https://ews-build.webkit.org/#/builders/164/builds/16332 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/7413 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/18108 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/81878 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/17845 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/18026 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/17900 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/page/FocusController.cpp
+++ b/Source/WebCore/page/FocusController.cpp
@@ -161,7 +161,7 @@ static void clearSelectionIfNeeded(LocalFrame* oldFocusedFrame, LocalFrame* newF
 
 class FocusNavigationScope {
 public:
-    Element* owner() const;
+    Variant<RefPtr<Element>, RefPtr<Frame>> owner() const;
     WEBCORE_EXPORT static FocusNavigationScope scopeOf(Node&);
     static FocusNavigationScope scopeOwnedByScopeOwner(Element&);
     static FocusNavigationScope scopeOwnedByIFrame(HTMLFrameOwnerElement&);
@@ -344,7 +344,7 @@ FocusNavigationScope::FocusNavigationScope(Element& element)
 {
 }
 
-Element* NODELETE FocusNavigationScope::owner() const
+Variant<RefPtr<Element>, RefPtr<Frame>> NODELETE FocusNavigationScope::owner() const
 {
     if (m_slotElement)
         return m_slotElement.get();
@@ -354,9 +354,12 @@ Element* NODELETE FocusNavigationScope::owner() const
         return shadowRoot->host();
     if (invokerForOpenPopover(m_treeScopeRootNode.get()))
         return downcast<Element>(m_treeScopeRootNode.get());
-    if (auto* frame = m_treeScopeRootNode->document().frame())
-        return frame->ownerElement();
-    return nullptr;
+    if (RefPtr frame = m_treeScopeRootNode->document().frame()) {
+        if (RefPtr ownerElement = frame->ownerElement())
+            return RefPtr<Element> { ownerElement };
+        return RefPtr<Frame> { frame->parent() };
+    }
+    return RefPtr<Element> { nullptr };
 }
 
 FocusNavigationScope FocusNavigationScope::scopeOf(Node& startingNode)
@@ -559,6 +562,42 @@ FocusableElementSearchResult FocusController::findFocusableElementStartingWithLo
     UserGestureIndicator gestureIndicator(IsProcessingUserGesture::Yes, document.get());
 
     return findFocusableElementInDocumentOrderStartingWithFrame(frame, document->documentElement(), nullptr, direction, focusEventData, InitialFocus::No, ContinuingRemoteSearch::Yes, shouldFocusElement);
+}
+
+FocusableElementSearchResult FocusController::findFocusableElementContinuingFromFrame(FocusDirection direction, WebCore::FrameIdentifier frameID, const FocusEventData& focusEventData, LocalFrame& localFrame, ShouldFocusElement shouldFocusElement)
+{
+    RefPtr ownerElement = localFrame.page()->chrome().client().frameOwnerElementForFrameID(frameID);
+
+    if (!ownerElement)
+        return { nullptr };
+
+    auto findResult = findFocusableElementAcrossFocusScope(direction, FocusNavigationScope::scopeOf(*ownerElement), ownerElement.get(), focusEventData, shouldFocusElement);
+
+    if (findResult.continuedSearchInRemoteFrame == ContinuedSearchInRemoteFrame::Yes)
+        return findResult;
+
+    if (!findResult.element) {
+        // We didn't find a node to focus, so we should try to pass focus to Chrome, as long as we are trying to move the focus.
+        if (shouldFocusElement == ShouldFocusElement::Yes) {
+            if (relinquishFocusToChrome(direction)) {
+                findResult.relinquishedFocusToChrome = RelinquishedFocusToChrome::Yes;
+                return findResult;
+            }
+        }
+
+        // Chrome doesn't want focus, so we should wrap focus.
+        // FIXME: We probably want to travel up the document tree
+        RefPtr localTopDocument = m_page->localTopDocument();
+        if (!localTopDocument)
+            return findResult;
+
+        findResult = findFocusableElementAcrossFocusScope(direction, FocusNavigationScope::scopeOf(*localTopDocument), nullptr, focusEventData, shouldFocusElement);
+
+        if (!findResult.element)
+            return findResult;
+    }
+
+    return findResult;
 }
 
 FocusableElementSearchResult FocusController::findFocusableElementDescendingIntoSubframes(FocusDirection direction, Element* startingElement, const FocusEventData& focusEventData, ShouldFocusElement shouldFocusElement)
@@ -783,28 +822,65 @@ FocusableElementSearchResult FocusController::findFocusableElementAcrossFocusSco
     }
 
     // If there's no focusable node to advance to, move up the focus scopes until we find one.
-    RefPtr owner = scope.owner();
-    while (owner) {
-        if (direction == FocusDirection::Backward && isFocusableScopeOwner(*owner, focusEventData))
-            return findFocusableElementDescendingIntoSubframes(direction, owner.get(), focusEventData, shouldFocusElement);
+    auto isValidOwner = [](const auto& ownerVariant) {
+        return WTF::switchOn(ownerVariant,
+            [](const RefPtr<Element>& owner) { return owner != nullptr; },
+            [](const RefPtr<Frame>& frame) { return frame != nullptr; }
+        );
+    };
+
+    auto owner = scope.owner();
+
+    auto handleElementOwner = [&](Element& element) -> FocusableElementSearchResult {
+        if (direction == FocusDirection::Backward && isFocusableScopeOwner(element, focusEventData))
+            return findFocusableElementDescendingIntoSubframes(direction, &element, focusEventData, shouldFocusElement);
 
         // If we're getting out of a popover backwards, focus the invoker itself instead of the node preceding it, if possible.
-        RefPtr invoker = invokerForOpenPopover(owner.get());
+        RefPtr invoker = invokerForOpenPopover(&element);
         if (invoker && direction == FocusDirection::Backward && invoker->isKeyboardFocusable(focusEventData))
             return { invoker.get() };
 
-        auto outerScope = FocusNavigationScope::scopeOf(invoker ? *invoker : *owner);
-        auto candidateInOuterScope = findFocusableElementWithinScope(direction, outerScope, invoker ? invoker.get() : owner.get(), focusEventData, shouldFocusElement);
-        if (candidateInOuterScope.element)
+        auto outerScope = FocusNavigationScope::scopeOf(invoker ? *invoker : element);
+        auto candidateInOuterScope = findFocusableElementWithinScope(direction, outerScope, invoker ? invoker.get() : &element, focusEventData, shouldFocusElement);
+        if (candidateInOuterScope.element || candidateInOuterScope.continuedSearchInRemoteFrame == ContinuedSearchInRemoteFrame::Yes)
             return candidateInOuterScope;
         owner = outerScope.owner();
+        return { };
+    };
+
+    while (isValidOwner(owner)) {
+        auto result = WTF::switchOn(owner,
+            [&](const RefPtr<Element>& element) -> FocusableElementSearchResult {
+                return handleElementOwner(*element);
+            },
+            [&](const RefPtr<Frame>& frame) -> FocusableElementSearchResult {
+                switch (frame->frameType()) {
+                case Frame::FrameType::Remote:
+
+                    if (shouldFocusElement == ShouldFocusElement::Yes) {
+                        RefPtr currentFrame = currentNode->document().frame();
+                        clearSelectionIfNeeded(currentFrame.get(), nullptr, nullptr);
+                        currentNode->document().setFocusedElement(nullptr);
+                    }
+                    downcast<RemoteFrame>(frame)->client().findFocusableElementContinuingFromFrame(direction, currentNode->document().frame()->frameID(), focusEventData, shouldFocusElement);
+                    return { nullptr, ContinuedSearchInRemoteFrame::Yes };
+                case Frame::FrameType::Local:
+                    if (RefPtr ownerElement = frame->ownerElement())
+                        return handleElementOwner(*ownerElement);
+                }
+                return { };
+            }
+        );
+        if (result.element || result.continuedSearchInRemoteFrame == ContinuedSearchInRemoteFrame::Yes)
+            return result;
+
     }
     return candidateInCurrentScope;
 }
 
 FocusableElementSearchResult FocusController::findFocusableElementWithinScope(FocusDirection direction, const FocusNavigationScope& scope, Node* start, const FocusEventData& focusEventData, ShouldFocusElement shouldFocusElement)
 {
-    // Starting node is exclusive.
+    // Starting node is excluded.
     auto candidate = direction == FocusDirection::Forward
         ? nextFocusableElementWithinScope(scope, start, focusEventData)
         : previousFocusableElementWithinScope(scope, start, focusEventData);
--- a/Source/WebCore/page/FocusController.h
+++ b/Source/WebCore/page/FocusController.h
@@ -88,6 +88,8 @@ public:
 
     WEBCORE_EXPORT FocusableElementSearchResult findFocusableElementStartingWithLocalFrame(FocusDirection, const FocusEventData&, LocalFrame&, ShouldFocusElement);
 
+    WEBCORE_EXPORT FocusableElementSearchResult findFocusableElementContinuingFromFrame(FocusDirection, WebCore::FrameIdentifier, const FocusEventData&, LocalFrame&, ShouldFocusElement);
+
 private:
     void setActiveInternal(bool);
     void setFocusedInternal(bool);
--- a/Source/WebCore/page/Frame.cpp
+++ b/Source/WebCore/page/Frame.cpp
@@ -360,6 +360,11 @@ bool Frame::isPrinting() const
     return m_isPrinting;
 }
 
+RefPtr<Frame> Frame::parent() const
+{
+    return tree().parent();
+}
+
 void Frame::setPrinting(bool printing, FloatSize pageSize, FloatSize originalPageSize, float maximumShrinkRatio, AdjustViewSize shouldAdjustViewSize, NotifyUIProcess notifyUIProcess)
 {
     m_isPrinting = printing;
--- a/Source/WebCore/page/RemoteFrameClient.h
+++ b/Source/WebCore/page/RemoteFrameClient.h
@@ -74,6 +74,7 @@ public:
     virtual void updateScrollingMode(ScrollbarMode scrollingMode) = 0;
     virtual void reportMixedContentViolation(bool blocked, const URL& target) = 0;
     virtual void findFocusableElementDescendingIntoRemoteFrame(FocusDirection, const FocusEventData&, ShouldFocusElement, CompletionHandler<void(FoundElementInRemoteFrame)>&&) = 0;
+    virtual void findFocusableElementContinuingFromFrame(FocusDirection, WebCore::FrameIdentifier, const FocusEventData&, ShouldFocusElement) = 0;
 
     virtual bool isWebRemoteFrameClient() const { return false; }
     virtual ~RemoteFrameClient() { }
--- a/Source/WebKit/UIProcess/WebFrameProxy.cpp
+++ b/Source/WebKit/UIProcess/WebFrameProxy.cpp
@@ -892,6 +892,11 @@ void WebFrameProxy::findFocusableElementDescendingIntoRemoteFrame(WebCore::Focus
     sendWithAsyncReply(Messages::WebFrame::FindFocusableElementDescendingIntoRemoteFrame(direction, focusEventData, shouldFocusElement), WTF::move(completionHandler));
 }
 
+void WebFrameProxy::findFocusableElementContinuingFromFrame(WebCore::FocusDirection direction, WebCore::FrameIdentifier frameID, const WebCore::FocusEventData& focusEventData, WebCore::ShouldFocusElement shouldFocusElement)
+{
+    send(Messages::WebFrame::FindFocusableElementContinuingFromFrame(direction, frameID, focusEventData, shouldFocusElement));
+}
+
 std::optional<SharedPreferencesForWebProcess> WebFrameProxy::sharedPreferencesForWebProcess() const
 {
     return process().sharedPreferencesForWebProcess();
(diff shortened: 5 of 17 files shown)


pxlcoder:
> Focus cannot from one remote frame to a sibling remote frame by pressing 'tab'.

I think a word is missing. Focus cannot _move_?

Source/WebKit/UIProcess/WebFrameProxy.cpp
```diff
@@ -876,6 +876,12 @@ void WebFrameProxy::findFocusableElementDescendingIntoRemoteFrame(WebCore::Focus
     sendWithAsyncReply(Messages::WebFrame::FindFocusableElementDescendingIntoRemoteFrame(direction, focusEventData, shouldFocusElement), WTF::move(completionHandler));
 }
 
+void WebFrameProxy::findFocusableElementContinutingFromFrame(WebCore::FocusDirection direction, WebCore::FrameIdentifier frameID, const WebCore::FocusEventData& focusEventData, WebCore::ShouldFocusElement shouldFocusElement)
+{
+    send(Messages::WebFrame::FindFocusableElementContinutingFromFrame(direction, frameID, focusEventData, shouldFocusElement));
+}
+
```
pxlcoder, line 899:
```suggestion
```

Tools/TestWebKitAPI/Tests/WebKitCocoa/FocusWebView.mm
```diff
@@ -212,7 +212,7 @@ HTTPServer server({
     }, HTTPServer::Protocol::HttpsProxy);
 
     RetainPtr configuration = server.httpsProxyConfiguration();
-    // FIXME: Enable site isolation here and make the test behave the same.
```
pxlcoder, line 215: Shouldn't we add `[[configuration preferences] _setSiteIsolationEnabled:YES];`?

Source/WebCore/page/Frame.cpp
```diff
@@ -371,6 +371,11 @@ bool Frame::isPrinting() const
     return m_isPrinting;
 }
 
+RefPtr<Frame> parent() const
```
aprotyas:
```suggestion
RefPtr<Frame> Frame::parent() const
```

Source/WebCore/page/FocusController.cpp
```diff
@@ -354,9 +354,12 @@ Element* FocusNavigationScope::owner() const
         return shadowRoot->host();
     if (invokerForOpenPopover(m_treeScopeRootNode.get()))
         return downcast<Element>(m_treeScopeRootNode.get());
-    if (auto* frame = m_treeScopeRootNode->document().frame())
-        return frame->ownerElement();
-    return nullptr;
+    if (RefPtr frame = m_treeScopeRootNode->document().frame()) {
+        if (RefPtr<Element> ownerElement = frame->ownerElement())
```
aprotyas:
```suggestion
        if (RefPtr ownerElement = frame->ownerElement())
```

Source/WebCore/page/FocusController.h
```diff
@@ -89,6 +89,8 @@ class FocusController final : public CanMakeCheckedPtr<FocusController> {
 
     WEBCORE_EXPORT FocusableElementSearchResult findFocusableElementStartingWithLocalFrame(FocusDirection, const FocusEventData&, LocalFrame&, ShouldFocusElement);
 
+    WEBCORE_EXPORT FocusableElementSearchResult findFocusableElementContinutingFromFrame(FocusDirection, WebCore::FrameIdentifier, const FocusEventData&, LocalFrame&, ShouldFocusElement);
```
aprotyas: minor typo in this function name
```suggestion
    WEBCORE_EXPORT FocusableElementSearchResult findFocusableElementContinuingFromFrame(FocusDirection, WebCore::FrameIdentifier, const FocusEventData&, LocalFrame&, ShouldFocusElement);
```

Source/WebKit/WebProcess/WebCoreSupport/WebChromeClient.h
```diff
@@ -549,6 +549,8 @@ class WebChromeClient final : public WebCore::ChromeClient {
     void clearAnimationsForActiveWritingToolsSession() final;
 #endif
 
+    WebCore::HTMLFrameOwnerElement* frameOwnerElementForFrameID(WebCore::FrameIdentifier) const;
```
aprotyas: This can be marked `final`
```suggestion
    WebCore::HTMLFrameOwnerElement* frameOwnerElementForFrameID(WebCore::FrameIdentifier) const final;
```

Source/WebCore/page/RemoteFrameClient.h
```diff
@@ -74,6 +74,7 @@ class RemoteFrameClient : public FrameLoaderClient {
     virtual void updateScrollingMode(ScrollbarMode scrollingMode) = 0;
     virtual void reportMixedContentViolation(bool blocked, const URL& target) = 0;
     virtual void findFocusableElementDescendingIntoRemoteFrame(FocusDirection, const FocusEventData&, ShouldFocusElement, CompletionHandler<void(FoundElementInRemoteFrame)>&&) = 0;
+    virtual void findFocusableElementContinutingFromFrame(FocusDirection, WebCore::FrameIdentifier, const FocusEventData&, ShouldFocusElement) = 0;
```
jesxilin:
```suggestion
    virtual void findFocusableElementContinuingFromFrame(FocusDirection, WebCore::FrameIdentifier, const FocusEventData&, ShouldFocusElement) = 0;
```
Same typo here among other instances

Source/WebKit/WebProcess/WebPage/WebFrame.cpp
```diff
@@ -1581,6 +1581,16 @@ void WebFrame::findFocusableElementDescendingIntoRemoteFrame(WebCore::FocusDirec
     completionHandler(foundElementInRemoteFrame);
 }
 
+void WebFrame::findFocusableElementContinutingFromFrame(WebCore::FocusDirection direction, WebCore::FrameIdentifier frameID, const WebCore::FocusEventData& focusEventData, WebCore::ShouldFocusElement shouldFocusElement)
+{
+    if (m_coreFrame) {
+        if (RefPtr localFrame = dynamicDowncast<LocalFrame>(m_coreFrame.get())) {
+            if (RefPtr page = localFrame->page())
```
jesxilin: Maybe to reduce nesting, it could be possible to do smth like this?

```suggestion
    if (!m_coreFrame)
        return;
    if (RefPtr localFrame = dynamicDowncast<LocalFrame>(m_coreFrame.get())) {
        if (RefPtr page = localFrame->page())
```

Tools/TestWebKitAPI/Tests/WebKitCocoa/FocusWebView.mm
megangardner: I don't think we want to force site isolation on, we want the test to run in both situations

Source/WebCore/page/FocusController.cpp
```diff
@@ -561,6 +564,38 @@ FocusableElementSearchResult FocusController::findFocusableElementStartingWithLo
...
+
+    if (!findResult.element) {
+        // We didn't find a node to focus, so we should try to pass focus to Chrome.
+        if (relinquishFocusToChrome(direction)) {
+            findResult.relinquishedFocusToChrome = RelinquishedFocusToChrome::Yes;
+            return findResult;
+        }
+
+        // Chrome doesn't want focus, so we should wrap focus.
+        // FIXME: We probably want to travel up the document tree
```
pxlcoder, line 589: Is there a known test case / scenario that is broken because we're not doing this?

Source/WebCore/page/FocusController.cpp
```diff
@@ -561,6 +564,38 @@ FocusableElementSearchResult FocusController::findFocusableElementStartingWithLo
...
+        RefPtr localTopDocument = m_page->localTopDocument();
+        if (!localTopDocument)
+            return findResult;
+        findResult = findFocusableElementAcrossFocusScope(direction, FocusNavigationScope::scopeOf(*localTopDocument), nullptr, focusEventData, shouldFocusElement);
```
pxlcoder, line 594: Nit: Readability.
```suggestion

        findResult = findFocusableElementAcrossFocusScope(direction, FocusNavigationScope::scopeOf(*localTopDocument), nullptr, focusEventData, shouldFocusElement);
```

Source/WebCore/page/FocusController.cpp
```diff
@@ -561,6 +564,38 @@ FocusableElementSearchResult FocusController::findFocusableElementStartingWithLo
...
+
+        if (!findResult.element)
+            return findResult;
+    }
+    return findResult;
```
pxlcoder, line 600: Nit: Readability.
```suggestion

    return findResult;
```

Source/WebCore/page/FocusController.cpp
```diff
@@ -783,28 +818,65 @@ FocusableElementSearchResult FocusController::findFocusableElementAcrossFocusSco
...
+                    if (RefPtr ownerElement = localFrame->ownerElement())
+                        return handleElementOwner(*ownerElement);
+                }
+                return { };
+            }
+        );
+
+        if (result.element || result.continuedSearchInRemoteFrame == ContinuedSearchInRemoteFrame::Yes)
+            return result;
+
```
pxlcoder, line 876: Nit: Extra newline.
```suggestion
```

Source/WebCore/page/FocusController.cpp
```diff
@@ -783,28 +818,65 @@ FocusableElementSearchResult FocusController::findFocusableElementAcrossFocusSco
...
+                        currentNode->document().setFocusedElement(nullptr);
+                    }
+                    remoteFrame->client().findFocusableElementContinuingFromFrame(direction, currentNode->document().frame()->frameID(), focusEventData, shouldFocusElement);
+                    return { nullptr, ContinuedSearchInRemoteFrame::Yes };
+                }
+                if (RefPtr localFrame = dynamicDowncast<LocalFrame>(frame)) {
+                    if (RefPtr ownerElement = localFrame->ownerElement())
+                        return handleElementOwner(*ownerElement);
+                }
+                return { };
```
pxlcoder, line 871: While this code is safe, the general structure of:

```
if (RefPtr remoteFrame = dynamicDowncast<RemoteFrame>(frame)) {
    ...
    return ...;
}

if (RefPtr localFrame = dynamicDowncast<LocalFrame>(frame)) {
    ...
    return ...;
}

return ...;
```

adds some confusion and hides an invariant.

`return { };` will never be reached since all `Frame`s are either local or remote. And if it was reached, this `while` loop would never terminate. So its inclusion adds confusion.

For that reason, I think it would be better to write this logic as a switch statement on `frame.frameType()`. In the `FrameType::Local` case, no downcast is necessary, since no `LocalFrame`-specific methods are used. Then, in the `FrameType::Remote` case you can use `downcast<>` rather than `dynamicDowncast<>`.

Source/WebCore/page/FocusController.cpp
megangardner: No more than there is for the "//FIXME: Implement sibling frame search by continuing here" that is a few lines later. I plan on writing a test to exercise this path in the coming days.

Source/WebCore/page/FocusController.cpp
megangardner: I see no difference here, what am I missing??

megangardner: I see the newline now.

Source/WebCore/page/FocusController.cpp
```diff
@@ -783,28 +820,65 @@ FocusableElementSearchResult FocusController::findFocusableElementAcrossFocusSco
...
+
+    while (isValidOwner(owner)) {
+        auto result = WTF::switchOn(owner,
+            [&](const RefPtr<Element>& element) -> FocusableElementSearchResult {
+                return handleElementOwner(*element);
+            },
+            [&](const RefPtr<Frame>& frame) -> FocusableElementSearchResult {
+                switch (frame->frameType()) {
+                case Frame::FrameType::Remote:
+
```
pxlcoder, line 859:
```suggestion
```

Source/WebCore/page/FocusController.cpp
```diff
@@ -783,28 +820,65 @@ FocusableElementSearchResult FocusController::findFocusableElementAcrossFocusSco
...
+                    if (shouldFocusElement == ShouldFocusElement::Yes) {
+                        RefPtr currentFrame = currentNode->document().frame();
+                        clearSelectionIfNeeded(currentFrame.get(), nullptr, nullptr);
+                        currentNode->document().setFocusedElement(nullptr);
+                    }
+                    downcast<RemoteFrame>(frame)->client().findFocusableElementContinuingFromFrame(direction, currentNode->document().frame()->frameID(), focusEventData, shouldFocusElement);
+                    return { nullptr, ContinuedSearchInRemoteFrame::Yes };
+                case Frame::FrameType::Local:
+                    if (RefPtr ownerElement = frame->ownerElement())
+                        return handleElementOwner(*ownerElement);
```
pxlcoder, line 869: If there's no ownerElement we'd have to stop the while loop somehow. Otherwise it would loop infinitely checking the same thing. Should we assign nullptr to `owner`? So that `isValidOwner(owner)` fails.

Source/WebCore/page/FocusController.cpp
```diff
@@ -561,6 +564,40 @@ FocusableElementSearchResult FocusController::findFocusableElementStartingWithLo
...
+    if (!findResult.element) {
+        // We didn't find a node to focus, so we should try to pass focus to Chrome.
+        if (relinquishFocusToChrome(direction)) {
+            findResult.relinquishedFocusToChrome = RelinquishedFocusToChrome::Yes;
+            return findResult;
+        }
+
+        // Chrome doesn't want focus, so we should wrap focus.
+        // FIXME: We probably want to travel up the document tree
+        RefPtr localTopDocument = m_page->localTopDocument();
```
rniwa, line 590: In the case main frame isn't a `LocalFrame`, don't we want to continue our search in the process where the same main frame is a local frame?

Source/WebCore/page/FocusController.cpp
```diff
@@ -344,7 +344,7 @@ FocusNavigationScope::FocusNavigationScope(Element& element)
 {
 }
 
-Element* FocusNavigationScope::owner() const
+Variant<RefPtr<Element>, RefPtr<Frame>> FocusNavigationScope::owner() const
```
rniwa: It's a bit spooky that there are two ways for this return value to be "invalid". Namely it could be `nullptr` `Frame` or `nullptr` `Element`. Perhaps we can use `std::optional<Variant<Ref<Element>, Ref<Frame>>>` instead?

Source/WebCore/page/FocusController.cpp
```diff
@@ -783,28 +820,65 @@ FocusableElementSearchResult FocusController::findFocusableElementAcrossFocusSco
...
+                return handleElementOwner(*element);
+            },
+            [&](const RefPtr<Frame>& frame) -> FocusableElementSearchResult {
+                switch (frame->frameType()) {
+                case Frame::FrameType::Remote:
+
+                    if (shouldFocusElement == ShouldFocusElement::Yes) {
+                        RefPtr currentFrame = currentNode->document().frame();
+                        clearSelectionIfNeeded(currentFrame.get(), nullptr, nullptr);
+                        currentNode->document().setFocusedElement(nullptr);
```
rniwa, line 863: Is this always right even when the starting node and the ending node are identical? e.g. we have exactly one input element and chrome doesn't accept focus.

Source/WebKit/WebProcess/WebCoreSupport/WebChromeClient.cpp
```diff
@@ -2357,6 +2359,19 @@ bool WebChromeClient::requiresScriptTrackingPrivacyProtections(const URL& url, c
...
 }
 
+HTMLFrameOwnerElement* WebChromeClient::frameOwnerElementForFrameID(FrameIdentifier frameID) const
+{
+    RefPtr webFrame = WebFrame::webFrame(frameID);
+    if (!webFrame)
+        return nullptr;
+
+    RefPtr remoteFrame = webFrame->coreRemoteFrame();
+    if (!remoteFrame)
```
rniwa: How does this early return make sense? This function should be able to return the owner element if a given frame is a local frame as well.
If we want to restrict this function to be only called in the situations where frameID is a remote frame in this process, we should probably assert that instead.

Source/WebCore/page/FocusController.cpp
megangardner: We're at the end of the known document, so we want to go start searching from the beginning, be that a local or remote frame.

Source/WebCore/page/FocusController.cpp
megangardner: I don't think we're in a situation to know if that is the case with where site isolation is right now. If we don't clear, then we're in danger of leaving something focused that shouldn't be. If we loop back around and there's nothing else to focus, we should focus the next thing again. I will write a test for the case you describe and fix it if it causes issues, though.

Source/WebKit/WebProcess/WebCoreSupport/WebChromeClient.cpp
megangardner: That's true. I wrote this function for the express purpose of getting remote frames, so I didn't consider other cases. I'll expand it to include local frames.

rniwa (approved):
Seems like an overall improvement over the status quo.

PR #58576, opened 2026-02-13
Reduce Ref counting operations in TextUtil::width

NathanSolomon733:
#### e20f666f105ba9f902201fb018ce20ceafe35fd8
<pre>
Reduce Ref counting operations in TextUtil::width
<a href="https://rdar.apple.com/170275347">rdar://170275347</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=307738">https://bugs.webkit.org/show_bug.cgi?id=307738</a>

Reviewed by Vitor Roriz.

Add NODELETE cachedPrimaryFont() and cachedCanTakeFixedPitchFastContentMeasuring()
to reduce ref-count operations in TextUtil::width().

TextUtil::width() is a hot function with many ref()/deref(). Each call to
FontCascade::primaryFont() uses
protect(m_fonts) / protect(fontSelector()), causing
ref/deref operations per call. Similarly,
canTakeFixedPitchFastContentMeasuring() uses protect() internally.

Add cachedPrimaryFont() which returns const Font* directly
from the cached weak pointer, and cachedCanTakeFixedPitchFastContentMeasuring()
which reads the cached TriState without protect(). On cache hit,
this eliminates all ref counting. On the cache miss,
fall back to the existing ref-counted methods. Cache hit is much more common than cache miss.

Accessing m_fonts without protect() and returning a raw pointer instead
of a Ref is safe because the FontCascade owns the FontCascadeFonts through
a RefPtr, and the FontCascadeFonts holds the Font alive through a
chain of strong references (FontRanges -&gt; FontAccessor -&gt; Font).
Since every caller of cachedPrimaryFont() is operating on a live
FontCascade, the returned Font pointer is valid for the duration of
the call.
The existing code already assumes m_fonts is non-null at the time of access because
protect(m_fonts)-&gt;primaryFont() would crash on a null m_fonts,
since protect() of a null RefPtr returns a null RefPtr.

The NODELETE annotation verifies that the called method
cannot free the returned object&apos;s memory.

A/B testing shows this is a small performance improvement.

* Source/WebCore/platform/graphics/Font.h:
(WebCore::Font::spaceWidth const):
* Source/WebCore/platform/graphics/FontCascade.h:
(WebCore::FontCascade::primaryFont const):
(WebCore::FontCascade::canTakeFixedPitchFastContentMeasuring const):
* Source/WebCore/platform/graphics/FontCascadeFonts.h:
(WebCore::FontCascadeFonts::cachedCanTakeFixedPitchFastContentMeasuring const):
(WebCore::FontCascadeFonts::cachedPrimaryFont const):

Canonical link: <a href="https://commits.webkit.org/309020@main">https://commits.webkit.org/309020@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/0be18f05043ca07cf2c01a0a5d6ca0dcb3bc270d

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/147905 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/20592 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/14186 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/156587 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/101320 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/9d4ac2f2-35f8-4fc2-8c9e-0f095cea8567) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/149779 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/21047 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/20495 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/114026 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/81315 "Passed tests") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/9a1b18ba-004e-4fef-aac8-c29a61a05e73) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/150866 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/16269 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/132855 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94790 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/dbab67e9-5437-40a0-91f3-c131881607f8) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/15415 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/13210 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/4027 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/125021 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/10752 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/158922 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/2057 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/12246 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/122060 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/20389 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/17139 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/122263 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/33510 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/20399 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/132552 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/76542 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/17750 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/9314 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/20007 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/83767 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/19737 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/19885 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/19793 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/platform/graphics/Font.h
+++ b/Source/WebCore/platform/graphics/Font.h
@@ -192,7 +192,7 @@ public:
 
     Path pathForGlyph(Glyph) const;
 
-    float spaceWidth(SyntheticBoldInclusion SyntheticBoldInclusion = SyntheticBoldInclusion::Incorporate) const
+    float NODELETE spaceWidth(SyntheticBoldInclusion SyntheticBoldInclusion = SyntheticBoldInclusion::Incorporate) const
     {
         return m_spaceWidth + (SyntheticBoldInclusion == SyntheticBoldInclusion::Incorporate ? syntheticBoldOffset() : 0);
     }
--- a/Source/WebCore/platform/graphics/FontCascade.h
+++ b/Source/WebCore/platform/graphics/FontCascade.h
@@ -386,6 +386,8 @@ private:
 
 inline const Font& FontCascade::primaryFont() const
 {
+    if (m_fonts->cachedPrimaryFont())
+        return *m_fonts->cachedPrimaryFont();
     WeakRef font = protect(m_fonts)->primaryFont(m_fontDescription, protect(fontSelector()).get());
     m_fontDescription.resolveFontSizeAdjustFromFontIfNeeded(protect(font));
     return font;
@@ -403,6 +405,9 @@ inline bool FontCascade::isFixedPitch() const
 
 inline bool FontCascade::canTakeFixedPitchFastContentMeasuring() const
 {
+    auto cachedCanTakeFixedPitch = m_fonts->cachedCanTakeFixedPitchFastContentMeasuring();
+    if (cachedCanTakeFixedPitch != TriState::Indeterminate)
+        return cachedCanTakeFixedPitch == TriState::True;
     return protect(m_fonts)->canTakeFixedPitchFastContentMeasuring(m_fontDescription, protect(fontSelector()).get());
 }
 
--- a/Source/WebCore/platform/graphics/FontCascadeFonts.h
+++ b/Source/WebCore/platform/graphics/FontCascadeFonts.h
@@ -115,6 +115,10 @@ public:
     bool isFixedPitch(const FontCascadeDescription&, FontSelector*);
 
     bool canTakeFixedPitchFastContentMeasuring(const FontCascadeDescription&, FontSelector*);
+    TriState NODELETE cachedCanTakeFixedPitchFastContentMeasuring() const
+    {
+        return m_canTakeFixedPitchFastContentMeasuring;
+    }
 
     bool isLoadingCustomFonts() const;
 
@@ -143,6 +147,7 @@ public:
     const TextShapingResult* getOrCreateCachedShapedText(const TextRun&, const FontCascade&, unsigned from, std::optional<unsigned> to, ForTextEmphasis);
 
     const Font& primaryFont(const FontCascadeDescription&, FontSelector*);
+    const Font* NODELETE cachedPrimaryFont() const { return m_cachedPrimaryFont.get(); }
     WEBCORE_EXPORT const FontRanges& realizeFallbackRangesAt(const FontCascadeDescription&, FontSelector*, unsigned fallbackIndex);
 
     void pruneSystemFallbacks();


Source/WebCore/layout/formattingContexts/inline/text/TextUtil.cpp
```diff
@@ -76,12 +83,26 @@ InlineLayoutUnit TextUtil::width(const InlineTextBox& inlineTextBox, const FontC
         ++to;
     auto width = 0.f;
     auto useSimplifiedContentMeasuring = inlineTextBox.canUseSimplifiedContentMeasuring();
+
+    std::optional<Ref<const Font>> fontHolder;
```
rniwa: Since this file is listed in `Source/WebCore/SaferCPPExpectations/UncheckedCallArgsCheckerExpectations`, call arguments aren't checked for memory safety. Based on that, I'm not sure how we can review this PR.

Source/WebCore/layout/formattingContexts/inline/text/TextUtil.cpp
```diff
@@ -114,8 +134,18 @@ InlineLayoutUnit TextUtil::width(const InlineTextItem& inlineTextItem, const Fon
     if (inlineTextItem.isWhitespace()) {
         CheckedRef inlineTextBox = inlineTextItem.inlineTextBox();
         auto singleWhiteSpace = from - to == 1 || !TextUtil::shouldPreserveSpacesAndTabs(inlineTextBox);
-        if (singleWhiteSpace)
-            return std::max(0.f, singleSpaceWidth(fontCascade, inlineTextBox->canUseSimplifiedContentMeasuring()));
+        if (singleWhiteSpace) {
+            std::optional<Ref<const Font>> fontHolder;
```
cdumez: a `std::optional<Ref<X>>` is not a common thing in WebKit at all. Why can't we use a `RefPtr<X>`?

NathanSolomon733: Oh, right. That's way better. That did not cross my mind for some reason.

Source/WebCore/layout/formattingContexts/inline/text/TextUtil.cpp
NathanSolomon733: I checked the history. TextUtil.cpp was added to UncheckedCallArgsCheckerExpectations in https://github.com/WebKit/WebKit/commit/c612e507973fe45e1c71c1893376818803781495
 as part of an addition of files when the static analyzer false negative was fixed for CanMakeCheckedPtrBase. I see in TextUtil.cpp we are passing CheckedRef<const InlineTextBox> to functions taking const Box& in shouldPreserveSpacesAndTabs(). I'd have to change this function to take in a CheckedRef. 
Although we'd have to remove TextUtil.cpp first from UncheckedCallArgsCheckerExpectations so I can even verify that these are all the places and that I've fixed them.
Should I even try to do this? Or do you see other issues?

rniwa: I think it would be good to get TextUtil.cpp "clean" before landing this change so that we can utilize static analysis to prove its soundness.

NathanSolomon733: Ok I made https://github.com/WebKit/WebKit/pull/59249/ which is ready for review.

I am A/B testing it now too because it may regress perf if we are adding another ref/deref operation at the call site of TextUtil::width(), which is the function I'm trying to optimize here.

NathanSolomon733: A/B testing results show that it is neutral.

Source/WebCore/layout/formattingContexts/inline/text/TextUtil.cpp
```diff
@@ -51,7 +51,11 @@ namespace Layout {
 
 InlineLayoutUnit TextUtil::singleSpaceWidth(const FontCascade& fontCascade, bool canUseSimplifiedContentMeasuring)
 {
-    auto width = canUseSimplifiedContentMeasuring ? fontCascade.primaryFont()->spaceWidth() : fontCascade.widthOfSpaceString();
+    InlineLayoutUnit width;
+    if (fontCascade.cachedPrimaryFont())
```
NathanSolomon733: Not sure if the static analyzer will let me assign a raw pointer variable here like 
if (auto* cachedFont = fontCascade.cachedPrimaryFont()) 
which is why I am repeating the calls to fontCascade.cachedPrimaryFont() here

NathanSolomon733: I haven't tried but I don't think it will accept it because fontCascade.widthOfSpaceString() below is a non-trivial function call. So, TextUtil::singleSpaceWidth which we are in is non-trivial. This will cause the analyzer to flag it.

Source/WebCore/platform/graphics/Font.h
```diff
@@ -192,7 +192,7 @@ class Font : public RefCounted<Font>, public CanMakeSingleThreadWeakPtr<Font> {
 
     Path pathForGlyph(Glyph) const;
 
-    float spaceWidth(SyntheticBoldInclusion SyntheticBoldInclusion = SyntheticBoldInclusion::Incorporate) const
+    float NODELETE spaceWidth(SyntheticBoldInclusion SyntheticBoldInclusion = SyntheticBoldInclusion::Incorporate) const
```
NathanSolomon733, line 195: This and other NODELETE annotions added in this PR might be unnecessary because the methods are all defined in the header so they are implicitly inlined. We may not actually need them to have the analyzer check for their trivialness. 
Correct me if I'm wrong though.

NathanSolomon733: Perhaps we can keep them to enforce they stay NODELETE in the future, since this PR's safety is dependent on that.

Source/WebCore/layout/formattingContexts/inline/text/TextUtil.cpp
Andrew-Fryer: `FontCascade::cachedPrimaryFont` is `inline` and `FontCasadeFonts::cachedPrimaryFont` is implicitly inline, so I think it'll just be an extra single load instruction, which will hit L1, so I don't think it matters at all for perf.

Source/WebCore/platform/graphics/Font.h
Andrew-Fryer: Yes, I believe the best practice is to keep `NODELETE` here.
See @rniwa 's reply here:
https://github.com/WebKit/WebKit/pull/59669/changes#r2869958692

Source/WebCore/platform/graphics/FontCascadeFonts.h
```diff
@@ -115,6 +113,12 @@ class FontCascadeFonts : public RefCounted<FontCascadeFonts> {
     bool isFixedPitch(const FontCascadeDescription&, FontSelector*);
 
     bool canTakeFixedPitchFastContentMeasuring(const FontCascadeDescription&, FontSelector*);
+    std::optional<bool> NODELETE cachedCanTakeFixedPitchFastContentMeasuring() const
+    {
+        if (m_canTakeFixedPitchFastContentMeasuring == TriState::Indeterminate)
+            return std::nullopt;
+        return m_canTakeFixedPitchFastContentMeasuring == TriState::True;
+    }
```
Andrew-Fryer, line 121: IMHO, it would be cleaner to simply operate on the `TriState` in `TextUtil::singleSpaceWidth` instead of converting to something that seems an awful lot like a tristate.

Source/WebCore/layout/formattingContexts/inline/text/TextUtil.cpp
```diff
@@ -78,7 +82,10 @@ InlineLayoutUnit TextUtil::width(const InlineTextBox& inlineTextBox, const FontC
     auto useSimplifiedContentMeasuring = inlineTextBox.canUseSimplifiedContentMeasuring();
     if (useSimplifiedContentMeasuring) {
         auto view = StringView(text).substring(from, to - from);
-        if (fontCascade.canTakeFixedPitchFastContentMeasuring())
+        auto fixedPitch = fontCascade.cachedCanTakeFixedPitchFastContentMeasuring();
```
Andrew-Fryer: If we're going to use `std::optional<bool>`, I think it's more readable to spell it out here.
(As I commented below, I think it's cleaner to use tristate directly here.)

Source/WebCore/layout/formattingContexts/inline/text/TextUtil.cpp
```diff
@@ -51,7 +51,11 @@ namespace Layout {
 
 InlineLayoutUnit TextUtil::singleSpaceWidth(const FontCascade& fontCascade, bool canUseSimplifiedContentMeasuring)
 {
-    auto width = canUseSimplifiedContentMeasuring ? fontCascade.primaryFont()->spaceWidth() : fontCascade.widthOfSpaceString();
+    InlineLayoutUnit width;
+    if (fontCascade.cachedPrimaryFont())
+        width = canUseSimplifiedContentMeasuring ? fontCascade.cachedPrimaryFont()->spaceWidth() : fontCascade.widthOfSpaceString();
+    else
+        width = canUseSimplifiedContentMeasuring ? fontCascade.primaryFont()->spaceWidth() : fontCascade.widthOfSpaceString();
```
Andrew-Fryer: IMHO, it's cleaner to write this as:
```
InlineLayoutUnit width;
if (canUseSimplifiedContentMeasuring)
    width = fontCascade.cachedPrimaryFont() ? fontCascade.cachedPrimaryFont()->spaceWidth() : fontCascade.primaryFont()->spaceWidth();
else
    width = fontCascade.widthOfSpaceString();
```

Source/WebCore/platform/graphics/Font.h
rniwa: Yeah, it's best to keep `NODELETE` annotation so that we can communicate to the future reader of the code that this code is supposed to be no-delete.

Andrew-Fryer:
> What is?

The interesting problem I was thinking of is that we want to avoid ref churn in the fast path and keep ref counts for the complex/slow path.
The solution here is to add separate methods for the fast path that the static analyzer can see are trivial.

Source/WebCore/layout/formattingContexts/inline/text/TextUtil.cpp
NathanSolomon733: Agreed.

Source/WebCore/platform/graphics/FontCascadeFonts.h
NathanSolomon733: I'm going to just return the TriState value. I thought TriState is like an internal detail that is worth hiding from the callers of FontCascadeFonts methods for simplicity, but it's not any simpler to use std::optional on second look.

Source/WebCore/layout/formattingContexts/inline/text/TextUtil.cpp
NathanSolomon733: Oh true.

NathanSolomon733:
FontCascade::primaryFont() was simplified to return a raw reference via https://bugs.webkit.org/show_bug.cgi?id=308955 using LIFETIME_BOUNDS, which allowed me to simplify the code more.

Andrew-Fryer:
>FontCascade::primaryFont() was simplified to return a raw reference via https://bugs.webkit.org/show_bug.cgi?id=308955 using LIFETIME_BOUNDS, which allowed me to simplify the code more.

Wow, that's much simpler; nice!
LGTM

Source/WebCore/platform/graphics/FontCascade.h
```diff
@@ -386,6 +386,8 @@ class FontCascade final : public CanMakeWeakPtr<FontCascade>, public CanMakeChec
 
 inline const Font& FontCascade::primaryFont() const
 {
+    if (m_fonts->cachedPrimaryFont())
+        return *m_fonts->cachedPrimaryFont();
     WeakRef font = protect(m_fonts)->primaryFont(m_fontDescription, protect(fontSelector()).get());
```
vitorroriz, line 391: I'm not sure I understand it.

FontCascadeFonts::primaryFont already returns the cachedPrimaryFont if possible.
Why is it ok to call m_fonts->cachedPrimaryFont() without protect but not m_fonts->primaryFont?

NathanSolomon733: We can do `m_fonts->cachedPrimaryFont()` because there is a garuntee by static analysis that `cachedPrimaryFont() `does not free memory. I put NODELETE on the method to have the analyzer verify this. If it did free memory, the analyzer would flag it. If` cachedPrimaryFont()` is verified as trivial, which it is, then it is safe to not need protect() around m_fonts because m_fonts will not get its ref count decremented for the duration of the call.

We cannot do `m_fonts->primaryFont()` because this function potentially frees memory, so it could mutate the ref count of m_fonts as some side effect. If you put NODELETE on primaryFont(), the analyzer will prove this function as being non-trivial and will not allow you to write that line without` protect()` around `m_fonts `

vitorroriz (approved):

PR #58629, opened 2026-02-13
Web Inspector: Elements: Support copying HTML text of multiple selected DOM nodes

rcaliman-apple:
#### 4237ebc104789ad73c146470fd84037c84ce7a0e
<pre>
Web Inspector: Elements: Support copying HTML text of multiple selected DOM nodes
<a href="https://bugs.webkit.org/show_bug.cgi?id=307799">https://bugs.webkit.org/show_bug.cgi?id=307799</a>
<a href="https://rdar.apple.com/169196441">rdar://169196441</a>

Reviewed by Devin Rousso.

The context menu options to copy HTML and HTML (formatted) are updated
to support copying the HTML text of all selected nodes from the DOM tree outline.

* Source/WebInspectorUI/UserInterface/Proxies/FormatterWorkerProxy.js:
(WI.FormatterWorkerProxy.prototype.formatJavaScript):
(WI.FormatterWorkerProxy.prototype.formatCSS):
(WI.FormatterWorkerProxy.prototype.formatHTML):
(WI.FormatterWorkerProxy.prototype.formatXML):
(WI.FormatterWorkerProxy.prototype.performAction):
* Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js:
(WI.DOMTreeElement.prototype.populateDOMNodeContextMenu):
(WI.DOMTreeElement.prototype._copyHTMLOfSelectedDOMNodes.async then):
(WI.DOMTreeElement.prototype._copyHTMLOfSelectedDOMNodes):

Canonical link: <a href="https://commits.webkit.org/307826@main">https://commits.webkit.org/307826@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/d166c0a3d878f2e2f3e43bc917a341c59cead9be

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/145669 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/18351 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/10237 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/154341 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/99306 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/18836 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/18244 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/112028 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/99306 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/148632 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/14394 "Passed tests") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/130838 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/92933 "Passed tests") | | [loading 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/a803c2dd-725a-4ce8-b076-ec4d24e272d6) 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/13703 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/11472 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/1788 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/123249 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/7668 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/156654 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/18201 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/8777 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/120029 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/18247 "Built successfully") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/15186 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/120381 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30853 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/16106 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/128954 "Passed tests") | [❌ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/73941 "Failed to checkout and rebase branch from PR 58629") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/17822 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/7089 "Passed tests") | | | 
| | [  ~~🛠 tv-sim~~](https://ews-build.webkit.org/#/builders/157/builds/17559 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/81602 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/17767 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/17622 "Built successfully") | | | |

### DIFF
--- a/Source/WebInspectorUI/UserInterface/Proxies/FormatterWorkerProxy.js
+++ b/Source/WebInspectorUI/UserInterface/Proxies/FormatterWorkerProxy.js
@@ -47,22 +47,22 @@ WI.FormatterWorkerProxy = class FormatterWorkerProxy
 
     formatJavaScript(sourceText, isModule, indentString, includeSourceMapData)
     {
-        this.performAction("formatJavaScript", ...arguments);
+        return this.performAction("formatJavaScript", ...arguments);
     }
 
     formatCSS(sourceText, indentString, includeSourceMapData)
     {
-        this.performAction("formatCSS", ...arguments);
+        return this.performAction("formatCSS", ...arguments);
     }
 
     formatHTML(sourceText, indentString, includeSourceMapData)
     {
-        this.performAction("formatHTML", ...arguments);
+        return this.performAction("formatHTML", ...arguments);
     }
 
     formatXML(sourceText, indentString, includeSourceMapData)
     {
-        this.performAction("formatXML", ...arguments);
+        return this.performAction("formatXML", ...arguments);
     }
 
     // Public
@@ -75,6 +75,30 @@ WI.FormatterWorkerProxy = class FormatterWorkerProxy
 
         console.assert(typeof actionName === "string", "performAction should always have an actionName");
         console.assert(typeof callback === "function", "performAction should always have a callback");
+
+        this._callbacks.set(callId, callback);
+        this._postMessage({callId, actionName, actionArguments});
+    }
+
+     performAction(actionName)
+     {
+        console.assert(typeof actionName === "string", "performAction should always have an actionName");
+
+        let callId = this._nextCallId++;
+
+        let callback = arguments[arguments.length - 1];
+        if (typeof callback !== "function") {
+            return new Promise((resolve, reject) => {
+                this._callbacks.set(callId, resolve);
+                this._postMessage({
+                    callId,
+                    actionName,
+                    actionArguments: Array.prototype.slice.call(arguments, 1),
+                });
+            });
+        }
+
+        let actionArguments = Array.prototype.slice.call(arguments, 1, arguments.length - 1);
 
         this._callbacks.set(callId, callback);
         this._postMessage({callId, actionName, actionArguments});
--- a/Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
+++ b/Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
@@ -868,21 +868,11 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
 
         if (!this.representedObject.destroyed && !this.representedObject.isPseudoElement()) {
             subMenus.copy.appendItem(WI.UIString("HTML"), () => {
-                this.representedObject.getOuterHTML()
-                .then((outerHTML) => {
-                    InspectorFrontendHost.copyText(outerHTML);
-                });
+                this._copyHTMLOfSelectedDOMNodes();
             });
 
             subMenus.copy.appendItem(WI.UIString("HTML (Formatted)"), () => {
-                this.representedObject.getOuterHTML()
-                .then((outerHTML) => {
-                    let workerProxy = WI.FormatterWorkerProxy.singleton();
-                    const includeSourceMapData = false;
-                    workerProxy.formatHTML(outerHTML, WI.indentString(), includeSourceMapData, ({formattedText}) => {
-                        InspectorFrontendHost.copyText(formattedText);
-                    });
-                });
+                this._copyHTMLOfSelectedDOMNodes({formatted: true});
             });
         }
 
@@ -932,6 +922,27 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
                 });
             }
         }
+    }
+
+    _copyHTMLOfSelectedDOMNodes({formatted} = {})
+    {
+        Promise.all(this.treeOutline.selectedTreeElements.map(async function(treeElement) {
+            let outerHTML = await treeElement.representedObject.getOuterHTML();
+            if (formatted) {
+                let workerProxy = WI.FormatterWorkerProxy.singleton();
+                const includeSourceMapData = false;
+                let {formattedText} = await workerProxy.formatHTML(outerHTML, WI.indentString(), includeSourceMapData);
+                outerHTML = formattedText || "";
+            }
+
+            return outerHTML;
+        }))
+        .then(function(outerHTMLs) {
+            InspectorFrontendHost.copyText(outerHTMLs.join("\n"));
+        })
+        .catch(function(error) {
+            WI.reportInternalError(error);
+        });
     }
 
     _startEditing()


Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -931,6 +922,63 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
         }
     }
 
+    _copyHTMLOfSelectedDOMNodes(formatted)
```
dcrousso: since this is optional please make it into a `{formatted} = {}`

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -931,6 +922,63 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
+    {
+        let abortController = new AbortController();
+        let abortSignal = AbortSignal.any([abortController.signal, AbortSignal.timeout(10_000)]);
+
+        WI.markInspectorBusy(abortSignal);
+
+        const copyHTMLText = (promiseResults) => {
+            // Might have timed-out midway while waiting to format some nodes. The output will be incomplete.
```
dcrousso: is this actually that likely?

if copying a bunch of stuff does for some reason take a while, the only thing this would affect is that we'd add something to the clipboard potentially out of order

id much rather have this PR just be about supporting copying multiple things and have a separate PR for deciding what to do when waiting for potentially long running operations

Source/WebInspectorUI/UserInterface/Views/Main.css
```diff
@@ -111,6 +111,10 @@ body.mac-platform.tahoe input[type=search]::-webkit-search-results-button {
     margin-inline: 3px;
 }
 
+body.busy {
+    cursor: wait;
```
dcrousso: if all this does is change the cursor UI (i.e. not actually block the user from clicking or whatnot) then what benefit does this really have?

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -931,6 +922,63 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
     }
 
+    _copyHTMLOfSelectedDOMNodes(formatted)
+    {
+        let abortController = new AbortController();
+        let abortSignal = AbortSignal.any([abortController.signal, AbortSignal.timeout(10_000)]);
+
+        WI.markInspectorBusy(abortSignal);
+
+        const copyHTMLText = (promiseResults) => {
```
dcrousso: this does not need to be an arrow function since it does not capture `this`

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -931,6 +922,63 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
+            // Might have timed-out midway while waiting to format some nodes. The output will be incomplete.
+            if (abortSignal.aborted)
+                return;
+
+            let HTMLText = promiseResults.reduce((accumulator, promiseResult) => {
+                return promiseResult.value ? accumulator + `\n` + promiseResult.value : accumulator;
```
dcrousso: can use a regular string `"\n"`

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -931,6 +922,63 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
+                            reject();
+                    });
+                });
+            });
+
+            Promise.allSettled(formatterPromises).then((results) => { copyHTMLText(results); });
+        })
+        .catch((error) => {
+            abortController.abort();
+            console.error(error);
```
dcrousso: please use `WI.reportInternalError` instead

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
rcaliman-apple: >is this actually that likely?

It is possible. I've been working on a separate patch with a file that takes 12+ seconds to prettify due to sheer size. It's an edge case, but it exits.

I think the UX is easier to reason about if you either get all content or no content on the clipboard. 

To be fair, this is a nice-to-have. I expect most copy operations to complete almost instantly.

Source/WebInspectorUI/UserInterface/Views/Main.css
rcaliman-apple: We don't show a "Copied!" message anywhere. In this context, it's meant to communicate that the request to copy HTML contents hasn't _yet_ put stuff on the clipboard. It's for those case of multi-second operations.

So you don't copy, switch to a text editor, paste, then be surprised nothing's there (or it's the previous clipboard contents)

dcrousso: that's assuming that the developer even notices the changed cursor

they'd also have to avoid anything else in Web Inspector that has a `cursor` set for it (which isn't everything but there's a good amount of things)

Source/WebInspectorUI/UserInterface/Base/Main.js
```diff
@@ -3176,6 +3176,22 @@ WI.engineeringSettingsAllowed = function() {
...
+
+    WI._busyStatePendingAbortCount ??= 0;
+    WI._busyStatePendingAbortCount++;
+
+    abortSignal.addEventListener("abort", () => {
+        WI._busyStatePendingAbortCount--;
+
+        if (WI._busyStatePendingAbortCount === 0)
+            document.body.classList.remove("busy");
+    })
```
the-chenergy:
```suggestion
    });
```

Source/WebInspectorUI/UserInterface/Base/Main.js
```diff
@@ -3176,6 +3176,22 @@ WI.engineeringSettingsAllowed = function() {
...
 
+WI.markInspectorBusy = function(abortSignal) {
+    console.assert(abortSignal instanceof AbortSignal, abortSignal);
+
+    document.body.classList.add("busy");
+
+    WI._busyStatePendingAbortCount ??= 0;
+    WI._busyStatePendingAbortCount++;
+
+    abortSignal.addEventListener("abort", () => {
```
the-chenergy: @dcrousso Can I ask when exactly do we prefer using `function ()` over lambda? Here we're also not capturing `this` (which we probably aren't ever in this `Main.js` file), so should we be using `function ()`?

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
dcrousso: ok sure i can see the pretty print case taking a while

but again id much rather have a more generalized solution since it's probably not an issue specific to this scenario (e.g. maybe some of the Export buttons could use this too?)

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -931,6 +922,63 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
+        WI.markInspectorBusy(abortSignal);
+
+        function copyHTMLText(promiseResults) {
+            // Might have timed-out midway while waiting to format some nodes. The output will be incomplete.
+            if (abortSignal.aborted)
+                return;
+
+            let HTMLText = promiseResults.reduce((accumulator, promiseResult) => {
+                return promiseResult.value ? accumulator + "\n" + promiseResult.value : accumulator;
+            }, "");
```
the-chenergy:
```suggestion
        let HTMLText = promiseResults
            .map(result => result.value)
            .filter(Boolean)
            .join("\n");
```

Source/WebInspectorUI/UserInterface/Base/Main.js
dcrousso: my general rule is to only ever use an arrow function when we actually need `this` or if it's a simple operation (e.g. `Array.prototype.map` where we want the `return` value and it can fit on a single line)

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -931,6 +922,63 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
+
+        Promise.allSettled(this.treeOutline.selectedTreeElements.map((treeElement) => treeElement.representedObject.getOuterHTML()))
+        .then((results) => {
+            if (!formatted) {
+                copyHTMLText(results);
+                return;
+            }
+
+            let workerProxy = WI.FormatterWorkerProxy.singleton();
+            const includeSourceMapData = false;
```
the-chenergy: I think we should put this right before its usage, right above `workerProxy.formatHTML`

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
rcaliman-apple: I agree this isn't a _great_ solution and it may be missed by some users. 

But it is _something_ to indicate an async operation that hasn't completed in a case where there's no other visual input. I've been uneasy with this opacity since implementing the HTML (Formatted) option that goes over to the formatter, knowing how long that can take.

We should have a better, holistic solution that covers stuff like Export too. 

(Maybe set progress messages shown somewhere: "Getting HTML text...", "Formatting HTML text...", "Copied to clipboard!")

But I don't think we should wait until then to provide some minimal visual indicator of long-running copy/formatting activity in this case. 

I'm inclined to keep this until implementing a better solution.

Source/WebInspectorUI/UserInterface/Views/Main.css
rcaliman-apple: Sure, but the user just chose a context menu option to "Copy". The very next operation is to attempt to paste that elsewhere. If they leave Web Inspector, we have no control. But if the clipboard is empty or stale and they come back and the cursor is `wait` there's a _chance_ they understand why. 

No visual indicator whatsoever leaves them stranded. And they'll likely repeat the operation and get the same result, expecting immediate copying to the clipboard.

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -931,6 +922,64 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
+        let abortSignal = AbortSignal.any([abortController.signal, AbortSignal.timeout(10_000)]);
+
+        WI.markInspectorBusy(abortSignal);
+
+        function copyHTMLText(promiseResults) {
+            // Might have timed-out midway while waiting to format some nodes. The output will be incomplete.
+            if (abortSignal.aborted)
+                return;
+
+            let HTMLText = promiseResults
```
burg: Weird variable naming. Can it be improved? `combinedHTMLText`?

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -931,6 +922,64 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
+                .map(result => result.value)
+                .filter(Boolean)
+                .join("\n");
+
+            InspectorFrontendHost.copyText(HTMLText);
+            abortController.abort();
+        }
+
+        Promise.allSettled(this.treeOutline.selectedTreeElements.map((treeElement) => treeElement.representedObject.getOuterHTML()))
+        .then((results) => {
```
burg: `results` is unclear in context, is this referring to a single node's outerHTML result or all of them? 

Second, the way that this is implemented, it has to wait on *get all getOuterHTML from backend* before starting to enqueue formatter work on the formatter proxy.

I think it would be preferable to have the promise chain work on a single `result` and then at the outer level it will do allSettled and copyText on multiple results. If any of the promises reject due to timeout, the node not being found, etc., then filter those out before calling copyText.

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -931,6 +922,64 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
+            if (!formatted) {
+                copyHTMLText(results);
```
burg: I think it would be better to return promise here with `result` and then the later `copyHTMLText` will be used whether or not its pretty-printed.

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -931,6 +922,64 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
+                        return;
+                    }
+
+                    // Give up if the timeout expired.
+                    abortSignal.addEventListener("abort", reject);
+
+                    let workerProxy = WI.FormatterWorkerProxy.singleton();
+                    const includeSourceMapData = false;
+
+                    workerProxy.formatHTML(result.value, WI.indentString(), includeSourceMapData, ({formattedText}) => {
```
burg: (maybe I'm just not used to this file, but)

It's a little weird to me that DOMTreeElement is getting knowledgeable about how to directly use workerProxy. Or maybe it just takes a lot of Promise Stuff to hook it up as-is?

Is there a better way to expose "async formatting as a service" to these context menu callbacks? Seems like it might be a good place to use async functions. Or maybe workerProxy.formatHTML can be adapted to use promises?

Source/WebInspectorUI/UserInterface/Base/Main.js
```diff
@@ -3176,6 +3176,22 @@ WI.engineeringSettingsAllowed = function() {
     return WI.isEngineeringBuild || WI.inspectorFrontendHostAllowsEngineeringSettings;
 }
 
+WI.markInspectorBusy = function(abortSignal) {
```
burg: I think this mechanism is a good use of AbortController but I would prefer to deliberate about its design separately from the actual enhancement.

If we keep this as interim, it should be added to AppController.js, as this is sort of an app-state concept.

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -931,6 +922,64 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
         }
     }
 
+    _copyHTMLOfSelectedDOMNodes({formatted} = {})
+    {
+        let abortController = new AbortController();
+        let abortSignal = AbortSignal.any([abortController.signal, AbortSignal.timeout(10_000)]);
```
burg: This timeout seems super generous. Alternatively, we expect it to be pathologically slow, per your test case.

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
rcaliman-apple: >results is unclear in context, is this referring to a single node's outerHTML result or all of them?

It's an array of objects containing all promises for `getOuterHTML` for each node, regardless of fulfillment success. Perhaps the name can be improved.

>Second, the way that this is implemented, it has to wait on get all getOuterHTML from backend before starting to enqueue formatter work on the formatter proxy.

This is intentional to preserve the order of the selected DOM nodes. It's been brought up as a requirement.
 
[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled#:~:text=The%20fulfillment%20value%20is%20an%20array%20of%20objects%2C%20each%20describing%20the%20outcome%20of%20one%20promise%20in%20the%20iterable%2C%20in%20the%20order%20of%20the%20promises%20passed%2C%20regardless%20of%20completion%20order)
> The fulfillment value is an array of objects, each describing the outcome of one promise in the iterable, in the order of the promises passed, regardless of completion order

Source/WebInspectorUI/UserInterface/Base/Main.js
rcaliman-apple: Ok, I'll take the AbortController and UI hint out of this patch and think of a better solution separately.

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
rcaliman-apple: It's 10s just because there can be some huge HTML chunks that need formatting. 
Most operations will fit well within 10s, but the value is there to cap any extreme edge cases.

Either way, I'm removing it from this patch.

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
rcaliman-apple: Removed the indicator in the latest patch. Will try a better solution separately

Source/WebInspectorUI/UserInterface/Views/Main.css
rcaliman-apple: Removed the indicator in the latest patch. Will try a better solution separately

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
rcaliman-apple: I added a new static method, `FormatterWorkerProxy.formatHTMLPromise()`, to wrap this complexity.

Source/WebInspectorUI/UserInterface/Proxies/FormatterWorkerProxy.js
```diff
@@ -43,6 +43,18 @@ WI.FormatterWorkerProxy = class FormatterWorkerProxy
         return FormatterWorkerProxy.instance;
     }
 
+    static formatHTMLPromise(sourceText, indentString, includeSourceMapData)
```
dcrousso: rather than create a new method id suggest you just modify `performAction` to return a `Promise` if not given a callback
```diff
 performAction(actionName)
 {
+    console.assert(typeof actionName === "string", "performAction should always have an actionName");
+
     let callId = this._nextCallId++;
+
     let callback = arguments[arguments.length - 1];
+    if (typeof callback !== "function") {
+        return new Promise((resolve, reject) => {
+            this._callbacks.set(callId, resolve);
+            this._postMessage({
+                callId,
+                actionName,
+                actionArguments: Array.prototype.slice.call(arguments, 1),
+            });
+        });
+    }
+
     let actionArguments = Array.prototype.slice.call(arguments, 1, arguments.length - 1);
 
-    console.assert(typeof actionName === "string", "performAction should always have an actionName");
-    console.assert(typeof callback === "function", "performAction should always have a callback");
-
     this._callbacks.set(callId, callback);
     this._postMessage({callId, actionName, actionArguments});
 }
```
that way we can just use the same method name without passing a callback to get a `Promise` (which also matches what we do for all backend commands and whatnot)

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -814,114 +814,102 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
-                this.representedObject.getOuterHTML()
-                .then((outerHTML) => {
-                    let workerProxy = WI.FormatterWorkerProxy.singleton();
-                    const includeSourceMapData = false;
-                    workerProxy.formatHTML(outerHTML, WI.indentString(), includeSourceMapData, ({formattedText}) => {
-                        InspectorFrontendHost.copyText(formattedText);
-                    });
-                });
+                const formatted = true;
+                this._copyHTMLOfSelectedDOMNodes({formatted});
```
dcrousso:
```suggestion
                this._copyHTMLOfSelectedDOMNodes({formatted: true});
```

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -934,6 +922,33 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
         }
     }
 
+    _copyHTMLOfSelectedDOMNodes({formatted} = {})
+    {
+        Promise.allSettled(this.treeOutline.selectedTreeElements.map((treeElement) => treeElement.representedObject.getOuterHTML()))
```
dcrousso: why do we have to wait for _everything_ to come from the backend before we start formatting any of them?
```js
Promise.all(this.treeOutline.selectedTreeElements.map(async function(treeElement) {
    let outerHTML = await treeElement.representedObject.getOuterHTML();
    if (formatted) {
        let workerProxy = WI.FormatterWorkerProxy.singleton();
        const includeSourceMapData = false;
        outerHTML = await workerProxy.formatHTML(outerHTML, WI.indentString(), includeSourceMapData);
    }
    return outerHTML;
})
.then(function(outerHTMLs) {
    InspectorFrontendHost.copyText(outerHTMLs.join("\n"));
})
.catch(function(error) {
    WI.reportInternalError(error);
});
```
as a nice benefit this is also a much closer implementation to `WI.DOMTreeContentView.prototype.handleCopyEvent`

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
```diff
@@ -934,6 +921,27 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
     }
 
+    _copyHTMLOfSelectedDOMNodes({formatted} = {})
+    {
+        Promise.all(this.treeOutline.selectedTreeElements.map(async function(treeElement) {
+            let outerHTML = await treeElement.representedObject.getOuterHTML();
+            if (formatted) {
+                let workerProxy = WI.FormatterWorkerProxy.singleton();
+                const includeSourceMapData = false;
+                let result = await workerProxy.formatHTML(outerHTML, WI.indentString(), includeSourceMapData);
```
dcrousso: NIT: let's destructure this
```js
let {formattedText} = ...

File: Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js [hunk 21]
```diff
@@ -934,6 +921,27 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
...
+                outerHTML = result.formattedText || "";
```
dcrousso: is it actually possible for `formattedText` to be empty?

dcrousso (approved):
nice work! thanks for iterating with me! :)

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
rcaliman-apple: `formattedText` can be `null` in case there's an error:
https://github.com/WebKit/WebKit/blob/main/Source/WebInspectorUI/UserInterface/Workers/Formatter/FormatterWorker.js#L111

And there's the edge case of copying an empty text node, sibling of a DOM node 😅

PR #58644, opened 2026-02-13
[WTF] Switch to SequesteredStack via new DeferredStack mechanism

Achierius:
#### 6d5322b005e5142df554852db841c860ca447b71
<pre>
[WTF] Switch to SequesteredStack via new DeferredStack mechanism
<a href="https://bugs.webkit.org/show_bug.cgi?id=307825">https://bugs.webkit.org/show_bug.cgi?id=307825</a>
<a href="https://rdar.apple.com/170333133">rdar://170333133</a>

Reviewed by Mark Lam.

It is desirable to be able to create threads whose backing stack-memory
is guarded by special memory protections. However, as we are doing this
wholly in userspace, the OS does not know to expect anything different
about the stack we pass to pthread_attr_setstack: this means that a call
to pthread_create using such a stack will necessarily fail, as the OS
does not engage the proper memory-access permissions prior to setting up
the new thread&apos;s stack.
So, what we instead want to do is be able to start a WTF::Thread with
one stack (provided by the OS, without special memory protections), then
&apos;teleport&apos; to the new protected stack via a trampoline.

What this patch does is add a new Kind to StackAllocationSpecification
which provides both a size (for the native OS stack -- we want this to
be small, as it&apos;s not going to be used for very much) and an allocation
(for the &apos;deferred&apos; stack that will be switched to during
initialization). It then makes SequesteredAutomaticThreads use this
mechanism for their protected stack.

In a subsequent patch, we will add the code to protect the special stack
and engage the proper memory protections prior to calling the
trampoline.

Canonical link: <a href="https://commits.webkit.org/307676@main">https://commits.webkit.org/307676@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/a9fc821379cd7c952497f2b084b6dbdb54a0b85d

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/145092 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/17773 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/9548 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/153763 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/98728 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/18256 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/17665 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/111569 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/98728 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/78062f38-9a49-492c-94ab-00cb3f53e68a) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/148055 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/13927 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/130328 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/92467 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/13295 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/11055 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/1208 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/137083 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/122815 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/7085 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/156075 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/5901 "Built successfully and passed tests") | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/17624 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/8173 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/119573 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/17670 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/14710 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/119903 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30761 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/15689 "Passed tests") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/128343 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/73288 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/17245 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/6602 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/176381 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/16981 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/81024 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/45372 "Passed tests") | | 
| | [  ~~🛠 watch~~](https://ews-build.webkit.org/#/builders/163/builds/17190 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/17045 "Built successfully") | | | |

### DIFF
--- a/Source/WTF/wtf/SequesteredAutomaticThread.cpp
+++ b/Source/WTF/wtf/SequesteredAutomaticThread.cpp
@@ -56,7 +56,7 @@ SequesteredAutomaticThread::~SequesteredAutomaticThread()
 
 StackAllocationSpecification SequesteredAutomaticThread::stackSpecification()
 {
-    return StackAllocationSpecification::CustomStack(m_stack.span());
+    return StackAllocationSpecification::DeferredStack(m_stack.span(), 1 * MB);
 }
 
 #endif // USE(PROTECTED_JIT_STACKS)
--- a/Source/WTF/wtf/StackAllocation.h
+++ b/Source/WTF/wtf/StackAllocation.h
@@ -39,6 +39,7 @@ public:
         Default = 0, // OS provided stack, OS default size
         SizeOnly, // OS provided stack, specified size
         SizeAndLocation, // Preallocated stack with known size
+        DeferredStack, // OS provided stack initially, then hop to user-preallocated stack
     };
 
     StackAllocationSpecification() = default;
@@ -57,23 +58,54 @@ public:
         return s;
     }
 
-    Kind kind() const { return m_kind; }
-    bool isKind(Kind kind) const { return this->kind() == kind; }
-
-    std::size_t sizeBytes() const
+    static StackAllocationSpecification DeferredStack(std::span<std::byte> deferredStack, std::size_t osStackSizeBytes)
     {
-        RELEASE_ASSERT(m_kind == Kind::SizeOnly || m_kind == Kind::SizeAndLocation);
+        StackAllocationSpecification s(Kind::DeferredStack);
+        s.storage.deferredStack.stack = deferredStack;
+        s.storage.deferredStack.osStackSize = osStackSizeBytes;
+        return s;
+    }
+
+    constexpr Kind kind() const { return m_kind; }
+    constexpr bool isKind(Kind kind) const { return this->kind() == kind; }
+
+    constexpr std::size_t osStackSize() const
+    {
+        if (m_kind == Kind::SizeOnly)
+            return storage.sizeBytes;
+        if (m_kind == Kind::DeferredStack)
+            return storage.deferredStack.osStackSize;
+        RELEASE_ASSERT_NOT_REACHED();
+    }
+
+    constexpr std::size_t effectiveSize() const
+    {
         if (m_kind == Kind::SizeOnly)
             return storage.sizeBytes;
         if (m_kind == Kind::SizeAndLocation)
             return storage.stack.size_bytes();
+        if (m_kind == Kind::DeferredStack)
+            return storage.deferredStack.stack.size_bytes();
         RELEASE_ASSERT_NOT_REACHED();
     }
 
-    std::span<std::byte> stackSpan() const
+    constexpr std::span<std::byte> stackSpan() const
     {
-        RELEASE_ASSERT(m_kind == Kind::SizeAndLocation);
-        return storage.stack;
+        if (m_kind == Kind::SizeAndLocation)
+            return storage.stack;
+        if (m_kind == Kind::DeferredStack)
+            return storage.deferredStack.stack;
+        RELEASE_ASSERT_NOT_REACHED();
+    }
+
+    void* stackOrigin() const
+    {
+        RELEASE_ASSERT(kind() == Kind::SizeAndLocation || kind() == Kind::DeferredStack);
+        void* bound = stackSpan().data();
+        WTF_ALLOW_UNSAFE_BUFFER_USAGE_BEGIN;
+        void* origin = reinterpret_cast<char*>(bound) + stackSpan().size_bytes();
+        WTF_ALLOW_UNSAFE_BUFFER_USAGE_END;
+        return origin;
     }
 
 private:
@@ -86,6 +118,10 @@ private:
         constexpr Storage() : sizeBytes(0) { }
         std::size_t sizeBytes;
         std::span<std::byte> stack;
+        struct {
+            std::span<std::byte> stack;
+            std::size_t osStackSize;
+        } deferredStack;
     } storage { };
 };
 
--- a/Source/WTF/wtf/StackSwitch.cpp
+++ b/Source/WTF/wtf/StackSwitch.cpp
@@ -1,0 +1,103 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL APPLE INC. OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ * OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include <wtf/StackSwitch.h>
+
+#include <wtf/Assertions.h>
+#include <wtf/Compiler.h>
+#include <wtf/InlineASM.h>
+#include <wtf/Threading.h>
+
+namespace WTF {
+
+// C-linkage wrapper for Thread::entryPointFinishSetup so it can be called from assembly
+extern "C" void threadEntryPointFinishSetupWrapper(void* context);
+extern "C" void threadEntryPointFinishSetupWrapper(void* context)
+{
+    Thread::entryPointFinishSetup(context);
+}
+
+#if CPU(ARM64E)
+
+// Trampoline that switches from the OS-provided stack to a custom sequestered stack.
+// Parameters:
+//   x0 = context pointer (passed as argument to Thread::entryPointFinishSetup)
+//   x1 = new stack pointer (top of stack, since stacks grow down)
+//
+// This function:
+// 1. Saves callee-saved registers on the current (OS) stack
+// 2. Switches to the new stack
+// 3. Sets up the rest of its frame (saved sp, lr) on the new stack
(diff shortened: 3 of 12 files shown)


Source/WTF/wtf/SequesteredAutomaticThread.cpp
```diff
@@ -56,7 +56,7 @@ SequesteredAutomaticThread::~SequesteredAutomaticThread()
 
 StackAllocationSpecification SequesteredAutomaticThread::stackSpecification()
 {
-    return StackAllocationSpecification::CustomStack(m_stack.span());
+    return StackAllocationSpecification::DeferredStack(m_stack.span(), 1 * MB);
```
MenloDorian, line 59: Hmmm, I wonder if we need to make this 1 MB runtime configurable.  For now, this is probably good enough if we only ever use this for JIT threads.

Source/WTF/wtf/StackAllocation.h
```diff
@@ -39,6 +39,7 @@ class StackAllocationSpecification {
         Default = 0, // OS provided stack, OS default size
         SizeOnly, // OS provided stack, specified size
         SizeAndLocation, // Preallocated stack with known size
+        DeferredStack, // OS provided stack initially, then hop to preallocated stack
```
MenloDorian: maybe "hop to preallocated stack" => "hop to user preallocated stack" instead?

I'm also not sure if DeferredStack is the best name (or I don't understand the rationale for the name).  But we can always rename later if we think of a better name.

Source/WTF/wtf/StackAllocation.h
```diff
@@ -57,23 +58,57 @@ class StackAllocationSpecification {
         return s;
     }
 
+    static StackAllocationSpecification DeferredStack(std::span<std::byte> stack, std::size_t osStackSizeBytes)
```
MenloDorian: Can't we use `stack.size()` instead of explicitly passing `osStackSizeBytes`?  It is the size of the span, right?  Oh, `osStackSizeBytes` is for the OS stack part as opposed to the user allocated stack.   I guess it makes sense that it's not the same as the span size then.  It may be clearer to rename `stack` to user allocated stack (or whatever nomenclature we use to refer to the "other" stack).

Source/WTF/wtf/StackAllocation.h
```diff
@@ -57,23 +58,57 @@ class StackAllocationSpecification {
...
+    {
+        StackAllocationSpecification s(Kind::DeferredStack);
+        s.storage.deferredStack.stack = stack;
+        s.storage.deferredStack.startingSize = osStackSizeBytes;
```
MenloDorian: It looks like `startingSize` is only relevant to the `deferredStack`.  Can we call this field `osStackSize` or `osStartingStackSize` instead?

Source/WTF/wtf/StackAllocation.h
```diff
@@ -57,23 +58,57 @@ class StackAllocationSpecification {
...
+        if (m_kind == Kind::SizeOnly)
+            return storage.sizeBytes;
+        if (m_kind == Kind::DeferredStack)
+            return storage.deferredStack.startingSize;
+        RELEASE_ASSERT_NOT_REACHED();
+    }
+
+    std::size_t effectiveSize() const
+    {
+        RELEASE_ASSERT(m_kind == Kind::SizeOnly || m_kind == Kind::SizeAndLocation || m_kind == Kind::DeferredStack);
```
MenloDorian: This RELEASE_ASSERT is redundant given the RELEASE_ASSERT_NOT_REACHED below.  You can change it to a Debug ASSERT or remove it.

Source/WTF/wtf/StackAllocation.h
```diff
@@ -57,23 +58,57 @@ class StackAllocationSpecification {
...
+    }
+
     Kind kind() const { return m_kind; }
     bool isKind(Kind kind) const { return this->kind() == kind; }
 
-    std::size_t sizeBytes() const
+    std::size_t osStackSize() const
     {
-        RELEASE_ASSERT(m_kind == Kind::SizeOnly || m_kind == Kind::SizeAndLocation);
+        RELEASE_ASSERT(m_kind == Kind::SizeOnly || m_kind == Kind::DeferredStack);
```
MenloDorian: Ditto, the RELEASE_ASSERT is redundant given the RELEASE_ASSERT_NOT_REACHED below.  Change to ASSERT or remove.

Source/WTF/wtf/StackAllocation.h
```diff
@@ -57,23 +58,57 @@ class StackAllocationSpecification {
...
+        if (m_kind == Kind::DeferredStack)
+            return storage.deferredStack.stack.size_bytes();
         RELEASE_ASSERT_NOT_REACHED();
     }
 
     std::span<std::byte> stackSpan() const
     {
-        RELEASE_ASSERT(m_kind == Kind::SizeAndLocation);
-        return storage.stack;
+        RELEASE_ASSERT(m_kind == Kind::SizeAndLocation || m_kind == Kind::DeferredStack);
```
MenloDorian: Ditto.

Source/WTF/wtf/StackAllocation.h
```diff
@@ -57,23 +58,57 @@ class StackAllocationSpecification {
...
+        if (m_kind == Kind::SizeAndLocation)
+            return storage.stack;
+        if (m_kind == Kind::DeferredStack)
+            return storage.deferredStack.stack;
+        RELEASE_ASSERT_NOT_REACHED();
+    }
+
+    void* stackOrigin() const
+    {
+        RELEASE_ASSERT(kind() == Kind::SizeAndLocation || kind() == Kind::DeferredStack);
```
MenloDorian, line 103: Ditto.

Source/WTF/wtf/Threading.cpp
```diff
@@ -228,6 +233,29 @@ void Thread::initializeInThread()
 
 void Thread::entryPoint(NewThreadContext* newThreadContext)
 {
+    std::optional<void*> stackSwitchTarget { };
+    {
+        MutexLocker locker(newThreadContext->mutex);
+        if (newThreadContext->stackSpec.kind() == StackAllocationSpecification::Kind::DeferredStack) {
+            // The OS-provided stack we start on is not our 'actual stack';
```
MenloDorian: nit: "actual stack" to "target work stack"?

Source/WTF/wtf/StackSwitch.cpp
```diff
@@ -0,0 +1,102 @@
...
+    // Switch to new stack (x0 contains new stack pointer)
+    "mov sp, x0" "\n"
+    // Create a proper frame on the new stack
+    // This frame chains back to the OS stack frame
+    // of our caller
+    "stp x29, x30, [sp, #-16]!" "\n"
+    "mov x29, sp" "\n"
+
+    // Set up arguments and call Thread::entryPointFinishSetup
+    "mov x0, x1" "\n" // x0 = context (first argument)
```
MenloDorian: If you flip the arguments for `callThreadEntryPointFinishSetupWithNewStack`, then `x0` will already be the correct argument here, and you can just `mov sp, x1` above instead.

Source/WTF/wtf/StackAllocation.h
Achierius: This one isn't redundant, since it doesn't have a RELEASE_ASSERT_NOT_REACHED below

Source/WTF/wtf/StackSwitch.cpp
```diff
@@ -0,0 +1,102 @@
...
+    "bl " SYMBOL_STRING(threadEntryPointFinishSetupWrapper) "\n"
+
```
MenloDorian, line 80: Don't you have to `ldp x29, x30, ...` here?  Otherwise, `lr` (which is volatile) will contain a junk value from callees.  Also, you won't have restored `fp`.

Source/WTF/wtf/StackAllocation.h
Achierius: Done

Source/WTF/wtf/StackAllocation.h
Achierius: Renamed

Source/WTF/wtf/StackAllocation.h
Achierius: Changed comment -- the rationale behind the name is that instead of installing it right away, we "defer" doing so until after the thread begins execution. But I agree, I'm not the biggest fan of the name, it's just what I was able to come up with -- happy to change later

MenloDorian (approved):
r=me with fixes.

Source/WTF/wtf/SequesteredAutomaticThread.cpp
Achierius: Yeah, my plan is to reduce it later -- I didn't want to take on more stability risk for the time being though, since this is only really consuming VA and I'm only deploying this to macOS, where we have plenty

PR #49437, opened 2025-08-15
[iOS] Find on page highlight is broken for -webkit-user-select:none styled elements (affects many sites, including nytimes.com, bizjournal.com, cnn.com, developer.mozilla.org)

megangardner:
#### 47db418a5ff00139a0068296c759b0c776ef904a
<pre>
[iOS] Find on page highlight is broken for -webkit-user-select:none styled elements (affects many sites, including nytimes.com, bizjournal.com, cnn.com, developer.mozilla.org)
<a href="https://bugs.webkit.org/show_bug.cgi?id=297441">https://bugs.webkit.org/show_bug.cgi?id=297441</a>
<a href="https://rdar.apple.com/8268929">rdar://8268929</a>

Reviewed by Aditya Keerthi.

When we try and make a text indicator for a selection that was a
range that is webkit-user-select-none then the selection collapses
and the indicator does not have an associated image.

We should fallback to creating an indicator of the range painted
without the alterations for the indicator, for now. This will allow us
to at least have somewhat of an indicator, even if it is not as obvious.

We also shouldn&apos;t block on the existence of a selection rect for painting,
as if we have text rects and an image, that is also workable for showing
an indicator.

* Source/WebCore/page/TextIndicator.cpp:
(WebCore::TextIndicator::createWithRange):
* Source/WebKit/WebProcess/WebPage/WebFoundTextRangeController.cpp:
(WebKit::WebFoundTextRangeController::drawRect):

Canonical link: <a href="https://commits.webkit.org/298785@main">https://commits.webkit.org/298785@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/1517e21592de1fdd948c8e1928889100377d066e

| Misc | iOS, visionOS, tvOS & watchOS  | macOS  | Linux |  Windows | Apple Internal |
| ----- | ---------------------- | ------- |  ----- |  --------- | ------ |
| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/116640 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/131/builds/36304 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/138/builds/26866 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/122714 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/67212 "Built successfully") | ⏳ 🛠 ios-apple 
[...]

PR #58696, opened 2026-02-14
Fix find in page for content with -webkit-user-select: none

colelao:
#### 635808d7603e771d21ea17bccf831aa917ad5b5a
<pre>
Fix find in page for content with -webkit-user-select: none
<a href="https://rdar.apple.com/85000322">rdar://85000322</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=307877">https://bugs.webkit.org/show_bug.cgi?id=307877</a>

Reviewed by Ryosuke Niwa and Megan Gardner.

The find-in-page feature relies on the selection API for finding strings. If
a string is found and it looks to be the same selection as the last found string,
it will redo the search.

The selection api will collapse user-select: none text, so when you have two matches
in a user-select: none block, both matches will be collapsed to the same point.

Because they are the same point, they look like the same selection, and the match
is skipped. To address this, I saved the start and end positions BEFORE
canonicalization, which represent the positions before adjusting to account for
user-select: none. Then, these saved points are used when comparing found strings.

* Source/WebCore/editing/Editor.cpp:
(WebCore::Editor::findString):
(WebCore::Editor::rangeOfString):
* Source/WebCore/editing/VisibleSelection.cpp:
(WebCore::VisibleSelection::rawRange const):
(WebCore::VisibleSelection::setBaseAndExtentToDeepEquivalents):
* Source/WebCore/editing/VisibleSelection.h:
(WebCore::VisibleSelection::rawStart const):
(WebCore::VisibleSelection::rawEnd const):

Canonical link: <a href="https://commits.webkit.org/308508@main">https://commits.webkit.org/308508@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/af18a7ac501331926e48eb34c9fd242dd8d349c1

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/147732 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/20417 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/14009 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/156415 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/101147 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/4734ddb0-115c-45af-9f7c-97b9382f6154) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/149605 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/20875 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/20317 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/113890 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/81218 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/84249d0d-5529-43ce-acce-7fb73ca5ba3c) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/150694 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/16130 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/132686 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94650 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/e868406f-113e-4773-899b-3a04912deb8c) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/15292 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/13072 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/3855 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/124888 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/10600 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/158750 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/1884 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/12081 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/121916 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/20216 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/16987 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/122117 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/31282 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/20227 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/132390 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/76342 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/17633 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/9161 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/19832 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/83594 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/19561 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/19712 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/19619 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/editing/Editor.cpp
+++ b/Source/WebCore/editing/Editor.cpp
@@ -4042,7 +4042,10 @@ std::optional<SimpleRange> Editor::findString(const String& target, FindOptions 
         document->updateLayoutIgnorePendingStylesheets({ LayoutOptions::TreatContentVisibilityAutoAsVisible, LayoutOptions::TreatRevealedWhenFoundAsVisible });
         Style::PostResolutionCallbackDisabler disabler(document);
         VisibleSelection selection = document->selection().selection();
-        resultRange = rangeOfString(target, selection.firstRange(), options);
+        auto referenceRange = selection.firstRange();
+        if (!referenceRange || referenceRange->collapsed())
+            referenceRange = selection.range();
+        resultRange = rangeOfString(target, referenceRange, options);
     }
 
     if (!resultRange)
@@ -4106,7 +4109,10 @@ std::optional<SimpleRange> Editor::rangeOfString(const String& target, const std
     // If we started in the reference range and the found range exactly matches the reference range, find again.
     // Build a selection with the found range to remove collapsed whitespace.
     // Compare ranges instead of selection objects to ignore the way that the current selection was made.
-    if (startInReferenceRange && VisibleSelection(resultRange).toNormalizedRange() == referenceRange) {
+    auto resultSelection = VisibleSelection(resultRange);
+    auto normalizedRange = resultSelection.toNormalizedRange();
+    auto resultRangeForComparison = !normalizedRange || normalizedRange->collapsed() ? resultSelection.range() : normalizedRange;
+    if (startInReferenceRange && resultRangeForComparison == referenceRange) {
         searchRange = makeRangeSelectingNodeContents(document);
         start(searchRange, options) = end(*referenceRange, options);
         if (shadowTreeRoot)
--- a/LayoutTests/editing/text-iterator/findString-selection-disabled-multiple-expected.txt
+++ b/LayoutTests/editing/text-iterator/findString-selection-disabled-multiple-expected.txt
@@ -1,0 +1,11 @@
+v_
+findme
+
+findme
+
+
+PASS find first string with selection disabled
+PASS find second string with selection disabled
+PASS find first string with selection enabled
+PASS find second string with selection enabled
+
--- a/LayoutTests/editing/text-iterator/findString-selection-disabled-multiple.html
+++ b/LayoutTests/editing/text-iterator/findString-selection-disabled-multiple.html
@@ -1,0 +1,43 @@
+<!DOCTYPE html>
+<html>
+<head>
+<title>Find multiple strings with selection disabled</title>v_
+<meta charset="utf-8">
+<script src="../../resources/testharness.js"></script>
+<script src="../../resources/testharnessreport.js"></script>
+<style type="text/css">
+ .selectionDisabled { -webkit-user-select: none; }
+ .selectionEnabled { -webkit-user-select: auto; }
+</style>
+</head>
+<body>
+    <div id="container">
+        <p id="first">findme</p>
+        <p id="second">findme</p>
+    </div>
+<script>
+
+async function testFindString(disableSelection, expectedNode) {
+    container.className = disableSelection ? "selectionDisabled" : "selectionEnabled";
+
+    const result = await testRunner.findString("findme", ["WrapAround"]);
+
+    const selection = window.getSelection();
+    const range = selection.getRangeAt(0);
+    const resultNode = range.startContainer;
+    assert_equals(resultNode.parentElement, expectedNode);
+}
+
+async function runTest(disableSelection, nodeId) {
+        const expectedNode = document.getElementById(nodeId);
+        await promise_test(() => testFindString(disableSelection, expectedNode), `find ${nodeId} string with selection ${disableSelection ? "disabled" : "enabled"}`);
+}
+
+for (disableSelection of [true, false]) {
+    runTest(disableSelection, "first");
+    runTest(disableSelection, "second");
+}
+
+</script>
+</body>
+</html>


Source/WebCore/page/TextIndicator.cpp
```diff
@@ -310,6 +316,13 @@ static bool initializeIndicator(TextIndicatorData& data, LocalFrame& frame, cons
     if (RefPtr document = frame.document())
         document->updateLayoutIgnorePendingStylesheets();
 
+    bool hasUserSelectNone = false;
+    if (auto* renderer = range.start.container->renderer())
```
aprotyas: nit: safercpp?

```suggestion
    if (CheckedPtr renderer = range.start.container->renderer())
```

Source/WebCore/editing/Editor.cpp
```diff
@@ -4053,8 +4053,11 @@ std::optional<SimpleRange> Editor::findString(const String& target, FindOptions
     if (!resultRange)
         return std::nullopt;
 
-    if (!options.contains(FindOption::DoNotSetSelection))
-        document->selection().setSelection(VisibleSelection(*resultRange));
+    if (!options.contains(FindOption::DoNotSetSelection)) {
+        VisibleSelection newSelection;
+        newSelection.setWithoutValidation(makeDeprecatedLegacyPosition(resultRange->start), makeDeprecatedLegacyPosition(resultRange->end));
```
rniwa: This is a very exotic thing to do. Why is this needed?
Which validation step is avoiding `-webkit-user-select: none`?

rniwa (changes requested):
We need a test one way or another.

Source/WebCore/editing/Editor.cpp
colelao: I think this is the right thing to do here. After a find operation, we want to select the found result. If we just construct a `VisibleSelection` like normal, the constructor will call `VisibleSelection::validate()`, which will set the start and end of the range to the same point, collapsing the range.

Without this change, matches will be skipped when navigating with the next/previous buttons during a find session.

It makes sense to do this without validation, as the range is coming from a trusted source (`Editor::rangeOfString()`).

Source/WebCore/page/TextIndicator.cpp
```diff
@@ -73,6 +73,20 @@ Ref<TextIndicator> TextIndicator::create(const TextIndicatorData& data)
...
 }
 
+static ALWAYS_INLINE bool rangeIndicatesCurrentSelection(const SimpleRange& range, Document& document)
+{
+    auto selectionRange = document.selection().selection().toNormalizedRange();
+    auto normalizedRange = VisibleSelection(range).toNormalizedRange();
+    return selectionRange && selectionRange == normalizedRange;
+}
+
+static ALWAYS_INLINE void handleBadSelection(const SimpleRange& range, Document& document, OptionSet<TextIndicatorOption>& options)
```
megangardner: I don't think this is a good name, the selection isn't bad, it's just in content that we have a harder time painting in our current setup. Maybe "checkForUnpaintedSelection"?

Source/WebCore/editing/Editor.cpp
```diff
@@ -4053,8 +4053,13 @@ std::optional<SimpleRange> Editor::findString(const String& target, FindOptions
     if (!resultRange)
         return std::nullopt;
 
-    if (!options.contains(FindOption::DoNotSetSelection))
-        document->selection().setSelection(VisibleSelection(*resultRange));
+    if (!options.contains(FindOption::DoNotSetSelection)) {
+        VisibleSelection newSelection;
+        const Position start = makeDeprecatedLegacyPosition(resultRange->start);
+        const Position end = makeDeprecatedLegacyPosition(resultRange->end);
```
megangardner: Can you use makeContainerOffsetPosition instead of these deprecated calls?

Source/WebCore/editing/Editor.cpp
rniwa: @colelao : Whether the range comes from a trusted source or not isn't relevant to why we validate `VisibleSelection`. `VisibleSelection` maintains some invariants to other clients.

We need to figure out which one of those invariants are causing this collapsing of VisibleSelection end points, and change the validation logic. Or else, editing code will sooner or later will do the same collapsing operation down the line.

colelao: @rniwa Maybe I am misunderstanding. I was under the impression that `VisibleSelection` should be collapsed for `-webkit-user-select: none` text. Should the constructor create a valid selection instead?

rniwa: Well, but then why are you working around it here? Either `VisibleSelection` represents the selection or not. We can't have it both ways.

megangardner: So, the issue is that because a VisibleSelection is so tied to the working of 'selection' that it is not valid to have a VisibleSelection that contains `user-select:none` I ran into this when trying to fix this problem on iOS. This means that in order for things like 'find' to work, we need to decouple the selection painting from selection, which is... very difficult. Which is why I was only able to get it to paint a rectangle. So in short, we can't have VisibleSelection represent a selection when we are trying to paint the find indicator (but we are using the selection code to make the TextIndicator). Decouping this will be a much bigger project, but doing a workaround for now is preferable to letting find skip over `user-select:none` content. See https://github.com/WebKit/WebKit/pull/49437/ for the work I did for iOS.

Source/WebCore/editing/FrameSelection.cpp
```diff
@@ -366,6 +366,8 @@ bool FrameSelection::setSelectionWithoutUpdatingAppearance(const VisibleSelectio
     bool closeTyping = options.contains(SetSelectionOption::CloseTyping);
     bool shouldClearTypingStyle = options.contains(SetSelectionOption::ClearTypingStyle);
 
+    m_isForFindInPage = options.contains(SetSelectionOption::ForFindInPage);
```
rniwa: I don't think we want to add this new option. Selection code is already very complex. Adding a new option like this is a foot gun on its own. As we discussed over slack, we should probably use `FrameSelection::associateLiveRange` instead.

Source/WebKit/WebProcess/WebPage/FindController.cpp
```diff
@@ -68,6 +69,7 @@ WTF_MAKE_TZONE_ALLOCATED_IMPL(FindController);
 
 FindController::FindController(WebPage* webPage)
     : m_webPage(webPage)
+    , m_findSession(DocumentFindSession::create())
```
rniwa: Do we really need a separate class for this? Why can't FindController just remember the last match?
Also it's very misleading for an object that exists per page to have "Document" prefix.

Source/WebCore/page/DocumentFindSession.h
```diff
@@ -0,0 +1,57 @@
...
+enum class TextIndicatorPresentationTransition : uint8_t;
+enum class TextIndicatorOption : uint16_t;
+
+class DocumentFindSession : public RefCounted<DocumentFindSession> {
+public:
+    WEBCORE_EXPORT static Ref<DocumentFindSession> create();
+
+    WEBCORE_EXPORT const std::optional<SimpleRange>& currentMatch() const;
+    WEBCORE_EXPORT void setCurrentMatch(std::optional<SimpleRange>&&);
+    WEBCORE_EXPORT LocalFrame* frameForCurrentMatch() const;
```
rniwa: Typically, we only prefix an object that exists one per document with "Document".
But why do we need this class in the first place? Why can't FindController hold onto this information?

Source/WebCore/page/DocumentFindSession.h
```diff
@@ -0,0 +1,57 @@
...
+#include <wtf/OptionSet.h>
+#include <wtf/RefCounted.h>
+
+namespace WebCore {
+class LocalFrame;
+class TextIndicator;
+enum class TextIndicatorPresentationTransition : uint8_t;
+enum class TextIndicatorOption : uint16_t;
+
+class DocumentFindSession : public RefCounted<DocumentFindSession> {
```
rniwa: Why does this object need to be `RefCounted` when it has a single owner model (`Page`)?

Source/WebKit/WebProcess/WebPage/FindController.cpp
```diff
@@ -143,7 +145,7 @@ RefPtr<LocalFrame> FindController::frameWithSelection(Page* page)
 void FindController::updateFindUIAfterPageScroll(bool found, const String& string, OptionSet<FindOptions> options, unsigned maxMatchCount, WebCore::DidWrap didWrap, std::optional<FrameIdentifier> idOfFrameContainingString, CompletionHandler<void(std::optional<WebCore::FrameIdentifier>, Vector<IntRect>&&, uint32_t, int32_t, bool)>&& completionHandler)
 {
     RefPtr webPage { m_webPage.get() };
-    RefPtr selectedFrame = frameWithSelection(protect(webPage->corePage()).get());
+    RefPtr selectedFrame = m_findSession->frameForCurrentMatch();
```
rniwa: Now we could have a mismatch between the frame which contains the selection and the frame which has the current "find" match. That seems problematic. Namely, all the code which currently gets the focused frame will no longer return the frame in which the current match is found.

colelao: @rniwa  I think this implies that we shouldn't rely on selections to do things like scrolling. I can add a way to do this with a SimpleRange instead, so we can decouple from selections here too.

Source/WebKit/WebProcess/WebPage/FindController.cpp
colelao: @rniwa I am hoping to have this new class take on more responsibility in the future. I also think this class can lead to some uniformity between the FindController and the WebFoundTextRangeController.

rniwa: What are those responsibilities? I don't think we should introduce this class unless we're absolutely sure we need it.

Source/WebKit/WebProcess/WebPage/FindController.cpp
rniwa: I'm not sure if that's the right thing to do here. I'm pretty sure finding text moves the selection to where the found selection is in other browser too.

Source/WebCore/editing/Editor.cpp
```diff
@@ -4042,7 +4042,10 @@ std::optional<SimpleRange> Editor::findString(const String& target, FindOptions
         document->updateLayoutIgnorePendingStylesheets({ LayoutOptions::TreatContentVisibilityAutoAsVisible, LayoutOptions::TreatRevealedWhenFoundAsVisible });
         Style::PostResolutionCallbackDisabler disabler(document);
         VisibleSelection selection = document->selection().selection();
-        resultRange = rangeOfString(target, selection.firstRange(), options);
+        auto referenceRange = selection.rawRange();
+        if (!referenceRange)
+            referenceRange = selection.firstRange();
+        resultRange = rangeOfString(target, referenceRange, options);
```
megangardner, line 4048: So maybe I'm not following, but this seems backwards? I would think you would want to default to the firstRange, but if that is collapsed, then use the raw range?

Source/WebCore/editing/Editor.cpp
```diff
@@ -4106,7 +4109,7 @@ std::optional<SimpleRange> Editor::rangeOfString(const String& target, const std
     // If we started in the reference range and the found range exactly matches the reference range, find again.
     // Build a selection with the found range to remove collapsed whitespace.
     // Compare ranges instead of selection objects to ignore the way that the current selection was made.
-    if (startInReferenceRange && VisibleSelection(resultRange).toNormalizedRange() == referenceRange) {
+    if (startInReferenceRange && VisibleSelection(resultRange).rawRange() == referenceRange) {
```
megangardner: I think we should only use the raw range if we have a collapsed Normalized range.

megangardner (approved):
Add a test, and I approve

Source/WebCore/editing/VisibleSelection.cpp
```diff
@@ -218,6 +218,14 @@ std::optional<SimpleRange> VisibleSelection::toNormalizedRange() const
     return makeSimpleRange(s, e);
 }
 
+std::optional<SimpleRange> VisibleSelection::rawRange() const
```
rniwa: How is this function different from `VisibleSelection::range()`?

Source/WebCore/editing/VisibleSelection.h
```diff
@@ -121,6 +121,8 @@ class VisibleSelection {
     // This function contracts the range around text and moves the caret upstream before returning the range.
     WEBCORE_EXPORT std::optional<SimpleRange> toNormalizedRange() const;
 
+    WEBCORE_EXPORT std::optional<SimpleRange> rawRange() const;
```
rniwa: Now this class has: `range()`, `firstRange()`, `toNormalizedRange()` and `rawRange()`, each of which returns a slightly different range. That's very confusing.

Source/WebCore/editing/Editor.cpp
```diff
@@ -4106,7 +4109,10 @@ std::optional<SimpleRange> Editor::rangeOfString(const String& target, const std
     // If we started in the reference range and the found range exactly matches the reference range, find again.
     // Build a selection with the found range to remove collapsed whitespace.
     // Compare ranges instead of selection objects to ignore the way that the current selection was made.
-    if (startInReferenceRange && VisibleSelection(resultRange).toNormalizedRange() == referenceRange) {
+    auto resultSelection = VisibleSelection(resultRange);
+    auto normalizedRange = resultSelection.toNormalizedRange();
```
rniwa, line 4113: Shouldn't this to be `firstRange()` instead although you're just preserving the existing behavior?
It doesn't make much sense that `Editor::findString` uses `firstRange` and `Editor::rangeOfString` uses `toNormalizedRange`.

rniwa (approved):

PR #54428, opened 2025-11-25
SVGAnimatedNumber properties should handle attribute removal and invalid values correctly

Ahmad-S792:
#### b99448c4aa6398dfad09cb896386e02ee1f49b70
<pre>
SVGAnimatedNumber properties should handle attribute removal and invalid values correctly
<a href="https://bugs.webkit.org/show_bug.cgi?id=303073">https://bugs.webkit.org/show_bug.cgi?id=303073</a>
<a href="https://rdar.apple.com/165378154">rdar://165378154</a>

Reviewed by NOBODY (OOPS!).

When SVG number attributes are removed or set to invalid values, they should
reset to their specified default values rather than 0. Previously, the code
used toFloat() which returns 0 for invalid input, causing properties with
non-zero defaults (e.g., slope=1, amplitude=1) to incorrectly become 0.

This patch switches to using parseNumber() which returns std::optional&lt;float&gt;,
allowing proper detection of parse failures and fallback to the correct default
values.

Additionally, for attributes parsed with parseNumberOptionalNumber() (like
baseFrequency, stdDeviation, kernelUnitLength), the code now explicitly handles
the parse failure case by resetting both X and Y components to their default values.

Affected elements and their defaults:
- SVGComponentTransferFunctionElement: slope/amplitude/exponent (default: 1),
  intercept/offset (default: 0)
- SVGFEDiffuseLightingElement: surfaceScale/diffuseConstant (default: 1),
  kernelUnitLengthX/Y (default: 0)
- SVGFEDropShadowElement: dx/dy/stdDeviationX/Y (default: 2)
- SVGFEGaussianBlurElement: stdDeviationX/Y (default: 0)
- SVGFEMorphologyElement: radiusX/Y (default: 0)
- SVGFETurbulenceElement: baseFrequencyX/Y (default: 0)
- SVGFESpecularLightingElement: surfaceScale/specularConstant/specularExponent (default: 1),
  kernelUnitLengthX/Y (default: 0)

* Source/WebCore/svg/SVGComponentTransferFunctionElement.cpp:
(WebCore::SVGComponentTransferFunctionElement::attributeChanged):
* Source/WebCore/svg/SVGFEDiffuseLightingElement.cpp:
(WebCore::SVGFEDiffuseLightingElement::attributeChanged):
* Source/WebCore/svg/SVGFEDropShadowElement.cpp:
(WebCore::SVGFEDropShadowElement::attributeChanged):
[...]

PR #58723, opened 2026-02-14
SVGAnimatedNumber properties should handle attribute removal and invalid values correctly

karlcow:
#### 1b0ef2d58624dacec83c2e360999a4f36df07b53
<pre>
SVGAnimatedNumber properties should handle attribute removal and invalid values correctly
<a href="https://bugs.webkit.org/show_bug.cgi?id=303073">https://bugs.webkit.org/show_bug.cgi?id=303073</a>
<a href="https://rdar.apple.com/165378154">rdar://165378154</a>

Reviewed by NOBODY (OOPS!).

Many SVG filter elements incorrectly fall back to 0 when attribute
parsing fails because attributeChanged() used toFloat() which silently
returns 0 for empty or invalid strings.

Replace toFloat() with parseNumber() which returns std::optional.
When parsing fails, reset the property to its spec-defined initial
value instead of 0.

Each element header now declares its initial values as static constexpr
constants, following the existing SVGFEConvolveMatrixElement pattern.

These could be consolidated into a shared SVGFilterDefaults.h in the
future.

* LayoutTests/imported/w3c/web-platform-tests/svg/types/scripted/SVGAnimatedNumber-initial-values-expected.txt:
* Source/WebCore/svg/SVGComponentTransferFunctionElement.cpp:
(WebCore::SVGComponentTransferFunctionElement::attributeChanged):
* Source/WebCore/svg/SVGComponentTransferFunctionElement.h:
* Source/WebCore/svg/SVGFECompositeElement.cpp:
(WebCore::SVGFECompositeElement::attributeChanged):
* Source/WebCore/svg/SVGFEConvolveMatrixElement.cpp:
(WebCore::SVGFEConvolveMatrixElement::attributeChanged):
* Source/WebCore/svg/SVGFEDiffuseLightingElement.cpp:
(WebCore::SVGFEDiffuseLightingElement::attributeChanged):
* Source/WebCore/svg/SVGFEDiffuseLightingElement.h:
* Source/WebCore/svg/SVGFEDisplacementMapElement.cpp:
(WebCore::SVGFEDisplacementMapElement::attributeChanged):
* Source/WebCore/svg/SVGFEDropShadowElement.cpp:
(WebCore::SVGFEDropShadowElement::attributeChanged):
* Source/WebCore/svg/SVGFEDropShadowElement.h:
* Source/WebCore/svg/SVGFEGaussianBlurElement.cpp:
(WebCore::SVGFEGaussianBlurElement::attributeChanged):
* Source/WebCore/svg/SVGFELightElement.cpp:
(WebCore::SVGFELightElement::attributeChanged):
* Source/WebCore/svg/SVGFELightElement.h:
* Source/WebCore/svg/SVGFEMorphologyElement.cpp:
(WebCore::SVGFEMorphologyElement::attributeChanged):
* Source/WebCore/svg/SVGFEOffsetElement.cpp:
(WebCore::SVGFEOffsetElement::attributeChanged):
* Source/WebCore/svg/SVGFESpecularLightingElement.cpp:
(WebCore::SVGFESpecularLightingElement::attributeChanged):
* Source/WebCore/svg/SVGFESpecularLightingElement.h:
* Source/WebCore/svg/SVGFETurbulenceElement.cpp:
(WebCore::SVGFETurbulenceElement::attributeChanged):
* Source/WebCore/svg/SVGGeometryElement.cpp:
(WebCore::SVGGeometryElement::attributeChanged):
* Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h:
(WebCore::SVGAnimatedPrimitiveProperty::setBaseValInternal):
(WebCore::SVGAnimatedPrimitiveProperty::SVGAnimatedPrimitiveProperty):
</pre>
https://github.com/WebKit/WebKit/commit/1b0ef2d58624dacec83c2e360999a4f36df07b53

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/155138 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/28398 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/21557 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/163898 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/108691 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/157011 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/28538 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/28246 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/120029 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/84800 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/158097 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/22281 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/139308 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/100722 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/21366 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/19421 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/11724 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/131028 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/17150 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/166376 "Built successfully") | | 
| | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/10403 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/18759 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/128132 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/27942 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/23459 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/128270 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/27866 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/138943 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/84575 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/23132 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/15738 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/27559 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/91662 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/27137 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/27367 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/27210 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/svg/SVGComponentTransferFunctionElement.cpp
+++ b/Source/WebCore/svg/SVGComponentTransferFunctionElement.cpp
@@ -26,6 +26,7 @@
 #include "SVGComponentTransferFunctionElementInlines.h"
 #include "SVGFEComponentTransferElement.h"
 #include "SVGNames.h"
+#include "SVGParserUtilities.h"
 #include "SVGPropertyOwnerRegistry.h"
 #include <wtf/NeverDestroyed.h>
 #include <wtf/TZoneMallocInlines.h>
@@ -63,19 +64,19 @@ void SVGComponentTransferFunctionElement::attributeChanged(const QualifiedName& 
         m_tableValues->baseVal()->parse(newValue);
         break;
     case AttributeNames::slopeAttr:
-        m_slope->setBaseValInternal(newValue.toFloat());
+        m_slope->setBaseValInternal(parseNumber(newValue), initialSlope);
         break;
     case AttributeNames::interceptAttr:
-        m_intercept->setBaseValInternal(newValue.toFloat());
+        m_intercept->setBaseValInternal(parseNumber(newValue));
         break;
     case AttributeNames::amplitudeAttr:
-        m_amplitude->setBaseValInternal(newValue.toFloat());
+        m_amplitude->setBaseValInternal(parseNumber(newValue), initialAmplitude);
         break;
     case AttributeNames::exponentAttr:
-        m_exponent->setBaseValInternal(newValue.toFloat());
+        m_exponent->setBaseValInternal(parseNumber(newValue), initialExponent);
         break;
     case AttributeNames::offsetAttr:
-        m_offset->setBaseValInternal(newValue.toFloat());
+        m_offset->setBaseValInternal(parseNumber(newValue));
         break;
     default:
         break;
--- a/Source/WebCore/svg/SVGComponentTransferFunctionElement.h
+++ b/Source/WebCore/svg/SVGComponentTransferFunctionElement.h
@@ -103,12 +103,17 @@ protected:
     bool rendererIsNeeded(const RenderStyle&) override { return false; }
     
 private:
+    // FIXME: These initial values should be consolidated in a shared SVGFilterDefaults.h.
+    static constexpr float initialSlope = 1;
+    static constexpr float initialAmplitude = 1;
+    static constexpr float initialExponent = 1;
+
     const Ref<SVGAnimatedEnumeration> m_type { SVGAnimatedEnumeration::create(this, ComponentTransferType::FECOMPONENTTRANSFER_TYPE_IDENTITY) };
     const Ref<SVGAnimatedNumberList> m_tableValues { SVGAnimatedNumberList::create(this) };
-    const Ref<SVGAnimatedNumber> m_slope { SVGAnimatedNumber::create(this, 1) };
+    const Ref<SVGAnimatedNumber> m_slope { SVGAnimatedNumber::create(this, initialSlope) };
     const Ref<SVGAnimatedNumber> m_intercept { SVGAnimatedNumber::create(this) };
-    const Ref<SVGAnimatedNumber> m_amplitude { SVGAnimatedNumber::create(this, 1) };
-    const Ref<SVGAnimatedNumber> m_exponent { SVGAnimatedNumber::create(this, 1) };
+    const Ref<SVGAnimatedNumber> m_amplitude { SVGAnimatedNumber::create(this, initialAmplitude) };
+    const Ref<SVGAnimatedNumber> m_exponent { SVGAnimatedNumber::create(this, initialExponent) };
     const Ref<SVGAnimatedNumber> m_offset { SVGAnimatedNumber::create(this) };
 };
 
--- a/Source/WebCore/svg/SVGFECompositeElement.cpp
+++ b/Source/WebCore/svg/SVGFECompositeElement.cpp
@@ -25,6 +25,7 @@
 #include "FEComposite.h"
 #include "NodeName.h"
 #include "SVGNames.h"
+#include "SVGParserUtilities.h"
 #include "SVGPropertyOwnerRegistry.h"
 #include <wtf/TZoneMallocInlines.h>
 
@@ -71,16 +72,16 @@ void SVGFECompositeElement::attributeChanged(const QualifiedName& name, const At
         Ref { m_in2 }->setBaseValInternal(newValue);
         break;
     case AttributeNames::k1Attr:
-        m_k1->setBaseValInternal(newValue.toFloat());
+        m_k1->setBaseValInternal(parseNumber(newValue));
         break;
     case AttributeNames::k2Attr:
-        m_k2->setBaseValInternal(newValue.toFloat());
+        m_k2->setBaseValInternal(parseNumber(newValue));
         break;
     case AttributeNames::k3Attr:
-        m_k3->setBaseValInternal(newValue.toFloat());
+        m_k3->setBaseValInternal(parseNumber(newValue));
         break;
     case AttributeNames::k4Attr:
-        m_k4->setBaseValInternal(newValue.toFloat());
+        m_k4->setBaseValInternal(parseNumber(newValue));
         break;
     default:
         break;
--- a/Source/WebCore/svg/SVGFEMorphologyElement.cpp
+++ b/Source/WebCore/svg/SVGFEMorphologyElement.cpp
@@ -69,6 +69,9 @@ void SVGFEMorphologyElement::attributeChanged(const QualifiedName& name, const A
         if (auto result = parseNumberOptionalNumber(newValue)) {
             m_radiusX->setBaseValInternal(result->first);
             m_radiusY->setBaseValInternal(result->second);
+        } else {
+            m_radiusX->setBaseValInternal(std::nullopt);
+            m_radiusY->setBaseValInternal(std::nullopt);
         }
         break;
     default:
--- a/Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
+++ b/Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
@@ -28,6 +28,7 @@
 #include "ExceptionOr.h"
 #include "SVGAnimatedProperty.h"
 #include "SVGSharedPrimitiveProperty.h"
+#include <optional>
 
 namespace WebCore {
 
@@ -57,6 +58,7 @@ public:
 
     // Used by SVGElement::parseAttribute().
     void setBaseValInternal(const PropertyType& baseVal) { m_baseVal->setValue(baseVal); }
+    void setBaseValInternal(std::optional<PropertyType> baseVal, const PropertyType& defaultValue = PropertyType { }) requires std::is_arithmetic_v<PropertyType> { m_baseVal->setValue(baseVal.value_or(defaultValue)); }
     const PropertyType& baseVal() const LIFETIME_BOUND { return m_baseVal->value(); }
 
     // Used by SVGAttributeAnimator::progress.
--- a/LayoutTests/imported/w3c/web-platform-tests/svg/types/scripted/SVGAnimatedNumber-initial-values-expected.txt
+++ b/LayoutTests/imported/w3c/web-platform-tests/svg/types/scripted/SVGAnimatedNumber-initial-values-expected.txt
@@ -1,12 +1,12 @@
 
-FAIL SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.slope (remove) assert_equals: initial after expected 1 but got 0
-FAIL SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.slope (invalid value) assert_equals: initial after expected 1 but got 0
+PASS SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.slope (remove)
+PASS SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.slope (invalid value)
 PASS SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.intercept (remove)
 PASS SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.intercept (invalid value)
-FAIL SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.amplitude (remove) assert_equals: initial after expected 1 but got 0
-FAIL SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.amplitude (invalid value) assert_equals: initial after expected 1 but got 0
-FAIL SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.exponent (remove) assert_equals: initial after expected 1 but got 0
-FAIL SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.exponent (invalid value) assert_equals: initial after expected 1 but got 0
+PASS SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.amplitude (remove)
+PASS SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.amplitude (invalid value)
+PASS SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.exponent (remove)
+PASS SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.exponent (invalid value)
 PASS SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.offset (remove)
 PASS SVGAnimatedNumber, initial values, SVGComponentTransferFunctionElement.prototype.offset (invalid value)
 PASS SVGAnimatedNumber, initial values, SVGFECompositeElement.prototype.k1 (remove)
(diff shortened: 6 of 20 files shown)


Source/WebCore/svg/SVGFEMorphologyElement.cpp
```diff
@@ -66,10 +66,7 @@ void SVGFEMorphologyElement::attributeChanged(const QualifiedName& name, const A
         Ref { m_in1 }->setBaseValInternal(newValue);
         break;
     case AttributeNames::radiusAttr:
-        if (auto result = parseNumberOptionalNumber(newValue)) {
-            Ref { m_radiusX }->setBaseValInternal(result->first);
-            Ref { m_radiusY }->setBaseValInternal(result->second);
-        }
+        SVGAnimatedPrimitiveProperty<float>::parseAndSetAttributePair(m_radiusX.get(), m_radiusY.get(), newValue);
```
Ahmad-S792: If we add `SVGAnimatedPrimitiveProperty` in the header, can we just call `parseAndSetAttributePair` directly?

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
```diff
@@ -134,16 +135,34 @@ class SVGAnimatedPrimitiveProperty : public SVGAnimatedProperty<SVGAnimatedPrimi
...
     }
 
+    // it will return the initial specified value at creation, but
+    // it will also return the value when parsing fails, example
+    // for invalid values
+    const PropertyType& initialValue() const { return m_initialValue; }
+
+    // Parses attribute value and sets baseVal, using initialValue as fallback on parse failure.
+    // Template specializations are provided for specific types (float, int, etc.).
+    void parseAndSetAttribute(const AtomString& value);
```
Ahmad-S792: You can drop arguments from header definition - `void parseAndSetAttribute(const AtomString&);`

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
```diff
@@ -134,16 +135,34 @@ class SVGAnimatedPrimitiveProperty : public SVGAnimatedProperty<SVGAnimatedPrimi
...
+
+    // Parses paired number attribute (e.g., "2 3" or "5").
+    // Single value sets both X and Y to same value.
+    // Parse failure resets both to their initial values.
+    // Used for: stdDeviationX/Y, kernelUnitLengthX/Y, radiusX/Y, etc.
+    static void parseAndSetAttributePair(SVGAnimatedPrimitiveProperty& propertyX,
```
Ahmad-S792: nit: Single line and also to fit, you can drop `arguments` from header definition.

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
```diff
@@ -156,6 +175,37 @@ class SVGAnimatedPrimitiveProperty : public SVGAnimatedProperty<SVGAnimatedPrimi
...
+    setBaseValInternal(parseNumber(value).value_or(m_initialValue));
+}
+
+// Parses paired number attribute (e.g., "2 3" or "5").
+// Single value sets both X and Y to same value.
+// Pair value sets X and Y individually.
+// Parse failure resets both to their initial values per SVG spec.
+template<>
+inline void SVGAnimatedPrimitiveProperty<float>::parseAndSetAttributePair(
+    SVGAnimatedPrimitiveProperty<float>& propertyX,
```
Ahmad-S792: nit: single line.

karlcow:
> Whether a property is animatable has nothing to do with its default value. Does this also create an awkward split where defaults now live in two places? I notice that some defaults, like `initialOrderValue` or `initialDivisorValue`, live in element headers. Effectively, this would create two sources of truth.

1. About `initialOrderValue` this is for the future PR about `SVGAnimatedInteger` as mentioned in the commit message. Or would you prefer to have everything done in one shot? This would be https://wpt.fyi/results/svg/types/scripted/SVGAnimatedInteger-initial-values.html to double check we keep the tests valid.
2. `initialDivisorValue`, this one has been forgottent indeed. 

About:

> I haven't fully formalized my thoughts on this yet, but there's a core problem here. The issue I see is that SVGAnimatedPrimitiveProperty is now being asked to handle two orthogonal concerns: animation (managing baseVal/animVal) and spec-defined initial values.

Yes. I see your point. Maybe not the best place for addressing this. I'm checking in the code if there is a better place to handle these initial values. 

Maybe 

* `SVGPropertyTraits` with specialization for each elements. So everything in one place.
* we put them in each of the element `.h` file. Or local to each element. 

I see pros and cons for both for maintenance.

karlcow:
but it would muddy the water for SVGPropertyTraits, make it too specific, not generic enough.

karlcow:
hmmm maybe we can adopt the same pattern than CSS has established with a centralized definition of properties. 
There are a bit more than 180 properties in SVG. 

something like: 

  1. SVGProperties.json 
  2. process-svg-properties.py like CSS a generator script and tools associated.
  3. SVGPropertyInitialValues.h
  
  but that would take a bit more time, a lot of rebasing each time someone edit something probably. I wonder what @nikolaszimmermann and @shallawa would think about this approach.

Or as I was saying we can move back the initial value close to each element and still solve the initial issue about duplication to solve the points made in https://github.com/WebKit/WebKit/pull/54428#discussion_r2559333426

karlcow:
Let's KISS (Keep It Simple Stupid).
I'll move back the initial values to the elements, and still keep the parse and set that fixes the initial issue. 
Then we can probably think about a way to handle these in the future?

karlcow:
let the bots run, then I will fix cosmetic comments.

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
karlcow: ok done. Thanks!

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
karlcow: ok done. 
(I find that less readable personally)
Thanks.

Ahmad-S792: I know but this is what we use throughout WebKit - at least what I have looked into. So just to stick with same standard.

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
```diff
@@ -134,6 +135,10 @@ class SVGAnimatedPrimitiveProperty : public SVGAnimatedProperty<SVGAnimatedPrimi
             m_animVal = nullptr;
     }
 
+    void parseAndSetAttribute(const AtomString&, const PropertyType&);
+
+    static void parseAndSetAttributePair(SVGAnimatedPrimitiveProperty&, SVGAnimatedPrimitiveProperty&, const AtomString&, const PropertyType&, const PropertyType&);
```
shallawa: These parsing functions should not be here. I suggest changing `setBaseValInternal()` to take `std::optional<PropertyType>`:

```
void setBaseValInternal(const std::optional<PropertyType>& baseVal) 
{ 
    m_baseVal->setValue(baseVal.value_or(m_initialValue));
}
```

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
```diff
@@ -156,6 +161,26 @@ class SVGAnimatedPrimitiveProperty : public SVGAnimatedProperty<SVGAnimatedPrimi
 
     const Ref<SVGSharedPrimitiveProperty<PropertyType>> m_baseVal;
     mutable RefPtr<SVGSharedPrimitiveProperty<PropertyType>> m_animVal;
+    PropertyType m_initialValue;
```
shallawa: Where is this member is initialized or used? I think this should be initialized by the constructor:
```
    SVGAnimatedPrimitiveProperty(SVGElement* contextElement)
        : Base(contextElement)
        , m_baseVal(SVGSharedPrimitiveProperty<PropertyType>::create())
        , m_initialValue(m_baseVal->value())
    {
    }

    SVGAnimatedPrimitiveProperty(SVGElement* contextElement, const PropertyType& value)
        : Base(contextElement)
        , m_baseVal(SVGSharedPrimitiveProperty<PropertyType>::create(value))
        , m_initialValue(value)
    {
    }
```

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
```diff
@@ -156,6 +161,26 @@ class SVGAnimatedPrimitiveProperty : public SVGAnimatedProperty<SVGAnimatedPrimi
...
+{
+    auto result = parseNumberOptionalNumber(value);
+    if (result) {
+        propertyX.setBaseValInternal(result->first);
+        propertyY.setBaseValInternal(result->second);
+    } else {
+        propertyX.setBaseValInternal(fallbackX);
+        propertyY.setBaseValInternal(fallbackY);
+    }
+}
```
shallawa: I do not think we need to do this template function specialization here

Source/WebCore/svg/SVGComponentTransferFunctionElement.cpp
```diff
@@ -37,6 +37,12 @@ WTF_MAKE_TZONE_ALLOCATED_IMPL(SVGComponentTransferFunctionElement);
 SVGComponentTransferFunctionElement::SVGComponentTransferFunctionElement(const QualifiedName& tagName, Document& document)
     : SVGElement(tagName, document, makeUniqueRef<PropertyRegistry>(*this))
 {
+    m_slope->setBaseValInternal(initialSlope());
+    m_intercept->setBaseValInternal(initialIntercept());
+    m_amplitude->setBaseValInternal(initialAmplitude());
+    m_exponent->setBaseValInternal(initialExponent());
+    m_offset->setBaseValInternal(initialOffset());
```
shallawa: Why are these calls needed even the members are initialized with these values when the class is created?

Source/WebCore/svg/SVGComponentTransferFunctionElement.cpp
```diff
@@ -63,19 +69,19 @@ void SVGComponentTransferFunctionElement::attributeChanged(const QualifiedName&
         m_tableValues->baseVal()->parse(newValue);
         break;
     case AttributeNames::slopeAttr:
-        m_slope->setBaseValInternal(newValue.toFloat());
+        m_slope->parseAndSetAttribute(newValue, initialSlope());
```
shallawa: I think this should be 
```
m_slope->setBaseValInternal(parseNumber(newValue));
```

Source/WebCore/svg/SVGComponentTransferFunctionElement.h
```diff
@@ -105,10 +111,10 @@ class SVGComponentTransferFunctionElement : public SVGElement {
 private:
     Ref<SVGAnimatedEnumeration> m_type { SVGAnimatedEnumeration::create(this, ComponentTransferType::FECOMPONENTTRANSFER_TYPE_IDENTITY) };
     Ref<SVGAnimatedNumberList> m_tableValues { SVGAnimatedNumberList::create(this) };
-    Ref<SVGAnimatedNumber> m_slope { SVGAnimatedNumber::create(this, 1) };
+    Ref<SVGAnimatedNumber> m_slope { SVGAnimatedNumber::create(this) };
```
shallawa: Why are the initial values are removed and put in the constructor? I think this should be:
```
Ref<SVGAnimatedNumber> m_slope { SVGAnimatedNumber::create(this, initialSlope()) };
```

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
shallawa: I take this back. This will increase the size of all the `SVGAnimatedPrimitiveProperty ` with the initial value size which should be the same for every property.

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
karlcow: @shallawa where would you do it? :) I'm happy to move it where it makes most sense, but that was to avoid the issues raised on https://github.com/WebKit/WebKit/pull/54428 for @Ahmad-S792 PR.

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
karlcow: Where do we move them?

Source/WebCore/svg/SVGComponentTransferFunctionElement.cpp
nikolaszimmermann: I agree, this shouldn't be needed here -- ideally it should stand beside the declaration of the property -- to not have information in two places for the same property.

Source/WebCore/svg/SVGComponentTransferFunctionElement.h
nikolaszimmermann: I fully agree.

Source/WebCore/svg/SVGFECompositeElement.cpp
```diff
@@ -71,16 +76,16 @@ void SVGFECompositeElement::attributeChanged(const QualifiedName& name, const At
         Ref { m_in2 }->setBaseValInternal(newValue);
         break;
     case AttributeNames::k1Attr:
-        Ref { m_k1 }->setBaseValInternal(newValue.toFloat());
+        Ref { m_k1 }->parseAndSetAttribute(newValue, initialK1());
         break;
     case AttributeNames::k2Attr:
-        Ref { m_k2 }->setBaseValInternal(newValue.toFloat());
+        Ref { m_k2 }->parseAndSetAttribute(newValue, initialK2());
```
nikolaszimmermann: m_k2 should somehow know it's internal value, so that we don't need to pass it here again -- this is error-prone to repeat it.

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
karlcow: modified.

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
karlcow: modified.

Source/WebCore/svg/SVGComponentTransferFunctionElement.cpp
karlcow: modified

Source/WebCore/svg/SVGComponentTransferFunctionElement.h
karlcow: modified.

Source/WebCore/svg/SVGFECompositeElement.cpp
karlcow: modified.

Source/WebCore/svg/SVGComponentTransferFunctionElement.cpp
karlcow: modified

Source/WebCore/svg/properties/SVGAnimatedPrimitiveProperty.h
```diff
@@ -156,6 +160,7 @@ class SVGAnimatedPrimitiveProperty : public SVGAnimatedProperty<SVGAnimatedPrimi
 
     const Ref<SVGSharedPrimitiveProperty<PropertyType>> m_baseVal;
     mutable RefPtr<SVGSharedPrimitiveProperty<PropertyType>> m_animVal;
+    PropertyType m_initialValue;
```
shallawa: This will increase the size of all the `SVGAnimatedPrimitiveProperty`  with the m_initialValue size. The initial value for a property should be the same for all instances.

karlcow: ha. You mean each time we create an instance of a filter we store an object for the initialValue and this more costly than a simple integer. It makes sense. 

So let's say we have 50 `<feDiffuseLighting>`. Each one of these creates 4 SVGAnimatedNumber properties (diffuseConstant, surfaceScale, kernelUnitLengthX, kernelUnitLengthY). A float is 4 bytes?

50 elements × 4 properties × 4 bytes = 800 bytes.

while with a 1 or 0, there is no cost at all. 

but if I set them individually on the properties. That would be a bad architecture choice and another place for bugs. Do we need a place which defines all the values in one place? 

Looking at https://drafts.csswg.org/filter-effects-1/

* SVGFEDiffuseLightingElement: diffuseConstant=1, surfaceScale=1
* SVGFESpecularLightingElement: specularConstant=1, specularExponent=1, surfaceScale=1
* SVGFELightElement: specularExponent=1
* SVGFEDropShadowElement: dx=2, dy=2, stdDeviationX=2, stdDeviationY=2
* SVGComponentTransferFunctionElement: slope=1, amplitude=1, exponent=1

karlcow: hmm… SVGFEConvolveMatrixElement has this, in the individual class. 

https://searchfox.org/wubkat/rev/9971986dcebbc9ad5713a058e7951288c455bbee/Source/WebCore/svg/SVGFEConvolveMatrixElement.h#102-106
```cpp
SVGFEConvolveMatrixElement(const QualifiedName&, Document&);

static constexpr int initialOrderValue = 3;
static constexpr float initialDivisorValue = 1;
static constexpr float initialKernelUnitLengthValue = 0;
```

hmmm but then we will repeat code for each class. What about putting them in a central place, in a new file `SVGFilterDefaults.h`

or can I use https://searchfox.org/wubkat/source/Source/WebCore/svg/SVGFilterPrimitiveStandardAttributes.h to store all the values?
ah no! `SVGComponentTransferFunctionElement` and `SVGFELightElement` don't inherit from `SVGFilterPrimitiveStandardAttributes`.

karlcow: With a new `SVGFilterDefaults.h`, we coul do:

```cpp
#pragma once

namespace WebCore::SVGFilterDefaults {

// https://drafts.csswg.org/filter-effects-1/
// Lighting filters (feDiffuseLighting, feSpecularLighting)
constexpr float surfaceScale = 1;
constexpr float diffuseConstant = 1;
constexpr float specularConstant = 1;
constexpr float specularExponent = 1;

// feDropShadow
constexpr float dropShadowDx = 2;
constexpr float dropShadowDy = 2;
constexpr float dropShadowStdDeviation = 2;

// feComponentTransfer (feFuncR, feFuncG, feFuncB, feFuncA)
constexpr float slope = 1;
constexpr float amplitude = 1;
constexpr float exponent = 1;

// feConvolveMatrix
constexpr int orderXY = 3;
constexpr float divisor = 1;

} // namespace WebCore::SVGFilterDefaults
  ```

karlcow: The centralization would avoid the duplication of 

* surfaceScale = 1 -> SVGFEDiffuseLightingElement and SVGFESpecularLightingElement                                                                                                      
* specularExponent = 1 -> SVGFESpecularLightingElement and SVGFELightElement

karlcow:
Closing this PR. 
Too many comments, too many changes. For anyone, feel free to take over with a new or multiple PRs.

PR #58748, opened 2026-02-15
Make Document's m_eventListenerCounts more sophisticated

annevk:
#### d2709687540ddca50d71751f2c2f85eb4bb594f7
<pre>
Make Document&apos;s m_eventListenerCounts more sophisticated
<a href="https://bugs.webkit.org/show_bug.cgi?id=307935">https://bugs.webkit.org/show_bug.cgi?id=307935</a>
<a href="https://rdar.apple.com/170910523">rdar://170910523</a>

Reviewed by Ryosuke Niwa.

Canonical link: <a href="https://commits.webkit.org/308284@main">https://commits.webkit.org/308284@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/63565f7ece69d20be7e7ca3a44f7e0643cb0dec2

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/146963 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/19643 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/13233 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/155645 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/100377 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/148837 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/20102 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/19544 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/113243 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/100377 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/149925 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/15489 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/132046 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94000 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/14701 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/12480 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/3087 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/124286 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/9959 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/157976 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/1107 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/11398 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/121267 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/19445 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/16325 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/121469 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/19454 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/131715 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/75413 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/22677 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/17036 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/8547 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/19060 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/82815 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/18790 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/18941 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/18849 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/dom/Document.cpp
+++ b/Source/WebCore/dom/Document.cpp
@@ -7228,21 +7228,25 @@ void Document::addListenerTypeIfNeeded(const AtomString& eventType)
     }
 }
 
-void Document::didAddEventListenersOfType(const AtomString& eventType, unsigned count)
+void Document::didAddEventListenersOfType(const AtomString& eventType, IsCapture isCapture, uint16_t count)
 {
     ASSERT(count);
     addListenerTypeIfNeeded(eventType);
-    auto result = m_eventListenerCounts.fastAdd(eventType, 0);
-    result.iterator->value += count;
+    auto& counts = m_eventListenerCounts.ensure(eventType, [] { return EventListenerCounts { }; }).iterator->value;
+    auto& field = isCapture == IsCapture::Yes ? counts.capturing : counts.bubbling;
+    field = std::min<uint32_t>(field + count, std::numeric_limits<uint16_t>::max());
 }
 
-void Document::didRemoveEventListenersOfType(const AtomString& eventType, unsigned count)
+void Document::didRemoveEventListenersOfType(const AtomString& eventType, IsCapture isCapture, uint16_t count)
 {
     ASSERT(count);
     ASSERT(m_eventListenerCounts.contains(eventType));
     auto it = m_eventListenerCounts.find(eventType);
-    ASSERT(it->value >= count);
-    it->value -= count;
+    auto& field = isCapture == IsCapture::Yes ? it->value.capturing : it->value.bubbling;
+    if (field != std::numeric_limits<uint16_t>::max()) [[likely]] {
+        ASSERT(field >= count);
+        field -= count;
+    }
 }
 
 HTMLFrameOwnerElement* Document::ownerElement() const
@@ -10066,7 +10070,7 @@ void Document::didAssociateFormControl(Element& element)
     if (!page)
         return;
     if (!page->chrome().client().shouldNotifyOnFormChanges()
-        && !hasEventListenersOfType(eventNames().webkitassociateformcontrolsEvent))
+        && !eventListenerCountsOfType(eventNames().webkitassociateformcontrolsEvent).hasAny())
         return;
 
     auto isNewEntry = m_associatedFormControls.add(element).isNewEntry;
--- a/Source/WebCore/dom/Document.h
+++ b/Source/WebCore/dom/Document.h
@@ -1117,10 +1117,20 @@ public:
     bool NODELETE hasListenerTypeForEventType(PlatformEventType) const;
     void addListenerTypeIfNeeded(const AtomString& eventType);
 
-    void didAddEventListenersOfType(const AtomString&, unsigned = 1);
-    void NODELETE didRemoveEventListenersOfType(const AtomString&, unsigned = 1);
+    enum class IsCapture : bool { No, Yes };
+
+    struct EventListenerCounts {
+        uint16_t capturing { 0 };
+        uint16_t bubbling { 0 };
+
+        bool hasAny() const { return capturing || bubbling; }
+        bool hasCapturing() const { return capturing; }
+    };
+
+    void didAddEventListenersOfType(const AtomString&, IsCapture, uint16_t count = 1);
+    void NODELETE didRemoveEventListenersOfType(const AtomString&, IsCapture, uint16_t count = 1);
     bool hasNodeWithEventListeners() const { return !m_eventListenerCounts.isEmpty(); }
-    bool hasEventListenersOfType(const AtomString& type) const { return m_eventListenerCounts.inlineGet(type); }
+    EventListenerCounts eventListenerCountsOfType(const AtomString& type) const { return m_eventListenerCounts.inlineGet(type); }
 
     bool hasConnectedPluginElements() { return m_connectedPluginElementCount; }
     void didConnectPluginElement() { ++m_connectedPluginElementCount; }
@@ -2654,7 +2664,7 @@ private:
     unsigned m_dataListElementCount { 0 };
 
     OptionSet<ListenerType> m_listenerTypes;
-    MemoryCompactRobinHoodHashMap<AtomString, unsigned> m_eventListenerCounts;
+    MemoryCompactRobinHoodHashMap<AtomString, EventListenerCounts> m_eventListenerCounts;
     unsigned m_connectedPluginElementCount { 0 };
 
     unsigned m_referencingNodeCount { 0 };
--- a/Source/WebCore/dom/EventDispatcher.cpp
+++ b/Source/WebCore/dom/EventDispatcher.cpp
@@ -88,21 +88,24 @@ static bool NODELETE isInShadowTree(EventTarget* target)
     return node && node->isInShadowTree();
 }
 
-static void dispatchEventInDOM(Event& event, const EventPath& path)
+static void dispatchEventInDOM(Event& event, const EventPath& path, const Document::EventListenerCounts& listenerCounts)
 {
-    // Invoke capturing event listeners in the reverse order.
-    for (size_t i = path.size(); i > 0; --i) {
-        const EventContext& eventContext = path.contextAt(i - 1);
-        if (eventContext.currentTarget() == eventContext.target())
-            event.setEventPhase(Event::AT_TARGET);
-        else
-            event.setEventPhase(Event::CAPTURING_PHASE);
-        eventContext.handleLocalEvents(event, EventTarget::EventInvokePhase::Capturing);
-        if (event.propagationStopped())
-            return;
+    if (listenerCounts.hasCapturing()) {
+        // Invoke capturing event listeners in the reverse order.
+        for (size_t i = path.size(); i > 0; --i) {
+            const EventContext& eventContext = path.contextAt(i - 1);
+            if (eventContext.currentTarget() == eventContext.target())
+                event.setEventPhase(Event::AT_TARGET);
+            else
+                event.setEventPhase(Event::CAPTURING_PHASE);
+            eventContext.handleLocalEvents(event, EventTarget::EventInvokePhase::Capturing);
+            if (event.propagationStopped())
+                return;
+        }
     }
 
-    // Invoke bubbling event listeners.
+    // Invoke bubbling event listeners. (We don't use listenerCounts here as handleLocalEvents()
+    // needs to update event and it's highly unlikely there's no bubble listeners.)
     size_t size = path.size();
     for (size_t i = 0; i < size; ++i) {
         const EventContext& eventContext = path.contextAt(i);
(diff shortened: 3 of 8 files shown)


Source/WebCore/dom/Document.h
```diff
@@ -1116,10 +1116,23 @@ class Document
     bool hasListenerTypeForEventType(PlatformEventType) const;
     void addListenerTypeIfNeeded(const AtomString& eventType);
 
-    void didAddEventListenersOfType(const AtomString&, unsigned = 1);
-    void didRemoveEventListenersOfType(const AtomString&, unsigned = 1);
+    struct EventListenerCounts {
+        uint16_t capturing { 0 };
```
rniwa, line 1123: Hm... with `uint16_t`, there is a chance of overflow... we should probably handle overflow gracefully.

Here's one scheme. Once we've reached UINT_MAX, we can stop incrementing any further and we'd never decrement such a count (since we don't know how many we've ignored).

Source/WebCore/dom/Document.h
```diff
@@ -1116,10 +1116,23 @@ class Document
...
+        uint16_t bubbling { 0 };
+
+        unsigned total() const { return capturing + bubbling; }
+        bool hasAny() const { return capturing || bubbling; }
+        bool hasCapturing() const { return capturing > 0; }
```
rniwa: Since the counter is uint16_t, just test `capturing`.

Source/WebCore/dom/EventDispatcher.cpp
```diff
@@ -88,21 +88,51 @@ static bool isInShadowTree(EventTarget* target)
     return node && node->isInShadowTree();
 }
 
-static void dispatchEventInDOM(Event& event, const EventPath& path)
+struct EventListenerInfo {
```
rniwa: Do we really need this separate struct? Can't we just use the one in Document?

Source/WebCore/dom/Document.h
```diff
@@ -1116,10 +1116,23 @@ class Document
...
+        bool hasBubbling() const { return bubbling > 0; }
+    };
+
+    void didAddEventListenersOfType(const AtomString&, bool useCapture, uint16_t count = 1);
```
rniwa: Please use enum class for the second argument.

rniwa (approved):

Source/WebCore/dom/Document.cpp
```diff
@@ -7228,21 +7228,26 @@ void Document::addListenerTypeIfNeeded(const AtomString& eventType)
...
+void Document::didAddEventListenersOfType(const AtomString& eventType, IsCapture isCapture, uint16_t count)
 {
     ASSERT(count);
     addListenerTypeIfNeeded(eventType);
-    auto result = m_eventListenerCounts.fastAdd(eventType, 0);
-    result.iterator->value += count;
+    auto& counts = m_eventListenerCounts.ensure(eventType, [] { return EventListenerCounts { }; }).iterator->value;
+    auto& field = isCapture == IsCapture::Yes ? counts.capturing : counts.bubbling;
+    if (field != std::numeric_limits<uint16_t>::max()) [[likely]]
+        field = std::min<uint32_t>(field + count, std::numeric_limits<uint16_t>::max());
```
Andrew-Fryer: I think this can overflow and wrap around before std::min extends to `uint32_t` because `field` and `count` are both `uint16_t`.

Andrew-Fryer: So, we need to cast before the addition.

Andrew-Fryer: However, I personally think it is cleanest to remove the check:
`if (field != std::numeric_limits<uint16_t>::max()) [[likely]]`
...just do:
`field = static_cast<uint16_t>(std::min<uint32_t>(static_cast<uint32_t>(field) + count, std::numeric_limits<uint16_t>::max()));`
I think clang will see through this and have good perf.

Source/WebCore/page/LocalDOMWindow.cpp
```diff
@@ -2289,7 +2296,7 @@ bool LocalDOMWindow::removeEventListener(const AtomString& eventType, EventListe
     auto& eventNames = WebCore::eventNames();
     auto typeInfo = eventNames.typeInfoForEvent(eventType);
     if (document) {
-        document->didRemoveEventListenersOfType(eventType);
+        document->didRemoveEventListenersOfType(eventType, options.capture ? Document::IsCapture::Yes : Document::IsCapture::No);
```
Andrew-Fryer, line 2299: I think we should merge `EventListenerOptions` and `Document::IsCapture` into the same enum, but I think it makes sense to do this in a follow up PR.

Source/WebCore/dom/Node.cpp
```diff
@@ -2512,16 +2519,18 @@ bool Node::removeEventListener(const AtomString& eventType, EventListener& liste
...
 {
     EventTarget::removeAllEventListeners();
-    enumerateEventListenerTypes([&](const AtomString& type, unsigned count) {
-        for (unsigned i = 0; i < count; ++i)
-            didRemoveEventListenerOfType(*this, type);
+    enumerateEventListenerTypes([&](const AtomString& type, uint16_t capturingCount, uint16_t bubblingCount) {
+        for (uint16_t i = 0; i < capturingCount; ++i)
+            didRemoveEventListenerOfType(*this, type, Document::IsCapture::Yes);
+        for (uint16_t i = 0; i < bubblingCount; ++i)
+            didRemoveEventListenerOfType(*this, type, Document::IsCapture::No);
```
Andrew-Fryer, line 2533: This doesn't look hot on traces, but perhaps we should fix this anyway.
(I think `didRemoveEventListenerOfType` should take a count that defaults to 1.)
I think we should do this in a follow up PR.

Source/WebCore/dom/EventListenerMap.h
```diff
@@ -83,8 +84,20 @@ class EventListenerMap {
...
     void enumerateEventListenerTypes(NOESCAPE const CallbackType& callback) const
     {
-        for (auto& entry : m_entries)
-            callback(entry.first, entry.second.size());
+        for (auto& entry : m_entries) {
+            uint16_t capturingCount = 0;
+            uint16_t bubblingCount = 0;
+            for (auto& listener : entry.second) {
+                if (listener->useCapture()) {
+                    if (capturingCount != std::numeric_limits<uint16_t>::max()) [[likely]]
```
Andrew-Fryer: I think we should accumulate in a `uint32_t` and then only check for overflow when we cast as we pass into `callback`.

Source/WebCore/dom/EventListenerMap.h
```diff
@@ -83,8 +84,20 @@ class EventListenerMap {
     template<typename CallbackType>
     void enumerateEventListenerTypes(NOESCAPE const CallbackType& callback) const
     {
-        for (auto& entry : m_entries)
-            callback(entry.first, entry.second.size());
+        for (auto& entry : m_entries) {
+            uint16_t capturingCount = 0;
+            uint16_t bubblingCount = 0;
+            for (auto& listener : entry.second) {
```
Andrew-Fryer, line 90: I think we should see if `enumerateEventListenerTypes` is hot, and if so, we should probably keep a separate total count so we don't need to accumulate here at all.

Andrew-Fryer: Hold on, I see:
`Vector<std::pair<AtomString, EventListenerVector>, 0, CrashOnOverflow, 4> m_entries;`
It seems that `m_entries` has "on average 1.63" elements (from 259920@main commit msg).
So, this might not be too bad, but maybe still worth it...? (I think we need to profile.)

Source/WebCore/dom/EventDispatcher.cpp
```diff
@@ -274,7 +283,7 @@ static void dispatchEventWithType(std::span<T* const> targets, Event& event)
     event.setTarget(RefPtr { targets.front() });
     event.setEventPath(eventPath);
     event.resetBeforeDispatch();
-    dispatchEventInDOM(event, eventPath);
+    dispatchEventInDOM(event, eventPath, { true, true });
```
Andrew-Fryer: IIUC, we're conservatively assuming bubbling and capturing here.
If that's right, I think `{ 1, 1 }` is more readable and that might be worthy of a comment so we know that we're being conservative...?

Source/WebCore/dom/EventDispatcher.cpp
```diff
@@ -88,21 +88,42 @@ static bool NODELETE isInShadowTree(EventTarget* target)
...
-        const EventContext& eventContext = path.contextAt(i - 1);
-        if (eventContext.currentTarget() == eventContext.target())
-            event.setEventPhase(Event::AT_TARGET);
-        else
-            event.setEventPhase(Event::CAPTURING_PHASE);
-        eventContext.handleLocalEvents(event, EventTarget::EventInvokePhase::Capturing);
-        if (event.propagationStopped())
-            return;
+    auto counts = document.eventListenerCountsOfType(event.type());
+    if (counts.capturing && counts.bubbling)
```
Andrew-Fryer: I personally don't understand why we're using the counts to know if we need to check legacy and why we didn't need to check legacy before this change.

Perhaps we should be doing something similar to what `EventTarget::fireEventListeners` does??
...that'd be something like: `if (auto& something = event.target()->eventTargetData()->eventListenerMap.find(event.type()))`

Source/WebCore/dom/Document.cpp
annevk: No this works because of https://en.cppreference.com/w/c/language/conversion.html#Integer_promotions

I like your idea of removing the branch completely though. So I'll go with your approach but I'm removing both casts as I'm pretty sure they're not needed.

Source/WebCore/dom/EventDispatcher.cpp
annevk: We did check this before. I'll move the method so the diff is clearer.

Source/WebCore/dom/EventListenerMap.h
annevk: This is not hot code.

PR #58817, opened 2026-02-17
Add wasm multimemory support to base instructions

asworkjsc:
#### 4ac3b92ed23f41a25bd955cafac8a8ca34263c30
<pre>
Add wasm multimemory support to base instructions
<a href="https://bugs.webkit.org/show_bug.cgi?id=277743">https://bugs.webkit.org/show_bug.cgi?id=277743</a>
<a href="https://rdar.apple.com/133394285">rdar://133394285</a>

Reviewed by Keith Miller.

This patch adds end-to-end support for multiple memories to wasm scalar
loads and stores, and memory.{size,grow,fill,copy,init}. This involves
modifying the wasm function parser to parse the memory index field, as
well as all classes that use the function parser to read the bytecode.
BBQ, OMG, and the IPInt metadata generator are modified to accept a memory
index argument. In all tiers, memory 0 is handled using a fast path that
keeps the old behavior to avoid performance regressions.

In all tiers loads and stores must materialize the pointer before
accessing it, and the pointer materialization code must fetch the memory
base and bounds checking size if the memory is not 0. Bulk memory
operations are generally handled using C++ functions so these have a
memory index argument added.

Tests: JSTests/wasm/stress/bulk-multimemory.js
       JSTests/wasm/stress/scalar-multimemory.js

* JSTests/wasm/function-tests/grow-memory.js:
* JSTests/wasm/references/parse_unreachable.js:
(invalidMemoryFillUnreachable):
(invalidMemoryCopyUnreachable):
* JSTests/wasm/stress/bulk-multimemory.js: Added.
(data.segment.second.local.0.local.1.local.2.memory.init.1.1.func.export.string_appeared_here.param.i32.result.i64.local.0.i64.load.1.data.string_appeared_here.data.string_appeared_here.async test):
* JSTests/wasm/stress/scalar-multimemory.js: Added.
(stores.to.memory.0.are.already.tested.elsewhere.func.export.string_appeared_here.param.i32.i32.local.0.local.1.i32.store.1.func.export.string_appeared_here.param.i32.i64.local.0.local.1.i64.store.1.func.export.string_appeared_here.param.i32.f32.local.0.local.1.f32.store.1.func.export.string_appeared_here.param.i32.f64.local.0.local.1.f64.store.1.func.export.string_appeared_here.param.i32.i32.local.0.local.1.i32.store8.1.func.export.string_appeared_here.param.i32.i32.local.0.local.1.i32.store16.1.func.export.string_appeared_here.param.i32.i64.local.0.local.1.i64.store8.1.func.export.string_appeared_here.param.i32.i64.local.0.local.1.i64.store16.1.func.export.string_appeared_here.param.i32.i64.local.0.local.1.i64.store32.1.async test):
* Source/JavaScriptCore/b3/B3AbstractHeapRepository.h:
* Source/JavaScriptCore/ftl/FTLLowerDFGToB3.cpp:
(JSC::FTL::DFG::LowerDFGToB3::compileCompareStrictEq):
* Source/JavaScriptCore/llint/InPlaceInterpreter.asm:
* Source/JavaScriptCore/llint/InPlaceInterpreter64.asm:
* Source/JavaScriptCore/runtime/ArrayBuffer.cpp:
(JSC::SharedArrayBufferContents::grow):
* Source/JavaScriptCore/runtime/BufferMemoryHandle.cpp:
(JSC::BufferMemoryHandle::registerInstance):
* Source/JavaScriptCore/wasm/WasmBBQJIT.cpp:
(JSC::Wasm::BBQJITImpl::BBQJIT::addGrowMemory):
(JSC::Wasm::BBQJITImpl::BBQJIT::addCurrentMemory):
(JSC::Wasm::BBQJITImpl::BBQJIT::addMemoryFill):
(JSC::Wasm::BBQJITImpl::BBQJIT::addMemoryCopy):
(JSC::Wasm::BBQJITImpl::BBQJIT::addMemoryInit):
(JSC::Wasm::BBQJITImpl::BBQJIT::loadWebAssemblyGlobalState):
(JSC::Wasm::BBQJITImpl::BBQJIT::emitTailCall):
(JSC::Wasm::BBQJITImpl::BBQJIT::addCall):
* Source/JavaScriptCore/wasm/WasmBBQJIT.h:
* Source/JavaScriptCore/wasm/WasmBBQJIT64.cpp:
(JSC::Wasm::BBQJITImpl::BBQJIT::load):
(JSC::Wasm::BBQJITImpl::BBQJIT::store):
(JSC::Wasm::BBQJITImpl::BBQJIT::restoreWebAssemblyGlobalStateAfterWasmCall):
(JSC::Wasm::BBQJITImpl::BBQJIT::addSIMDLoad):
(JSC::Wasm::BBQJITImpl::BBQJIT::addSIMDStore):
(JSC::Wasm::BBQJITImpl::BBQJIT::addSIMDLoadExtend):
(JSC::Wasm::BBQJITImpl::BBQJIT::addSIMDLoadPad):
* Source/JavaScriptCore/wasm/WasmBBQJIT64.h:
(JSC::Wasm::BBQJITImpl::BBQJIT::emitCheckAndPrepareAndMaterializePointerApply):
* Source/JavaScriptCore/wasm/WasmBinding.cpp:
(JSC::Wasm::wasmToWasm):
* Source/JavaScriptCore/wasm/WasmBinding.h:
* Source/JavaScriptCore/wasm/WasmConstExprGenerator.cpp:
* Source/JavaScriptCore/wasm/WasmEntryPlan.cpp:
(JSC::Wasm::EntryPlan::generateWasmToWasmStubs):
(JSC::Wasm::EntryPlan::generateWasmToJSStubs):
* Source/JavaScriptCore/wasm/WasmFunctionIPIntMetadataGenerator.cpp:
(JSC::Wasm::FunctionIPIntMetadataGenerator::addMemoryIndex):
* Source/JavaScriptCore/wasm/WasmFunctionIPIntMetadataGenerator.h:
* Source/JavaScriptCore/wasm/WasmFunctionParser.h:
(JSC::Wasm::FunctionParser&lt;Context&gt;::load):
(JSC::Wasm::FunctionParser&lt;Context&gt;::store):
(JSC::Wasm::FunctionParser&lt;Context&gt;::parseMemoryFillImmediate):
(JSC::Wasm::FunctionParser&lt;Context&gt;::parseMemoryCopyImmediates):
(JSC::Wasm::FunctionParser&lt;Context&gt;::parseMemoryInitImmediates):
(JSC::Wasm::FunctionParser&lt;Context&gt;::parseExpression):
(JSC::Wasm::FunctionParser&lt;Context&gt;::parseUnreachableExpression):
* Source/JavaScriptCore/wasm/WasmIPIntGenerator.cpp:
(JSC::Wasm::IPIntGenerator::load):
(JSC::Wasm::IPIntGenerator::store):
(JSC::Wasm::IPIntGenerator::addGrowMemory):
(JSC::Wasm::IPIntGenerator::addCurrentMemory):
(JSC::Wasm::IPIntGenerator::addMemoryFill):
(JSC::Wasm::IPIntGenerator::addMemoryCopy):
(JSC::Wasm::IPIntGenerator::addMemoryInit):
* Source/JavaScriptCore/wasm/WasmIPIntGenerator.h:
* Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.cpp:
(JSC::IPInt::WASM_IPINT_EXTERN_CPP_DECL):
* Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.h:
* Source/JavaScriptCore/wasm/WasmIRGeneratorHelpers.h:
(JSC::Wasm::buildEntryBufferForCatch):
* Source/JavaScriptCore/wasm/WasmLimits.h:
* Source/JavaScriptCore/wasm/WasmMemory.cpp:
(JSC::Wasm::Memory::grow):
* Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp:
(JSC::Wasm::OMGIRGenerator::reloadMemoryRegistersFromInstance):
(JSC::Wasm::OMGIRGenerator::emitIndirectCall):
(JSC::Wasm::OMGIRGenerator::addGrowMemory):
(JSC::Wasm::OMGIRGenerator::addCurrentMemory):
(JSC::Wasm::OMGIRGenerator::addMemoryFill):
(JSC::Wasm::OMGIRGenerator::addMemoryInit):
(JSC::Wasm::OMGIRGenerator::addMemoryCopy):
(JSC::Wasm::OMGIRGenerator::emitCheckAndPreparePointer):
(JSC::Wasm::OMGIRGenerator::load):
(JSC::Wasm::OMGIRGenerator::store):
(JSC::Wasm::OMGIRGenerator::atomicLoad):
(JSC::Wasm::OMGIRGenerator::atomicStore):
(JSC::Wasm::OMGIRGenerator::atomicBinaryRMW):
(JSC::Wasm::OMGIRGenerator::atomicCompareExchange):
(JSC::Wasm::OMGIRGenerator::addSIMDLoad):
(JSC::Wasm::OMGIRGenerator::addSIMDStore):
(JSC::Wasm::OMGIRGenerator::addSIMDLoadSplat):
(JSC::Wasm::OMGIRGenerator::addSIMDLoadLane):
(JSC::Wasm::OMGIRGenerator::addSIMDStoreLane):
(JSC::Wasm::OMGIRGenerator::addSIMDLoadExtend):
(JSC::Wasm::OMGIRGenerator::addSIMDLoadPad):
(JSC::Wasm::OMGIRGenerator::emitDirectCall):
* Source/JavaScriptCore/wasm/WasmOperations.cpp:
(JSC::Wasm::JSC_DEFINE_NOEXCEPT_JIT_OPERATION):
* Source/JavaScriptCore/wasm/WasmOperations.h:
* Source/JavaScriptCore/wasm/WasmOperationsInlines.h:
(JSC::Wasm::growMemory):
(JSC::Wasm::memorySize):
(JSC::Wasm::memoryInit):
(JSC::Wasm::memoryFill):
(JSC::Wasm::memoryCopy):
* Source/JavaScriptCore/wasm/WasmSectionParser.cpp:
(JSC::Wasm::SectionParser::parseMemoryHelper):
* Source/JavaScriptCore/wasm/debugger/WasmMemoryHandler.cpp:
(JSC::Wasm::MemoryHandler::readMemoryData):
(JSC::Wasm::MemoryHandler::handleWasmMemoryRegionInfo):
* Source/JavaScriptCore/wasm/js/JSToWasm.cpp:
(JSC::Wasm::createJSToWasmJITShared):
(JSC::Wasm::FunctionSignature::jsToWasmICEntrypoint const):
* Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.cpp:
(JSC::JSWebAssemblyInstance::JSWebAssemblyInstance):
(JSC::JSWebAssemblyInstance::memoryInit):
* Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h:
* Source/JavaScriptCore/wasm/js/WasmToJS.cpp:
(JSC::Wasm::materializeImportJSCell):
(JSC::Wasm::wasmToJS):
* Source/JavaScriptCore/wasm/js/WasmToJS.h:

Canonical link: <a href="https://commits.webkit.org/309850@main">https://commits.webkit.org/309850@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/7a5ab0e97605ed1f8fb5bc42e20611e1c1d4b654

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/151762 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/24543 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/18113 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/160504 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/105219 "Built successfully") | [loading 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/430e769a-f0c3-457c-ae54-1ecc76d9b6d7) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/25036 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/24836 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/117227 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/83193 "Passed tests") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/11cb46c4-21f6-40ca-aa24-4582d6d2dc56) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/154722 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/19383 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/136184 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/97942 "Passed tests") | | [❌ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/45c60c37-6f60-48ed-bbde-471914b5a8b9) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/18468 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/16405 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/8339 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/143766 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/128098 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/14088 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/162968 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/12563 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/6117 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/15679 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/125246 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/24342 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/20464 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/125427 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/24343 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/135884 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/80918 "The change is no longer eligible for processing.") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/23320 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/20464 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/12659 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/183373 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/23959 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/88245 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/46768 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/23651 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/23811 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/23711 "Built successfully") | | | |


Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -70,7 +70,7 @@ class BaselineData;
 }
 
 // The layout of a JSWebAssemblyInstance is
-//     { struct JSWebAssemblyInstance }[ WasmOrJSImportableFunctionCallLinkInfo ][ Wasm::Table* ][ Global::Value ][ Wasm::BaselineData* ][ WebAssemblyGCStructure* ][ Allocator* ]
+//     { struct JSWebAssemblyInstance }[ WasmOrJSImportableFunctionCallLinkInfo ][ Wasm::Table* ][ UGPRPair ][ Global::Value ][ Wasm::BaselineData* ][ WebAssemblyGCStructure* ][ Allocator* ]
```
kmiller68: It's not obvious what this pair is without additional context.  In this case I'd just make a `struct WasmMemoryBaseAndBound` for clarity. I would also put memory zero in here, removing the existing labels. Lastly, it should probably be the first trailing array since the previous two are only used for calls, which are less common than memory access. This makes the IPInt code simpler and doesn't matter for JIT code since the size of these vectors are all constant for a given ModuleInfo anyway.

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -468,6 +495,11 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
     CalleeBits m_builtinCalleeBits[WASM_BUILTIN_COUNT];
     Wasm::ExceptionType m_exception { Wasm::ExceptionType::Termination };
     uint32_t m_debugId { 0 };
+
+public:
+    // IPInt doesn't support function calls with more than 4 parameters
+    uint8_t m_scratchMemoryIndex0;
+    uint8_t m_scratchMemoryIndex1;
```
kmiller68: Remove, if you need more parameters you can just add another call macro that takes 6 arguments. But I don't think you need this with the other changes.

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -440,9 +466,10 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
     FixedVector<WriteBarrier<JSWebAssemblyMemory>> m_memories;
     FixedVector<WriteBarrier<JSWebAssemblyTable>> m_tables;
     StackManager::Mirror m_stackMirror;
-    CagedPtr<Gigacage::Primitive, void> m_cachedMemory; // only memory 0 is cached
+    CagedPtr<Gigacage::Primitive, void> m_cachedMemory; // memory 0 is cached here
     size_t m_cachedBoundsCheckingSize { 0 };
     size_t m_cachedMemorySize { 0 };
+    FixedVector<UGPRPair> m_cachedMemoryBaseAndSizeVector;
```
kmiller68: Remove, you have this as a trailing array now.

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -440,9 +466,10 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
     FixedVector<WriteBarrier<JSWebAssemblyMemory>> m_memories;
     FixedVector<WriteBarrier<JSWebAssemblyTable>> m_tables;
     StackManager::Mirror m_stackMirror;
-    CagedPtr<Gigacage::Primitive, void> m_cachedMemory; // only memory 0 is cached
+    CagedPtr<Gigacage::Primitive, void> m_cachedMemory; // memory 0 is cached here
     size_t m_cachedBoundsCheckingSize { 0 };
     size_t m_cachedMemorySize { 0 };
```
kmiller68: Remove and use the ones from the trailing array.

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -214,11 +222,19 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
...
 #endif
-            // only memory 0 is cached
+            // only memory 0 is cached here
             m_cachedMemorySize = m_wasmMemory->size();
             m_cachedMemory = CagedPtr<Gigacage::Primitive, void>(m_wasmMemory->basePointer());
             m_cachedIsMemory64 = moduleInformation().memory(0).isMemory64();
             ASSERT(m_wasmMemory->basePointer() == cachedMemory());
+
+            // when the wasm instance is being initialized, this will crash
+            // base and size are loaded for the first time in setMemory
```
kmiller68: If the single memory case worked what changed about the multi-memory case?

Source/JavaScriptCore/wasm/WasmLimits.h
kmiller68, line 40: Nit: Can you add a link to the JS-API Wasm spec document for the limits?

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
```diff
@@ -951,15 +953,65 @@ macro loadMemoryOffsetAndAdvanceMC(dstReg, tmpReg, instrLenReg)
 .done:
 end
 
+macro ldstAdvanceMCAndMakePointer(ptrReg, instrLenReg, size, wasmAddrReg, scratch, preserve)
```
kmiller68: We generally avoid using abbreviations like this since it can be hard to know what they stand for at first glance.

```suggestion
macro loadStoreAdvanceMCAndMakePointer(ptrReg, instrLenReg, size, wasmAddrReg, scratch, preserve)
```

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
```diff
@@ -951,15 +953,65 @@ macro loadMemoryOffsetAndAdvanceMC(dstReg, tmpReg, instrLenReg)
...
+    leap size - 1[ptrReg], wasmAddrReg # use wasmAddrReg as scratch
+    bpb wasmAddrReg, boundsCheckingSize, .boundsCheckSucceeded
+    ipintException(OutOfBoundsMemoryAccess)
+.boundsCheckSucceeded:
+    addp memoryBase, ptrReg
+
+    # if memory index != 0, restore MB and BC
+    btiz scratch, .doneRestoringMBAndBC
+    pop boundsCheckingSize, memoryBase
+.doneRestoringMBAndBC:
```
kmiller68: If you make the changes I suggested in the instance this just becomes:

```suggestion
# memAddress contains the address after adding the appropriate base after this macro falls through.
macro preparePointerAndCheckMemoryBound(memAddress, memIndex, scratch1, scratch2,  size, outOfBoundsLabel)
    btinz memIndex, .checkFromInstanceCache
    # You'll have to add these two to offlineasm (the latter only for arm64) it shouldn't be hard though.
    baddpo memoryBase, memAddress, outOfBoundsLabel 
    baddpo memAddress, size - 1, scratch1, outOfBoundsLabel
    bpaeq scratch1, boundsCheckingSize, outOfBoundsLabel
    jmp .continue
.checkFromInstanceCache:
    mulp sizeof(WasmMemoryBaseAndBound), memIndex
    # Or something similar e.g. sizeof(JSWebAssemblyInstance) to get the (now fixed) base of the cached base and bounds.
    loadPairp (constexpr JSWebAssemblyInstance::offsetOfCachedMemoryBaseSize(0))[memIndex], scratch1, scratch2
    baddpo scratch1, memAddress, outOfBoundsLabel
    baddpo memAddress, size - 1, scratch1, outOfBoundsLabel
    bpaeq scratch1, scratch2, outOfBoundsLabel
.continue
end

macro loadMemoryIndexAndOffsetThenAdvanceMC(dstMemIndexReg, dstOffsetReg, instrLenReg)
    loadb IPInt::MemoryIndexMetadata::memoryIndex[MC], dstMemIndexReg
	loadb JSWebAssemblyInstance::m_cachedIsMemory64[wasmInstance], dstOffsetReg
	btiz dstOffsetReg, .memory32
	loadq IPInt::Const64Metadata::value[MC], dstOffsetReg
    loadb IPInt::Const64Metadata::instructionLength[MC], instrLenReg
	advanceMC(constexpr (sizeof(IPInt::MemoryIndexMetadata) + sizeof(IPInt::Const64Metadata))), 
	jmp .done
.memory32:
	loadi IPInt::Const32Metadata::value[MC], dstOffsetReg
    loadb IPInt::Const32Metadata::instructionLength[MC], instrLenReg
	advanceMC(constexpr (sizeof(IPInt::MemoryIndexMetadata) + sizeof(IPInt::Const32Metadata)))
.done:
end
```

kmiller68 (commented):
Some comments

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
asworkjsc: IIRC the first time it's called is during instantiation before the memory vector is initialized. At first there was no memory vector so no need for that loop. After I put it in I found the easiest way to avoid the crash was to add this special case to check if the memory vector accessed by getMemoryBaseAndSize had not been filled yet.

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
asworkjsc: I think the second macro needs to advance MC by sizeof(MemoryIndexMetadata) before trying to access the Const{32|64}Metadata record at MC

kmiller68: Ah yeah it needs to add `sizeof(MemoryIndexMetadata)` so e.g. `loadi sizeof MemoryIndexMetadata  + IPInt::Const32Metadata::value[MC], dstOffsetReg`

asworkjsc: The first macro tries to compare the computed pointer to the memory size, when it should compare the computed wasm address (not added to the base pointer) to the memory size

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -470,6 +475,8 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
     uint32_t m_debugId { 0 };
 };
 
+constexpr ptrdiff_t WASM_INSTANCE_OFFSET_OF_CACHED_BASE_SIZE = JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(0);
```
kmiller68: Let's remove this and just use `(constexpr JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(0))` in IPInt

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
```diff
@@ -971,14 +959,67 @@ macro loadMemoryOffsetAndAdvanceMC(dstReg, tmpReg, instrLenReg)
...
 end
 
+macro preparePointerAndCheckMemoryBound(memAddress, memIndex, offsetReg, scratch, size)
+    if not (ARM64 or ARM64E)
+        break # unclear if x86 supports addps and bcs the same way as ARM64
+    end
+    # memAddress contains the address after adding the appropriate base after this macro falls through.
+    btinz memIndex, .checkFromInstanceCache
+    addps offsetReg, memAddress
+    bcs .outOfBounds
```
kmiller68: Let's do what preparePointerAndCheckMemoryBound does and do `addp` then `bpb`. Then this works for X86. And include the add the `# FIXME: make this baddpc` (FIXME added).

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
```diff
@@ -971,14 +959,67 @@ macro loadMemoryOffsetAndAdvanceMC(dstReg, tmpReg, instrLenReg)
...
+    bpaeq scratch, boundsCheckingSize, .outOfBounds
+    addp memoryBase, memAddress
+    jmp .continue
+.checkFromInstanceCache:
+    addps offsetReg, memAddress
+    bcs .outOfBounds
+    mulp constexpr (sizeof(JSWebAssemblyInstance::WasmMemoryBaseAndSize)), memIndex
+    addp wasmInstance, memIndex # computing pointer wasmInstance + offset of cached base/size vector + memoryIndex * sizeof(WasmMemoryBaseAndSize)
+    loadp constexpr JSC::WASM_INSTANCE_OFFSET_OF_CACHED_BASE_SIZE[memIndex], scratch
+    loadp constexpr (JSC::WASM_INSTANCE_OFFSET_OF_CACHED_BASE_SIZE + 8)[memIndex], offsetReg # reuse offsetReg for memory size
```
kmiller68: This can be `loadPair (constexpr JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(0))[memIndex], scratch, offsetReg`

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -188,37 +201,30 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
...
-    bool memoryInit(uint64_t dstAddress, uint32_t srcAddress, uint32_t length, uint32_t dataSegmentIndex);
+    bool memoryInit(uint64_t dstAddress, uint32_t srcAddress, uint32_t length, uint32_t dataSegmentIndex, uint8_t memoryIndex);
 
     void dataDrop(uint32_t dataSegmentIndex);
 
-    void* cachedMemory() const { return m_cachedMemory.getMayBeNull(); }
-    size_t cachedBoundsCheckingSize() const { return m_cachedBoundsCheckingSize; }
-    size_t cachedMemorySize() const { return m_cachedMemorySize; }
+    void* cachedMemory0() const { return m_memories[0]->memory().basePointer(); }
+    size_t cachedMemory0BoundsCheckingSize() const { return m_memories[0]->memory().size(); }
```
Achierius: This will crash if there are no memories, unlike in the old code

Source/JavaScriptCore/wasm/WasmBBQJIT64.cpp
```diff
@@ -3779,7 +3781,8 @@ void NODELETE BBQJIT::notifyFunctionUsesSIMD()
         RELEASE_ASSERT_NOT_REACHED();
     }
 
-    result = emitCheckAndPrepareAndMaterializePointerApply(pointer, uoffset, sizeof(double), [&](auto location) -> Value {
+    uint8_t memoryIndex = 0; // FIXME(wasm-multimemory)
```
Achierius, line 3804: What's to be fixed here?

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
```diff
@@ -971,14 +959,67 @@ macro loadMemoryOffsetAndAdvanceMC(dstReg, tmpReg, instrLenReg)
...
+    advanceMC(constexpr (sizeof(IPInt::MemoryIndexMetadata)))
+    btiz dstOffsetReg, .memory32
+    loadq IPInt::Const64Metadata::value[MC], dstOffsetReg
+    loadb IPInt::Const64Metadata::instructionLength[MC], instrLenReg
+    advanceMC(constexpr (sizeof(IPInt::Const64Metadata)))
+    jmp .done
+.memory32:
+    loadi IPInt::Const32Metadata::value[MC], dstOffsetReg
+    loadb IPInt::Const32Metadata::instructionLength[MC], instrLenReg
+    advanceMC(constexpr (sizeof(IPInt::Const32Metadata)))
```
kmiller68:
```suggestion
    const memoryIndexSize = sizeof IPInt::MemoryIndexMetadata
    btiz dstOffsetReg, .memory32
    loadq memoryIndexSize + IPInt::Const64Metadata::value[MC], dstOffsetReg
    loadb memoryIndexSize + IPInt::Const64Metadata::instructionLength[MC], instrLenReg
    advanceMC(memoryIndexSize + (sizeof IPInt::Const64Metadata))
    jmp .done
.memory32:
    loadi memoryIndexSize + IPInt::Const32Metadata::value[MC], dstOffsetReg
    loadb memoryIndexSize + IPInt::Const32Metadata::instructionLength[MC], instrLenReg
    advanceMC(memoryIndexSize + (sizeof IPInt::Const32Metadata))
```

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
```diff
@@ -971,14 +959,67 @@ macro loadMemoryOffsetAndAdvanceMC(dstReg, tmpReg, instrLenReg)
...
+.done:
+end
+
+macro loadStoreAdvanceMCAndMakePointer(instrLenReg, size, wasmAddrReg, scratch1, scratch2, scratch3)
```
kmiller68: Nit: I would parameterize this macro as:

```suggestion
macro loadStoreAdvanceMCAndMakePointer(instrLenReg, wasmAddrReg, size, scratch1, scratch2, scratch3)
```

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -134,6 +135,18 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
...
     MemoryMode memoryMode() const { return memory()->memory().mode(); }
 
+    struct WasmMemoryBaseAndSize {
+        UCPURegister base;
+        UCPURegister size;
+    };
+
+    WasmMemoryBaseAndSize getMemoryBaseAndSize(unsigned i)
+    {
+        // IPInt handles return value registers as r0 and r1
```
kmiller68: I don't think you need this comment anymore.

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -188,37 +201,30 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
...
 
-    bool memoryInit(uint64_t dstAddress, uint32_t srcAddress, uint32_t length, uint32_t dataSegmentIndex);
+    bool memoryInit(uint64_t dstAddress, uint32_t srcAddress, uint32_t length, uint32_t dataSegmentIndex, uint8_t memoryIndex);
 
     void dataDrop(uint32_t dataSegmentIndex);
 
-    void* cachedMemory() const { return m_cachedMemory.getMayBeNull(); }
-    size_t cachedBoundsCheckingSize() const { return m_cachedBoundsCheckingSize; }
-    size_t cachedMemorySize() const { return m_cachedMemorySize; }
+    void* cachedMemory0() const { return m_memories[0]->memory().basePointer(); }
```
Achierius: Given that you're not caching these values separately, is there a reason to keep the cachedMemory0 functions around rather than just going through getMemoryBaseAndSize? Naively it looks like they're doing the same thing.

Depending on how perf sensitive this is I could imagine wanting a fast-path for these, but that would require bringing back something like the m_cachedMemorySize member.

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
Achierius: Why not check for whether the memory vector is empty here, rather than having different behavior per caller?

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -188,37 +201,30 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
...
+    size_t cachedMemory0BoundsCheckingSize() const { return m_memories[0]->memory().size(); }
+    size_t cachedMemory0Size() const { return !m_memories.size() ? 0 : cachedMemory0BoundsCheckingSize(); }
 
-    void updateCachedMemory()
+    void updateCachedMemory(bool isUpdatingCache)
```
Achierius: This is confusing -- what would "updateCachedMemory" do if not update the cache?

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -188,37 +201,30 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
...
-            // on ARMv7 yet. In order to get more of the test suite to run, we
-            // can still use a shared memory by using bounds checking, by using
-            // the actual size here, but this means we cannot grow the shared
-            // memory safely in case it's used by multiple threads. Once the
-            // signal handler are available, m_cachedBoundsCheckingSize should
-            // be set to use m_wasmMemory->mappedCapacity() like other platforms,
-            // and at that point growing the shared memory will be safe.
-            m_cachedBoundsCheckingSize = m_wasmMemory->size();
-#else
-            m_cachedBoundsCheckingSize = m_wasmMemory->mappedCapacity();
```
Achierius, line 215: For the 0th memory (or more generally, any memories in the signalling-mode) don't you still want to consider the mappedCapacity?

Source/JavaScriptCore/wasm/WasmBBQJIT64.cpp
asworkjsc: Adding multimemory support to SIMD and atomics - right now memoryIndex is hardcoded to 0 for instructions that are not supported yet

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -1930,61 +1930,83 @@ auto OMGIRGenerator::emitIndirectCall(Value* calleeInstance, Value* calleeCode,
...
-auto OMGIRGenerator::addCurrentMemory(ExpressionType& result) -> PartialResult
+auto OMGIRGenerator::addCurrentMemory(ExpressionType& result, uint8_t memoryIndex) -> PartialResult
 {
-    static_assert(sizeof(std::declval<Memory*>()->size()) == sizeof(uintptr_t), "codegen relies on this size");
+    if (!memoryIndex) {
+        static_assert(sizeof(std::declval<Memory*>()->size()) == sizeof(uintptr_t), "codegen relies on this size");
 
-    Value* size = m_currentBlock->appendNew<MemoryValue>(m_proc, Load, pointerType(), origin(), instanceValue(), safeCast<int32_t>(JSWebAssemblyInstance::offsetOfCachedMemorySize()));
-    m_heaps.decorateMemory(&m_heaps.JSWebAssemblyInstance_cachedMemorySize, size);
+        Value* size = m_currentBlock->appendNew<MemoryValue>(m_proc, Load, pointerType(), origin(), instanceValue(), safeCast<int32_t>(JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(0) + sizeof(UCPURegister)));
```
kmiller68: We should definitely have an abstract heap for these.

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -134,6 +135,18 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
 
     MemoryMode memoryMode() const { return memory()->memory().mode(); }
 
+    struct WasmMemoryBaseAndSize {
+        UCPURegister base;
+        UCPURegister size;
+    };
```
kmiller68, line 153:
```suggestion
    struct WasmMemoryBaseAndSize {
        void* base;
        size_t size;
    };
```

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -188,37 +201,30 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
...
+    bool memoryInit(uint64_t dstAddress, uint32_t srcAddress, uint32_t length, uint32_t dataSegmentIndex, uint8_t memoryIndex);
 
     void dataDrop(uint32_t dataSegmentIndex);
 
-    void* cachedMemory() const { return m_cachedMemory.getMayBeNull(); }
-    size_t cachedBoundsCheckingSize() const { return m_cachedBoundsCheckingSize; }
-    size_t cachedMemorySize() const { return m_cachedMemorySize; }
+    void* cachedMemory0() const { return m_memories[0]->memory().basePointer(); }
+    size_t cachedMemory0BoundsCheckingSize() const { return m_memories[0]->memory().size(); }
+    size_t cachedMemory0Size() const { return !m_memories.size() ? 0 : cachedMemory0BoundsCheckingSize(); }
```
kmiller68: You can't load this from the memory. That would be semantically incorrect with shared memory grow. These have to load from the cached memory. Also is BoundsCheckingSize actually the same as cachedMemory size for shared memories? This should probably be covered in the commit message for posterity.


```suggestion
    void* cachedMemory0() const { return cachedMemoryBaseSizePairs()[0].base; }
    size_t cachedMemory0BoundsCheckingSize() const { return cachedMemoryBaseSizePairs()[0].size; }
    size_t cachedMemory0Size() const { return !m_memories.size() ? 0 : cachedMemory0BoundsCheckingSize(); }
```

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
asworkjsc: Might as well get rid of them, since I got rid of the cachedMemory fields

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
asworkjsc: I'll get rid of those methods since I already got rid of the backing fields

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
asworkjsc: All the places where those accessors are used assume that there is a valid memory to begin with, otherwise the base and size would be invalid

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
asworkjsc: Since m_memories is a FixedVector it is allocated up front with invalid data, so the easiest fix was to explicitly pass an argument into the method. Now that the cached fields are gone I can probably get rid of the boolean argument to this method

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
asworkjsc: I'm getting rid of the argument, looks like it's unnecessary since the one place where this method is called with false will now only update m_cachedIsMemory64, which will eventually be obsolete since it assumes there is only a single memory

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
asworkjsc: I forgot we still have to keep that for x86, let me add that to getMemoryBaseAndSize

asworkjsc: When I use mappedCapacity in place of size it causes crashes; I'm still not sure why mappedCapacity didn't cause a crash in the old code, since it seems like size should be the correct method on all platforms

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
```diff
@@ -958,14 +946,63 @@ macro loadMemoryOffsetAndAdvanceMC(dstReg, tmpReg, instrLenReg)
...
+.memory32:
+    loadi memoryIndexSize + IPInt::Const32Metadata::value[MC], instrLenReg # reuse instrLenReg to store offset
+    baddpc(instrLenReg, wasmAddrReg, .outOfBounds)
+    move size - 1, scratch2
+    baddpc(wasmAddrReg, scratch2, .outOfBounds)
+    loadb memoryIndexSize + IPInt::Const32Metadata::instructionLength[MC], instrLenReg
+    loadb IPInt::MemoryIndexMetadata::memoryIndex[MC], scratch # scratch contains memory index now
+    advanceMC(memoryIndexSize + sizeof IPInt::Const32Metadata)
+.doneLoadingWasmAddress:
+
```
kmiller68, line 981: If you're going to have a new line I'd do for consistency with the rest of the code:

```suggestion

.doneLoadingWasmAddress:
```

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
```diff
@@ -958,14 +946,63 @@ macro loadMemoryOffsetAndAdvanceMC(dstReg, tmpReg, instrLenReg)
...
+    bpaeq scratch2, boundsCheckingSize, .outOfBounds # scratch2 contains wasm address + size - 1
+    addp memoryBase, wasmAddrReg
+    jmp .done
+
+.memoryIsNotZero:
+    mulp constexpr (sizeof(JSWebAssemblyInstance::WasmMemoryBaseAndSize)), scratch
+    addp wasmInstance, scratch # computing instance + offset of base/size vector + memory index * sizeof(base size pair)
+    loadp (constexpr (JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(0) + sizeof(void*))) [scratch], scratch2 # bounds checking size
+    bpaeq wasmAddrReg, scratch2, .outOfBounds
+    loadp (constexpr (JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(0))) [scratch], scratch2 # memory base
```
kmiller68:
```suggestion
    FIXME: Use a loadPair here
    loadp (constexpr (JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(0) + sizeof(void*))) [wasmInstance, scratch], scratch2 # bounds checking size
    bpaeq wasmAddrReg, scratch2, .outOfBounds
    loadp (constexpr (JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(0))) [wasmInstance, scratch], scratch2 # memory base
```

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
```diff
@@ -1282,20 +1297,28 @@ end)
...
+    loadb IPInt::MemoryIndexMetadata::memoryIndex[MC], t0
+    advanceMC(constexpr (sizeof(IPInt::MemoryIndexMetadata)))
+    btinz t0, .callMemorySize
+    loadp constexpr (JSWebAssemblyInstance::offsetOfCurrentMemory0())[wasmInstance], t0 # size of memory 0
+    jmp .doneLoadingMemorySize
+.callMemorySize:
+    move t0, a1
+    operationCall(macro() cCall2(_ipint_extern_memory_size) end)
+    move r0, t0
+.doneLoadingMemorySize:
```
kmiller68, line 1323:
```suggestion
    loadb IPInt::MemoryIndexMetadata::memoryIndex[MC], t0
    advanceMC(constexpr (sizeof(IPInt::MemoryIndexMetadata)))
    mulp (mulp constexpr (sizeof(JSWebAssemblyInstance::WasmMemoryBaseAndSize)), t0
    loadp constexpr (JSWebAssemblyInstance::offsetOfCurrentMemory0())[wasmInstance, t0], t0 # bounds checking size
```

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -188,37 +216,37 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
...
 
-    void* cachedMemory() const LIFETIME_BOUND { return m_cachedMemory.getMayBeNull(); }
-    size_t cachedBoundsCheckingSize() const { return m_cachedBoundsCheckingSize; }
-    size_t cachedMemorySize() const { return m_cachedMemorySize; }
+    // FIXME(wasm-memory64): eventually m_cachedIsMemory64 should be removed
+    // since it assumes there is only a single memory
+    void updateCachedIsMemory64()
+    {
+        m_cachedIsMemory64 = moduleInformation().memory(0).isMemory64();
+    }
```
kmiller68: Why does this need to be updated? The fact that a memory is memory64 is fixed.

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -107,17 +107,40 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
...
+            // Shared memory requires signaling memory which is not available
+            // on ARMv7 yet. In order to get more of the test suite to run, we
+            // can still use a shared memory by using bounds checking, by using
+            // the actual size here, but this means we cannot grow the shared
+            // memory safely in case it's used by multiple threads. Once the
+            // signal handler are available, m_cachedBoundsCheckingSize should
+            // be set to use m_memories[i]->mappedCapacity() like other platforms,
+            // and at that point growing the shared memory will be safe.
+            m_memories[i]->size();
+#else
```
kmiller68: We don't need this ARMv7 no longer supports wasm.

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -298,25 +326,23 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
...
-    static constexpr ptrdiff_t offsetOfCachedMemorySize() { return OBJECT_OFFSETOF(JSWebAssemblyInstance, m_cachedMemorySize); }
     static constexpr ptrdiff_t offsetOfCachedTable0Buffer() { return OBJECT_OFFSETOF(JSWebAssemblyInstance, m_cachedTable0Buffer); }
     static constexpr ptrdiff_t offsetOfCachedTable0Length() { return OBJECT_OFFSETOF(JSWebAssemblyInstance, m_cachedTable0Length); }
     static constexpr ptrdiff_t offsetOfTemporaryCallFrame() { return OBJECT_OFFSETOF(JSWebAssemblyInstance, m_temporaryCallFrame); }
     static constexpr ptrdiff_t offsetOfBuiltinCalleeBits() { return OBJECT_OFFSETOF(JSWebAssemblyInstance, m_builtinCalleeBits); }
     static constexpr ptrdiff_t offsetOfCachedIsMemory64() { return OBJECT_OFFSETOF(JSWebAssemblyInstance, m_cachedIsMemory64); }
+    static constexpr ptrdiff_t offsetOfCurrentMemory0() { return OBJECT_OFFSETOF(JSWebAssemblyInstance, m_currentMemory0); }
 
     // Tail accessors.
-    static_assert(sizeof(WasmOrJSImportableFunctionCallLinkInfo) == WTF::roundUpToMultipleOf<sizeof(uint64_t)>(sizeof(WasmOrJSImportableFunctionCallLinkInfo)), "We rely on this for the alignment to be correct");
```
kmiller68, line 311: Why is this `static_assert` not needed anymore?

Source/JavaScriptCore/wasm/WasmBBQJIT64.h
```diff
@@ -88,7 +116,8 @@ auto BBQJIT::emitCheckAndPrepareAndMaterializePointerApply(Value pointer, uint64
         pointerLocation = loadIfNecessary(pointer);
     ASSERT(pointerLocation.isGPR());
 
-    switch (m_mode) {
+    // FIXME: for now we conservatively force bounds checking for nonzero memories
+    switch (memoryIndex ? MemoryMode::BoundsChecking : m_mode) {
```
kmiller68: We should be set upstream and should be an ASSERT here. See how it was done for memory64.

Source/JavaScriptCore/wasm/WasmBBQJIT64.h
```diff
@@ -88,7 +116,8 @@ auto BBQJIT::emitCheckAndPrepareAndMaterializePointerApply(Value pointer, uint64
         pointerLocation = loadIfNecessary(pointer);
     ASSERT(pointerLocation.isGPR());
 
-    switch (m_mode) {
+    // FIXME: for now we conservatively force bounds checking for nonzero memories
```
kmiller68: I don't think this is a FIXME, we'll probably never change that.

Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.cpp
```diff
@@ -565,19 +565,24 @@ WASM_IPINT_EXTERN_CPP_DECL(table_grow, IPIntStackEntry* sp, TableGrowMetadata* m
...
     WasmSlowPathWithoutCallFrameTracer tracer(instance->vm());
-    WASM_RETURN_TWO(reinterpret_cast<void*>(Wasm::growMemory(instance, delta)), 0);
+    WASM_RETURN_TWO(reinterpret_cast<void*>(Wasm::growMemory(instance, delta, memoryIndex)), 0);
 }
 
-WASM_IPINT_EXTERN_CPP_DECL(memory_init, int32_t dataIndex, IPIntStackEntry* sp)
+WASM_IPINT_EXTERN_CPP_DECL(memory_size, uint8_t memoryIndex)
+{
+    IPINT_RETURN(Wasm::memorySize(instance, memoryIndex));
+}
```
kmiller68, line 578: Don't think you need this.

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
asworkjsc: That one is split into multiple loads so that there is no need for another scratch register, since t5 aliases with PL on x86, so we reuse scratch2. There is no worry about a performance regression since this code path only ever fires if memoryIndex != 0

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
asworkjsc: currentMemory0 is different from the base-size-pair vector so that indexed load is illegal. The bounds checking size is the mappedCapacity, not the legal memory size, so I have to do a proper lookup of the memory size. currentMemory0 caches the result of that call for memory 0 specifically to avoid potential performance regressions when executing memory.size 0

Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.cpp
asworkjsc: IPInt has to make a C call for memory.size with memoryIndex != 0

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
kmiller68: It could be `loadPairp (constexpr (JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(0) + sizeof(void*))) [wasmInstance, scratch], scratch, scratch2` as scratch isn't needed anymore after the load.

asworkjsc: Probably doable but then we need a special case for x86 since loadpair x, y, [x] would be split into load x, [x]; load y, [x] which overwrites x after the first instruction. Is the suggestion only that I should leave a FIXME comment instead of writing out the loadpair instruction?

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
asworkjsc: As of right now since we don't officially support memory64, IPInt always checks m_cachedIsMemory64 no matter what memory it's accessing. This will be a correctness problem when memory64 is enabled so eventually it will have to be changed somehow, but that's outside the scope of multimemory support

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
kmiller68: The suggestion is just for the FIXME and to remove the extra add instruction. 

The loads can also be done in any order on X86 so if the base register is one of the destinations that load happens second. Or it could just fail to compile.

asworkjsc: All right, FIXME it is, already removed the add and used the reg+reg addressing mode

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
asworkjsc: That static_assert deals with the first trailing array, which is why it now mentions WasmMemoryBaseAndSize instead of WasmOrJSImportableFunctionCallLinkInfo

Source/JavaScriptCore/wasm/WasmSectionParser.cpp
```diff
@@ -401,6 +403,11 @@ auto SectionParser::parseMemoryHelper(bool isImport) -> PartialResult
         ASSERT(!maximum || *maximum >= initial);
         WASM_PARSER_FAIL_IF(!PageCount::isValid(initial), "Memory's initial page count of "_s, initial, " is invalid"_s);
 
+        // FIXME(wasm-memory64): for now IPInt checks m_cachedIsMemory64 (flag if memory 0 is 64-bit)
+        // no matter which memory is being accessed
+        if (isMemory64)
+            RELEASE_ASSERT(!m_info->memoryCount(), "if using memory64 then multiple memories are illegal for now");
```
kmiller68: Let's `WASM_PARSER_FAIL_IF` rather than crash.

Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.cpp
```diff
@@ -565,19 +565,24 @@ WASM_IPINT_EXTERN_CPP_DECL(table_grow, IPIntStackEntry* sp, TableGrowMetadata* m
...
 {
     WasmSlowPathWithoutCallFrameTracer tracer(instance->vm());
-    WASM_RETURN_TWO(reinterpret_cast<void*>(Wasm::growMemory(instance, delta)), 0);
+    WASM_RETURN_TWO(reinterpret_cast<void*>(Wasm::growMemory(instance, delta, memoryIndex)), 0);
 }
 
-WASM_IPINT_EXTERN_CPP_DECL(memory_init, int32_t dataIndex, IPIntStackEntry* sp)
+WASM_IPINT_EXTERN_CPP_DECL(memory_size, uint8_t memoryIndex)
+{
+    IPINT_RETURN(Wasm::memorySize(instance, memoryIndex));
```
kmiller68, line 577: Nit: Add an `ASSERT_WITH_MESSAGE(memoryIndex, "memory zero should go through the fast path");`

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -134,6 +146,11 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
 
     MemoryMode memoryMode() const { return memory()->memory().mode(); }
 
+    struct WasmMemoryBaseAndSize {
```
kmiller68, line 150: I think we'll probably have to add a cached size here for compatibility with shared-growable memory. Let's add a FIXME to follow up.

Source/JavaScriptCore/wasm/WasmBBQJIT64.h
```diff
@@ -88,7 +116,8 @@ auto BBQJIT::emitCheckAndPrepareAndMaterializePointerApply(Value pointer, uint64
         pointerLocation = loadIfNecessary(pointer);
     ASSERT(pointerLocation.isGPR());
 
-    switch (m_mode) {
+    // conservatively force bounds checking for nonzero memories
+    switch (memoryIndex ? MemoryMode::BoundsChecking : m_mode) {
     case MemoryMode::BoundsChecking: {
         // We're not using signal handling only when the memory is not shared.
         // Regardless of signaling, we must check that no memory access exceeds the current memory size.
```
kmiller68, line 115: Seems like we should just check for `m_info.memory(memoryIndex)` here. Ditto on the `RELEASE_ASSERT` for Signaling below.

Source/JavaScriptCore/wasm/WasmBBQJIT64.h
```diff
@@ -131,12 +161,13 @@ auto BBQJIT::emitCheckAndPrepareAndMaterializePointerApply(Value pointer, uint64
...
+        if (m_info.memory(memoryIndex).isMemory64())
+            return functor(CCallHelpers::BaseIndex(wasmScratchGPR, pointerLocation.asGPR(), CCallHelpers::TimesOne, static_cast<int32_t>(uoffset)));
+        return functor(CCallHelpers::BaseIndex(wasmScratchGPR, pointerLocation.asGPR(), CCallHelpers::TimesOne, static_cast<int32_t>(uoffset), CCallHelpers::Extend::ZExt32));
     }
 
-    if (m_info.theOnlyMemory().isMemory64())
+    if (m_info.memory(memoryIndex).isMemory64())
         m_jit.addPtr(wasmBaseMemoryPointer, pointerLocation.asGPR(), wasmScratchGPR);
     else
         m_jit.addZeroExtend64(wasmBaseMemoryPointer, pointerLocation.asGPR(), wasmScratchGPR);
```
kmiller68: We should fix the X86 code below too: `m_info.memory(memoryIndex).isMemory64()`

Source/JavaScriptCore/wasm/WasmBBQJIT64.h
```diff
@@ -56,14 +58,40 @@ auto BBQJIT::emitCheckAndPrepareAndMaterializePointerApply(Value pointer, uint64
...
+        // overwrite wasmBaseMemoryPointer and wasmBoundsCheckingSizeRegister
+        m_jit.loadPtr(
+            Address(GPRInfo::wasmContextInstancePointer, JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(memoryIndex)),
+            wasmBaseMemoryPointer
+        );
+        m_jit.loadPtr(
+            Address(GPRInfo::wasmContextInstancePointer, JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(memoryIndex) + sizeof(UCPURegister)),
+            wasmBoundsCheckingSizeRegister
+        );
+    }
```
kmiller68, line 76:
```suggestion
    std::optional<ScratchScope<2, 0>> nonZeroMemoryScratches;
    GPRReg baseRegister = wasmBaseMemoryPointer;
    GPRReg boundsCheckingSizeRegister = wasmBoundsCheckingSizeRegister;
    if (memoryIndex) {
        nonZeroMemoryScratches.emplace(*this);
        baseRegister = nonZeroMemoryScratches->gpr(0);
        boundsCheckingSizeRegister = nonZeroMemoryScratches->gpr(1);
        m_jit.loadPtr(
            Address(GPRInfo::wasmContextInstancePointer, JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(memoryIndex)),
            baseRegister
        );
        m_jit.loadPtr(
            Address(GPRInfo::wasmContextInstancePointer, JSWebAssemblyInstance::offsetOfCachedMemoryBaseSizePair(memoryIndex) + sizeof(UCPURegister)),
            boundsCheckingSizeRegister
        );
    }
```

Source/JavaScriptCore/wasm/WasmBBQJIT64.h
```diff
@@ -131,12 +161,13 @@ auto BBQJIT::emitCheckAndPrepareAndMaterializePointerApply(Value pointer, uint64
     bool canUseOffsetForm = uoffset <= static_cast<uint64_t>(std::numeric_limits<int32_t>::max()) && B3::Air::Arg::isValidAddrForm(B3::Air::Move, static_cast<int32_t>(uoffset), Width::Width128);
 #if CPU(ARM64)
     if (canUseOffsetForm) {
-        if (m_info.theOnlyMemory().isMemory64())
-            return functor(CCallHelpers::BaseIndex(wasmBaseMemoryPointer, pointerLocation.asGPR(), CCallHelpers::TimesOne, static_cast<int32_t>(uoffset)));
-        return functor(CCallHelpers::BaseIndex(wasmBaseMemoryPointer, pointerLocation.asGPR(), CCallHelpers::TimesOne, static_cast<int32_t>(uoffset), CCallHelpers::Extend::ZExt32));
+        m_jit.move(wasmBaseMemoryPointer, wasmScratchGPR);
```
kmiller68: I think you don't need this move. Even without the change above, the scope exit runs after functor has returned.

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -440,16 +451,14 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
...
-    size_t m_cachedMemorySize { 0 };
     Wasm::FuncRefTable::Function* m_cachedTable0Buffer { nullptr };
     uint32_t m_cachedTable0Length { 0 };
     const Ref<Wasm::Module> m_module;
     const Ref<const Wasm::ModuleInformation> m_moduleInformation;
     RefPtr<Wasm::InstanceAnchor> m_anchor;
     RefPtr<SourceProvider> m_sourceProvider;
-    bool m_cachedIsMemory64 { false };
+    bool m_cachedIsMemory64 { false }; // FIXME(wasm-memory64): rename this to cachedMemory0IsMemory64 or something similar
+    uint64_t m_currentMemory0; // memory.size for memory 0, handled specially to avoid performance regressions
```
kmiller68: This name reads like it points to the memory handle itself not is a cache of the size.
```suggestion
    uint64_t m_cachedMemory0Size; // memory.size for memory 0, handled specially to avoid performance regressions
```

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -2352,44 +2392,78 @@ inline void OMGIRGenerator::emitWriteBarrier(Value* cell)
...
+
+    // if memoryIndex != 0, force bounds checking
+
+    // FIXME: is this redundant?
+    if (sumOverflows<uint32_t>(offset, sizeOfOperation)) {
+        B3::PatchpointValue* throwException = m_currentBlock->appendNew<B3::PatchpointValue>(m_proc, B3::Void, origin());
+        throwException->setGenerator([this, origin = this->origin()](CCallHelpers& jit, const B3::StackmapGenerationParams&) {
+            this->emitExceptionCheck(jit, origin, ExceptionType::OutOfBoundsMemoryAccess);
+        });
     }
```
kmiller68: I think this is redundant for 32-bit but it wouldn't be for 64-bit.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -1930,61 +1930,85 @@ auto OMGIRGenerator::emitIndirectCall(Value* calleeInstance, Value* calleeCode,
     return { };
 }
 
-auto OMGIRGenerator::addGrowMemory(ExpressionType delta, ExpressionType& result) -> PartialResult
+auto OMGIRGenerator::addGrowMemory(ExpressionType delta, ExpressionType& result, uint8_t memoryIndex) -> PartialResult
 {
+    const auto& memoryIndexValue = push(constant(Int32, memoryIndex));
     result = push(callWasmOperation(m_currentBlock, Int32, operationGrowMemory,
-        instanceValue(), get(delta)));
+        instanceValue(), get(delta), get(memoryIndexValue)));
```
kmiller68: This seems wrong. `push` is for when values are pushed to the wasm value stack (albeit somewhat redundant these days as it only is there for debugging and there are other ways to get that). This is an immediate so you can just do: 


```suggestion
result = push(callWasmOperation(m_currentBlock, Int32, operationGrowMemory,
        instanceValue(), get(delta), constant(Int32, memoryIndex));
```

Ditto for the other places.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -2352,44 +2392,78 @@ inline void OMGIRGenerator::emitWriteBarrier(Value* cell)
...
+    auto* memorySize = m_currentBlock->appendNew<MemoryValue>(
+        m_proc, Load, pointerType(), origin(), instanceValue(),
+        safeCast<int32_t>(baseAndSizeOffset + sizeof(uintptr_t)));
+
+    uint64_t lastLoadedOffset = static_cast<uint64_t>(offset);
+    lastLoadedOffset += static_cast<uint64_t>(sizeOfOperation - 1);
+    auto boundsCheckOffset = constant(Int64, lastLoadedOffset);
     pointer = m_currentBlock->appendNew<Value>(m_proc, ZExt32, origin(), pointer);
-    return m_currentBlock->appendNew<WasmAddressValue>(m_proc, origin(), pointer, GPRInfo::wasmBaseMemoryPointer);
+    Value* outOfBounds = m_currentBlock->appendNew<Value>(m_proc, Above, origin(), m_currentBlock->appendNew<Value>(m_proc, Add, origin(), pointer, boundsCheckOffset), memorySize);
```
kmiller68, line 2395: Let's add a FIXME: to make this a CheckAdd for memory64.

kmiller68 (commented):
This could definitely use a more robust commit message explaining what you changed and why.

Source/JavaScriptCore/wasm/WasmBBQJIT64.h
asworkjsc: MemoryInformation doesn't store the memory mode so that's why I decided to use m_mode to store the mode for memory 0 and conservatively use MemoryMode::BoundsChecking otherwise. I can rename m_mode to m_memory0Mode or something similar in a later patch for clarity if that would help

Source/JavaScriptCore/wasm/js/JSWebAssemblyInstance.h
```diff
@@ -188,37 +206,28 @@ class JSWebAssemblyInstance final : public JSNonFinalObject {
...
-    bool memoryInit(uint64_t dstAddress, uint32_t srcAddress, uint32_t length, uint32_t dataSegmentIndex);
+    bool memoryInit(uint64_t dstAddress, uint32_t srcAddress, uint32_t length, uint32_t dataSegmentIndex, uint8_t memoryIndex);
 
     void dataDrop(uint32_t dataSegmentIndex);
 
-    void* cachedMemory() const LIFETIME_BOUND { return m_cachedMemory.getMayBeNull(); }
-    size_t cachedBoundsCheckingSize() const { return m_cachedBoundsCheckingSize; }
-    size_t cachedMemorySize() const { return m_cachedMemorySize; }
-
     void updateCachedMemory()
```
kmiller68: Nit: Should probably be `updateCachedMemories`.

kmiller68 (approved):
r=me with nome more change

PR #58826, opened 2026-02-17
AX: On macOS with site isolation enabled, VoiceOver cannot traverse into iframes because UIElementsForSearchPredicate only searches within the same process

twilco:
#### c1e16773122e8d52ef7b55c00404eb530575f4a4
<pre>
AX: On macOS with site isolation enabled, VoiceOver cannot traverse into iframes because UIElementsForSearchPredicate only searches within the same process
<a href="https://bugs.webkit.org/show_bug.cgi?id=307994">https://bugs.webkit.org/show_bug.cgi?id=307994</a>
<a href="https://rdar.apple.com/170487304">rdar://170487304</a>

Reviewed by Joshua Hoffman.

This change implements cross-process accessibility search to support VoiceOver
navigation across site-isolated iframes. When a search encounters a remote frame,
it sends async IPC to the child process to continue the search, then continues its
search in its own process. The parent process is also queried via async IPC,
since the search may require moving beyond this frame (e.g. asking for the next
element at the last element in an iframe). When children + parent processes return
results, we merge them back in tree order, resolving AccessibilityRemoteToken results into
NSAccessibilityRemoteUIElements as necessary.

The search uses a single absolute deadline established when the top-level search begins,
rather than giving each nested frame its own independent timeout. When dispatching to child
frames, the remaining time until the deadline (minus IPC overhead buffer) is passed along,
ensuring deeply nested frame hierarchies share the original timeout budget. This prevents
timeout multiplication where N levels of nesting would otherwise allow N × timeout total
wait time, while still guaranteeing each frame gets at least a minimum search duration.

Other key changes:
  - Add AXCrossProcessSearch.{h,cpp} with coordination infrastructure for parallel
    IPC dispatch and cascading timeouts (described above) across nested frames

  - Refactor AXSearchManager to return AccessibilitySearchResultStream that records
    both local results and remote frame placeholders in tree order

  - Add ChromeClient methods for performAccessibilitySearchInRemoteFrame() and
    continueAccessibilitySearchFromChildFrame() to enable bidirectional search.

  - Add uiElementsForSearchPredicate() test helper for verifying multi-result searches.

  - Track accessibility enabled state in WebPageCreationParameters so newly-created
    web processes for site-isolated frames start with accessibility being enabled.
    This was required to make my three new tests pass because accessibility must be
    enabled by the time they got a cross-process accessibility search request.

  - Searched frames are tracked by ID and will not be searched multiple times.

New tests were added to exercise various interesting scenarios:
  - http/tests/site-isolation/accessibility/cross-process-search-headings.html
   * Ensures proper merging of local and remote results.
  - http/tests/site-isolation/accessibility/cross-process-search-nested-iframes.html
   * Verifies search progresses through multiple layers of iframes.
  - http/tests/site-isolation/accessibility/cross-process-search-traversal.html
   * Ensures that we can traverse all the way forwards and back through
     the all content on the page.

* LayoutTests/accessibility/ax-object-destroyed-on-reload-expected.txt:
* LayoutTests/accessibility/ax-object-destroyed-on-reload.html:
* LayoutTests/http/tests/site-isolation/accessibility/cross-process-search-headings-expected.txt: Added.
* LayoutTests/http/tests/site-isolation/accessibility/cross-process-search-headings.html: Added.
* LayoutTests/http/tests/site-isolation/accessibility/cross-process-search-nested-iframes-expected.txt: Added.
* LayoutTests/http/tests/site-isolation/accessibility/cross-process-search-nested-iframes.html: Added.
* LayoutTests/http/tests/site-isolation/accessibility/cross-process-search-traversal-expected.txt: Added.
* LayoutTests/http/tests/site-isolation/accessibility/cross-process-search-traversal.html: Added.
* LayoutTests/http/tests/site-isolation/accessibility/resources/iframe-with-headings.html: Added.
* LayoutTests/http/tests/site-isolation/accessibility/resources/nested-iframe-inner.html: Added.
* LayoutTests/http/tests/site-isolation/accessibility/resources/nested-iframe-middle.html: Added.
* LayoutTests/platform/ios/accessibility/ax-object-destroyed-on-reload-expected.txt:
* Source/WebCore/Sources.txt:
* Source/WebCore/accessibility/AXAnnouncementTypes.h: Added.
(WebCore::AXCoreObject::partialOrder):
(WebCore::AXCoreObject::findMatchingObjectsWithin):
* Source/WebCore/accessibility/AXCrossProcessSearch.cpp: Added.
(WebCore::canDoRemoteSearch):
(WebCore::spinRunLoopUntil):
(WebCore::AXCrossProcessSearchCoordinator::waitWithTimeout):
(WebCore::mergeStreamResults):
(WebCore::computeRemainingTimeout):
(WebCore::dispatchRemoteFrameSearch):
(WebCore::performCrossProcessSearch):
(WebCore::performSearchWithCrossProcessCoordination):
(WebCore::mergeParentSearchResults):
(WebCore::ParentFrameSearchContext::signal):
(WebCore::ParentFrameSearchContext::waitWithTimeout):
(WebCore::ParentFrameSearchContext::markParentDispatched):
(WebCore::ParentFrameSearchContext::didDispatchParent const):
(WebCore::ParentFrameSearchContext::setParentTokens):
(WebCore::ParentFrameSearchContext::takeParentTokens):
(WebCore::performSearchWithParentCoordination):
* Source/WebCore/accessibility/AXCrossProcessSearch.h: Added.
(WebCore::AXCrossProcessSearchCoordinator::create):
(WebCore::AXCrossProcessSearchCoordinator::addPendingRequest):
(WebCore::AXCrossProcessSearchCoordinator::markSearchComplete):
(WebCore::AXCrossProcessSearchCoordinator::responseReceived):
(WebCore::AXCrossProcessSearchCoordinator::storeRemoteResults):
(WebCore::AXCrossProcessSearchCoordinator::takeRemoteResults):
(WebCore::AXCrossProcessSearchCoordinator::markFrameAsSearched):
(WebCore::AXCrossProcessSearchCoordinator::checkCompletion):
* Source/WebCore/accessibility/AXLiveRegionManager.h:
(): Deleted.
* Source/WebCore/accessibility/AXLogger.cpp:
(WebCore::operator&lt;&lt;):
(WebCore::streamAXCoreObject):
(WebCore::AXObjectCache::getOrCreateSlow):
* Source/WebCore/accessibility/AXObjectCache.h:
(WebCore::AccessibilityRemoteToken::AccessibilityRemoteToken): Deleted.
* Source/WebCore/accessibility/AXRemoteFrame.h:
* Source/WebCore/accessibility/AXSearchManager.cpp:
(WebCore::AXSearchManager::findMatchingObjectsAsStream):
(WebCore::AXSearchManager::findMatchingObjectsInternalAsStream):
(WebCore::AXSearchManager::findMatchingRange):
(WebCore::AXSearchManager::matchWithResultsLimit): Deleted.
(WebCore::AXSearchManager::findMatchingObjectsInternal): Deleted.
* Source/WebCore/accessibility/AXSearchManager.h:
(WebCore::AccessibilitySearchCriteriaIPC::AccessibilitySearchCriteriaIPC):
(WebCore::AccessibilitySearchCriteriaIPC::toSearchCriteria const):
(WebCore::SearchResultEntry::localResult):
(WebCore::SearchResultEntry::remoteFrame):
(WebCore::SearchResultEntry::isLocalResult const):
(WebCore::SearchResultEntry::isRemoteFrame const):
(WebCore::SearchResultEntry::objectIfLocalResult const):
(WebCore::SearchResultEntry::frameID const):
(WebCore::SearchResultEntry::streamIndex const):
(WebCore::SearchResultEntry::SearchResultEntry):
(WebCore::AccessibilitySearchResultStream::appendLocalResult):
(WebCore::AccessibilitySearchResultStream::appendRemoteFrame):
(WebCore::AccessibilitySearchResultStream::entries const):
(WebCore::AccessibilitySearchResultStream::entryCount const):
(WebCore::AccessibilitySearchResultStream::setResultsLimit):
(WebCore::AccessibilitySearchResultStream::resultsLimit const):
(WebCore::AccessibilitySearchResultStream::nextIndex const):
(WebCore::AccessibilitySearchResult::local):
(WebCore::AccessibilitySearchResult::remote):
(WebCore::AccessibilitySearchResult::isLocal const):
(WebCore::AccessibilitySearchResult::isRemote const):
(WebCore::AccessibilitySearchResult::objectIfLocalResult const):
(WebCore::AccessibilitySearchResult::remoteToken const):
(WebCore::AccessibilitySearchResult::AccessibilitySearchResult):
(WebCore::AXSearchManager::findMatchingObjects): Deleted.
* Source/WebCore/accessibility/AXStitchUtilities.h:
* Source/WebCore/accessibility/AccessibilityMockObject.h:
* Source/WebCore/accessibility/AccessibilityNodeObject.cpp:
* Source/WebCore/accessibility/AccessibilityObject.cpp:
(WebCore::AccessibilityObject::findMatchingObjectsWithin):
(WebCore::AccessibilityObject::findMatchingObjects): Deleted.
* Source/WebCore/accessibility/AccessibilityObject.h:
* Source/WebCore/accessibility/AccessibilityRemoteToken.h: Added.
(WebCore::AccessibilityRemoteToken::AccessibilityRemoteToken):
* Source/WebCore/accessibility/AccessibilityScrollView.cpp:
(WebCore::AccessibilityScrollView::detachRemoteParts):
* Source/WebCore/accessibility/AccessibilityScrollView.h:
* Source/WebCore/accessibility/AccessibilitySearchCriteriaIPC.h: Added.
* Source/WebCore/accessibility/ios/WebAccessibilityObjectWrapperIOS.mm:
(-[WebAccessibilityObjectWrapper accessibilityFindMatchingObjects:]):
(WebCore::AXIsolatedObject::findMatchingObjects): Deleted.
* Source/WebCore/accessibility/isolatedtree/AXIsolatedObject.h:
(WebCore::AXIsolatedTree::applyPendingChangesLocked):
* Source/WebCore/accessibility/isolatedtree/mac/AXIsolatedObjectMac.mm:
(WebCore::appendPlatformProperties):
(remoteElementFromToken):
(searchResultsToNSArray):
(performSearchWithRemoteFrames):
(-[WebAccessibilityObjectWrapper _accessibilityHitTestResolvingRemoteFrame:callback:]):
(-[WebAccessibilityObjectWrapper accessibilityAttributeValue:forParameter:]):
* Source/WebCore/page/ChromeClient.cpp:
(WebCore::ChromeClient::performAccessibilitySearchInRemoteFrame):
(WebCore::ChromeClient::continueAccessibilitySearchFromChildFrame):
* Source/WebCore/page/FrameTree.h:
(headers_for_type):
(WebKit::WebPageProxy::performAccessibilitySearchInRemoteFrame):
(WebKit::WebPageProxy::continueAccessibilitySearchFromChildFrame):
(WebKit::WebChromeClient::performAccessibilitySearchInRemoteFrame):
(WebKit::WebChromeClient::continueAccessibilitySearchFromChildFrame):
(WebKit::tokenFromWrapper):
(WebKit::convertSearchResultsToRemoteTokens):
(WebKit::WebPage::performAccessibilitySearchInRemoteFrame):
(WebKit::WebPage::continueAccessibilitySearchInParentFrame):
* Tools/WebKitTestRunner/InjectedBundle/AccessibilityUIElement.cpp:
(WTR::AccessibilityUIElement::uiElementsForSearchPredicate):
* Tools/WebKitTestRunner/InjectedBundle/AccessibilityUIElement.h:
* Tools/WebKitTestRunner/InjectedBundle/Bindings/AccessibilityUIElement.idl:
* Tools/WebKitTestRunner/InjectedBundle/atspi/AccessibilityUIElementAtspi.cpp:
(WTR::AccessibilityUIElementAtspi::uiElementsForSearchPredicate):
* Tools/WebKitTestRunner/InjectedBundle/atspi/AccessibilityUIElementAtspi.h:
* Tools/WebKitTestRunner/InjectedBundle/ios/AccessibilityUIElementIOS.h:
* Tools/WebKitTestRunner/InjectedBundle/ios/AccessibilityUIElementIOS.mm:
(WTR::AccessibilityUIElementIOS::uiElementsForSearchPredicate):
* Tools/WebKitTestRunner/InjectedBundle/mac/AccessibilityUIElementMac.h:
* Tools/WebKitTestRunner/InjectedBundle/mac/AccessibilityUIElementMac.mm:
(WTR::AccessibilityUIElementMac::attributeValue const):
(WTR::AccessibilityUIElementMac::allAttributes):
(WTR::AccessibilityUIElementMac::setBoolAttributeValue):
(WTR::AccessibilityUIElementMac::setValue):
(WTR::AccessibilityUIElementMac::isAttributeSupported):
(WTR::AccessibilityUIElementMac::isValid const):
(WTR::AccessibilityUIElementMac::uiElementsForSearchPredicate):
(WTR::AccessibilityUIElementMac::setSelectedTextRange):
(WTR::AccessibilityUIElementMac::invokeCustomActionAtIndex):
(WTR::AccessibilityUIElementMac::setSelectedTextMarkerRange):
(WTR::AccessibilityUIElementMac::setSelectedChild const):
(WTR::AccessibilityUIElementMac::setSelectedChildAtIndex const):
(WTR::AccessibilityUIElementMac::removeSelectionAtIndex const):
(WTR::AccessibilityUIElementMac::takeFocus):
(WTR::AccessibilityUIElementMac::resetSelectedTextMarkerRange):
* Tools/WebKitTestRunner/InjectedBundle/playstation/AccessibilityUIElementPlayStation.cpp:
(WTR::AccessibilityUIElementPlayStation::uiElementsForSearchPredicate):
* Tools/WebKitTestRunner/InjectedBundle/playstation/AccessibilityUIElementPlayStation.h:
* Tools/WebKitTestRunner/InjectedBundle/win/AccessibilityUIElementWin.cpp:
(WTR::AccessibilityUIElementWin::uiElementsForSearchPredicate):
* Tools/WebKitTestRunner/InjectedBundle/win/AccessibilityUIElementWin.h:

Canonical link: <a href="https://commits.webkit.org/308321@main">https://commits.webkit.org/308321@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/2562f373e83cde3ef157b3f5f237d00cae2035af

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/147070 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/19751 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/13341 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/155752 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/100484 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/148944 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/20209 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/19651 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/113336 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/80862 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/150032 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/15582 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/132128 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94092 "Passed tests") | | ⏳ 🛠 vision-apple 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/146394 "Passed tests") | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/14787 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/12554 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/3194 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/124374 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/10049 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/158083 "Built successfully") | | 
| | [❌ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/1214 "Found 1 new failure in accessibility/AccessibilityRenderObject.cpp") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/11489 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/121360 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/19552 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/16408 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/121561 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/31151 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/19561 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/131806 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/75520 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/17122 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/8624 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/19167 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/82922 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/18897 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/19048 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/18956 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/accessibility/AXCoreObject.h
+++ b/Source/WebCore/accessibility/AXCoreObject.h
@@ -32,6 +32,7 @@
 #include <WebCore/CharacterRange.h>
 #include <WebCore/Color.h>
 #include <WebCore/ColorConversion.h>
+#include <WebCore/FrameIdentifier.h>
 #include <WebCore/HTMLTextFormControlElement.h>
 #include <WebCore/InputType.h>
 #include <WebCore/LayoutRect.h>
@@ -144,7 +145,7 @@ enum class AccessibilityDetachmentType { CacheDestroyed, ElementDestroyed, Eleme
 enum class AccessibilityConversionSpace { Screen, Page };
 
 // FIXME: This should be replaced by AXDirection (or vice versa).
-enum class AccessibilitySearchDirection {
+enum class AccessibilitySearchDirection : uint8_t {
     Next = 1,
     Previous,
 };
@@ -634,7 +635,8 @@ public:
     bool isFrame() const;
 #if PLATFORM(COCOA)
     virtual RetainPtr<id> remoteFramePlatformElement() const = 0;
-    virtual pid_t remoteFrameProcessIdentifier() const = 0;
+    virtual pid_t remoteFramePID() const = 0;
+    virtual std::optional<FrameIdentifier> remoteFrameID() const = 0;
 #endif
     virtual bool hasRemoteFrameChild() const = 0;
 
@@ -856,7 +858,8 @@ public:
     AXCoreObject* parentObjectIncludingCrossFrame() const;
     AXCoreObject* parentObjectUnignoredIncludingCrossFrame() const;
 
-    virtual AccessibilityChildrenVector findMatchingObjects(AccessibilitySearchCriteria&&) = 0;
+    // Finds objects within |this| object matching the given search criteria.
+    virtual AccessibilityChildrenVector findMatchingObjectsWithin(AccessibilitySearchCriteria&&);
     virtual bool isDescendantOfRole(AccessibilityRole) const = 0;
     AXCoreObject* selfOrFirstTextDescendant();
 
--- a/Source/WebCore/accessibility/AXCrossProcessSearch.cpp
+++ b/Source/WebCore/accessibility/AXCrossProcessSearch.cpp
@@ -1,0 +1,452 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1.  Redistributions of source code must retain the above copyright
+ *     notice, this list of conditions and the following disclaimer.
+ * 2.  Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "AXCrossProcessSearch.h"
+
+#include <WebCore/AXCoreObject.h>
+#include <WebCore/AXObjectCache.h>
+#include <WebCore/AXTreeStoreInlines.h>
+#include <WebCore/Chrome.h>
+#include <WebCore/ChromeClient.h>
+#include <WebCore/LocalFrame.h>
+#include <WebCore/Page.h>
+#include <wtf/MainThread.h>
+#include <wtf/MonotonicTime.h>
+#include <wtf/RefCounted.h>
+#include <wtf/StdLibExtras.h>
+#include <wtf/threads/BinarySemaphore.h>
+
+#if PLATFORM(COCOA)
+#include <CoreFoundation/CFRunLoop.h>
+#endif
+
+#if PLATFORM(MAC)
+#define PLATFORM_SUPPORTS_REMOTE_SEARCH 1
+#else
+#define PLATFORM_SUPPORTS_REMOTE_SEARCH 0
+#endif
+
+namespace WebCore {
+
+static bool canDoRemoteSearch(const std::optional<AXTreeID>& treeID)
+{
+#if PLATFORM_SUPPORTS_REMOTE_SEARCH
+    return treeID.has_value();
+#else
+    UNUSED_PARAM(treeID);
+    return false;
+#endif // PLATFORM_SUPPORTS_REMOTE_SEARCH
+}
+
+// Spins the run loop on the main thread while waiting for a condition to become true.
+// In the future, we could consider changing callers to implement a solution that doesn't
+// require polling as done in this function, since polling can be inefficient.
+template<typename Predicate>
+static DidTimeout spinRunLoopUntil(Predicate&& isComplete, Seconds timeout)
+{
+    AX_ASSERT(isMainThread());
+
+    auto deadline = MonotonicTime::now() + timeout;
+    while (MonotonicTime::now() < deadline) {
+        if (isComplete())
+            return DidTimeout::No;
+#if PLATFORM(COCOA)
+        CFRunLoopRunInMode(kCFRunLoopDefaultMode, 0.02, true);
+#else
+        Thread::yield();
+#endif
+    }
+    return isComplete() ? DidTimeout::No : DidTimeout::Yes;
+}
+
+DidTimeout AXCrossProcessSearchCoordinator::waitWithTimeout(Seconds timeout)
+{
+    auto isComplete = [this] {
+        return m_searchComplete.load(std::memory_order_acquire) && !m_pendingCount.load(std::memory_order_acquire);
+    };
+
+    // If search is already complete with no pending requests, return immediately.
+    if (isComplete())
+        return DidTimeout::No;
+
+    if (isMainThread()) {
+        // On the main thread, we can't block on a semaphore because IPC callbacks
+        // need to run on the main thread. Instead, spin the run loop.
+        return spinRunLoopUntil(isComplete, timeout);
+    }
+
+    // On background threads (e.g., the accessibility thread), we can safely
+    // block on the semaphore.
+    return m_semaphore.waitFor(timeout) ? DidTimeout::No : DidTimeout::Yes;
+}
+
+// Helper to merge stream entries into AccessibilitySearchResults.
+// If coordinator is provided, also pulls in remote results for RemoteFrame entries.
+static AccessibilitySearchResults mergeStreamResults(const Vector<SearchResultEntry>& entries, unsigned limit, AXCrossProcessSearchCoordinator* coordinator)
+{
+    AccessibilitySearchResults results;
+    for (const auto& entry : entries) {
+        if (results.size() >= limit)
+            break;
+
+        if (RefPtr object = entry.objectIfLocalResult())
+            results.append(AccessibilitySearchResult::local(object.releaseNonNull()));
+        else if (coordinator) {
+            // The search result was from an AXRemoteFrame we contain. Pull
+            // AccessibilityRemoteTokens from the search coordinator and
+            // convert them into results.
+            auto tokens = coordinator->takeRemoteResults(entry.streamIndex());
+            for (auto& token : tokens) {
+                if (results.size() >= limit)
+                    break;
+                results.append(AccessibilitySearchResult::remote(WTF::move(token)));
+            }
+        }
+    }
+    return results;
+}
+
+#if PLATFORM_SUPPORTS_REMOTE_SEARCH
+// Computes remaining timeout from an absolute deadline, accounting for IPC overhead.
+// Returns std::nullopt if the deadline has already passed (so callers can skip the search).
+// Returns at least crossProcessSearchMinimumTimeout to ensure deeply nested frames
+// always get some time to search.
+static std::optional<Seconds> computeRemainingTimeout(std::optional<MonotonicTime> deadline)
+{
+    if (!deadline)
+        return crossProcessSearchTimeout;
+
+    auto remaining = *deadline - MonotonicTime::now() - crossProcessSearchIPCOverhead;
+    if (remaining <= 1_ms)
+        return std::nullopt;
+    return std::max(crossProcessSearchMinimumTimeout, remaining);
+}
+
+// Dispatches an IPC request to search a remote frame.
+// The coordinator's responseReceived() will be called when the response arrives (or on failure).
+static void dispatchRemoteFrameSearch(Ref<AXCrossProcessSearchCoordinator> coordinator, FrameIdentifier frameID, AccessibilitySearchCriteriaIPC criteria, size_t streamIndex, AXTreeID treeID)
+{
+    ensureOnMainThread([coordinator = WTF::move(coordinator), frameID, criteria = WTF::move(criteria), streamIndex, treeID]() mutable {
+        AX_ASSERT(isMainThread());
+
+        WeakPtr cache = AXTreeStore<AXObjectCache>::axObjectCacheForID(treeID);
+        RefPtr page = cache ? cache->page() : nullptr;
+        if (!page) {
+            coordinator->responseReceived();
+            return;
+        }
+
+        page->chrome().client().performAccessibilitySearchInRemoteFrame(frameID, criteria,
+            [coordinator = WTF::move(coordinator), streamIndex](Vector<AccessibilityRemoteToken>&& tokens) mutable {
+                coordinator->storeRemoteResults(streamIndex, WTF::move(tokens));
+                coordinator->responseReceived();
+            });
+    });
+}
+#endif // PLATFORM_SUPPORTS_REMOTE_SEARCH
+
+AccessibilitySearchResults performCrossProcessSearch(AccessibilitySearchResultStream&& stream, const AccessibilitySearchCriteriaIPC& criteriaForIPC, std::optional<AXTreeID> treeID, unsigned originalLimit, std::optional<FrameIdentifier> requestingFrameID)
+{
+    if (!canDoRemoteSearch(treeID)) {
+        UNUSED_PARAM(criteriaForIPC);
+        UNUSED_PARAM(requestingFrameID);
+        return mergeStreamResults(stream.entries(), originalLimit, nullptr);
+    }
+
+#if PLATFORM_SUPPORTS_REMOTE_SEARCH
+    // Calculate how many results to request from each remote frame.
+    // We need to account for local results that precede each remote frame in tree order.
+    Vector<std::pair<const SearchResultEntry*, unsigned>> remoteFrameRequests;
+    unsigned localCountSoFar = 0;
+    for (const auto& entry : stream.entries()) {
+        if (entry.isLocalResult())
+            ++localCountSoFar;
+        else {
+            // For this remote frame, request enough to potentially fill remaining quota.
+            unsigned remaining = originalLimit > localCountSoFar ? originalLimit - localCountSoFar : 0;
+            // If we've already filled our quota with local results before this remote frame,
+            // we don't need to query it.
+            if (remaining > 0)
+                remoteFrameRequests.append({ &entry, remaining });
+        }
+    }
+
+    if (remoteFrameRequests.isEmpty()) {
+        // All remote frames were skipped because local results filled the quota.
+        return mergeStreamResults(stream.entries(), originalLimit, nullptr);
+    }
+
+    // We have remote frames to query. Create a coordinator for synchronization.
+    Ref coordinator = AXCrossProcessSearchCoordinator::create();
+
+    if (requestingFrameID) {
+        // Pre-populate with requesting frame to prevent re-searching it.
+        coordinator->markFrameAsSearched(*requestingFrameID);
+    }
+
+    // Dispatch IPC for each remote frame.
+    for (const auto& [entry, maxResults] : remoteFrameRequests) {
+        if (!entry->frameID()) {
+            // No frame ID, nothing to dispatch.
+            continue;
+        }
+
+        // Skip frames we've already searched.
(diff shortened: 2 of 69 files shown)


Source/WebCore/accessibility/AccessibilityRemoteToken.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <wtf/UUID.h>
+#include <wtf/Vector.h>
+
+namespace WebCore {
+
+// Token used to identify accessibility objects across process boundaries.
+// On Mac, this is an opaque byte array from NSAccessibilityRemoteUIElement.
+// On other platforms, it's a UUID and process ID pair.
+struct AccessibilityRemoteToken {
```
minorninth, line 35: Thanks for abstracting this!

Source/WebCore/accessibility/AXCoreObject.h
```diff
@@ -144,7 +145,7 @@ enum class AccessibilityDetachmentType { CacheDestroyed, ElementDestroyed, Eleme
 enum class AccessibilityConversionSpace { Screen, Page };
 
 // FIXME: This should be replaced by AXDirection (or vice versa).
-enum class AccessibilitySearchDirection {
+enum class AccessibilitySearchDirection : uint8_t {
```
minorninth, line 148: FYI: This might make sense if AccessibilitySearchDirection is in a packed struct and saves a significant number of bytes, but could actually hurt performance in other cases (adds a truncation/masking instruction every time a value that's not already uint8 is assigned).

Source/WebCore/accessibility/AXCrossProcessSearch.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+#if PLATFORM(COCOA)
+#include <CoreFoundation/CFRunLoop.h>
+#endif
+
+namespace WebCore {
+
+// Spins the run loop on the main thread while waiting for a condition to become true.
+template<typename Predicate>
+static DidTimeout spinRunLoopUntil(Predicate&& isComplete, Seconds timeout)
```
minorninth, line 67: This seems totally reasonable for now, but should we add a comment or file a radar about updating this to exit the run loop immediately when the condition is hit, enabling using a longer timeout rather than polling?

Source/WebCore/accessibility/AXCrossProcessSearch.cpp
```diff
@@ -0,0 +1,436 @@
...
+
+        if (RefPtr object = entry.objectIfLocalResult())
+            results.append(AccessibilitySearchResult::local(object.releaseNonNull()));
+        else if (coordinator) {
+            // The search result was from an AXRemoteFrame we contain. Pull
+            // AccessibilityRemoteTokens from the search coordinator and
+            // convert them into results.
+            auto tokens = coordinator->takeRemoteResults(entry.streamIndex());
+            for (auto& token : tokens) {
+                if (results.size() >= limit)
```
minorninth, line 122: This feels weirdly asymmetric to be checking the size at the top of the outer for loop, and also at the top of the inner for loop.

What about instead, just check after each append:

```
for (const auto& entry : entries) {
    if (RefPtr object = entry.objectIfLocalResult()) {
      results.append(...);
      if (results.size() >= limit)
        break;
    } else if (coordinator) {
      ...
      for (auto& token : tokens) {
        results.append(...);
        if (results.size() >= limit)
          break;
      }
    }

```

Source/WebCore/accessibility/AXCrossProcessSearch.cpp
```diff
@@ -0,0 +1,436 @@
...
+// Computes remaining timeout from an absolute deadline, accounting for IPC overhead.
+// Returns at least crossProcessSearchMinimumTimeout to ensure deeply nested frames
+// always get some time to search.
+static Seconds computeRemainingTimeout(std::optional<MonotonicTime> deadline)
+{
+    if (!deadline)
+        return crossProcessSearchTimeout;
+
+    auto remaining = *deadline - MonotonicTime::now() - crossProcessSearchIPCOverhead;
+    return std::max(crossProcessSearchMinimumTimeout, remaining);
```
minorninth, line 144: Let's say a process receives the IPC super late, after the query timed out. We should probably just skip the search in that case, otherwise when a process finally becomes unblocked it could waste time servicing a long queue of expired queries.

Source/WebCore/accessibility/AXCrossProcessSearch.cpp
```diff
@@ -0,0 +1,436 @@
...
+    auto criteriaForIPC = AccessibilitySearchCriteriaIPC(criteria);
+    std::optional treeID = anchorObject.treeID();
+
+    // If no deadline has been set, set one now. This establishes the timeout budget
+    // for the entire search tree, ensuring nested frames share the same deadline.
+    if (!criteriaForIPC.deadline)
+        criteriaForIPC.deadline = MonotonicTime::now() + crossProcessSearchTimeout;
+
+    if (!treeID) {
+        // No tree ID means we can't coordinate with remote frames.
```
minorninth: Would it be worth merging this with the non-Mac case? Like basically have a canDoRemoteSearch flag that's always false on non-Mac, but if there's no treeID, Mac could set that flag - then have a single fallback block.

Another idea is that for debugging we might want to have a flag to disable remote search.

Source/WebCore/accessibility/AXCrossProcessSearch.cpp
```diff
@@ -0,0 +1,436 @@
...
+        if (mergedResults.size() + localCount >= limit)
+            break;
+        mergedResults.append(AccessibilitySearchResult::remote(WTF::move(token)));
+    }
+    mergedResults.appendVector(WTF::move(localResults));
+    return mergedResults;
+}
+
+#if PLATFORM(MAC)
+// Ref-counted class for safe parent search coordination across threads.
```
minorninth: Does Parent here mean "Parent Frame"? Or does it mean the caller of the IPC? I get the issue with preventing UAF if the callback comes after the timeout, but it's not clear what "parent" here refers to, it feels like it should actually relate to whoever initiated the search.

Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
```diff
@@ -605,6 +611,100 @@ String plainTextForDisplay(const std::optional<SimpleRange>& range)
...
+
+static Vector<WebCore::AccessibilityRemoteToken> convertSearchResultsToRemoteTokens(WebCore::AccessibilitySearchResults&& searchResults)
+{
+    Vector<WebCore::AccessibilityRemoteToken> results;
+    for (auto& result : searchResults) {
+        if (auto* object = result.objectIfLocalResult()) {
+            if (auto token = tokenFromWrapper(protect(object->wrapper())))
+                results.append(WTF::move(*token));
+        } else if (result.isRemote())
+            results.append(*result.remoteToken());
```
minorninth, line 634: One thought: it probably wouldn't be that wasteful to store a tiny bit more info here other than the remote token - you could also return the role of the result, for example, enabling tests to provide more info than just a bunch of remote tokens.

minorninth (commented):
Great work! A few ideas and questions, but no architectural concerns, this looks like a very comprehensive design

Source/WebCore/accessibility/AXCoreObject.h
twilco: Totally makes sense! The motivation for doing this is because we are now sending this over IPC between web content processes, and having an explicitly defined size is required for that.

Source/WebCore/accessibility/AXCrossProcessSearch.cpp
twilco: Your suggestion would break out of the inner for loop, but not the outer one, which is why the code is structured as-is. Happy to adjust though if you have another idea!

Source/WebCore/accessibility/AXCrossProcessSearch.cpp
twilco: Added a comment!

Source/WebCore/accessibility/AXCrossProcessSearch.cpp
twilco: Good point! Will fix in latest commit.

Source/WebCore/accessibility/AXCrossProcessSearch.cpp
twilco: It does mean "Parent frame" — it's the context we send to the parent frame via IPC to continue the search from where we, a child frame, are. I will rename this to ParentFrameSearchContext since maybe that could help avoid confusion?

twilco: Also improved the code comment to help better describe what's happening here.

Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
twilco: I explored this. I agree that it wouldn't be overly wasteful or costly to do it, but the mechanics of piping the information through ended up being a bit complicated. I think a better option may just be porting these tests to be client tests in a later PR, where we can get the role for free.

Source/WebCore/accessibility/AXCrossProcessSearch.cpp
twilco: Fixed!

hoffmanjoshua (approved):

PR #58854, opened 2026-02-17
Use tri-state mayBlockNetworkRequest for controlling blocking latch mode in Network Process

sysrqb:
#### ff0faf2c4d52aae405395b052b7c1472333ce48d
<pre>
Use tri-state mayBlockNetworkRequest for controlling blocking latch mode in Network Process
<a href="https://bugs.webkit.org/show_bug.cgi?id=308052">https://bugs.webkit.org/show_bug.cgi?id=308052</a>
<a href="https://rdar.apple.com/170551008">rdar://170551008</a>

Reviewed by Charlie Wolfe.

In 307413@main, the blocking behavior that happens in the Network Process may
still happen when the associated feature flag is disabled. This patch changes
the NetworkResourceLoadParameter to be a std::optional and let&apos;s us
differentiate between when the feature is disabled and when the domain is not
present on the fingerprinting script list (but the feature is still enabled).

Covered by existing tests.

* Source/WebKit/NetworkProcess/NetworkLoadChecker.cpp:
(WebKit::NetworkLoadChecker::shouldBlockForTrackingPolicy):
* Source/WebKit/NetworkProcess/NetworkResourceLoadParameters.h:
* Source/WebKit/NetworkProcess/NetworkResourceLoadParameters.serialization.in:
* Source/WebKit/WebProcess/Network/WebLoaderStrategy.cpp:
(WebKit::addParametersShared):

Canonical link: <a href="https://commits.webkit.org/307760@main">https://commits.webkit.org/307760@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/156c6fb1b5cd73a4f3c44883e88ef249934fec95

| Misc | iOS, visionOS, tvOS & watchOS  | macOS  | Linux |  Windows | Apple Internal |
| ----- | ---------------------- | ------- |  ----- |  --------- | ------ |
| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/145419 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/18101 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/9914 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/154091 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/99056 "Built successfully") | ⏳ 🛠 ios-apple 
[...]

PR #58856, opened 2026-02-17
[webkitcorepy] Add support for dynamic configuration files

JonWBedard:
#### 49542a2a2a2e3b5f5c2730dbe420f4ffcfd9fc2b
<pre>
[webkitcorepy] Add support for dynamic configuration files
<a href="https://bugs.webkit.org/show_bug.cgi?id=308058">https://bugs.webkit.org/show_bug.cgi?id=308058</a>
<a href="https://rdar.apple.com/problem/170556228">rdar://problem/170556228</a>

Reviewed by NOBODY (OOPS!).

WIP (OOPS!).
</pre>
https://github.com/WebKit/WebKit/commit/49542a2a2a2e3b5f5c2730dbe420f4ffcfd9fc2b

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/147353 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/20038 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/13629 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/156035 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/100768 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/149226 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/20494 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/19938 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/113570 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/81000 "Exiting early after 60 failures. 15458 tests run. 60 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/282ccda9-3809-486f-b6c1-7c972eb9f41e) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/150315 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/15783 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/132349 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94329 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/c7060feb-849e-463a-afe8-e138caca80ec) 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/146675 "Passed tests") | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/14964 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/12752 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/3476 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/124559 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/10285 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/158367 "Built successfully") | | 
| | | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/11733 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/121598 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/19837 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/16641 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/121797 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/19848 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/132042 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/75826 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/17325 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/8829 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/19452 "Built successfully") | | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/19182 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/19333 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/19240 "Built successfully") | | | |


Tools/Scripts/libraries/webkitconfigpy/webkitconfigpy/config.py
```diff
@@ -0,0 +1,248 @@
...
+    def __ne__(self, other):
+        result = self.__eq__(other)
+        if result is NotImplemented:
+            return result
+        return not result
+
+    def keys(self):
+        if isinstance(self.data, dict):
+            return self.data.keys()
+        raise AttributeError("'Config' object has no attribute 'keys' (data is not a dict)")
```
gsnedders, line 303: It would be much better (from a type safety point of view) if we actually returned different types, and then a type checker would know if you could call `keys` or not.

It's not even clear to me we gain anything from returning a `Config` object rather than just the bare object? All we're doing is carrying `mode` with the object, but I'm a bit dubious about how much value that provides? I'd expect typically the relevant code knows whether it wants to re-serialize it as JSON or YAML?

Tools/Scripts/libraries/webkitconfigpy/webkitconfigpy/config.py
```diff
@@ -0,0 +1,248 @@
...
+            raise NotImplemented
+
+    def __eq__(self, other):
+        if isinstance(other, Config):
+            return self.data == other.data
+        elif isinstance(other, (dict, list)):
+            return self.data == other
+        return NotImplemented
+
+    def __ne__(self, other):
```
gsnedders, line 294: There's basically never any reason to define an `__ne__` like this in Python 3 — because there's a default implementation which does this already.

Tools/Scripts/libraries/webkitconfigpy/webkitconfigpy/config.py
```diff
@@ -0,0 +1,248 @@
...
+    INDENT = 4
+    IMPLICIT_VARIABLE_RE = re.compile(r'^[a-zA-Z][a-zA-Z0-9_-]*$')
+
+    class Mode(Enum):
+        JSON = 'json'
+        YAML = 'yaml'
+
+    @classmethod
+    def loads(cls, string, mode=None):
+        if mode == cls.Mode.YAML or (mode is None and not string.strip().startswith(('{', "["))):
```
gsnedders, line 44: Given YAML is a superset of JSON, this isn't really a guarantee at all.

We should just define one of the modes to be the default, rather than trying to do some sort of magic.

quanghuynh10111-png (commented):
Fixes #58854

Tools/Scripts/libraries/webkitconfigpy/webkitconfigpy/config.py
```diff
@@ -0,0 +1,248 @@
...
+class Config(object):
+    INDENT = 4
+    IMPLICIT_VARIABLE_RE = re.compile(r'^[a-zA-Z][a-zA-Z0-9_-]*$')
+
+    class Mode(Enum):
+        JSON = 'json'
+        YAML = 'yaml'
+
+    @classmethod
+    def loads(cls, string, mode=None):
```
gsnedders, line 43: I'd _love_ if we could stop writing/landing large amounts of code without type annotations.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,248 @@
...
+from unittest import mock
+
+
+class Config(object):
+    INDENT = 4
+    IMPLICIT_VARIABLE_RE = re.compile(r'^[a-zA-Z][a-zA-Z0-9_-]*$')
+
+    class Mode(Enum):
+        JSON = 'json'
+        YAML = 'yaml'
```
gsnedders: Do we actually want the enum to have these values? I would've expected that the value would be opaque, so we could just use `enum.auto()`?

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,248 @@
...
+from enum import Enum
+from typing import Any, Dict, Optional
+from unittest import mock
+
+
+class Config(object):
+    INDENT = 4
+    IMPLICIT_VARIABLE_RE = re.compile(r'^[a-zA-Z][a-zA-Z0-9_-]*$')
+
+    class Mode(Enum):
```
gsnedders: Having `webkitcorepy.config.Config.Mode` seems a bit redundant; isn't it implied to be the mode of the config from being in the module? I'd just move it to `webkitcorepy.config.Mode`.

Tools/Scripts/libraries/webkitconfigpy/webkitconfigpy/config.py
```diff
@@ -0,0 +1,248 @@
...
+import json
+import os
+import re
+
+from enum import Enum
+from typing import Any, Dict, Optional
+from unittest import mock
+
+
+class Config(object):
```
gsnedders, line 34: `(object)` is redundant in Python 3:

```suggestion
class Config:
```

Tools/Scripts/libraries/webkitconfigpy/webkitconfigpy/config.py
```diff
@@ -0,0 +1,248 @@
...
+
+    class Mode(Enum):
+        JSON = 'json'
+        YAML = 'yaml'
+
+    @classmethod
+    def loads(cls, string, mode=None):
+        if mode == cls.Mode.YAML or (mode is None and not string.strip().startswith(('{', "["))):
+            import yaml
+            data = yaml.safe_load(string) or {}
```
gsnedders, line 46: Why do we have `or {}` here?

Tools/Scripts/libraries/webkitconfigpy/webkitconfigpy/config.py
```diff
@@ -0,0 +1,248 @@
...
+            return cls(data, mode=cls.Mode.YAML)
+        else:
+            data = json.loads(string)
+            return cls(data, mode=cls.Mode.JSON)
+
+    @classmethod
+    def load(cls, file, mode=None):
+        if isinstance(file, str):
```
gsnedders, line 54: Can we keep the interface closer to most `load` class methods in the stdlib, and require a file-like object?

Tools/Scripts/libraries/webkitconfigpy/webkitconfigpy/config.py
```diff
@@ -0,0 +1,248 @@
...
+                        mode = cls.Mode.YAML
+
+                if mode is None:
+                    raise ValueError("mode parameter required when loading from file object without inferrable extension")
+
+            if mode == cls.Mode.JSON:
+                data = json.load(file)
+            else:
+                import yaml
+                data = yaml.safe_load(file) or {}
```
gsnedders, line 90: Again, why the `or {}`?

Tools/Scripts/libraries/webkitconfigpy/webkitconfigpy/config.py
```diff
@@ -0,0 +1,248 @@
...
+
+        return cls(data, mode=mode)
+
+    @classmethod
+    def _renderValue(cls, template, context):
```
gsnedders, line 165: Per prior offline discussion, I'd _really_ like to see some example of why you believe that overriding this is worthwhile versus using `$let` and similar before I'm willing to r+ this.

Tools/Scripts/libraries/webkitconfigpy/webkitconfigpy/config.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def __init__(self, data=None, mode=None, context=None):
+        super().__init__()
+
+        data = data or {}
+        mode = mode or Config.Mode.JSON
+        context = context or {}
+
+        if not isinstance(data, (dict, list)):
+            raise TypeError(f"data must be dict or list, not {type(data).__name__}")
```
gsnedders, line 244: But that isn't true if it's come from JSON or YAML?

Tools/Scripts/libraries/webkitcorepy/render-config
```diff
@@ -0,0 +1,70 @@
...
+        if sys.stdin.isatty():
+            print("Error: No input provided. Specify a file or pipe content to stdin.", file=sys.stderr)
+            return 1
+        content = sys.stdin.read()
+        config = Config.loads(content)
+    else:
+        config = Config.load(args.config_file)
+
+    if args.format:
+        config.mode = dict(json=Config.Mode.JSON, yaml=Config.Mode.YAML)[args.format]
```
gsnedders: Can we please not use `dict` for things like this? But also, see my comment about `--format`.

Tools/Scripts/libraries/webkitconfigpy/render-config
```diff
@@ -0,0 +1,70 @@
...
+        'config_file',
+        nargs='?',
+        default=None,
+        help='Path to JSON or YAML configuration file (or read from stdin if omitted)'
+    )
+    parser.add_argument(
+        '--format',
+        choices=['json', 'yaml'],
+        help='Output format (input format is auto-detected)'
+    )
```
gsnedders, line 52: As mentioned below, we can avoid having to do any post-processing here if we just specify an appropriate `type` argument:

```suggestion
    parser.add_argument(
        '--format',
        choices=[mode.name.lower() for mode in Config.Mode],
        type=lambda v: Config.Mode(v.upper()),
        help='Output format (input format is auto-detected)'
    )
```

Tools/Scripts/libraries/webkitconfigpy/render-config
```diff
@@ -0,0 +1,70 @@
...
+
+libraries = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+AutoInstall.set_directory(os.path.join(libraries, 'autoinstalled', 'python-{}-{}'.format(sys.version_info[0], platform.machine())))
+
+def main(*args, **kwargs):
+    parser = argparse.ArgumentParser(description='Render a templated JSON or YAMl config')
+    parser.add_argument(
+        'config_file',
+        nargs='?',
+        default=None,
```
gsnedders, line 43:
```suggestion
        default='-',
```

Tools/Scripts/libraries/webkitconfigpy/webkitconfigpy/config.py
```diff
@@ -0,0 +1,248 @@
...
+        return f"Config({self.data!r}, mode={self.mode!r})"
+
+    def __str__(self):
+        if self.mode == Config.Mode.JSON:
+            return json.dumps(self.data, indent=self.INDENT, ensure_ascii=False)
+        elif self.mode == Config.Mode.YAML:
+            import yaml
+            return yaml.dump(self.data, default_flow_style=False, allow_unicode=True, indent=self.INDENT)
+        else:
+            raise NotImplemented
```
gsnedders, line 285: This feels a bit surprising as a `__str__`? Like I can't disagree it's a string representation of the object, but it's not where I'd first look if I were trying to re-serialize the object?

gsnedders (changes requested):
This is a lot of comments, but in short:

 * I still want to see a justifying example for the local_context stuff, rather than using `$let`. Effectively forking a library, especially when we don't have a trivial way to run all the upstream tests against our fork, seems very vulnerable to subtle later breakage.
 * Can we return different objects depending on what the data is, because then type-checking etc. all becomes _a lot_ simpler, and avoids a lot of things where we currently rely on runtime checks?
 * Do we actually want a single class owning parsing and serialising, and owning all of this regardless of the input file type?

Tools/Scripts/libraries/webkitconfigpy/webkitconfigpy/config.py
JonWBedard: The specific use-case I'm thinking of here is `buildbot`, where I think it quite likely that we will start mixing JSON and YAML, where we want the derived object to be JSON but the source object will likely be YAML in the near future.

emw-apple (changes requested):
We're talking more offline, but as some early high-level feedback, I'm not a big fan of bolting extra constructs on top of json-e. Seems like you are proposing permanent fork which will make everything harder to maintain and more confusing to explain. Can we embrace the core language more?

For the two language constructs you've added:
- `$foreach` is `$map` with different syntax
- `$exec` could be implemented with a context function, like:
   ```python
   >>> import jsone
   >>> from subprocess import check_output
   >>> jsone.render(
   ...     template={'whoami': {'$eval': 'system("whoami")'}},
   ...     context={'system': lambda x: check_output(x, text=True).rstrip()}
   ... )
   {'whoami': 'emw'}
   ```

PR #58880, opened 2026-02-17
Optimize CSS d property

nullhook:
#### bd56a695fa5f
<pre>
Optimize CSS d property
<a href="https://bugs.webkit.org/show_bug.cgi?id=304761">https://bugs.webkit.org/show_bug.cgi?id=304761</a>
<a href="https://rdar.apple.com/167300890">rdar://167300890</a>

Reviewed by NOBODY (OOPS!).

The CSS d property for SVG path elements caused a performance regression
because CSSPathValue creation is expensive. This cost was incurred for
every path element during style resolution, even when CSS doesn&apos;t use the d property.

This change defers creation of the d presentational hint until we know CSS
actually references the d property. In the common case where CSS doesn&apos;t
touch d, no CSSPathValue is created and the path uses its attribute value
directly.

The d attribute is intentionally not mapped in cssPropertyIdForSVGAttributeName
to keep it out of the generic presentational hint machinery. Instead,
SVGPathElement provides a separately cached presentational hint that
ElementRuleCollector adds only when author rules contain the d property.

This fully recovers the performance regression.

                          Baseline    After
  React-Stockcharts-SVG   3.32%       0.23%
  PanTheChart             1.70%       0.07%
  ZoomTheChart            6.54%       0.36%

* Source/WebCore/rendering/svg/RenderSVGPath.cpp:
(WebCore::RenderSVGPath::styleDidChange):
* Source/WebCore/rendering/svg/legacy/LegacyRenderSVGPath.cpp:
(WebCore::LegacyRenderSVGPath::styleDidChange):
* Source/WebCore/style/ElementRuleCollector.cpp:
(WebCore::Style::ElementRuleCollector::matchedRulesContainProperty const):
(WebCore::Style::ElementRuleCollector::matchAllRules):
* Source/WebCore/style/ElementRuleCollector.h:
* Source/WebCore/style/StyleExtractorCustom.h:
(WebCore::Style::PropertyExtractorAdaptor&lt;CSSPropertyD&gt;::computedValue const):
(WebCore::Style::ExtractorCustom::extractD):
(WebCore::Style::ExtractorCustom::extractDSerialization):
* Source/WebCore/style/computed/data/StyleSVGLayoutData.cpp:
(WebCore::Style::SVGLayoutData::SVGLayoutData):
(WebCore::Style::SVGLayoutData::operator== const):
(WebCore::Style::SVGLayoutData::dumpDifferences const):
(WebCore::Style::operator&lt;&lt;):
* Source/WebCore/style/computed/data/StyleSVGLayoutData.h:
(WebCore::SVGElement::cssPropertyIdForSVGAttributeName):
(WebCore::SVGPathElement::svgAttributeChanged):
(WebCore::SVGPathElement::pathByteStream const):
(WebCore::SVGPathElement::path const):
(WebCore::SVGPathElement::presentationalHintStyleForDAttribute const):
(WebCore::SVGPathElement::buildDPresentationalHintStyle const):
(WebCore::SVGPathElement::invalidateDPresentationalHintCache):
(WebCore::SVGPathElement::collectPresentationalHintsForAttribute): Deleted.
(WebCore::SVGPathElement::collectExtraStyleForPresentationalHints): Deleted.
(WebCore::SVGPathElement::collectDPresentationalHint): Deleted.
* Source/WebCore/svg/SVGPathElement.h:
</pre>
https://github.com/WebKit/WebKit/commit/bd56a695fa5f64c24f44ac528cda080f73bdfb37

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/145980 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/18669 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/10816 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/154659 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/99516 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/563ba822-6495-4d59-b767-8704107f37bb) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/147855 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/19149 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/18559 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/112291 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/80386 "1 flakes 1 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b50dbd67-b045-4448-a671-165cac38fd7c) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/148943 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/14671 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/131135 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/93191 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/9e79c31c-b27a-4208-af87-9f7d609254a8) 
| | [❌ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/13953 "Found 2 new test failures: imported/w3c/web-platform-tests/css/css-cascade/all-prop-revert-layer-noop.html?include=6 imported/w3c/web-platform-tests/css/css-cascade/all-prop-revert-noop.html?include=6 (failure)") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/11710 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/2104 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/123509 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/8069 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/156970 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/191 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/9309 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/120304 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/18503 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/15470 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/120639 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/18536 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/129484 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/74214 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/16343 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/7426 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/18122 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/81890 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/17858 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/18038 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/17918 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/style/ElementRuleCollector.cpp
+++ b/Source/WebCore/style/ElementRuleCollector.cpp
@@ -44,6 +44,8 @@
 #include "HTMLElement.h"
 #include "HTMLSlotElement.h"
 #include "SVGElement.h"
+#include "SVGElementTypeHelpers.h"
+#include "SVGPathElement.h"
 #include "SelectorCheckerTestFunctions.h"
 #include "SelectorCompiler.h"
 #include "SelectorMatchingState.h"
@@ -855,6 +857,15 @@ static inline bool compareRules(MatchedRule r1, MatchedRule r2)
     return r1.ruleData->position() < r2.ruleData->position();
 }
 
+bool ElementRuleCollector::matchedRulesContainProperty(CSSPropertyID propertyID) const
+{
+    for (const auto& matchedRule : m_matchedRules) {
+        if (matchedRule.ruleData->styleRule().properties().findPropertyIndex(propertyID) >= 0)
+            return true;
+    }
+    return false;
+}
+
 void ElementRuleCollector::sortMatchedRules()
 {
     std::ranges::sort(m_matchedRules, compareRules);
@@ -897,8 +908,19 @@ void ElementRuleCollector::matchAllRules(bool matchAuthorAndUserStyles, bool inc
         clearMatchedRules();
 
         collectMatchingRules(DeclarationOrigin::Author);
-        sortMatchedRules();
 
+        // If CSS uses the d property, add the d attribute presentational hint for proper cascade.
+        // This must be done BEFORE transferring author rules so that author rules can override
+        // the presentational hint. This avoids creating CSSPathValue when CSS doesn't touch
+        // the d property (which is the common case).
+        if (RefPtr pathElement = dynamicDowncast<SVGPathElement>(element())) {
+            if (matchedRulesContainProperty(CSSPropertyD)) {
+                if (RefPtr dHintStyle = pathElement->presentationalHintStyleForDAttribute())
+                    addElementStyleProperties(dHintStyle.get(), RuleSet::cascadeLayerPriorityForPresentationalHints);
+            }
+        }
+
+        sortMatchedRules();
         transferMatchedRules(DeclarationOrigin::Author, ScopeOrdinal::Element);
 
         // Inline style behaves as if it has higher specificity than any rule.
--- a/Source/WebCore/style/computed/data/StyleSVGLayoutData.h
+++ b/Source/WebCore/style/computed/data/StyleSVGLayoutData.h
@@ -62,6 +62,7 @@ public:
     SVGCoordinateComponent x;
     SVGCoordinateComponent y;
     SVGPathData d;
+    bool hasExplicitlySetD { false };
 
 private:
     SVGLayoutData();
--- a/Source/WebCore/svg/SVGPathElement.cpp
+++ b/Source/WebCore/svg/SVGPathElement.cpp
@@ -149,8 +149,13 @@ void SVGPathElement::svgAttributeChanged(const QualifiedName& attrName)
             path->setNeedsShapeUpdate();
 
         updateSVGRendererForElementChange();
-        if (document().settings().cssDPropertyEnabled())
-            setPresentationalHintStyleIsDirty();
+
+        if (document().settings().cssDPropertyEnabled()) {
+            invalidateDPresentationalHintCache();
+            if (renderer() && renderer()->style().hasExplicitlySetD())
+                invalidateStyle();
+        }
+
         invalidateResourceImageBuffersIfNeeded();
         return;
     }
@@ -239,9 +244,11 @@ const SVGPathByteStream& SVGPathElement::pathByteStream() const
 {
     if (document().settings().cssDPropertyEnabled()) {
         if (CheckedPtr renderer = this->renderer()) {
-            if (auto& pathFunction = renderer->style().d().tryPath())
-                return pathFunction->parameters.data.byteStream;
-            return SVGPathByteStream::empty();
+            if (renderer->style().hasExplicitlySetD()) {
+                if (auto& pathFunction = renderer->style().d().tryPath())
+                    return pathFunction->parameters.data.byteStream;
+                return SVGPathByteStream::empty();
+            }
         }
     }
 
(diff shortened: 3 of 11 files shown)


Source/WebCore/style/ElementRuleCollector.cpp
```diff
@@ -895,6 +906,17 @@ void ElementRuleCollector::matchAllRules(bool matchAuthorAndUserStyles, bool inc
...
+        // If CSS uses the d property, add the d attribute presentational hint for proper cascade.
+        // This must be done BEFORE transferring author rules so that author rules can override
+        // the presentational hint. This avoids creating CSSPathValue when CSS doesn't touch
+        // the d property (which is the common case).
+        if (matchedRulesContainProperty(CSSPropertyD)) {
+            if (RefPtr pathElement = dynamicDowncast<SVGPathElement>(element())) {
+                if (RefPtr dHintStyle = pathElement->presentationalHintStyleForDAttribute())
+                    addElementStyleProperties(dHintStyle.get(), RuleSet::cascadeLayerPriorityForPresentationalHints);
+            }
+        }
```
anttijk: This code should be next to other presentational hint collection code. It should not do things like `if (matchedRulesContainProperty(CSSPropertyD))` but just add the properties unconditionally. We'll sort of what overrides what during cascading.

You can just do
```
addElementStyleProperties(pathElement->presentationalHintStyleForDAttribute(), RuleSet::cascadeLayerPriorityForPresentationalHints)
```
`addElementStyleProperties` handles null testing already.

Source/WebCore/style/computed/data/StyleSVGLayoutData.h
```diff
@@ -62,6 +62,7 @@ class SVGLayoutData : public RefCounted<SVGLayoutData> {
     SVGCoordinateComponent x;
     SVGCoordinateComponent y;
     SVGPathData d;
+    bool hasExplicitlySetD { false };
```
anttijk, line 65: Why do we care if it is explicitly set? When is it implicitly set?

Source/WebCore/style/ElementRuleCollector.cpp
```diff
@@ -851,6 +853,15 @@ static inline bool compareRules(MatchedRule r1, MatchedRule r2)
...
 }
 
+bool ElementRuleCollector::matchedRulesContainProperty(CSSPropertyID propertyID) const
+{
+    for (const auto& matchedRule : m_matchedRules) {
+        if (matchedRule.ruleData->styleRule().properties().findPropertyIndex(propertyID) >= 0)
+            return true;
+    }
+    return false;
+}
```
anttijk, line 867: This is unnecessary, slow and wrong to have in ElementRuleCollector (as noted elsewhere).

Source/WebCore/svg/SVGPathElement.cpp
```diff
@@ -252,40 +258,42 @@ Path SVGPathElement::path() const
...
-    ASSERT(document().settings().cssDPropertyEnabled());
-    // In the case of the `d` property, we want to avoid providing a string value since it will require
-    // the path data to be parsed again and path data can be unwieldy.
-    auto property = cssPropertyIdForSVGAttributeName(SVGNames::dAttr, document().settings());
+    auto style = MutableStyleProperties::create(SVGAttributeMode);
     // The fill rule value passed here is not relevant for the `d` property.
     auto cssPathValue = CSSPathValue::create(CSS::PathFunction { CSS::Keyword::Nonzero { }, CSS::Path::Data { Ref { m_pathSegList }->currentPathByteStream() } });
-    addPropertyToPresentationalHintStyle(style, property, WTF::move(cssPathValue));
+    style->setProperty(CSSPropertyD, WTF::move(cssPathValue));
+    return style->immutableCopy();
```
anttijk: Alternatively we could just keep the MutableStyleProperties and mutate the existing cache entry if the properties change.

Source/WebCore/svg/SVGPathElement.cpp
```diff
@@ -252,40 +258,42 @@ Path SVGPathElement::path() const
...
+}
+
+void SVGPathElement::invalidateDPresentationalHintCache()
+{
+    m_dPresentationalHintStyleIsDirty = true;
+    m_cachedDPresentationalHintStyle = nullptr;
```
anttijk: The separate bit may be unncessary as nullptr m_cachedDPresentationalHintStyle already indicates that the cache is dirty.

Source/WebCore/style/computed/data/StyleSVGLayoutData.h
nullhook: We need to distinguish between when CSS sets `d: none` to render nothing and when CSS didn't touch `d` to fallback using attribute value. Without this flag, both cases have `style().d()` returning none (initial value), but they should behave differently.

Source/WebCore/style/ElementRuleCollector.cpp
nullhook: isn't the null check in `addElementStyleProperties` too late?

when we call `presentationalHintStyleForDAttribute()`, if the cache is dirty, it effectively creates `CSSPathValue` and this is expensive and we're trying avoid it.

without the guard every `path` element will call `presentationalHintStyleForDAttribute` and effectively create `CSSPathValue` even if CSS doesn't use `d` (common case)

weinig:
I think this is the wrong approach and it will be hard to get all the corner cases right here, as an invariant we typically hold is that that the ComputedValue is the source of truth of style. Things like inheritance and animation/transitions all depend on the ComputedValue holding the correct values.

I'd like to understand more about why creating a CSSPathValue is so expensive. From looking at the code, it seems like just a standard allocation.

nullhook:
An observation about the CSS d syntax is: `d: path(fill-rule, "M 0 0 L 100 100");`. The `CSSPathValue` is modeled around this idea. It consists of a `CSS::PathFunction` > `CSS::Path` (which includes the fill rule and data) > `CSS::Path::Data` and `SVGPathByteStream`.

This chain makes sense when representing the CSS-side of a path value. However, for presentational hints coming from the SVG d attribute (`<path d="..." />`), why are we forced go through this chain?

I haven't fully formalized my thoughts around this yet but we could add a `Variant<CSS::PathFunction, WeakPtr<SVGPathElement>>` in CSSPathValue to store either owned path data (for CSS) or a reference to the element (for presentational hints).

When collecting presentational hints from SVGPathElement, we could do something like:

```
addPropertyToPresentationalHintStyle(style, property, CSSPathValue::createFromElement(*this));
```

and defer the construction of the full `CSS::PathFunction` wrapper until `CSSPathValue::path()` is called.

PR #58956, opened 2026-02-18
[GTK][WPE] Implement GPU atlas creation and replay substitution for batched raster image uploads

nikolaszimmermann:
#### 48bf47fef51c61baf918a618ab851081493bf1ca
<pre>
[GTK][WPE] Implement GPU atlas creation and replay substitution for batched raster image uploads
<a href="https://bugs.webkit.org/show_bug.cgi?id=308166">https://bugs.webkit.org/show_bug.cgi?id=308166</a>

This builds on the previous atlas layout computation to complete
the texture atlas pipeline for batching raster image uploads in
the Skia painting engine.

During the recording phase, raster images are collected and packed
into atlas layouts. After recording, SkiaPaintingEngine now creates
GPU atlases on the main thread by acquiring BitmapTextures from the
texture pool and composing the raster image pixels into them. Two
composition paths are supported: a DMA-buf path that memory-maps
the GPU buffer for direct pixel writes (optimized - dispatched to a
dedicated composition worker thread), and a GL fallback path that
uses BitmapTexture::updateContents() synchronously. A single GL
fence is inserted after all atlas uploads to allow worker threads
to wait for completion before using the textures.

The DMA-buf atlas composition path is enabled by default on
platforms supporting GBM. It can be disabled at runtime by setting
the WEBKIT_DISABLE_DMABUF_ATLAS environment variable to 1.

Async DMA-buf atlas composition is synchronized using an
AtlasCompositionFence countdown latch (Lock + Condition). A single
fence is shared across all atlas compositions for a given recording,
with addPending()/signal() counting dispatched work and wait()
blocking replay workers until all compositions complete.

On the replay side, SkiaReplayCanvas now creates per-worker
SkiaReplayAtlas wrappers that rewrap the pre-uploaded GPU textures
for each worker thread&apos;s GrDirectContext. During picture playback,
onDrawImage2() and onDrawImageRect2() intercept raster image draws
and substitute them with atlas texture draws, mapping source
coordinates into atlas space.

All atlas draw paths use kStrict_SrcRectConstraint to prevent
texture bleeding artifacts. When drawing from a standalone image,
kFast_SrcRectConstraint is safe because sampling beyond the source
rect just clamps to the image&apos;s own edge pixels. However, when
drawing a sub-region from an atlas texture, sampling beyond the
source rect would read pixels from adjacent images in the atlas,
producing visible edge artifacts. kStrict_SrcRectConstraint tells
Skia to clamp texture sampling strictly within the source rect.

Images with non-sRGB color spaces (e.g. linear RGB from SVG filter
results rendered with color-interpolation-filters: linearRGB) are
converted to sRGB during atlas composition. The atlas texture is
always tagged sRGB, so without conversion the raw pixel data would
be misinterpreted, causing visible color shifts (darker colors).
The conversion uses Skia&apos;s built-in readPixels() with an sRGB
target SkImageInfo. sRGB and untagged images use a zero-copy
peekPixels() fast path.

Covered by existing tests.

* Source/WebCore/platform/Skia.cmake:
* Source/WebCore/platform/SourcesSkia.txt:
* Source/WebCore/platform/graphics/NativeImage.h:
* Source/WebCore/platform/graphics/skia/GraphicsContextSkia.cpp:
(WebCore::GraphicsContextSkia::drawNativeImage):
(WebCore::GraphicsContextSkia::beginRecording):
(WebCore::GraphicsContextSkia::endRecording):
* Source/WebCore/platform/graphics/skia/NativeImageSkia.cpp:
(WebCore::NativeImage::uniqueID const):
* Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.cpp: Added.
(WebCore::SkiaGPUAtlas::SkiaGPUAtlas):
(WebCore::SkiaGPUAtlas::create):
(WebCore::SkiaGPUAtlas::uploadImages):
* Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.h: Added.
(WebCore::SkiaGPUAtlas::backendTexture const):
(WebCore::SkiaGPUAtlas::imageToRect const):
(WebCore::SkiaGPUAtlas::size const):
(WebCore::SkiaGPUAtlas::atlasTexture const):
* Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp:
(WebCore::SkiaPaintingEngine::SkiaPaintingEngine):
(WebCore::SkiaPaintingEngine::createAtlas):
(WebCore::SkiaPaintingEngine::record):
(WebCore::SkiaPaintingEngine::replay):
(WebCore::SkiaPaintingEngine::shouldUseDMABufAtlasTextures):
* Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.h:
(WebCore::SkiaPaintingEngine::useThreadedRendering const):
* Source/WebCore/platform/graphics/skia/SkiaRecordingResult.cpp:
(WebCore::SkiaRecordingResult::waitForUploadFence):
(WebCore::SkiaRecordingResult::waitForUploadCondition):
* Source/WebCore/platform/graphics/skia/SkiaRecordingResult.h:
(WebCore::AtlasUploadCondition::create):
(WebCore::AtlasUploadCondition::addPending):
(WebCore::AtlasUploadCondition::signal):
(WebCore::AtlasUploadCondition::wait):
(WebCore::AtlasUploadCondition::WTF_GUARDED_BY_LOCK):
* Source/WebCore/platform/graphics/skia/SkiaReplayAtlas.cpp: Added.
(WebCore::SkiaReplayAtlas::SkiaReplayAtlas):
(WebCore::SkiaReplayAtlas::create):
(WebCore::SkiaReplayAtlas::rectForImage const):
* Source/WebCore/platform/graphics/skia/SkiaReplayAtlas.h: Copied from Source/WebCore/platform/graphics/skia/SkiaRecordingResult.cpp.
(WebCore::SkiaReplayAtlas::atlasTexture const):
* Source/WebCore/platform/graphics/skia/SkiaReplayCanvas.cpp:
(WebCore::SkiaReplayCanvas::SkiaReplayCanvas):
(WebCore::SkiaReplayCanvas::create):
(WebCore::SkiaReplayCanvas::onDrawImage2):
(WebCore::SkiaReplayCanvas::onDrawImageRect2):
* Source/WebCore/platform/graphics/skia/SkiaReplayCanvas.h:
* Source/WebCore/platform/graphics/texmap/BitmapTexture.cpp:
(WebCore::BitmapTexture::BitmapTexture):
(WebCore::BitmapTexture::reset):
Fix a problem with the untested code path, pixelFormat was wrong
when using the BGRALayout flag

Canonical link: <a href="https://commits.webkit.org/308458@main">https://commits.webkit.org/308458@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/ec68e8347dba3f48a6523553ba4fff6a364755cf

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/147638 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/20323 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/13914 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/156320 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/101053 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/149511 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/20780 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/20223 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/113807 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/101053 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/150600 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/16049 "Passed tests") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/132606 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94568 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/15214 "Passed tests") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/12999 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/3761 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/124810 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/10523 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/158654 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/1790 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/11995 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/121833 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/20122 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/16902 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/122034 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/31238 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/20133 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/132304 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/76238 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/17573 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/9084 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/19737 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/83500 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/19467 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/19618 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/19525 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/platform/graphics/skia/GraphicsContextSkia.cpp
+++ b/Source/WebCore/platform/graphics/skia/GraphicsContextSkia.cpp
@@ -283,8 +283,8 @@ void GraphicsContextSkia::drawNativeImage(const NativeImage& nativeImage, const 
         return;
 
     // Collect raster images for atlas batching during recording.
-    if (m_contextMode == ContextMode::RecordingMode && !image->isTextureBacked() && m_atlasLayoutBuilder) {
-        // FIXME: Remove m_atlasLayoutBuilder check and turn into ASSERT(m_atlasLayoutBuilder), once atlas mode is activated.
+    if (m_contextMode == ContextMode::RecordingMode && m_renderingMode == RenderingMode::Accelerated && !image->isTextureBacked()) {
+        ASSERT(m_atlasLayoutBuilder);
         m_atlasLayoutBuilder->collectRasterImage(image);
     }
 
@@ -1236,8 +1236,11 @@ void GraphicsContextSkia::beginRecording()
 {
     ASSERT(m_contextMode == ContextMode::PaintingMode);
     m_contextMode = ContextMode::RecordingMode;
-    // FIXME: Enable atlas mode, once upstreaming completed.
-    // m_atlasLayoutBuilder = makeUnique<SkiaImageAtlasLayoutBuilder>();
+
+    if (m_renderingMode == RenderingMode::Accelerated)
+        m_atlasLayoutBuilder = makeUnique<SkiaImageAtlasLayoutBuilder>();
+    else
+        ASSERT(!m_atlasLayoutBuilder);
 }
 
 SkiaRecordingData GraphicsContextSkia::endRecording()
@@ -1245,7 +1248,6 @@ SkiaRecordingData GraphicsContextSkia::endRecording()
     ASSERT(m_contextMode == ContextMode::RecordingMode);
     m_contextMode = ContextMode::PaintingMode;
 
-    // FIXME: Remove m_atlasLayoutBuilder check and turn into ASSERT(m_atlasLayoutBuilder), once atlas mode is activated.
     Vector<Ref<SkiaImageAtlasLayout>> atlasLayouts;
     if (m_atlasLayoutBuilder) {
         atlasLayouts = m_atlasLayoutBuilder->finalize();
--- a/Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.cpp
+++ b/Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.cpp
@@ -1,0 +1,139 @@
+/*
+ * Copyright (C) 2025, 2026 Igalia S.L.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL APPLE INC. OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ * OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "SkiaGPUAtlas.h"
+
+#if USE(SKIA)
+#include "SkiaImageAtlasLayout.h"
+#if USE(GBM)
+#include "MemoryMappedGPUBuffer.h"
+#endif
+#include <wtf/TZoneMallocInlines.h>
+
+WTF_IGNORE_WARNINGS_IN_THIRD_PARTY_CODE_BEGIN
+#include <skia/core/SkColorSpace.h>
+#include <skia/core/SkPixmap.h>
+#include <skia/gpu/ganesh/gl/GrGLBackendSurface.h>
+WTF_IGNORE_WARNINGS_IN_THIRD_PARTY_CODE_END
+
+#if USE(LIBEPOXY)
+#include <epoxy/gl.h>
+#else
+#include <GLES3/gl3.h>
+#endif
+
+namespace WebCore {
+
+WTF_MAKE_TZONE_ALLOCATED_IMPL(SkiaGPUAtlas);
+
+SkiaGPUAtlas::SkiaGPUAtlas(Ref<BitmapTexture>&& atlasTexture, GrBackendTexture&& backendTexture, const SkiaImageAtlasLayout& layout, const IntSize& size)
+    : m_atlasTexture(WTF::move(atlasTexture))
+    , m_backendTexture(WTF::move(backendTexture))
+    , m_layout(layout)
+    , m_size(size)
+{
+    m_imageToRect.reserveInitialCapacity(layout.entries().size());
+    for (const auto& entry : layout.entries()) {
+        const auto& rect = entry.atlasRect;
+        m_imageToRect.add(entry.rasterImage.get(), SkRect::MakeXYWH(rect.x(), rect.y(), rect.width(), rect.height()));
+    }
+}
+
+SkiaGPUAtlas::~SkiaGPUAtlas() = default;
+
+RefPtr<SkiaGPUAtlas> SkiaGPUAtlas::create(const SkiaImageAtlasLayout& layout, Ref<BitmapTexture>&& atlasTexture)
+{
+    const auto& atlasSize = layout.atlasSize();
+    if (atlasSize.isEmpty())
+        return nullptr;
+
+    RELEASE_ASSERT(atlasSize == atlasTexture->size());
+
+    GrGLTextureInfo externalTexture;
+    externalTexture.fTarget = GL_TEXTURE_2D;
+    externalTexture.fID = atlasTexture->id();
+    externalTexture.fFormat = GL_RGBA8;
+    auto backendTexture = GrBackendTextures::MakeGL(atlasSize.width(), atlasSize.height(), skgpu::Mipmapped::kNo, externalTexture);
+    if (!backendTexture.isValid())
+        return nullptr;
+
+    return adoptRef(*new SkiaGPUAtlas(WTF::move(atlasTexture), WTF::move(backendTexture), layout, atlasSize));
+}
+
+bool SkiaGPUAtlas::uploadImages()
+{
+    Vector<uint8_t> conversionBuffer;
+
+    // Returns pixel data for atlas upload — either a zero-copy reference to the original
+    // pixels (fast path) or a converted sRGB copy (for non-sRGB images).
+    auto pixelDataInSRGB = [&conversionBuffer](const sk_sp<SkImage>& image) -> std::optional<std::pair<const void*, size_t>> {
+        SkPixmap pixmap;
+        if (!image->peekPixels(&pixmap))
+            return std::nullopt;
+
+        if (auto* colorSpace = image->colorSpace(); !colorSpace || colorSpace->isSRGB())
+            return std::pair { pixmap.addr(), pixmap.rowBytes() };
+
+        // Convert to sRGB using Skia's built-in color space conversion.
+        auto srgbInfo = SkImageInfo::Make(image->width(), image->height(), image->colorType(), image->alphaType(), SkColorSpace::MakeSRGB());
+        size_t srgbRowBytes = srgbInfo.minRowBytes();
+        conversionBuffer.resize(srgbInfo.computeMinByteSize());
+
+        if (!image->readPixels(static_cast<GrDirectContext*>(nullptr), srgbInfo, conversionBuffer.mutableSpan().data(), srgbRowBytes, 0, 0))
+            return std::nullopt;
+
+        return std::pair { static_cast<const void*>(conversionBuffer.mutableSpan().data()), srgbRowBytes };
+    };
+
+#if USE(GBM)
+    if (auto* gpuBuffer = m_atlasTexture->memoryMappedGPUBuffer()) {
+        if (gpuBuffer->isLinear() || gpuBuffer->isVivanteSuperTiled()) {
+            auto writeScope = makeGPUBufferWriteScope(*gpuBuffer);
+            if (!writeScope)
+                return false;
+
+            for (const auto& entry : m_layout.entries()) {
+                if (auto pixels = pixelDataInSRGB(entry.rasterImage))
+                    gpuBuffer->updateContents(*writeScope, pixels->first, entry.atlasRect, pixels->second);
+            }
+
+            return true;
+        }
+    }
+#endif
+
+    // GL fallback: use BitmapTexture::updateContents() per entry.
+    for (const auto& entry : m_layout.entries()) {
+        if (auto pixels = pixelDataInSRGB(entry.rasterImage))
+            m_atlasTexture->updateContents(pixels->first, entry.atlasRect, IntPoint::zero(), pixels->second, PixelFormat::BGRA8);
+    }
+
+    return true;
+}
+
+} // namespace WebCore
+
+#endif // USE(SKIA)
--- a/Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
+++ b/Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
@@ -37,12 +37,18 @@
 #include "PlatformDisplay.h"
 #include "ProcessCapabilities.h"
 #include "RenderingMode.h"
+#include "SkiaGPUAtlas.h"
+#include "SkiaImageAtlasLayout.h"
 #include "SkiaRecordingResult.h"
 #include "SkiaReplayCanvas.h"
 WTF_IGNORE_WARNINGS_IN_THIRD_PARTY_CODE_BEGIN
+#include <skia/core/SkColorSpace.h>
 #include <skia/core/SkPictureRecorder.h>
+#include <skia/core/SkSurface.h>
 #include <skia/gpu/ganesh/GrBackendSurface.h>
+#include <skia/gpu/ganesh/GrDirectContext.h>
 #include <skia/gpu/ganesh/SkImageGanesh.h>
+#include <skia/gpu/ganesh/SkSurfaceGanesh.h>
 WTF_IGNORE_WARNINGS_IN_THIRD_PARTY_CODE_END
 #include <wtf/NumberOfCores.h>
 #include <wtf/SystemTracing.h>
@@ -70,13 +76,13 @@ SkiaPaintingEngine::SkiaPaintingEngine()
         m_texturePool = makeUnique<BitmapTexturePool>();
 
         if (auto numberOfGPUThreads = numberOfGPUPaintingThreads())
-            m_workerPool = WorkerPool::create("SkiaGPUWorker"_s, numberOfGPUThreads);
+            m_paintingWorkerPool = WorkerPool::create("SkiaGPUWorker"_s, numberOfGPUThreads);
 
         return;
     }
 
     if (auto numberOfCPUThreads = numberOfCPUPaintingThreads())
-        m_workerPool = WorkerPool::create("SkiaCPUWorker"_s, numberOfCPUThreads);
+        m_paintingWorkerPool = WorkerPool::create("SkiaCPUWorker"_s, numberOfCPUThreads);
 }
 
 SkiaPaintingEngine::~SkiaPaintingEngine() = default;
@@ -121,6 +127,50 @@ Ref<CoordinatedTileBuffer> SkiaPaintingEngine::createBuffer(RenderingMode render
     return CoordinatedUnacceleratedTileBuffer::create(size, contentsOpaque ? CoordinatedTileBuffer::NoFlags : CoordinatedTileBuffer::SupportsAlpha);
 }
 
+RefPtr<SkiaGPUAtlas> SkiaPaintingEngine::createAtlas(const SkiaImageAtlasLayout& layout, AtlasUploadCondition& uploadCondition)
+{
+    if (!m_texturePool)
+        return { };
+
+    const auto& atlasSize = layout.atlasSize();
+
+    OptionSet<BitmapTexture::Flags> textureFlags { BitmapTexture::Flags::UseBGRALayout, BitmapTexture::Flags::NearestFiltering, BitmapTexture::Flags::SupportsAlpha };
+    bool isDMABufBackedTexture = false;
+#if USE(GBM)
+    if (shouldUseDMABufAtlasTextures()) {
+        isDMABufBackedTexture = true;
+        textureFlags.add({ BitmapTexture::Flags::BackedByDMABuf, BitmapTexture::Flags::ForceLinearBuffer });
+    }
+#endif
+
+    // Verify the texture actually has DMA-buf backing. BitmapTexture silently
+    // falls back to GL if DMA-buf allocation fails, but we must not dispatch
+    // GL operations to the upload worker thread (which has no GL context).
+    auto texture = m_texturePool->acquireTexture(atlasSize, textureFlags);
+    if (!texture->memoryMappedGPUBuffer())
+        isDMABufBackedTexture = false;
+
+    auto atlas = SkiaGPUAtlas::create(layout, WTF::move(texture));
+    if (!atlas)
+        return nullptr;
+
+    // GL path: upload synchronously.
+    if (!isDMABufBackedTexture) [[unlikely]] {
+        atlas->uploadImages();
+        return atlas;
+    }
+
+    // DMA-buf path: create atlas without uploading, dispatch pixel writes to worker.
+    if (!m_uploadWorkQueue)
+        m_uploadWorkQueue = WorkQueue::create("AtlasUpload"_s);
+    uploadCondition.addPending();
+    m_uploadWorkQueue->dispatch([atlas = Ref { *atlas }, condition = Ref { uploadCondition }]() mutable {
+        atlas->uploadImages();
+        condition->signal();
+    });
+    return atlas;
+}
+
 Ref<CoordinatedTileBuffer> SkiaPaintingEngine::paint(const GraphicsLayerCoordinated& layer, const IntRect& dirtyRect, bool contentsOpaque, float contentsScale)
 {
     // ### Synchronous rendering on main thread ###
@@ -155,7 +205,7 @@ Ref<SkiaRecordingResult> SkiaPaintingEngine::record(const GraphicsLayerCoordinat
 {
     // ### Asynchronous rendering on worker threads ###
     ASSERT(useThreadedRendering());
-    ASSERT(m_workerPool);
+    ASSERT(m_paintingWorkerPool);
 
     auto renderingMode = canPerformAcceleratedRendering() ? RenderingMode::Accelerated : RenderingMode::Unaccelerated;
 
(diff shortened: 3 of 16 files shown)


Source/WebCore/platform/graphics/skia/GraphicsContextSkia.cpp
```diff
@@ -1236,16 +1234,18 @@ void GraphicsContextSkia::beginRecording()
 {
     ASSERT(m_contextMode == ContextMode::PaintingMode);
     m_contextMode = ContextMode::RecordingMode;
-    // FIXME: Enable atlas mode, once upstreaming completed.
-    // m_atlasLayoutBuilder = makeUnique<SkiaImageAtlasLayoutBuilder>();
+
+    if (ProcessCapabilities::canUseAcceleratedBuffers() && PlatformDisplay::sharedDisplay().skiaGLContext())
```
carlosgcampos: Can't we just check m_renderingMode here?

Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include <epoxy/gl.h>
+#else
+#include <GLES3/gl3.h>
+#endif
+
+namespace WebCore {
+
+WTF_MAKE_TZONE_ALLOCATED_IMPL(SkiaGPUAtlas);
+
+SkiaGPUAtlas::SkiaGPUAtlas(Ref<BitmapTexture>&& atlasTexture, GrBackendTexture&& backendTexture, ImageToRectMap&& imageToRect, IntSize size)
```
carlosgcampos: I would either use IntSize&& and move or const IntSize& and copy.

Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.cpp
```diff
@@ -0,0 +1,180 @@
...
+#if USE(GBM)
+    if (auto* gpuBuffer = atlasTexture.memoryMappedGPUBuffer()) {
+        if (gpuBuffer->isLinear() || gpuBuffer->isVivanteSuperTiled()) {
+            auto writeScope = makeGPUBufferWriteScope(*gpuBuffer);
+            if (!writeScope)
+                return false;
+
+            for (const auto& entry : layout.entries()) {
+                if (auto pixels = PixelDataInSRGB::create(entry.rasterImage, conversionBuffer))
+                    gpuBuffer->updateContents(*writeScope, pixels.data, entry.atlasRect, pixels.rowBytes);
```
carlosgcampos: I think the PixelDataInSRGB is a bit weird, it receives a const to a vector and also its contents (or not) as return value. I understand this is to keep the data alive for the context of the caller, but it's not obvious. Can't we use BitmapTexture::updateContents that receives a NativeImage? just adding the color conversion there?

Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+static GrBackendTexture createBackendTextureFromBitmapTexture(BitmapTexture& texture)
+{
+    GrGLTextureInfo externalTexture;
+    externalTexture.fTarget = GL_TEXTURE_2D;
+    externalTexture.fID = texture.id();
+    externalTexture.fFormat = GL_RGBA8;
+
+    const auto& size = texture.size();
+    return GrBackendTextures::MakeGL(size.width(), size.height(), skgpu::Mipmapped::kNo, externalTexture);
```
carlosgcampos: Don't we need to make a context current here?

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
```diff
@@ -121,6 +128,50 @@ Ref<CoordinatedTileBuffer> SkiaPaintingEngine::createBuffer(RenderingMode render
...
+{
+    if (!m_texturePool)
+        return { };
+
+    const auto& atlasSize = layout.atlasSize();
+
+    OptionSet<BitmapTexture::Flags> textureFlags;
+    textureFlags.add(BitmapTexture::Flags::UseBGRALayout);
+    textureFlags.add(BitmapTexture::Flags::NearestFiltering);
+    textureFlags.add(BitmapTexture::Flags::SupportsAlpha);
```
carlosgcampos:
```suggestion
    OptionSet<BitmapTexture::Flags> textureFlags = { BitmapTexture::Flags::UseBGRALayout, BitmapTexture::Flags::NearestFiltering, BitmapTexture::Flags::SupportsAlpha };
```

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
```diff
@@ -121,6 +128,50 @@ Ref<CoordinatedTileBuffer> SkiaPaintingEngine::createBuffer(RenderingMode render
...
+    bool isDMABufBackedTexture = false;
+#if USE(GBM)
+    if (shouldUseDMABufAtlasTextures()) {
+        textureFlags.add(BitmapTexture::Flags::BackedByDMABuf);
+        textureFlags.add(BitmapTexture::Flags::ForceLinearBuffer);
```
carlosgcampos:
```suggestion
        textureFlags.add({ BitmapTexture::Flags::BackedByDMABuf, BitmapTexture::Flags::ForceLinearBuffer });
```

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
```diff
@@ -166,10 +217,46 @@ Ref<SkiaRecordingResult> SkiaPaintingEngine::record(const GraphicsLayerCoordinat
...
     WTFEndSignpost(this, RecordTile);
 
-    return SkiaRecordingResult::create(WTF::move(picture), WTF::move(recordingData), recordRect, renderingMode, contentsOpaque, contentsScale);
+    auto result = SkiaRecordingResult::create(WTF::move(picture), WTF::move(recordingData), recordRect, renderingMode, contentsOpaque, contentsScale);
+
+    // Prepare GPU atlases on main thread before dispatching to workers.
+    // Textures are acquired from BitmapTexturePool which handles recycling.
+    if (result->hasAtlasLayouts()) {
+        if (auto* grContext = PlatformDisplay::sharedDisplay().skiaGrContext()) {
+            PlatformDisplay::sharedDisplay().skiaGLContext()->makeContextCurrent();
```
carlosgcampos: Aha, here is where we make the context current then, right?

Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.cpp
nikolaszimmermann: Done on the call-site.

Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.cpp
nikolaszimmermann: We cannot use BitmapTexture::updateContents() as that - if MemoryGPUBuffer backed - creates a write buffer scope every time - the idea here is to batch that: obtaining a single scope, doing all the operations, and releasing it.
But we can still discuss how to make PixelDataInSRGB nicer

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
nikolaszimmermann: Correct. We make it current here, call createAtlas, which calls SkiaGPUAtlas::create, which calls createBackendTextureFromBitmapTexture.

nikolaszimmermann: Is that too convoluted?

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.h
```diff
@@ -65,8 +69,14 @@ class SkiaPaintingEngine {
 private:
     Ref<CoordinatedTileBuffer> createBuffer(RenderingMode, const IntSize&, bool contentsOpaque) const;
     void paintIntoGraphicsContext(const GraphicsLayer&, GraphicsContext&, const IntRect&, bool contentsOpaque, float contentsScale) const;
+    RefPtr<SkiaGPUAtlas> createAtlas(const SkiaImageAtlasLayout&, AtlasCompositionFence*);
 
+    // Painting worker pool (GPU: 2 threads, CPU: nCores/2).
```
nikolaszimmermann: Maybe I should instead rename the m_workerPool to m_paintingWorkerPool, and drop the comment...

Source/WebCore/platform/graphics/skia/SkiaRecordingResult.cpp
```diff
@@ -61,6 +61,33 @@ void SkiaRecordingResult::waitForFenceIfNeeded(const SkImage& image)
...
 
+void SkiaRecordingResult::waitForUploadFence()
+{
+    if (m_uploadFence)
+        m_uploadFence->serverWait();
+}
+
+void SkiaRecordingResult::setGPUAtlases(Vector<Ref<SkiaGPUAtlas>>&& atlases)
+{
+    m_gpuAtlases = WTF::move(atlases);
```
nikolaszimmermann: Does inlining make sense here? I had WTFLogAlways statements here before, and now the bodies are 1-liners...

Source/WebCore/platform/graphics/skia/SkiaReplayCanvas.cpp
```diff
@@ -42,13 +42,34 @@ SkiaReplayCanvas::SkiaReplayCanvas(const IntSize& size, const RefPtr<SkiaRecordi
...
+    // GPU atlases were uploaded on main thread; we just rewrap for this worker's context.
+    if (m_recording && m_recording->hasGPUAtlases()) {
+        auto* glContext = PlatformDisplay::sharedDisplay().skiaGLContext();
+        if (!glContext || !glContext->makeContextCurrent())
+            return;
+
+        auto* grContext = PlatformDisplay::sharedDisplay().skiaGrContext();
+        RELEASE_ASSERT(grContext);
+
+        m_recording->waitForCompositionFence();
```
nikolaszimmermann: Maybe it's confusing to call that a fence, since it's not a GL fence... WDYT?

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.h
nikolaszimmermann: I like that, did it.

Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.cpp
nikolaszimmermann: I folded it into a lambda function instead.

Source/WebCore/platform/graphics/skia/SkiaRecordingResult.cpp
nikolaszimmermann: Did that.

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
```diff
@@ -68,15 +74,16 @@ SkiaPaintingEngine::SkiaPaintingEngine()
 {
     if (canPerformAcceleratedRendering()) {
         m_texturePool = makeUnique<BitmapTexturePool>();
+        m_compositionWorkerPool = WorkerPool::create("AtlasComposition"_s, 1);
```
carlosgcampos: Why is this a worker pool? why not WorkQueue since we always one just one thread?

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
```diff
@@ -121,6 +128,46 @@ Ref<CoordinatedTileBuffer> SkiaPaintingEngine::createBuffer(RenderingMode render
...
+#endif
+
+    // Verify the texture actually has DMA-buf backing. BitmapTexture silently
+    // falls back to GL if DMA-buf allocation fails, but we must not dispatch
+    // GL operations to the composition worker thread (which has no GL context).
+    auto texture = m_texturePool->acquireTexture(atlasSize, textureFlags);
+    if (!texture->memoryMappedGPUBuffer())
+        isDMABufBackedTexture = false;
+
+    if (isDMABufBackedTexture && m_compositionWorkerPool) {
```
carlosgcampos: so the composition thread is only used for dmabuf? why do we need a thread only in that case? how can it be nullptr at this point? Could we create this thread on demand?

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
```diff
@@ -121,6 +128,46 @@ Ref<CoordinatedTileBuffer> SkiaPaintingEngine::createBuffer(RenderingMode render
     return CoordinatedUnacceleratedTileBuffer::create(size, contentsOpaque ? CoordinatedTileBuffer::NoFlags : CoordinatedTileBuffer::SupportsAlpha);
 }
 
+RefPtr<SkiaGPUAtlas> SkiaPaintingEngine::createAtlas(const SkiaImageAtlasLayout& layout, AtlasCompositionFence* compositionFence)
```
carlosgcampos: Why is AtlasCompositionFence a pointer? I don't see how this can be called with nullptr

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
```diff
@@ -121,6 +128,46 @@ Ref<CoordinatedTileBuffer> SkiaPaintingEngine::createBuffer(RenderingMode render
...
+    // falls back to GL if DMA-buf allocation fails, but we must not dispatch
+    // GL operations to the composition worker thread (which has no GL context).
+    auto texture = m_texturePool->acquireTexture(atlasSize, textureFlags);
+    if (!texture->memoryMappedGPUBuffer())
+        isDMABufBackedTexture = false;
+
+    if (isDMABufBackedTexture && m_compositionWorkerPool) {
+        // DMA-buf path: create atlas without composition, dispatch pixel writes to worker.
+        auto atlas = SkiaGPUAtlas::createWithoutComposition(layout, WTF::move(texture));
+        if (atlas && compositionFence) {
```
carlosgcampos: How can be compositionFence nullptr?

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
```diff
@@ -166,10 +213,46 @@ Ref<SkiaRecordingResult> SkiaPaintingEngine::record(const GraphicsLayerCoordinat
...
+            gpuAtlases.reserveInitialCapacity(result->atlasLayouts().size());
+
+            for (const auto& layout : result->atlasLayouts()) {
+                if (auto atlas = createAtlas(layout.get(), compositionFence.ptr()))
+                    gpuAtlases.append(atlas.releaseNonNull());
+            }
+
+            if (!gpuAtlases.isEmpty()) {
+                result->setGPUAtlases(WTF::move(gpuAtlases));
+                result->setCompositionFence(WTF::move(compositionFence));
```
carlosgcampos: These are only called here, could we make the fence a parameter of setGPUAtlases?

Source/WebCore/platform/graphics/skia/SkiaRecordingResult.h
```diff
@@ -44,6 +46,40 @@ namespace WebCore {
 
 using SkiaImageToFenceMap = HashMap<const SkImage*, std::unique_ptr<GLFence>>;
 
+class AtlasCompositionFence : public ThreadSafeRefCounted<AtlasCompositionFence> {
```
carlosgcampos: I would call this AtlasCompositionCondition, I find Fence a bit confusing, I thought it was a gl fence

Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.cpp
```diff
@@ -0,0 +1,172 @@
...
+}
+
+static bool composeImagesIntoTexture(BitmapTexture& atlasTexture, const SkiaImageAtlasLayout& layout)
```
carlosgcampos: I think I would avoid using compose and composition for the atlas, it might be confusing. What about using upload? uploadImagesToTexture, createWihoutUploading, AtlasUploadCondition, etc.

Source/WebCore/platform/graphics/skia/SkiaReplayCanvas.cpp
```diff
@@ -42,13 +42,34 @@ SkiaReplayCanvas::SkiaReplayCanvas(const IntSize& size, const RefPtr<SkiaRecordi
...
+        m_recording->waitForUploadFence();
+
+        const auto& gpuAtlases = m_recording->gpuAtlases();
+        m_atlases.reserveInitialCapacity(gpuAtlases.size());
+
+        for (const auto& gpuAtlas : gpuAtlases) {
+            if (auto atlas = SkiaReplayAtlas::create(gpuAtlas, grContext))
```
carlosgcampos: Why do we pass the gr context? it could be get from create(), no?

Source/WebCore/platform/graphics/skia/SkiaReplayAtlas.cpp
```diff
@@ -0,0 +1,81 @@
...
+    , m_rewrappedTexture(WTF::move(rewrappedTexture))
+    , m_imageToRect(WTF::move(imageToRect))
+{
+}
+
+SkiaReplayAtlas::~SkiaReplayAtlas() = default;
+
+std::unique_ptr<SkiaReplayAtlas> SkiaReplayAtlas::create(const SkiaGPUAtlas& gpuAtlas, GrDirectContext* grContext)
+{
+    if (!gpuAtlas.backendTexture().isValid() || !grContext)
```
carlosgcampos: I think we could get grContext from here. In current patch it can't be nullptr here, since there's a release assert in the caller.

Source/WebCore/platform/graphics/skia/SkiaReplayAtlas.cpp
```diff
@@ -0,0 +1,81 @@
...
+    auto rewrapped = SkImages::BorrowTextureFrom(grContext, gpuAtlas.backendTexture(), kTopLeft_GrSurfaceOrigin, kRGBA_8888_SkColorType, kPremul_SkAlphaType, SkColorSpace::MakeSRGB());
+    if (!rewrapped)
+        return nullptr;
+
+    // Copy the image-to-rect mapping to ensure safe lifetime (SkiaGPUAtlas may be destroyed first).
+    ImageToRectMap imageToRectCopy = gpuAtlas.imageToRect();
+    return std::unique_ptr<SkiaReplayAtlas>(new SkiaReplayAtlas(gpuAtlas.atlasTexture(), WTF::move(rewrapped), WTF::move(imageToRectCopy)));
+}
+
+std::optional<SkRect> SkiaReplayAtlas::rectForImage(const SkImage* image) const
```
carlosgcampos: Could this be a reference instead  of a pointer?

Source/WebCore/platform/graphics/skia/SkiaReplayCanvas.cpp
```diff
@@ -175,6 +196,21 @@ void SkiaReplayCanvas::onDrawGlyphRunList(const sktext::GlyphRunList& glyphRunLi
...
 void SkiaReplayCanvas::onDrawImage2(const SkImage* image, SkScalar x, SkScalar y, const SkSamplingOptions& sampling, const SkPaint* paint)
 {
+    // Check for atlas substitution for raster images.
+    if (image && !image->isTextureBacked()) {
+        for (auto& atlas : m_atlases) {
+            if (auto atlasRect = atlas->rectForImage(image)) {
+                // Draw from atlas instead of original image.
+                const auto& atlasTexture = atlas->atlasTexture();
+                if (!atlasTexture)
+                    break;
```
carlosgcampos: Can this happen? Shouldn't be avoid to create an atlas without a valid texture?

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
nikolaszimmermann: It's only used for dmabuf -- the standard code path. The fallback uses GL from main thread. The null check is a left-over, however, we can actually create it on demand, as a WorkQueue, as you suggested elsewhere. Will adapt

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
nikolaszimmermann: It cannot.

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
nikolaszimmermann: Yeah, makes sense.

Source/WebCore/platform/graphics/skia/SkiaRecordingResult.h
nikolaszimmermann: Makes sense, will adapt.

Source/WebCore/platform/graphics/skia/SkiaReplayCanvas.cpp
nikolaszimmermann: Right.

Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.cpp
nikolaszimmermann: Sounds good.

Source/WebCore/platform/graphics/skia/SkiaReplayCanvas.cpp
nikolaszimmermann: Turned into assertion.

Source/WebCore/platform/graphics/skia/SkiaGPUAtlas.cpp
```diff
@@ -0,0 +1,172 @@
...
+    , m_imageToRect(WTF::move(imageToRect))
+    , m_size(size)
+{
+}
+
+SkiaGPUAtlas::~SkiaGPUAtlas() = default;
+
+static ImageToRectMap buildImageToRectMap(const SkiaImageAtlasLayout& layout)
+{
+    ImageToRectMap imageToRect;
```
nikolaszimmermann: `imageToRect.reserveInitialCapacity(layout.entries().size());` is missing.

Source/WebCore/platform/graphics/skia/SkiaReplayAtlas.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    // Always rewrap the texture for this worker context.
+    // The underlying GL texture was created on the main thread context but is
+    // shareable via GL context sharing. However, the Skia SkImage wrapper is
+    // context-specific and must be rewrapped for each worker's GrDirectContext.
+    auto rewrapped = SkImages::BorrowTextureFrom(grContext, gpuAtlas.backendTexture(), kTopLeft_GrSurfaceOrigin, kRGBA_8888_SkColorType, kPremul_SkAlphaType, SkColorSpace::MakeSRGB());
+    if (!rewrapped)
+        return nullptr;
+
+    // Copy the image-to-rect mapping to ensure safe lifetime (SkiaGPUAtlas may be destroyed first).
```
nikolaszimmermann: Ah this no longer holds true, I changed the architecture several times and this is a relict...

Source/WebCore/platform/graphics/skia/SkiaReplayAtlas.h
```diff
@@ -0,0 +1,75 @@
...
+    const sk_sp<SkImage>& atlasTexture() const { return m_rewrappedTexture; }
+
+    // Lookup: original raster image -> rect within atlas texture.
+    // Returns nullopt if image is not in this atlas.
+    std::optional<SkRect> rectForImage(const SkImage&) const;
+
+private:
+    SkiaReplayAtlas(Ref<BitmapTexture>&&, sk_sp<SkImage>&&, ImageToRectMap&&);
+
+    Ref<BitmapTexture> m_atlasTexture;
```
nikolaszimmermann: We can hold a SkiaGPUAtlas here instead, it's cleaner, since we don't need to copy the ImageToRectMap at all.

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
```diff
@@ -121,6 +127,48 @@ Ref<CoordinatedTileBuffer> SkiaPaintingEngine::createBuffer(RenderingMode render
...
+    // falls back to GL if DMA-buf allocation fails, but we must not dispatch
+    // GL operations to the upload worker thread (which has no GL context).
+    auto texture = m_texturePool->acquireTexture(atlasSize, textureFlags);
+    if (!texture->memoryMappedGPUBuffer())
+        isDMABufBackedTexture = false;
+
+    if (isDMABufBackedTexture) {
+        // DMA-buf path: create atlas without uploading, dispatch pixel writes to worker.
+        auto atlas = SkiaGPUAtlas::createWithoutUploading(layout, WTF::move(texture));
+        if (atlas) {
```
carlosgcampos:
```suggestion
        if (auto atlas = SkiaGPUAtlas::createWithoutUploading(layout, WTF::move(texture))) {
```

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
```diff
@@ -166,10 +214,45 @@ Ref<SkiaRecordingResult> SkiaPaintingEngine::record(const GraphicsLayerCoordinat
...
     auto picture = pictureRecorder.finishRecordingAsPicture();
     WTFEndSignpost(this, RecordTile);
 
-    return SkiaRecordingResult::create(WTF::move(picture), WTF::move(recordingData), recordRect, renderingMode, contentsOpaque, contentsScale);
+    auto result = SkiaRecordingResult::create(WTF::move(picture), WTF::move(recordingData), recordRect, renderingMode, contentsOpaque, contentsScale);
+
+    // Prepare GPU atlases on main thread before dispatching to workers.
+    // Textures are acquired from BitmapTexturePool which handles recycling.
+    if (result->hasAtlasLayouts()) {
+        if (auto* grContext = PlatformDisplay::sharedDisplay().skiaGrContext()) {
```
carlosgcampos: I think we can assert instead, if we are here is because we checked the gr context before, no?

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
```diff
@@ -254,6 +338,26 @@ unsigned SkiaPaintingEngine::numberOfGPUPaintingThreads()
...
 }
 
+bool SkiaPaintingEngine::shouldUseDMABufAtlasTextures()
+{
+#if USE(GBM)
+    static std::once_flag onceFlag;
+    static bool shouldUseDMABufAtlas = true;
+
+    std::call_once(onceFlag, [] {
+        if (const char* envString = getenv("WEBKIT_USE_DMABUF_ATLAS")) {
```
carlosgcampos: Since the default is to use dmabuf, I would call this `WEBKIT_DISABLE_DMABUF_ATLAS` and inverse the logic

carlosgcampos (approved):

Source/WebCore/platform/graphics/skia/SkiaPaintingEngine.cpp
nikolaszimmermann: Makes sense.

carlosgcampos (approved):

PR #59008, opened 2026-02-19
[GLib] Avoid accessing NetworkMDNSRegister after Avahi proxy creation is cancelled

spenap:
#### 6bf92f788fbb70ae21d4b8b5addcf9a05ac53959
<pre>
[GLib] Avoid accessing NetworkMDNSRegister after Avahi proxy creation is cancelled
<a href="https://bugs.webkit.org/show_bug.cgi?id=308222">https://bugs.webkit.org/show_bug.cgi?id=308222</a>

Reviewed by Philippe Normand.

When the NetworkMDNSRegister destructor runs, it calls
g_cancellable_cancel, which will cancel any ongoing asynchronos calls
where m_cancellable participates.

Calls in NetworkMDNSRegister::registerMDNSName are safe, as they only
access memory they&apos;ve allocated, but g_dbus_proxy_new_for_bus in the
constructor keeps a pointer to &quot;this&quot; that would have been invalid after
the destructor has completed.

This reorders the callback so that the instance is only accessed when no
errors are reported.

* Source/WebKit/NetworkProcess/glib/NetworkMDNSRegisterGLib.cpp:
(WebKit::NetworkMDNSRegister::NetworkMDNSRegister):

Canonical link: <a href="https://commits.webkit.org/307846@main">https://commits.webkit.org/307846@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/083b44706250a0c760dc75e7b92e1de62f0771a5

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/145702 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/18384 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/10293 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/154374 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/99341 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/18869 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/18277 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/112052 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/99341 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/148665 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/14424 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/130866 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/92958 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/13727 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/11509 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/1821 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/123275 "Passed tests") | | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/156687 "Built successfully") | 
| | [⏳ 🛠 ios-safer-cpp ](https://ews-build.webkit.org/#/builders/Apple-iOS-26-Safer-CPP-Checks-EWS "Waiting in queue, processing has not started yet") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/8830 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/120054 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/18234 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/15215 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/120405 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30865 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/18280 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/129009 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/73974 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/16133 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/7121 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/17856 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/81635 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/17592 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/17800 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/17655 "Built successfully") | | |

### DIFF
--- a/Source/WebKit/NetworkProcess/glib/NetworkMDNSRegisterGLib.cpp
+++ b/Source/WebKit/NetworkProcess/glib/NetworkMDNSRegisterGLib.cpp
@@ -46,10 +46,12 @@ NetworkMDNSRegister::NetworkMDNSRegister(NetworkConnectionToWebProcess& connecti
     m_cancellable = adoptGRef(g_cancellable_new());
     g_dbus_proxy_new_for_bus(G_BUS_TYPE_SYSTEM, G_DBUS_PROXY_FLAGS_NONE, nullptr, "org.freedesktop.Avahi", "/", "org.freedesktop.Avahi.Server", m_cancellable.get(), [](GObject*, GAsyncResult* result, gpointer userData) {
         GUniqueOutPtr<GError> error;
-        auto self = reinterpret_cast<NetworkMDNSRegister*>(userData);
-        self->m_dbusProxy = adoptGRef(g_dbus_proxy_new_for_bus_finish(result, &error.outPtr()));
-        if (!error)
+        auto proxy = adoptGRef(g_dbus_proxy_new_for_bus_finish(result, &error.outPtr()));
+        if (!error) {
+            auto self = static_cast<NetworkMDNSRegister*>(userData);
+            self->m_dbusProxy = WTF::move(proxy);
             return;
+        }
 
 #if PLATFORM(GTK)
         // Check if the connection to the system bus was refused, don't log an error when that


philn (approved):

Source/WebKit/NetworkProcess/glib/NetworkMDNSRegisterGLib.cpp
```diff
@@ -46,10 +46,12 @@ NetworkMDNSRegister::NetworkMDNSRegister(NetworkConnectionToWebProcess& connecti
     m_cancellable = adoptGRef(g_cancellable_new());
     g_dbus_proxy_new_for_bus(G_BUS_TYPE_SYSTEM, G_DBUS_PROXY_FLAGS_NONE, nullptr, "org.freedesktop.Avahi", "/", "org.freedesktop.Avahi.Server", m_cancellable.get(), [](GObject*, GAsyncResult* result, gpointer userData) {
         GUniqueOutPtr<GError> error;
-        auto self = reinterpret_cast<NetworkMDNSRegister*>(userData);
-        self->m_dbusProxy = adoptGRef(g_dbus_proxy_new_for_bus_finish(result, &error.outPtr()));
-        if (!error)
+        auto proxy = adoptGRef(g_dbus_proxy_new_for_bus_finish(result, &error.outPtr()));
+        if (!error) {
+            auto self = static_cast<NetworkMDNSRegister*>(userData);
```
csaavedra, line 51: This is still not entirely safe. If an asynchronous lambda captures `this`, it needs to either keep a strong reference or use a weak one to check that the object is still alive when the lambda is called. See https://github.com/WebKit/WebKit/wiki/Safer-CPP-Guidelines#capture-protectedthis-or-weakthis-in-asynchronous-lambdas-where-you-need-to-use-this

carlosgcampos: But in this case `this` is not captured by the lambda. What we normally do in this cases when using the gio async apis is that we pass `this` as user data, then we cancel the cancellable in the destructor and the async callback calls the `finish()` and checks if error is cancelled in which case it returns earlky without using the user data.

spenap: Claudio is technically right, and I should have shared more of my reasoning together with the patch, because I thought of that.

Using a strong reference would have caused a circular reference -- the destructor wouldn't be called, and then the only way out would be for the callback to run, and only then the destructor would run as a _noop_.

Using a weak reference is indeed the right solution, but that requires inheriting from "CanMakeWeakPtr" (or something similar). I decided not to push that approach to avoid changing the shared header. As Carlos says, here we don't touch the user data and it's _currently_ safe. If the code is refactored, the risk might be reintroduced.

carlosgcampos: The approach of checking cancellable error first is safe too and it's already fused in other parts

csaavedra: Yes, I see the logic now, if there is a cancellable and is always cancelled in the destructor then checking it first is safe enough.

PR #59096, opened 2026-02-20
Web Inspector: Network: Request / Response menu ignores user’s previous setting

rcaliman-apple:
#### f01b6d195b0b57832b2b6e915d88e55df7d3c5d4
<pre>
Web Inspector: Network: Request / Response menu ignores user’s previous setting
<a href="https://bugs.webkit.org/show_bug.cgi?id=308310">https://bugs.webkit.org/show_bug.cgi?id=308310</a>
<a href="https://rdar.apple.com/108231795">rdar://108231795</a>

Reviewed by BJ Burg.

The existing logic in `WI.ResourceClusterContentView` to reuse the last chosen
content view for a resource has a timing issue. If it&apos;s a custom request content view,
those views are created asynchronously after getting the resource content and determining
if they&apos;re applicable. By this time, the default content view, Request, is used as a
fallback on the first call to `WI.ResourceClusterContentView.prototype._showContentViewForIdentifier()`.
Any last known saved preference per-resource is overwritten with the fallback.

This is fixed by first waiting for the custom content views to load.

However, users have requested remembering content view preferences per resource type
instead of per single resource. This patch introduces a persistent mapping by MIME type
to the last used content view type identifer to restore a user&apos;s prefered content view.

* Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js:
(WI.ResourceClusterContentView.prototype.attached):
(WI.ResourceClusterContentView.prototype.restoreFromCookie):
(WI.ResourceClusterContentView.prototype.showRequest):
(WI.ResourceClusterContentView.prototype.showResponse):
(WI.ResourceClusterContentView.prototype._showContentViewForIdentifier):
(WI.ResourceClusterContentView.prototype._pathComponentSelected):
(WI.ResourceClusterContentView.prototype._resourceLoadingDidFinish):
(WI.ResourceClusterContentView.prototype._getMIMETypeForPreferences):
(WI.ResourceClusterContentView.prototype._getPreferredContentViewIdentifier):
(WI.ResourceClusterContentView.prototype._setPreferredContentViewIdentifier):
(WI.ResourceClusterContentView.prototype._tryEnableCustomResponseContentViews):
(WI.ResourceClusterContentView):

Canonical link: <a href="https://commits.webkit.org/308142@main">https://commits.webkit.org/308142@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/b1ca5741fe4ff5495db8690aca59aeb265226914

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/146579 "Passed style check") | [  ~~🛠 ios~~](https://ews-build.webkit.org/#/builders/159/builds/19258 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 mac~~](https://ews-build.webkit.org/#/builders/168/builds/12777 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/155242 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/99969 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/19717 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/19161 "Built successfully") | [  ~~🧪 wpe-wk2~~](https://ews-build.webkit.org/#/builders/34/builds/112938 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/99969 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/9c278cb9-423b-4075-80dd-9f4cfe0e6dcf) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/149542 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/162/builds/15181 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/131716 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/93685 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/14436 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/12207 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/2690 "Built successfully") | | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/124015 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-wk1~~](https://ews-build.webkit.org/#/builders/167/builds/9578 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/157567 "Built successfully") | | 
| | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/717 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-wk2~~](https://ews-build.webkit.org/#/builders/169/builds/11012 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/1/builds/120949 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [  ~~🛠 vision~~](https://ews-build.webkit.org/#/builders/153/builds/19061 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/16012 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/121156 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/31030 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/19068 "Built successfully") | [  ~~🧪 mac-wk2-stress~~](https://ews-build.webkit.org/#/builders/8/builds/131330 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/74870 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [  ~~🧪 vision-wk2~~](https://ews-build.webkit.org/#/builders/164/builds/16795 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/8240 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/18677 "Built successfully") | [  ~~🛠 mac-safer-cpp~~](https://ews-build.webkit.org/#/builders/120/builds/82424 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [  ~~🛠 tv-sim~~](https://ews-build.webkit.org/#/builders/157/builds/18407 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [  ~~🛠 watch~~](https://ews-build.webkit.org/#/builders/163/builds/18557 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [  ~~🛠 watch-sim~~](https://ews-build.webkit.org/#/builders/152/builds/18466 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | |

### DIFF
--- a/Source/WebInspectorUI/UserInterface/Base/Setting.js
+++ b/Source/WebInspectorUI/UserInterface/Base/Setting.js
@@ -216,6 +216,7 @@ WI.settings = {
     searchFromSelection: new WI.Setting("search-from-selection", false),
     searchRegularExpression: new WI.Setting("search-regular-expression", false),
     selectedNetworkDetailContentViewIdentifier: new WI.Setting("network-detail-content-view-identifier", "preview"),
+    resourceContentViewIdentifierForMIMEType: new WI.Setting("resource-content-view-identifier-for-mime-type", {}),
     sourceMapsEnabled: new WI.Setting("source-maps-enabled", true),
     showConsoleMessageTimestamps: new WI.Setting("show-console-message-timestamps", false),
     showCSSPropertySyntaxInDocumentationPopover: new WI.Setting("show-css-property-syntax-in-documentation-popover", false),
--- a/Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
+++ b/Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
@@ -51,13 +51,7 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
             this._tryEnableCustomRequestContentViews();
         }
 
-        // FIXME: Since a custom response content view may only become available after a response is received
-        // we need to figure out a way to restore / prefer the custom content view. For example if users
-        // always want to prefer the JSON view to the normal Response text view.
-
-        this._currentContentViewSetting = new WI.Setting("resource-current-view-" + this._resource.url.hash, ResourceClusterContentView.Identifier.Response);
-
-        this._tryEnableCustomResponseContentViews();
+        this._enableCustomResponseContentViewsPromise = this._tryEnableCustomResponseContentViews();
     }
 
     // Public
@@ -85,7 +79,14 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
         if (this._shownInitialContent)
             return;
 
-        this._showContentViewForIdentifier(this._currentContentViewSetting.value);
+        this._enableCustomResponseContentViewsPromise
+        .then(() => {
+            let identifier = this._getPreferredContentViewIdentifier();
+            this._showContentViewForIdentifier(identifier);
+        })
+        .catch((error) => {
+            WI.reportInternalError(error);
+        });
     }
 
     closed()
@@ -97,39 +98,39 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
 
     restoreFromCookie(cookie)
     {
-        let contentView = this._showContentViewForIdentifier(cookie[WI.ResourceClusterContentView.ContentViewIdentifierCookieKey]);
+        let contentViewIdentifier = cookie[WI.ResourceClusterContentView.ContentViewIdentifierCookieKey] || this._getPreferredContentViewIdentifier();
 
-        if (contentView.revealPosition) {
-            let textRangeToSelect = null;
-            if (!isNaN(cookie.startLine) && !isNaN(cookie.startColumn) && !isNaN(cookie.endLine) && !isNaN(cookie.endColumn))
-                textRangeToSelect = new WI.TextRange(cookie.startLine, cookie.startColumn, cookie.endLine, cookie.endColumn);
+        this._enableCustomResponseContentViewsPromise.then(() => {
+            let contentView = this._showContentViewForIdentifier(contentViewIdentifier);
 
-            let position = null;
-            if (!isNaN(cookie.lineNumber) && !isNaN(cookie.columnNumber))
-                position = new WI.SourceCodePosition(cookie.lineNumber, cookie.columnNumber);
-            else if (textRangeToSelect)
-                position = textRangeToSelect.startPosition();
+            if (contentView.revealPosition) {
+                let textRangeToSelect = null;
+                if (!isNaN(cookie.startLine) && !isNaN(cookie.startColumn) && !isNaN(cookie.endLine) && !isNaN(cookie.endColumn))
+                    textRangeToSelect = new WI.TextRange(cookie.startLine, cookie.startColumn, cookie.endLine, cookie.endColumn);
 
-            let scrollOffset = null;
-            if (!isNaN(cookie.scrollOffsetX) && !isNaN(cookie.scrollOffsetY))
-                scrollOffset = new WI.Point(cookie.scrollOffsetX, cookie.scrollOffsetY);
+                let position = null;
+                if (!isNaN(cookie.lineNumber) && !isNaN(cookie.columnNumber))
+                    position = new WI.SourceCodePosition(cookie.lineNumber, cookie.columnNumber);
+                else if (textRangeToSelect)
+                    position = textRangeToSelect.startPosition();
 
-            if (position)
-                contentView.revealPosition(position, {...cookie, textRangeToSelect, scrollOffset});
-        }
+                let scrollOffset = null;
+                if (!isNaN(cookie.scrollOffsetX) && !isNaN(cookie.scrollOffsetY))
+                    scrollOffset = new WI.Point(cookie.scrollOffsetX, cookie.scrollOffsetY);
+
+                if (position)
+                    contentView.revealPosition(position, {...cookie, textRangeToSelect, scrollOffset});
+            }
+        });
     }
 
     showRequest()
     {
-        this._shownInitialContent = true;
-
         return this._showContentViewForIdentifier(ResourceClusterContentView.Identifier.Request);
     }
 
     showResponse()
     {
-        this._shownInitialContent = true;
-
         return this._showContentViewForIdentifier(ResourceClusterContentView.Identifier.Response);
     }
 
@@ -337,7 +338,7 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
         return null;
     }
 
-    _showContentViewForIdentifier(identifier)
+    _showContentViewForIdentifier(identifier, saveAsPreference)
     {
         let contentViewToShow = null;
 
@@ -381,14 +382,18 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
 
         console.assert(contentViewToShow);
 
-        this._currentContentViewSetting.value = this._identifierForContentView(contentViewToShow);
+        if (saveAsPreference)
+            this._setPreferredContentViewIdentifier(identifier);
+
+        this._shownInitialContent = true;
 
         return this.contentViewContainer.showContentView(contentViewToShow);
     }
 
     _pathComponentSelected(event)
     {
-        this._showContentViewForIdentifier(event.data.pathComponent.representedObject);
+        const saveAsPreference = true;
+        this._showContentViewForIdentifier(event.data.pathComponent.representedObject, saveAsPreference);
     }
 
     _resourceTypeDidChange(event)
@@ -408,7 +413,7 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
 
     _resourceLoadingDidFinish(event)
     {
-        this._tryEnableCustomResponseContentViews();
+        this._enableCustomResponseContentViewsPromise = this._tryEnableCustomResponseContentViews();
     }
 
     _canUseJSONContentViewForContent(content)
@@ -463,6 +468,41 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
         return mimeType;
     }
 
+    _getMIMETypeForPreferences()
+    {
+        if (!this._resource.mimeType)
+            return null;
+
+        return parseMIMEType(this._resource.mimeType).type;
+    }
+
+    _getPreferredContentViewIdentifier()
+    {
+        let mimeType = this._getMIMETypeForPreferences();
+        if (!mimeType)
+            return null;
+
+        let preferences = WI.settings.resourceContentViewIdentifierForMIMEType.value;
+        return preferences[mimeType] || null;
+    }
+
+    _setPreferredContentViewIdentifier(identifier)
+    {
+        let mimeType = this._getMIMETypeForPreferences();
+        if (!mimeType)
+            return;
+
+        let preferences = WI.settings.resourceContentViewIdentifierForMIMEType.value;
+
+        if (identifier === ResourceClusterContentView.Identifier.Response)
+            delete preferences[mimeType];
+        else
+            preferences[mimeType] = identifier;
+
+        // `preferences` is a referenced object. Changing its contents and simply setting it as a new value won't qualify as value change so it won't be persisted.
+        WI.settings.resourceContentViewIdentifierForMIMEType.value = structuredClone(preferences);
+    }
+
     _tryEnableCustomRequestContentViews()
     {
         let content = this._resource.requestData;
@@ -503,13 +543,13 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
     _tryEnableCustomResponseContentViews()
     {
         if (!this._resource.hasResponse())
-            return;
+            return Promise.resolve();
 
         // WebSocket resources already use a "custom" response content view.
         if (this._resource instanceof WI.WebSocketResource)
-            return;
+            return Promise.resolve();
 
-        this._resource.requestContent()
+        return this._resource.requestContent()
         .then(({error, content}) => {
             if (error || typeof content !== "string")
                 return;


Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -51,13 +51,7 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
             this._tryEnableCustomRequestContentViews();
         }
 
-        // FIXME: Since a custom response content view may only become available after a response is received
-        // we need to figure out a way to restore / prefer the custom content view. For example if users
-        // always want to prefer the JSON view to the normal Response text view.
-
-        this._currentContentViewSetting = new WI.Setting("resource-current-view-" + this._resource.url.hash, ResourceClusterContentView.Identifier.Response);
```
rcaliman-apple, line 58: Should we add clean-up method that removes all the obsolete settings?

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -463,6 +465,43 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
...
+    }
+
+    _getPreferredContentViewIdentifier()
+    {
+        let mimeType = this._getMIMETypeForPreferences();
+        if (!mimeType)
+            return null;
+
+        let preferences = WI.settings.resourceContentViewIdentifierForMIMEType.value;
+        return preferences[mimeType] || null;
```
the-chenergy, line 486: Should this default to `ResourceClusterContentView.Identifier.Response` instead of null, to match how the setting gets updated below?

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
the-chenergy: I think that's a good idea! Although in general I can also see delaying this clean-up to a follow-up patch being better, just in case we revert this patch so we don't accidentally erase some user's settings

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: I think we should keep this as `null` because it's about the _preferred_ identifier. We handle the `null` just as any other unknown identifier in the `default` case in `_showContentViewForIdentifier()`.

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -463,6 +465,43 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
...
+    _setPreferredContentViewIdentifier(identifier)
+    {
+        let mimeType = this._getMIMETypeForPreferences();
+        if (!mimeType)
+            return;
+
+        let preferences = WI.settings.resourceContentViewIdentifierForMIMEType.value;
+
+        if (identifier === ResourceClusterContentView.Identifier.Response) {
+            console.log("cleanup for", mimeType, preferences)
```
AnujPanta1: is this left on purpose ?

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -503,13 +542,13 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
     _tryEnableCustomResponseContentViews()
     {
         if (!this._resource.hasResponse())
-            return;
+            return Promise.reject();
```
AnujPanta1: I'm curious if we are in a state where there's no resource, the `attached()` method catches a rejection and calls error, but this feels like a normal state to be in, maybe we do a Promise.resolve() here ?

Source/WebInspectorUI/UserInterface/Base/Setting.js
```diff
@@ -216,6 +216,7 @@ WI.settings = {
     searchFromSelection: new WI.Setting("search-from-selection", false),
     searchRegularExpression: new WI.Setting("search-regular-expression", false),
     selectedNetworkDetailContentViewIdentifier: new WI.Setting("network-detail-content-view-identifier", "preview"),
+    resourceContentViewIdentifierForMIMEType: new WI.Setting("resource-content-view-identifer-for-mime-type", {}),
```
AnujPanta1: typo: identifier

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -85,7 +79,14 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
         if (this._shownInitialContent)
             return;
 
-        this._showContentViewForIdentifier(this._currentContentViewSetting.value);
+        this._enableCustomResponseContentViewsPromise
```
AnujPanta1, line 82: `attached()` awaits the promise before showing content, but `restoreFromCookie()` still calls `_showContentViewForIdentifier()` synchronously, so if `restoreFromCookie()` fires before the promise resolves, the view might get shown twice from the cookie, then again when the .then() in `attached()` runs the `_shownInitialContent` check in `attached()` happens before the `.then()`, so it won't prevent the second call. could this like cause a flicker or overwrite what the cookie restored ?

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -381,14 +379,18 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
 
         console.assert(contentViewToShow);
 
-        this._currentContentViewSetting.value = this._identifierForContentView(contentViewToShow);
+        if (saveAsPreference)
+            this._setPreferredContentViewIdentifier(identifier);
+
+        this._shownInitialContent = true;
```
AnujPanta1, line 388: not 100% sure but I think ` _shownInitialContent` now gets set on every code path that shows a view, including restoreFromCookie(), probably fine but we use it for guards I think

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: Oops!

burg (approved):
Nice. Please address Anuj's feedbacks before merging, thanks.

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: Great catch! 👍

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: That's right! The empty resource is handled elsewhere, just like the check for `WI.WebSocketResource` which in the previous version of the code also returned `undefined`.

`Promise.resolve()` makes sense.

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: Makes sense! 👍

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -337,7 +338,7 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
         return null;
     }
 
-    _showContentViewForIdentifier(identifier)
+    _showContentViewForIdentifier(identifier, saveAsPreference)
```
dcrousso, line 341: since `saveAsPreference` is optional lets put it in a `options = {}`
```js
_showContentViewForIdentifier(identifier, {saveAsPreference} = {})
```
that way callers have to actually provide the key instead of creating a `const`

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
dcrousso: i wonder if `_shownInitialContent` is even really needed anymore now that `WI.View` is a thing (i.e. it was added _way_ before we had any sort of system)

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -85,7 +79,14 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
...
 
-        this._showContentViewForIdentifier(this._currentContentViewSetting.value);
+        this._enableCustomResponseContentViewsPromise
+        .then(() => {
+            let identifier = this._getPreferredContentViewIdentifier();
+            this._showContentViewForIdentifier(identifier);
+        })
+        .catch((error) => {
+            WI.reportInternalError(error);
+        });
```
dcrousso, line 89: are there actually any realistic errors possible here (i.e. not counting syntax errors or whatnot since those should be fixed right now)?

we usually only use `WI.reportInternalError` for things like protocol stuff since that can actually have problems due to async comms

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -463,6 +468,41 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
...
+
+        let preferences = WI.settings.resourceContentViewIdentifierForMIMEType.value;
+
+        if (identifier === ResourceClusterContentView.Identifier.Response)
+            delete preferences[mimeType];
+        else
+            preferences[mimeType] = identifier;
+
+        // `preferences` is a referenced object. Changing its contents and simply setting it as a new value won't qualify as value change so it won't be persisted.
+        WI.settings.resourceContentViewIdentifierForMIMEType.value = structuredClone(preferences);
```
dcrousso, line 503: we actually have `WI.Setting.prototype.save` just for this purpose

more importantly tho, i kinda feel like it'd be better to have a separate `WI.Setting` for each mime type instead of stuffing it all onto an object so that each time we read from it (or especially when we write to it) we dont have to deserialize (or serialize)

additionally, since this is something pretty specific to `WI.ResourceClusterContentView` id have it be limited in scope to that class
```js
WI.ResourceClusterContentView._preferredContentViewIdentifierSettingForMIMEType ??= new Map;
let setting = WI.ResourceClusterContentView._preferredContentViewIdentifierSettingForMIMEType.getOrInsertComputed(mimeType, () => new WI.Setting("resource-current-view-" + mimeType, null));
```

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -503,13 +543,13 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
     _tryEnableCustomResponseContentViews()
     {
         if (!this._resource.hasResponse())
-            return;
+            return Promise.resolve();
```
dcrousso, line 546: i realize it'd keep the diff cleaner to do it this way, but why not make the entire function `async` so we can deindent everything?

dcrousso (commented):
sorry for the late review!  just a few suggestions if you dont mind a followup :)

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: > (i.e. not counting syntax errors or whatnot since those should be fixed right now)?
Aside from those potential errors, no.

There was a `Promise.reject()` return in an earlier version of the patch. It felt wrong not to `catch()` that. But that's removed now.

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: Actually, we don't need `_shownInitialContent` and `attached()` either.

Showing a `WI.ResourceClusterContentView` **always** calls `WI.ResourceClusterContentView .prototype.restoreFromCookie()`.

The code flow is via `WI.ContentViewContainer.showBackForwardEntryForIndex()` -> `_showEntry()` which creates the view and adds it as a subview (that automatically calls `attached()`)

Screenshot 2026-02-25 at 15 32 03

then calls `restoreFromCookie()` on it a bit later in the call chain.

Screenshot 2026-02-25 at 15 32 47

It turns out both `attached()` and `restoreFromCookie()` are called every time. We can remove `attached()`. 

Since the cookie is _always_ empty for responses, "restoring from cookie" will use any saved preferred content view.

The only time the cookie is set with `WI.ResourceClusterContentView.ContentViewIdentifierCookieKey` is for explicitly showing the request:

https://github.com/WebKit/WebKit/blob/025460e0fccc6e0c4e20c23c29c4b432d7a97db2/Source/WebInspectorUI/UserInterface/Base/Main.js#L1532

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: >we actually have WI.Setting.prototype.save just for this purpose

I saw it used from the `WI.Setting` value setter and standalone from places like:

https://github.com/WebKit/WebKit/blob/025460e0fccc6e0c4e20c23c29c4b432d7a97db2/Source/WebInspectorUI/UserInterface/Views/TabBrowser.js#L361

But both imply knowing that the value is an object that, when its properties change, won't pass the check in the setter:

```
if (this._value === value)
    return;
```

It actually tripped me for a few minutes. So I preferred to make it explicit with `structuredClone()` and a comment. 

> more importantly tho, i kinda feel like it'd be better to have a separate WI.Setting for each mime type instead of stuffing it all onto an object so that each time we read from it (or especially when we write to it) we dont have to deserialize (or serialize)

That's a nice idea!

I don't expect this object to grow beyond a few entries, a few tens at worst (how many mime-types can we expect to encounter? 😅). But having individual key/value settings per mime type is more elegant and removes the need to know about special `save()` handling of that object. 

Will add it in a follow-up.

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: That's right. I'll add it in a follow-up.

PR #59107, opened 2026-02-20
Remove non-standard `display: block-ruby`

weinig:
#### b8670f8cdf8e4bc2c211b2ea4c5b8dbf59284a8b
<pre>
Remove non-standard `display: block-ruby`
<a href="https://bugs.webkit.org/show_bug.cgi?id=308317">https://bugs.webkit.org/show_bug.cgi?id=308317</a>

Reviewed by Tim Nguyen.

Remove support for parsing and serializing display: `block-ruby`. It should
always be represented as the space separated pair `block ruby`.

Also added more conservative support for `display` CSSValueConversion, so
any valid pair that might make its way in from the CSS typedom gets handled
correctly.

To test this, a new off by default setting, CSSRubyDisplayTypesInAuthorStylesEnabled,
was added to enable use of ruby display types outside of UA style sheets.

Test: fast/ruby/ruby-display-types.html
* LayoutTests/fast/ruby/ruby-display-types-expected.txt: Added.
* LayoutTests/fast/ruby/ruby-display-types.html: Added.
* Source/WebCore/css/CSSValueKeywords.in:
* Source/WebCore/css/parser/CSSParserContext.cpp:
* Source/WebCore/css/parser/CSSParserContext.h:
* Source/WebCore/css/parser/CSSPropertyParserConsumer+Display.cpp:
* Source/WebCore/css/parser/CSSPropertyParserConsumer+Display.h:
* Source/WebCore/style/StyleAdjuster.cpp:
* Source/WebCore/style/values/display/StyleDisplay.cpp:
* Source/WebCore/style/values/display/StyleDisplay.h:

Canonical link: <a href="https://commits.webkit.org/308000@main">https://commits.webkit.org/308000@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/d6b2e21b0a164a1c84ac8702d36fc49314b8cee0

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/146187 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/18864 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/11437 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/154856 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/99653 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/148062 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/19336 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/18759 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/112486 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/80474 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/149150 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/14836 "Passed tests") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/131345 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/93357 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/14102 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/11864 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/2302 "Built successfully") | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/138158 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/123668 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/8546 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/157175 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/6979 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/346 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/9886 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/120509 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/18682 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/15647 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/120810 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/30949 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/18702 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/130095 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/74410 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/16488 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/7701 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/177486 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/18303 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/82054 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/45546 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/18035 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/18201 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/18092 "Built successfully") | | |

### DIFF
--- a/LayoutTests/fast/ruby/ruby-display-types.html
+++ b/LayoutTests/fast/ruby/ruby-display-types.html
@@ -1,0 +1,29 @@
+<!DOCTYPE html><!-- webkit-test-runner [ CSSRubyDisplayTypesInAuthorStylesEnabled=true ] -->
+<html>
+<head>
+  <script src="../../resources/testharness.js"></script>
+  <script src="../../resources/testharnessreport.js"></script>
+</head>
+<body>
+<script>
+function test_ruby_display_type(specified, expected) {
+  test((t) => {
+    const target = document.createElement("div");
+    document.body.append(target);
+    t.add_cleanup(() => {
+      target.remove();
+    });
+    target.style['display'] = specified;
+    assert_equals(target.style['display'], expected);
+    assert_equals(window.getComputedStyle(target)['display'], expected);
+  }, `display: ${specified}`);
+}
+
+test_ruby_display_type('block ruby', 'block ruby');
+test_ruby_display_type('inline ruby', 'ruby');
+test_ruby_display_type('ruby', 'ruby');
+test_ruby_display_type('ruby-base', 'ruby-base');
+test_ruby_display_type('ruby-text', 'ruby-text');
+</script>
+</body>
+</html>
--- a/Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
+++ b/Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
@@ -1375,6 +1375,20 @@ CSSRhythmicSizingEnabled:
     WebCore:
       default: false
 
+CSSRubyDisplayTypesInAuthorStylesEnabled:
+  type: bool
+  status: unstable
+  category: css
+  humanReadableName: "CSS Ruby Display Types in Author Styles"
+  humanReadableDescription: "Enable CSS Ruby Display types in author styles"
+  defaultValue:
+    WebKitLegacy:
+      default: false
+    WebKit:
+      default: false
+    WebCore:
+      default: false
+
 CSSScrollAnchoringEnabled:
   type: bool
   status: stable
--- a/Source/WebCore/css/parser/CSSParserContext.cpp
+++ b/Source/WebCore/css/parser/CSSParserContext.cpp
@@ -48,6 +48,7 @@ static void NODELETE applyUASheetBehaviorsToContext(CSSParserContext& context)
 {
     // FIXME: We should turn all of the features on from their WebCore Settings defaults.
     context.cssAppearanceBaseEnabled = true;
+    context.cssRubyDisplayTypesEnabled = true;
     context.cssTextTransformMathAutoEnabled = true;
     context.popoverAttributeEnabled = true;
     context.propertySettings.cssInputSecurityEnabled = true;
@@ -113,6 +114,7 @@ CSSParserContext::CSSParserContext(const Settings& settings)
     , targetTextPseudoElementEnabled { settings.targetTextPseudoElementEnabled() }
     , htmlEnhancedSelectEnabled { settings.htmlEnhancedSelectEnabled() }
     , cssRandomFunctionEnabled { settings.cssRandomFunctionEnabled() }
+    , cssRubyDisplayTypesEnabled { settings.cssRubyDisplayTypesInAuthorStylesEnabled() }
     , cssTreeCountingFunctionsEnabled { settings.cssTreeCountingFunctionsEnabled() }
     , cssURLModifiersEnabled { settings.cssURLModifiersEnabled() }
     , cssURLIntegrityModifierEnabled { settings.cssURLIntegrityModifierEnabled() }
@@ -152,6 +154,7 @@ void add(Hasher& hasher, const CSSParserContext& context)
         context.targetTextPseudoElementEnabled,
         context.htmlEnhancedSelectEnabled,
         context.cssRandomFunctionEnabled,
+        context.cssRubyDisplayTypesEnabled,
         context.cssTreeCountingFunctionsEnabled,
         context.cssURLModifiersEnabled,
         context.cssURLIntegrityModifierEnabled,
--- a/Source/WebCore/css/parser/CSSPropertyParserConsumer+Display.cpp
+++ b/Source/WebCore/css/parser/CSSPropertyParserConsumer+Display.cpp
@@ -39,43 +39,7 @@
 namespace WebCore {
 namespace CSSPropertyParserHelpers {
 
-enum class DisplayOutside : uint8_t { NoOutside, Block, Inline };
-enum class DisplayInside  : uint8_t { NoInside, Flow, FlowRoot, Table, Flex, Grid, GridLanes, Ruby };
-
-} // namespace CSSPropertyParserHelpers
-} // namespace WebCore
-
-namespace WTF {
-
-template<> struct EnumTraits<WebCore::CSSPropertyParserHelpers::DisplayOutside> {
-    using values = EnumValues<
-        WebCore::CSSPropertyParserHelpers::DisplayOutside,
-        WebCore::CSSPropertyParserHelpers::DisplayOutside::NoOutside,
-        WebCore::CSSPropertyParserHelpers::DisplayOutside::Block,
-        WebCore::CSSPropertyParserHelpers::DisplayOutside::Inline
-    >;
-};
-
-template<> struct EnumTraits<WebCore::CSSPropertyParserHelpers::DisplayInside> {
-    using values = EnumValues<
-        WebCore::CSSPropertyParserHelpers::DisplayInside,
-        WebCore::CSSPropertyParserHelpers::DisplayInside::NoInside,
-        WebCore::CSSPropertyParserHelpers::DisplayInside::Flow,
-        WebCore::CSSPropertyParserHelpers::DisplayInside::FlowRoot,
-        WebCore::CSSPropertyParserHelpers::DisplayInside::Table,
-        WebCore::CSSPropertyParserHelpers::DisplayInside::Flex,
-        WebCore::CSSPropertyParserHelpers::DisplayInside::Grid,
-        WebCore::CSSPropertyParserHelpers::DisplayInside::GridLanes,
-        WebCore::CSSPropertyParserHelpers::DisplayInside::Ruby
-    >;
-};
-
-} // namespace WTF
-
-namespace WebCore {
-namespace CSSPropertyParserHelpers {
-
-using DisplayOutsideInsideMap = EnumeratedArray<DisplayOutside, EnumeratedArray<DisplayInside, CSSValueID>>;
+using DisplayOutsideInsideMap = EnumeratedArray<DisplayOutside, EnumeratedArray<DisplayInside, std::pair<CSSValueID, CSSValueID>>>;
 
 consteval DisplayOutsideInsideMap NODELETE makeDisplayOutsideInsideMap()
 {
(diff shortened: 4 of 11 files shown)


Source/WebCore/css/parser/CSSPropertyParserConsumer+Display.cpp
```diff
@@ -111,14 +75,14 @@ consteval DisplayOutsideInsideMap NODELETE makeDisplayOutsideInsideMap()
...
-    result[Inline][Grid]         = CSSValueInlineGrid;
-    result[Inline][GridLanes]    = CSSValueInlineGridLanes;
-    result[Inline][Ruby]         = CSSValueRuby;
+    result[Inline][NoInside]     = { CSSValueInline,          CSSValueInvalid };
+    result[Inline][Flow]         = { CSSValueInline,          CSSValueInvalid };
+    result[Inline][FlowRoot]     = { CSSValueInlineBlock,     CSSValueInvalid };
+    result[Inline][Table]        = { CSSValueInlineTable,     CSSValueInvalid };
+    result[Inline][Flex]         = { CSSValueInlineFlex,      CSSValueInvalid };
+    result[Inline][Grid]         = { CSSValueInlineGrid,      CSSValueInvalid };
+    result[Inline][GridLanes]    = { CSSValueInlineGridLanes, CSSValueInvalid };
```
nt1m, line 84: For inline-grid-lanes, there’s no backwards compat to worry about, so we generally want to prefer the 2 value syntax serialization.

Source/WebCore/css/parser/CSSPropertyParserConsumer+Display.cpp
```diff
@@ -290,7 +267,7 @@ RefPtr<CSSValue> consumeDisplay(CSSParserTokenRange& range, CSS::PropertyParserS
         return CSSPrimitiveValue::create(range.consumeIncludingWhitespace().id());
     case CSSValueRubyBase:
     case CSSValueRubyText:
-        if (!isUASheetBehavior(state.context.mode))
+        if (!isUASheetBehavior(state.context.mode) && !state.context.cssRubyDisplayTypesInAuthorStylesEnabled)
```
nt1m: You can remove the UA sheet check and just enable the new pref in applyUASheetBehaviorsToContext

Source/WebCore/css/parser/CSSPropertyParserConsumer+Display.h
```diff
@@ -38,9 +39,39 @@ struct PropertyParserState;
...
 // MARK: <'display'> consuming
 // https://drafts.csswg.org/css-display/#propdef-display
 RefPtr<CSSValue> consumeDisplay(CSSParserTokenRange&, CSS::PropertyParserState&);
 
 } // namespace CSSPropertyParserHelpers
 } // namespace WebCore
+
+namespace WTF {
+
+template<> struct EnumTraits<WebCore::CSSPropertyParserHelpers::DisplayOutside> {
```
nt1m, line 54: There’s quite a bit of duplication that feels quite unfortunate. Can we do anything about it?

Source/WebCore/style/StyleAdjuster.cpp
```diff
@@ -335,8 +335,11 @@ static bool shouldInlinifyForRuby(const RenderStyle& style, const RenderStyle& p
     return hasRubyParent && !style.hasOutOfFlowPosition() && style.floating() == Float::None;
 }
 
-static bool hasUnsupportedRubyDisplay(Display display, const Element* element)
+static bool hasUnsupportedRubyDisplay(Display display, const Element* element, const Document& document)
 {
+    if (document.settings().cssRubyDisplayTypesInAuthorStylesEnabled())
```
nt1m, line 341: You shouldn’t need to pass the document in this method. Either there’s no element, and true is returned. Either the pref is off, and true is returned (e.g. the display type is unsupported), and you can get the doc from the element.

Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
```diff
@@ -1375,6 +1375,20 @@ CSSRhythmicSizingEnabled:
     WebCore:
       default: false
 
+CSSRubyDisplayTypesInAuthorStylesEnabled:
+  type: bool
+  status: unstable
```
nt1m, line 1380: internal maybe? Like the internal-auto-based pref

Source/WebCore/css/parser/CSSPropertyParserConsumer+Display.h
weinig: If I think of a way that doesn't result in more code, I will update it.

Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
weinig: Seems like something we will want to turn on one day when it is more fully supported. Why would internal be better?

Source/WebCore/style/StyleAdjuster.cpp
weinig: Ok.

weinig: Actually, I think I want to 

"return false" when:
   - display: block ruby
   - there is no element 
   - the setting is enabled

"return true" when:
   - display: block ruby
   - there is no element 
   - the setting is disabled.

So I need some way to get the setting when the element is not available.

Source/WebCore/css/parser/CSSPropertyParserConsumer+Display.cpp
weinig: Ok.

Source/WebCore/css/parser/CSSPropertyParserConsumer+Display.cpp
weinig: Not changing that in this PR.

Source/WebCore/css/parser/CSSPropertyParserConsumer+Display.h
weinig: Wait, do you just mean the EnumTraits? That's just how all enum traits are.

Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
nt1m: Hmm, ok I could be convinced either way. Usually we tend to do `testable` or `internal`. `unstable` only when we want to hide it from users or disable in tests.

Source/WebCore/css/parser/CSSParserContext.cpp
```diff
@@ -48,6 +48,7 @@ static void NODELETE applyUASheetBehaviorsToContext(CSSParserContext& context)
 {
     // FIXME: We should turn all of the features on from their WebCore Settings defaults.
     context.cssAppearanceBaseEnabled = true;
+    context.cssRubyDisplayTypesInAuthorStylesEnabled = true;
```
nt1m: Makes more sense to name the pref `CSSRubyDisplayTypesEnabled` now that we're calling it from this method. (the author styles part also doesn't matter once we display the pref in the feature flags pane)

Source/WebCore/style/StyleAdjuster.cpp
nt1m: I'm not sure you want to `return false` when the setting is enabled. We adjust the style here precisely because the layout engine doesn't support it. `return false` means we signal the layout engine that it's always support the value.

As a result, I still think we should just grab the doc from the element, and `return true` when the setting is enabled.

nt1m (approved):

Source/WebCore/style/StyleAdjuster.cpp
weinig: That’s exactly what I want here. The point of the check is to make it available as the computed style.

LayoutTests/fast/ruby/ruby-display-types.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html><!-- webkit-test-runner [ CSSRubyDisplayTypesInAuthorStylesEnabled=true ] -->
```
nt1m, line 1:
```suggestion
<!DOCTYPE html><!-- webkit-test-runner [ CSSRubyDisplayTypesEnabled=true ] -->
```

PR #59155, opened 2026-02-21
Support user agent part pseudo-elements in getComputedStyle()

annevk:
#### f2441fbb57d8bc69d7c21c18f9e0dbb19d4627b0
<pre>
Support user agent part pseudo-elements in getComputedStyle()
<a href="https://bugs.webkit.org/show_bug.cgi?id=308380">https://bugs.webkit.org/show_bug.cgi?id=308380</a>

Reviewed by Antti Koivisto.

This enables support for ::picker(select) and ::file-selector-button
and the like, while not adding support for non-standard
pseudo-elements.

We do this by eagerly looking up the relevant element in the shadow
tree before creating a CSSComputedStyleDeclaration instance, but when
there is no such element we make use of the new StandardUserAgentPart
to pick up author-defined styles which need to be reflected as shown in
the various tests.

New test:

    <a href="https://github.com/web-platform-tests/wpt/pull/57959">https://github.com/web-platform-tests/wpt/pull/57959</a>
    <a href="https://github.com/web-platform-tests/wpt/pull/58009">https://github.com/web-platform-tests/wpt/pull/58009</a>

Canonical link: <a href="https://commits.webkit.org/308118@main">https://commits.webkit.org/308118@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/8714896f9f5f635a8ec60dea4807758bdb7fe77e

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/146492 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/19165 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/12674 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/155156 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/99895 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/697939a4-bcec-4f95-b350-78c007cfc6d8) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/148367 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/19626 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/19069 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/112795 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/99895 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [❌ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/9f7fb18d-a3f0-4814-8f81-b20cd63b08b0) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/149455 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/15096 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/131634 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/93609 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/1962ec91-6a8b-45d6-ba23-eb84bc1d571f) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/14357 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/12121 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/2600 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/123928 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/9500 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/157480 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/633 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/10931 "Passed tests") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/1/builds/120835 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/18969 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/15920 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/121076 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/18983 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/131252 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/74772 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/22604 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/16724 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/8161 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/18590 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/82340 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/18319 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/18475 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/18377 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/css/SelectorChecker.cpp
+++ b/Source/WebCore/css/SelectorChecker.cpp
@@ -202,7 +202,7 @@ void SelectorChecker::CheckingContext::setRequestedPseudoElement(Style::PseudoEl
 {
     hasRequestedPseudoElement = true;
     pseudoElementType = pseudoElementIdentifier.type;
-    pseudoElementNameArgument = pseudoElementIdentifier.nameArgument;
+    pseudoElementNameArgument = pseudoElementIdentifier.nameOrPart;
 }
 
 std::optional<Style::PseudoElementIdentifier> SelectorChecker::CheckingContext::requestedPseudoElement() const
@@ -342,6 +342,13 @@ SelectorChecker::MatchResult SelectorChecker::matchRecursively(CheckingContext& 
             // In functional pseudo-classes like :is()/where(), pseudo-elements are always disabled.
             if (context.inFunctionalPseudoClass)
                 return MatchResult::fails(Match::SelectorFailsCompletely);
+            if (context.requestedPseudoElement && context.requestedPseudoElement->type == PseudoElementType::UserAgentPartFallback) {
+                if (context.selector->value() != context.requestedPseudoElement->nameOrPart)
+                    return MatchResult::fails(Match::SelectorFailsLocally);
+                collectedPseudoElements.add(PseudoElementType::UserAgentPartFallback);
+                matchType = MatchType::VirtualPseudoElementOnly;
+                break;
+            }
             if (CheckedPtr root = context.element->containingShadowRoot()) {
                 if (root->mode() != ShadowRootMode::UserAgent)
                     return MatchResult::fails(Match::SelectorFailsLocally);
@@ -356,14 +363,20 @@ SelectorChecker::MatchResult SelectorChecker::matchRecursively(CheckingContext& 
             // In functional pseudo-classes like :is()/where(), pseudo-elements are always disabled.
             if (context.inFunctionalPseudoClass)
                 return MatchResult::fails(Match::SelectorFailsCompletely);
+            auto* stringList = context.selector->stringList();
+            ASSERT(stringList && !stringList->isEmpty());
+            auto part = makeString("picker("_s, stringList->at(0), ')');
+            if (context.requestedPseudoElement && context.requestedPseudoElement->type == PseudoElementType::UserAgentPartFallback) {
+                if (part != context.requestedPseudoElement->nameOrPart)
+                    return MatchResult::fails(Match::SelectorFailsLocally);
+                collectedPseudoElements.add(PseudoElementType::UserAgentPartFallback);
+                matchType = MatchType::VirtualPseudoElementOnly;
+                break;
+            }
             CheckedPtr root = context.element->containingShadowRoot();
             if (!root || root->mode() != ShadowRootMode::UserAgent)
                 return MatchResult::fails(Match::SelectorFailsLocally);
 
-            auto* stringList = context.selector->stringList();
-            ASSERT(stringList && !stringList->isEmpty());
-
-            auto part = makeString("picker("_s, stringList->at(0), ')');
             if (context.element->userAgentPart() != part)
                 return MatchResult::fails(Match::SelectorFailsLocally);
 
@@ -504,11 +517,15 @@ SelectorChecker::MatchResult SelectorChecker::matchRecursively(CheckingContext& 
             return MatchResult::updateWithMatchType(result, matchType);
         }
     case CSSSelector::Relation::ShadowDescendant:  {
-        CheckedPtr host = context.element->shadowHost();
-        if (!host)
-            return MatchResult::fails(Match::SelectorFailsCompletely);
-
-        nextContext.element = host.get();
+        // When matching a user-agent-part pseudo-element via style resolution (no backing
+        // element found), the element is the originating element itself. Skip the shadow host
+        // traversal and match the rest of the selector against it directly.
+        if (!collectedPseudoElements.contains(PseudoElementType::UserAgentPartFallback)) {
+            CheckedPtr host = context.element->shadowHost();
+            if (!host)
+                return MatchResult::fails(Match::SelectorFailsCompletely);
+            nextContext.element = host.get();
+        }
         nextContext.firstSelectorOfTheFragment = nextContext.selector;
         nextContext.isSubjectOrAdjacentElement = false;
         EnumSet<PseudoElementType> ignoredPseudoElements;
@@ -1356,7 +1373,7 @@ bool SelectorChecker::checkOne(CheckingContext& checkingContext, LocalContext& c
                 return true;
             if (requestedPseudoElement->type != PseudoElementType::Highlight || !selector.stringList())
                 return false;
-            return selector.stringList()->first() == requestedPseudoElement->nameArgument;
+            return selector.stringList()->first() == requestedPseudoElement->nameOrPart;
 
         case CSSSelector::PseudoElement::ViewTransitionGroup:
         case CSSSelector::PseudoElement::ViewTransitionImagePair:
@@ -1370,7 +1387,7 @@ bool SelectorChecker::checkOne(CheckingContext& checkingContext, LocalContext& c
 
             auto& list = *selector.stringList();
             auto& name = list.first();
-            if (name != starAtom() && name != requestedPseudoElement->nameArgument)
+            if (name != starAtom() && name != requestedPseudoElement->nameOrPart)
                 return false;
 
             if (list.size() == 1)
@@ -1674,7 +1691,7 @@ bool SelectorChecker::checkViewTransitionPseudoClass(const CheckingContext& chec
         return false;
     case PseudoElementType::ViewTransitionGroup: {
         if (RefPtr activeViewTransition = element.document().activeViewTransition()) {
-            if (activeViewTransition->namedElements().find(pseudoIdentifier->nameArgument))
+            if (activeViewTransition->namedElements().find(pseudoIdentifier->nameOrPart))
                 return activeViewTransition->namedElements().size() == 1;
         }
         return false;
@@ -1683,14 +1700,14 @@ bool SelectorChecker::checkViewTransitionPseudoClass(const CheckingContext& chec
         return true;
     case PseudoElementType::ViewTransitionOld: {
         if (RefPtr activeViewTransition = element.document().activeViewTransition()) {
-            if (auto* capturedElement = activeViewTransition->namedElements().find(pseudoIdentifier->nameArgument))
+            if (auto* capturedElement = activeViewTransition->namedElements().find(pseudoIdentifier->nameOrPart))
                 return !capturedElement->newElement;
         }
         return false;
     }
     case PseudoElementType::ViewTransitionNew: {
         if (RefPtr activeViewTransition = element.document().activeViewTransition()) {
-            if (auto* capturedElement = activeViewTransition->namedElements().find(pseudoIdentifier->nameArgument))
+            if (auto* capturedElement = activeViewTransition->namedElements().find(pseudoIdentifier->nameOrPart))
                 return !capturedElement->oldImage;
         }
         return false;
--- a/Source/WebCore/css/parser/CSSSelectorParser.cpp
+++ b/Source/WebCore/css/parser/CSSSelectorParser.cpp
@@ -807,6 +807,16 @@ std::unique_ptr<MutableCSSSelector> CSSSelectorParser::consumeAttribute(CSSParse
     return selector;
 }
 
+static AtomString consumePickerArgument(CSSParserTokenRange& block)
+{
+    auto& ident = block.consumeIncludingWhitespace();
+    if (ident.type() != IdentToken || !block.atEnd())
+        return nullAtom();
+    if (ident.value() != "select"_s)
+        return nullAtom();
+    return ident.value().toAtomString();
+}
+
 std::unique_ptr<MutableCSSSelector> CSSSelectorParser::consumePseudo(CSSParserTokenRange& range)
 {
     ASSERT(range.peek().type() == ColonToken);
@@ -1001,12 +1011,8 @@ std::unique_ptr<MutableCSSSelector> CSSSelectorParser::consumePseudo(CSSParserTo
         }
 
         case CSSSelector::PseudoElement::Picker: {
-            auto& ident = block.consumeIncludingWhitespace();
-            if (ident.type() != IdentToken || !block.atEnd())
-                return nullptr;
-
-            auto argument = ident.value().toAtomString();
-            if (argument != "select"_s)
+            auto argument = consumePickerArgument(block);
+            if (argument.isNull())
                 return nullptr;
 
             selector->setStringList({ { argument } });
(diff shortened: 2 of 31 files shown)


Source/WebCore/css/parser/CSSSelectorParser.cpp
```diff
@@ -1505,6 +1511,11 @@ std::pair<bool, std::optional<Style::PseudoElementIdentifier>> CSSSelectorParser
             return { };
         return { true, Style::PseudoElementIdentifier { *CSSSelector::stylePseudoElementTypeFor(*pseudoElement), ident.value().toAtomString() } };
     }
+    case CSSSelector::PseudoElement::Picker: {
+        if (consumePickerArgument(block).isNull())
+            return { };
+        return { true, std::nullopt, true };
+    }
```
nt1m, line 1533: I'd prefer using the dot syntax so we can distinguish the arguments:
```
{
   .isParsable = true,
   .pseudoElementIdentifier = ...
   .isPickerSelect = true
}
```

Source/WebCore/css/parser/CSSSelectorParser.h
```diff
@@ -44,9 +45,11 @@ class CSSSelectorList;
 class StyleSheetContents;
 class StyleRule;
 
-namespace Style {
-struct PseudoElementIdentifier;
-}
+struct ParsedPseudoElement {
+    bool isParsable { false };
+    std::optional<Style::PseudoElementIdentifier> identifier;
+    bool isPickerSelect { false };
```
nt1m: Since these are 3 mutually exclusive states, I think we should just do: `using ParsedPseudoElement = Variant<NotParsable, std::optional<Style::PseudoElementIdentifier>, IsPickerSelect>` (or something like that). Then the code can use `WTF::switchOn()` which should be easier to read.

If we generalize this to all user agent parts, we could have `Variant<NotParsable, std::optional<Style::PseudoElementIdentifier>, AtomString>`

weinig:
I think it is a mistake to special case this one selector, as it usually leads to more complications down the road for people less familiar with the code. Putting checks for specific pseudo selector in LocalDOMWindow seems particularly objectionable.

What would it take to do this in a way that doesn't require special casing? What data structures need to be updated?

Source/WebCore/page/LocalDOMWindow.cpp
```diff
@@ -1651,16 +1653,26 @@ StyleMedia& LocalDOMWindow::styleMedia()
     return *m_media;
 }
 
+static Element* NODELETE pickerSelectPseudoElement(Element& element)
+{
+    if (auto* selectElement = dynamicDowncast<HTMLSelectElement>(element))
+        return selectElement->pickerPopoverElement();
+    return nullptr;
+}
```
nt1m: Seems possible to generalize this to all user agent parts I feel like. Where we do a descendant walk and find the element with the right user agent part, and early return if there's no UA shadow tree. We could short circuit for ::picker(select). 

And regarding parsing, you pretty much have the UA part string as input already (just need to strip the `::`)

Source/WebCore/css/parser/CSSSelectorParser.cpp
annevk: I've seen this break on GCC.

Source/WebCore/page/LocalDOMWindow.cpp
annevk: But do we want to support this for all user agent parts? I think it makes sense to support it for the standardized ones, but I'm not sure we should generalize it.

Source/WebCore/css/parser/CSSSelectorParser.cpp
nt1m: It breaks only when you omit arguments.

Source/WebCore/page/LocalDOMWindow.cpp
nt1m: It shouldn't be hard to allowlist to the standardized one if we want to

nt1m: Rejecting prefixed ones should be an accurate way to do so

annevk: I think I'm going with a safelist for now since what if we decide that there are multiple in a single user agent shadow tree? Probably requires some case-by-case decision making.

Source/WebCore/dom/Element.cpp
```diff
@@ -4739,7 +4739,7 @@ const RenderStyle* Element::resolveComputedStyle(ResolveComputedStyleMode mode)
     return computedStyle;
 }
 
-const RenderStyle& Element::resolvePseudoElementStyle(const Style::PseudoElementIdentifier& pseudoElementIdentifier)
+const RenderStyle* Element::resolvePseudoElementStyle(const Style::PseudoElementIdentifier& pseudoElementIdentifier)
```
nt1m, line 4749: This is only used by internal callers. We should either assert this:
`ASSERT(pseudoElementIdentifier.type != PseudoElementType::StandardUserAgentPart)` or change this to support UA parts (and potentially use it in `getComputedStyle`)

We should definitely not return nullptr though.

Source/WebCore/css/parser/CSSSelectorParser.cpp
```diff
@@ -1481,8 +1487,15 @@ std::pair<bool, std::optional<Style::PseudoElementIdentifier>> CSSSelectorParser
     if (token.type() == IdentToken) {
         range.consume();
         if (!range.atEnd() || CSSSelector::pseudoElementRequiresArgument(*pseudoElement))
-            return { };
-        return { true, pseudoElementIdentifierFor(*pseudoElement) };
+            return std::nullopt;
+        auto styleId = pseudoElementIdentifierFor(*pseudoElement);
+        if (styleId)
+            return styleId;
```
nt1m: nit: please always name these variables `pseudoElementIdentifier` to match the rest of the code base.

Source/WebCore/style/PseudoElementUtilities.h
```diff
@@ -0,0 +1,43 @@
...
+#include <wtf/RefPtr.h>
+
+namespace WebCore {
+
+class Element;
+
+struct ResolvedComputedPseudoElement {
+    RefPtr<Element> element { nullptr };
+    std::optional<Style::PseudoElementIdentifier> identifier { std::nullopt };
+};
```
nt1m, line 41: This is pretty much the Styleable interface, but maybe we shouldn't use it to avoid confusion about different ways UA parts invoke Styleable (sometimes it uses the UA part element directly).

Source/WebCore/page/LocalDOMWindow.cpp
```diff
@@ -1656,36 +1657,36 @@ Ref<CSSStyleDeclaration> LocalDOMWindow::getComputedStyle(Element& element, cons
...
 {
     if (!isCurrentlyDisplayedInFrame())
         return nullptr;
 
-    // FIXME: This parser context won't get the right settings without a document.
-    auto parserContext = document() ? CSSSelectorParserContext { *protect(document()) } : CSSSelectorParserContext { CSSParserContext { HTMLStandardMode } };
-    auto [pseudoElementIsParsable, pseudoElementIdentifier] = CSSSelectorParser::parsePseudoElement(pseudoElement, parserContext);
-    if (!(pseudoElementIsParsable || (pseudoElementIdentifier && !pseudoElementIdentifier->nameArgument.isNull())) && !pseudoElement.isEmpty())
+    Ref document = *this->document();
+    auto pseudoId = CSSSelectorParser::parsePseudoElement(pseudoElement, CSSSelectorParserContext { document });
```
nt1m: ditto here, we should name this `pseudoElementIdentifier`

Source/WebCore/style/PseudoElementIdentifier.h
```diff
@@ -35,22 +35,23 @@ namespace WebCore::Style {
 struct PseudoElementIdentifier {
     PseudoElementType type;
 
-    // highlight name for ::highlight or view transition name for view transition pseudo elements.
-    AtomString nameArgument { nullAtom() };
+    // Highlight name for ::highlight, view transition name for view transition pseudo-elements,
+    // or user agent part name for StandardUserAgentPart (e.g., "file-selector-button", "picker(select)").
+    AtomString nameOrPart { nullAtom() };
```
nt1m, line 40: nit: Maybe it should just be `name`. `nameOrPart` doesn't add much

Source/WebCore/rendering/style/RenderStyleConstants.h
```diff
@@ -71,6 +71,8 @@ enum class PseudoElementType : uint8_t {
     ViewTransitionOld,
     ViewTransitionNew,
 
+    StandardUserAgentPart,
```
nt1m: We should ASSERT at all the methods which take an PseudoElementIdentifier where this is not used (because they take the UA part element instead)

Source/WebCore/css/SelectorChecker.cpp
nt1m, line 1: I'm not familiar with the changes in this file, would be great to get Antti to review this part

Source/WebCore/dom/Element.cpp
annevk: This is needed to support `getComputedStyle(someDiv, '::file-selector-button')`. In this case we need to ask the style system for styles, but when there are none, we need to return an empty `CSSStyleDeclaration`. Asserting here would make tests crash.

Source/WebCore/style/PseudoElementIdentifier.h
annevk: I like the hint it provides about its dual purpose. We have done this elsewhere too.

Source/WebCore/dom/Element.cpp
nt1m: I guess my question is: how we end up when we end up hitting this branch ?

```
        if (pseudoElementIdentifier.type == PseudoElementType::StandardUserAgentPart)
            return nullptr;
```

Source/WebCore/style/PseudoElementIdentifier.h
anttijk: @weinig would probably use `Variant<>` to make only legal combos expressible but this seems fine for now.

Source/WebCore/style/PseudoElementUtilities.h
anttijk, line 41: This should be in `Style` namespace like PseudoElementIdentifier and in WebCore/style directory.

Source/WebCore/page/LocalDOMWindow.cpp
```diff
@@ -1656,36 +1657,36 @@ Ref<CSSStyleDeclaration> LocalDOMWindow::getComputedStyle(Element& element, cons
...
-    auto parserContext = document() ? CSSSelectorParserContext { *protect(document()) } : CSSSelectorParserContext { CSSParserContext { HTMLStandardMode } };
-    auto [pseudoElementIsParsable, pseudoElementIdentifier] = CSSSelectorParser::parsePseudoElement(pseudoElement, parserContext);
-    if (!(pseudoElementIsParsable || (pseudoElementIdentifier && !pseudoElementIdentifier->nameArgument.isNull())) && !pseudoElement.isEmpty())
+    Ref document = *this->document();
+    auto pseudoElementIdentifier = CSSSelectorParser::parsePseudoElement(pseudoElement, CSSSelectorParserContext { document });
+    // Reject new features, such as ::highlight(name) and ::file-selector-button.
+    if (pseudoElementIdentifier && !pseudoElementIdentifier->nameOrPart.isNull())
+        pseudoElementIdentifier = std::nullopt;
+    if (!pseudoElementIdentifier && !pseudoElement.isEmpty())
         return nullptr;
```
anttijk, line 1677: The existing code appears to do something different in the case where pseudo element is parseable but not recognized. Why is it ok to not differentiate this case?

anttijk (approved):

Source/WebCore/style/PseudoElementIdentifier.h
annevk: For this one it seems hard without splitting AtomString. Of course, there is historical precedent for splitting the atom.

Source/WebCore/page/LocalDOMWindow.cpp
annevk: The one known legacy caller for `getMatchedCSSRules` might have disappeared (looking into that) and I'm pretty certain it doesn't care about pseudo-elements at all. It would also require `parsePseudoElement` to return more state if we wanted to fully match the original contract.

PR #59220, opened 2026-02-23
counters evaluation with style containment

islandryu:
#### 5a4d55dbf634942e0e8d338f6c5f4b625395bba4
<pre>
counters evaluation with style containment
<a href="https://bugs.webkit.org/show_bug.cgi?id=308446">https://bugs.webkit.org/show_bug.cgi?id=308446</a>

Reviewed by NOBODY (OOPS!).

Fixed an issue where counters from outside the style containment were not displayed when contain: style was used.

This behavior should be covered by imported/w3c/web-platform-tests/css/css-contain/counter-scoping-001.html and imported/w3c/web-platform-tests/css/css-contain/counter-scoping-002.html

* Source/WebCore/rendering/CounterNode.cpp:
(WebCore::CounterNode::CounterNode):
(WebCore::CounterNode::create):
(WebCore::CounterNode::shouldShowZero const):
* Source/WebCore/rendering/CounterNode.h:
* Source/WebCore/rendering/RenderCounter.cpp:
(WebCore::ancestorStyleContainmentObject):
(WebCore::planCounter):
(WebCore::makeCounterNode):
(WebCore::RenderCounter::originalText const):
</pre>
https://github.com/WebKit/WebKit/commit/5a4d55dbf634942e0e8d338f6c5f4b625395bba4

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/148843 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/21556 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/15125 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/157529 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/102273 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/150716 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/22009 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/21434 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/114745 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/81714 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/151803 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/16942 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/133596 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/95507 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/16053 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/13903 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/5297 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/125654 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/11516 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/159867 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/3003 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/13038 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/122804 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/21358 "Built successfully") | [❌ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/17915 "Found 2 new test failures: accessibility/display-contents/aria-grid.html accessibility/mac/text-operation/text-operation-capitalize.html (failure)") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/123034 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/21366 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/133310 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/77554 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/18328 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/10071 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/20968 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/84770 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/20700 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/20847 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/20756 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/rendering/CounterNode.cpp
+++ b/Source/WebCore/rendering/CounterNode.cpp
@@ -32,10 +32,11 @@
 
 namespace WebCore {
 
-CounterNode::CounterNode(RenderElement& owner, OptionSet<CounterNode::Type> type, int value)
+CounterNode::CounterNode(RenderElement& owner, OptionSet<CounterNode::Type> type, int value, std::optional<Plan> plan)
     : m_type(type)
     , m_value(value)
     , m_owner(owner)
+    , m_plan(plan)
 {
 }
 
@@ -92,9 +93,9 @@ RenderElement& CounterNode::owner() const
     return m_owner.get();
 }
 
-Ref<CounterNode> CounterNode::create(RenderElement& owner, OptionSet<CounterNode::Type> type, int value)
+Ref<CounterNode> CounterNode::create(RenderElement& owner, OptionSet<CounterNode::Type> type, int value, std::optional<Plan> plan)
 {
-    return adoptRef(*new CounterNode(owner, type, value));
+    return adoptRef(*new CounterNode(owner, type, value, plan));
 }
 
 CounterNode* CounterNode::nextInPreOrderAfterChildren(const CounterNode* stayWithin) const
@@ -337,6 +338,15 @@ void CounterNode::removeChild(CounterNode& oldChild)
 
     if (next)
         next->recount();
+}
+
+bool CounterNode::shouldShowZero() const
+{
+    for (auto* node = this; node; node = node->parent()) {
+        if (node->m_plan.has_value())
+            return true;
+    }
+    return false;
 }
 
 #if ENABLE(TREE_DEBUGGING)
--- a/Source/WebCore/rendering/CounterNode.h
+++ b/Source/WebCore/rendering/CounterNode.h
@@ -21,6 +21,7 @@
 
 #pragma once
 
+#include <optional>
 #include <wtf/CheckedPtr.h>
 #include <wtf/Forward.h>
 #include <wtf/OptionSet.h>
@@ -44,12 +45,17 @@ class RenderElement;
 class CounterNode : public RefCounted<CounterNode>, public CanMakeSingleThreadWeakPtr<CounterNode> {
 public:
     enum class Type : uint8_t { Increment, Reset, Set };
+    struct Plan {
+        OptionSet<Type> type;
+        int value { 0 };
+    };
 
-    static Ref<CounterNode> create(RenderElement&, OptionSet<Type>, int value);
+    static Ref<CounterNode> create(RenderElement&, OptionSet<Type>, int value, std::optional<Plan>);
     ~CounterNode();
     bool actsAsReset() const { return hasResetType() || !m_parent; }
     bool hasResetType() const { return m_type.contains(Type::Reset); }
     bool hasSetType() const { return m_type.contains(Type::Set); }
+    bool shouldShowZero() const;
     int value() const { return m_value; }
     int countInParent() const { return m_countInParent; }
     RenderElement& NODELETE owner() const;
@@ -74,7 +80,7 @@ public:
     void removeChild(CounterNode&);
 
 private:
-    CounterNode(RenderElement&, OptionSet<Type>, int value);
+    CounterNode(RenderElement&, OptionSet<Type>, int value, std::optional<Plan>);
     int NODELETE computeCountInParent() const;
     // Invalidates the text in the renderer of this counter, if any,
     // and in the renderers of all descendants of this counter, if any.
@@ -92,6 +98,7 @@ private:
     SingleThreadWeakPtr<CounterNode> m_nextSibling;
     SingleThreadWeakPtr<CounterNode> m_firstChild;
     SingleThreadWeakPtr<CounterNode> m_lastChild;
+    std::optional<Plan> m_plan;
 };
 
 } // namespace WebCore
--- a/Source/WebCore/rendering/RenderCounter.cpp
+++ b/Source/WebCore/rendering/RenderCounter.cpp
@@ -30,6 +30,7 @@
 #include "HTMLNames.h"
 #include "HTMLOListElement.h"
 #include "RenderChildIterator.h"
+#include "RenderElement.h"
 #include "RenderElementStyleInlines.h"
 #include "RenderElementInlines.h"
 #include "RenderListItem.h"
@@ -37,6 +38,7 @@
 #include "RenderStyle.h"
 #include "RenderView.h"
 #include <wtf/StdLibExtras.h>
+#include <wtf/text/AtomString.h>
 #include <wtf/TZoneMallocInlines.h>
 
 #if ENABLE(TREE_DEBUGGING)
@@ -58,6 +60,16 @@ static CounterMaps& counterMaps()
 {
     static NeverDestroyed<CounterMaps> staticCounterMaps;
     return staticCounterMaps;
+}
+
+static RenderElement* ancestorStyleContainmentObject(const RenderElement& element)
+{
+    SUPPRESS_UNCOUNTED_LOCAL auto* ancestor = element.parent();
+    for (; ancestor; ancestor = ancestor->parent()) {
+        if (ancestor->shouldApplyStyleContainment())
+            break;
+    }
+    return ancestor;
 }
 
 // Descend into a subtree to find its last RenderElement in pre-order, stopping at style containment boundaries.
@@ -175,12 +187,7 @@ static CounterDirectives listItemCounterDirectives(RenderElement& renderer)
     return { };
 }
 
-struct CounterPlan {
-    OptionSet<CounterNode::Type> type;
-    int value;
-};
-
-static std::optional<CounterPlan> planCounter(RenderElement& renderer, const AtomString& identifier)
+static std::optional<CounterNode::Plan> planCounter(RenderElement& renderer, const AtomString& identifier)
 {
     // We must have a generating node or else we cannot have a counter.
     RefPtr generatingElement = renderer.generatingElement();
@@ -217,11 +224,11 @@ static std::optional<CounterPlan> planCounter(RenderElement& renderer, const Ato
         type.add(CounterNode::Type::Increment);
 
     if (directives.setValue)
-        return CounterPlan { type, *directives.setValue };
+        return CounterNode::Plan { type, *directives.setValue };
     if (directives.resetValue)
-        return CounterPlan { type, saturatedSum<int>(*directives.resetValue, directives.incrementValue.value_or(0)) };
+        return CounterNode::Plan { type, saturatedSum<int>(*directives.resetValue, directives.incrementValue.value_or(0)) };
     if (directives.incrementValue)
-        return CounterPlan { type, *directives.incrementValue };
+        return CounterNode::Plan { type, *directives.incrementValue };
     return std::nullopt;
 }
 
@@ -374,7 +381,7 @@ static CounterNode* makeCounterNode(RenderElement& renderer, const AtomString& i
     auto& maps = counterMaps();
 
     auto type = plan ? plan->type : OptionSet<CounterNode::Type> { };
-    auto newNode = CounterNode::create(renderer, type, plan ? plan->value : 0);
+    auto newNode = CounterNode::create(renderer, plan ? plan->type : OptionSet<CounterNode::Type> { }, plan ? plan->value : 0, plan);
 
     auto place = findPlaceForCounter(renderer, identifier, type);
     if (place.parent)
@@ -439,22 +446,42 @@ String RenderCounter::originalText() const
     if (!m_counterNode)
         return emptyString();
 
-    RefPtr counterNode = m_counterNode.get();
-    int value = counterNode->actsAsReset() ? counterNode->value() : counterNode->countInParent();
+    String text;
+    for (RefPtr current = m_counterNode; current; ) {
+        RefPtr child = current;
+        int value = child->actsAsReset() ? child->value() : child->countInParent();
 
-    auto counterText = [&](int value) {
-        return counterStyle()->text(value, writingMode());
-    };
-    auto text = counterText(value);
-    if (!m_counter.separator.isNull()) {
-        if (!counterNode->actsAsReset())
-            counterNode = counterNode->parent();
-        while (RefPtr parent = counterNode->parent()) {
-            text = makeString(counterText(counterNode->countInParent()), m_counter.separator, text);
-            counterNode = parent;
+        auto counterText = [&](int value) {
+            return counterStyle()->text(value, writingMode());
+        };
+        auto newText = counterText(value);
+        if (!m_counter.separator.isNull()) {
+            if (!child->actsAsReset())
+                child = child->parent();
+            while (RefPtr parent = child->parent()) {
+                newText = makeString(counterText(child->countInParent()), m_counter.separator, newText);
+                child = parent;
+            }
         }
-    }
 
+        if (newText != "0"_s || current->shouldShowZero()) {
+            if (!text.isEmpty())
+                text = makeString(newText, m_counter.separator, text);
+            else
+                text = newText;
+        }
+
+        if (m_counter.separator.isNull())
+            return text.isEmpty() ? newText : text;
+
+        if (auto ancestorContainment = ancestorStyleContainmentObject(current->owner())) {
+            if (auto renderer = ancestorContainment) {
+                current = makeCounterNode(*renderer, m_counter.identifier, true);
+                continue;
+            }
+        }
+        return text.isEmpty() ? newText : text;
+    }
     return text;
 }
 
--- a/LayoutTests/TestExpectations
+++ b/LayoutTests/TestExpectations
@@ -5874,8 +5874,6 @@ imported/w3c/web-platform-tests/css/css-contain/content-visibility/content-visib
 imported/w3c/web-platform-tests/css/css-contain/content-visibility/content-visibility-hidden-scrollTop-left-width-height.html [ Failure ]
 
 # Style containment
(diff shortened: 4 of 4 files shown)


Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -400,7 +401,7 @@ static CounterNode* makeCounterNode(RenderElement& renderer, const AtomString& i
 
     auto& maps = counterMaps();
 
-    auto newNode = CounterNode::create(renderer, plan ? plan->type : OptionSet<CounterNode::Type> { }, plan ? plan->value : 0);
+    auto newNode = CounterNode::create(renderer, plan ? plan->type : OptionSet<CounterNode::Type> { }, plan ? plan->value : 0, !!plan);
```
brentfulgham: It really seems like the `CounterNode::create(...)` method should take a `std::optional<CounterPlan>` and use it to populate the `CounterNode` constructor.

Source/WebCore/rendering/CounterNode.cpp
```diff
@@ -339,6 +340,15 @@ void CounterNode::removeChild(CounterNode& oldChild)
         next->recount();
 }
 
+bool CounterNode::shouldShowZero()
+{
+    if (m_explicitValue)
+        return true;
+    if (this->parent())
+        return this->parent()->shouldShowZero();
```
brentfulgham: Would prefer something like:
```
if (auto parent = this->parent())
    return parent->shouldShowZero();
```

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -465,7 +466,14 @@ String RenderCounter::originalText() const
...
         return emptyString();
 
-    RefPtr child = m_counterNode.get();
+    auto text = originalTextImpl(*m_counterNode);
+    return text;
+}
+
+String RenderCounter::originalTextImpl(CounterNode& currentNode) const
+{
+
```
brentfulgham, line 466: Nit: Please remove this extra newline.

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -479,6 +487,21 @@ String RenderCounter::originalText() const
             text = makeString(counterText(child->countInParent()), m_counter.separator, text);
             child = parent;
         }
+        auto ancestorContainment = ancestorStyleContainmentObject(*currentNode.owner().element());
+        if (ancestorContainment) {
```
brentfulgham: Prefer:
```
if (auto ancestorContainment = ancestorStyleContainmentObject(*currentNode.owner().element())) {
```

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -479,6 +487,21 @@ String RenderCounter::originalText() const
...
+            auto renderer = ancestorContainment->renderer();
+            if (renderer) {
```
brentfulgham: Prefer:
```
if (auto renderer = ancestorContainment->renderer()) {
```

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -479,6 +487,21 @@ String RenderCounter::originalText() const
...
+                auto counterNode = makeCounterNode(*renderer, m_counter.identifier, true);
+                if (counterNode) {
```
brentfulgham: Prefer:
```
if (auto counterNode = makeCounterNode(*renderer, m_counter.identifier, true)) {
```

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -465,7 +466,14 @@ String RenderCounter::originalText() const
     if (!m_counterNode)
         return emptyString();
 
-    RefPtr child = m_counterNode.get();
+    auto text = originalTextImpl(*m_counterNode);
+    return text;
```
brentfulgham: Seems like this should just be:

```
return originalTextImpl(*m_counterNode);
```

Source/WebCore/rendering/CounterNode.h
```diff
@@ -45,11 +45,12 @@ class CounterNode : public RefCounted<CounterNode>, public CanMakeSingleThreadWe
 public:
     enum class Type : uint8_t { Increment, Reset, Set };
 
-    static Ref<CounterNode> create(RenderElement&, OptionSet<Type>, int value);
+    static Ref<CounterNode> create(RenderElement&, OptionSet<Type>, int value, bool explicitValue);
     ~CounterNode();
     bool actsAsReset() const { return hasResetType() || !m_parent; }
     bool hasResetType() const { return m_type.contains(Type::Reset); }
     bool hasSetType() const { return m_type.contains(Type::Set); }
+    bool shouldShowZero();
```
darinadler:
```suggestion
    bool shouldShowZero() const;
```

Source/WebCore/rendering/RenderCounter.h
```diff
@@ -50,6 +50,7 @@ class RenderCounter final : public RenderText {
     
     ASCIILiteral renderName() const override;
     String originalText() const override;
+    String originalTextImpl(CounterNode&) const;
```
darinadler: Much prefer a name made out of words. I understand that you think of this as "the implementation of `originalText`" but I think we could name it `originalTextForNode` or `computeOriginalText`.

Also, it seems like this function calls itself. Is there some guarantee such calls won't recurse or could this use stack proportional to the size of a document in some pathological case?

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -479,6 +487,21 @@ String RenderCounter::originalText() const
...
             child = parent;
         }
+        auto ancestorContainment = ancestorStyleContainmentObject(*currentNode.owner().element());
+        if (ancestorContainment) {
+            auto renderer = ancestorContainment->renderer();
+            if (renderer) {
+                auto counterNode = makeCounterNode(*renderer, m_counter.identifier, true);
+                if (counterNode) {
+                    String containmentText = originalTextImpl(*counterNode);
+                    if (!containmentText.isEmpty()) {
```
darinadler:
```suggestion
                    if (String containmentText = originalTextImpl(*counterNode); !containmentText.isEmpty()) {
```

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -479,6 +487,21 @@ String RenderCounter::originalText() const
             text = makeString(counterText(child->countInParent()), m_counter.separator, text);
             child = parent;
         }
+        auto ancestorContainment = ancestorStyleContainmentObject(*currentNode.owner().element());
+        if (ancestorContainment) {
+            auto renderer = ancestorContainment->renderer();
+            if (renderer) {
+                auto counterNode = makeCounterNode(*renderer, m_counter.identifier, true);
```
darinadler: We usually avoid `bool` arguments to avoid mysterious things at call sites like this `true` here.

darinadler (approved):

Source/WebCore/rendering/CounterNode.h
```diff
@@ -40,16 +41,22 @@ namespace WebCore {
 
 class RenderCounter;
 class RenderElement;
+struct CounterPlan;
```
darinadler: This is probably left over from an earlier version and now wrong. We don't need both `WebCore::CounterPlan` and `WebCore::CounterNode::CounterPlan`.

Source/WebCore/rendering/CounterNode.h
```diff
@@ -40,16 +41,22 @@ namespace WebCore {
...
 
 class CounterNode : public RefCounted<CounterNode>, public CanMakeSingleThreadWeakPtr<CounterNode> {
 public:
     enum class Type : uint8_t { Increment, Reset, Set };
+    struct CounterPlan {
```
darinadler: Could consider just calling this `Plan` since it's a member of `CounterNode`.

Source/WebCore/rendering/CounterNode.h
```diff
@@ -40,16 +41,22 @@ namespace WebCore {
...
+        OptionSet<CounterNode::Type> type;
```
darinadler:
```suggestion
        OptionSet<Type> type;
```

Source/WebCore/rendering/CounterNode.h
```diff
@@ -40,16 +41,22 @@ namespace WebCore {
...
+        int value;
```
darinadler: Consider a default value here?
```suggestion
        int value { 0 };
```
Perhaps cleaner to not support uninitialized values.

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -465,22 +461,49 @@ String RenderCounter::originalText() const
...
-    auto text = counterText(value);
-    if (!m_counter.separator.isNull()) {
-        if (!child->actsAsReset())
-            child = child->parent();
-        while (RefPtr parent = child->parent()) {
-            text = makeString(counterText(child->countInParent()), m_counter.separator, text);
-            child = parent;
+    auto& currentNode = *m_counterNode;
+    String text;
+    for (RefPtr<CounterNode> current = &currentNode; current; ) {
```
darinadler: Can we write it like this?
```suggestion
    for (RefPtr current = m_currentNode; current; ) {
```
What is the local `currentNode` for?

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -465,22 +461,49 @@ String RenderCounter::originalText() const
...
+                text = makeString(newText, m_counter.separator, text);
+            else
+                text = newText;
+        }
+
+        if (m_counter.separator.isNull()) {
+            if (text.isEmpty())
+                text = newText;
+            break;
         }
```
darinadler: Maybe write it like this?
```suggestion
        if (m_counter.separator.isNull()
            return text.isEmpty() ? newText : text;
```

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -465,22 +461,49 @@ String RenderCounter::originalText() const
...
+        auto ancestorContainment = ancestorStyleContainmentObject(*current->owner().element());
+        if (ancestorContainment) {
+            if (auto renderer = ancestorContainment->renderer()) {
+                current = makeCounterNode(*renderer, m_counter.identifier, true);
+                continue;
+            }
+        }
+        if (text.isEmpty())
+            text = newText;
+        break;
```
darinadler: Maybe write this?
```suggestion
        return text.isEmpty() ? newText : text;
```

Source/WebCore/rendering/CounterNode.cpp
```diff
@@ -27,15 +27,16 @@
 #include "RenderElement.h"
 #include "RenderObjectDocument.h"
 #include "RenderView.h"
+#include <optional>
```
darinadler: I think this needs to be included in the header, not the `.cpp` file.

Source/WebCore/rendering/CounterNode.cpp
```diff
@@ -27,15 +27,16 @@
...
 #include <stdio.h>
 #include <wtf/CheckedArithmetic.h>
 
 namespace WebCore {
-
```
darinadler, line 34: Should not remove this blank line.

Source/WebCore/rendering/CounterNode.cpp
```diff
@@ -339,6 +340,15 @@ void CounterNode::removeChild(CounterNode& oldChild)
         next->recount();
 }
 
+bool CounterNode::shouldShowZero() const
+{
+    if (m_plan.has_value())
+        return true;
+    if (auto parent = this->parent())
+        return parent->shouldShowZero();
+    return false;
```
darinadler: I suggest we write this as a loop instead of recursion.
```suggestion
    for (auto* node = this; node; node = node->parent()) {
        if (node->m_plan.has_value())
            return true;
    }
    return false;
```

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -39,6 +39,7 @@
 #include "RenderStyle.h"
 #include "RenderView.h"
 #include <wtf/StdLibExtras.h>
+#include <wtf/text/WTFString.h>
```
darinadler: Given this file uses `AtomString`, I would think that's the header we would need to include.

darinadler (approved):

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -60,6 +61,23 @@ static CounterMaps& counterMaps()
     return staticCounterMaps;
 }
 
+static Element* ancestorStyleContainmentObject(const Element& element)
```
islandryu: This function was removed in this PR, but I have restored it.
https://github.com/WebKit/WebKit/pull/59629

islandryu (commented):
Thank you for the review.
I have made the changes based on comments.

Source/WebCore/rendering/RenderCounter.cpp
islandryu: Should we create a new enum?

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -60,6 +61,23 @@ static CounterMaps& counterMaps()
...
+{
+    auto* pseudoElement = dynamicDowncast<PseudoElement>(element);
+    SUPPRESS_UNCOUNTED_LOCAL auto* ancestor = pseudoElement ? pseudoElement->hostElement() : element.parentElement();
+    while (ancestor) {
```
darinadler: Fine as is but would be slightly better as a `for` loop.
```suggestion
    // FIXME: This should use parentInComposedTree but for now matches the rest of RenderCounter.
    for (; ancestor; ancestor = ancestor->parentElement()) {
```

darinadler (approved):

Source/WebCore/rendering/RenderCounter.cpp
nt1m: Please don’t use element based look ups and use renderer look ups instead. The code written here will be inaccurate for pseudo-elements like ::checkmark/::picker-icon.

nt1m (changes requested):

Source/WebCore/rendering/RenderCounter.cpp
islandryu: Thanks for the feedback. 
I changed the implementation to perform the lookup using RenderElement.

Source/WebCore/rendering/RenderCounter.cpp
```diff
@@ -60,6 +62,17 @@ static CounterMaps& counterMaps()
     return staticCounterMaps;
 }
 
+static RenderElement* ancestorStyleContainmentObject(const RenderElement& element)
+{
+    SUPPRESS_UNCOUNTED_LOCAL auto* ancestor = element.parent();
+    // FIXME: This should use parentInComposedTree but for now matches the rest of RenderCounter.
```
nt1m: The fixme is no longer relevant with renderer based look ups.

islandryu: Removed

islandryu: @nt1m 
I'd appreciate it if you could take a look when you have time.

PR #59266, opened 2026-02-23
Introduce LocalFrameViewDestructionObserver.

sammygill:
#### 6bbd56c5ecd76938fe9cb0d2b52c0b7afa72e416
<pre>
Introduce LocalFrameViewDestructionObserver.
<a href="https://bugs.webkit.org/show_bug.cgi?id=308484">https://bugs.webkit.org/show_bug.cgi?id=308484</a>
<a href="https://rdar.apple.com/171008275">rdar://171008275</a>

Reviewed by NOBODY (OOPS!).

This patch introduces LocalFrameViewDestructionObserver, which is a
simple helper class that is used to be notified when a particular
LocalFrameView is being destroyed. This class and API follow a very
similar pattern to FrameDestructionObserver, which is also used in a few
different places throughout the codebase. We apply this new class by
starting to use it inside of RenderView to observe the LocalFrameView
that is associated with it.

Use of this new class is fairly straightforward since it has a very
small API.

(WebCore::LocalFrameView::~LocalFrameView):
We want to make sure that we notify any observers that are currently
observing the LocalFrameView that it is being destroyed so that they can
respond properly.

* Source/WebCore/page/LocalFrameViewDestructionObserver.cpp: Added.
(WebCore::LocalFrameViewDestructionObserver::LocalFrameViewDestructionObserver):
In order to begin observing a LocalFrameView, clients will create an observer by
passing in the associated LocalFrameView to the constructor. This new
observer will notify the LocalFrameView of its creation so that it can
begin keeping track of it. We do this by storing them in a SingleThreadWeakHashSet
since the LocalFrameView is not responsible for any of its observers and
should not care if any of them disappear.

(WebCore::LocalFrameViewDestructionObserver::~LocalFrameViewDestructionObserver):
When the client decides it no longer needs the observer, we will
unobserve the associated LocalFrameView so that it can perform its
bookkeeping properly. Destroying the observer is one way to accomplish
this but there is also an unobserve API that could be used.

(WebCore::LocalFrameViewDestructionObserver::frameViewIsBeingDestroyed):
This pure virtual function is the primary way that observers will get
notified that their associated LocalFrameView is being destroyed. They
will need to implement this since the context will likely play a large
part in how they want to respond to this event. The LocalFrameView will
go through any of the observers that are still alive and call this
method inside of its destructor.

(WebCore::LocalFrameViewDestructionObserver::observeFrameView):
(WebCore::LocalFrameViewDestructionObserver::unobserveFrameView):
Clients can arbitrarily observe and unobserve LocalFrameViews through the
dedicated public methods which inform the LocalFrameView properly.

* Source/WebCore/rendering/RenderView.cpp:
(WebCore::RenderView::FrameViewDestructionObserver::FrameViewDestructionObserver):
In RenderView we create a new nested class: FrameViewDestructionObserver
which inherits from LocalFrameViewDestructionObserver since RenderView
itself cannot inherit directly from it. That would result in two
different ways it would inherit from CanMakeSingleThreadWeakPtr.
</pre>
https://github.com/WebKit/WebKit/commit/6bbd56c5ecd76938fe9cb0d2b52c0b7afa72e416

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/146467 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/19140 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/12649 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/155131 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/99878 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/148342 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/19602 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/19044 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/112755 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/80598 "Passed tests") | [❌ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/a50ff1b4-a30b-4cd6-9443-4263a68bb568) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/149430 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/15073 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/131602 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/93588 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/cace7b35-b16e-4bbe-8267-cd3ce5e8c608) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/14336 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/12101 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/2576 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/123906 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/9467 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/157455 "Built successfully") | | 
| | [❌ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/611 "Found 33 new failures in rendering/RenderSearchField.cpp, rendering/svg/SVGPaintServerHandlingInlines.h, rendering/BackgroundPainter.cpp, page/LocalFrameViewDestructionObserver.cpp, inspector/agents/page/PageTimelineAgent.cpp, rendering/RenderLayerModelObject.cpp, rendering/RenderReplaced.cpp, rendering/RenderMultiColumnFlow.cpp, rendering/RenderBlockFlow.cpp, rendering/svg/legacy/LegacyRenderSVGResourceSolidColor.cpp ...") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/10898 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/120804 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/18946 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/15899 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/121052 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/18961 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/131219 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/74750 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/16706 "Passed tests") | [❌ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/8130 "Found 1 new test failure: scrollingcoordinator/mac/latching/horizontal-overflow-back-swipe.html (failure)") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/18568 "Built successfully") | [❌ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/82318 "Failed to build and analyze WebKit") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/18297 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/18453 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/18354 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/page/LocalFrameViewDestructionObserver.cpp
+++ b/Source/WebCore/page/LocalFrameViewDestructionObserver.cpp
@@ -1,0 +1,58 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL APPLE INC. OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ * OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "LocalFrameViewDestructionObserver.h"
+
+#include "LocalFrameView.h"
+
+namespace WebCore {
+
+LocalFrameViewDestructionObserver::LocalFrameViewDestructionObserver(LocalFrameView& frameView)
+{
+    observeFrameView(frameView);
+}
+
+LocalFrameViewDestructionObserver::~LocalFrameViewDestructionObserver()
+{
+    unobserveFrameView();
+}
+
+void LocalFrameViewDestructionObserver::observeFrameView(LocalFrameView& frameView)
+{
+    if (m_frameView)
+        unobserveFrameView();
+    m_frameView = &frameView;
+    m_frameView->addDestructionObserver(*this);
+}
+
+void LocalFrameViewDestructionObserver::unobserveFrameView()
+{
+    ASSERT(m_frameView);
+    if (m_frameView)
+        m_frameView->removeDestructionObserver(*this);
+}
+
+} // namespace WebCore
--- a/Source/WebCore/rendering/RenderView.cpp
+++ b/Source/WebCore/rendering/RenderView.cpp
@@ -43,6 +43,7 @@
 #include "LegacyRenderSVGRoot.h"
 #include "LocalFrame.h"
 #include "LocalFrameView.h"
+#include "LocalFrameViewDestructionObserver.h"
 #include "Logging.h"
 #include "NodeInlines.h"
 #include "NodeTraversal.h"
@@ -81,9 +82,19 @@ namespace WebCore {
 
 WTF_MAKE_TZONE_ALLOCATED_IMPL(RenderView);
 
+RenderView::FrameViewDestructionObserver::FrameViewDestructionObserver(LocalFrameView& frameView)
+    : LocalFrameViewDestructionObserver(frameView)
+{
+}
+
+void RenderView::FrameViewDestructionObserver::frameViewIsBeingDestroyed()
+{
+    WTFReportBacktraceWithPrefix("RenderView::LocalFrameViewDestructionObserver::frameViewIsBeingDestroyed()");
+}
+
 RenderView::RenderView(Document& document, RenderStyle&& style)
     : RenderBlockFlow(Type::View, document, WTF::move(style))
-    , m_frameView(*document.view())
+    , m_frameViewDestructionObserver(makeUniqueRef<FrameViewDestructionObserver>(*document.view()))
     , m_initialContainingBlock(makeUniqueRef<Layout::InitialContainingBlock>(RenderStyle::clone(this->style())))
     , m_layoutState(makeUniqueRef<Layout::LayoutState>(document, m_initialContainingBlock, Layout::LayoutState::Type::Primary, LayoutIntegration::layoutWithFormattingContextForBox, LayoutIntegration::formattingContextRootLogicalWidthForType, LayoutIntegration::formattingContextRootLogicalHeightForType, LayoutIntegration::layoutWithFormattingContextForBlockInInline))
     , m_selection(*this)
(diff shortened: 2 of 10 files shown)


Source/WebCore/page/LocalFrameViewDestructionObserver.cpp
```diff
@@ -23,14 +23,37 @@
...
+#include "LocalFrameView.h"
 
-namespace WGSL {
+namespace WebCore {
 
-class ShaderModule;
+LocalFrameViewDestructionObserver::LocalFrameViewDestructionObserver(LocalFrameView& frameView)
+    : m_frameView(frameView)
+{
+    observeFrameView(frameView);
```
brentfulgham, line 35: Doesn't this cause us to `unobserverFrameView()` then set the frame view, since we initialized `m_frameView` above? Seems like we should just initialize `m_frameView` to `nullptr` or perhaps not initialize it at all in the initializer list, since we will be using the helper function for that purpose.

Source/WebCore/rendering/RenderView.h
```diff
@@ -250,6 +258,7 @@ class RenderView final : public RenderBlockFlow {
     void updateInitialContainingBlockSize();
 
     const CheckedRef<LocalFrameView> m_frameView;
```
brentfulgham: If we are following the pattern of `FrameDestructionObserver` we wouldn't hold `m_frameView` separately, since `m_frameViewDestructionObserver` owns the frame reference.

This might mean you need to expose an accessor in `m_frameViewDestructionObserver` so that you can ask the frame questions.

Source/WebCore/page/LocalFrameViewDestructionObserver.cpp
sammygill: Yeah we should just use the helper function for this so I removed the initializer

Source/WebCore/rendering/RenderView.h
sammygill: Good point! Made some modifications so now `RenderView::frameView` goes through the observer since it does not hold onto it anymore

brentfulgham (approved):
Looks good! r=me

Source/WebCore/rendering/RenderView.cpp
```diff
@@ -81,9 +82,19 @@ namespace WebCore {
...
 WTF_MAKE_TZONE_ALLOCATED_IMPL(RenderView);
 
+RenderView::FrameViewDestructionObserver::FrameViewDestructionObserver(LocalFrameView& frameView)
+    : LocalFrameViewDestructionObserver(frameView)
+{
+}
+
+void RenderView::FrameViewDestructionObserver::frameViewIsBeingDestroyed()
+{
+    WTFReportBacktraceWithPrefix("RenderView::LocalFrameViewDestructionObserver::frameViewIsBeingDestroyed()");
```
rniwa, line 92: I don't think we want to do this here.

PR #59419, opened 2026-02-25
Web Inspector: Network: Follow-up for Bug 308310 Request / Response menu ignores user’s previous setting

rcaliman-apple:
#### db14f805f786480be3828341eefdc7823cea91de
<pre>
Web Inspector: Network: Follow-up for Bug 308310 Request / Response menu ignores user’s previous setting
<a href="https://bugs.webkit.org/show_bug.cgi?id=308642">https://bugs.webkit.org/show_bug.cgi?id=308642</a>
<a href="https://rdar.apple.com/171168680">rdar://171168680</a>

Reviewed by Devin Rousso.

Replace the singular &quot;resource-content-view-identifier-for-mime-type&quot; setting
with individual settings per mime type.

Remove `WI.ResourceClusterContentView.prototype.attached()` with its logic to show
a prefered or default content view because `WI.ResourceClusterContentView.prototype.restoreFromCookie()`,
which already handles this, is always called.

This view is always shown via the call chain `WI.ContentViewContainer.showBackForwardEntryForIndex()` -&gt; `_showEntry()`
which creates the view, attaches it, then calls `restoreFromCookie()` on it.

(WI.Setting.prototype.save):

Drive-by: check for equality between primitives (string, number, boolean),
not just shallow equal objects, when determining if the new value and default value
are identical in order to remove the entry from local storage.

* Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js:
(WI.ResourceClusterContentView.prototype._showContentViewForIdentifier):
(WI.ResourceClusterContentView.prototype._pathComponentSelected):
(WI.ResourceClusterContentView.prototype._getPreferredContentViewIdentifier):
(WI.ResourceClusterContentView.prototype._getPreferredContentViewSetting):
(WI.ResourceClusterContentView.prototype._setPreferredContentViewIdentifier):
(WI.ResourceClusterContentView.prototype.async _tryEnableCustomResponseContentViews):
(WI.ResourceClusterContentView):
(WI.ResourceClusterContentView.prototype.attached): Deleted.
(WI.ResourceClusterContentView.prototype.closed): Deleted.
(WI.ResourceClusterContentView.prototype._tryEnableCustomResponseContentViews): Deleted.

Canonical link: <a href="https://commits.webkit.org/308344@main">https://commits.webkit.org/308344@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/927b5af78366a07133f3a6b7468bf63c5908d91e

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/146997 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/19677 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/13267 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/155679 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/100411 "Built successfully") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/20136 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/19578 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/113271 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/80830 "Passed tests") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/bf132114-f9f2-4319-866e-e40bda1c0187) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/149959 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/15521 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/132070 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94027 "Passed tests") | | [loading 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/a8004150-7d72-49c5-b84b-3ecaecf0a817) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/14728 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/12505 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/3121 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/124315 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/9988 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/158010 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/1141 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/11427 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/121294 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/19479 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/16354 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/121495 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/31165 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/19488 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/131744 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/75447 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/17063 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/8570 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/19094 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/82849 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/18824 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/18975 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/18883 "Built successfully") | | | |

### DIFF
--- a/Source/WebInspectorUI/UserInterface/Base/Setting.js
+++ b/Source/WebInspectorUI/UserInterface/Base/Setting.js
@@ -118,7 +118,7 @@ WI.Setting = class Setting extends WI.Object
         if (!window.InspectorTest && window.localStorage) {
             let key = WI.Setting._localStorageKeyPrefix + this._name;
             try {
-                if (Object.shallowEqual(this._value, this._defaultValue))
+                if (this._value === this._defaultValue || Object.shallowEqual(this._value, this._defaultValue))
                     window.localStorage.removeItem(key);
                 else
                     window.localStorage.setItem(key, JSON.stringify(this._value));
--- a/Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
+++ b/Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
@@ -72,30 +72,6 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
         return components.concat(currentContentView.selectionPathComponents);
     }
 
-    attached()
-    {
-        super.attached();
-
-        if (this._shownInitialContent)
-            return;
-
-        this._enableCustomResponseContentViewsPromise
-        .then(() => {
-            let identifier = this._getPreferredContentViewIdentifier();
-            this._showContentViewForIdentifier(identifier);
-        })
-        .catch((error) => {
-            WI.reportInternalError(error);
-        });
-    }
-
-    closed()
-    {
-        super.closed();
-
-        this._shownInitialContent = false;
-    }
-
     restoreFromCookie(cookie)
     {
         let contentViewIdentifier = cookie[WI.ResourceClusterContentView.ContentViewIdentifierCookieKey] || this._getPreferredContentViewIdentifier();
@@ -338,7 +314,7 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
         return null;
     }
 
-    _showContentViewForIdentifier(identifier, saveAsPreference)
+    _showContentViewForIdentifier(identifier, {saveAsPreference} = {})
     {
         let contentViewToShow = null;
 
@@ -385,15 +361,12 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
         if (saveAsPreference)
             this._setPreferredContentViewIdentifier(identifier);
 
-        this._shownInitialContent = true;
-
         return this.contentViewContainer.showContentView(contentViewToShow);
     }
 
     _pathComponentSelected(event)
     {
-        const saveAsPreference = true;
-        this._showContentViewForIdentifier(event.data.pathComponent.representedObject, saveAsPreference);
+        this._showContentViewForIdentifier(event.data.pathComponent.representedObject, {saveAsPreference: true});
     }
 
     _resourceTypeDidChange(event)
@@ -482,8 +455,15 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
         if (!mimeType)
             return null;
 
-        let preferences = WI.settings.resourceContentViewIdentifierForMIMEType.value;
-        return preferences[mimeType] || null;
+        return this._getPreferredContentViewSetting(mimeType).value;
+    }
+
+    _getPreferredContentViewSetting(mimeType)
+    {
+        WI.ResourceClusterContentView._preferredContentViewIdentifierSettingForMIMEType ??= new Map;
+        let setting = WI.ResourceClusterContentView._preferredContentViewIdentifierSettingForMIMEType.getOrInsertComputed(mimeType, () => new WI.Setting("resource-preferred-content-view-" + mimeType, ResourceClusterContentView.Identifier.Response));
+
+        return setting;
     }
 
     _setPreferredContentViewIdentifier(identifier)
@@ -492,15 +472,13 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
         if (!mimeType)
             return;
 
-        let preferences = WI.settings.resourceContentViewIdentifierForMIMEType.value;
+        let setting = this._getPreferredContentViewSetting(mimeType);
 
-        if (identifier === ResourceClusterContentView.Identifier.Response)
-            delete preferences[mimeType];
-        else
-            preferences[mimeType] = identifier;
+        if (identifier === setting.defaultValue)
+            WI.ResourceClusterContentView._preferredContentViewIdentifierSettingForMIMEType.delete(mimeType);
 
-        // `preferences` is a referenced object. Changing its contents and simply setting it as a new value won't qualify as value change so it won't be persisted.
-        WI.settings.resourceContentViewIdentifierForMIMEType.value = structuredClone(preferences);
+        // Resetting to the default value automatically removes the setting from persistent storage.
+        setting.value = identifier;
     }
 
     _tryEnableCustomRequestContentViews()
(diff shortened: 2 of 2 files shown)


Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -72,30 +72,6 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
         return components.concat(currentContentView.selectionPathComponents);
     }
 
-    attached()
```
dcrousso, line 75: what about the workflow of when switching between tabs?  does that go through this instead of `restoreFromCookie`? im fine removing this if it's not needed, but it does seem somewhat surprising that we don't even need `attached` anymore

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -492,15 +472,12 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
         if (!mimeType)
             return;
 
-        let preferences = WI.settings.resourceContentViewIdentifierForMIMEType.value;
+        let setting = this._getPreferredContentViewSetting(mimeType);
 
         if (identifier === ResourceClusterContentView.Identifier.Response)
-            delete preferences[mimeType];
+            WI.Setting.migrateValue(setting.name);
```
dcrousso: this is not the intended use of `WI.Setting.migrateValue`

`WI.Setting` will already delete itself from `localStorage` if the value is equal to the default

if you want this to be automatic then id either suggest having this be `setting.value = null;` or change the default value to be `WI.ResourceClusterContentView.Identifier.Response` in `_getPreferredContentViewSetting`

dcrousso (commented):
thanks for the followup!  just a few questions and corrections

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: All entry points I can find that show the `WI.ResourceClusterContentView` end up calling `restoreFromCookie` on it. The entry point is aways via `ContentViewContainer.showContentView()`

Showing the Sources tab for the first time, the `NavigationSidebarPanel.attach()` initiates it:

Screenshot 2026-02-26 at 12 40 28

Returning to the Sources tab after it was already opened:
Screenshot 2026-02-26 at 12 41 51

Switching to the Preview tab for a resource in the Network tab:
Screenshot 2026-02-26 at 12 43 35

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: > this is not the intended use of `WI.Setting.migrateValue`

Fair.

> `WI.Setting` will already delete itself from `localStorage` if the value is equal to the default

Neat! But there's a bug in the current implementation 😅:
`Object.isShallowEqual("response", "response") => false`

Fixed in the latest patch.

We might have a lot of default primitives unintentionally persisted in local storage 😄. 
 
> if you want this to be automatic then id either suggest having this be `setting.value = null;` or change the default value to be `WI.ResourceClusterContentView.Identifier.Response` in `_getPreferredContentViewSetting`

Using the `WI.ResourceClusterContentView.Identifier.Response` as the default and checking the identifier against `setting.defaultValue` makes it nice and explicit. Love it! 👍

Source/WebInspectorUI/UserInterface/Base/Setting.js
```diff
@@ -118,7 +118,7 @@ WI.Setting = class Setting extends WI.Object
         if (!window.InspectorTest && window.localStorage) {
             let key = WI.Setting._localStorageKeyPrefix + this._name;
             try {
-                if (Object.shallowEqual(this._value, this._defaultValue))
+                if (this._value === this._defaultValue || Object.shallowEqual(this._value, this._defaultValue))
```
dcrousso, line 121: i wonder if we should just be changing `Object.shallowEqual` instead
```diff
+if (a === b)
+    return true
+
 if (!(a instanceof Object) || !(b instanceof Object))
     return false;
 
-if (a === b)
-    return true;
-
```

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
```diff
@@ -492,15 +472,13 @@ WI.ResourceClusterContentView = class ResourceClusterContentView extends WI.Clus
...
-        if (identifier === ResourceClusterContentView.Identifier.Response)
-            delete preferences[mimeType];
-        else
-            preferences[mimeType] = identifier;
+        if (identifier === setting.defaultValue)
+            WI.ResourceClusterContentView._preferredContentViewIdentifierSettingForMIMEType.delete(mimeType);
 
-        // `preferences` is a referenced object. Changing its contents and simply setting it as a new value won't qualify as value change so it won't be persisted.
-        WI.settings.resourceContentViewIdentifierForMIMEType.value = structuredClone(preferences);
+        // Resetting to the default value automatically removes the setting from persistent storage.
```
dcrousso, line 480: this comment is unnecessary (and also reveals an internal detail of `WI.Setting` which is not really super relevant here)

dcrousso (approved):

Source/WebInspectorUI/UserInterface/Base/Setting.js
rcaliman-apple: I'd keep that unchanged to avoid any side effects that require a reject. It looks designed to operate on objects.

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: I disagree. We are relying on an internal detail of `WI.Setting` to actually clear the setting from storage. 
In absence of a `WI.Setting.clear()` or `.delete()`, I prefer to call out this special behavior.

dcrousso: why does `WI.ResourceClusterContentView` care about whether a setting is cleared from memory or not?

heck it doesn't even know the details about how `WI.Setting` works (i.e. how does it know whether it reads from `localStorage` or `indexedDB` or etc.?)

all `WI.ResourceClusterContentView` cares about is that the data is somehow preserved, which is all `WI.Setting` promises to do

whether and/or when that preserved data is deleted is not it's concern

Source/WebInspectorUI/UserInterface/Base/Setting.js
dcrousso: from a quick search it seems like all other current uses of `Object.shallowEqual` (and `Array.shallowEqual`) are using object (and array) values, so it probably would be fine to change it, but i agree that doing that separately would be better

Source/WebInspectorUI/UserInterface/Views/ResourceClusterContentView.js
rcaliman-apple: >why does WI.ResourceClusterContentView care about whether a setting is cleared from memory or not?

It doesn't care. Functionality is unaffected if that setting lingers or goes away.
It's just nice to clean up unused stuff from local storage.

I want to follow-up with a patch to clean-up all obsolete settings for:
`"resource-current-view-" + this._resource.url.hash`
https://github.com/WebKit/WebKit/pull/59096/changes#r2834307174

PR #59463, opened 2026-02-26
Web Inspector: Previewing resources in Web Inspector displays "An error occurred trying to load the resource" when Preserve Log is enabled

roynos:
#### 4ec1778cf0a7cd79ed4b2d08324cbd8fc49c72b9
<pre>
Web Inspector: Previewing resources in Web Inspector displays &quot;An error occurred trying to load the resource&quot; when Preserve Log is enabled
<a href="https://rdar.apple.com/171216835">rdar://171216835</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=308680">https://bugs.webkit.org/show_bug.cgi?id=308680</a>

Reviewed by Devin Rousso and BJ Burg.

Changed `Preseve Log` in Web Inspector&apos;s Network tab to say `Preserve Requests`. Plumbed state through InspectorNetworkAgent.

Test: inspector/network/clear-resource-data-on-navigate.html

* LayoutTests/inspector/network/clear-resource-data-on-navigate-expected.txt: Added.
* LayoutTests/inspector/network/clear-resource-data-on-navigate.html: Added.
* Source/JavaScriptCore/inspector/protocol/Network.json:
* Source/WebCore/inspector/agents/InspectorNetworkAgent.cpp:
(WebCore::InspectorNetworkAgent::setClearResourceDataOnNavigate):
(WebCore::InspectorNetworkAgent::mainFrameNavigated):
* Source/WebCore/inspector/agents/InspectorNetworkAgent.h:
* Source/WebInspectorUI/UserInterface/Base/Main.js:
* Source/WebInspectorUI/UserInterface/Views/NetworkTableContentView.js:
(WI.NetworkTableContentView.prototype._handleOtherFiltersNavigationItemContextMenu):

Canonical link: <a href="https://commits.webkit.org/308559@main">https://commits.webkit.org/308559@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/d1a8a9094270b7605fb568249fa5bbe8bb0957f4

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/147705 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/20390 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/13982 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/156388 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/101120 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/149578 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/20848 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/20290 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/113860 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/81207 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/150667 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/16099 "Passed tests") | [⏳ 🧪 api-mac](https://ews-build.webkit.org/#/builders/API-Tests-macOS-EWS "Waiting to run tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94620 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/15263 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/13049 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/3828 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/139675 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/124858 "Passed tests") | [⏳ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/macOS-Sequoia-Release-WK1-Tests-EWS "Waiting to run tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/158722 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/8493 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/1857 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/12048 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/121889 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/20189 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/16959 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/122090 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/31313 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/20200 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/132357 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/76315 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/17612 "Passed tests") | [⏳ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/macOS-Sequoia-Release-WK2-Intel-Tests-EWS "Waiting to run tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/179127 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/19805 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/83567 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/45906 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/19534 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/19685 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/19592 "Built successfully") | | | |


Source/JavaScriptCore/inspector/protocol/Network.json
```diff
@@ -186,7 +186,10 @@
     "commands": [
         {
             "name": "enable",
-            "description": "Enables network tracking, network events will now be delivered to the client."
+            "description": "Enables network tracking, network events will now be delivered to the client.",
+            "parameters": [
+                { "name": "preserveLog", "type": "boolean", "optional": true, "description": "Whether to preserve the network log across page navigations / reloads." }
```
burg: I don't think we need this?

Source/JavaScriptCore/inspector/protocol/Network.json
```diff
@@ -218,6 +221,14 @@
                 { "name": "disabled", "type": "boolean", "description": "Whether to prevent usage of the resource cache." }
             ]
         },
+        {
+            "name": "setPreserveLog",
+            "description": "Toggles whether the network log is preserved across page navigations / reloads.",
```
burg: Preserve Log means preserve *console* logs and not clear them when navigations happen.

burg: But this could be a generic 'setResourceStoragePolicy' command with 'clearOnMainFrameNavigation': true.

burg (changes requested):
It seems costly to effectively save all resources encountered by the Page, without cache eviction. Is that what it's supposed to do?

Please add tests to exercise the new command.

Source/JavaScriptCore/inspector/protocol/Network.json
dcrousso: yeah i think something more like this makes sense as "preserve log" is not quite the right terminology

i'd probably make it an `enum` with `ClearOnMainFrameNavigation` and `KeepForever` as options

we'd also wanna make it so that if it changes we invalidate everything not for the current page

and to avoid consuming a crazy amount of resources we should verify that the cache can evict the least recently used resource when needed

roynos: Yes. I am in favor or `clearOnMainFrameNavigation` .

Source/JavaScriptCore/inspector/protocol/Network.json
roynos: This is needed to synchronize the state at start up. This actually solves the case where:

1. User loads webpage A
2. User opens Web Inspector
3. User navigates to webpage B (without clicking on webpage A resources).

Resources from webpage A will get deleted in the backend

burg: Unless Web Inspector is open/the inspector backend has a frontend connection, it shouldn't be saving resources of webpage A at all. This is by design.
If webpage A has not InstrumentingAgents enabled, then NetworkResourcesData will not get filled up with webpage A's resources, because it only receives those callbacks (willSendRequest, didReceiveResponse, etc) during the actual loading events when the page loads.

Obviously that's little solace to a developer who knows that the resource is very likely still cached.

It would be better to backfill the NetworkResourcesData from CachedResources and their saved request/response objects once Web Inspector opens and calls Network.enable. Not quite sure how that would be done- perhaps replaying the callbacks like is already done for WebSockets. See the code below.

As a practical matter, you should probably tackle these as two parts (backfilling from cached resources on current page, and turning off cache clearing on main frame navigation) as separate enhancements. They seem orthogonal, and the latter is much simpler to implement.

```
Inspector::Protocol::ErrorStringOr<void> InspectorNetworkAgent::enable()
{
    m_enabled = true;
    Ref { m_instrumentingAgents.get() }->setEnabledNetworkAgent(this);

    {
        Locker locker { WebSocket::allActiveWebSocketsLock() };

        for (Ref webSocket : activeWebSockets()) {
            RefPtr document = dynamicDowncast<Document>(webSocket->scriptExecutionContext());
            if (!document)
                continue;

            RefPtr channel = webSocket->channel();

            auto identifier = channel->progressIdentifier();
            didCreateWebSocket(identifier, webSocket->url());

            auto cookieRequestHeaderFieldValue = [document = WeakPtr<Document, WeakPtrImplWithEventTargetData> { document }](const URL& url) -> String {
                if (!document || !document->page())
                    return { };
                return document->page()->cookieJar().cookieRequestHeaderFieldValue(*document, url);
            };
            willSendWebSocketHandshakeRequest(identifier, channel->clientHandshakeRequest(WTF::move(cookieRequestHeaderFieldValue)));

            if (channel->isConnected())
                didReceiveWebSocketHandshakeResponse(identifier, channel->serverHandshakeResponse());

            if (webSocket->readyState() == WebSocket::CLOSED)
                didCloseWebSocket(identifier);
        }
    }

    return { };
}
```

roynos: > If webpage A has not InstrumentingAgents enabled, then NetworkResourcesData will not get filled up with webpage A's resources, because it only receives those callbacks (willSendRequest, didReceiveResponse, etc) during the actual loading events when the page loads.

 I understand why it doesn't disappear if you click on it before navigating to webpage B. How come the requestBody appears in Preview when you click on webpage A's document?

Source/WebInspectorUI/UserInterface/Base/Main.js
```diff
@@ -2681,6 +2682,17 @@ WI._resourceCachingDisabledSettingChanged = function(event)
     }
 };
 
+WI._clearNetworkOnNavigateSettingChanged = function(event)
+{
+    for (let target of WI.targets) {
+        if (target instanceof WI.FrameTarget)
+            continue;
```
dcrousso: @burg @the-chenergy is this necessary?

Source/WebInspectorUI/UserInterface/Controllers/NetworkManager.js
```diff
@@ -210,9 +210,15 @@ WI.NetworkManager = class NetworkManager extends WI.Object
...
         if (target.hasDomain("Network")) {
-            target.NetworkAgent.enable();
+            if (target.hasCommand("Network.setClearNetworkOnNavigate"))
+                target.NetworkAgent.enable(WI.settings.clearNetworkOnNavigate.value);
+            else
+                target.NetworkAgent.enable();
             target.NetworkAgent.setResourceCachingDisabled(WI.settings.resourceCachingDisabled.value);
 
+            if (target.hasCommand("Network.setClearNetworkOnNavigate"))
+                target.NetworkAgent.setClearNetworkOnNavigate(WI.settings.clearNetworkOnNavigate.value);
```
dcrousso: the fact that you invoke this immediately after `Network.enable` means that there is only an extremely small amount of time after that where the backend has not gotten this yet (since IPC is guaranteed to be received in the order sent

additionally, Web Inspector will pause network activity whenever a new target is created until it's done setting up all instrumentation on whatnot on that target (see `WebPageInspectorController::shouldPauseLoading`)

as such, please remove the `clearNetworkOnNavigate` argument from `Network.enable`

Source/WebInspectorUI/UserInterface/Views/NetworkTableContentView.js
```diff
@@ -2724,7 +2724,7 @@ WI.NetworkTableContentView = class NetworkTableContentView extends WI.ContentVie
 
     _handleOtherFiltersNavigationItemContextMenu(contextMenu)
     {
-        contextMenu.appendCheckboxItem(WI.UIString("Preserve Log"), () => {
+        contextMenu.appendCheckboxItem(WI.UIString("Preserve Requests"), () => {
```
dcrousso: id suggest "Preserve Activity" or something like that

also, we should use the three argument form of `WI.UIString` so that localizers have more info
```js
contextMenu.appendCheckboxItem(WI.UIString("Preserve Activity", "Preserve Activity @ Network Tab", "Label for option that controls whether captured network activity is cleared when the page navigates."), () => {
```

roynos: Noted. I opted for "Preserve Requests" because Requests is what is used in the Network tab's UI.

burg:
This is looking much better. Two things left:
- remove enable argument per Devin
- prefer `clearResourceData` over `clearNetwork`

Source/WebCore/inspector/agents/InspectorNetworkAgent.cpp
```diff
@@ -841,7 +842,8 @@ Inspector::Protocol::ErrorStringOr<void> InspectorNetworkAgent::disable()
     m_interceptionEnabled = false;
     m_intercepts.clear();
     Ref { m_instrumentingAgents.get() }->setEnabledNetworkAgent(nullptr);
-    m_resourcesData->clear();
+    if (m_clearNetworkOnNavigate)
+        m_resourcesData->clear();
```
dcrousso: we should always be clearing the `NetworkResourcesData` whenever the frontend invokes `Network.disable` as that's used to disable all instrumentation of network activity (i.e. the frontend would only get notified of any network events thereafter if `Network.enable` is invoked again)

burg (approved):

dcrousso (approved):

Source/JavaScriptCore/inspector/protocol/Network.json
```diff
@@ -218,6 +218,14 @@
                 { "name": "disabled", "type": "boolean", "description": "Whether to prevent usage of the resource cache." }
             ]
         },
+        {
+            "name": "setClearResourceDataOnNavigate",
+            "description": "Toggles whether resource data is cleared on page navigations / reloads.",
+            "targetTypes": ["page"],
+            "parameters": [
+                { "name": "clearResourceDataOnNavigate", "type": "boolean", "description": "Whether to clear resource data on navigations." }
```
dcrousso, line 226: would be nice to mention that the default is `true` so that when looking back at older protocol files in the future we can know what it was expecting

PR #59717, opened 2026-03-02
Accept none and normal values for position-anchor CSS property

surajt6:
#### 91728e0e57bb8b1344d63e29f5f2c981d0d0dd0e
<pre>
Accept none and normal values for position-anchor CSS property

<a href="https://bugs.webkit.org/show_bug.cgi?id=308981">https://bugs.webkit.org/show_bug.cgi?id=308981</a>

Reviewed by Antti Koivisto.

<a href="https://drafts.csswg.org/css-anchor-position-1/#position-anchor">https://drafts.csswg.org/css-anchor-position-1/#position-anchor</a>
Add support for parsing, style computation and layout of none and normal
values for position-anchor CSS property. The new default value is normal.
position-anchor: normal and position-area: none is equivalent to position-anchor: none.

Update tests that expected position-anchor:auto by default.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/anchor-scroll-implicit-001.html:
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/anchor-scroll-implicit-002.html:
* LayoutTests/imported/w3c/web-platform-tests/html/dom/elements/global-attributes/the-anchor-attribute-003.tentative.html:

Update parsing and computed style tests
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-basics-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-basics.html:

Add new layout tests
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-001-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-001.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-002-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-002.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-none-pseudo-element-implicit-001-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-none-pseudo-element-implicit-001.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-none-pseudo-element-implicit-002-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-none-pseudo-element-implicit-002.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-none-pseudo-element-named-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-none-pseudo-element-named.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-normal-pseudo-element-implicit-001-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-normal-pseudo-element-implicit-001.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-normal-pseudo-element-implicit-002-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-normal-pseudo-element-implicit-002.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-normal-pseudo-element-implicit-003-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-normal-pseudo-element-implicit-003.html: Added.

Layout Changes
* Source/WebCore/style/AnchorPositionEvaluator.cpp:
(WebCore::Style::AnchorPositionEvaluator::findAnchorsForAnchorPositionedElement):
If the element has none as it&apos;s default anchor then do not add the implicit anchor
to the anchorToAnchorPositionedMap
(WebCore::Style::AnchorPositionEvaluator::updateAnchorPositioningStatesAfterInterleavedLayout):
(WebCore::Style::AnchorPositionEvaluator::isImplicitAnchor):

Parsing and style computation
* Source/WebCore/style/AnchorPositionEvaluator.h:
* Source/WebCore/style/values/anchor-position/StylePositionAnchor.cpp: Copied from Source/WebCore/style/values/anchor-position/StylePositionAnchor.h.
(WebCore::Style::CSSValueConversion&lt;PositionAnchor&gt;::operator):
* Source/WebCore/style/values/anchor-position/StylePositionAnchor.h:
(WebCore::Style::PositionAnchor::PositionAnchor):
(WebCore::Style::PositionAnchor::isNormal const):
(WebCore::Style::PositionAnchor::isNone const):
(WebCore::Style::PositionAnchor::isAuto const):
(WebCore::Style::PositionAnchor::isName const):
(WebCore::Style::PositionAnchor::tryName const):
(WebCore::Style::PositionAnchor::switchOn const):

Fix for failing build
* Source/WebCore/style/values/filter-effects/StyleAppleColorFilter.h:

Add new header and source file
* Source/WebCore/Sources.txt:

# Conflicts:
#       Source/WebCore/Sources.txt

Canonical link: <a href="https://commits.webkit.org/312378@main">https://commits.webkit.org/312378@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/c0e8a76a95a7a8340caf1299036793a0ac0b1f51

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/159732 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/33199 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/26306 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/168589 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/114113 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/161601 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/33304 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/33203 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/123767 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/114113 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/162690 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/26029 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/143471 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/104409 "Passed tests") | 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/25080 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/23553 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/16353 "Built successfully") | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/134772 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/21243 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/171082 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/17100 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/22880 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/132024 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/32878 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/27631 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/132080 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35745 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/32863 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/143037 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/90947 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/26703 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/19850 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/32372 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/98768 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/31869 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/32116 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/32020 "Built successfully") | | |

### DIFF
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-001.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-001.html
@@ -1,0 +1,23 @@
+<!doctype html>
+<title>position-anchor auto for ::after pseudo element</title>
+<link rel="help" href="https://drafts.csswg.org/css-anchor-position-1/#position-anchor"/>
+<link rel="author" title="Suraj Thanugundla" href="mailto:contact@surajt.com" />
+<link rel="match" href="../reference/ref-filled-green-100px-square-only.html">
+<style>
+    .anchor {
+        width: 100px;
+        height: 50px;
+        background: green;
+    }
+
+    .anchor::after {
+        width: 100px;
+        height: 50px;
+        background: green;
+        content: "";
+        position: absolute;
+        position-anchor: auto;
+        position-area: block-end;
+    }
+</style>
+<div class="anchor"></div>
--- a/Source/WebCore/css/CSSProperties.json
+++ b/Source/WebCore/css/CSSProperties.json
@@ -13099,8 +13099,10 @@
        },
         "position-anchor": {
             "animation-type": "discrete",
-            "initial": "auto",
+            "initial": "normal",
             "values": [
+                "normal",
+                "none",
                 "auto"
             ],
             "codegen-properties": {
@@ -13108,7 +13110,7 @@
                 "computed-style-storage-path": ["m_nonInheritedData", "rareData"],
                 "computed-style-storage-kind": "reference",
                 "computed-style-type": "Style::PositionAnchor",
-                "parser-grammar": "auto | <dashed-ident>"
+                "parser-grammar": "normal | none | auto | <dashed-ident>"
             },
             "specification": {
                 "category": "css-anchor-position",
--- a/Source/WebCore/style/AnchorPositionEvaluator.cpp
+++ b/Source/WebCore/style/AnchorPositionEvaluator.cpp
@@ -1222,11 +1222,17 @@ static AnchorsForAnchorName collectAnchorsForAnchorName(const Document& document
     return anchorsForAnchorName;
 }
 
-static AnchorElements findAnchorsForAnchorPositionedElement(const Styleable& anchorPositioned, const HashSet<ResolvedScopedName>& anchorNames, const AnchorsForAnchorName& anchorsForAnchorName)
+static AnchorElements findAnchorsForAnchorPositionedElement(const Styleable& anchorPositioned, const RenderStyle& anchorPositionedStyle, const HashSet<ResolvedScopedName>& anchorNames, const AnchorsForAnchorName& anchorsForAnchorName)
 {
     AnchorElements anchorElements;
 
     for (auto& anchorName : anchorNames) {
+        auto isImplicitAnchorName = anchorName.name() == implicitAnchorElementName().name;
+        auto isDefaultAnchorNone = anchorPositionedStyle.positionAnchor().isNone()
+            || (anchorPositionedStyle.positionAnchor().isNormal() && anchorPositionedStyle.positionArea().isNone());
+        if (isImplicitAnchorName && isDefaultAnchorNone)
+            continue;
+
         auto anchor = findLastAcceptableAnchorWithName(anchorName, anchorPositioned, anchorsForAnchorName);
         anchorElements.add(anchorName, anchor);
     }
@@ -1255,7 +1261,7 @@ void AnchorPositionEvaluator::updateAnchorPositioningStatesAfterInterleavedLayou
         case AnchorPositionResolutionStage::FindAnchors: {
             if (renderer) {
                 // FIXME: Remove the redundant anchorElements member. The mappings are available in anchorPositionedToAnchorMap.
-                state->anchorElements = findAnchorsForAnchorPositionedElement(*anchorPositioned, state->anchorNames, anchorsForAnchorName);
+                state->anchorElements = findAnchorsForAnchorPositionedElement(*anchorPositioned, renderer->style(), state->anchorNames, anchorsForAnchorName);
                 if (isLayoutTimeAnchorPositioned(renderer->style()))
                     renderer->setNeedsLayout();
 
@@ -1699,7 +1705,7 @@ bool AnchorPositionEvaluator::isImplicitAnchor(const RenderStyle& style)
         if (!pseudoElementStyle)
             return false;
         // If we have an explicit anchor name then there is no need for an implicit anchor.
-        if (!pseudoElementStyle->positionAnchor().isAuto())
+        if (pseudoElementStyle->positionAnchor().isName())
             return false;
 
         return pseudoElementStyle->usesAnchorFunctions() || isLayoutTimeAnchorPositioned(*pseudoElementStyle);
--- a/Source/WebCore/style/values/anchor-position/StylePositionAnchor.cpp
+++ b/Source/WebCore/style/values/anchor-position/StylePositionAnchor.cpp
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2026 Suraj Thanugundla <contact@surajt.com>
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1.  Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ * 2.  Redistributions in binary form must reproduce the above copyright
+ * notice, this list of conditions and the following disclaimer in the
+ * documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+ * ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "StylePositionAnchor.h"
+
+#include "CSSKeywordValue.h"
+#include "StyleBuilderState.h"
+
+namespace WebCore {
+namespace Style {
+
+auto CSSValueConversion<PositionAnchor>::operator()(BuilderState& state, const CSSValue& value) -> PositionAnchor
+{
+    if (auto* keywordValue = dynamicDowncast<CSSKeywordValue>(value)) {
+        switch (keywordValue->valueID()) {
+        case CSSValueNormal:
+            return CSS::Keyword::Normal { };
+        case CSSValueNone:
+            return CSS::Keyword::None { };
+        case CSSValueAuto:
+            return CSS::Keyword::Auto { };
+        default:
+            break;
+        }
+    }
+
+    return { toStyleFromCSSValue<AnchorName>(state, value) };
+}
+
+} // namespace Style
+} // namespace WebCore
--- a/Source/WebCore/style/values/anchor-position/StylePositionAnchor.h
+++ b/Source/WebCore/style/values/anchor-position/StylePositionAnchor.h
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2025 Samuel Weinig <sam@webkit.org>
+ * Copyright (C) 2026 Suraj Thanugundla <contact@surajt.com>
  * 
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -30,14 +31,56 @@
 namespace WebCore {
 namespace Style {
 
-// <'position-anchor'> = auto | <anchor-name>
+// <'position-anchor'> = normal | none | auto | <anchor-name>
 // https://drafts.csswg.org/css-anchor-position-1/#propdef-position-anchor
-struct PositionAnchor : ValueOrKeyword<AnchorName, CSS::Keyword::Auto> {
-    using Base::Base;
+struct PositionAnchor {
+    PositionAnchor(CSS::Keyword::Normal keyword)
+        : m_value { keyword }
+    {
+    }
 
-    bool isAuto() const { return isKeyword(); }
-    bool isName() const { return isValue(); }
-    std::optional<AnchorName> tryName() const { return tryValue(); }
+    PositionAnchor(CSS::Keyword::None keyword)
+        : m_value { keyword }
+    {
+    }
+
+    PositionAnchor(CSS::Keyword::Auto keyword)
+        : m_value { keyword }
+    {
+    }
+
+    PositionAnchor(AnchorName&& value)
+        : m_value { WTF::move(value) }
+    {
+    }
+
+    bool isNormal() const { return WTF::holdsAlternative<CSS::Keyword::Normal>(m_value); }
+    bool isNone() const { return WTF::holdsAlternative<CSS::Keyword::None>(m_value); }
+    bool isAuto() const { return WTF::holdsAlternative<CSS::Keyword::Auto>(m_value); }
+    bool isName() const { return WTF::holdsAlternative<AnchorName>(m_value); }
+
+    std::optional<AnchorName> tryName() const
+    {
+        if (auto* name = std::get_if<AnchorName>(&m_value))
+            return *name;
+        return { };
+    }
+
+    template<typename... F> decltype(auto) switchOn(F&&... f) const
+    {
+        return WTF::switchOn(m_value, std::forward<F>(f)...);
+    }
+
+    bool operator==(const PositionAnchor&) const = default;
+
+private:
+    Variant<CSS::Keyword::Normal, CSS::Keyword::None, CSS::Keyword::Auto, AnchorName> m_value;
+};
+
+// MARK: - Conversion
+
+template<> struct CSSValueConversion<PositionAnchor> {
+    auto operator()(BuilderState&, const CSSValue&) -> PositionAnchor;
 };
 
 } // namespace Style
--- a/Source/WebCore/style/values/filter-effects/StyleAppleColorFilter.h
+++ b/Source/WebCore/style/values/filter-effects/StyleAppleColorFilter.h
@@ -38,6 +38,8 @@
 
 namespace WebCore {
 
+class Color;
+
 namespace CSS {
 struct AppleColorFilter;
 struct AppleColorFilterValue;
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/anchor-scroll-implicit-001.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/anchor-scroll-implicit-001.html
@@ -9,8 +9,7 @@
   position: fixed;
   width: 100px;
   height: 50px;
-  left: anchor(left);
-  top: anchor(bottom);
+  position-area: center bottom;
   background: green;
   border: none;
   padding: 0;
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/anchor-scroll-implicit-002.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/anchor-scroll-implicit-002.html
@@ -9,8 +9,7 @@
   position: absolute;
   width: 100px;
   height: 50px;
-  left: anchor(left);
-  top: anchor(bottom);
+  position-area: center bottom;
   background: green;
   border: none;
   padding: 0;
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-001-expected.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-001-expected.html
@@ -1,0 +1,10 @@
+<!doctype html>
+<link rel="help" href="https://drafts.csswg.org/css-anchor-position-1/#position-anchor"/>
+<style>
+    div {
+        background-color: green;
+        height: 100px;
+        width: 100px;
+    }
+</style>
+<div></div>
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-002-expected.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-002-expected.html
@@ -1,0 +1,10 @@
+<!doctype html>
+<link rel="help" href="https://drafts.csswg.org/css-anchor-position-1/#position-anchor"/>
+<style>
+    div {
+        background-color: green;
+        height: 100px;
+        width: 100px;
+    }
+</style>
+<div></div>
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-002.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-002.html
@@ -1,0 +1,23 @@
+<!doctype html>
+<title>position-anchor auto for ::after pseudo element</title>
+<link rel="help" href="https://drafts.csswg.org/css-anchor-position-1/#position-anchor"/>
+<link rel="author" title="Suraj Thanugundla" href="mailto:contact@surajt.com" />
+<link rel="match" href="../reference/ref-filled-green-100px-square-only.html">
+<style>
+    .anchor {
+        width: 100px;
+        height: 50px;
+        background: green;
+    }
+
+    .anchor::after {
+        width: 100px;
+        height: 50px;
+        background: green;
+        content: "";
+        position: absolute;
+        position-anchor: auto;
+        top: anchor(bottom);
+    }
+</style>
+<div class="anchor"></div>
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-basics-expected.txt
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-basics-expected.txt
@@ -1,5 +1,6 @@
 
-FAIL e.style['position-anchor'] = "none" should set the property value assert_not_equals: property should be set got disallowed value ""
+PASS e.style['position-anchor'] = "normal" should set the property value
+PASS e.style['position-anchor'] = "none" should set the property value
 PASS e.style['position-anchor'] = "auto" should set the property value
 PASS e.style['position-anchor'] = "--foo" should set the property value
 PASS e.style['position-anchor'] = "foo-bar" should not set the property value
@@ -7,10 +8,11 @@ PASS e.style['position-anchor'] = "--foo --bar" should not set the property valu
 PASS e.style['position-anchor'] = "--foo, --bar" should not set the property value
 PASS e.style['position-anchor'] = "100px" should not set the property value
 PASS e.style['position-anchor'] = "100%" should not set the property value
-FAIL Property position-anchor value 'none' assert_true: 'none' is a supported value for position-anchor. expected true got false
+PASS Property position-anchor value 'normal'
+PASS Property position-anchor value 'none'
 PASS Property position-anchor value 'auto'
 PASS Property position-anchor value '--foo'
-FAIL Property position-anchor has initial value none assert_equals: expected "none" but got "auto"
+PASS Property position-anchor has initial value normal
 PASS Property position-anchor does not inherit
 PASS CSS Transitions with transition-behavior:allow-discrete: property <position-anchor> from [--foo] to [auto] at (-0.3) should be [--foo]
 PASS CSS Transitions with transition-behavior:allow-discrete: property <position-anchor> from [--foo] to [auto] at (0) should be [--foo]
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-basics.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-basics.html
@@ -15,7 +15,8 @@
 
 <script>
 // position-anchor: <anchor-element>
-// <anchor-element> = none | auto | <dashed-ident>
+// <anchor-element> = normal | none | auto | <dashed-ident>
+test_valid_value('position-anchor', 'normal');
 test_valid_value('position-anchor', 'none');
 test_valid_value('position-anchor', 'auto');
 test_valid_value('position-anchor', '--foo');
@@ -26,13 +27,14 @@ test_invalid_value('position-anchor', '100px');
 test_invalid_value('position-anchor', '100%');
 
 // Computed value: as specified
+test_computed_value('position-anchor', 'normal');
 test_computed_value('position-anchor', 'none');
 test_computed_value('position-anchor', 'auto');
 test_computed_value('position-anchor', '--foo');
 
-// Initial: none
+// Initial: normal
 // Inherited: no
-assert_not_inherited('position-anchor', 'none', '--foo');
+assert_not_inherited('position-anchor', 'normal', '--foo');
 
 // Animation type: discrete
 test_no_interpolation({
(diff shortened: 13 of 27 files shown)


Source/WebCore/style/values/anchor-position/StylePositionAnchor.h
```diff
@@ -30,14 +31,49 @@
...
+        : m_value { keyword }
+    {
+    }
+
+    PositionAnchor(AnchorName&& value)
+        : m_value { WTF::move(value) }
+    {
+    }
+
+    bool isAuto() const { return WTF::holdsAlternative<CSS::Keyword::Auto>(m_value); }
```
weinig, line 59: Please add an isNone() predicate here as well.

weinig (changes requested):
Unlike when adding new properties, for existing properties, please don't add just the parsing/style computation support, as that leaves things in a bad state.

Presumably, somewhere in layout needs to check if the value is `none` as well.

weinig:
Are you saying what none and auto have identical behavior for layout?

fantasai (changes requested):
The initial value is not `none`. See https://www.w3.org/TR/css-anchor-position-1/#position-anchor and https://github.com/w3c/csswg-drafts/issues/13067

Source/WebCore/style/values/anchor-position/StylePositionAnchor.h
surajt6: Sure, I'll add it.

surajt6:
> The initial value is not `none`. See https://www.w3.org/TR/css-anchor-position-1/#position-anchor and [w3c/csswg-drafts#13067](https://github.com/w3c/csswg-drafts/issues/13067)

My bad, I missed that. I'll update it in the next revision of this PR.

nt1m:
> > Though we do need to add `position-anchor: auto` to popover and other features using implicit anchors to avoid breaking those
> 
> Sure, I'll make the change to the popover stylesheet rules.
> 
> > (I think the spec PR should document what user agent stylesheet rules need to be added)
> 
> @nt1m Are there any other user agent stylesheet rules that need to be changed? I can't think of any. FYI, [Chrome](https://chromium-review.googlesource.com/c/chromium/src/+/7155368/9/third_party/blink/renderer/core/html/resources/html.css#b1629) and [Firefox](https://github.com/mozilla-firefox/firefox/commit/ba9378d24791#diff-a95947c9444c5e3d20039f8a5288461667e2739a983699c0aaf9700edae9d0b8) just modify the popover styles, with Chrome modifying `menulist` rules as well.
> 
> As for documenting the user agent stylesheet rules to changed on the spec PR, should I just comment it on the spec PR - [w3c/csswg-drafts#13067](https://github.com/w3c/csswg-drafts/issues/13067)?

popover should be sufficient

nt1m:
To summarize, for this PR I think we need to:
- implement the `normal` value (and its related behavior), use it as initial value
- implement the `none` value (what you have here should suffice)
- add `position-anchor: auto` to popover.css

surajt6:
I think we need to update `PositionedLayoutConstraints::needsAnchor()` to check for `none` as it guards the use of `AnchorPositionEvaluator::defaultAnchorForBox()` - https://searchfox.org/wubkat/source/Source/WebCore/rendering/PositionedLayoutConstraints.cpp#60

LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-none-pseudo-element-implicit.html
```diff
@@ -0,0 +1,24 @@
+<!doctype html>
+<title>position-anchor none for ::after pseudo element</title>
+<link
```
fantasai: Reftests need a `<link rel=match>` pointing at their reference. In this case, I think with a few tweaks, you can just point to `../css/reference/ref-filled-green-100px-square.xht`.

LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-none-pseudo-element-implicit.html
```diff
@@ -0,0 +1,24 @@
...
+    rel="help"
+    href="https://drafts.csswg.org/css-anchor-position-1/#position-anchor"
+/>
+<style>
+    .anchor {
+        width: 100px;
+        height: 100px;
+        background: blue;
```
fantasai: This will be overwritten by the green box, right? Let's make it red instead of blue to highlight visibility as a failure.

LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-normal-pseudo-element-implicit.html
```diff
@@ -0,0 +1,24 @@
+<!doctype html>
+<title>position-anchor normal for ::after pseudo element</title>
+<link
```
fantasai: Need a reference link here, too.

If you want to re-use the green square reference, you could make these both 50px-tall green rectangles. There's no requirement to re-use existing references, but in some cases it's handy--as long as it doesn't make the test harder to understand. (I've seen some tests that contort themselves a lot to match an existing reference, and that's not a good practice imho.)

Source/WebCore/css/CSSProperties.json
```diff
@@ -13133,16 +13133,18 @@
        },
         "position-anchor": {
             "animation-type": "discrete",
-            "initial": "auto",
+            "initial": "normal",
             "values": [
-                "auto"
+                "auto",
+                "none",
+                "normal"
```
fantasai: Nit: Suggest listing 'normal' first, since it's the initial value.

Source/WebCore/css/CSSProperties.json
```diff
@@ -13133,16 +13133,18 @@
...
             ],
             "codegen-properties": {
                 "settings-flag": "cssAnchorPositioningEnabled",
                 "computed-style-storage-path": ["m_nonInheritedData", "rareData"],
                 "computed-style-storage-kind": "reference",
                 "computed-style-type": "Style::PositionAnchor",
-                "parser-grammar": "auto | <dashed-ident>"
+                "parser-grammar": "auto | none | normal | <dashed-ident>"
```
fantasai: Same here.

Source/WebCore/rendering/PositionedLayoutConstraints.cpp
```diff
@@ -100,6 +100,9 @@ void PositionedLayoutConstraints::computeInsets()
 
 bool PositionedLayoutConstraints::needsAnchor() const
 {
+    if (m_style.positionAnchor().isNone())
+        return false;
+
```
fantasai: This check should go into `Style::AnchorPositionEvaluator::defaultAnchorForBox()`. It's not just PositionedLayoutConstraints that needs to fail to look up the default anchor, but also `anchor()` and everything else that tries to look up a default anchor box.

Source/WebCore/style/AnchorPositionEvaluator.cpp
```diff
@@ -1664,8 +1664,12 @@ bool AnchorPositionEvaluator::isImplicitAnchor(const RenderStyle& style)
...
         if (!pseudoElementStyle)
             return false;
-        // If we have an explicit anchor name then there is no need for an implicit anchor.
-        if (!pseudoElementStyle->positionAnchor().isAuto())
+
+        // pseudo-element has no implicit anchor element for certain values of position-anchor and position-area
+        // https://www.w3.org/TR/css-anchor-position-1/#position-anchor
+        if (pseudoElementStyle->positionAnchor().isNone() || pseudoElementStyle->positionAnchor().isName())
+            return false;
+        if (pseudoElementStyle->positionAnchor().isNormal() && pseudoElementStyle->positionArea().isNone())
```
fantasai: I think I'd like to get @anttijk's take on whether to have this check here, or only in Style::AnchorPositionEvaluator::defaultAnchorForBox(). But if we are having it here, it should be a short circuit at the top of the function, not just for pseudo-elements.

fantasai (changes requested):
Noted a few things to adjust. I also am wondering if we are invalidating things correctly somewhere? Because I'm not seeing this modify any invalidation code... but I'm also not successfully finding existing invalidation code...

It would be good to get @anttijk's review on the style calculation parts.

LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-normal-pseudo-element-implicit.html
surajt6: I can update the test to re-use the green square reference

Source/WebCore/css/CSSProperties.json
surajt6: Sure,  I'll keep that in mind. I listed them alphabetically.

Source/WebCore/style/AnchorPositionEvaluator.cpp
surajt6: I will look deeper into this as well. I'm working on wrapping my head around and trying to understand anchor positioning layout code.

Source/WebCore/rendering/PositionedLayoutConstraints.cpp
surajt6: This check doesn’t apply to `anchor()` as I think `anchor()` follows a different code path to lookup default anchors. The anchor lookup is done in [￼`AnchorPositionEvaluator::evaluate`￼](https://searchfox.org/wubkat/source/Source/WebCore/style/AnchorPositionEvaluator.cpp#857) itself. More specifically, it’s in [`AnchorPositionEvaluator::findAnchorForAnchorFunctionAndAttemptResolution`￼](https://searchfox.org/wubkat/source/Source/WebCore/style/AnchorPositionEvaluator.cpp#747).  We should also have an a check for `position-area` in `findAnchorForAnchorFunctionAndAttemptResolution` and return no element so that `anchor()` does not use the implicit anchor as it’s default anchor.

Source/WebCore/style/AnchorPositionEvaluator.cpp
surajt6: I think this check should go in `AnchorPositionEvaluator::defaultAnchorForBox` and not in `AnchorPositionEvaluator::isImplicitAnchor`. The [spec says](https://www.w3.org/TR/css-anchor-position-1/#valdef-position-anchor-none) that the box has no default anchor when the when `position-anchor` is `none`. It means that a box will still have an implicit anchor but it’s just not used as it’s default anchor. The only check here should be to verify if `position-anchor` is an anchor name, in which case the box doesn’t need an implicit anchor. Please lmk if I’m missing anything.

Source/WebCore/style/AnchorPositionEvaluator.cpp
```diff
@@ -759,6 +759,12 @@ CheckedPtr<RenderBoxModelObject> AnchorPositionEvaluator::findAnchorForAnchorFun
     if (!isValid())
         return { };
 
+    // https://www.w3.org/TR/css-anchor-position-1/#propdef-position-anchor
+    auto defaultAnchorNone = style.positionAnchor().isNone()
+        || (style.positionAnchor().isNormal() && style.positionArea().isNone());
```
fantasai: I wonder if this logic (which we use in two places at least) could be wrapped into a helper function on RenderStyle or something.

Source/WebCore/style/AnchorPositionEvaluator.cpp
anttijk, line 1699: I don't think popover invokers with `position-anchor:none` should be implicit anchors either.

Source/WebCore/style/AnchorPositionEvaluator.cpp
```diff
@@ -1665,7 +1671,7 @@ bool AnchorPositionEvaluator::isImplicitAnchor(const RenderStyle& style)
         if (!pseudoElementStyle)
             return false;
         // If we have an explicit anchor name then there is no need for an implicit anchor.
-        if (!pseudoElementStyle->positionAnchor().isAuto())
+        if (pseudoElementStyle->positionAnchor().isName())
             return false;
```
anttijk, line 1709: Why this change? `auto` is the only value that should allow implicit anchors.

Source/WebCore/style/AnchorPositionEvaluator.cpp
```diff
@@ -1685,6 +1691,12 @@ CheckedPtr<RenderBoxModelObject> AnchorPositionEvaluator::defaultAnchorForBox(co
     if (!box.element())
         return nullptr;
 
+    // https://www.w3.org/TR/css-anchor-position-1/#propdef-position-anchor
+    auto defaultAnchorNone = box.style().positionAnchor().isNone()
+        || (box.style().positionAnchor().isNormal() && box.style().positionArea().isNone());
+    if (defaultAnchorNone)
+        return nullptr;
```
anttijk: This is wrong place for the test. If anchorPositionedToAnchorMap was in correct state there would be no mapping there.

Source/WebCore/style/AnchorPositionEvaluator.cpp
anttijk: Might be better to model this by having `defaultAnchorName()` return `nullopt`

anttijk (changes requested):

Source/WebCore/style/AnchorPositionEvaluator.cpp
fantasai: No, we're also adding `normal` in this patch, which sometimes allows them and sometimes doesn't.

Source/WebCore/style/AnchorPositionEvaluator.cpp
surajt6: Makes sense. It ensures anchorPositionedToAnchorMap is in correct state.

Source/WebCore/style/AnchorPositionEvaluator.cpp
surajt6: > The *position-anchor* property specifies the ***default anchor element***, which is used by *position-area*, *position-try*, and (by default) all *anchor functions* applied to **this element**.

The [spec](https://drafts.csswg.org/css-anchor-position-1/#position-anchor) says `position-anchor` only effects the default anchor of element the property is applied to. 

The implicit anchor for a popover is its invoker. If we don’t want that, then we use `position-anchor:none` on the popover. The check in  `defaultAnchorName` should take care of that.

Why do we need to check the style of the invoker for `position-anchor:none`? My assumption is that the anchor positioned element (and not the anchor) gets to choose it’s default anchor based on the value of `position-anchor`. So we should check for `position-anchor: none` on the anchor positioned element rather than the anchor.

Consider the following example where the element `#button::after` uses `#button` as it’s implicit anchor -
```html
<button id=button popovertarget=popover>Button</button>
<div popover id=popover>Popover</div>
<style>
#button::after {
    width: 100px;
    height: 100px;
    background: green;
    content: "";
    position: absolute;
    position-anchor: auto;
    position-area: top;
}
#button {
  position:relative;
  top:100px;
  left:100px;
  position-anchor:none;
}
#popover {
  border:1px solid black;
  inset: auto;
  margin:0;
  padding:0;
  position-area: top left;
}
</style>
```

If we add `position-anchor:none` check for popover invoker (`#button`) then `isImplicitAnchor` will return `false` even though `#button` is an implicit anchor for `#button::after`. Please lmk if this makes sense.

Source/WebCore/style/AnchorPositionEvaluator.cpp
surajt6: +1 to @fantasai. a combination of `position-anchor: normal` and `position-area: <value>` allows implicit anchors.

Source/WebCore/style/AnchorPositionEvaluator.cpp
```diff
@@ -767,10 +767,14 @@ CheckedPtr<RenderBoxModelObject> AnchorPositionEvaluator::findAnchorForAnchorFun
         return WTF::makeUnique<AnchorPositionedState>();
     }).iterator->value.get();
 
+    auto defaultAnchorName = AnchorPositionEvaluator::defaultAnchorName(style);
```
surajt6: I used `AnchorPositionEvaluator::defaultAnchorName` instead of just `defaultAnchorName`. Otherwise, the following compiler error shows up

```
Variable 'defaultAnchorName' declared with deduced type 'auto' cannot appear in its own initializer 
(clang auto_variable_cannot_appear_in_own_initializer)
```

Source/WebCore/style/AnchorPositionEvaluator.cpp
```diff
@@ -1692,7 +1703,11 @@ CheckedPtr<RenderBoxModelObject> AnchorPositionEvaluator::defaultAnchorForBox(co
     if (it == anchorPositionedMap.end())
         return nullptr;
 
-    auto anchorName = ResolvedScopedName::createFromScopedName(element, defaultAnchorName(box.style()));
+    auto defaultAnchorName = AnchorPositionEvaluator::defaultAnchorName(box.style());
+    if (!defaultAnchorName)
+        return nullptr;
```
surajt6: The only reason for having this check is because `defaultAnchorName` now returns an optional value. The predicate in `if` just before this code evaluates to `true` if `position-anchor: none` and control exits from this function. `defaultAnchorName` is guaranteed to not be empty at this point. I can remove the `if` check and the call to `defaultAnchorName` if needed.

Source/WebCore/style/AnchorPositionEvaluator.cpp
surajt6: Agreed. If we move this check to `defaultAnchorName` then it's not needed here. I added a note on this section on the latest version of this PR.

Source/WebCore/style/values/filter-effects/StyleAppleColorFilter.h
```diff
@@ -37,6 +37,8 @@
 
 namespace WebCore {
 
+class Color;
+
```
fantasai, line 42: Is this change needed?

Source/WebCore/style/AnchorPositionEvaluator.cpp
fantasai: Let's not call this function when we don't need to.

Source/WebCore/style/AnchorPositionEvaluator.cpp
```diff
@@ -1334,9 +1338,11 @@ void AnchorPositionEvaluator::updateAnchorPositionedStateForDefaultAnchorAndPosi
         return makeUnique<AnchorPositionedState>();
     }).iterator->value.get();
 
-    if (shouldResolveDefaultAnchor) {
+    auto defaultAnchorName = AnchorPositionEvaluator::defaultAnchorName(style);
```
fantasai: Same here.

fantasai (changes requested):

Source/WebCore/style/values/filter-effects/StyleAppleColorFilter.h
surajt6: If I dont add this then build fails with `StyleAppleColorFilter.h:92:34: error: no type named 'Color' in namespace 'WebCore'`. It’s needed since I added a new file and I think its because of unified sources - [Unified Build System - WebKit Documentation](https://docs.webkit.org/Deep%20Dive/Build/UnifiedBuilds.html#build-failures-with-unified-sources)

Source/WebCore/style/AnchorPositionEvaluator.cpp
surajt6: Please check my recent comment on the PR for response.

Source/WebCore/style/AnchorPositionEvaluator.cpp
surajt6: Please check my recent comment on the PR for response.

surajt6:
> Please make sure to run the export script for web platform tests (export-w3c-test-changes)

Sure, will do. Thanks!

fantasai (commented):
Changes look good to me. :) Would be good if @anttijk took a look, though.

anttijk:
I'm fine with this if @tuankiet65 is fine with it.

anttijk (approved):

Source/WebCore/style/values/anchor-position/StylePositionAnchor.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "StylePositionAnchor.h"
+
+#include "StyleBuilderState.h"
+
+namespace WebCore {
+namespace Style {
+
+auto CSSValueConversion<PositionAnchor>::operator()(BuilderState& state, const CSSValue& value) -> PositionAnchor
+{
+    if (auto* primitiveValue = dynamicDowncast<CSSPrimitiveValue>(value)) {
```
weinig: This will need to use CSSKeywordValue now.

Source/WebCore/style/values/filter-effects/StyleAppleColorFilter.h
tuankiet65: Make sense, I'll consider this resolved.

Source/WebCore/style/AnchorPositionEvaluator.cpp
```diff
@@ -1249,11 +1249,17 @@ static AnchorsForAnchorName collectAnchorsForAnchorName(const Document& document
...
+AnchorElements AnchorPositionEvaluator::findAnchorsForAnchorPositionedElement(const Element& anchorPositionedElement, const RenderStyle& style, const HashSet<ResolvedScopedName>& anchorNames, const AnchorsForAnchorName& anchorsForAnchorName)
 {
     AnchorElements anchorElements;
 
     for (auto& anchorName : anchorNames) {
+        auto isImplicitAnchorName = anchorName.name() == implicitAnchorElementName().name;
+        auto isDefaultAnchorNone = style.positionAnchor().isNone()
+            || (style.positionAnchor().isNormal() && style.positionArea().isNone());
+        if (isImplicitAnchorName && isDefaultAnchorNone)
+            continue;
```
tuankiet65, line 1234: So we keep how `defaultAnchorName` works (returns name if `position-anchor` is name, or return implicit anchor name if it's `none/auto/normal`), then here we reject the implicit anchor name if it's either `none` or `normal` and `position-area` is not specified.

This works but I think it's rather inelegant, I'd rather that we change `defaultAnchorName` instead. But it seems like a big change, because we'd have to make `defaultAnchorName` returns optional, and change its caller to accomodate that (and maybe change the caller of callers too...). So I'll say let's go with this for now, and we can clean it up like I said later.

Source/WebCore/style/AnchorPositionEvaluator.cpp
tuankiet65: Yeah I'm not sure what Antti means by that, I think `position-anchor: none` doesn't change if the popover invoker is implicit anchor or not.

Source/WebCore/style/AnchorPositionEvaluator.cpp
tuankiet65: Make sense, consider this resolved.

LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-001.html
```diff
@@ -0,0 +1,23 @@
+<!doctype html>
+<title>position-anchor auto for ::after pseudo element</title>
+<link rel="help" href="https://drafts.csswg.org/css-anchor-position-1/#position-anchor"/>
+<link rel="author" name="Suraj Thanugundla" href="mailto:contact@surajt.com" />
```
tuankiet65: The `name` attribute should be `title`, you should fix this in other test cases too:

```
<link rel="author" title="Suraj Thanugundla" href="mailto:contact@surajt.com" />
```

tuankiet65 (commented):
Thank you for the (very exhaustive!) test cases

Source/WebCore/style/values/anchor-position/StylePositionAnchor.cpp
surajt6: Will update and push a new revision.

LayoutTests/imported/w3c/web-platform-tests/css/css-anchor-position/position-anchor-auto-pseudo-element-implicit-001.html
surajt6: Sure, I will fix it and send a new revision for the PR

surajt6:
@fantasai @anttijk @tuankiet65 @weinig thank you for reviewing this PR. I will update it with corrections and send a new revision shortly :)

tuankiet65 (commented):
Looks good, should be ready to merge. Antti already approved this, so you should add the `merge-queue` label, and the bot will come by to merge it. Congratulations on your first PR 🎉 

On exporting to WPT: you should use `Tools/Scripts/export-w3c-test-changes` to export new tests to WPT. `--help` shows you how to use it, but for reference, I use `Tools/Scripts/export-w3c-test-changes -c -g HEAD -b [bugzilla bug id]`.
* If you export before merging this PR, the WPT bot won't approve the WPT export PR until this PR has been merged. (there could be a few hours delay)
* If you export after merging this PR, the WPT bot should approve instantly.

I like to export before merging to see if the new tests also pass on Chrome/Firefox, as WPT CI will run your tests against other browsers.

Do you want to follow up with a clean up? (refer to [my comment here](https://github.com/WebKit/WebKit/pull/59717#discussion_r3133910676))

surajt6:
@tuankiet65 thanks for getting back. I can keep this PR open until the wpt tests land. I will raise a follow up PR to clean up as you suggested. 

I tried running the `export-w3c-test-changes` command but it failed with - `error: html/dom/elements/global-attributes/the-anchor-attribute-003.tentative.html: does not exist in index`. I had to remove the `the-anchor-attribute-003.tentative.html` file from the commit and try it again. It worked and I raised a PR in the wpt repo - https://github.com/web-platform-tests/wpt/pull/59532.

Turns out `the-anchor-attribute-003.tentative.html` test was removed from wpt on 08/21/25 - https://github.com/web-platform-tests/wpt/commit/90bf88acc8231c71479b44d9733fca3e779c1b0b. Should I remove that test from WebKit's repo as well?

tuankiet65:
> I will raise a follow up PR to clean up as you suggested.

Thank you

> It worked and I raised a PR in the wpt repo - https://github.com/web-platform-tests/wpt/pull/59532.

Usually the WPT bot will say "it will be approved automatically once the downstream patch is r+" for WebKit exports, but it didn't. Maybe it's because you aren't a member in WPT group. I r+ it already, but let's wait for Firefox to review too as you modified an Interop test. (`css/css-anchor-position/position-anchor-basics.html`)

If you don't know, [Interop](https://wpt.fyi/interop-2026) is an effort to ensure every browsers implement important features. This is done by picking features each year, then scores browsers based on how many tests they pass. The change to the interop test might affect other browsers' scores, hence the extra scrutiny. I think Firefox will be ok with the change, so you can merge this PR first (of course fix the conflicts first) and deal with the WPT export later

> Turns out the-anchor-attribute-003.tentative.html test was removed from wpt on 08/21/25 - https://github.com/web-platform-tests/wpt/commit/90bf88acc8231c71479b44d9733fca3e779c1b0b. Should I remove that test from WebKit's repo as well?

You can do that, but the smarter thing is to import the entire folder to sync it with WPT, you can do that with `import-w3c-tests`. Here you would want to import `‎html/dom/elements/global-attributes/`, which is the directory that contains the removed text. Basically: you start from a clean checkout, run the script, add the changes, create PR for it.

surajt6:
@tuankiet65 I've updated two wpt tests to use `position-area: center bottom;` as mentioned in https://github.com/web-platform-tests/wpt/pull/59532. Please help land/merge these changes if you think they're ready as I can't add any labels to webkit PRs. I'm not on the Contributors list yet and I will add myself soon.

> you can do that with import-w3c-tests

Will do. Thanks for pointing it out. To summarize, 2 follow ups needed for this PR - one to cleanup the deleted wpt test and another is to refactor as you suggested.

tuankiet65:
> Please help land/merge these changes if you think they're ready as I can't add any labels to webkit PRs

Yes I forgot that non-contributors can't do that 😅 I've added merge-queue for you.

> To summarize, 2 follow ups needed for this PR - one to cleanup the deleted wpt test and another is to refactor as you suggested.

Sounds good.

tuankiet65:
Not really affecting the code, but you should clean up the commit message next time, there's this bits in the message:

```
# Conflicts:
#       Source/WebCore/Sources.txt
```

PR #59726, opened 2026-03-02
[GStreamer] Expose VIDEO_DECODING_LIMIT through an environment variable

eocanha:
#### de98f6bb5014db381fd8fe98ecf8a5977d6c3521
<pre>
[GStreamer] Expose VIDEO_DECODING_LIMIT through an environment variable
<a href="https://bugs.webkit.org/show_bug.cgi?id=308969">https://bugs.webkit.org/show_bug.cgi?id=308969</a>

Reviewed by Alicia Boya Garcia and Xabier Rodriguez-Calvar.

The RDK-E downstream platform would like to have a single unified build
for many kind of devices. For that reason, the platform compile-time
VIDEO_DECODING_LIMIT restriction aren&apos;t appropriate anymore, since the
same build can be used for devices with different runtime limits. A way
to specify the decoding limit ar runtime would be desirable.

An environment variable looks like a better way to do it than a runtime
setting, since the limit is unlikely to change at runtime and likely to
change per device (with a value consistent on every execution).
Moreover, having a setting would mean extra changes in the dependent
packages that embed WebKit, while having en environmente variable would
be less intrusive and would only mean a change in the (easier to
maintain) startup scripts.

This change adds WEBKIT_GST_VIDEO_DECODING_LIMIT env on top of
VIDEO_DECODING_LIMIT compile time option to limit video decoding
capabilities. WEBKIT_GST_VIDEO_DECODING_LIMIT can now be set in runtime
and it takes precedence over compile time definition
VIDEO_DECODING_LIMIT.

See: <a href="https://github.com/WebPlatformForEmbedded/WPEWebKit/pull/1625">https://github.com/WebPlatformForEmbedded/WPEWebKit/pull/1625</a>
     <a href="https://github.com/WebPlatformForEmbedded/WPEWebKit/pull/1626">https://github.com/WebPlatformForEmbedded/WPEWebKit/pull/1626</a>

Original author: Andrzej Surdej &lt;Andrzej_Surdej@comcast.com&gt;

* Source/WebCore/platform/graphics/gstreamer/GStreamerRegistryScanner.cpp:
(videoDecoderLimitsDefaults): Deleted.

* Source/WebCore/platform/graphics/gstreamer/GStreamerRegistryScanner.cpp:
(WebCore::parseVideoDecodingLimit): Use a StringView to prevent extra copies.
(WebCore::resolveVideoDecodingLimits): Decide the right decoding limits considering the build time option and the environment variable value.
(WebCore::GStreamerRegistryScanner::isContentTypeSupported const): Parse env var and only require the build ifdefs when strictly needed.
(videoDecoderLimitsDefaults): Deleted.

Canonical link: <a href="https://commits.webkit.org/308552@main">https://commits.webkit.org/308552@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/7fb4f6e1beac17bc87ddb893e4ea4e4168694a0e

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/147849 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/20534 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/14127 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/156532 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/101264 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/149722 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/20992 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/20438 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/113983 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/81281 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/150811 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/16231 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/132797 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94744 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/15378 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/13165 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/3972 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/124983 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/10701 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/158867 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/2001 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/12192 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/122013 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/20333 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/17093 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/122214 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/31309 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/20344 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/132498 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/76483 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/17720 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/9263 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/19949 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/83711 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/19678 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/19829 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/19736 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/platform/graphics/gstreamer/GStreamerRegistryScanner.cpp
+++ b/Source/WebCore/platform/graphics/gstreamer/GStreamerRegistryScanner.cpp
@@ -44,7 +44,11 @@
 #include "VideoEncoderPrivateGStreamer.h"
 #endif
 
-namespace {
+namespace WebCore {
+
+GST_DEBUG_CATEGORY_STATIC(webkit_media_gst_registry_scanner_debug);
+#define GST_CAT_DEFAULT webkit_media_gst_registry_scanner_debug
+
 struct VideoDecodingLimits {
     unsigned mediaMaxWidth = 0;
     unsigned mediaMaxHeight = 0;
@@ -56,22 +60,18 @@ struct VideoDecodingLimits {
     {
     }
 };
-}
 
-#ifdef VIDEO_DECODING_LIMIT
-static std::optional<VideoDecodingLimits> videoDecoderLimitsDefaults()
+// Parses a video decoding limit string in format WIDTHxHEIGHT@FRAMERATE.
+static std::optional<VideoDecodingLimits> parseVideoDecodingLimit(const StringView& videoDecodingLimit)
 {
-    // VIDEO_DECODING_LIMIT should be in format: WIDTHxHEIGHT@FRAMERATE.
-    String videoDecodingLimit(String::fromUTF8(VIDEO_DECODING_LIMIT));
-
     if (videoDecodingLimit.isEmpty())
         return { };
 
     Vector<String> entries;
 
-    // Extract frame rate part from the VIDEO_DECODING_LIMIT: WIDTHxHEIGHT@FRAMERATE.
-    videoDecodingLimit.split('@', [&entries](StringView item) {
-        entries.append(item.toString());
+    // Extract frame rate part: WIDTHxHEIGHT@FRAMERATE.
+    videoDecodingLimit.toStringWithoutCopying().split('@', [&entries](StringView item) {
+        entries.append(item.toStringWithoutCopying());
     });
 
     if (entries.size() != 2)
@@ -99,12 +99,39 @@ static std::optional<VideoDecodingLimits> videoDecoderLimitsDefaults()
 
     return { VideoDecodingLimits(width.value(), height.value(), frameRate.value()) };
 }
+
+// Returns the active VideoDecodingLimits, resolved once at first call.
+// WEBKIT_GST_VIDEO_DECODING_LIMIT env var takes precedence over the compile-time VIDEO_DECODING_LIMIT.
+// Format for both: WIDTHxHEIGHT@FRAMERATE (e.g. "1920x1080@30").
+static VideoDecodingLimits* resolveVideoDecodingLimits()
+{
+    static std::optional<VideoDecodingLimits> limits;
+    static std::once_flag onceFlag;
+    std::call_once(onceFlag, [] {
+        if (const char* envLimit = g_getenv("WEBKIT_GST_VIDEO_DECODING_LIMIT")) {
+            GST_DEBUG("WEBKIT_GST_VIDEO_DECODING_LIMIT env var is set: %s", envLimit);
+            limits = parseVideoDecodingLimit(StringView::fromLatin1(envLimit));
+            if (!limits)
+                GST_WARNING("Parsing WEBKIT_GST_VIDEO_DECODING_LIMIT env var failed: %s", envLimit);
+        }
+#ifdef VIDEO_DECODING_LIMIT
+        if (!limits) {
+            GST_DEBUG("VIDEO_DECODING_LIMIT compile-time definition is set: %s", VIDEO_DECODING_LIMIT);
+            limits = parseVideoDecodingLimit(String::fromLatin1(VIDEO_DECODING_LIMIT));
+            if (!limits) {
+                GST_WARNING("Parsing VIDEO_DECODING_LIMIT failed: %s", VIDEO_DECODING_LIMIT);
+                ASSERT_NOT_REACHED();
+                return;
+            }
+        }
 #endif
-
-namespace WebCore {
-
-GST_DEBUG_CATEGORY_STATIC(webkit_media_gst_registry_scanner_debug);
-#define GST_CAT_DEFAULT webkit_media_gst_registry_scanner_debug
+        if (limits) {
+            GST_DEBUG("Video decoding limits: max width=%u, max height=%u, max frame rate=%u",
+                limits->mediaMaxWidth, limits->mediaMaxHeight, limits->mediaMaxFrameRate);
+        }
+    });
+    return limits ? &*limits : nullptr;
+}
 
 // We shouldn't accept media that the player can't actually play.
 // AAC supports up to 96 channels.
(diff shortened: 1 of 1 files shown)


Source/WebCore/platform/graphics/gstreamer/GStreamerRegistryScanner.cpp
philn: Maybe this should be a CStringView or similar, I am deferring to strings expert @calvaris

eocanha: Have into account that this code has to be backported down to 2.38, where the current implementation of CStringView might not be present (and won't ever be).

ntrrgc (approved):
LGTM.

Source/WebCore/platform/graphics/gstreamer/GStreamerRegistryScanner.cpp
calvaris: We have this into account, for upstream it should use CStringView. When backporting downstream you should use String::fromLatin1.

calvaris (approved):
Please change what I requested before landing.

Source/WebCore/platform/graphics/gstreamer/GStreamerRegistryScanner.cpp
calvaris: This should also receive CStringView.

eocanha: My understanding is that [String::fromUTF8(const char*)](https://github.com/WebKit/WebKit/blob/main/Source/WTF/wtf/text/WTFString.h#L279) will [internally create](https://github.com/WebKit/WebKit/blob/main/Source/WTF/wtf/StdLibExtras.h#L922) an std::span<const char> using [unsafeSpan()](https://github.com/WebKit/WebKit/blob/main/Source/WTF/wtf/text/StringCommon.h#L55), which doesn't allocate new memory nor generates copies (since span is just [a way to refer to a region of memory in the same way (pointer, length) used to do in C](https://www.cppstories.com/2023/span-cpp20/)).

With that in mind, we're already avoiding copies by using the String::fromUTF8. What advantage would a CStringView bring, given that we don't have any way to use all the String helper functions (such as [split()](https://github.com/WebKit/WebKit/pull/59726/changes/BASE..9aa28dcfffd3b4f3ea4a6f5de573f2b165afa763#diff-f1e6d0a710906abde361f4b0ee720809486455e7ac160d389f36afa1a939994aL73))?

calvaris: You need to [dig](https://github.com/WebKit/WebKit/blob/main/Source/WTF/wtf/text/WTFString.cpp#L55) [much](https://github.com/WebKit/WebKit/blob/main/Source/WTF/wtf/text/StringImpl.cpp#L282) [deeper](https://github.com/WebKit/WebKit/blob/main/Source/WTF/wtf/text/StringImpl.cpp#L262) into the stack call to see why String::fromUTF8 is creating a copy when it checks for the character conversion.

You're right in one thing I missed. You can't do good split on a CStringView. I'd suggest you try StringView::fromLatin1 to see if it's confortable with current code and if not, you can land with String, otherwise please use StringView.

calvaris, line 106: Another necessary change is that you need to move this after the declaration of the debug category or log won't show up properly.

calvaris (approved):
.

Source/WebCore/platform/graphics/gstreamer/GStreamerRegistryScanner.cpp
eocanha: As we talked in private (about CStringView not having .split() and StringView being a good alternative), I'm using StringView here.

aperezdc:
Backported into `webkitglib/2.52` as commit f18d440eaafe7b9b9f4cdb7aa6ab598072b27a1e

PR #59807, opened 2026-03-03
Consolidate lazy load observers

credod:
#### f4ad91fe0531241659e2982d2254526243fac2b7
<pre>
Reviving <a href="https://github.com/WebKit/WebKit/pull/59807">https://github.com/WebKit/WebKit/pull/59807</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=309078">https://bugs.webkit.org/show_bug.cgi?id=309078</a>

Reviewed by NOBODY (OOPS!).

Consolidates LazyLoadImageObserver, LazyLoadVideoObserver, LazyLoadFrameObserver, and LazyLoadModelObserver to a single LazyLoadElementObserver.

* Source/WebCore/Modules/model-element/HTMLModelElement.cpp:
(WebCore::HTMLModelElement::~HTMLModelElement):
(WebCore::HTMLModelElement::stop):
(WebCore::HTMLModelElement::insertionSteps):
(WebCore::HTMLModelElement::removingSteps):
(WebCore::HTMLModelElement::lazyLoadIntersectionCallbackInvoked):
(WebCore::HTMLModelElement::viewportIntersectionChanged): Deleted.
* Source/WebCore/Modules/model-element/HTMLModelElement.h:
* Source/WebCore/Modules/model-element/LazyLoadModelObserver.cpp: Removed.
* Source/WebCore/Modules/model-element/LazyLoadModelObserver.h: Removed.
* Source/WebCore/Sources.txt:
(WebCore::Document::commonTeardown):
(WebCore::Document::lazyLoadElementObserver):
(WebCore::Document::lazyLoadImageObserver): Deleted.
(WebCore::Document::lazyLoadModelObserver): Deleted.
(WebCore::Document::lazyLoadVideoObserver): Deleted.
* Source/WebCore/html/HTMLIFrameElement.cpp:
(WebCore::HTMLIFrameElement::attributeChanged):
(WebCore::HTMLIFrameElement::referrerPolicy const):
(WebCore::HTMLIFrameElement::shouldLoadFrameLazily):
(WebCore::HTMLIFrameElement::isLazyLoadObserverActive const):
(WebCore::HTMLIFrameElement::loadDeferredFrame):
(WebCore::HTMLIFrameElement::lazyLoadIntersectionCallbackInvoked):
(WebCore::HTMLIFrameElement::lazyLoadFrameObserver): Deleted.
* Source/WebCore/html/HTMLIFrameElement.h:
* Source/WebCore/html/HTMLImageElement.cpp:
(WebCore::HTMLImageElement::lazyLoadIntersectionCallbackInvoked):
* Source/WebCore/html/HTMLImageElement.h:
* Source/WebCore/html/HTMLVideoElement.cpp:
(WebCore::HTMLVideoElement::~HTMLVideoElement):
(WebCore::HTMLVideoElement::create):
(WebCore::HTMLVideoElement::didMoveToNewDocument):
(WebCore::HTMLVideoElement::lazyLoadIntersectionCallbackInvoked):
(WebCore::HTMLVideoElement::viewportIntersectionChanged): Deleted.
* Source/WebCore/html/HTMLVideoElement.h:
* Source/WebCore/html/LazyLoadElementObserver.cpp: Renamed from Source/WebCore/html/LazyLoadImageObserver.cpp.
(WebCore::LazyLoadElementObserver::observe):
(WebCore::LazyLoadElementObserver::unobserve):
(WebCore::LazyLoadElementObserver::intersectionObserver):
(WebCore::LazyLoadElementObserver::isObserved const):
* Source/WebCore/html/LazyLoadElementObserver.h: Renamed from Source/WebCore/html/LazyLoadImageObserver.h.
* Source/WebCore/html/LazyLoadFrameObserver.cpp: Removed.
* Source/WebCore/html/LazyLoadFrameObserver.h: Removed.
* Source/WebCore/html/LazyLoadVideoObserver.cpp: Removed.
* Source/WebCore/html/LazyLoadVideoObserver.h: Removed.
* Source/WebCore/loader/ImageLoader.cpp:
(WebCore::ImageLoader::didUpdateCachedImage):
(WebCore::ImageLoader::notifyFinished):
(WebCore::ImageLoader::resetLazyImageLoading):
</pre>


Source/WebCore/html/LazyLoadElementObserver.cpp
```diff
@@ -74,25 +83,25 @@ class LazyImageLoadIntersectionObserverCallback final : public IntersectionObser
...
         observer->unobserve(element);
 }
 
-IntersectionObserver* LazyLoadImageObserver::intersectionObserver(Document& document)
+IntersectionObserver* LazyLoadElementObserver::intersectionObserver(Document& document)
 {
     if (!m_observer) {
-        auto callback = LazyImageLoadIntersectionObserverCallback::create(document);
+        auto callback = LazyLoadIntersectionObserverCallback::create(document);
         static NeverDestroyed<const String> lazyLoadingScrollMarginFallback(MAKE_STATIC_STRING_IMPL("100%"));
```
credod, line 106: One thing to call out here is that the lazy loading behavior for `<iframe>` will change. `LazyLoadFrameObserver` [didn't previously specify custom `scrollMargin`](https://github.com/WebKit/WebKit/blob/8af162c62e03384c11d74f8a4301b66790e90e0c/Source/WebCore/html/LazyLoadFrameObserver.cpp#L112), meaning that `<iframe>` elements would only "intersect" when they entered the viewport. Here, we are setting this value to be `100%`, which is what was done for [`LazyLoadImageObserver`](https://github.com/WebKit/WebKit/blob/fbd2a11004f46e9ace6f44ba52ab7863d547e8db/Source/WebCore/html/LazyLoadImageObserver.cpp#L96) and [`LazyLoadModelObserver`](https://github.com/WebKit/WebKit/blob/fbd2a11004f46e9ace6f44ba52ab7863d547e8db/Source/WebCore/Modules/model-element/LazyLoadModelObserver.cpp#L99). 

The [HTML spec](https://html.spec.whatwg.org/#lazy-loading-attributes:lazy-load-intersection-observer) suggests that we should be reusing the lazy load observer instance across `` and `<iframe>`:

> For [img](https://html.spec.whatwg.org/#the-img-element) and [iframe](https://html.spec.whatwg.org/#the-iframe-element) elements that [will lazy load](https://html.spec.whatwg.org/#will-lazy-load-element-steps), these steps are run from the [lazy load intersection observer](https://html.spec.whatwg.org/#lazy-load-intersection-observer)'s callback or when their [lazy loading attribute](https://html.spec.whatwg.org/#lazy-loading-attribute) is set to the [Eager](https://html.spec.whatwg.org/#attr-loading-eager-state) state. This causes the element to continue loading.

Therefore, this change would ensure that Webkit is following this standard outlined by the spec but I'm curious to know what people think about this behavioral change.

smfr: Seems OK, but I'd like the behavior change to be confirmed by a new test (or a newly passing test). Are there any WPT for iframe lazy loading that use `scroll-margin`?

Source/WebCore/html/LazyLoadElementObserver.cpp
```diff
@@ -58,12 +65,14 @@ class LazyImageLoadIntersectionObserverCallback final : public IntersectionObser
...
 
         for (auto& entry : entries) {
-            if (!entry->isIntersecting())
-                continue;
-            if (RefPtr element = dynamicDowncast<HTMLImageElement>(entry->target())) {
-                element->loadDeferredImage();
-                element->document().lazyLoadImageObserver().unobserve(*element, element->document());
-            }
+            if (RefPtr element = dynamicDowncast<HTMLImageElement>(entry->target()))
+                element->lazyLoadIntersectionCallbackInvoked(entry->isIntersecting());
```
smfr, line 69: Maybe Element could have a virtual `lazyLoadIntersectionCallbackInvoked`, so you don't need to `dynamicDowncast` here

smfr, line 69: Maybe Element could have a virtual `lazyLoadIntersectionCallbackInvoked`, so you don't need to `dynamicDowncast` here

smfr (approved):

Source/WebCore/html/LazyLoadElementObserver.cpp
credod: WPT tests typically seem to allow [generous scroll margins](https://github.com/web-platform-tests/wpt/blob/a40a308d4aa553aa7fd2785391a1dce4858859b1/html/semantics/embedded-content/the-iframe-element/iframe-loading-lazy.html#L84) so that browsers can implement lazy-loading to their liking. I did run existing WPT tests against this patch and seems like this implementation matches existing behavior, and additionally gets us passing a few more tests actually:

| `<iframe>` loading tests in Safari | `<iframe>` loading tests in MiniBrowser (with changes from this patch) |
|--------|--------|
| image | image |

Source/WebCore/html/LazyLoadElementObserver.cpp
credod: That could certainly clean things up here, but it kind of feels like overkill for `Element` to be aware of this, since it only impacts a few specific elements for now. However, I'm happy to make this change if we think this is a better approach!

Source/WebCore/Modules/model-element/HTMLModelElement.cpp
```diff
@@ -1802,6 +1802,11 @@ void HTMLModelElement::viewportIntersectionChanged(bool isIntersecting)
     visibilityStateChanged();
 }
 
+void HTMLModelElement::lazyLoadIntersectionCallbackInvoked(bool isIntersecting)
+{
+    viewportIntersectionChanged(isIntersecting);
```
weinig: I think you can merge `viewportIntersectionChanged()` into this function since this is the only caller.

weinig (approved):

Source/WebCore/html/shadow/ProgressShadowElement.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+Ref<ProgressValueElement> ProgressValueElement::create(Document& document)
+{
+    Ref result = adoptRef(*new ProgressValueElement(document));
+    ScriptDisallowedScope::EventAllowedScope resultScope { result };
+    result->setUserAgentPart(UserAgentParts::webkitProgressValue());
+    return result;
+}
+
+}
```
annevk: I think rebasing went wrong here. This file shouldn't have been added?

Source/WebCore/html/LazyLoadElementObserver.cpp
```diff
@@ -74,25 +83,26 @@ class LazyImageLoadIntersectionObserverCallback final : public IntersectionObser
     }
 };
 
-void LazyLoadImageObserver::observe(Element& element)
+void LazyLoadElementObserver::observe(Element& element)
 {
-    auto& observer = element.document().lazyLoadImageObserver();
-    RefPtr intersectionObserver = observer.intersectionObserver(protect(element.document()));
+    Ref protectedDocument = element.document();
```
annevk:
```suggestion
    Ref document = element.document();
```
We don't use the "protected" prefix except for "protectedThis" essentially.

Source/WebCore/html/LazyLoadElementObserver.cpp
```diff
@@ -103,9 +113,10 @@ IntersectionObserver* LazyLoadImageObserver::intersectionObserver(Document& docu
     return m_observer.get();
 }
 
-bool LazyLoadImageObserver::isObserved(Element& element) const
+bool LazyLoadElementObserver::isObserved(Element& element) const
 {
-    return m_observer && m_observer->isObserving(element);
+    Ref protectedElement = element;
+    return m_observer && m_observer->isObserving(protectedElement);
```
annevk:
```suggestion
    return m_observer && m_observer->isObserving(protect(element));
```

Source/WebCore/html/shadow/ProgressShadowElement.cpp
credod: good catch! fixed!

PR #59821, opened 2026-03-03
[Dictation] Add DictationStreamingOpacity document marker for streaming partial results

ishaanbhasin04:
#### e7fe98f3c9cee9a91f54e086bf2f94172cf737eb
<pre>
[Dictation] Add DictationStreamingOpacity document marker for streaming partial results
<a href="https://bugs.webkit.org/show_bug.cgi?id=309109">https://bugs.webkit.org/show_bug.cgi?id=309109</a>
<a href="https://rdar.apple.com/169476197">rdar://169476197</a>

Reviewed by NOBODY (OOPS!).

Add a new DocumentMarkerType::DictationStreamingOpacity that renders text at a
variable alpha. This is used to show unconfirmed speech recognition text at
reduced opacity while the recognizer is still processing.

The marker carries an opacity value and a UUID identifier for targeted removal.
The rendering pipeline reads the opacity from the marker data and applies it via
context.setAlpha(), following the same pattern as TransparentContent and
DraggedContent markers.

IPC messages (SetDictationStreamingOpacity, ClearDictationStreamingOpacity) are
in the PLATFORM(COCOA) block. The WebProcess handler locates the hypothesis text
by walking backward from the cursor and applies the marker to the streaming
sub-range. WKContentView SPI entry points are added for the UIProcess.

The cursor-walking logic is extracted into findDictatedTextRangeBeforeCursor,
shared with replaceDictatedText.

* Source/WebCore/dom/DocumentMarker.h:
* Source/WebCore/dom/DocumentMarkerController.cpp:
(WebCore::DocumentMarkerController::addDictationStreamingOpacityMarker):
(WebCore::DocumentMarkerController::removeDictationStreamingOpacityMarkers):
(WebCore::DocumentMarkerController::removeAllDictationStreamingOpacityMarkers):
* Source/WebCore/dom/DocumentMarkerController.h:
* Source/WebCore/rendering/MarkedText.cpp:
(WebCore::MarkedText::collectForDictationStreamingOpacity):
* Source/WebCore/rendering/MarkedText.h:
* Source/WebCore/rendering/StyledMarkedText.cpp:
(WebCore::resolveStyleForMarkedText):
* Source/WebCore/rendering/TextBoxPainter.cpp:
(WebCore::TextBoxPainter::paintForegroundAndDecorations):
(WebCore::TextBoxPainter::paintForeground):
(WebCore::TextBoxPainter::createDecorationPainter):
(WebCore::TextBoxPainter::paintPlatformDocumentMarkers):
* Source/WebKit/UIProcess/Cocoa/WebPageProxyCocoa.mm:
(WebKit::WebPageProxy::setDictationStreamingOpacity):
(WebKit::WebPageProxy::clearDictationStreamingOpacity):
* Source/WebKit/UIProcess/ios/WKContentViewInteraction.mm:
(-[WKContentView _setDictationStreamingOpacity:forHypothesisText:streamingRange:identifier:]):
(-[WKContentView _clearDictationStreamingOpacity]):
(WebKit::WebPage::findDictatedTextRangeBeforeCursor):
(WebKit::WebPage::setDictationStreamingOpacity):
(WebKit::WebPage::clearDictationStreamingOpacity):
* Source/WebKit/WebProcess/WebPage/ios/WebPageIOS.mm:
(WebKit::WebPage::replaceDictatedText):
* Tools/TestWebKitAPI/Tests/WebCore/MarkedText.cpp:
(WebCore::operator&lt;&lt;):
</pre>
https://github.com/WebKit/WebKit/commit/e7fe98f3c9cee9a91f54e086bf2f94172cf737eb

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/147950 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/20635 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/14231 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/156630 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/101363 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/149823 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/21092 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/20539 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/114059 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/81337 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/150910 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/16305 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/132886 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/94822 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/15451 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/13247 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/4070 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/125057 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/10775 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/158965 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/2100 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/12276 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/122092 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/20434 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/17177 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/122295 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/20442 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/132583 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/76585 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/17777 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/9343 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/20051 "Built successfully") | [❌ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/83810 "Found 8 new failures in WebProcess/WebPage/Cocoa/WebPageCocoa.mm and found 1 fixed file: rendering/MarkedText.h") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/19780 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/19928 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/19837 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/dom/DocumentMarkerController.h
+++ b/Source/WebCore/dom/DocumentMarkerController.h
@@ -68,6 +68,9 @@ public:
     WEBCORE_EXPORT bool addMarker(Node&, DocumentMarker&&);
     void addDraggedContentMarker(const SimpleRange&);
     WEBCORE_EXPORT void addTransparentContentMarker(const SimpleRange&, WTF::UUID);
+    WEBCORE_EXPORT void addDictationStreamingOpacityMarker(const SimpleRange&, float opacity, WTF::UUID);
+    WEBCORE_EXPORT void removeDictationStreamingOpacityMarkers(const WTF::UUID&);
+    WEBCORE_EXPORT void removeAllDictationStreamingOpacityMarkers();
 
     void copyMarkers(Node& source, OffsetRange, Node& destination);
     bool hasMarkers() const;
--- a/Source/WebCore/rendering/MarkedText.cpp
+++ b/Source/WebCore/rendering/MarkedText.cpp
@@ -353,4 +353,27 @@ Vector<MarkedText> MarkedText::collectForDraggedAndTransparentContent(const Docu
     });
 }
 
+Vector<MarkedText> MarkedText::collectForDictationStreamingOpacity(const RenderText& renderer, const TextBoxSelectableRange& selectableRange)
+{
+    if (!renderer.textNode())
+        return { };
+
+    CheckedPtr markerController = renderer.document().markersIfExists();
+    if (!markerController)
+        return { };
+
+    auto markers = markerController->markersFor(*renderer.textNode(), DocumentMarkerType::DictationStreamingOpacity);
+    if (markers.isEmpty())
+        return { };
+
+    Vector<MarkedText> result;
+    result.reserveInitialCapacity(markers.size());
+    for (auto& marker : markers) {
+        auto [clampedStart, clampedEnd] = selectableRange.clamp(marker->startOffset(), marker->endOffset());
+        if (clampedStart < clampedEnd)
+            result.append({ clampedStart, clampedEnd, MarkedText::Type::DictationStreamingOpacity, marker.get() });
+    }
+    return result;
+}
+
 }
--- a/Source/WebCore/rendering/TextBoxPainter.cpp
+++ b/Source/WebCore/rendering/TextBoxPainter.cpp
@@ -438,6 +438,9 @@ void TextBoxPainter::paintForegroundAndDecorations()
             auto markedTextsForTransparentContent = MarkedText::collectForDraggedAndTransparentContent(DocumentMarkerType::TransparentContent, m_renderer, m_selectableRange);
             if (!markedTextsForTransparentContent.isEmpty())
                 markedTexts.appendVector(WTF::move(markedTextsForTransparentContent));
+            auto markedTextsForDictationStreamingOpacity = MarkedText::collectForDictationStreamingOpacity(m_renderer, m_selectableRange);
+            if (!markedTextsForDictationStreamingOpacity.isEmpty())
+                markedTexts.appendVector(WTF::move(markedTextsForDictationStreamingOpacity));
         }
     }
     // The selection marked text acts as a placeholder when computing the marked texts for the gaps...
@@ -654,7 +657,8 @@ void TextBoxPainter::paintForeground(const StyledMarkedText& markedText)
         m_isCombinedText ? &downcast<RenderCombineText>(m_renderer.get()) : nullptr
     };
 
-    bool isTransparentMarkedText = markedText.type == MarkedText::Type::DraggedContent || markedText.type == MarkedText::Type::TransparentContent;
+    bool isTransparentMarkedText = markedText.type == MarkedText::Type::DraggedContent || markedText.type == MarkedText::Type::TransparentContent
+        || markedText.type == MarkedText::Type::DictationStreamingOpacity;
     GraphicsContextStateSaver stateSaver(context, markedText.style.textStyles.strokeWidth > 0 || isTransparentMarkedText);
     if (isTransparentMarkedText)
         context.setAlpha(markedText.style.alpha);
@@ -716,7 +720,8 @@ TextDecorationPainter TextBoxPainter::createDecorationPainter(const StyledMarked
     // Note that if the text is truncated, we let the thing being painted in the truncation
     // draw its own decoration.
     GraphicsContextStateSaver stateSaver { context, false };
-    bool isTransparentContent = markedText.type == MarkedText::Type::DraggedContent || markedText.type == MarkedText::Type::TransparentContent;
+    bool isTransparentContent = markedText.type == MarkedText::Type::DraggedContent || markedText.type == MarkedText::Type::TransparentContent
+        || markedText.type == MarkedText::Type::DictationStreamingOpacity;
     if (isTransparentContent || !clipOutRect.isEmpty()) {
         stateSaver.save();
         if (isTransparentContent)
@@ -1240,6 +1245,8 @@ void TextBoxPainter::paintPlatformDocumentMarkers()
     // the other marked texts when being subdivided so that they do not get painted.
     Vector<MarkedText> allMarkedTexts;
     allMarkedTexts.appendVector(transparentContentMarkedTexts);
+    auto dictationStreamingOpacityMarkedTexts = MarkedText::collectForDictationStreamingOpacity(m_renderer, m_selectableRange);
+    allMarkedTexts.appendVector(dictationStreamingOpacityMarkedTexts);
     allMarkedTexts.appendVector(markedTexts);
     if (textDecorationLineSpellingErrorAsMarkedText)
         allMarkedTexts.append(*textDecorationLineSpellingErrorAsMarkedText);
@@ -1250,6 +1257,7 @@ void TextBoxPainter::paintPlatformDocumentMarkers()
         switch (markedText.type) {
         case MarkedText::Type::DraggedContent:
         case MarkedText::Type::TransparentContent:
+        case MarkedText::Type::DictationStreamingOpacity:
             continue;
 
         default:
--- a/Source/WebKit/UIProcess/WebPageProxy.h
+++ b/Source/WebKit/UIProcess/WebPageProxy.h
@@ -1383,6 +1383,8 @@ public:
     void addDictationAlternative(WebCore::TextAlternativeWithRange&&);
     void dictationAlternativesAtSelection(CompletionHandler<void(Vector<WebCore::DictationContext>&&)>&&);
     void clearDictationAlternatives(Vector<WebCore::DictationContext>&&);
+    void setDictationStreamingOpacity(const String& hypothesisText, const WebCore::CharacterRange& streamingRangeInHypothesis, float opacity, const WTF::UUID&);
+    void clearDictationStreamingOpacity();
 
     void hasMarkedText(CompletionHandler<void(bool)>&&);
     void getMarkedRangeAsync(CompletionHandler<void(const EditingRange&)>&&);
--- a/Source/WebKit/UIProcess/ios/WKContentViewInteraction.mm
+++ b/Source/WebKit/UIProcess/ios/WKContentViewInteraction.mm
@@ -116,6 +116,7 @@
 #import <MobileCoreServices/UTCoreTypes.h>
 #import <UniformTypeIdentifiers/UTCoreTypes.h>
 #import <WebCore/AppHighlight.h>
+#import <WebCore/CharacterRange.h>
 #import <WebCore/ColorCocoa.h>
 #import <WebCore/ColorSerialization.h>
 #import <WebCore/CompositionHighlight.h>
@@ -6000,6 +6001,27 @@ static void logTextInteraction(const char* methodName, UIGestureRecognizer *loup
 
     _autocorrectionContextNeedsUpdate = YES;
     protect(_page)->replaceDictatedText(oldText, newText);
+}
+
+- (void)_setDictationStreamingOpacity:(CGFloat)opacity forHypothesisText:(NSString *)hypothesisText streamingRange:(NSRange)streamingRange identifier:(NSUUID *)identifier
+{
+    if (!_page || !identifier || !hypothesisText)
+        return;
+
+    auto uuid = WTF::UUID::fromNSUUID(identifier);
+    if (!uuid)
+        return;
+
+    WebCore::CharacterRange characterRange { streamingRange.location, streamingRange.length };
+    protect(_page)->setDictationStreamingOpacity(hypothesisText, characterRange, static_cast<float>(opacity), *uuid);
+}
+
+- (void)_clearDictationStreamingOpacity
+{
+    if (!_page)
+        return;
+
+    protect(_page)->clearDictationStreamingOpacity();
 }
 
 // The completion handler should pass the rect of the correction text after replacing the input text, or nil if the replacement could not be performed.
--- a/Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
+++ b/Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
@@ -139,6 +139,7 @@
 #import <WebCore/UTIRegistry.h>
 #import <WebCore/UTIUtilities.h>
 #import <WebCore/UserTypingGestureIndicator.h>
+#import <WebCore/VisibleUnits.h>
 #import <WebCore/WebAccessibilityObjectWrapperMac.h>
 #import <WebCore/markup.h>
 #import <pal/spi/cocoa/LaunchServicesSPI.h>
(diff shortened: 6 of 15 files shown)


Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
```diff
@@ -545,6 +546,54 @@ String plainTextForDisplay(const std::optional<SimpleRange>& range)
...
+        return;
+    }
+
+    // Find the hypothesis text by walking backward from the cursor, matching how
+    // replaceDictatedText locates text in the document.
+    VisiblePosition position = frame->selection().selection().start();
+    for (auto i = numGraphemeClusters(hypothesisText); i; --i)
+        position = position.previous();
+    if (position.isNull())
+        position = startOfDocument(protect(frame->document()));
```
whsieh: This looks like it could be merged with the existing logic in `replaceDictatedText`, into a new helper method.

Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
```diff
@@ -545,6 +546,54 @@ String plainTextForDisplay(const std::optional<SimpleRange>& range)
...
+    RefPtr frame = m_page->focusController().focusedOrMainFrame();
+    if (!frame) {
+        WEBPAGE_RELEASE_LOG(ViewState, "setDictationStreamingOpacity - no focused frame");
+        return;
+    }
+
+    if (frame->selection().isNone()) {
+        WEBPAGE_RELEASE_LOG(ViewState, "setDictationStreamingOpacity - selection is none");
+        return;
+    }
```
whsieh, line 574: Presumably we probably also want to return early, in the case where the selection isn't editable? (IIUC this is only for editable text, right?)

whsieh (commented):
Nice work — this is a solid start! I think it should be possible to add an API or layout test for this as well.

Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
ishaanbhasin04: Updated.

Source/WebKit/WebProcess/WebPage/ios/WebPageIOS.mm
```diff
@@ -2208,24 +2208,19 @@ static bool insideImageOverlay(const VisiblePosition& position)
...
     }
-    VisiblePosition position = frame->selection().selection().start();
-    for (auto i = numGraphemeClusters(oldText); i; --i)
-        position = position.previous();
-    if (position.isNull())
-        position = startOfDocument(protect(frame->document()));
-    auto range = makeSimpleRange(position, frame->selection().selection().start());
 
-    if (plainTextForContext(range) != oldText)
+    auto range = findDictatedTextRangeBeforeCursor(*frame, oldText);
```
ishaanbhasin04, line 2217: Using common logic in findDictatedTextRangeBeforeCursor now.

Source/WebCore/dom/DocumentMarkerController.h
```diff
@@ -68,6 +68,9 @@ class DocumentMarkerController final : public CanMakeCheckedPtr<DocumentMarkerCo
     WEBCORE_EXPORT bool addMarker(Node&, DocumentMarker&&);
     void addDraggedContentMarker(const SimpleRange&);
     WEBCORE_EXPORT void addTransparentContentMarker(const SimpleRange&, WTF::UUID);
+    WEBCORE_EXPORT void addDictationStreamingOpacityMarker(const SimpleRange&, float opacity, WTF::UUID);
+    WEBCORE_EXPORT void removeDictationStreamingOpacityMarkers(const WTF::UUID&);
```
darinadler, line 72: No reason to take `WTF::UUID` for `add` but `const WTF::UUID&` for `remove`. We should choose one and be consistent. The code already isn’t consistent. There’s no good reason to take `const SimpleRange&` and `WTF::UUID` without using `const&`. In both cases, the `const&` reduces reference count churn a bit, but doesn’t do have additional compelling value.

Source/WebCore/rendering/TextBoxPainter.cpp
```diff
@@ -438,6 +438,9 @@ void TextBoxPainter::paintForegroundAndDecorations()
             auto markedTextsForTransparentContent = MarkedText::collectForDraggedAndTransparentContent(DocumentMarkerType::TransparentContent, m_renderer, m_selectableRange);
             if (!markedTextsForTransparentContent.isEmpty())
                 markedTexts.appendVector(WTF::move(markedTextsForTransparentContent));
+            auto markedTextsForDictationStreamingOpacity = MarkedText::collectForDictationStreamingOpacity(m_renderer, m_selectableRange);
+            if (!markedTextsForDictationStreamingOpacity.isEmpty())
+                markedTexts.appendVector(WTF::move(markedTextsForDictationStreamingOpacity));
```
darinadler, line 443: I think we should use a more terse style; the local variable doesn’t need a long name, but also it isn’t needed since `appendVector` already does nothing if passed an empty vector, and don’t need to write WTF::move if there isn’t a local variable:
```suggestion
            markedTexts.appendVector(MarkedText::collectForDictationStreamingOpacity(m_renderer, m_selectableRange));
```
Could also change the code just above in the same way.

Source/WebCore/rendering/TextBoxPainter.cpp
```diff
@@ -654,7 +657,8 @@ void TextBoxPainter::paintForeground(const StyledMarkedText& markedText)
         m_isCombinedText ? &downcast<RenderCombineText>(m_renderer.get()) : nullptr
     };
 
-    bool isTransparentMarkedText = markedText.type == MarkedText::Type::DraggedContent || markedText.type == MarkedText::Type::TransparentContent;
+    bool isTransparentMarkedText = markedText.type == MarkedText::Type::DraggedContent || markedText.type == MarkedText::Type::TransparentContent
+        || markedText.type == MarkedText::Type::DictationStreamingOpacity;
```
darinadler, line 661: As this list of types gets longer it might be worthwhile to use a helper function rather than having this list repeated.

Source/WebCore/rendering/TextBoxPainter.cpp
```diff
@@ -1240,6 +1245,8 @@ void TextBoxPainter::paintPlatformDocumentMarkers()
     // the other marked texts when being subdivided so that they do not get painted.
     Vector<MarkedText> allMarkedTexts;
     allMarkedTexts.appendVector(transparentContentMarkedTexts);
+    auto dictationStreamingOpacityMarkedTexts = MarkedText::collectForDictationStreamingOpacity(m_renderer, m_selectableRange);
+    allMarkedTexts.appendVector(dictationStreamingOpacityMarkedTexts);
```
darinadler, line 1249: This would read better without a local variable:
```suggestion
    allMarkedTexts.appendVector(MarkedText::collectForDictationStreamingOpacity(m_renderer, m_selectableRange));
```

Source/WebKit/UIProcess/ios/WKContentViewInteraction.mm
```diff
@@ -6002,6 +6003,27 @@ - (void)replaceDictatedText:(NSString*)oldText withText:(NSString *)newText
     protect(_page)->replaceDictatedText(oldText, newText);
 }
 
+- (void)_setDictationStreamingOpacity:(CGFloat)opacity forHypothesisText:(NSString *)hypothesisText streamingRange:(NSRange)streamingRange identifier:(NSUUID *)identifier
+{
+    if (!_page || !identifier || !hypothesisText)
+        return;
+
+    auto uuid = WTF::UUID::fromNSUUID(identifier);
```
darinadler, line 6011: I would call this `parsedIdentifier` or `convertedIdentifier` or `coreIdentifier`.
```suggestion
    auto parsedIdentifier = WTF::UUID::fromNSUUID(identifier);
```

Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
```diff
@@ -545,6 +546,58 @@ String plainTextForDisplay(const std::optional<SimpleRange>& range)
...
+
+    auto streamingRange = resolveCharacterRange(*hypothesisRange, streamingRangeInHypothesis);
+    if (streamingRange.collapsed()) {
+        WEBPAGE_RELEASE_LOG(ViewState, "setDictationStreamingOpacity - resolved streaming range is collapsed");
+        return;
+    }
+
+    protect(frame->document()->markers())->removeDictationStreamingOpacityMarkers(identifier);
+    protect(frame->document()->markers())->addDictationStreamingOpacityMarker(streamingRange, opacity, identifier);
```
darinadler, line 589: Slightly better to use a local variable, I think?
```suggestion
    Ref markers = frame->document()->markers();
    markers->removeDictationStreamingOpacityMarkers(identifier);
    markers->addDictationStreamingOpacityMarker(streamingRange, opacity, identifier);
```

Source/WebKit/WebProcess/WebPage/WebPage.h
```diff
@@ -1345,6 +1345,9 @@ class WebPage final : public API::ObjectImpl<API::Object::Type::BundlePage>, pub
     void addDictationAlternative(const String& text, WebCore::DictationContext, CompletionHandler<void(bool)>&&);
     void dictationAlternativesAtSelection(CompletionHandler<void(Vector<WebCore::DictationContext>&&)>&&);
     void clearDictationAlternatives(Vector<WebCore::DictationContext>&&);
+    std::optional<WebCore::SimpleRange> findDictatedTextRangeBeforeCursor(WebCore::LocalFrame&, const String&);
```
darinadler, line 1348: I suggest we make this function private, rather than mixing it in with the public functions.

Source/WebKit/WebProcess/WebPage/WebPage.h
```diff
@@ -1345,6 +1345,9 @@ class WebPage final : public API::ObjectImpl<API::Object::Type::BundlePage>, pub
...
+    void setDictationStreamingOpacity(const String& hypothesisText, const WebCore::CharacterRange& streamingRangeInHypothesis, float opacity, const WTF::UUID&);
```
darinadler, line 1349: This could take a `WebCore::CharacterRange` rather than a `const WebCore::CharacterRange&` since it’s a simple structure made entirely out of scalars.

Source/WebKit/UIProcess/WebPageProxy.h
```diff
@@ -1383,6 +1383,8 @@ class WebPageProxy final : public API::ObjectImpl<API::Object::Type::Page>, publ
     void addDictationAlternative(WebCore::TextAlternativeWithRange&&);
     void dictationAlternativesAtSelection(CompletionHandler<void(Vector<WebCore::DictationContext>&&)>&&);
     void clearDictationAlternatives(Vector<WebCore::DictationContext>&&);
+    void setDictationStreamingOpacity(const String& hypothesisText, const WebCore::CharacterRange& streamingRangeInHypothesis, float opacity, const WTF::UUID&);
```
darinadler, line 1386: This could take a `StringView` rather than a `const String&` and it should. This should take a `WebCore::CharacterRange` rather than a `const WebCore::CharacterRange&` since it’s a simple structure made entirely out of scalars.

Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
```diff
@@ -545,6 +546,58 @@ String plainTextForDisplay(const std::optional<SimpleRange>& range)
     }, DocumentMarkerType::DictationAlternatives);
 }
 
+std::optional<SimpleRange> WebPage::findDictatedTextRangeBeforeCursor(LocalFrame& frame, const String& text)
+{
+    VisiblePosition position = frame.selection().selection().start();
+    for (auto i = numGraphemeClusters(text); i; --i)
+        position = position.previous();
```
darinadler, line 553: For the code that we moved here (it’s not new, but used more than before) we should confirm that `VisiblePosition::previous` moves one grapheme cluster at a time, ideally by including test cases that involve combining characters and non-BMP characters to prove we have that right.

Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
```diff
@@ -545,6 +546,58 @@ String plainTextForDisplay(const std::optional<SimpleRange>& range)
...
+
+void WebPage::setDictationStreamingOpacity(const String& hypothesisText, const WebCore::CharacterRange& streamingRangeInHypothesis, float opacity, const WTF::UUID& identifier)
+{
+    RefPtr frame = m_page->focusController().focusedOrMainFrame();
+    if (!frame) {
+        WEBPAGE_RELEASE_LOG(ViewState, "setDictationStreamingOpacity - no focused frame");
+        return;
+    }
+
+    if (frame->selection().isNone() || !frame->selection().selection().isContentEditable()) {
```
darinadler, line 571: We may need code to confirm that `streamingRange` is entirely within a content editable area; it might not be enough to check that we _start_ with a selection within a content editable area. Worth thinking about how to test that edge case.

Source/WebCore/rendering/MarkedText.cpp
```diff
@@ -353,4 +353,27 @@ Vector<MarkedText> MarkedText::collectForDraggedAndTransparentContent(const Docu
...
+    if (!renderer.textNode())
+        return { };
+
+    CheckedPtr markerController = renderer.document().markersIfExists();
+    if (!markerController)
+        return { };
+
+    auto markers = markerController->markersFor(*renderer.textNode(), DocumentMarkerType::DictationStreamingOpacity);
+    if (markers.isEmpty())
+        return { };
```
darinadler, line 367: I don’t think this check is needed. Pretty sure the code below will also efficiently return an empty vector.

darinadler (approved):
I have a variety of code style comments, but mostly this looks right to me.

PR #59947, opened 2026-03-05
Parse media controls UA stylesheets once per document using adoptedStyleSheets

NathanSolomon733:
#### 5c180c290003e2549b9ece2d9f32def732444b70
<pre>
Parse media controls UA stylesheets once per document using adoptedStyleSheets
<a href="https://bugs.webkit.org/show_bug.cgi?id=309214">https://bugs.webkit.org/show_bug.cgi?id=309214</a>
<a href="https://rdar.apple.com/171764828">rdar://171764828</a>

Reviewed by NOBODY (OOPS!).

Media controls stylesheets were injected as inline &lt;style&gt; elements in the UA shadow root, causing the
same CSS to be re-parsed for every media element. Parse them once per document in C++ and share the
result across all media elements via adoptedStyleSheets on the shadow root.

Parse in C++ with UASheetMode rather than constructing stylesheets from JS, which uses HTMLStandardMode.
In HTMLStandardMode, CSS values like display: ruby and display: ruby-text are gated behind
cssRubyDisplayTypesInAuthorStylesEnabled and silently dropped. There is no equivalent mechanism for
adopted stylesheets created from JS, so parsing in C++ allows the parser mode to be set explicitly.

Expose the pre-parsed sheets to JS via host.uaStyleSheets, and let a new StyleSheetSupport class in
JavaScript manage the shadowRoot.adoptedStyleSheets list. This keeps the logic (caching,
diffing, assembling the sheet list) in JS along with the rest of the media controls UI code, while C++
handles the UASheetMode parse and exposes the data to JS through IDL getters.

Caption user preference overrides and VTT track-level stylesheets are managed in JS using
constructed stylesheets (new CSSStyleSheet() + replaceSync()). These do not need UASheetMode because
they are author CSS. Adopted stylesheets take precedence over inline &lt;style&gt; elements in the cascade,
so caption and track sheets must also be adopted stylesheets to still be able to override the
UA rules. JS reacts to text track change events and a C++ notification (captionStyleSheetsDidChange)
when system caption preferences change.

Adopted stylesheets cannot match pseudo-elements in the shadow tree the way inline &lt;style&gt; elements
can, so change text track CSS selectors from pseudo-element selectors
(e.g. video::-webkit-media-text-track-container) to attribute selectors (e.g. [useragentpart=&quot;...&quot;]).

When running the VideoElementWithControlsCreation micro benchmark  before vs after this commit, the median
time improved from 193ms to 94ms per iteration (51% faster). Memory usage decreased from ~327MB to
~260MB per iteration (20% less memory).

No new tests because this performance change should not affect correctness.
VideoElementWithControlsCreation.html stresses this code change already.

Original PR credits and optimization idea go to Ryan Reno.

* LayoutTests/media/modern-media-controls/media-controller/media-controller-single-container-expected.txt:
* LayoutTests/media/modern-media-controls/media-controller/media-controller-single-container.html:
* Source/WebCore/CMakeLists.txt:
* Source/WebCore/DerivedSources-input.xcfilelist:
* Source/WebCore/DerivedSources.make:
* Source/WebCore/Modules/mediacontrols/MediaControlsHost.cpp:
(WebCore::MediaControlsHost::uaStyleSheets const):
(WebCore::MediaControlsHost::captionPreferencesStyleSheet const):
(WebCore::MediaControlsHost::showingTextTrackStyleSheets const):
* Source/WebCore/Modules/mediacontrols/MediaControlsHost.h:
* Source/WebCore/Modules/mediacontrols/MediaControlsHost.idl:
* Source/WebCore/Modules/modern-media-controls/controls/text-tracks.css:
([useragentpart=&quot;-webkit-media-text-track-container&quot;]):
(:host(video[controls]) [useragentpart=&quot;-webkit-media-text-track-container&quot;].visible-controls-bar):
(video::-webkit-media-text-track-container): Deleted.
(video[controls]::-webkit-media-text-track-container.visible-controls-bar): Deleted.
* Source/WebCore/Modules/modern-media-controls/js-files:
* Source/WebCore/Modules/modern-media-controls/main.js:
(createControls):
* Source/WebCore/Modules/modern-media-controls/media/media-controller.js:
(MediaController):
(MediaController.prototype.captionStyleSheetsDidChange):
* Source/WebCore/Modules/modern-media-controls/media/stylesheet-support.js: Added.
(StyleSheetSupport):
(StyleSheetSupport.prototype.get control):
(StyleSheetSupport.prototype.get mediaEvents):
(StyleSheetSupport.prototype.get tracksToMonitor):
(StyleSheetSupport.prototype.handleEvent):
(StyleSheetSupport.prototype.captionStyleSheetsDidChange):
(StyleSheetSupport.prototype._syncAdoptedStyleSheets):
* Source/WebCore/css/CSSStyleSheet.cpp:
(WebCore::CSSStyleSheet::createFromContents):
* Source/WebCore/css/CSSStyleSheet.h:
* Source/WebCore/dom/DocumentMediaElement.cpp:
(WebCore::DocumentMediaElement::ensureMediaControlsStyleSheets):
* Source/WebCore/dom/DocumentMediaElement.h:
* Source/WebCore/html/HTMLMediaElement.cpp:
(WebCore::HTMLMediaElement::updateTextTrackDisplay):
(WebCore::HTMLMediaElement::captionPreferencesChanged):
(WebCore::HTMLMediaElement::callControllerJSMethod):
(WebCore::HTMLMediaElement::ensureMediaControls):
* Source/WebCore/html/HTMLMediaElement.h:
* Source/WebCore/html/track/VTTCue.cpp:
(WebCore::VTTCue::obtainCSSBoxes):
</pre>
https://github.com/WebKit/WebKit/commit/5c180c290003e2549b9ece2d9f32def732444b70

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/161918 "Passed style check") | [loading 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/35392 "Build is in progress. Recent messages:OS: Tahoe (26.2), Xcode: 26.2; Skipping applying patch since patch_id isn't provided; Checked out pull request; Compiled WebKit (warnings)") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/28495 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/170821 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/118289 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/163787 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/35494 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/35394 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/125629 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/88531 "17 failures") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/164877 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/27955 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/145437 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/106225 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/27004 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/25518 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/18542 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/136697 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/23192 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/173310 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/20405 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/24833 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/133931 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/35066 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/29605 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/133936 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/35050 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/144987 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/97145 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/28469 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/21812 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/34560 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/102045 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/34061 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/34303 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/34209 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/html/HTMLMediaElement.cpp
+++ b/Source/WebCore/html/HTMLMediaElement.cpp
@@ -7940,8 +7940,10 @@ void HTMLMediaElement::configureTextTrackDisplay(TextTrackVisibilityCheckType ch
 
 void HTMLMediaElement::updateTextTrackDisplay()
 {
-    if (ensureMediaControls())
+    if (ensureMediaControls()) {
+        callControllerJSMethod("captionStyleSheetsDidChange"_s);
         m_mediaControlsHost->updateTextTrackContainer();
+    }
 }
 
 void HTMLMediaElement::updateTextTrackRepresentationImageIfNeeded()
@@ -8111,6 +8113,8 @@ void HTMLMediaElement::captionPreferencesChanged()
 
     if (RefPtr mediaControlsHost = m_mediaControlsHost.get())
         mediaControlsHost->captionPreferencesChanged();
+
+    callControllerJSMethod("captionStyleSheetsDidChange"_s);
 
     if (RefPtr player = m_player)
         player->tracksChanged();
@@ -8942,6 +8946,38 @@ void HTMLMediaElement::setControllerJSProperty(ASCIILiteral propertyName, JSC::J
     });
 }
 
+void HTMLMediaElement::callControllerJSMethod(ASCIILiteral methodName)
+{
+    DocumentMediaElement::from(protect(document())).setupAndCallMediaControlsJS([this, methodName](JSDOMGlobalObject& globalObject, JSC::JSGlobalObject& lexicalGlobalObject, ScriptController&, DOMWrapperWorld&) {
+        auto& vm = globalObject.vm();
+        auto scope = DECLARE_THROW_SCOPE(vm);
+        if (!m_mediaControlsHost)
+            return false;
+
+        auto controllerValue = m_mediaControlsHost->controllerWrapper().getValue();
+        RETURN_IF_EXCEPTION(scope, false);
+        auto* controllerObject = controllerValue.toObject(&lexicalGlobalObject);
+        RETURN_IF_EXCEPTION(scope, false);
+
+        auto functionValue = controllerObject->get(&lexicalGlobalObject, JSC::Identifier::fromString(vm, methodName));
+        if (scope.exception()) [[unlikely]]
+            return false;
+        if (functionValue.isUndefinedOrNull())
+            return false;
+
+        auto* function = functionValue.toObject(&lexicalGlobalObject);
+        RETURN_IF_EXCEPTION(scope, false);
+
+        auto callData = JSC::getCallData(function);
+        if (callData.type == JSC::CallData::Type::None)
+            return false;
+
+        JSC::call(&lexicalGlobalObject, function, callData, controllerObject, JSC::MarkedArgumentBuffer());
+        RETURN_IF_EXCEPTION(scope, false);
+        return true;
+    });
+}
+
 bool HTMLMediaElement::ensureMediaControls()
 {
     if (m_controlsState == ControlsState::Ready)
@@ -8981,8 +9017,10 @@ bool HTMLMediaElement::ensureMediaControls()
             auto mediaJSWrapper = toJS(&lexicalGlobalObject, &globalObject, *this);
             auto mediaControlsHostJSWrapper = toJS(&lexicalGlobalObject, &globalObject, *m_mediaControlsHost);
 
+            Ref shadowRoot = ensureUserAgentShadowRoot();
+
             JSC::MarkedArgumentBuffer argList;
-            argList.append(toJS(&lexicalGlobalObject, &globalObject, Ref { ensureUserAgentShadowRoot() }));
+            argList.append(toJS(&lexicalGlobalObject, &globalObject, shadowRoot));
             argList.append(mediaJSWrapper);
             argList.append(mediaControlsHostJSWrapper);
             ASSERT(!argList.hasOverflowed());
--- a/LayoutTests/media/modern-media-controls/media-controller/media-controller-single-container-expected.txt
+++ b/LayoutTests/media/modern-media-controls/media-controller/media-controller-single-container-expected.txt
@@ -3,8 +3,7 @@ Testing the MediaController has a single container for captions and media contro
 On success, you will see a series of "PASS" messages, followed by "TEST COMPLETE".
 
 
-PASS shadowRoot.childElementCount is >= 2
-PASS shadowRoot.firstElementChild.localName is "style"
+PASS shadowRoot.childElementCount is 1
 PASS shadowRoot.lastElementChild.localName is "div"
 PASS shadowRoot.lastElementChild.className is "media-controls-container"
 PASS shadowRoot.lastElementChild.childElementCount is 2
--- a/LayoutTests/media/modern-media-controls/media-controller/media-controller-single-container.html
+++ b/LayoutTests/media/modern-media-controls/media-controller/media-controller-single-container.html
@@ -9,8 +9,7 @@ description("Testing the <code>MediaController</code> has a single container for
 const media = document.querySelector("video");
 const shadowRoot = window.internals.shadowRoot(media);
 
-shouldBeGreaterThanOrEqual("shadowRoot.childElementCount", "2");
-shouldBeEqualToString("shadowRoot.firstElementChild.localName", "style");
+shouldBe("shadowRoot.childElementCount", "1");
 shouldBeEqualToString("shadowRoot.lastElementChild.localName", "div");
 shouldBeEqualToString("shadowRoot.lastElementChild.className", "media-controls-container");
 
--- a/Source/WebCore/CMakeLists.txt
+++ b/Source/WebCore/CMakeLists.txt
@@ -2638,6 +2638,7 @@ set(MODERN_MEDIA_CONTROLS_SCRIPTS
     "${WEBCORE_DIR}/Modules/modern-media-controls/controls/adwaita-fullscreen-media-controls.js"
     "${WEBCORE_DIR}/Modules/modern-media-controls/controls/adwaita-layout-traits.js"
     "${WEBCORE_DIR}/Modules/modern-media-controls/media/media-controller-support.js"
+    "${WEBCORE_DIR}/Modules/modern-media-controls/media/stylesheet-support.js"
     "${WEBCORE_DIR}/Modules/modern-media-controls/media/airplay-support.js"
     "${WEBCORE_DIR}/Modules/modern-media-controls/media/audio-support.js"
     "${WEBCORE_DIR}/Modules/modern-media-controls/media/close-support.js"
--- a/Source/WebCore/DerivedSources-input.xcfilelist
+++ b/Source/WebCore/DerivedSources-input.xcfilelist
@@ -772,6 +772,7 @@ $(PROJECT_DIR)/Modules/modern-media-controls/media/close-support.js
 $(PROJECT_DIR)/Modules/modern-media-controls/media/controls-visibility-support.js
 $(PROJECT_DIR)/Modules/modern-media-controls/media/fullscreen-support.js
 $(PROJECT_DIR)/Modules/modern-media-controls/media/media-controller-support.js
+$(PROJECT_DIR)/Modules/modern-media-controls/media/stylesheet-support.js
 $(PROJECT_DIR)/Modules/modern-media-controls/media/media-controller.js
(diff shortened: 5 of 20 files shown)


Source/WebCore/html/HTMLMediaElement.cpp
```diff
@@ -8910,12 +8910,24 @@ bool HTMLMediaElement::ensureMediaControls()
...
+            Ref shadowRoot = ensureUserAgentShadowRoot();
+
             JSC::MarkedArgumentBuffer argList;
-            argList.append(toJS(&lexicalGlobalObject, &globalObject, Ref { ensureUserAgentShadowRoot() }));
+            argList.append(toJS(&lexicalGlobalObject, &globalObject, shadowRoot));
             argList.append(mediaJSWrapper);
             argList.append(mediaControlsHostJSWrapper);
             ASSERT(!argList.hasOverflowed());
 
+            // Adopt the cached UA stylesheets on the shadow root before JS runs.
```
rniwa: But why?

Source/WebCore/html/HTMLMediaElement.cpp
```diff
@@ -8910,12 +8910,24 @@ bool HTMLMediaElement::ensureMediaControls()
...
+            // These were parsed with UASheetMode in DocumentMediaElement so that
+            // UA-gated values are not silently dropped.
```
rniwa: Why do we need to know about this? What effect does it have?

Source/WebCore/html/shadow/MediaControlTextTrackContainerElement.cpp
```diff
@@ -101,6 +102,81 @@ static bool compareCueIntervalForDisplay(const CueInterval& one, const CueInterv
...
+    bool isNewSheet = false;
+    if (!sheet) {
+        auto result = CSSStyleSheet::create(document, CSSStyleSheet::Init { });
+        if (result.hasException())
+            return false;
+        sheet = result.releaseReturnValue();
+        isNewSheet = true;
+    }
+    if (css != cachedCSS) {
+        sheet->replaceSync(String { css });
```
rniwa: This will parse CSS each time we call this.
Why can't we also cache the result when the contents are identical?

Source/WebCore/html/HTMLMediaElement.cpp
NathanSolomon733: We need to do this before JS runs because previously, the function in main.js `createControls() `created <style> elements and injected them into the shadow root. This PR causes the stylesheets to now be parsed once in C++ (in DocumentMediaElement) and reused. So, the C++ side has to adopt them onto the shadow root before calling createControls, because the JS no longer does it. If we do not do this here, there would be no styles from the UA style sheet on the shadow root

Source/WebCore/html/HTMLMediaElement.cpp
NathanSolomon733: We don't need to know about it here on second thought. 
This comment makes more sense inside `ensureMediaControlsStyleSheets()` where the `parserContext` is created and used.
I wanted to explain why I created a parserContext inside `ensureMediaControlsStyleSheets` with UASheetMode. If we parse in C++, we need to use UASheetMode so that UA styles with internal properties and values parse correctly. This was in response to a failing test case `media/track/webvtt-ruby.html.`

More broadly, this is why I put the parsing of adoptedStyleSheets in C++ instead of JS. If I created CSSStyleSheet() objects in JS, they'd parse in HTMLStandardMode. There is no JS API to specify UASheetMode. For example, I had tried this in main.js:
```
function createControls(shadowRoot, media, host)
  {                                                                                                                                                                                                              
      if (host) {
          if (!cachedUAStyleSheets) {                                                                                                                                                                            
              cachedUAStyleSheets = host.shadowRootStyleSheets.map(cssText => {
                  const styleSheet = new CSSStyleSheet();
                  styleSheet.replaceSync(cssText);
                  return styleSheet;
              });
          }
          shadowRoot.adoptedStyleSheets = cachedUAStyleSheets;
      }

      controller = new MediaController(shadowRoot, media, host);
      if (host)
          host.controller = controller;
  }
```
But it did not work for this reason above.

Source/WebCore/html/HTMLMediaElement.cpp
rniwa: Then a better comment will be something like
> Insert stylesheets before JS runs since createControls in main.js expects style to be there already.

Source/WebCore/html/HTMLMediaElement.cpp
rniwa: So then a better comment will be that
> Use UASheetMode in order to parse internal properties like XYZ properly.

Source/WebCore/html/shadow/MediaControlTextTrackContainerElement.cpp
NathanSolomon733: Am I not already caching the result? I cache the resulted CSSStyleSheet as member variables m_captionOverrideSheet and m_trackStyleSheet on MediaControlTextTrackContainerElement. I only modify them when `(css != cachedCSS)`, which is their stored String representations that are also member vars.
Or are you suggesting another caching that I'm not realizing?

rniwa: What I meant is that `collectActiveTrackStyleSheets` aggregates stylesheets into one giant string. So if the list of stylesheets changes, then we'd need to re-parse them even if many of them stayed identical.

NathanSolomon733: I see now. I'll implement this change.

PR #60035, opened 2026-03-06
Font-faces with different styles should not be mixed in a CSSSegmentedFontFace

brentfulgham:
#### d026972d8525840d8dc263f1b40941125d4efcd2
<pre>
Font-faces with different styles should not be mixed in a CSSSegmentedFontFace
<a href="https://bugs.webkit.org/show_bug.cgi?id=120499">https://bugs.webkit.org/show_bug.cgi?id=120499</a>
<a href="https://rdar.apple.com/32566994">rdar://32566994</a>

Reviewed by Vitor Roriz.

This patch is inspired by a Blink change: <a href="https://chromium.googlesource.com/chromium/blink/+/7823ceba809d5c05bffeeb9e732e92fe76fab7e9">https://chromium.googlesource.com/chromium/blink/+/7823ceba809d5c05bffeeb9e732e92fe76fab7e9</a>

Before this patch, when a glyph for the character to be rendered is not
in the matched @font-face, glyph from other @font-face (with different
style) in the family or system font with the same name were used as fallback.
This behavior does not follow the font matching algorithm
(<a href="http://www.w3.org/TR/css3-fonts/#font-matching-algorithm)">http://www.w3.org/TR/css3-fonts/#font-matching-algorithm)</a>:

&quot;6. If no matching face exists or the matched face does not contain a glyph
    for the character to be rendered, the next family name is selected and
    the previous three steps repeated. Glyphs from other faces in the family
    are not considered.&quot;

Previously, CSSFontFaceSet::fontFace() was sorting all candidate faces by distance and
adding all of them to the CSSSegmentedFontFace. This allowed faces with worse-matching
traits to supply glyphs for characters covered by their unicode-range, violating the spec.

The fix uses FontSelectionAlgorithm&apos;s filter to eliminate &quot;non-best&quot; candidates before
populating the segmented face. Only faces that tie at the best distance for all three
properties (width, style, weight) are eligible to supply glyphs.

This work also revealed that the test expectation for fast/text/font-weight-download-3.html
was wrong. The previous expected file showed &quot;Ahem.otf&quot; being downloaded, which was
incorrect because it had a font-weight of 900 (while the test restricts weight to 100).

A similar issue needed to be corrected it fast/text/font-weight-fallback.html. The test
references two fonts, one that matches the requested weight, the other having the
characters needed to render the content. Neither font should be used, according to the
spec, and it should fall back to a system font (which is what Firefox and Chrome do).
This test was revised to confirm this behavior works according to spec.

Because of GTK&apos;s installed fonts (at least on the bots) the test results had to be
rebaselined because the corrected font fallback logic resulted in a slightly different
output (Courier is not a viable choice since is not a good match to the required weight
700) so it resolves to a system font.

Test: imported/w3c/web-platform-tests/css/css-fonts/font-face-unicode-range-stretch-isolation.html

* LayoutTests/fast/text/font-weight-download-3-expected.txt: Rebaselined.
* LayoutTests/fast/text/font-weight-fallback-expected.html: Rebaselined.
* LayoutTests/fast/text/font-weight-fallback.html: Correct test.
* LayoutTests/imported/w3c/web-platform-tests/css/css-fonts/font-face-unicode-range-stretch-isolation-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-fonts/font-face-unicode-range-stretch-isolation-ref.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-fonts/font-face-unicode-range-stretch-isolation.html: Added.
* LayoutTests/platform/glib/fast/css/font-face-multiple-faces-expected.txt: Rebaselined.
* LayoutTests/platform/gtk/fast/css/font-face-multiple-faces-expected.png: Rebaselined.
* LayoutTests/platform/gtk/fast/text/font-weight-fallback-expected.html: Added.
* LayoutTests/platform/wpe/fast/text/font-weight-fallback-expected.html: Added.
* Source/WebCore/css/CSSFontFaceSet.cpp:
(WebCore::CSSFontFaceSet::fontFace):
* Source/WebCore/platform/graphics/FontSelectionAlgorithm.cpp:
(WebCore::FontSelectionAlgorithm::FontSelectionAlgorithm):
(WebCore::FontSelectionAlgorithm::ensureEliminatedCapabilities):
(WebCore::FontSelectionAlgorithm::eliminatedCapabilities):
(WebCore::FontSelectionAlgorithm::indexOfBestCapabilities):
* Source/WebCore/platform/graphics/FontSelectionAlgorithm.h:

Canonical link: <a href="https://commits.webkit.org/309121@main">https://commits.webkit.org/309121@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/c7921fb0089db7b5bd565c57481f09f7d53b5654

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/149455 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/22173 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/15751 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/158157 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/102887 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/7ca90673-cc15-47e4-a066-f0e299a189f8) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/151328 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/22624 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/22050 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/115265 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/81980 "Passed tests") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/619ef6ea-29ef-4ee2-bf87-8923c3e5578d) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/152415 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/17418 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/134119 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/96009 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/050bea9d-9397-467d-9fe9-8f041a2b623f) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/16514 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/14403 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/6000 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/126123 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/12052 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/160634 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/3628 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/13591 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/123300 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/21976 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/18445 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/123513 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/33576 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/21983 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/133844 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/78197 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/18717 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/10595 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/21583 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/85404 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/21314 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/21466 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/21371 "Built successfully") | | | |


LayoutTests/fast/text/font-face-unicode-range-stretch-isolation.html
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
+<html class="reftest-wait">
+<head>
+<meta charset="utf-8">
+<link rel="author" title="WebKit" href="https://webkit.org">
+<link rel="help" href="https://drafts.csswg.org/css-fonts-3/#font-style-matching">
+<link rel="match" href="font-face-unicode-range-stretch-isolation-ref.html">
```
vitorroriz, line 5: There is no `font-face-unicode-range-stretch-isolation-ref.html` file, do we need a match tag if we aren't exporting it to WPT?

Source/WebCore/css/CSSFontFaceSet.cpp
```diff
@@ -505,6 +505,18 @@ CSSSegmentedFontFace* CSSFontFaceSet::fontFace(FontSelectionRequest request, con
...
+        // Per CSS Fonts 5.2.6: Only faces with the best-matching width, style, and weight are eligible
+        // to supply glyphs, even via unicode-range.
+        auto eliminated = fontSelectionAlgorithm.eliminatedCapabilities();
+        RELEASE_ASSERT(eliminated.size() == candidateFontFaces.size());
+        size_t eliminatedIndex = 0;
+        candidateFontFaces.removeAllMatching([&](auto&) {
+            return eliminated[eliminatedIndex++];
+        });
+        if (candidateFontFaces.isEmpty())
+            return face;
```
vitorroriz, line 518: I was thinking, some lines above we already have logic to exclude candidates when appending face. Shouldn't we do the same for this case instead of appending everything and then having to filter them out, which also requires us to keep a list of eliminated candidates?

LayoutTests/fast/text/font-face-unicode-range-stretch-isolation.html
brentfulgham: Oh, that is a mistake. It should be `LayoutTests/fast/text/font-face-unicode-range-stretch-isolation-expected.html`. I'll fix that.

Source/WebCore/css/CSSFontFaceSet.cpp
brentfulgham: I'm not sure? We were previously building the list of eliminated candidates each time `FontCache::createFontPlatformData(...)` was called. We now cache that calculation and use it here to filter out non-matches.

Maybe we could include the `eliminatedCapabilities` list check while building the set of font capabilities above, but I'm not sure if that would be an improvement.

brentfulgham: I don't think we can merge these loops, unless I'm misunderstanding how it works.

`eliminatedCapabilities()` works by calling filterCapability on width, style, and then weight. Each step invokes bestValue(), which scans all candidates to find the minimum distance for that property among the
  survivors so far. Only then does it eliminate candidates whose distance is worse than that minimum.                                      
                  
So I don't think we can decide whether to eliminate a candidate until we have the full set, so we can be sure we are only eliminated the worst matches.

vitorroriz (approved):
Nice patch! Maybe we can make this a WPT as it is enforcing the spec

PR #60062, opened 2026-03-06
[WTF] Add a NotificationPoint API

aoikonomopoulos:
#### e1346ce7c2cd25ddaf5967842bf1c18cef111f0d
<pre>
[WTF] Add a NotificationPoint API
<a href="https://bugs.webkit.org/show_bug.cgi?id=309350">https://bugs.webkit.org/show_bug.cgi?id=309350</a>

Reviewed by NOBODY (OOPS!).

Add WTF::NotificationPoint that essentially abstracts the Darwin
notify_* API. On USE(GLIB) platforms, implement the same semantics in
terms of D-Bus.

Since on D-Bus we need an owned name and a path to an object, have
NotificationPoint take those, but join them together to form the same
key as currently used for Darwin. The idea is that the external API
shouldn&apos;t change at all there.
</pre>
https://github.com/WebKit/WebKit/commit/1a480936ef200ea57eee59228b8837d1d009e58a

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/160255 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/33730 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/26829 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/169157 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/114636 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/162124 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/33829 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/33728 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/124239 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/87794 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/163213 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/26483 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/143941 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/104838 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/25536 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/24032 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/16877 "Built successfully") | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/152311 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/135228 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/21704 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/171635 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/21092 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/17639 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/23344 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/132491 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/33401 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/28126 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/132517 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/33386 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/143499 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/91671 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/27135 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/20313 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/192654 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/32895 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/99308 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/49595 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/32393 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/32640 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/32543 "Built successfully") | | |

### DIFF
--- a/Source/JavaScriptCore/runtime/VM.cpp
+++ b/Source/JavaScriptCore/runtime/VM.cpp
@@ -145,6 +145,7 @@
 #include "WeakGCMapInlines.h"
 #include "WideningNumberPredictionFuzzerAgent.h"
 #include <wtf/CryptographicallyRandomNumber.h>
+#include <wtf/NotificationPoint.h>
 #include <wtf/ProcessID.h>
 #include <wtf/ReadWriteLock.h>
 #include <wtf/SimpleStats.h>
@@ -402,16 +403,13 @@ WTF_ALLOW_UNSAFE_BUFFER_USAGE_END
         pathOut.print("JSCProfile-", getCurrentProcessID(), "-", m_perBytecodeProfiler->databaseID(), ".json");
         static NeverDestroyed<CString> pathOutString = pathOut.toCString();
 
-#if PLATFORM(COCOA)
+#if HAVE(NOTIFICATIONPOINT)
         static std::once_flag registerFlag;
         std::call_once(registerFlag, [this]() {
             int pid = getpid();
-            const char* key = "com.apple.WebKit.bytecode.profiler";
             dataLogF("<BYTECODE.STAT><%d> Registering callback for dumping profiles, dumping to %s.\n", pid, pathOutString->data());
-            dataLogF("<BYTECODE.STAT><%d> Use `notifyutil -v -p %s` to dump statistics.\n", pid, key);
-
-            int token;
-            notify_register_dispatch(key, &token, mainDispatchQueueSingleton(), ^(int) {
+            auto result = NotificationPoint::create("bytecode.profiler"_s, [this]() {
+                int pid = getpid();
                 dataLogF("<BYTECODE.STAT><%d> Dumping\n", pid);
                 if (!m_perBytecodeProfiler->save(pathOutString->data()))
                     dataLogF("<BYTECODE.STAT><%d> Failed to dump to %s. Do you need to add a sandbox extension? ((allow file-write* (subpath \"/private/tmp/\")) in WebProcess.sb.in\n", pid, pathOutString->data());
@@ -419,6 +417,13 @@ WTF_ALLOW_UNSAFE_BUFFER_USAGE_END
                     dataLogF("<BYTECODE.STAT><%d> Dumped to %s\n", pid, pathOutString->data());
                 dataLogF("<BYTECODE.STAT><%d> Dumping finished\n", pid);
             });
+            if (result.has_value()) {
+                RefPtr<NotificationPoint> point = result.value();
+#if PLATFORM(COCOA)
+                dataLogF("<BYTECODE.STAT><%d> Use `notifyutil -v -p %s` to dump statistics.\n", pid, point->key().utf8().data());
+#endif
+                (void)point.leakRef();
+            }
         });
 #endif
 
--- a/Source/JavaScriptCore/wasm/WasmOpcodeCounter.cpp
+++ b/Source/JavaScriptCore/wasm/WasmOpcodeCounter.cpp
@@ -30,13 +30,9 @@
 #include <wtf/Atomics.h>
 #include <wtf/DataLog.h>
 #include <wtf/NeverDestroyed.h>
+#include <wtf/NotificationPoint.h>
 #include <wtf/Vector.h>
 
-#if PLATFORM(COCOA)
-#include <notify.h>
-#include <unistd.h>
-#include <wtf/darwin/DispatchExtras.h>
-#endif
 
 WTF_ALLOW_UNSAFE_BUFFER_USAGE_BEGIN
 
@@ -101,18 +97,22 @@ void WasmOpcodeCounter::dump()
 
 void WasmOpcodeCounter::registerDispatch()
 {
-#if PLATFORM(COCOA)
+#if HAVE(NOTIFICATIONPOINT)
     static std::once_flag registerFlag;
     std::call_once(registerFlag, [&]() {
         int pid = getpid();
-        const char* key = "com.apple.WebKit.wasm.op.stat";
         dataLogF("<WASM.OP.STAT><%d> Registering callback for wasm opcode statistics.\n", pid);
-        dataLogF("<WASM.OP.STAT><%d> Use `notifyutil -v -p %s` to dump statistics.\n", pid, key);
 
-        int token;
-        notify_register_dispatch(key, &token, mainDispatchQueueSingleton(), ^(int) {
+        auto result = NotificationPoint::create("wasm.op.stat"_s, []() {
             WasmOpcodeCounter::singleton().dump();
         });
+        if (result.has_value()) {
+            RefPtr<NotificationPoint> point = result.value();
+#if PLATFORM(COCOA)
+            dataLogF("<WASM.OP.STAT><%d> Use `notifyutil -v -p %s` to dump statistics.\n", pid, point->key().utf8().data());
+#endif
+            (void)point.leakRef();
+        }
     });
 #endif
 }
--- a/Source/WTF/WTF.xcodeproj/project.pbxproj
+++ b/Source/WTF/WTF.xcodeproj/project.pbxproj
@@ -173,6 +173,7 @@
 		5FAD3AE221B9636600BEE178 /* URLHelpers.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 5FAD3AE121B9636600BEE178 /* URLHelpers.cpp */; };
 		6311592628989A55006A9A12 /* LibraryPathDiagnostics.h in Headers */ = {isa = PBXBuildFile; fileRef = 6311592428989A55006A9A12 /* LibraryPathDiagnostics.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		6311592728989A55006A9A12 /* LibraryPathDiagnostics.mm in Sources */ = {isa = PBXBuildFile; fileRef = 6311592528989A55006A9A12 /* LibraryPathDiagnostics.mm */; };
+        6311592728989A8500680A12 /* NotificationPoint.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 6311592725289A8500680A24 /* NotificationPoint.cpp */; };
 		63F9BED42898D96400371416 /* CFPrivSPI.h in Headers */ = {isa = PBXBuildFile; fileRef = 63F9BED32898D96400371416 /* CFPrivSPI.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		70A993FE1AD7151300FA615B /* SymbolRegistry.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 70A993FC1AD7151300FA615B /* SymbolRegistry.cpp */; };
 		70ECA60D1B02426800449739 /* AtomStringImpl.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 70ECA60A1B02426800449739 /* AtomStringImpl.cpp */; };
@@ -345,6 +346,7 @@
 		DD3DC89327A4BF8E007E5B61 /* BackwardsGraph.h in Headers */ = {isa = PBXBuildFile; fileRef = DCEE22041CEB9869000C2396 /* BackwardsGraph.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		DD3DC89427A4BF8E007E5B61 /* MathExtras.h in Headers */ = {isa = PBXBuildFile; fileRef = A8A472C9151A825B004123FF /* MathExtras.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		DD3DC89527A4BF8E007E5B61 /* NotFound.h in Headers */ = {isa = PBXBuildFile; fileRef = 7CEAE5AC1EA6E10F00DB6890 /* NotFound.h */; settings = {ATTRIBUTES = (Private, ); }; };
+        DD3DC89528A4B48E007E5B61 /* NotificationPoint.h in Headers */ = {isa = PBXBuildFile; fileRef = 7CEAED2C1EA6E1AF00DB6890 /* NotificationPoint.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		DD3DC89627A4BF8E007E5B61 /* SinglyLinkedList.h in Headers */ = {isa = PBXBuildFile; fileRef = A8A4730B151A825B004123FF /* SinglyLinkedList.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		DD3DC89727A4BF8E007E5B61 /* Markable.h in Headers */ = {isa = PBXBuildFile; fileRef = 304CA4E41375437EBE931D03 /* Markable.h */; settings = {ATTRIBUTES = (Private, ); }; };
 		DD3DC89927A4BF8E007E5B61 /* ProcessID.h in Headers */ = {isa = PBXBuildFile; fileRef = 0FC4488216FE9FE100844BE9 /* ProcessID.h */; settings = {ATTRIBUTES = (Private, ); }; };
@@ -1456,6 +1458,7 @@
 		5FAD3AE121B9636600BEE178 /* URLHelpers.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = URLHelpers.cpp; sourceTree = "<group>"; };
 		6311592428989A55006A9A12 /* LibraryPathDiagnostics.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = LibraryPathDiagnostics.h; sourceTree = "<group>"; };
 		6311592528989A55006A9A12 /* LibraryPathDiagnostics.mm */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.objcpp; path = LibraryPathDiagnostics.mm; sourceTree = "<group>"; };
+        6311592725289A8500680A24 /* NotificationPoint.cpp */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.cpp; path = NotificationPoint.cpp; sourceTree = "<group>"; };
 		63F9BED32898D96400371416 /* CFPrivSPI.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = CFPrivSPI.h; sourceTree = "<group>"; };
 		70A993FC1AD7151300FA615B /* SymbolRegistry.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = SymbolRegistry.cpp; sourceTree = "<group>"; };
 		70A993FD1AD7151300FA615B /* SymbolRegistry.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = SymbolRegistry.h; sourceTree = "<group>"; };
@@ -1497,6 +1500,7 @@
 		7CD4C26F1E2C82B900929470 /* StringConcatenateNumbers.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = StringConcatenateNumbers.h; sourceTree = "<group>"; };
 		7CDD7FF9186D2A54007433CD /* IteratorRange.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = IteratorRange.h; sourceTree = "<group>"; };
 		7CEAE5AC1EA6E10F00DB6890 /* NotFound.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = NotFound.h; sourceTree = "<group>"; };
+        7CEAED2C1EA6E1AF00DB6890 /* NotificationPoint.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = NotificationPoint.h; sourceTree = "<group>"; };
 		7CEF23EB23D1257900F8B904 /* PlatformCallingConventions.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = PlatformCallingConventions.h; sourceTree = "<group>"; };
 		7CFAF3C923D0AF1500D21BDD /* PlatformUse.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = PlatformUse.h; sourceTree = "<group>"; };
 		7E29C33D15FFD79B00516D61 /* ObjCRuntimeExtras.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = ObjCRuntimeExtras.h; sourceTree = "<group>"; };
@@ -2221,6 +2225,7 @@
 				6311592428989A55006A9A12 /* LibraryPathDiagnostics.h */,
 				6311592528989A55006A9A12 /* LibraryPathDiagnostics.mm */,
 				46335BEE2E778B1300860375 /* NetworkOSObject.h */,
+                6311592725289A8500680A24 /* NotificationPoint.cpp */,
 				53FC70CE23FB950C005B1990 /* OSLogPrintStream.h */,
 				53FC70CF23FB950C005B1990 /* OSLogPrintStream.mm */,
 				449EEC7D2D93788B008E7C80 /* TypeCastsOSObject.h */,
@@ -2603,6 +2608,7 @@
 				E3E64F0B22813428001E55B4 /* Nonmovable.h */,
 				526AEC911F6B4E5C00695F5D /* NoTailCalls.h */,
 				7CEAE5AC1EA6E10F00DB6890 /* NotFound.h */,
+                7CEAED2C1EA6E1AF00DB6890 /* NotificationPoint.h */,
 				9B56058F2D9691ED00FCE33E /* NoVirtualDestructorBase.h */,
 				A8A472D5151A825B004123FF /* NumberOfCores.cpp */,
 				A8A472D6151A825B004123FF /* NumberOfCores.h */,
@@ -3759,6 +3765,7 @@
 				DD3DC96127A4BF8E007E5B61 /* NoTailCalls.h in Headers */,
 				DD3DC89527A4BF8E007E5B61 /* NotFound.h in Headers */,
 				4448265228D18CA600D916F9 /* NotificationCenterCF.h in Headers */,
+                DD3DC89528A4B48E007E5B61 /* NotificationPoint.h in Headers */,
 				9B5605902D9691ED00FCE33E /* NoVirtualDestructorBase.h in Headers */,
 				DDF306E327C08654006A526F /* NSLocaleSPI.h in Headers */,
 				DDF306E227C08654006A526F /* NSObjCRuntimeSPI.h in Headers */,
@@ -4562,6 +4569,7 @@
 				A8A473EC151A825B004123FF /* MetaAllocator.cpp in Sources */,
 				0F66B28C1DC97BAB004A1D3F /* MonotonicTime.cpp in Sources */,
 				51B07F542AB8797300E14EE9 /* NativePromise.cpp in Sources */,
+                6311592728989A8500680A12 /* NotificationPoint.cpp in Sources */,
 				5CC0EE8A2162BC2200A1A842 /* NSURLExtras.mm in Sources */,
 				A8A473F4151A825B004123FF /* NumberOfCores.cpp in Sources */,
 				465F34932CD824AF000963B1 /* ObjCRuntimeExtras.mm in Sources */,
--- a/Source/WTF/wtf/FastMalloc.cpp
+++ b/Source/WTF/wtf/FastMalloc.cpp
@@ -30,6 +30,7 @@
 #include <string.h>
 #include <wtf/Atomics.h>
 #include <wtf/CheckedArithmetic.h>
+#include <wtf/NotificationPoint.h>
 #include <wtf/PageBlock.h>
 
 #if OS(WINDOWS)
@@ -52,10 +53,6 @@
 #include <wtf/NeverDestroyed.h>
 #include <wtf/SetForScope.h>
 #include <wtf/StackShot.h>
-
-#if PLATFORM(COCOA)
-#include <notify.h>
-#endif
 
 #endif
 
@@ -207,10 +204,13 @@ MallocCallTracker& MallocCallTracker::singleton()
 
 MallocCallTracker::MallocCallTracker()
 {
-    int token;
-    notify_register_dispatch("com.apple.WebKit.dumpUntrackedMallocs", &token, mainDispatchQueueSingleton(), ^(int) {
+#if HAVE(NOTIFICATIONPOINT)
+    auto point = NotificationPoint::create("dumpUntrackedMallocs"_s, []() {
         MallocCallTracker::singleton().dumpStats();
     });
+    if (point.has_value())
+        (void)point.value().leakRef();
+#endif
 }
 
 void MallocCallTracker::recordMalloc(void* address, size_t allocationSize)
--- a/Source/WTF/wtf/NotificationPoint.h
+++ b/Source/WTF/wtf/NotificationPoint.h
@@ -1,0 +1,137 @@
+/*
+ *  Copyright (C) 2026 Igalia. S.L. . All rights reserved.
+ *
+ *  This library is free software; you can redistribute it and/or
+ *  modify it under the terms of the GNU Lesser General Public
+ *  License as published by the Free Software Foundation; either
+ *  version 2 of the License, or (at your option) any later version.
+ *
+ *  This library is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ *  Lesser General Public License for more details.
+ *
+ *  You should have received a copy of the GNU Lesser General Public
+ *  License along with this library; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ *
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <wtf/CanMakeWeakPtr.h>
+#include <wtf/Expected.h>
+#include <wtf/Function.h>
+#include <wtf/HashMap.h>
+#include <wtf/HashSet.h>
+#include <wtf/Platform.h>
+#include <wtf/RefPtr.h>
+#include <wtf/ThreadSafeRefCounted.h>
+#include <wtf/Vector.h>
+#include <wtf/WeakPtr.h>
+#include <wtf/text/MakeString.h>
+#include <wtf/text/WTFString.h>
+
+#if OS(DARWIN)
+#include <dispatch/dispatch.h>
+#include <notify.h>
+#elif USE(GLIB)
+#include <gio/gio.h>
+#include <glib.h>
+#endif
+
+#if PLATFORM(COCOA) || USE(GLIB)
+#define HAVE_NOTIFICATIONPOINT 1
+#endif
+
+namespace WTF {
+
+class OwnedNotificationPoint;
+
+class NotificationPoint : public ThreadSafeRefCounted<NotificationPoint>
+#if USE(GLIB)
+, public CanMakeWeakPtr<NotificationPoint>
+#endif
+{
+public:
+    enum class Error : int {
+        ConnectionError,
+        InvalidArgument,
+        PermissionError,
+        Other,
+    };
+    WTF_EXPORT_PRIVATE static Expected<RefPtr<NotificationPoint>, Error> create(ASCIILiteral providedPath, Function<void()>&& callback);
+    WTF_EXPORT_PRIVATE static Expected<RefPtr<NotificationPoint>, Error> createWithName(ASCIILiteral name, ASCIILiteral providedPath, Function<void()>&& callback = nullptr);
+    WTF_EXPORT_PRIVATE Expected<bool, Error> getAndClearNotificationsPosted();
+    bool isLive() const;
+    WTF_EXPORT_PRIVATE ~NotificationPoint();
+    WTF_EXPORT_PRIVATE Expected<uint64_t, Error> getState();
+    WTF_EXPORT_PRIVATE static Expected<uint64_t, Error> getState(ASCIILiteral providedName, String providedPath);
+    WTF_EXPORT_PRIVATE Expected<void, Error> setState(uint64_t);
+    WTF_EXPORT_PRIVATE static Expected<void, Error> setState(ASCIILiteral providedName, String providedPath, uint64_t state);
+    WTF_EXPORT_PRIVATE Expected<void, Error> notify();
+    WTF_EXPORT_PRIVATE static Expected<void, Error> notify(ASCIILiteral providedName, String providedPath);
+    static bool testWTFisEmpty();
+
+#if OS(DARWIN)
+    // Unfortunately, moving this def to the .cpp file causes `tap installapi`
+    // (in the XCode build) to fail, since it apparently uses an .mm file to
+    // import things, but dispatch_queue_t expands to NSObject<> there but a
+    // struct on the C++ side, so the decl doesn't match the definition..
+    static void testWTFsetDispatchQueue(dispatch_queue_t queue)
+    {
+        s_dispatchQueue = queue;
+    }
+    const String& key() const;
+#elif USE(GLIB)
+    void actionCallback();
+#endif // OS(DARWIN)
+
+private:
+#if OS(DARWIN)
+    static void dispatchCallback(const String&);
+    NotificationPoint(String key, int token);
+    int m_token;
+    const String m_key;
+    static std::optional<dispatch_queue_t> s_dispatchQueue;
+#elif USE(GLIB)
+    NotificationPoint(OwnedNotificationPoint*, Function<void()>&&);
+    bool m_notificationsPosted;
+    OwnedNotificationPoint* m_ownedNotificationPoint;
+    Function<void()> m_callback;
+#endif // OS(DARWIN)
+};
+
+#if USE(GLIB)
+class OwnedNotificationPoint {
+    friend class NotificationPoint;
+public:
+    OwnedNotificationPoint(String name, String path) WTF_REQUIRES_LOCK(ownedNotificationPointLock);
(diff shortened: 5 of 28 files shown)


Source/WTF/wtf/glib/NotificationPoint.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+NotificationPointStatus NotificationPointStatus::fromPlatformStatus(int platformStatus)
+{
+    return NotificationPointStatus(NotificationPointStatus::Other, platformStatus);
+}
+
+
+void NotificationPoint::onBusAcquired(GDBusConnection* connection, const char* name, void*)
+{
+    g_message("Acquired bus: %s", name);
```
aperezdc: Any reason to not use a WTF log channel? I think the `Process` one could be suitable here, then we can go with the usual `export WEBKIT_DEBUG='Process'` to selectively enable the logging:


```suggestion
    RELEASE_LOG_INFO(Process, "Acquired bus: %s", name);
```

...and similarly for the rest of logging statements. Also many of them look like debugging used during development of the feature, so I suppose a number of them will go away before landing. Right?

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,620 @@
...
+    static void destroy(RefPtr<NotificationPoint> np)
+    {
+        auto iter = s_pendingNotificationPoints.find(np->m_name);
+        if (iter != s_pendingNotificationPoints.end()) {
+            auto map = iter->value;
+            map.remove(np->m_path);
+            if (map.isEmpty()) {
+                s_pendingNotificationPoints.remove(np->m_name);
+                if (!s_notificationPoints.contains(np->m_name)) {
+                    // We don't want to owm this name any more and it
```
aperezdc:
```suggestion
                    // We don't want to own this name any more and it
```

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,620 @@
...
+    {
+        return NotificationPointStatus(Unset);
+    }
+    WTF_EXPORT_PRIVATE static NotificationPointStatus fromPlatformStatus(int);
+private:
+    enum GenericStatus m_status;
+    int m_platformStatus;
+};
+
+#if OS(DARWIN)
```
aperezdc, line 77: Would it be much trouble to split the implementation in files named `NotificationPointDarwin.cpp` and `NotificationPointDBus.cpp`, with the `NotificationPoint.h` header having the minimum possible amount of platform-specific definitions? That would follow the same pattern of other existing modules (e.g. `FileSystem.h` + `FileSystemGLib.c` + `FileSystemCF` + ...).

That would make it easier to have a “fallback” no-op implementation, which I would at least enable for `!OS(DARWIN) || !USE(GLIB) || OS(ANDROID)`... and also to add a `NotificationPointAndroid.cpp` implementation in the future (yes, I am thinking already that D-Bus is not particularly suitable there; and we can abuse system properties for notifications like the init system does)

Source/WTF/wtf/glib/NotificationPoint.cpp
aoikonomopoulos: Yes, I only expect a few of the output statements to remain in the version for landing.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Oops, thanks.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: I'll see what I can do, thanks for the feedback!

Source/JavaScriptCore/runtime/VM.cpp
```diff
@@ -414,16 +415,16 @@ WTF_ALLOW_UNSAFE_BUFFER_USAGE_END
...
 
-#if PLATFORM(COCOA)
+#if PLATFORM(COCOA) || USE(GLIB)
         static std::once_flag registerFlag;
         std::call_once(registerFlag, [this]() {
             int pid = getpid();
             const char* key = "com.apple.WebKit.bytecode.profiler";
             dataLogF("<BYTECODE.STAT><%d> Registering callback for dumping profiles, dumping to %s.\n", pid, pathOutString->data());
-            dataLogF("<BYTECODE.STAT><%d> Use `notifyutil -v -p %s` to dump statistics.\n", pid, key);
+            dataLogF("<BYTECODE.STAT><%d> Use `notifyutil -v -p %s` to dump statistics.\n", pid, key); // XXX
```
Scony: leftover

Source/WebCore/PAL/pal/Logging.cpp
```diff
@@ -26,27 +26,21 @@
...
 #include "Logging.h"
 
-#include <wtf/text/WTFString.h>
-
-#if PLATFORM(COCOA)
-#include <notify.h>
-#include <wtf/BlockPtr.h>
-#include <wtf/darwin/DispatchExtras.h>
+#if USE(GLIB)
+#include <unistd.h>
```
Scony: is this one really necessary?

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,620 @@
...
+        m_platformStatus = platformStatus;
+    }
+    bool isError()
+    {
+        return m_status;
+    }
+    enum GenericStatus genericStatus()
+    {
+        return m_status;
+    }
```
Scony: const

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,620 @@
...
+
+    static void setDispatchQueue(dispatch_queue_t queue)
+    {
+        s_dispatchQueue = queue;
+    }
+    static RefPtr<NotificationPoint> create(String name,
+                                            String path,
+                                            NotificationPointStatus* status,
+                                            Function<void(NotificationPoint*)>&& cb = nullptr)
```
Scony: consider `callback`

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,620 @@
...
+    NotificationPointStatus getAndClearNotificationsPosted(bool* notificationsPosted)
+    {
+        int value;
+        auto status = NotificationPointStatus::fromPlatformStatus(notify_check(m_token, &value));
+        if (!status.isError())
+            *notificationsPosted = static_cast<bool>(value);
+        return status;
+    }
+
+    bool isLive()
```
Scony: const

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,620 @@
...
+                    // no need to unown it.
+                    s_ownerIDs.remove(np->m_name);
+                }
+            }
+            return;
+        }
+        auto map = s_notificationPoints.find(np->m_name);
+        RELEASE_ASSERT(map != s_notificationPoints.end());
+        if (!map->value.remove(np->m_path))
+            RELEASE_ASSERT_NOT_REACHED();
```
Scony: perhaps `RELEASE_ASSERT(map->value.remove(np->m_path))`

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,620 @@
...
+        if (map->value.isEmpty()) {
+            if (!s_notificationPoints.remove(np->m_name))
+                RELEASE_ASSERT_NOT_REACHED();
```
Scony: same

Scony (commented):
Really nice piece of work.

I haven't checked everything yet, but left a few comments.

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,620 @@
...
+                1000,
+                nullptr,
+                &error)) {
+            g_warning("Couldn't setState for path %s under %s: %s", path.utf8().data(), name.utf8().data(), error->message);
+            return NotificationPointStatus::fromPlatformStatus(error->code);
+        }
+        return NotificationPointStatus::ok();
+    }
+
+    NotificationPointStatus notify()
```
justinmichaud: This is unused except for testing, right? I think we could move all of the API that is not currently used in the engine to private, that way we don't need to support it for future backends.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Thanks for all the suggestions @Scony, should all be integrated now.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Well, I haven't updated all users of notify_* yet; only a few, as a demonstration. Agreed that if something is clearly not used after I've gone through the code base, we can remove it.

Source/JavaScriptCore/wasm/WasmOpcodeCounter.cpp
```diff
@@ -101,16 +97,15 @@ void WasmOpcodeCounter::dump()
 
 void WasmOpcodeCounter::registerDispatch()
 {
-#if PLATFORM(COCOA)
+#if HAVE(NOTIFICATIONPOINT)
     static std::once_flag registerFlag;
     std::call_once(registerFlag, [&]() {
         int pid = getpid();
         const char* key = "com.apple.WebKit.wasm.op.stat";
```
darinadler: I think you want to delete this.

Source/JavaScriptCore/wasm/WasmOpcodeCounter.cpp
```diff
@@ -101,16 +97,15 @@ void WasmOpcodeCounter::dump()
...
         dataLogF("<WASM.OP.STAT><%d> Registering callback for wasm opcode statistics.\n", pid);
         dataLogF("<WASM.OP.STAT><%d> Use `notifyutil -v -p %s` to dump statistics.\n", pid, key);
```
darinadler: I think you want to do the same kind of pattern here as in `VM.cpp`.

Source/WTF/wtf/darwin/NotificationPoint.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+NotificationPointStatus NotificationPoint::setState(uint64_t state)
+{
+    return NotificationPointStatus::fromPlatformStatus(notify_set_state(m_token, state));
+}
+
+NotificationPointStatus NotificationPoint::setState(String name, String path, uint64_t value)
+{
+    int token;
+    auto key = makeString(name, "."_s, path);
```
darinadler:
```suggestion
    auto key = makeString(name, '.', path);
```

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+    }
+    WTF_EXPORT_PRIVATE static NotificationPointStatus fromPlatformStatus(int);
+private:
+    GenericStatus m_status;
+    int m_platformStatus;
+};
+
+class NotificationPoint : public ThreadSafeRefCounted<NotificationPoint> {
+public:
+    WTF_EXPORT_PRIVATE static const String platformOwnerName();
```
darinadler: This should be `ASCIILiteral`. But also I don’t think "platform owner name" is a good concept. Maybe "path prefix"?

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+    WTF_EXPORT_PRIVATE static RefPtr<NotificationPoint> create(String name, String providedPath, NotificationPointStatus*, Function<void(NotificationPoint*)>&& cb = nullptr);
```
darinadler: Seems like the prefix should be built into this class rather than having each caller call `platformOwnerName()`. I would move that function inside the class.

To return specifics of an error, we should use `WTF::Expected<>`, not an out pointer.

I don’t understand why the callback function receives a `NotificationPoint*`. None of the clients are using it.

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+    NotificationPointStatus getAndClearNotificationsPosted(bool* notificationsPosted);
```
darinadler: To return an error we should use `WTF::Expected`, not make this a `get` function that takes a pointer.

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+    bool isLive() const;
+    WTF_EXPORT_PRIVATE static void destroy(RefPtr<NotificationPoint>);
```
darinadler: This should take a `NotificationPoint&`, not a `RefPtr<NotificationPoint>`, but also, why would we need a public `destroy` function for a reference counted class?

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+    NotificationPointStatus setState(uint64_t value);
+    static NotificationPointStatus setState(String name, String providedPath, uint64_t value);
+    NotificationPointStatus notify();
+    static NotificationPointStatus notify(String name, String providedPath);
+    static bool isEmpty();
+
+    inline const Function<void(NotificationPoint*)>& callback() const
+    {
+        return m_callback;
+    }
```
darinadler: This should not be a public function. Instead the code that calls it should be a private static member function.

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+class NotificationPoint : public ThreadSafeRefCounted<NotificationPoint> {
+public:
+    WTF_EXPORT_PRIVATE static const String platformOwnerName();
+    WTF_EXPORT_PRIVATE static RefPtr<NotificationPoint> create(String name, String providedPath, NotificationPointStatus*, Function<void(NotificationPoint*)>&& cb = nullptr);
+    NotificationPointStatus getAndClearNotificationsPosted(bool* notificationsPosted);
+    bool isLive() const;
+    WTF_EXPORT_PRIVATE static void destroy(RefPtr<NotificationPoint>);
+    WTF_EXPORT_PRIVATE ~NotificationPoint();
+    NotificationPointStatus getState(uint64_t* value);
+    static NotificationPointStatus getState(String name, String providedPath, uint64_t* value);
```
darinadler: Why do we need this?

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+    NotificationPointStatus setState(uint64_t value);
+    static NotificationPointStatus setState(String name, String providedPath, uint64_t value);
```
darinadler: Why do we need this?

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+    NotificationPointStatus notify();
+    static NotificationPointStatus notify(String name, String providedPath);
```
darinadler: Why do we need this?

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+
+class NotificationPoint : public ThreadSafeRefCounted<NotificationPoint> {
+public:
+    WTF_EXPORT_PRIVATE static const String platformOwnerName();
+    WTF_EXPORT_PRIVATE static RefPtr<NotificationPoint> create(String name, String providedPath, NotificationPointStatus*, Function<void(NotificationPoint*)>&& cb = nullptr);
+    NotificationPointStatus getAndClearNotificationsPosted(bool* notificationsPosted);
+    bool isLive() const;
+    WTF_EXPORT_PRIVATE static void destroy(RefPtr<NotificationPoint>);
+    WTF_EXPORT_PRIVATE ~NotificationPoint();
+    NotificationPointStatus getState(uint64_t* value);
```
darinadler: To return an error we should use `WTF::Expected`, not make this a `get` function that takes a pointer.

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+    static NotificationPointStatus getState(String name, String providedPath, uint64_t* value);
+    NotificationPointStatus setState(uint64_t value);
```
darinadler: If this is called `setState`, then why is the argument named `value`?

Source/WTF/wtf/darwin/NotificationPoint.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+NotificationPointStatus NotificationPoint::getState(uint64_t* state)
+{
+    return NotificationPointStatus::fromPlatformStatus(notify_get_state(m_token, state));
+}
+
+NotificationPointStatus NotificationPoint::getState(String name, String path, uint64_t* value)
+{
+    int token;
+    auto key = makeString(name, "."_s, path);
```
darinadler:
```suggestion
    auto key = makeString(name, '.', path);
```

Source/WTF/wtf/darwin/NotificationPoint.cpp
```diff
@@ -0,0 +1,235 @@
...
+static void dispatchCallback(int token);
+
+const String NotificationPoint::platformOwnerName()
+{
+    return "com.apple.WebKit"_s;
+}
+
+NotificationPoint::NotificationPoint(String key, int token, Function<void(NotificationPoint*)>&& callback)
+    : m_token(token)
+    , m_key(key)
```
darinadler, line 55:
```suggestion
    , m_key(WTF::move(key))
    , m_callback(WTF::move(callback))
```

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+    // Unfortunately, moving this def to the .cpp file causes `tap installapi`
+    // (in the XCode build) to fail, since it apparently uses an .mm file to
+    // import things, but dispatch_queue_t expands to NSObject<> there but a
+    // struct on the C++ side, so the decl doesn't match the definition..
+    static void setDispatchQueue(dispatch_queue_t queue)
+    {
+        s_dispatchQueue = queue;
+    }
+    static std::optional<dispatch_queue_t> s_dispatchQueue;
+    const String& key() const;
```
darinadler, line 86: This should not be a public function.

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+#if OS(DARWIN)
+    // Unfortunately, moving this def to the .cpp file causes `tap installapi`
+    // (in the XCode build) to fail, since it apparently uses an .mm file to
+    // import things, but dispatch_queue_t expands to NSObject<> there but a
+    // struct on the C++ side, so the decl doesn't match the definition..
+    static void setDispatchQueue(dispatch_queue_t queue)
+    {
+        s_dispatchQueue = queue;
+    }
+    static std::optional<dispatch_queue_t> s_dispatchQueue;
```
darinadler: This should not be a public data member.

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+    {
+        return m_callback;
+    }
+
+#if OS(DARWIN)
+    // Unfortunately, moving this def to the .cpp file causes `tap installapi`
+    // (in the XCode build) to fail, since it apparently uses an .mm file to
+    // import things, but dispatch_queue_t expands to NSObject<> there but a
+    // struct on the C++ side, so the decl doesn't match the definition..
+    static void setDispatchQueue(dispatch_queue_t queue)
```
darinadler: This is used only for testing. Do we really need it? If we do, can we give it a name that makes it clear it’s only for testing?

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+    bool isLive() const;
+    WTF_EXPORT_PRIVATE static void destroy(RefPtr<NotificationPoint>);
+    WTF_EXPORT_PRIVATE ~NotificationPoint();
+    NotificationPointStatus getState(uint64_t* value);
+    static NotificationPointStatus getState(String name, String providedPath, uint64_t* value);
+    NotificationPointStatus setState(uint64_t value);
+    static NotificationPointStatus setState(String name, String providedPath, uint64_t value);
+    NotificationPointStatus notify();
+    static NotificationPointStatus notify(String name, String providedPath);
+    static bool isEmpty();
```
darinadler: This is used only for testing. We should consider not having it.

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+
+    inline const String& path() const
+    {
+        return m_path;
+    }
+
+    inline const GSimpleActionGroup* actionGroup() const
+    {
+        return m_actionGroup;
+    }
```
darinadler: Exposing all these as public functions is not a good idea.

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+
+    NotificationPointStatus setNotificationsPosted();
+    void setExportID(guint);
+#endif // OS(DARWIN)
+
+private:
+#if OS(DARWIN)
+    NotificationPoint(String key, int token, Function<void(NotificationPoint*)>&& callback);
```
darinadler: The key argument should be `ASCIILiteral`, not `String`.

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+    int m_token;
+    const String m_key;
+#elif USE(GLIB)
+    NotificationPoint(String, String, GSimpleActionGroup*, GSimpleAction*, Function<void(NotificationPoint*)>&&);
+    const String m_name;
+    const String m_path;
```
darinadler: These should all be `ASCIILiteral`, not `String`.

Source/WTF/wtf/NotificationPoint.h
```diff
@@ -0,0 +1,166 @@
...
+        return NotificationPointStatus(ConnectionError, platformStatus);
+    }
+    static NotificationPointStatus unset()
+    {
+        return NotificationPointStatus(Unset);
+    }
+    WTF_EXPORT_PRIVATE static NotificationPointStatus fromPlatformStatus(int);
+private:
+    GenericStatus m_status;
+    int m_platformStatus;
```
darinadler: This is only set, never fetched. I suggest we don’t store something that we never use. I think we don’t need a structure for this at all. We can just use an enumeration.

darinadler (changes requested):
Seems like a good idea but needs quite a bit of work. I made a first round of comments.

The strangest thing here is the `destroy` function. We should not have a reference counted object, a destructor, and a `destroy` function. Perhaps there’s some other name for the operation other than `destroy`.

Source/JavaScriptCore/wasm/WasmOpcodeCounter.cpp
aoikonomopoulos: Right, thanks.

Source/JavaScriptCore/wasm/WasmOpcodeCounter.cpp
aoikonomopoulos: Yup, missed that.

Source/WTF/wtf/darwin/NotificationPoint.cpp
aoikonomopoulos: :+1:

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Yah, this only matters on d-bus platforms, where this prefix is used as the "owned name" on the bus. Perhaps rendered moot by your later suggestion.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Yah, I only left the prefix in for future flexibility, there shouldn't be any current need for it AFAICT.

WTF::Expected, right, thank you.

The NotificationPoint* is only there because Darwin allows you to cancel the token in the callback, and it felt like something that someone might want to use in the future. I'm happy to drop it if you think it's overkill.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: :+1:

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Hopefully explained above (internal tracking list to avoid duplicate notification events and (on darwin) to look up the notification point for the token we're given in the block). If you have a preference between renaming the function and changing the semantics (or some third option), I'm open to all suggestions.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Fair enough, I may have gone a bit too far in unburdening NotificationPoint.h from platform-specific static member functions :-)

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: A reasonable question. I haven't updated all users of notify_* in this PR, since this was intended for review. There are a couple of users of `notify_{g,s}et_state` in the tree. I'll make sure to convert those too in the updated PR.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: See above.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Same as above, there are users of notify_post() in the tree.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: :+1:

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Right, will rename to `state`.

Source/WTF/wtf/darwin/NotificationPoint.cpp
aoikonomopoulos: :+1:

Source/WTF/wtf/darwin/NotificationPoint.cpp
aoikonomopoulos: Ah, thanks.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Hmm, `key()` is only there for (preexisting) users that want to print out instructions on  how to trigger the NotificationPoint. If we drop those messages (perhaps in favor of a link to a wiki page?), it can go.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Oops, thanks.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Sure thing. IIRC it was needed because TestWTF turns off concurrent event handling, but I can double-check.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Maybe we can only expose it when compiling TestWTF? Would be great to test that future changes don't introduce any leaks.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Yah, can remove them in favor of making some static functions static class functions.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Good point, thanks.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: :+1:

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Works for me, I just didn't want to limit the options of future users, but I guess they can add more cases to a `NotificationPointStatus` enum as needed.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Uh, actually this won't work since `::create` needs to synthesize the path out of two components, right?

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Same here. It could work for path but the owned name should contain the PID (or some unique identifier) on d-bus.

Source/WebCore/PAL/pal/Logging.cpp
aoikonomopoulos: Nope, thanks.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Thanks, changed a bunch of places away from RELEASE_ASSERT_NOT_REACHED().

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Renamed it to ::cancel FWIW.

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Removed -- now ::create uses this internally and there's a new ::createWithName that can override it (we want to do that in `TestWTF` anyway).

Source/WTF/wtf/NotificationPoint.h
aoikonomopoulos: Renamed it to `testWTFisEmpty`.

Source/WebKit/NetworkProcess/NetworkProcess.h
darinadler: Seems unfortunate that a `NotificationPoint` is now a reference counted object instead of just an object that wraps a token. A lot more memory, but I suppose it’s OK since there are so few of these.

No need to write `{ nullptr }` in a case like this.

Source/WTF/wtf/darwin/NotificationPoint.cpp
darinadler: No need for `= std::nullopt`.

darinadler, line 25: We normally leave a blank line here.

darinadler: It’s OK but a bit unconventional in our project to have these kinds of forward declarations. Instead we often just put these kinds of functions at the top of the file.

darinadler: This should take a `const String&` for the key; I’d actually prefer an `ASCIILiteral`, but you had some reason that you didn’t want to go with that, but I’m not sure what it is. I don’t think we need to support computed keys rather than literals. If there was some reason instead that we decide to take `String` then we’d want to `WTF::move` it.

darinadler: This should take a `StringView`, `const String&`, or `ASCIILiteral` since all it does with this path is concatenate it.

darinadler: I think these arguments should both be `ASCIILiteral`, not `String`. We don’t need to support passing computed strings and it’s better to be more efficient.

darinadler: Seems like there is a race here. What prevents us from overwriting a notification point that is set by another thread? The contains check was a long time ago and we didn’t keep the mutex locked.

darinadler: Strange line breaking here.

darinadler: It seems wasteful to make a map just so we can check if a notification point by this name already exists, when `notify_register_dispatch` and `notify_register_check` already do that and return an error in that case.

darinadler: This function should take a `NotificationPoint&`, not a `RefPtr<NotificationPoint>`, and so it should just be a member function named `cancel()` with no argument.

darinadler: Unnecessary to look up the callback by key here. We capture `key` but we could have instead allocated the `NotificationPoint` and captured the pointer to that.

darinadler (changes requested):
I didn’t review this whole thing, and didn’t look at the GLib implementation yet. I think the design is not a lightweight enough wrapper around `libnotify`. It’s good to allow other platforms to implement this, and abstract it away, but the wrapper is too heavy and adds unnecessary overhead without obvious benefit. Too much `String` allocation, an unnecessary map. Our goal should be to stay much lighter.

Source/WTF/wtf/darwin/NotificationPoint.cpp
aoikonomopoulos: ASCIILiteral isn't great on glib cause we want the name to be "org.webkitgtk-%d". I guess technically I can have `createWithName()` unconditionally concat the PID on `USE(GLIB)` platforms, it just felt weird to go that for for a handful of objects that don't allocate that often.
If this were in a fast path somewhere we could also look into compile-time concatenation of ASCIILiterals, but that seems like overkill for this use case? I'll implement what's described in the previous paragraph then.

aoikonomopoulos: So, I just double-checked in case I was mistaken about this but AFAICT `notify_register_dispatch` doesn't error out when you try to register a second callback for the same key ([example program](https://gist.github.com/aoikonomopoulos/ac8504a8197d062cca879993159bd287)).

I understand why that makes sense for a system library but, for WebKit, it sounds more like a misuse of the API. Full disclosure, it'd mean changes to the glib implementation to chain callbacks etc. Doable but I'd only do that if someone feels we want those semantics.

aoikonomopoulos: Sure. Perhaps we can capture the callback directly and skip `m_callback` then, I'll have to look into whether I can move it into the block though.

Source/WTF/WTF.xcodeproj/project.pbxproj
nikolaszimmermann: Typo in the comment, s/.cp/.cpp/

Source/WTF/wtf/darwin/NotificationPoint.cpp
aoikonomopoulos: Hmm. We're using `pointsByKey` in this line which is the `DataMutexLocker`, right? So AFAICT the `DataMutexLocker` is still in scope so the mutex is still locked?

Source/WTF/WTF.xcodeproj/project.pbxproj
aoikonomopoulos: Thank you eagle-eyed @nikolaszimmermann!

Source/WTF/wtf/darwin/NotificationPoint.cpp
aoikonomopoulos: Done.

aoikonomopoulos:
@darinadler Should be much more lightweight on the Darwin side now, would be great if you could take another look!

Source/WTF/wtf/glib/NotificationPoint.cpp
justinmichaud, line 386: Is this a UAF?

>If you break out of a such a while loop using g_variant_iter_loop() then you must free or unreference all the unpacked values as you would with g_variant_get(). Failure to do so will cause a memory leak.

justinmichaud:
I feel like there is a lot of code to go through here that I can't imagine ever getting used. Since it seems like some of that code does still have bugs, I really think that we should focus on the smaller subset that we already have a use for, or have a concrete plan to use in the next few months.

Source/WTF/wtf/darwin/NotificationPoint.cpp
justinmichaud, line 72: What does this do?

Source/WTF/wtf/NotificationPoint.h
justinmichaud, line 129: GRefPtr?

Source/WTF/wtf/glib/NotificationPoint.cpp
justinmichaud, line 252: Do we have a test for this?

justinmichaud, line 130: Why not remove?

justinmichaud:
I feel like there is a lot of potential for the lifecycle of things here to be wrong. Can we not just remove all of that and make this be a never destroyed singleton? Do we actually have a use case that requires this extra complexity?

Source/WTF/wtf/glib/NotificationPoint.cpp
justinmichaud, line 275: Is it ever possible for this to trigger? Dbus issues shouldn't take down the browser

justinmichaud, line 498: Should we be asserting that we are on the main thread?

Source/WebKit/NetworkProcess/NetworkProcess.cpp
justinmichaud, line 273: No need for the if?

Source/WTF/wtf/FastMalloc.cpp
justinmichaud, line 212: Can we use NeverDestroyed for these?

Source/WTF/wtf/glib/NotificationPoint.cpp
aoikonomopoulos: Isn't this backwards? From the quoted text, the implication I get is that you have a reference to the unpacked value (in `&returnedValue`) and, if you stop iterating, you're responsible for freeing it. This seems clear from the `g_variant_valist_get*()` implementations in https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gvariant.c, which all take a `gboolean free` argument (whether to free the previous value), which is false in the first iteration and true thereafter.

But this looks like a leak of the GVariant pointed to by `returnedValue` for sure, thanks!

aoikonomopoulos: New notification points are only created when holding `OwnedNotificationPoint::ownedNotificationPointLock`, so this should be thread-safe, no?

Source/WTF/wtf/FastMalloc.cpp
aoikonomopoulos: Sure, though that would mean introducing a name for something that's never referenced by name again 🤷

Source/WebKit/NetworkProcess/NetworkProcess.cpp
aoikonomopoulos: Ah, thanks.

Source/WTF/wtf/darwin/NotificationPoint.cpp
aoikonomopoulos: It's just a defensive initialization, happy to remove it.

Source/WTF/wtf/glib/NotificationPoint.cpp
aoikonomopoulos: Indeed.

aoikonomopoulos: This shouldn't be possible, cause we only end up here if the `OwnedNotificationPoint` is in `ownedNotificationPoints`, which means it has been exported.

But now I noticed that `OwnedNotificationPoint::processPendingForName` only prints out a warning when that happens and still moves the point to `ownedNotificationPoints`. I'll fix that bit, but if your point is that we should try to plod along here regardless, I can do that as well.

aoikonomopoulos: Do you mean testing that the notification point callback doesn't fire when we `NotificationPoint::notify` on its name/path after we've destroyed it? If so then: nope, and will add such a test.

aoikonomopoulos:
> I feel like there is a lot of code to go through here that I can't imagine ever getting used. Since it seems like some of that code does still have bugs, I really think that we should focus on the smaller subset that we already have a use for, or have a concrete plan to use in the next few months.

I hear you. ATM, it's `m_mediaStreamingActivityNotificationPoint` which has a non-static lifecycle and uses `->setState()`. Technically we could leave that using `notify_*` and further simplify NotificationPoint. As long as we don't mind having overlapping APIs on Darwin, that would work as well.

On the Darwin side, this is now a very thin wrapper around `notify_*`, so hopefully that's Obviously Correct (tm). On the glib side, I'll see if I can force the d-bus "error" callbacks to run by force-closing the connection to see if we can get some more test coverage there.

PR #60097, opened 2026-03-06
[webkitcorepy] Add support for dynamic configuration files

JonWBedard:
#### 202dd979c6efc278bd38506b3e3ff8ca5f75eccb
<pre>
[webkitcorepy] Add support for dynamic configuration files
<a href="https://bugs.webkit.org/show_bug.cgi?id=308058">https://bugs.webkit.org/show_bug.cgi?id=308058</a> <a href="https://rdar.apple.com/problem/170556228">rdar://problem/170556228</a>

Reviewed by Aakash Jain.

Build a dynamic configuration renderer based on the json-e module.

* Tools/Scripts/libraries/webkitcorepy/render-config: Add stand-alone script to render a dynamic configuration.
* Tools/Scripts/libraries/webkitcorepy/webkitcorepy/__init__.py: Add pyyaml and jsone.
* Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py: Added.
(Config.Mode): Enum for different display modes.
(Config.render): Recursively render a dictionary. If a $context is provided, render each child with this function.
Otherwise, invoke jsone&apos;s render on all children.
(Config.render.dynamic_pop):
(Config.loads): Load YAML or JSON configuration from a string.
(Config.load): Load YAML or JSON configuration from a file.
(Config.__init__): Pass display mode on construction.
(Config.dump): Print the rendered configuration to a file.
(Config.dumps): Print the rendered configuration as a string.
* Tools/Scripts/libraries/webkitcorepy/webkitcorepy/tests/config_unittest.py: Added.
(ConfigRenderTestCase): Added.
(ConfigLoadsTestCase): Added.
(ConfigLoadTestCase): Added.
(ConfigDumpsTestCase): Added.
(ConfigPopTestCase): Added.
* Tools/Scripts/libraries/webkitcorepy/webkitcorepy/tests/example-config.yaml: Add an example configuration to
demonstrate the features of the configuration renderer.

Canonical link: <a href="https://commits.webkit.org/311395@main">https://commits.webkit.org/311395@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/01d925a15d3b6898d105b9aa80df2dd6c7e93612

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/156432 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/29767 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/22949 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/165253 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/110512 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/158303 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/29900 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/29771 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/121152 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85165 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/159390 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/23374 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/140487 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/101821 "Passed tests") | | ⏳ 🛠 vision-apple 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/156180 "Passed tests") | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/22438 "Build was cancelled. Recent messages:OS: Tahoe (26.2), Xcode: 26.2; Skipping applying patch since patch_id isn't provided; Checked out pull request; Skipped layout-tests; 4 new passes 2 flakes 13 failures; Uploaded test results; 4 new passes 1 flakes 15 failures; Running compile-webkit-without-change") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/20616 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/13025 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/132118 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/18318 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/167735 "Built successfully") | | 
| | | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/19931 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/129276 "Passed tests") | | 
| [✅ 🧪 services](https://ews-build.webkit.org/#/builders/28/builds/155830 "Passed tests") | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/29368 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/24684 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/129387 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/29290 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/140112 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/87086 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/23877 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/24213 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/16911 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/29000 "Built successfully") | | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/28526 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/28754 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/28650 "Built successfully") | | | |

### DIFF
--- a/Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
+++ b/Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
@@ -1,0 +1,155 @@
+# Copyright (C) 2026 Apple Inc. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1.  Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+# 2.  Redistributions in binary form must reproduce the above copyright
+#    notice, this list of conditions and the following disclaimer in the
+#    documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+# ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import io
+import json
+import os
+from collections.abc import Iterable, Mapping
+from enum import Enum, auto
+from typing import Callable, Dict, IO, List, Optional, Tuple, Union
+
+
+class Config(dict[str, 'Config.Values']):
+    class Mode(Enum):
+        JSON = auto()
+        YAML = auto()
+
+    CONTEXT_SYMBOL: str = '$context'
+    Values = Union[str, int, float, bool, None, List['Config.Values'], Dict[str, 'Config.Values']]
+
+    @classmethod
+    def render(cls, value: 'Config.Values', context: Optional[Dict[str, Union['Config.Values', Callable[..., 'Config.Values']]]] = None) -> 'Config.Values':
+        import jsone
+        from jsone.render import parse as jsone_parse
+
+        def dynamic_pop(jsone_context: Mapping[str, Union['Config.Values', Callable[..., 'Config.Values']]], collection: Union[List['Config.Values'], Mapping[str, 'Config.Values']], *args: Union[int, str]) -> 'Config.Values':
+            """Pop an item from a mapping or list which matches some criteria, usable as a jsone built-in.
+
+            For mappings, requires a key argument.
+
+            For lists, accepts an optional index (int) or a jsone filter expression
+            (str) that is evaluated against each item to find the first match. Intended
+            to allow, for example, a configuration to progressively consume a pre-existing
+            list of machines with listed properties.
+            """
+            if isinstance(collection, list):
+                if args and isinstance(args[0], str):
+                    filter_expr = args[0]
+                    for i, item in enumerate(collection):
+                        subcontext = dict(jsone_context)
+                        if isinstance(item, Mapping):
+                            subcontext.update(item)
+                        else:
+                            subcontext['it'] = item
+                        if jsone_parse(filter_expr, subcontext):
+                            return collection.pop(i)
+                    raise ValueError(f'No item in list matches filter: {filter_expr!r}')
+                return collection.pop(int(args[0])) if args else collection.pop(0)
+            if isinstance(collection, Mapping):
+                if not args:
+                    raise TypeError('pop() on a mapping requires a key argument')
+                return collection.pop(args[0])
+            raise TypeError('pop() requires a list or mapping')
+
+        # Treat pop as a jsone built-in so we can apply a filter against it
+        dynamic_pop._jsone_builtin = True
+
+        context = dict(context or {})
+        context.setdefault('pop', dynamic_pop)
+
+        if not isinstance(value, Mapping):
+            return jsone.render(value, context=context)
+
+        if cls.CONTEXT_SYMBOL in value:
+            result = Config()
+            context = dict(**context)
+            context.update(cls.render(value[cls.CONTEXT_SYMBOL], context=context))
+            for key, content in value.items():
+                if key == cls.CONTEXT_SYMBOL:
+                    continue
+                result[key] = cls.render(content, context=context)
+                if isinstance(result[key], list):
+                    context[key] = [x for x in result[key]]
+                elif isinstance(result[key], Mapping):
+                    result[key] = dict(**result[key])
+                    context[key] = dict(**result[key])
+                else:
+                    context[key] = result[key]
+            return result
+
+        result = jsone.render(value, context=context)
+        if isinstance(result, Mapping):
+            return Config(**result)
+        return result
+
+    @classmethod
+    def loads(cls, string: str, mode: Optional['Config.Mode'] = None) -> 'Config':
+        return cls.load(io.StringIO(string), mode=mode)
+
+    @classmethod
+    def load(cls, file: IO[str], mode: Optional['Config.Mode'] = None) -> 'Config':
+        if mode is None and hasattr(file, 'name'):
+            ext = os.path.splitext(file.name)[1].lower()
+            if ext == '.json':
+                mode = cls.Mode.JSON
+            elif ext in ('.yaml', '.yml'):
+                mode = cls.Mode.YAML
+        if mode is cls.Mode.JSON:
+            result = cls.render(json.load(file))
+        else:
+            import yaml
+
+            # yaml.safe_load_all returns a generator, not a sequence.
+            documents = list(yaml.safe_load_all(file))
+            if not documents:
+                result = cls()
+            elif len(documents) == 1:
+                if not isinstance(documents[0], dict):
+                    raise TypeError('yaml sub-document is not a dictionary')
+                result = cls.render(documents[0])
+            else:
+                data = {cls.CONTEXT_SYMBOL: documents[0]}
+                for doc in documents[1:]:
+                    if not isinstance(doc, dict):
+                        raise TypeError('yaml sub-document is not a dictionary')
+                    data.update(doc)
+                result = cls.render(data)
+            mode = cls.Mode.YAML
+        result.mode = mode
+        return result
+
+    def __init__(self, mapping: Union[Mapping[str, 'Config.Values'], Iterable[Tuple[str, 'Config.Values']], None] = None, **kwargs: 'Config.Values') -> None:
+        super().__init__(**kwargs) if mapping is None else super().__init__(mapping, **kwargs)
+        self.mode: 'Config.Mode' = self.Mode.YAML
+
+    def dump(self, file: IO[str], mode: Optional['Config.Mode'] = None, indent: int = 4) -> None:
+        mode = mode or self.mode
+        if mode is self.Mode.JSON:
+            json.dump(dict(self), file, indent=indent)
+        else:
+            import yaml
+            yaml.dump(dict(self), file, indent=indent, default_flow_style=False)
+
+    def dumps(self, mode: Optional['Config.Mode'] = None, indent: int = 4) -> str:
+        file = io.StringIO()
+        self.dump(file, mode=mode, indent=indent)
+        return file.getvalue()
--- a/Tools/Scripts/libraries/webkitcorepy/webkitcorepy/tests/example-config.yaml
+++ b/Tools/Scripts/libraries/webkitcorepy/webkitcorepy/tests/example-config.yaml
@@ -1,0 +1,48 @@
+# Context:
+machine_counts:
+  - model: Mac16,10
+    count: 10
+  - model: Mac14,3
+    count: 10
+names:
+  $map: {$eval: "range(100, 120)"}
+  each(i): bot${i}
+versions:
+  macos:
+    [Tahoe, Sequoia, Sonoma]
+build_types: [Release, Debug]
+---
+workers:
+  $flatten:
+    $map: {$eval: machine_counts}
+    each(entry):
+      $map: {$eval: "range(0, entry.count)"}
+      each(n):
+        name: {$eval: pop(names)}
+        model: {$eval: entry.model}
+
+builders:
+  $flatten:
+    $map: {$eval: versions.macos}
+    each(version):
+      $map: {$eval: build_types}
+      each(build_type):
+        name: Apple-${version}-${build_type}-Build
+        factory: BuildFactory
+        platform: mac-${lowercase(version)}
+        configuration: {$eval: lowercase(build_type)}
+        architectures: [x86_64, arm64]
+        workernames:
+          - $if: "build_type == 'Debug'"
+            then: {$eval: "pop(workers, \"model == 'Mac16,10'\").name"}
+            else: {$eval: "pop(workers, \"model == 'Mac14,3'\").name"}
+          - $if: "build_type == 'Debug'"
+            then: {$eval: "pop(workers, \"model == 'Mac16,10'\").name"}
+            else: {$eval: "pop(workers, \"model == 'Mac14,3'\").name"}
+          - $if: "build_type == 'Debug'"
+            then: {$eval: "pop(workers, \"model == 'Mac16,10'\").name"}
+            else: {$eval: "pop(workers, \"model == 'Mac14,3'\").name"}
+
+unused-workers:
+  $map: {$eval: workers}
+  each(w): {$eval: w.name}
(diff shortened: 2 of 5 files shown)


Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,139 @@
...
+            return Config(**result)
+        return result
+
+    @classmethod
+    def loads(cls, string: str, mode: Optional['Config.Mode'] = None) -> 'Config':
+        if mode is cls.Mode.JSON:
+            result = cls.render(json.loads(string))
+        else:
+            import yaml
+            documents = list(yaml.safe_load_all(string))
```
gsnedders: ~~Why `list` here?~~

Ah, while [the docs say](https://pyyaml.org/wiki/PyYAMLDocumentation#:~:text=safe_load_all(stream)%20parses%20the%20given%20stream%20and%20returns%20a%20sequence%20of%20Python%20objects%20corresponding%20to%20the%20documents%20in%20the%20stream.):

> `safe_load_all(stream)` parses the given `stream` and returns a sequence of Python objects corresponding to the documents in the stream.

The docs are in fact incorrect: https://github.com/yaml/pyyaml.org/issues/18

Can you add a comment just to make the reason for the `list()` here explicit, and that it returns a generator, not a sequence.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,139 @@
...
+                result = cls()
+            elif len(documents) == 1:
+                result = cls.render(documents[0])
+            else:
+                data = {cls.CONTEXT_SYMBOL: documents[0]}
+                for doc in documents[1:]:
+                    if not isinstance(doc, dict):
+                        raise TypeError('yaml sub-document is not a dictionary')
+                    data.update(doc)
+                print(data.keys())
```
gsnedders: Left-over debug output?

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,139 @@
...
+                for i, item in enumerate(collection):
+                    subcontext = dict(jsone_context)
+                    if isinstance(item, dict):
+                        subcontext.update(item)
+                    else:
+                        subcontext['it'] = item
+                    if jsone_parse(filter_expr, subcontext):
+                        return collection.pop(i)
+                raise ValueError(f'No item in list matches filter: {filter_expr!r}')
+            return cls.pop(collection, *args)
```
gsnedders: Does it actually make sense to have this within the class? The only reason to have it here is the `cls.pop` reference, but I'm not sure that makes sense either, per my other comment.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,139 @@
...
+            documents = list(yaml.safe_load_all(string))
+            if not documents:
+                result = cls()
+            elif len(documents) == 1:
+                result = cls.render(documents[0])
+            else:
+                data = {cls.CONTEXT_SYMBOL: documents[0]}
+                for doc in documents[1:]:
+                    if not isinstance(doc, dict):
+                        raise TypeError('yaml sub-document is not a dictionary')
```
gsnedders, line 133: Do we not want a similar check in the `len(documents) == 1` case?

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @classmethod
+    def load(cls, file: IO[str], mode: Optional['Config.Mode'] = None) -> 'Config':
+        if mode is None and hasattr(file, 'name'):
+            ext = os.path.splitext(file.name)[1].lower()
+            if ext == '.json':
+                mode = cls.Mode.JSON
+            elif ext in ('.yaml', '.yml'):
+                mode = cls.Mode.YAML
+        result = cls.loads(file.read(), mode=mode)
```
gsnedders: Nit: in general, I'd probably lean towards doing this the other way around — use `io.StringIO` to wrap something we already have as a string and then call `load` with that, because then we fall into what's a faster parsing path versus loading the entire file into memory and then parsing it.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,139 @@
...
+import enum
+import json
+import os
+from typing import Any, IO, Optional, Union
+
+
+class Config(dict):
+    class Mode(enum.Enum):
+        JSON = enum.auto()
+        YAML = enum.auto()
```
gsnedders: I would probably, like the Python docs do, and do `from enum import Enum, auto` and avoid all the `enum` here.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
gsnedders: Why not, similar to a few lines above:

```suggestion
            return collection.pop(*args)
```

Or are we relying on `cls.pop(collection, *args)` to raise TypeError if `collection` isn't a `dict` instance, which it might not be?

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,139 @@
...
+                raise TypeError('pop() on a dict requires a key argument')
+            return collection.pop(args[0])
+        raise TypeError('pop() requires a list or dict')
+
+    @classmethod
+    def render(cls, value: Any, context: Optional[dict[str, Any]] = None) -> Any:
+        import jsone
+        from jsone.render import parse as jsone_parse
+
+        def dynamic_pop(jsone_context: dict, collection: Union[list[Any], dict[Any, Any]], *args: Union[int, str]) -> Any:
```
gsnedders: A docstring would be really helpful here, because I'm struggling to understand what this is trying to do.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,139 @@
...
+
+class Config(dict):
+    class Mode(enum.Enum):
+        JSON = enum.auto()
+        YAML = enum.auto()
+
+    CONTEXT_SYMBOL: str = '$context'
+
+    @classmethod
+    def pop(cls, collection: Union[list[Any], dict[Any, Any]], *args: Union[int, str]) -> Any:
```
gsnedders: This violates the Liskov Substitution Principle, as `(cls: Type[Config], collection: Union[list[Any], dict[Any, Any]], *args: Union[int, str]) -> Any` isn't contravariant of `(self: dict[_, _V], key: object) -> _V`.

It would be *much* nicer to avoid creating a `dict` subclass which neither supports the `dict` interface nor even the smaller `collections.abc.MutableMapping` interface.

aj062 (approved):
rs=me

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,139 @@
...
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import enum
+import json
+import os
+from typing import Any, IO, Optional, Union
+
+
+class Config(dict):
```
gsnedders: In general, subclassing `dict` isn't great — because `dict` has a lot of magic internally, and the fact we have an incompatible override of `dict.pop` shows some of the issue here.

I wonder if it would make more sense to split this class up into two — one which contains all the logic for loading a config, and another much thinner `collections.abc.MutableMapping` (do we even want configs after loading to be mutable? or would we be better with `collections.abc.Mapping`?) subclass for the rendered config, which more-or-less just stores the `mode` and provides `dump`/`dumps` methods, and otherwise just delegates to the object it is wrapping?

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/tests/example-config.yaml
```diff
@@ -0,0 +1,48 @@
+#Context:
```
gsnedders:
```suggestion
# Context:
```

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,139 @@
...
+            elif ext in ('.yaml', '.yml'):
+                mode = cls.Mode.YAML
+        result = cls.loads(file.read(), mode=mode)
+        return result
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        self.mode: 'Config.Mode' = self.Mode.YAML
+
+    def dumps(self, mode: Optional['Config.Mode'] = None, indent: int = 4) -> str:
```
gsnedders: No `dump`?

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,139 @@
...
+        if isinstance(collection, list):
+            return collection.pop(int(args[0])) if args else collection.pop(0)
+        if isinstance(collection, dict):
+            if not args:
+                raise TypeError('pop() on a dict requires a key argument')
+            return collection.pop(args[0])
+        raise TypeError('pop() requires a list or dict')
+
+    @classmethod
+    def render(cls, value: Any, context: Optional[dict[str, Any]] = None) -> Any:
```
gsnedders: It would probably be nicer to allow any `Mapping`, not just `dict`, for the context object?

gsnedders (changes requested):
A lot of these are just questions, because I'm not really sure what we want for a lot of this!

I'd also like to see a lot less use of the `Any` type, because it significantly weakens the type guarantees we're providing.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
JonWBedard: I don't know that we _actually_ violate Liskov's Substitution Principle, since this is a classmethod  rather than an instance method, but the point stands. This method should have a different name.

JonWBedard: Actually, should just be put inside the dynamic method entirely.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
JonWBedard: It does not, and I figured that out from your other comment before getting to this one.

JonWBedard: We are, but this point is valid, and we can move the `TypeError` into `dynamic_pop`

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
JonWBedard: After replacing the `.pop` method (which was absolutely a mistake, that should never have been a classmethod in the first place), I don't really see why we shouldn't subclass `dict`. It's true that `dict` has magic in it, but all that magic is to our benefit, I think. We really only have 3 "extras" on-top of dictionary now:
- Mode
- dump (which uses Mode)
- dumps (which calls dump)

(I think there is no good reason to _forbid_ modification of a config once it's loaded, even though I also can't think of a reason I would want to write code that did that)

gsnedders: I don't think it's a massive deal — though I will say that having immutable data structures often makes code easier to reason about, because they can't unexpectedly change when you call another function!

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,155 @@
...
+
+import io
+import json
+import os
+from collections.abc import Iterable, Mapping
+from enum import Enum, auto
+from typing import Callable, Dict, IO, List, Optional, Tuple, Union
+
+
+class Config(dict):
```
gsnedders: This should be:

```suggestion
class Config(dict[str, 'Config.Values']):
```

Rather than the default `dict[Any, Any]`.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,155 @@
...
+    class Mode(Enum):
+        JSON = auto()
+        YAML = auto()
+
+    CONTEXT_SYMBOL: str = '$context'
+    Values = Union[str, int, float, bool, None, List['Config.Values'], Dict[str, 'Config.Values']]
+
+    @classmethod
+    def render(cls, value: 'Config.Values', context: Optional[Dict[str, Union['Config.Values', Callable]]] = None) -> 'Config.Values':
```
gsnedders:
```suggestion
    def render(cls, value: 'Config.Values', context: Optional[Dict[str, Union['Config.Values', Callable[..., 'Config.Values']]]] = None) -> 'Config.Values':
```

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
```diff
@@ -0,0 +1,155 @@
...
+        import jsone
+        from jsone.render import parse as jsone_parse
+
+        def dynamic_pop(jsone_context: Mapping, collection: Union[List['Config.Values'], Mapping[str, 'Config.Values']], *args: Union[int, str]) -> 'Config.Values':
```
gsnedders:
```suggestion
        def dynamic_pop(jsone_context: Mapping[str, Union['Config.Values', Callable[..., 'Config.Values']]], collection: Union[List['Config.Values'], Mapping[str, 'Config.Values']], *args: Union[int, str]) -> 'Config.Values':
```

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/config.py
JonWBedard: I understand that, I think the reason I'm against an immutable data structure here is we're not really dealing with a complex object with a lifecycle. This is mostly just a dictionary which you load from a datasource and then do whatever you need to do. The only reason we're overriding `dict` at all is to keep the bit of encoding information around.

PR #60115, opened 2026-03-07
[CMake][GLIB] Support PGO features in regular CMake builds with clang

lauromoura:
#### 94556bd2d2ae842db20073afba23f52412dc6d81
<pre>
[CMake][GLIB] Support PGO features in regular CMake builds with clang
<a href="https://bugs.webkit.org/show_bug.cgi?id=309318">https://bugs.webkit.org/show_bug.cgi?id=309318</a>

PGO (Profile Guided Optimization) is usually performed in two phases,
which in WebKit are controlled by the following two cmake options:

- ENABLE_LLVM_PROFILE_GENERATION: Tells clang to add IR-profiling
  instrumentation code to the build. When we run the instrumented
  binaries, this code will collect the profile data and save into
  .profraw files at the PGO_PROFILE_DIR directory (overridden by
  LLVM_PROFILE_FILE envvar). These files are then merged into a single
  .profdata file.
- USE_PGO_PROFILE: Tells clang to use the profdata at PGO_PROFILE_PATH
  as an example of the expected workload, so it can make better
  predictions when generating the final build.

While GCC also supports `-fprofile-generate` and `-fprofile-use`, it
appeared much more strict during compilation, emitting various warnings
related to missing/mismatched coverage, string formatters, etc.. It also
had issues with profile generation in distributed builds, generating the
wrong annotations in remote compilation jobs. So we&apos;ll focus on clang in
this initial patch, adding GCC support in the future.

By default, the instrumented build will save the profiles using clang&apos;s
default path, but this can be overridden at compile time with
`-DPGO_PROFILE_DIR=PATH`, and at runtime with `LLVM_PROFILE_FILE`. You
might want to set it to an absolute path for distributed builds with
icecc, as the usage of `-fprofile-generate` might lead it to think we&apos;re
generating files during compilation, forcing local jobs.

During instrumented builds, there were some problems with the symbols
when compiling executables and linking, so actual instrumentation is
compiled only into WEBKIT_LIBRARY and WEBKIT_FRAMEWORK. Given the hot
paths are usually on the library itself, should not be a problem. Also,
instrumented builds add multiple __llvm_profile_filename definitions,
that clash when LTO is enabled. Given they have the same value, taken
from PGO_PROFILE_DIR, it&apos;s safe to turn this into a non-fatal warning.

By default, clang dumps the instrumented data in the `atexit` handler,
which might not be called in forced terminations. For example,
`ProcessLauncherGLib.cpp` uses SIGKILL to terminate the children. That
means that we could potentially lose the data from the WebProcess even
if the UIProcess exits cleanly.

This commit adds a SIGTERM handler in auxiliary process to explicitly
dump the collected data. It also makes the ProcessLauncher send first
SIGTERM, and wait 500ms before sending SIGKILL, which should be enough
to save the profile. Note that this only affects instrumented builds.

* Source/WebKit/Shared/unix/AuxiliaryProcessMain.cpp:
(WebKit::AuxiliaryProcess::platformInitialize):
* Source/WebKit/UIProcess/Launcher/glib/ProcessLauncherGLib.cpp:
(WebKit::ProcessLauncher::terminateProcess):
* Source/cmake/OptionsCommon.cmake:
* Source/cmake/WebKitCommon.cmake:
* Source/cmake/WebKitFeatures.cmake:

Canonical link: <a href="https://commits.webkit.org/310954@main">https://commits.webkit.org/310954@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/abaf0df8c380b01cf6f73aa4269f2936cb4feca8

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/155381 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/28641 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/21800 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/164143 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/109178 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/28781 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/28491 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/120250 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/109178 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/158340 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/22449 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/139537 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/100940 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/21535 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/19633 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/11972 "Built successfully") | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/147431 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/131204 "Passed tests") | | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/166621 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/16212 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/10789 "Built successfully") | [❌ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/18979 "Found 1 new test failure: imported/w3c/web-platform-tests/fetch/api/crashtests/huge-fetch.any.sharedworker.html (failure)") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/128359 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/28185 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/23667 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/128495 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/34875 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/28109 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/139162 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/85677 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/23333 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/15959 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/187266 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/27803 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/91906 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/48013 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/27380 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/27610 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/27453 "Built successfully") | | |


Source/cmake/WebKitCommon.cmake
```diff
@@ -258,6 +258,84 @@ if (NOT HAS_RUN_WEBKIT_COMMON)
...
+    if (ENABLE_LLVM_PROFILE_GENERATION AND NOT COMPILER_IS_CLANG)
+        message(FATAL_ERROR
+            "ENABLE_LLVM_PROFILE_GENERATION requires Clang. "
+            "GCC PGO support may be added in a future patch.")
+    endif ()
+
+    if (USE_PGO_PROFILE AND NOT COMPILER_IS_CLANG)
+        message(FATAL_ERROR
+            "USE_PGO_PROFILE requires Clang. "
+            "GCC PGO support may be added in a future patch.")
```
aperezdc: Same here:


```suggestion
        message(FATAL_ERROR "USE_PGO_PROFILE requires Clang.")
```

Source/cmake/WebKitCommon.cmake
```diff
@@ -258,6 +258,84 @@ if (NOT HAS_RUN_WEBKIT_COMMON)
         find_package(Gperf 3.0.1 REQUIRED)
     endif ()
 
+    # -------------------------------------------------------------------------
+    # PGO (Profile-Guided Optimization) support
+    # -------------------------------------------------------------------------
+    if (ENABLE_LLVM_PROFILE_GENERATION AND NOT COMPILER_IS_CLANG)
+        message(FATAL_ERROR
+            "ENABLE_LLVM_PROFILE_GENERATION requires Clang. "
+            "GCC PGO support may be added in a future patch.")
```
aperezdc: I think the second line is not really needed, it doesn't really add much to the message... and the build may not be even using GCC.


```suggestion
        message(FATAL_ERROR "ENABLE_LLVM_PROFILE_GENERATION requires Clang.")
```

Source/cmake/WebKitCommon.cmake
```diff
@@ -258,6 +258,84 @@ if (NOT HAS_RUN_WEBKIT_COMMON)
...
+
+    if (USE_PGO_PROFILE AND NOT COMPILER_IS_CLANG)
+        message(FATAL_ERROR
+            "USE_PGO_PROFILE requires Clang. "
+            "GCC PGO support may be added in a future patch.")
+    endif ()
+
+    # Phase 1: Profile Generation - build instrumented binary that writes .profraw files
+    if (ENABLE_LLVM_PROFILE_GENERATION AND COMPILER_IS_CLANG AND NOT MSVC)
+        include(CheckCXXSourceCompiles)
```
aperezdc, line 273: You need to save the check state flags with [cmake_push_check_state()](https://cmake.org/cmake/help/latest/module/CMakePushCheckState.html):


```suggestion
        include(CheckCXXSourceCompiles)
        cmake_push_check_state()
```

Source/cmake/WebKitCommon.cmake
```diff
@@ -258,6 +258,84 @@ if (NOT HAS_RUN_WEBKIT_COMMON)
...
+        set(CMAKE_REQUIRED_FLAGS "-fprofile-generate")
+        set(CMAKE_REQUIRED_LINK_OPTIONS "-fprofile-generate")
+        check_cxx_source_compiles("int main() { return 0; }" HAVE_CLANG_PROFILE_RUNTIME)
+        unset(CMAKE_REQUIRED_FLAGS)
+        unset(CMAKE_REQUIRED_LINK_OPTIONS)
```
aperezdc: ...and then here use `cmake_pop_check_state()`:


```suggestion
        cmake_pop_check_state()
```

Source/cmake/WebKitCommon.cmake
```diff
@@ -258,6 +258,84 @@ if (NOT HAS_RUN_WEBKIT_COMMON)
...
+        if (PGO_PROFILE_DIR)
+            set(_PGO_GENERATE_FLAG "-fprofile-generate=${PGO_PROFILE_DIR}")
+        else ()
+            set(_PGO_GENERATE_FLAG "-fprofile-generate")
+        endif ()
+
+        set(WEBKIT_PGO_COMPILE_OPTIONS "${_PGO_GENERATE_FLAG}" CACHE INTERNAL "")
+
+        # LTO builds error out on duplicate __llvm_profile_filename definitions.
+        set(PGO_LINK_FLAGS "${_PGO_GENERATE_FLAG}")
```
aperezdc: Why not setting `WEBKIT_PGO_COMPILE_OPTIONS` directly and skip the temporary variable polluting the global scope? Also, I don't know why this variable has a `WEBKIT_` prefix but the others do not have it, let's remove it:


```suggestion
        if (PGO_PROFILE_DIR)
            set(PGO_COMPILE_OPTIONS "-fprofile-generate=${PGO_PROFILE_DIR}" CACHE INTERNAL "")
        else ()
            set(PGO_COMPILE_OPTIONS "-fprofile-generate" CACHE INTERNAL "")
        endif ()

        # LTO builds error out on duplicate __llvm_profile_filename definitions.
        set(PGO_LINK_FLAGS "${PGO_COMPILE_OPTIONS}")
```

Source/cmake/WebKitCommon.cmake
```diff
@@ -258,6 +258,84 @@ if (NOT HAS_RUN_WEBKIT_COMMON)
...
+        if (LD_VARIANT STREQUAL "BFD" OR LD_VARIANT STREQUAL "GOLD" OR LD_VARIANT STREQUAL "LLD" OR LD_VARIANT STREQUAL "MOLD")
+            string(APPEND PGO_LINK_FLAGS " -Wl,--allow-multiple-definition")
+        endif ()
+        set(CMAKE_EXE_LINKER_FLAGS "${PGO_LINK_FLAGS} ${CMAKE_EXE_LINKER_FLAGS}")
+        set(CMAKE_SHARED_LINKER_FLAGS "${PGO_LINK_FLAGS} ${CMAKE_SHARED_LINKER_FLAGS}")
+        set(CMAKE_MODULE_LINKER_FLAGS "${PGO_LINK_FLAGS} ${CMAKE_MODULE_LINKER_FLAGS}")
```
aperezdc: Right above you used `string(APPEND ..)` so why not `string(PREPEND ..)` here? 😉 

```suggestion
        string(PREPEND CMAKE_EXE_LINKER_FLAGS "${PGO_LINK_FLAGS}")
        string(PREPEND CMAKE_SHARED_LINKER_FLAGS "${PGO_LINK_FLAGS}")
        string(PREPEND CMAKE_MODULE_LINKER_FLAGS "${PGO_LINK_FLAGS}")
```

Source/cmake/WebKitCommon.cmake
```diff
@@ -258,6 +258,84 @@ if (NOT HAS_RUN_WEBKIT_COMMON)
...
+            message(FATAL_ERROR "PGO_PROFILE_PATH does not exist: ${PGO_PROFILE_PATH}")
+        endif ()
+
+        # profile counter mismatches (e.g., source changed since profiling) trigger backend-plugin warnings,
+        # which would break the build with our usage of -Werror in CI
+        set(CMAKE_C_FLAGS "-fprofile-use=${PGO_PROFILE_PATH} -Wno-error=backend-plugin ${CMAKE_C_FLAGS}")
+        set(CMAKE_CXX_FLAGS "-fprofile-use=${PGO_PROFILE_PATH} -Wno-error=backend-plugin ${CMAKE_CXX_FLAGS}")
+        set(CMAKE_EXE_LINKER_FLAGS "-fprofile-use=${PGO_PROFILE_PATH} ${CMAKE_EXE_LINKER_FLAGS}")
+        set(CMAKE_SHARED_LINKER_FLAGS "-fprofile-use=${PGO_PROFILE_PATH} ${CMAKE_SHARED_LINKER_FLAGS}")
+        set(CMAKE_MODULE_LINKER_FLAGS "-fprofile-use=${PGO_PROFILE_PATH} ${CMAKE_MODULE_LINKER_FLAGS}")
```
aperezdc: These could also use `string(PREPEND ..)`.

Source/WebKit/Shared/unix/AuxiliaryProcessMain.cpp
```diff
@@ -92,6 +96,17 @@ void AuxiliaryProcess::platformInitialize(const AuxiliaryProcessInitializationPa
...
     RELEASE_ASSERT(!sigaction(SIGPIPE, &signalAction, nullptr));
+#if ENABLE(LLVM_PROFILE_GENERATION)
+    // __llvm_profile_dump() is not async-signal-safe but losing profile
+    // data from auxiliary processes would make PGO collection impractical.
+    if (__llvm_profile_dump) {
+        signalAction.sa_handler = [](int) {
+            __llvm_profile_dump();
+            _exit(0);
+        };
+        RELEASE_ASSERT(!sigaction(SIGTERM, &signalAction, nullptr));
```
aperezdc: Instead of this, you could use `g_unix_signal_add()` to handle the signal in the main thread's event loop, and then it is perfectly fine to call `__llvm_profile_dump()` even if it is not async-signal-safe. After completing the dump, the callback can call `AuxiliaryProcess::terminate()` to make the process stop the run loop and exit.

Something like this (untested):


```suggestion
    if (__llvm_profile_dump) {
        g_unix_signal_add(SIGTERM, [](void* data) -> gboolean {
            __llvm_profile_dump();
            static_cast<AuxiliaryProcess*>(data)->terminate();
            return G_SOURCE_REMOVE;
        }, this);
```

Of course this would depend on `USE(GLIB)` but I think that's fine.

Source/WebKit/UIProcess/Launcher/glib/ProcessLauncherGLib.cpp
```diff
@@ -289,7 +289,22 @@ void ProcessLauncher::terminateProcess()
...
+    // Ensure auxiliary processes have some time to flush their PGO data.
+    kill(m_processID, SIGTERM);
+    auto* pid = new ProcessID(m_processID);
+    g_timeout_add_full(G_PRIORITY_DEFAULT, 500, [](gpointer data) -> gboolean {
+        auto* pid = static_cast<ProcessID*>(data);
+        kill(*pid, SIGKILL);
+        return G_SOURCE_REMOVE;
+    }, pid, [](gpointer data) {
+        delete static_cast<ProcessID*>(data);
+    });
```
aperezdc: This could pass the PID directly as userdata to avoid allocating an extra `ProcessID`:


```suggestion
    g_timeout_add_full(G_PRIORITY_DEFAULT, 500, [](gpointer data) -> gboolean {
        kill(GPOINTER_TO_SIZE(data), SIGKILL);
        return G_SOURCE_REMOVE;
    }, GSIZE_TO_POINTER(m_processID), nullptr);
    });
```

aperezdc (changes requested):
Looks good overall but I think there are places that need a bit of improvement. Please take a look at the comments.

Source/cmake/WebKitCommon.cmake
```diff
@@ -257,6 +259,78 @@ if (NOT HAS_RUN_WEBKIT_COMMON)
...
+
+        if (PGO_PROFILE_DIR)
+            set(PGO_COMPILE_OPTIONS "-fprofile-generate=${PGO_PROFILE_DIR}" CACHE INTERNAL "")
+        else ()
+            set(PGO_COMPILE_OPTIONS "-fprofile-generate" CACHE INTERNAL "")
+        endif ()
+
+        # LTO builds error out on duplicate __llvm_profile_filename definitions.
+        set(PGO_LINK_FLAGS "${PGO_COMPILE_OPTIONS}")
+        if (LD_VARIANT STREQUAL "BFD" OR LD_VARIANT STREQUAL "GOLD" OR LD_VARIANT STREQUAL "LLD" OR LD_VARIANT STREQUAL "MOLD")
```
aperezdc: I think that a check for this could go in `WebKitCommon.cmake`, we already have a few checks for linker features (around line 88). Something like this:

```cmake
if (LD_USAGE MATCHES "--allow-multiple-definitions")
    set(LD_SUPPORTS_ALLOW_MULTIPLE_DEFINITIONS TRUE)
endif ()
```

and then here we can do the following, instead of doing a long check for every possible linker variant:

```suggestion
        if (LD_SUPPORTS_ALLOW_MULTIPLE_DEFINITIONS)
```

Source/cmake/WebKitCommon.cmake
```diff
@@ -257,6 +259,78 @@ if (NOT HAS_RUN_WEBKIT_COMMON)
...
+            message(FATAL_ERROR "USE_PGO_PROFILE is ON but PGO_PROFILE_PATH is not set")
+        endif ()
+        if (NOT EXISTS "${PGO_PROFILE_PATH}")
+            message(FATAL_ERROR "PGO_PROFILE_PATH does not exist: ${PGO_PROFILE_PATH}")
+        endif ()
+
+        # profile counter mismatches (e.g., source changed since profiling) trigger backend-plugin warnings,
+        # which would break the build with our usage of -Werror in CI
+        string(PREPEND CMAKE_C_FLAGS "-fprofile-use=${PGO_PROFILE_PATH} -Wno-error=backend-plugin ")
+        string(PREPEND CMAKE_CXX_FLAGS "-fprofile-use=${PGO_PROFILE_PATH} -Wno-error=backend-plugin ")
```
aperezdc: We have a `WEBKIT_PREPEND_GLOBAL_COMPILER_FLAGS` macro that may be used for this, and will also check whether the compiler actually works with the flag:


```suggestion
        WEBKIT_PREPEND_GLOBAL_COMPILER_FLAGS("-fprofile-use=${PGO_PROFILE_PATH}" "-Wno-error=backend-plugin")
```

aperezdc (approved):
Thanks for the update @lauromoura! This LGTM, with a couple of suggestions in case you might want to apply them before landing.

lauromoura:
> Thanks for the update @lauromoura! This LGTM, with a couple of suggestions in case you might want to apply them before landing.

Thanks for the review! I've updated the patch with both suggestions. Let's leave one last EWS run before landing.

aperezdc:
Backported into `webkitglib/2.52` as commit 064af56802ababec165b4ba1d446dadb1c0686e4

PR #60199, opened 2026-03-09
[GTK][WPE] `WTFEndSignpost`'s format argument is ignored

obyknovenius:
#### 9f2e3189a7581e2e48297c79f4dbe85833343c7c
<pre>
[GTK][WPE] `WTFEndSignpost`&apos;s format argument is ignored
<a href="https://bugs.webkit.org/show_bug.cgi?id=309497">https://bugs.webkit.org/show_bug.cgi?id=309497</a>

The third argument of `WTFEndSignpost` is a `printf`-like format string,
but it is ignored in the Sysprof implementation.

This patch combines begin/end signpost descriptions into a single string
separated by &quot;|&quot; character, so that both parts are visible in Sysprof.

It also adds &quot;handled&quot; and &quot;coalesced with&quot; information to the mouse
event handling signposts, which is only available at the end of the
handling routine.

* Source/WTF/wtf/glib/SysprofAnnotator.h:
(WebKit::WebPageProxy::mouseEventHandlingCompleted):

Canonical link: <a href="https://commits.webkit.org/309480@main">https://commits.webkit.org/309480@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/485abb61067568bdd52b9c76a3fcf93190b2ca83

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/150813 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/23571 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/17142 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/159536 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/104247 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/24003 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/23779 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/116406 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/82655 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/153773 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/18516 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/135298 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/97134 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/17613 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/15564 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/7383 "Built successfully") | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/142796 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/127227 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/13222 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/162010 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/11611 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/5130 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/14779 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/124408 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/23375 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/19611 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/124604 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/33815 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/23364 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/135017 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/79752 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/19674 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/11779 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/182338 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/22975 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/86775 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/46588 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/22687 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/22839 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/22741 "Built successfully") | | |

### DIFF
--- a/Source/WTF/wtf/glib/SysprofAnnotator.h
+++ b/Source/WTF/wtf/glib/SysprofAnnotator.h
@@ -65,43 +65,79 @@ public:
 
     void beginMark(const void* pointer, std::span<const char> name, const char* description, ...) WTF_ATTRIBUTE_PRINTF(4, 5)
     {
-        auto key = std::make_pair(pointer, static_cast<const void*>(name.data()));
+        auto time = SYSPROF_CAPTURE_CURRENT_TIME;
 
-        WTF_ALLOW_UNSAFE_BUFFER_USAGE_BEGIN
         Vector<char> buffer(1024);
-        va_list args;
-        va_start(args, description);
-        vsnprintf(buffer.mutableSpan().data(), buffer.size(), description, args);
-        va_end(args);
-        WTF_ALLOW_UNSAFE_BUFFER_USAGE_END
 
-        auto value = std::make_pair(SYSPROF_CAPTURE_CURRENT_TIME, WTF::move(buffer));
+        if (!description || description[0] == '\0') {
+            buffer.resize(0);
+            buffer.append('\0');
+        } else {
+            va_list args, copyArgs;
+            va_start(args, description);
+            va_copy(copyArgs, args);
+
+            auto descriptionLength = vsnprintf(nullptr, 0, description, args);
+            va_end(args);
+            buffer.resize(descriptionLength + 1);
+
+            vsnprintf(buffer.mutableSpan().data(), descriptionLength + 1, description, copyArgs);
+            va_end(copyArgs);
+        }
+
+        auto key = std::make_pair(pointer, static_cast<const void*>(name.data()));
+        auto value = std::make_pair(time, WTF::move(buffer));
 
         Locker locker { m_lock };
-        m_ongoingMarks.set(key, value);
+        m_ongoingMarks.set(key, WTF::move(value));
     }
 
     void endMark(const void* pointer, std::span<const char> name, const char* description, ...) WTF_ATTRIBUTE_PRINTF(4, 5)
     {
+        auto time = SYSPROF_CAPTURE_CURRENT_TIME;
+
         auto key = std::make_pair(pointer, static_cast<const void*>(name.data()));
         std::optional<TimestampAndString> value;
 
         {
             Locker locker { m_lock };
-
-            TimestampAndString v = m_ongoingMarks.take(key);
-            if (v.first)
-                value = WTF::move(v);
+            value = m_ongoingMarks.takeOptional(key);
         }
 
         if (value) {
             int64_t startTime = std::get<0>(*value);
-            const Vector<char>& description = std::get<1>(*value);
-            sysprof_collector_mark(startTime, SYSPROF_CAPTURE_CURRENT_TIME - startTime, m_processName, name.data(), description[0] ? description.span().data() : nullptr);
+            Vector<char>& buffer = std::get<1>(*value);
+
+            if (description && description[0] != '\0') {
+                va_list args, copyArgs;
+                va_start(args, description);
+                va_copy(copyArgs, args);
+
+                auto descriptionLength = vsnprintf(nullptr, 0, description, args);
+                va_end(args);
+
+                if (descriptionLength > 0) {
+                    bool needsSeparator = buffer.size() > 1;
+                    static constexpr auto separator = " | "_span;
+
+                    auto oldSize = buffer.size();
+                    buffer.resize(oldSize + (needsSeparator ? separator.size() : 0) + descriptionLength);
+
+                    auto span = buffer.mutableSpan().subspan(oldSize - 1);
+                    if (needsSeparator) {
+                        memcpySpan(span, separator);
+                        skip(span, separator.size());
+                    }
+                    vsnprintf(span.data(), descriptionLength + 1, description, copyArgs);
+                }
+                va_end(copyArgs);
+            }
+
+            sysprof_collector_mark(startTime, time - startTime, m_processName, name.data(), buffer.span().data());
         } else {
             va_list args;
             va_start(args, description);
-            sysprof_collector_mark_vprintf(SYSPROF_CAPTURE_CURRENT_TIME, 0, m_processName, name.data(), description, args);
+            sysprof_collector_mark_vprintf(time, 0, m_processName, name.data(), description, args);
             va_end(args);
         }
     }
--- a/Source/WebKit/UIProcess/WebPageProxy.cpp
+++ b/Source/WebKit/UIProcess/WebPageProxy.cpp
@@ -11830,11 +11830,11 @@ void WebPageProxy::mouseEventHandlingCompleted(std::optional<WebEventType> event
     }
 
 #if PLATFORM(GTK) || PLATFORM(WPE)
-    WTFEndSignpost(event.signpostIdentifier(), HandleMouseEvent);
+    WTFEndSignpost(event.signpostIdentifier(), HandleMouseEvent, "handled: %s", handled ? "yes" : "no");
     for (auto& coalescedEvent : event.coalescedEvents()) {
         if (coalescedEvent.signpostIdentifier() == event.signpostIdentifier())
             continue;
-        WTFEndSignpost(coalescedEvent.signpostIdentifier(), HandleMouseEvent);
+        WTFEndSignpost(coalescedEvent.signpostIdentifier(), HandleMouseEvent, "coalesced with: %" PRIuPTR, event.signpostIdentifier());
     }
 #endif
 


Source/WTF/wtf/glib/SysprofAnnotator.h
aperezdc: Here pick the number of characters actually formatted, then adjust the size of the `Vector`:


```suggestion
        auto formattedBytes = vsnprintf(buffer.mutableSpan().data(), buffer.size(), description, args);
        buffer.resize(formattedBytes + 1); // Include nul terminator.
```

This way, `buffer.size()` is the amount of actual data. Note that this does **not** change the amount of memory *reserved* for the `Vector`: only what it reports as the size. Then later we can keep adding to the format there: we will still have `buffer.capacity() - buffer.size()` bytes left to append the end description.

Source/WTF/wtf/glib/SysprofAnnotator.h
```diff
@@ -88,16 +88,29 @@ class SysprofAnnotator final {
...
-            sysprof_collector_mark(startTime, SYSPROF_CAPTURE_CURRENT_TIME - startTime, m_processName, name.data(), description[0] ? description.span().data() : nullptr);
+            const Vector<char>& beginMarkBuffer = std::get<1>(*value);
+
+            WTF_ALLOW_UNSAFE_BUFFER_USAGE_BEGIN
+            Vector<char> endMarkBuffer(1024);
+            va_list args;
+            va_start(args, description);
+            vsnprintf(endMarkBuffer.mutableSpan().data(), endMarkBuffer.size(), description, args);
+            va_end(args);
+            WTF_ALLOW_UNSAFE_BUFFER_USAGE_END
```
csaavedra: We are trying very hard not to have unsafe buffer usage, so if possible I'd advocate for an alternative that doesn't require `vsnprintf()`, unless that can't be avoided of course.

Source/WTF/wtf/glib/SysprofAnnotator.h
```diff
@@ -88,16 +88,29 @@ class SysprofAnnotator final {
...
-            if (v.first)
-                value = WTF::move(v);
+            value = m_ongoingMarks.takeOptional(key);
         }
 
         if (value) {
             int64_t startTime = std::get<0>(*value);
-            const Vector<char>& description = std::get<1>(*value);
-            sysprof_collector_mark(startTime, SYSPROF_CAPTURE_CURRENT_TIME - startTime, m_processName, name.data(), description[0] ? description.span().data() : nullptr);
+            const Vector<char>& beginMarkBuffer = std::get<1>(*value);
```
aperezdc: Rename this as `markBuffer` and remove the `const`:


```suggestion
            Vector<char>& markBuffer = std::get<1>(*value);
```

Source/WTF/wtf/glib/SysprofAnnotator.h
```diff
@@ -88,16 +88,29 @@ class SysprofAnnotator final {
...
+
+            WTF_ALLOW_UNSAFE_BUFFER_USAGE_BEGIN
+            Vector<char> endMarkBuffer(1024);
```
aperezdc: Now, here, if the `description` format specifier is empty, we do nothing. Otherwise, do something like this:

```suggestion
            va_list args;
            va_start(args, description);

            // Do not add the trailing mark description if there is none,
            // or if we do not have space for it.
            auto usedBytes = markBuffer.size() - 1;  // Ignore the trailing nul.
            auto availableBytes = markBuffer.capacity() - usedBytes;
            if (description && *description && availableBytes >= 6) {
                // Buy us some free space, as in “free beer” ^_^
                markBuffer.resize(markBuffer.capacity());
                auto remainingBuffer = markBuffer.mutableSpan().subspan(usedBytes);
                if (availableBytes == 6) {
                    // Append only " | ..." to indicate that there was a
                    // description for the end of the mark, but it did not
                    // fit.
                    static constexpr auto ellipsis = " | ...\0"_span;
                    memcpySpan(remainingBuffer, ellipsis);
                    usedBytes += ellipsis.size();
                } else {
                    // Append " | " into markBuffer.
                    static constexpr separator = " | "_span;
                    memcpySpan(remainingBuffer, separator);
                    skip(remainingBuffer, separator.size());
                    usedBytes += separator.size();
                    
                    // Format description directly into markBuffer.
                    usedBytes += vsnprintf(remainingBuffer.data(), remainingBuffer.size(), description, args);
                    usedBytes++;  // Nul terminator, not accounted by vsnprintf()
                }
            }
            va_end(args);
            
            // At this point, the markBuffer Vector contains the formatted
            // mark-start description, and if it fits and was provided, the
            // formatted mark-end description. Resizing here is not really
            // needed because it's zero-terminated, but here I'm illustrating
            // the point.
            markBuffer.resize(usedBytes);
            
            // Nothing else to do, we can send the mark to Sysprof =)
            sysprof_collector_mark(startTime, SYSPROF_CAPTURE_CURRENT_TIME - startTime, m_processName, name.data(), !markBuffer.isEmpty() && markBuffer[0] ? arkBuffer.span().data() : nullptr);
```

aperezdc (commented):
I think we can completely avoid additional heap allocations (for `endMarkBuffer`) by reusing the _same_ `Vector` where we already have stored the mark-start description. We are making the buffer for each mark-start 1 KiB and that is _lots_ (one could even advocate for a smaller size) so we can keep appending there.

Below I have outlined in the comments how this can be done; it requires being a bit careful to keep track of how many bytes we have already written at each step, in order doing linear scans to find the nul-terminator for the string. We can take advantage of the fact that _reducing_ the size of a `Vector` below its allocation size won't actually reallocate memory at all.

Source/WTF/wtf/glib/SysprofAnnotator.h
aperezdc: I am not sure what may be causing the unsafe buffer usage warnings, but it should not be `vsnprintf()`: we use it in other places without needing to suppress warnings.

Source/WTF/wtf/glib/SysprofAnnotator.h
obyknovenius: It's very efficient but has a couple of flaws:
- If formatted string doesn't fit into the buffer `vsnprintf()` returns the number of characters which would have been written to the final string if enough space had been available.
- It always adds " | " separator even if resulted string is empty.

obyknovenius:
@aperezdc I took some inspiration from your suggestion but implemented it a bit differently. `WTFBeginSignpost()`/`WTFEndSignpost()` can now handle description of any length. If total length doesn't exceed 1024 bytes they allocate only once.

Source/WTF/wtf/glib/SysprofAnnotator.h
```diff
@@ -47,61 +47,88 @@ class SysprofAnnotator final {
...
-            const Vector<char>& description = std::get<1>(*value);
-            sysprof_collector_mark(startTime, SYSPROF_CAPTURE_CURRENT_TIME - startTime, m_processName, name.data(), description[0] ? description.span().data() : nullptr);
+            Vector<char>& buffer = std::get<1>(*value);
+
+            va_list args, copyArgs;
+            va_start(args, format);
+            va_copy(copyArgs, args);
+
+            auto length = vsnprintf(nullptr, 0, format, args);
+            va_end(args);
```
aperezdc: This crossed my mind when I made the suggestion of reusing the same `Vector`/buffer; but I specifically did not write about it because:

* In some architectures `va_copy()` may actually do a memory copy.
* We are calling the formatting function twice instead of once, and formatting is notoriously slow (specially when involving floating point numbers: this is why we have `Source/WTF/wtf/{dtoa,fast_float}/` in the tree.
* We are preallocating 1 KiB for each mark, a full 1024 bytes: the console default width is 80 characters, a wide console has ~130. Are we going to fill eight full “wide console lines” for each mark? I highly doubt it.
* Again, this is performance-sensitive code, we want mark recording to be as performance-neutral as possble.

(The last point is why I wrote earlier that arguably allocating 1 KiB for each mark is a bit of overkill.)

Source/WTF/wtf/glib/SysprofAnnotator.h
```diff
@@ -47,61 +47,88 @@ class SysprofAnnotator final {
         return SYSPROF_CAPTURE_CURRENT_TIME + timeDelta.microsecondsAs<int64_t>();
     }
 
-    void instantMark(std::span<const char> name, const char* description, ...) WTF_ATTRIBUTE_PRINTF(3, 4)
+    void instantMark(std::span<const char> name, const char* format, ...) WTF_ATTRIBUTE_PRINTF(3, 4)
```
aperezdc: Could we please leave the argument name as “description” to avoid unneeded churn in the diff? The same goes for the over locations where the name is changed without any other name to the code.

If the rename is seen as important (not in my opinion), let's do it in a separate PR.

aperezdc (commented):
I think this still needs one last sping, thanks for working on it @obyknovenius!

Source/WTF/wtf/glib/SysprofAnnotator.h
```diff
@@ -88,16 +90,39 @@ class SysprofAnnotator final {
...
+            auto usedBytes = buffer.size() - 1;
+            auto availableBytes = buffer.capacity() - usedBytes;
+
+            buffer.resize(buffer.capacity());
+            auto destinationSpan = buffer.mutableSpan().subspan(usedBytes);
+
+            static constexpr auto separator = " | "_span;
+            if (usedBytes > 0) {
+                RELEASE_ASSERT(availableBytes > separator.size());
+
```
aperezdc: If there is no space left to add the separator, this will always fail the assertion and abort. Let's only append the separator when it fits:

```suggestion
            if (usedBytes > 0 && availableBytes > separator.size()) {
```

Source/WTF/wtf/glib/SysprofAnnotator.h
```diff
@@ -88,16 +90,39 @@ class SysprofAnnotator final {
...
+                memcpySpan(destinationSpan, separator);
```
aperezdc: Keeping the invariant that the “in progress” buffer is always null-terminated after each modification makes things easier to reason about.

Therefore, I would prefer to append `" | \0"_span` here (note the null-terminator), to ensure that the terminator is always present after each modification to the buffer...

Source/WTF/wtf/glib/SysprofAnnotator.h
```diff
@@ -88,16 +90,39 @@ class SysprofAnnotator final {
...
+            value = m_ongoingMarks.takeOptional(key);
         }
 
         if (value) {
             int64_t startTime = std::get<0>(*value);
-            const Vector<char>& description = std::get<1>(*value);
-            sysprof_collector_mark(startTime, SYSPROF_CAPTURE_CURRENT_TIME - startTime, m_processName, name.data(), description[0] ? description.span().data() : nullptr);
+            Vector<char>& buffer = std::get<1>(*value);
+
+            auto usedBytes = buffer.size() - 1;
```
aperezdc: I think we only want to go through the work of appending the mark-end message if there is a message. This should be all wrapped in a `description[0] != '\0'` check:


```suggestion
            if (description[0] != '\0') {
                auto usedBytes = buffer.size() - 1;
```

Source/WTF/wtf/glib/SysprofAnnotator.h
```diff
@@ -88,16 +90,39 @@ class SysprofAnnotator final {
...
+            va_list args;
+            va_start(args, description);
+            auto descriptionLength = vsnprintf(destinationSpan.data(), availableBytes, description, args);
+            va_end(args);
+
+            RELEASE_ASSERT(descriptionLength >= 0 && descriptionLength < static_cast<int>(availableBytes));
+
+            if (!descriptionLength)
+                buffer.mutableSpan()[usedBytes] = '\0';
+
```
aperezdc: ...and then this is not needed because `vsnprintf()` will add the null-terminator itself if there was anything to format.

```suggestion
```

aperezdc: Oh, wow! After re-reading this I now understood how the original code worked: if nothing was formatted by `vsprintf()` and it returned zero, then this re-adds the null-terminator _before_ the separator. That is counterintuitive 😰.

There was anyway a problem: when `availableBytes` is zero, `vsnprintf()` will not modify the buffer; but will return a non-zero result (the number of characters that would have been used for formatting), and then this code would _not_ add the null-terminator. Then the string without the terminator would be passed down to Sysprof and eventually we would see a crash.

Reminder: let's not call `vsnprintf()` when `availableBytes` is zero!

Source/WTF/wtf/glib/SysprofAnnotator.h
obyknovenius: In the previous review you raised a concert about the size of preallocated buffer for a mark. But because we can't know upfront what size is enough I've added `RELEASE_ASSERTS` to check that any message we use will fit. I expect that when somebody adds a message that exceeds the current size of the buffer it won't go unnoticed, and we can decide what to do - increase the size or shorten the message.

Source/WTF/wtf/glib/SysprofAnnotator.h
obyknovenius: I considered this idea but discarded it because there's always following `vsnprintf` call that always writes the null terminator. Adding it to the separator won't hurt but will make the code event more convoluted:
```
destinationSpan = destinationSpan.subspan(separator.size() - 1);
availableBytes -= (separator.size() - 1);
```
I assume that the final message fits the buffer or the assertion is hit if it's not.

Source/WTF/wtf/glib/SysprofAnnotator.h
obyknovenius: That's a good optimization. I'll add it.

Source/WTF/wtf/glib/SysprofAnnotator.h
obyknovenius: `RELEASE_ASSERT(descriptionLength >= 0 && descriptionLength < static_cast<int>(availableBytes));` fails if the message exceeds the buffer capacity. We should either increase the buffer size or use shorter message.

obyknovenius:
```
if (!descriptionLength)
    buffer.mutableSpan()[usedBytes] = '\0';
```
is needed to remove trailing separator if the message was effectively empty, e.g. [`endMark(nullptr, tracePointCodeName(code).spanIncludingNullTerminator(), "%s", "");`](https://github.com/WebKit/WebKit/blob/ff70c312f1bb9b293ae92c92fdbd864c4f0d0dff/Source/WTF/wtf/glib/SysprofAnnotator.h#L240)

Source/WTF/wtf/glib/SysprofAnnotator.h
aperezdc: > In the previous review you raised a concert about the size of preallocated buffer for a mark. But because we can't know upfront what size is enough I've added `RELEASE_ASSERTS` to check that any message we use will fit.

I think it's nasty to crash and stop a person's perf-debugging session just because, by chance, a string didn't fit into a buffer.

> I expect that when somebody adds a message that exceeds the current size of the buffer it won't go unnoticed, and we can decide what to do - increase the size or shorten the message.

And I expect that if the message is too long it will be truncated, maybe having an indication of the truncation, and that is why [already in my very first suggestion](https://github.com/WebKit/WebKit/pull/60199#discussion_r2911543433) there's this comment:

```cpp
                    // Append only " | ..." to indicate that there was a
                    // description for the end of the mark, but it did not
                    // fit.
```

We can discuss which approach is better in theory; but think about this: you are debugging something or analyzing performance with Sysprof, and suddenly WebKit crashes when doing a capture... now to continue working you need to edit the WebKit source code to change the buffer size, rebuild, and then continue where you left off.

Do you prefer to have some marks with an ellipsis (`...`) at the end of the label, or to have to rebuild? If you see the ellipsis and _absolutely_ need the missing information, you can rebuild as well; but if you don't need the truncated part, you don't get your flow broken 😉

Source/WTF/wtf/glib/SysprofAnnotator.h
aperezdc: > ```
> if (!descriptionLength)
>     buffer.mutableSpan()[usedBytes] = '\0';
> ```
> 
> is needed to remove trailing separator if the message was effectively empty, e.g. [`endMark(nullptr, tracePointCodeName(code).spanIncludingNullTerminator(), "%s", "");`](https://github.com/WebKit/WebKit/blob/ff70c312f1bb9b293ae92c92fdbd864c4f0d0dff/Source/WTF/wtf/glib/SysprofAnnotator.h#L240)

As written in my second comment: I understood that; but it's hard to reason about. Again, I think it's better to:

* Only add to the buffer things we know that will fit, and
* Always keep the buffer null-terminated every time it gets modified.

Source/WTF/wtf/glib/SysprofAnnotator.h
aperezdc: ...except that the `vsnprintf()` call may not format anything at all and leave the buffer without a null terminator (as I wrote [in this comment](https://github.com/WebKit/WebKit/pull/60199#discussion_r2921499195)).

You can “see” that behaviour here: https://godbolt.org/z/znx4qW1Ez

Source/WTF/wtf/glib/SysprofAnnotator.h
obyknovenius: Assuming that the current buffer size is correct and covers existing marks, only new marks can overflow. If we use ellipsis, the useful information can be cut off, and the developer may not even know why. The API doesn't explain that there's a limit. Perf-debugging session is broken, but now you also need to understand where to increase the buffer size. When explicit crash can at least give you a hint.

Source/WTF/wtf/glib/SysprofAnnotator.h
```diff
@@ -65,43 +65,79 @@ class SysprofAnnotator final {
...
-        va_list args;
-        va_start(args, description);
-        vsnprintf(buffer.mutableSpan().data(), buffer.size(), description, args);
-        va_end(args);
-        WTF_ALLOW_UNSAFE_BUFFER_USAGE_END
 
-        auto value = std::make_pair(SYSPROF_CAPTURE_CURRENT_TIME, WTF::move(buffer));
+        if (!description || description[0] == '\0') {
+            buffer.resize(1);
+            buffer.mutableSpan()[0] = '\0';
```
aperezdc: Nit:


```suggestion
            buffer.append('\0');
```

Source/WTF/wtf/glib/SysprofAnnotator.h
```diff
@@ -65,43 +65,79 @@ class SysprofAnnotator final {
...
+                    bool needsSeparator = buffer.size() > 1;
+                    static constexpr auto separator = " | "_span;
+
+                    auto oldSize = buffer.size();
+                    buffer.resize(oldSize + (needsSeparator ? separator.size() : 0) + descriptionLength);
+
+                    auto destinationSpan = buffer.mutableSpan().subspan(oldSize - 1);
+                    if (needsSeparator) {
+                        memcpySpan(destinationSpan, separator);
+                        destinationSpan = destinationSpan.subspan(separator.size());
```
aperezdc: Small nit... we have a helper for this:


```suggestion
                        skip(destinationSpan, separator.size());
```

aperezdc (approved):
LGTM with a couple of small nits that you may want to take a look into before landing.

For the record, after some discussion with @obyknovenius, we agreed to go with the double call to `[v]snprintf()` to determine how much space is needed in the buffers. He noticed that Sysprof itself does the same when it takes care of formatting &mdash;e.g. when using `sysprof_collector_mark_vprintf()`&mdash; so given that Sysprof does that, we would not be doing worse when we take care of the formatting in WebKit before passing the preformatted strings to Sysprof.

Source/WTF/wtf/glib/SysprofAnnotator.h
csaavedra: I have removed these guards locally and the code compiles without warnings with clang-21,  @obyknovenius why were they needed?

csaavedra: Nevermind, I realized I was working on an outdated main.

PR #60231, opened 2026-03-10
git-webkit: add `create-bug` and `tracker-metadata` subcommands

ddkilzer:
#### 00af9f0cd69117d359726aace2d4490d29bf1e44
<pre>
git-webkit: add `create-bug` and `tracker-metadata` subcommands
&lt;<a href="https://bugs.webkit.org/show_bug.cgi?id=309535">https://bugs.webkit.org/show_bug.cgi?id=309535</a>&gt;
&lt;<a href="https://rdar.apple.com/problem/172135894">rdar://problem/172135894</a>&gt;

Reviewed by Pascoe and Jonathan Bedard.

Add two subcommands to provide a scriptable bug tracker interface,
primarily to support external tooling such as LLM-based skills.

`git-webkit create-bug` creates a new bug on the configured tracker.
It optionally links the bug to an existing radar or auto-creates a
new radar for it, and determines whether the Security product and
component are required based on the provided radar and the title and
description.

`git-webkit tracker-metadata` queries the tracker for its available
products, components, and keywords -- including descriptions -- so
external tools can discover and validate choices before calling
create-bug without extra round-trips.

Both `create-bug` and the existing `branch` subcommand share radar
URL parsing through the new `radar.Tracker.parse_id()` classmethod,
which uses `urllib.parse.urlparse()` to handle all radar URL forms
(rdar://, radar://, <a href="https://rdar.apple.com/)">https://rdar.apple.com/)</a> and extract numeric
IDs.  This replaces the `RES` compiled-regex list that
`from_string()` previously relied on, and eliminates the duplicate
`urlparse` logic that `parse_radar_arg()` previously contained.

Covered by newly added unit tests.

* Tools/Scripts/libraries/webkitbugspy/webkitbugspy/bugzilla.py:
(Tracker.set):
- Add `cc` parameter.
(Tracker.SECURITY_KEYWORDS):
(Tracker.check_security_keywords):
(Tracker.prompt_security_classification):
(Tracker.classify_from_radar):
(Tracker.valid_keywords):
- Return a dict mapping keyword names to their descriptions rather
  than a list, so callers can surface contextual help.
* Tools/Scripts/libraries/webkitbugspy/webkitbugspy/mocks/bugzilla.py:
* Tools/Scripts/libraries/webkitbugspy/webkitbugspy/radar.py:
(Tracker.parse_id):
- Extract radar URL parsing into a shared classmethod that accepts
  rdar://, radar://, and <a href="https://rdar.apple.com/">https://rdar.apple.com/</a> forms with optional
  angle brackets and ampersand-separated multi-ID URLs.  Return the
  numeric ID as a string (single) or list of strings (multiple), or
  None on no match.
(Tracker.from_string):
- Delegate to parse_id() instead of iterating the RES regex list.
  For multi-ID URLs, take the first ID to preserve existing behavior.
(Tracker.RES):
- Remove the compiled-regex list, now superseded by parse_id().
* Tools/Scripts/libraries/webkitbugspy/webkitbugspy/tests/bugzilla_security_unittest.py: Add.
(TestBugzillaSecurityKeywords):
(TestBugzillaClassifyFromRadar):
* Tools/Scripts/libraries/webkitbugspy/webkitbugspy/tests/radar_unittest.py:
(TestRadarParseId):
- Cover all single-ID URL forms, angle brackets, ampersand multi-ID
  URLs, bare integers, and invalid inputs.
(TestRadarFromString):
- Regression tests verifying from_string() behavior is unchanged
  after the parse_id() refactor.
* Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/__init__.py:
(main):
* Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/branch.py:
(Branch.main):
* Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py: Add.
(CreateBug):
(CreateBug.parser):
(CreateBug.parse_radar_arg):
- Delegate URL parsing to radar.Tracker.parse_id() instead of
  reimplementing urlparse logic.  Accept <a href="https://rdar.apple.com/N">https://rdar.apple.com/N</a>
  in addition to rdar:// and radar:// forms.  Normalize all outputs
  to <a href="https://rdar.apple.com/N">rdar://N</a> format.
(CreateBug.main):
(CreateBug._post_create_operations):
- Filter the radar importer from the explicit --cc list so cc_radar()
  can add it in the correct order (after the &lt;<a href="https://rdar.apple.com/N">rdar://N</a>&gt; comment and
  InRadar keyword), preventing a duplicate radar from being created.
* Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/tracker_metadata.py: Add.
(TrackerMetadata):
(TrackerMetadata.parser):
(TrackerMetadata.main):
(TrackerMetadata.get_tracker_metadata):
(TrackerMetadata.output_result):
* Tools/Scripts/libraries/webkitscmpy/webkitscmpy/test/create_bug_unittest.py: Add.
(TestCreateBug):
(TestCreateBug.test_radar_importer_not_double_cced):
(TestCreateBugRadarArgValidation):
- Update <a href="https://rdar.apple.com/problem/N">rdar://problem/N</a> tests to expect normalized <a href="https://rdar.apple.com/N">rdar://N</a> output.
  Move <a href="https://rdar.apple.com/N">https://rdar.apple.com/N</a> from invalid to valid.
* Tools/Scripts/libraries/webkitscmpy/webkitscmpy/test/tracker_metadata_unittest.py: Add.
(TestTrackerMetadata):

Canonical link: <a href="https://commits.webkit.org/311151@main">https://commits.webkit.org/311151@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/06bc2dcbcc3da81cea15fecd2ad7ec1f6b1a81f5

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/156186 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/29521 "Built successfully") | [  ~~🛠 mac~~](https://ews-build.webkit.org/#/builders/168/builds/22703 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/165007 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/110264 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/158057 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/29654 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/29524 "Built successfully") | [  ~~🧪 wpe-wk2~~](https://ews-build.webkit.org/#/builders/34/builds/120941 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/110264 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/159144 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/162/builds/23143 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/140257 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-wpe~~](https://ews-build.webkit.org/#/builders/41/builds/101616 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | ⏳ 🛠 vision-apple 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/155505 "Passed tests") | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/22223 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/20386 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/12779 "Built successfully") | | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/131884 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-wk1~~](https://ews-build.webkit.org/#/builders/167/builds/18088 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/167486 "Built successfully") | | 
| | | [  ~~🧪 mac-wk2~~](https://ews-build.webkit.org/#/builders/169/builds/19701 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/1/builds/129060 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| [✅ 🧪 services](https://ews-build.webkit.org/#/builders/28/builds/155584 "Passed tests") | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/29122 "Built successfully") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/24449 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-gtk~~](https://ews-build.webkit.org/#/builders/21/builds/129180 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/29044 "Built successfully") | [  ~~🧪 mac-wk2-stress~~](https://ews-build.webkit.org/#/builders/8/builds/139882 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/86811 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/23774 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/23999 "Passed tests") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/16681 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/28753 "Built successfully") | | | | 
| | [  ~~🛠 tv-sim~~](https://ews-build.webkit.org/#/builders/157/builds/28280 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [  ~~🛠 watch~~](https://ews-build.webkit.org/#/builders/163/builds/28508 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/28404 "Built successfully") | | | |

### DIFF
--- a/Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
+++ b/Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
@@ -1,0 +1,252 @@
+# Copyright (C) 2026 Apple Inc. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1.  Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+# 2.  Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in the
+#     documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS BE LIABLE FOR
+# ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import sys
+
+from .command import Command
+
+from webkitbugspy import Tracker, bugzilla, radar
+from webkitscmpy import log
+
+
+class CreateBug(Command):
+    name = 'create-bug'
+    aliases = []
+    help = 'Create a new bug on the configured bug tracker'
+
+    @classmethod
+    def parser(cls, parser, loggers=None):
+        parser.add_argument(
+            '--title', dest='title', type=str, required=True,
+            help='Bug title/summary',
+        )
+        parser.add_argument(
+            '-F', '--file', dest='description_file', type=str, default=None,
+            help='Path to file containing bug description',
+        )
+        parser.add_argument(
+            '--component', dest='component', type=str, default=None,
+            help='WebKit component (prompts if not specified)',
+        )
+        parser.add_argument(
+            '--project', dest='project', type=str, default=None,
+            help='Product name (prompts if not specified)',
+        )
+        parser.add_argument(
+            '--keywords', dest='keywords', type=str, default=None,
+            help='Comma-separated list of keywords',
+        )
+        parser.add_argument(
+            '--cc', dest='cc', type=str, default=None,
+            help='Comma-separated list of email addresses to CC',
+        )
+        parser.add_argument(
+            '--assignee', dest='assignee', type=str, default=None,
+            help='Username to assign bug to (defaults to current user)',
+        )
+        parser.add_argument(
+            '--depends-on', dest='depends_on', type=int, action='append', default=None,
+            help='Bug ID that this bug depends on (can be specified multiple times)',
+        )
+        parser.add_argument(
+            '--blocks', dest='blocks', type=int, action='append', default=None,
+            help='Bug ID that this bug blocks (can be specified multiple times)',
+        )
+        parser.add_argument(
+            '--see-also', dest='see_also', action='append', default=None,
+            help='URL to add to See Also (can be specified multiple times)',
+        )
+        parser.add_argument(
+            '--radar', dest='radar', type=str, default=None,
+            help='Radar ID to associate when creating this bug (e.g., rdar://12345678).',
+        )
+
+    @classmethod
+    def parse_radar_arg(cls, value):
+        """Validate and normalize a --radar argument to a single rdar:// string.
+
+        Accepts a bare integer ID, rdar://NNNNN, rdar://problem/NNNNN,
+        or https://rdar.apple.com/NNNNN.
+        Raises ValueError for multiple IDs or unrecognized formats.
+        Returns the normalized 'rdar://NNNNN' string.
+        """
+        value = value.strip()
+
+        # Strip optional surrounding angle brackets (e.g., <rdar://N>)
+        if value.startswith('<') and value.endswith('>'):
+            value = value[1:-1].strip()
+
+        # Bare integer ID
+        if value.isdigit():
+            return 'rdar://{}'.format(value)
+
+        # Space or comma — multiple IDs
+        if ' ' in value or ',' in value:
+            raise ValueError('only one radar ID is accepted')
+
+        # Delegate URL parsing to radar.Tracker.parse_id()
+        result = radar.Tracker.parse_id(value)
+        if result is None:
+            raise ValueError("unrecognized format; expected a numeric ID or rdar://NNNNN")
+        if isinstance(result, list):
+            raise ValueError('only one radar ID is accepted')
+        return 'rdar://{}'.format(result)
+
+    @classmethod
+    def main(cls, args, repository, **kwargs):
+        tracker = Tracker.instance()
+        if not tracker:
+            sys.stderr.write("No bug tracker configured\n")
+            return 1
+
+        # Ensure credentials are available
+        if getattr(tracker, 'credentials', None):
+            tracker.credentials(required=True, validate=True)
+
+        # Read description from file
+        description = None
+        if args.description_file:
+            try:
+                with open(args.description_file, 'r') as f:
+                    description = f.read()
+            except IOError as e:
+                sys.stderr.write("Failed to read description file '{}': {}\n".format(
+                    args.description_file, e))
+                return 1
+
+        if not description:
+            sys.stderr.write("Bug description is required (use -F/--file)\n")
+            return 1
+
+        # Start with user-specified project/component
+        project = args.project
+        component = args.component
+
+        # Check for radar and apply security override if redacted
+        radar_issue = None
+        security_forced = False
+        if args.radar:
+            try:
+                radar_str = cls.parse_radar_arg(args.radar)
+            except ValueError as e:
+                sys.stderr.write("Invalid --radar value '{}': {}\n".format(args.radar, e))
+                return 1
+            radar_issue = Tracker.from_string(radar_str)
+            if radar_issue:
+                project, component, security_forced = bugzilla.Tracker.classify_from_radar(
+                    radar_issue, project, component,
+                )
+
+        # If not forced to Security, check for security keywords in title + description
+        if not security_forced:
+            matches = bugzilla.Tracker.check_security_keywords(args.title, description)
+            if matches:
+                project, component = bugzilla.Tracker.prompt_security_classification(
+                    matches, project, component,
+                )
+
+        # Parse keywords
+        keywords = None
+        if args.keywords:
+            keywords = [k.strip() for k in args.keywords.split(',')]
+
+        # Create the bug (tracker.create prompts for project/component if not specified)
+        try:
+            issue = tracker.create(
+                title=args.title,
+                description=description,
+                project=project,
+                component=component,
+                keywords=keywords,
+            )
+        except Exception as e:
+            sys.stderr.write("Failed to create bug: {}\n".format(e))
+            return 1
+
+        if not issue:
+            sys.stderr.write("Failed to create bug\n")
+            return 1
+
+        print("Created bug: {}".format(issue.link))
+
+        # Post-creation operations
+        cls._post_create_operations(issue, tracker, args, radar_issue=radar_issue)
+
+        return 0
+
+    @classmethod
+    def _post_create_operations(cls, issue, tracker, args, radar_issue=None):
+        """Handle CC, assignee, see-also, depends-on, blocks, and radar linking after bug creation."""
+
+        # Handle assignee
+        if args.assignee:
+            try:
+                assignee_user = tracker.user(username=args.assignee)
+                if assignee_user:
+                    tracker.set(issue, assignee=assignee_user)
+                    print("Assigned to: {}".format(args.assignee))
+            except Exception as e:
+                log.warning("Failed to assign: {}".format(e))
+
+        # Handle CC list
+        if args.cc:
+            cc_list = [e.strip() for e in args.cc.split(',')]
+            # Remove the radar importer if present — cc_radar() will handle it
+            # to ensure the <rdar://N> comment and InRadar keyword come first.
+            if isinstance(tracker, bugzilla.Tracker) and getattr(tracker, 'radar_importer', None):
+                importer = tracker.radar_importer
+                importer_ids = set(filter(None, [importer.username, importer.email] + list(importer.emails)))
+                cc_list = [cc for cc in cc_list if cc not in importer_ids]
+            if cc_list:
+                try:
+                    tracker.set(issue, cc=cc_list)
+                    print("Added CC: {}".format(', '.join(cc_list)))
+                except Exception as e:
+                    log.warning("Failed to add CC: {}".format(e))
+
+        # Handle See Also
+        if args.see_also:
+            try:
+                tracker.set(issue, see_also=args.see_also)
+                print("Added See Also: {}".format(', '.join(args.see_also)))
+            except Exception as e:
+                log.warning("Failed to add See Also: {}".format(e))
+
+        # Handle relationships
+        for attr, label, apply in [
+            ('depends_on', 'depends on', lambda bug_id: tracker.relate(issue, depends_on=tracker.issue(bug_id))),
+            ('blocks', 'blocks', lambda bug_id: tracker.relate(issue, blocks=tracker.issue(bug_id))),
+        ]:
+            for bug_id in getattr(args, attr) or []:
+                try:
+                    apply(bug_id)
+                    print("Set {}: Bug {}".format(label, bug_id))
+                except Exception as e:
+                    log.warning("Failed to set {} {}: {}".format(attr, bug_id, e))
+
+        # Link to radar: post <rdar://N> comment, add InRadar keyword, CC importer.
+        # cc_radar handles the safe ordering to avoid duplicate radar creation.
+        if isinstance(tracker, bugzilla.Tracker):
+            try:
+                tracker.cc_radar(issue, radar=radar_issue)
+            except Exception as e:
+                log.warning("Failed to cc radar importer: {}".format(e))
--- a/Tools/Scripts/libraries/webkitbugspy/webkitbugspy/bugzilla.py
+++ b/Tools/Scripts/libraries/webkitbugspy/webkitbugspy/bugzilla.py
@@ -48,6 +48,82 @@ class Tracker(GenericTracker):
     NAME = 'Bugzilla'
     DEFAULT_TIMEOUT = 30
 
+    # Security keywords to detect in title/description.
+    # These trigger a prompt to use Security product.
+    SECURITY_KEYWORDS = [
+        # Web platform security
+        r'\bXSS\b', r'cross[- ]?site[- ]?scripting',
+        r'\bCSRF\b', r'cross[- ]?site[- ]?request[- ]?forgery',
+        r'SQL[- ]?injection', r'command[- ]?injection', r'code[- ]?injection',
+        r'same[- ]?origin', r'\bCORS\b',
+        # Memory safety
+        r'use[- ]?after[- ]?free', r'\bUAF\b',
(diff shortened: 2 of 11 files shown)


Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/bugzilla_metadata.py
```diff
@@ -0,0 +1,183 @@
...
+import sys
+from datetime import datetime
+
+from .command import Command
+
+from webkitbugspy import Tracker, bugzilla
+from webkitscmpy import log
+
+
+class BugzillaMetadata(Command):
```
JonWBedard: Couple thoughts on this class.

First, no reason to limit this to Bugzilla (it should work for any configured bug tracker).

Second, this command (which is basically just "fetch-bug") should be mostly iterating over the properties of `Issue`. I think it's reasonable for a V1 of this command to omit some data that we don't think is needed. I would recommend having a class variable called "PROPERTIES" which defines something like:
```
PROPERTIES = ['title', 'link', 'state', 'products', 'components', 'keywords'...]
```
and then you could have a `parser` that does something like this:
```
def parser(cls, parser, loggers=None):
    ...
    parser.add_argument(
            '--properties',,
            dest='properties', action='append', default=None,
            help='List of properties to output', choices=cls.PROPERTIES,
        )
```

Then your program logic could iterate through all the defined properties. I think the advantage of this architecture would be that adding new properties is trivial, since I'm sure there will be more properties folks want to add in the future.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
```diff
@@ -0,0 +1,289 @@
...
+
+
+class CreateBug(Command):
+    name = 'create-bug'
+    aliases = []
+    help = 'Create a new bug on the configured Bugzilla tracker'
+
+    # Security keywords to detect in title/description
+    # These trigger a prompt to use Security product
+    SECURITY_KEYWORDS = [
```
JonWBedard: Shouldn't these belong to the `bugzilla.py` class? Seems like the logic to trigger a security classification should apply to bugs filed via `git-webkit` too.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
```diff
@@ -0,0 +1,289 @@
...
+
+    @classmethod
+    def main(cls, args, repository, **kwargs):
+        tracker = Tracker.instance()
+        if not tracker:
+            sys.stderr.write("No Bugzilla tracker configured\n")
+            return 1
+
+        if not isinstance(tracker, bugzilla.Tracker):
+            sys.stderr.write("This command only works with Bugzilla trackers\n")
```
JonWBedard: Seems like this would work for any bug tracker?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
```diff
@@ -0,0 +1,289 @@
...
+        if not description:
+            sys.stderr.write("Bug description is required (use -F/--file)\n")
+            return 1
+
+        # Start with user-specified project/component
+        project = args.project
+        component = args.component
+        security_forced = False
+
+        # Check for radar and apply security override if redacted
```
JonWBedard, line 144: I actually think this logic should be owned by `bugzilla` and apply to `webkitscmpy`'s `branch` command as well (which is where we create bugs in `git-webkit`)

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
```diff
@@ -0,0 +1,289 @@
...
+            try:
+                assignee_user = tracker.user(username=args.assignee)
+                if assignee_user:
+                    tracker.set(issue, assignee=assignee_user)
+                    print("Assigned to: {}".format(args.assignee))
+            except Exception as e:
+                log.warning("Failed to assign: {}".format(e))
+
+        # Handle CC list
+        if args.cc:
```
JonWBedard, line 211: Should always cc the radar importer.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
```diff
@@ -0,0 +1,289 @@
...
+            cc_list = [e.strip() for e in args.cc.split(',')]
+            try:
+                tracker.set(issue, cc=cc_list)
+                print("Added CC: {}".format(', '.join(cc_list)))
+            except Exception as e:
+                log.warning("Failed to add CC: {}".format(e))
+
+        # Handle See Also
```
JonWBedard, line 226: Seems like `see_also`, `depends_on` and `blocks` should be a for-loop or function, since they all share the same implementation.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/test/bugzilla_metadata_unittest.py
```diff
@@ -0,0 +1,231 @@
...
+            ))
+        # Verify JSON output
+        output = captured.stdout.getvalue()
+        data = json.loads(output)
+        self.assertIn('products', data)
+        self.assertIn('keywords', data)
+
+    def test_metadata_text_output(self):
+        """Test --format text output"""
+        with OutputCapture() as captured, mocks.local.Git(self.path), bmocks.Bugzilla(
```
burg: This test case is wild. :D

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
burg: is this blocking or can be fixed in a follow-on?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
```diff
@@ -0,0 +1,289 @@
...
+            return 1
+
+        # Start with user-specified project/component
+        project = args.project
+        component = args.component
+        security_forced = False
+
+        # Check for radar and apply security override if redacted
+        radar_issue = None
+        if args.radar:
```
burg, line 147: Help message says that the radar and bugzilla get linked, but in actuality the radar is used to grab the security  classification. Can it also update the bug with a <rdar://nnnn> comment?

If possible would be great to have a test case for the --radar argument.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
```diff
@@ -0,0 +1,289 @@
...
+            '--keywords', dest='keywords', type=str, default=None,
+            help='Comma-separated list of keywords',
+        )
+        parser.add_argument(
+            '--cc', dest='cc', type=str, default=None,
+            help='Comma-separated list of email addresses to CC',
+        )
+        parser.add_argument(
+            '--assignee', dest='assignee', type=str, default=None,
+            help='Email address to assign bug to (defaults to current user)',
```
burg: This says email address, but it takes a username. Can you fix the help string?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
burg: Also, there's no validation for the format of --radar argument.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
burg: If we don't comment the linked radar and add InRadar before cc'ing the importer, two radars will be created. I don't think we want that? Not sure how to factor that into this command.

burg (changes requested):
This looks good overall, I would like to address the issues and add test for --radar before checking in though.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/bugzilla_metadata.py
ddkilzer: I renamed this command from "bugzilla-metadata" to "tracker-metadata".  The intent of this subcommand is to get a list of products and components (Bugzilla) or components and versions (Radar) with descriptions so a tool may present a list of components to choose from for a new bug, or an LLM may make intelligent suggestions based on a bug description.  (Similar for keywords.)

I think I preserved the multi-tracker architecture, but I did not implement the radar part of this, so I am not sure if I did it correctly.  Does some code still need to move out of `tracker_metadata.py` into a `bugzilla_tracker.py` submodule?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
ddkilzer: Done. Moved `SECURITY_KEYWORDS`, `check_security_keywords()`, `prompt_security_classification()`, and `classify_from_radar()` into `webkitbugspy/bugzilla.Tracker`. Both `create-bug` and `branch` now call the shared methods.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
ddkilzer: Done. Removed this guard, but there is a minor `isinstance()` check left in `_post_create_operations()`. Is that a concern?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
ddkilzer: Done. Moved the radar-based security classification into `bugzilla.Tracker.classify_from_radar()` and the keyword scan into `check_security_keywords()` / `prompt_security_classification()`. Both `create-bug` and `branch` now delegate to these shared methods.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
ddkilzer: Done. `_post_create_operations()` now calls `tracker.cc_radar(issue, radar=radar_issue)`, which handles the correct ordering — posts the `<rdar://N>` comment, adds the `InRadar` keyword, then CCs the importer — to avoid triggering duplicate radar creation.

Additionally, if the caller explicitly passes the importer address via `--cc`, it is filtered out before the explicit CC call so the importer is not double-CCed. Preventing the double-CC added an `isinstance()` check, though. Is there a "general" rule that adding `isinstance()` is a code smell, and should be extracted into a tracker-specific method?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
ddkilzer: Done. `depends_on` and `blocks` are now a single loop using lambdas to capture the differing `tracker.relate()` calls. `see_also` is left separate since it uses `tracker.set()` and operates on the full list rather than individual IDs.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/test/bugzilla_metadata_unittest.py
ddkilzer: The subcommand was redesigned as `tracker-metadata` and the old test file was replaced with `tracker_metadata_unittest.py`. Is there a specific concern that you wanted addressed?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
ddkilzer: Done. The `--radar` help text has been updated. The argument is now validated with `urllib.parse` (bare IDs and `rdar://` URL forms accepted; multiple IDs rejected with a clear error). The `<rdar://N>` comment is posted via `cc_radar()` as part of the radar linking flow. Tests added for both the validation and the `--radar` integration path.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
ddkilzer: Done. Help string updated to say "username" instead of "email address".

pascoej (approved):

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
JonWBedard: Adding an `isinstance()` check is a bit of a code-smell, but the sort of check you're doing here is exactly what I would expect we use it for. The question I always ask myself is "is there a way to express this idea in a generic way". In this case, there clearly isn't (it's weird that CCing a specific email _also_ means "import" in Bugzilla, and Radar doesn't have a similar concept, nor would any other hypothetical bug tracker have that same concept)

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
JonWBedard, line 84: This looks pretty close to `radar.py`'s `Tracker.from_string`. I'm find with keeping this function around, but I _think_ it might be duplicative.

JonWBedard (approved):

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/create_bug.py
ddkilzer: Refactored (now-shared) parsing code into `radar.Tracker.parse_id()` method.  Thanks!

PR #60299, opened 2026-03-10
Panning a PDF frequently gets into a state where it rubber bands back to where you started

aj062:
#### 8828f24e7c4da08c591a057765bcdfed5bb05f00
<pre>
Panning a PDF frequently gets into a state where it rubber bands back to where you started
<a href="https://bugs.webkit.org/show_bug.cgi?id=309596">https://bugs.webkit.org/show_bug.cgi?id=309596</a>
<a href="https://rdar.apple.com/156854435">rdar://156854435</a>

Reviewed by Abrar Rahman Protyasha.

When a full-frame PDF plugin on iOS delegates scrolling to the main frame,
the plugin&apos;s own UIScrollView should have no scrollable area.
However, expandedIntSize rounding in contentsSize() can produce a width 1 pixel
larger than the plugin&apos;s visible width. This makes the plugin&apos;s UIScrollView
horizontally scrollable, causing it to capture pan gestures and rubber-band instead
of letting the main frame handle horizontal scrolling.

This is a regression from 290778@main (d8546c304004), which removed the explicit CSS
width synchronization between contentsSize() and the plugin element. Without that
synchronization, the rounding difference between expandedIntSize (ceilf) and the
CSS layout width becomes observable.

Fix by setting contentsSize().width() to the plugin&apos;s visible width when
shouldSizeToFitContent() is true, since in this mode the contents width should
always equal the plugin width.

* Source/WebKit/WebProcess/Plugins/PDF/UnifiedPDF/UnifiedPDFPlugin.mm:
(WebKit::UnifiedPDFPlugin::contentsSize const):
* Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm:
(TestWebKitAPI::TEST): Added an API test.

Canonical link: <a href="https://commits.webkit.org/309264@main">https://commits.webkit.org/309264@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/5faa26cacf39794c4d278e7c7b9d6d5d2f926c15

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/150029 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/22755 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/16348 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/158740 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/103463 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/555bf932-64ac-4405-a747-b3d3080fe1e9) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/151902 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/23202 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/22880 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/115732 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/82214 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/d8dbfc23-f077-4003-a333-ad191983a872) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/152989 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/17839 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/134597 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/96461 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/28b1c674-fc3b-46a6-ab6b-ab83652b7dee) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/16939 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/14890 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/6586 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/126554 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/12521 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/161214 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/4305 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/14073 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/123733 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/22556 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/18937 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/123935 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/33667 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/22561 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/134316 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/78805 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/19079 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/11072 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/22161 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/85989 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/21891 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/22043 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/21949 "Built successfully") | | | |

### DIFF
--- a/Source/WebKit/WebProcess/Plugins/PDF/UnifiedPDF/UnifiedPDFPlugin.mm
+++ b/Source/WebKit/WebProcess/Plugins/PDF/UnifiedPDF/UnifiedPDFPlugin.mm
@@ -1432,7 +1432,18 @@ IntSize UnifiedPDFPlugin::contentsSize() const
 
     auto size = m_documentLayout.scaledContentsSize();
     size.scale(m_scaleFactor);
-    return expandedIntSize(size);
+    auto result = expandedIntSize(size);
+
+    // In sizeToFitContent mode (iOS full-frame PDF), the contents width is
+    // the plugin's visible width. Use it directly to avoid a rounding artifact
+    // where expandedIntSize rounds up by 1 pixel, making the plugin's own
+    // UIScrollView horizontally scrollable and causing rubber-banding.
+    if (shouldSizeToFitContent() && m_size.width() > 0) {
+        ASSERT(std::abs((result - m_size).width()) <= 1);
+        result.setWidth(m_size.width());
+    }
+
+    return result;
 }
 
 unsigned UnifiedPDFPlugin::heightForPageAtIndex(PDFDocumentLayout::PageIndex pageIndex) const
--- a/Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
+++ b/Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
@@ -1187,6 +1187,32 @@ UNIFIED_PDF_TEST(ScrollPositionAfterChangeToTwoUpContinuous)
     EXPECT_EQ([webView scrollView].contentOffset, CGPointZero);
 }
 
+UNIFIED_PDF_TEST(PluginScrollViewIsNotHorizontallyScrollableForFullFramePDF)
+{
+    // Use width 391 specifically: for test.pdf (page width 129.6pt, contentWidth 161.6),
+    // the layout scale 391/161.6 produces a scaled width of ~391.000031 in float32,
+    // which expandedIntSize rounds up to 392 — a 1-pixel mismatch that makes the
+    // plugin's UIScrollView horizontally scrollable without the fix.
+    RetainPtr webView = adoptNS([[TestWKWebView alloc] initWithFrame:CGRectMake(0, 0, 391, 800) configuration:configurationForWebViewTestingUnifiedPDF().get()]);
+    RetainPtr request = [NSURLRequest requestWithURL:[NSBundle.test_resourcesBundle URLForResource:@"test" withExtension:@"pdf"]];
+    [webView synchronouslyLoadRequest:request.get()];
+
+    __block bool stable = false;
+    [webView _doAfterNextStablePresentationUpdate:^{
+        stable = true;
+    }];
+    TestWebKitAPI::Util::run(&stable);
+
+    RetainPtr childScrollView = dynamic_objc_cast<UIScrollView>([webView wkFirstSubviewWithClass:NSClassFromString(@"WKChildScrollView")]);
+    ASSERT_NOT_NULL(childScrollView);
+
+    CGSize contentSize = [childScrollView contentSize];
+    CGRect bounds = [childScrollView bounds];
+    EXPECT_GT(contentSize.width, 0);
+    EXPECT_GT(bounds.size.width, 0);
+    EXPECT_LE(contentSize.width, bounds.size.width);
+}
+
 #endif // PLATFORM(IOS_FAMILY)
 
 #if HAVE(UIKIT_WITH_MOUSE_SUPPORT)


Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
```diff
@@ -1420,6 +1420,32 @@ static void checkKeyboardScrollability(TestWKWebView *webView)
 }
 #endif
 
+#if PLATFORM(IOS_FAMILY)
+UNIFIED_PDF_TEST(PluginScrollViewIsNotHorizontallyScrollableForFullFramePDF)
+{
+    RetainPtr webView = adoptNS([[TestWKWebView alloc] initWithFrame:CGRectMake(0, 0, 400, 800) configuration:configurationForWebViewTestingUnifiedPDF().get()]);
```
hortont424: Does this test actually fail in the "before" case?

Source/WebKit/WebProcess/Plugins/PDF/UnifiedPDF/UnifiedPDFPlugin.mm
```diff
@@ -1432,7 +1432,19 @@ static String mutationObserverNotificationString()
...
+
+    // When sizing to fit content (iOS full-frame PDF), the layout scale is
+    // computed so scaled content width equals the available width. Floating-point
+    // arithmetic can produce a value slightly above the available width, which
+    // expandedIntSize rounds up by 1 pixel. This makes the plugin's own
+    // UIScrollView horizontally scrollable, capturing pan gestures and causing
+    // rubber-banding instead of letting the main frame handle scrolling.
+    // Clamp width to the plugin's visible width to eliminate this artifact.
+    if (shouldSizeToFitContent() && m_size.width() > 0)
+        result.setWidth(std::min(result.width(), m_size.width()));
```
hortont424: Is there a reason to be so precious with the fix? In sizeToFitContents mode, shouldn't our width always be the plugin's width? (If so, maybe the fix can be simpler, and if not, how is sometimes clamping it to the plugin's width ok?)

aprotyas: I need to go back in time 1 year, but I think there was a reason why we didn't just set to plugin width.

aprotyas: (thinking...)

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
aj062: @hortont424 Good catch!
This test didn't actually fail in the "before" case. Just updated it, it now fails in the "before" case and passes in "after" case.

Source/WebKit/WebProcess/Plugins/PDF/UnifiedPDF/UnifiedPDFPlugin.mm
aj062: Looking at the history, the original `sizeToFitContentsIfNeeded()` (from  [288932@main](https://commits.webkit.org/288932@main)) set the plugin element's CSS width from `contentsSize()`:

```cpp
pluginElement->setInlineStyleProperty(CSSPropertyWidth, size.width(), CSSUnitType::CSS_PX);
pluginElement->setInlineStyleProperty(CSSPropertyHeight, size.height(), CSSUnitType::CSS_PX);
```

In that design, `contentsSize()` was the source of truth for the plugin's width. So having `contentsSize()` return `m_size.width()` would have been circular (`contentsSize()` → `m_size` → layout → CSS width from `contentsSize()` → ...). 

But [290778@main](https://commits.webkit.org/290778@main) removed the CSS width setting to fix the landscape rotation issue. Now the plugin uses `width: 100%` from the PluginDocument stylesheet, so `m_size.width()` is independently determined by the viewport. There's no circular dependency anymore, and `m_size.width()` is the natural source of truth for width.

I just simplified the fix to just `result.setWidth(m_size.width())`. In sizeToFitContent mode, the layout scale is computed as `availableSize.width() / contentWidth`, so the contents width should always equal the plugin width. We're just enforcing that directly instead of relying on float32 arithmetic to round-trip back to an exact integer.

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
aj062: Here is that test failing without my fix ("after" case): https://ews-build.webkit.org/#/builders/13/builds/126483

It passed with my fix in this PR: https://ews-build.webkit.org/#/builders/13/builds/126484

Source/WebKit/WebProcess/Plugins/PDF/UnifiedPDF/UnifiedPDFPlugin.mm
```diff
@@ -1432,7 +1432,16 @@ static String mutationObserverNotificationString()
...
     size.scale(m_scaleFactor);
-    return expandedIntSize(size);
+    auto result = expandedIntSize(size);
+
+    // In sizeToFitContent mode (iOS full-frame PDF), the contents width is
+    // the plugin's visible width. Use it directly to avoid a rounding artifact
+    // where expandedIntSize rounds up by 1 pixel, making the plugin's own
+    // UIScrollView horizontally scrollable and causing rubber-banding.
+    if (shouldSizeToFitContent() && m_size.width() > 0)
+        result.setWidth(m_size.width());
```
aprotyas: we should probably assert that the widths are within 1 unit of each other.
```suggestion
    if (shouldSizeToFitContent() && m_size.width() > 0) {
        ASSERT(std::abs((result - m_size).width()) <= 1);
        result.setWidth(m_size.width());
    }
```

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
```diff
@@ -1420,6 +1420,45 @@ static void checkKeyboardScrollability(TestWKWebView *webView)
 }
 #endif
 
+#if PLATFORM(IOS_FAMILY)
```
aprotyas: There's an iOS family block of tests around L1190 if you want to just group this there: https://github.com/WebKit/WebKit/pull/60299/changes#diff-aebcb202cbe6ad97c711bdb14b134a4b0a5dff448b4589add22467f0ce65076bR1190

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
```diff
@@ -1420,6 +1420,45 @@ static void checkKeyboardScrollability(TestWKWebView *webView)
...
+    // Use width 391 specifically: for test.pdf (page width 129.6pt, contentWidth 161.6),
+    // the layout scale 391/161.6 produces a scaled width of ~391.000031 in float32,
+    // which expandedIntSize rounds up to 392 — a 1-pixel mismatch that makes the
+    // plugin's UIScrollView horizontally scrollable without the fix.
+    RetainPtr webView = adoptNS([[TestWKWebView alloc] initWithFrame:CGRectMake(0, 0, 391, 800) configuration:configurationForWebViewTestingUnifiedPDF().get()]);
+    RetainPtr request = [NSURLRequest requestWithURL:[NSBundle.test_resourcesBundle URLForResource:@"test" withExtension:@"pdf"]];
+    [webView synchronouslyLoadRequest:request.get()];
+
+    // Wait for stable presentation to ensure the scrolling tree is fully committed
+    // and the plugin's WKChildScrollView has its geometry set.
```
aprotyas: This comment is unnecessary, the `-_doAfterNextStablePresentationUpdate:` intent is straightforward.

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
```diff
@@ -1420,6 +1420,45 @@ static void checkKeyboardScrollability(TestWKWebView *webView)
...
+    [webView _doAfterNextStablePresentationUpdate:^{
+        stable = true;
+    }];
+    TestWebKitAPI::Util::run(&stable);
+
+    // The plugin's WKChildScrollView should not be horizontally scrollable.
+    // A rounding mismatch in contentsSize() could make contentSize.width
+    // 1 pixel larger than bounds.width, causing the plugin's scroll view
+    // to capture horizontal pan gestures and rubber-band (rdar://156854435).
+    UIView *childScrollView = [webView wkFirstSubviewWithClass:NSClassFromString(@"WKChildScrollView")];
```
aprotyas:
```suggestion
    RetainPtr childScrollView = [webView wkFirstSubviewWithClass:NSClassFromString(@"WKChildScrollView")];
```

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
```diff
@@ -1420,6 +1420,45 @@ static void checkKeyboardScrollability(TestWKWebView *webView)
...
+    __block bool stable = false;
+    [webView _doAfterNextStablePresentationUpdate:^{
+        stable = true;
+    }];
+    TestWebKitAPI::Util::run(&stable);
+
+    // The plugin's WKChildScrollView should not be horizontally scrollable.
+    // A rounding mismatch in contentsSize() could make contentSize.width
+    // 1 pixel larger than bounds.width, causing the plugin's scroll view
+    // to capture horizontal pan gestures and rubber-band (rdar://156854435).
```
aprotyas: I think we needn't reiterate the bug description. Plus, the expectation that the plugin's WKChildScrollView should not be horizontally scrollable is the whole point of the forthcoming assertions, so I would suggest we remove this comment.

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
```diff
@@ -1420,6 +1420,45 @@ static void checkKeyboardScrollability(TestWKWebView *webView)
...
+    UIView *childScrollView = [webView wkFirstSubviewWithClass:NSClassFromString(@"WKChildScrollView")];
+    EXPECT_NOT_NULL(childScrollView);
+    if (!childScrollView)
+        return;
```
aprotyas: You can write `ASSERT_NOT_NULL` instead, which will fail and terminate this test.
```suggestion
    ASSERT_NOT_NULL(childScrollView);
```

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
```diff
@@ -1420,6 +1420,45 @@ static void checkKeyboardScrollability(TestWKWebView *webView)
...
+
+    auto *scrollView = dynamic_objc_cast<UIScrollView>(childScrollView);
+    EXPECT_NOT_NULL(scrollView);
+    if (!scrollView)
+        return;
```
aprotyas: I don't think we need to cast this to a `UIScrollView`? I would remove this whole bit.

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
```diff
@@ -1420,6 +1420,45 @@ static void checkKeyboardScrollability(TestWKWebView *webView)
...
+
+    EXPECT_GT(scrollView.contentSize.width, 0);
+    EXPECT_GT(scrollView.bounds.size.width, 0);
+    EXPECT_LE(scrollView.contentSize.width, scrollView.bounds.size.width);
```
aprotyas:
```suggestion
    FloatSize contentSize { [childScrollView contentSize] };
    FloatRect bounds { [childScrollView bounds] };
    EXPECT_GT(contentSize.width(), 0);
    EXPECT_GT(bounds.width(), 0);
    EXPECT_LE(contentSize.width(), bounds.width());
```

aprotyas (commented):
The change looks good to me. I had some comments about the test case.

Source/WebKit/WebProcess/Plugins/PDF/UnifiedPDF/UnifiedPDFPlugin.mm
aj062: Done!

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
aj062: Done!

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
aj062: Done!

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
aj062: Done!

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
aj062: wkFirstSubviewWithClass: returns UIView *, and contentSize is a UIScrollView property, that's why we are doing the cast, without this getting a compile error. Moved it into the dynamic_objc_cast on the same line instead of a separate variable.

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
aj062: Done!

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
aj062: Done!

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
aj062: Using FloatSize/FloatRect here would require an extra import and conversion from the CGSize/CGRect returned by UIKit. I didn't do it in this iteration, but happy to change it if you feel strongly about it.

Made other changes you suggested here.

aprotyas: I don't feel strongly about it.

The conversion and the multiple fetches are roughly the same cost 🤷 but I'd at least consider storing `scrollView.contentSize` and `scrollView.bounds.size` out into separate variables first.

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
```diff
@@ -1187,6 +1187,32 @@ static BOOL isBackground(id self)
...
+    __block bool stable = false;
+    [webView _doAfterNextStablePresentationUpdate:^{
+        stable = true;
+    }];
+    TestWebKitAPI::Util::run(&stable);
+
+    RetainPtr childScrollView = [webView wkFirstSubviewWithClass:NSClassFromString(@"WKChildScrollView")];
+    ASSERT_NOT_NULL(childScrollView);
+
+    CGSize contentSize = [childScrollView contentSize];
```
aprotyas, line 1209: I don't see the dynamicCast anymore, is this still fine?

aj062: sorry, my `git-webkit pr` with that change got stuck at a prompt in background, while I was typing the comment in https://github.com/WebKit/WebKit/pull/60299#discussion_r2933778522. Just updated the PR to include the dynamicCast change.

Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm
aj062: For reference, this was the compile error:
```
Tools/TestWebKitAPI/Tests/WebKitCocoa/UnifiedPDFTests.mm:1209:43: 
error: 'UIView' may not respond to 'contentSize' [-Werror]
 1209 |     CGSize contentSize = [childScrollView contentSize];
      |                           ~~~~~~~~~~~~~~~ ^
1 error generated.
```

aprotyas (approved):

PR #60368, opened 2026-03-11
Fix outside list-marker painting with relative-positioned block links

CGQAQ:
#### d184010b1c5533e37b700e227571b73183af02aa
<pre>
Fix outside list-marker painting with relative-positioned block links
<a href="https://bugs.webkit.org/show_bug.cgi?id=18289">https://bugs.webkit.org/show_bug.cgi?id=18289</a>

Reviewed by Alan Baradlay.

Outside list markers can end up structurally nested under descendant renderers that paint in a different self-painting layer from the associated RenderListItem. For the reported case, a relatively positioned display:block link with a negative offset paints its background over the marker’s visual area, but the marker was still painted from the descendant painting path, so the bullet showed through.

Fix:
The change keeps layout behavior unchanged and only adjusts painting:
 - detect outside markers that need to be painted from the associated list-item layer
 - skip painting those markers in the descendant self-painting-layer pass
 - explicitly paint them from the associated list-item layer instead
 - preserve the correct visual position by computing the paint offset from the marker’s ancestor chain

This restores the expected paint order:
 - the list item paints the outside marker
 - the relatively positioned link paints afterward
 - the link background correctly covers the marker area

Tests: fast/lists/outside-marker-covered-by-relpos-block-link-render-tree.html
       imported/w3c/web-platform-tests/css/css-lists/outside-marker-covered-by-relpos-block-link-ref.html
       imported/w3c/web-platform-tests/css/css-lists/outside-marker-covered-by-relpos-block-link.html

* LayoutTests/imported/w3c/web-platform-tests/css/css-lists/outside-marker-covered-by-relpos-block-link-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-lists/outside-marker-covered-by-relpos-block-link-ref.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-lists/outside-marker-covered-by-relpos-block-link.html: Added.
* Source/WebCore/rendering/RenderListItem.cpp:
(WebCore::paintOffsetForMarkerFromAssociatedListItem):
(WebCore::RenderListItem::paint):
* Source/WebCore/rendering/RenderListMarker.cpp:
(WebCore::RenderListMarker::shouldPaintInAssociatedListItemLayer const):
(WebCore::RenderListMarker::paintFromAssociatedListItemLayer):
(WebCore::RenderListMarker::paint):
* Source/WebCore/rendering/RenderListMarker.h:

Canonical link: <a href="https://commits.webkit.org/309188@main">https://commits.webkit.org/309188@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/99e595d7a665f34ac4efdd5897321deca0520a9f

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/149755 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/22474 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/16055 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/158460 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/103187 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/151628 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/22924 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/22564 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/115519 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/82112 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/152715 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/17646 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/134381 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/96260 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/16742 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/14662 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/6306 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/126350 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/12312 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/160937 "Built successfully") | 
| | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/4021 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/13854 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/123542 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/22276 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/18704 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/123748 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/33620 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/22283 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/134105 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/78508 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/18921 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/10856 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/21884 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/85704 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/21614 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/21766 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/21671 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/rendering/RenderListMarker.cpp
+++ b/Source/WebCore/rendering/RenderListMarker.cpp
@@ -43,6 +43,7 @@
 #include "StyleListStyleType.h"
 #include "StyleScope.h"
 #include "TextUtil.h"
+#include <wtf/Assertions.h>
 #include <wtf/StackStats.h>
 #include <wtf/TZoneMallocInlines.h>
 #include <wtf/text/MakeString.h>
@@ -82,6 +83,33 @@ RenderListMarker::RenderListMarker(RenderListItem& listItem, RenderStyle&& style
 
 // Do not add any code in below destructor. Add it to willBeDestroyed() instead.
 RenderListMarker::~RenderListMarker() = default;
+
+bool RenderListMarker::shouldPaintInAssociatedListItemLayer() const
+{
+    if (isInside())
+        return false;
+
+    auto* associatedListItem = listItem();
+    if (!associatedListItem) {
+        ASSERT_NOT_REACHED("Marker must have an associated list item");
+        return false;
+    }
+
+    for (auto* ancestor = parent(); ancestor && ancestor != associatedListItem; ancestor = ancestor->parent()) {
+        if (!ancestor->hasSelfPaintingLayer())
+            continue;
+        if (ancestor->isRenderFragmentedFlow())
+            return false;
+        return ancestor->isPositioned();
+    }
+
+    return false;
+}
+
+void RenderListMarker::paintFromAssociatedListItemLayer(PaintInfo& paintInfo, const LayoutPoint& paintOffset)
+{
+    paint(paintInfo, paintOffset);
+}
 
 void RenderListMarker::willBeDestroyed()
 {
@@ -192,6 +220,9 @@ void RenderListMarker::paintDisclosureMarker(GraphicsContext& context, const Flo
 void RenderListMarker::paint(PaintInfo& paintInfo, const LayoutPoint& paintOffset)
 {
     if (paintInfo.phase != PaintPhase::Foreground && paintInfo.phase != PaintPhase::Accessibility)
+        return;
+
+    if (shouldPaintInAssociatedListItemLayer() && paintInfo.enclosingSelfPaintingLayer() == enclosingLayer())
         return;
 
     if (style().usedVisibility() != Visibility::Visible)
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-lists/outside-marker-covered-by-relpos-block-link-expected.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-lists/outside-marker-covered-by-relpos-block-link-expected.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Reference: outside marker covered by relatively positioned block link</title>
+<link rel="author" title="Jason Leo" href="mailto:m.jason.liu@outlook.com">
+<style>
+body {
+  background: white;
+  font: 16px/1.4 sans-serif;
+}
+
+ul {
+  list-style-position: outside;
+  padding-inline-start: 40px;
+}
+
+.hide-marker::marker {
+  color: transparent;
+}
+
+.LinkContainer {
+  width: 290px;
+}
+
+.faqviewmore {
+    background-color: #efefef;
+    display: block;
+    left: -22px;
+    padding-left: 22px;
+    position: relative;
+}
+</style>
+
+<div class="LinkContainer">
+  <ul>
+    <li><a href="#">Why doesn't my map print?</a></li>
+    <li class="hide-marker"><a class="faqviewmore" href="#">View more questions</a></li>
+  </ul>
+</div>
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-lists/outside-marker-covered-by-relpos-block-link-ref.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-lists/outside-marker-covered-by-relpos-block-link-ref.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Reference: outside marker covered by relatively positioned block link</title>
+<link rel="author" title="Jason Leo" href="mailto:m.jason.liu@outlook.com">
+<style>
+body {
+  background: white;
+  font: 16px/1.4 sans-serif;
+}
+
+ul {
+  list-style-position: outside;
+  padding-inline-start: 40px;
+}
+
+.hide-marker::marker {
+  color: transparent;
+}
+
+.LinkContainer {
+  width: 290px;
+}
+
+.faqviewmore {
+    background-color: #efefef;
+    display: block;
+    left: -22px;
+    padding-left: 22px;
+    position: relative;
+}
+</style>
+
+<div class="LinkContainer">
+  <ul>
+    <li><a href="#">Why doesn't my map print?</a></li>
+    <li class="hide-marker"><a class="faqviewmore" href="#">View more questions</a></li>
+  </ul>
+</div>
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-lists/outside-marker-covered-by-relpos-block-link.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-lists/outside-marker-covered-by-relpos-block-link.html
@@ -1,0 +1,38 @@
+<!DOCTYPE html>
+<meta charset="utf-8">
+<title>CSS Lists: outside marker covered by relatively positioned block link</title>
+<link rel="author" title="CGQAQ" href="mailto:cgqaq@chromium.org">
+<link rel="help" href="https://www.w3.org/TR/CSS22/generate.html#lists">
+<link rel="help" href="https://bugs.webkit.org/show_bug.cgi?id=18289">
+<link rel="match" href="outside-marker-covered-by-relpos-block-link-ref.html">
+<meta name="assert" content="An outside list marker must not remain visibly painted when a relatively positioned display:block link extends into the marker area and paints over it.">
+<style>
+body {
+  background: white;
+  font: 16px/1.4 sans-serif;
+}
+
+ul {
+  list-style-position: outside;
+  padding-inline-start: 40px;
+}
+
+.LinkContainer {
+  width: 290px;
+}
+
+.faqviewmore {
+    background-color: #efefef;
+    display: block;
+    left: -22px;
+    padding-left: 22px;
+    position: relative;
+}
+</style>
+
+<div class="LinkContainer">
+  <ul>
+    <li><a href="#">Why doesn't my map print?</a></li>
+    <li><a class="faqviewmore" href="#">View more questions</a></li>
+  </ul>
+</div>
--- a/Source/WebCore/rendering/RenderListItem.cpp
+++ b/Source/WebCore/rendering/RenderListItem.cpp
@@ -93,6 +93,18 @@ bool isHTMLListElement(const Node& node)
     return isAnyOf<HTMLUListElement, HTMLOListElement>(node);
 }
 
+static LayoutPoint paintOffsetForMarkerFromAssociatedListItem(const RenderListMarker& marker, const RenderListItem& listItem, const LayoutPoint& listItemPaintOffset)
+{
+    auto markerParentPaintOffset = listItemPaintOffset;
+    for (auto* ancestor = marker.parent(); ancestor && ancestor != &listItem; ancestor = ancestor->parent()) {
+        auto* box = dynamicDowncast<RenderBox>(*ancestor);
+        if (!box)
+            break;
+        markerParentPaintOffset.moveBy(box->location());
+    }
+    return markerParentPaintOffset;
+}
+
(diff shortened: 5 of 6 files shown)


Source/WebCore/rendering/updating/RenderTreeBuilderList.cpp
```diff
@@ -94,7 +94,17 @@ static RenderBlock* parentCandidateForMarker(RenderListItem& listItemRenderer, c
...
-    return parentCandidate ? parentCandidate : fallbackParent;
+    auto* candidate = parentCandidate ? parentCandidate : fallbackParent;
+    if (!candidate)
+        return nullptr;
+
+    // An outside marker should not become part of a relatively or stickily positioned descendant's box.
+    // Keep the line-alignment benefits of the nested candidate, but bubble it above any in-flow positioned ancestor.
+    for (auto* ancestor = candidate; ancestor && ancestor != &listItemRenderer; ancestor = dynamicDowncast<RenderBlock>(ancestor->parent())) {
+        if (ancestor->isInFlowPositioned())
+            candidate = dynamicDowncast<RenderBlock>(ancestor->parent());
```
alanbaradlay: `findParentOfEmptyOrFirstLineBox` has strict rules of whether a container is eligible for placing the list marker inside and this logic here implies that any container between the candidate and the list item is also eligible (well minus inflow positioned containers). Is this a valid assumption?
Couldn't we simply skip inflow positioned renderers in `findParentOfEmptyOrFirstLineBox` instead? also should we only skip inflow positioned containers where we _actually_  translate the position?

CGQAQ: I did not know enough context. I have updated the fix based on your suggestions. How about now?

Source/WebCore/rendering/updating/RenderTreeBuilderList.cpp
```diff
@@ -66,6 +66,13 @@ static std::pair<RenderBlock*, RenderBlock*> findParentOfEmptyOrFirstLineBox(Ren
             break;
 
         auto& blockChild = downcast<RenderBlock>(child);
+        if (!marker.isInside() && blockChild.isInFlowPositioned() && !blockChild.offsetForInFlowPosition().isZero()) {
```
alanbaradlay: `offsetForInFlowPosition` seems to be relying on layout bits, while we are in the middle of tree building (maybe checking against computed style is sufficient here).

CGQAQ: Done

Source/WebCore/rendering/updating/RenderTreeBuilderList.cpp
```diff
@@ -66,6 +66,17 @@ static std::pair<RenderBlock*, RenderBlock*> findParentOfEmptyOrFirstLineBox(Ren
             break;
 
         auto& blockChild = downcast<RenderBlock>(child);
+        auto& blockChildStyle = blockChild.style();
+        if (!marker.isInside()
+            && blockChildStyle.position() == PositionType::Relative
```
alanbaradlay: ...and now we are missing sticky. `isInFlowPositioned()` check was fine, what needed fixing was `offsetForInFlowPosition`.

CGQAQ: ...OK, done

LayoutTests/fast/lists/outside-marker-covered-by-relpos-block-link-render-tree-expected.txt
```diff
@@ -0,0 +1,17 @@
...
+            RenderListMarker at (-16,1) size 6x16: bullet
+            RenderInline {A} at (0,1) size 160x16 [color=#0000EE]
+              RenderText {#text} at (0,1) size 160x16
+                text run at (0,1) width 160: "first item"
+          RenderListItem {LI} at (40,18) size 250x18
+layer at (22,18) size 250x18
+  RenderBlock (relative positioned) {A} at (0,0) size 250x18 [bgcolor=#EFEFEF]
+    RenderListMarker at (-16,1) size 6x16: bullet
+    RenderText {#text} at (18,1) size 144x16
+      text run at (18,1) width 144: "view more"
```
alanbaradlay: would prefer ref test (tree dumps are considered legacy due to their fragile nature)

Source/WebCore/rendering/RenderListMarker.cpp
```diff
@@ -83,6 +83,31 @@ RenderListMarker::RenderListMarker(RenderListItem& listItem, RenderStyle&& style
...
+            return false;
+        return ancestor->isPositioned();
+    }
+
+    return false;
+}
+
+void RenderListMarker::paintFromAssociatedListItemLayer(PaintInfo& paintInfo, const LayoutPoint& paintOffset)
+{
+    paint(paintInfo, paintOffset);
```
alanbaradlay, line 111: could you explain why we can't find a valid container for the marker instead?

CGQAQ: Before this change, I tried to reparent the marker to the ListItem, which caused the marker to take the whole line. I tried a lot of ways, but it didn't work out. So I ended up fixing the issue by not changing the render tree structure, only letting the list item paint the marker.

LayoutTests/fast/lists/outside-marker-covered-by-relpos-block-link-render-tree-expected.txt
CGQAQ: Removed

Source/WebCore/rendering/RenderListMarker.cpp
CGQAQ: Is it acceptable?

alanbaradlay: it is certainly doable, but paint-time-only adjustment(s) could lead to weird artifacts (or more like disconnect) caused by the fact that layout is still impacted by the marker position but now we paint it "somewhere else".
A simple example is where the marker inflates the line box.

```
<style>
li::marker {
   font-size: 40px;
}

div {
  outline: 1px solid blue;
  left: -10px;
  position: relative;
}
</style>
<ul><li><div>XXXX
```
This would not be a regression, since the layout is already broken in the same way. It may actually fix some clipping issues where a marker becomes invisible when it is (incorrectly) placed under a block with `overflow: clip`.

Source/WebCore/rendering/RenderListMarker.cpp
```diff
@@ -83,6 +83,31 @@ RenderListMarker::RenderListMarker(RenderListItem& listItem, RenderStyle&& style
 // Do not add any code in below destructor. Add it to willBeDestroyed() instead.
 RenderListMarker::~RenderListMarker() = default;
 
+bool RenderListMarker::shouldPaintInAssociatedListItemLayer() const
+{
+    if (isInside())
+        return false;
+
+    auto* associatedListItem = listItem();
+    if (!associatedListItem)
```
alanbaradlay: maybe assert here as this is an unexpected state?

Source/WebCore/rendering/RenderListMarker.cpp
```diff
@@ -83,6 +83,31 @@ RenderListMarker::RenderListMarker(RenderListItem& listItem, RenderStyle&& style
...
+        return false;
+
+    for (auto* ancestor = parent(); ancestor && ancestor != associatedListItem; ancestor = ancestor->parent()) {
+        if (!ancestor->hasSelfPaintingLayer())
+            continue;
+        if (ancestor->isRenderFragmentedFlow())
+            return false;
+        return ancestor->isPositioned();
```
alanbaradlay, line 103: all kinds of positioned?

alanbaradlay (approved):

Source/WebCore/rendering/RenderListMarker.cpp
CGQAQ: I think so, if the ancestor is positioned, we want the marker to be rendered in the ListItem pass instead, to let the positioned content be correctly rendered on top of the marker

Source/WebCore/rendering/RenderListMarker.cpp
CGQAQ: OK, got it. I'm new to WebKit and will look into the renderer code more; hopefully, I can fix the broken part in the future. Thanks for your patient guidance

Source/WebCore/rendering/RenderListMarker.cpp
CGQAQ: You're right, replaced with RELEASE_ASSERT with useful info

alanbaradlay: please use ASSERT_NOT_REACHED() (debug only) only as I am not convinced we should crash the web content process (potentially causing data loss) just because a marker is detached from the tree -but it is certainly an unexpected state so debug builds should catch that.

Source/WebCore/rendering/RenderListMarker.cpp
alanbaradlay: thank you for taking the time to fix this long-standing bug.

Source/WebCore/rendering/RenderListMarker.cpp
CGQAQ: Done!

PR #60512, opened 2026-03-12
[Github.com] Skip full style resolution for non-var() descendants during custom property changes

NathanSolomon733:
#### f3c4bda80ceabecfa33c6df0bfe067b311e6dcba
<pre>
[Github.com] Skip full style resolution for non-var() descendants during custom property changes
<a href="https://bugs.webkit.org/show_bug.cgi?id=309838">https://bugs.webkit.org/show_bug.cgi?id=309838</a>
<a href="https://rdar.apple.com/167267369">rdar://167267369</a>

Reviewed by NOBODY (OOPS!).

Resizing the sidebar on GitHub changes a CSS custom property on a parent with ~15,000
descendants. Previously all descendants got full selector matching and cascade even though the
vast majority do not reference the changed property via var().

This patch adds a fast path. When a descendant does not reference the changed property, its
existing style is cloned, the parent&apos;s inherited custom properties are copied onto the clone,
and the result is handed to Style::Update. This skips selector matching, property application,
and the MatchedDeclarationsCache. Elements that reference the changed property still get full
resolution.

Split inherited property comparison in determineChanges so custom property changes produce a
separate Change::CustomProperty flag instead of being folded into Change::Inherited.
The fast path activates when the parent&apos;s only change is to custom properties.

Track which custom property names each element references via var() or declares using a 31-bit
bloom filter stored on NonInheritedData. Each name is hashed into a single bit of the filter
via maskForCustomPropertyName.
Also use 1 bit to track if the element declares its own inherited custom properties.
Filter bits are accumulated on the Builder during property
application and written to NonInheritedData once after all properties are applied, avoiding
copy-on-write on the shared style data.

When a parent element&apos;s inherited custom properties change, compute a bloom filter of the changed names and
propagate it to descendants via the TreeResolver. Each descendant ANDs its own filter with the changed filter. A
zero result means the element does not reference any changed property and can take the fast path.
A nonzero result means the element may reference a changed property and needs full resolution.

The optimization is conservative. Bloom filter false positives cause unnecessary full
resolution but never skip elements that need it. If a parent has any non-custom-property
inherited change alongside the custom property change, the optimization falls back to full
resolution for all children. We also do not take the optimization if the element itself
declares its own inherited custom properties.

A microbenchmark was added to test this optimization. The median time on a Macbook Pro at desk
decreased from 556 ms per iteration to 5 ms per iteration.

* LayoutTests/fast/css/variables/custom-property-fastpath-preserves-custom-properties-expected.txt: Added.
* LayoutTests/fast/css/variables/custom-property-fastpath-preserves-custom-properties.html: Added.
* PerformanceTests/CSS/CustomPropertySubtreeInvalidation.html: Added.
* Source/WebCore/style/StyleBuilder.cpp:
(WebCore::Style::Builder::addCustomPropertyFilterBit):
(WebCore::Style::Builder::commitCustomPropertyState):
(WebCore::Style::Builder::applyCustomProperty):
* Source/WebCore/style/StyleBuilder.h:
(WebCore::Style::Builder::customPropertyFilter const):
(WebCore::Style::Builder::declaresInheritedCustomProperty const):
* Source/WebCore/style/StyleChange.cpp:
(WebCore::Style::determineChanges):
(WebCore::Style::operator&lt;&lt;):
* Source/WebCore/style/StyleChange.h:
* Source/WebCore/style/StyleResolver.cpp:
(WebCore::Style::Resolver::styleForKeyframe const):
(WebCore::Style::Resolver::styleForPage):
(WebCore::Style::Resolver::applyMatchedProperties):
* Source/WebCore/style/StyleSubstitutionResolver.cpp:
(WebCore::Style::SubstitutionResolver::substituteVariableFunction):
(WebCore::Style::SubstitutionResolver::trySimpleSubstitution):
* Source/WebCore/style/StyleTreeResolver.cpp:
(WebCore::Style::TreeResolver::computeDescendantsToResolve const):
(WebCore::Style::TreeResolver::resolveAgainInDifferentContext):
(WebCore::Style::TreeResolver::applyCascadeAfterAnimation):
(WebCore::Style::TreeResolver::resolveComposedTree):
* Source/WebCore/style/StyleTreeResolver.h:
(WebCore::Style::ComputedStyle::setInheritedCustomPropertiesFrom):
(WebCore::Style::ComputedStyle::nonFastPathInheritedEqual const):
(WebCore::Style::ComputedStyle::nonFastPathInheritedEqualIgnoringCustomProperties const):
(WebCore::Style::ComputedStyle::inheritedCustomPropertiesEqual const):
* Source/WebCore/style/computed/StyleComputedStyleBase.cpp:
(WebCore::Style::ComputedStyleBase::customPropertyFilter const):
(WebCore::Style::ComputedStyleBase::setCustomPropertyFilter):
(WebCore::Style::ComputedStyleBase::declaresInheritedCustomProperty const):
(WebCore::Style::ComputedStyleBase::setDeclaresInheritedCustomProperty):
(WebCore::Style::ComputedStyleBase::maskForCustomPropertyName):
* Source/WebCore/style/computed/data/StyleCustomPropertyData.cpp:
(WebCore::Style::CustomPropertyData::changedCustomPropertiesFilter const):
* Source/WebCore/style/computed/data/StyleCustomPropertyData.h:
* Source/WebCore/style/computed/data/StyleInheritedRareData.cpp:
(WebCore::Style::InheritedRareData::operator== const):
(WebCore::Style::InheritedRareData::equalIgnoringCustomProperties const):
* Source/WebCore/style/computed/data/StyleInheritedRareData.h:
* Source/WebCore/style/computed/data/StyleNonInheritedData.cpp:
(WebCore::Style::NonInheritedData::NonInheritedData):
* Source/WebCore/style/computed/data/StyleNonInheritedData.h:
</pre>
https://github.com/WebKit/WebKit/commit/f3c4bda80ceabecfa33c6df0bfe067b311e6dcba

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/166765 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/40255 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/33836 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/175813 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/124023 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/ac5a8fda-2325-4b39-96d8-cf9b11c66c36) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/168631 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/40688 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/40263 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/129670 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/91328 "1 flakes 2 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/fc6f6e02-3640-437a-baa0-8cf681b2b3e4) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/169724 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/32069 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/149841 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/110195 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b4998190-55aa-464d-befb-12452e59ae94) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/31045 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/29745 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/24549 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/141016 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/27538 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/178353 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/31247 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/29245 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/138040 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/40325 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/33893 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/137953 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/41013 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/149336 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/103810 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/33233 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/26201 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/39524 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/112598 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/38920 "Built successfully") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/4292 "") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/39165 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/39063 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/style/StyleBuilder.cpp
+++ b/Source/WebCore/style/StyleBuilder.cpp
@@ -119,6 +119,20 @@ void Builder::applyAllProperties()
     adjustAfterApplying();
 }
 
+void Builder::addCustomPropertyFilterBit(const AtomString& name)
+{
+    m_customPropertyFilter |= ComputedStyleBase::maskForCustomPropertyName(name);
+}
+
+void Builder::commitCustomPropertyState()
+{
+    auto& style = m_state->style();
+    if (m_customPropertyFilter)
+        style.setCustomPropertyFilter(m_customPropertyFilter);
+    if (m_declaresInheritedCustomProperty)
+        style.setDeclaresInheritedCustomProperty();
+}
+
 // Top priority properties affect resolution of high priority properties.
 void Builder::applyTopPriorityProperties()
 {
@@ -441,10 +455,14 @@ void Builder::applyCustomProperty(const AtomString& name, Variant<Ref<const Styl
 {
     auto& style = m_state->style();
 
+    addCustomPropertyFilterBit(name);
+
     auto registeredCustomProperty = m_state->registeredProperty(name);
 
     auto applyValue = [&](Ref<const CustomProperty>&& valueToApply) {
         bool isInherited = !registeredCustomProperty || registeredCustomProperty->inherits;
+        if (isInherited)
+            m_declaresInheritedCustomProperty = true;
         state().style().setCustomPropertyValue(WTF::move(valueToApply), isInherited);
     };
 
--- a/Source/WebCore/style/StyleChange.h
+++ b/Source/WebCore/style/StyleChange.h
@@ -41,7 +41,8 @@ enum class Change : uint8_t {
     Inherited = 1 << 1,
     FastPathInherited = 1 << 2,
     Container = 1 << 3,
-    Renderer = 1 << 4
+    Renderer = 1 << 4,
+    InheritedCustomProperty = 1 << 5,
 };
 
 constexpr OptionSet<Change> inheritedChanges() { return { Change::Inherited, Change::FastPathInherited }; }
--- a/Source/WebCore/style/StyleTreeResolver.cpp
+++ b/Source/WebCore/style/StyleTreeResolver.cpp
@@ -65,6 +65,7 @@
 #include "StyleAdjuster.h"
 #include "StyleBuilder.h"
 #include "StyleComputedStyle+SettersInlines.h"
+#include "StyleCustomPropertyData.h"
 #include "StyleFontSizeFunctions.h"
 #include "StyleOriginatedTimelinesController.h"
 #include "StylePositionTryFallbackTactic.h"
@@ -279,7 +280,7 @@ static bool affectsRenderedSubtree(Element& element, const Style::ComputedStyle&
     return false;
 }
 
-auto TreeResolver::computeDescendantsToResolve(const ElementUpdate& update, const Style::ComputedStyle* existingStyle, Validity validity) const -> DescendantsToResolve
+auto TreeResolver::computeDescendantsToResolve(ElementUpdate& update, const Style::ComputedStyle* existingStyle, Validity validity) const -> DescendantsToResolve
 {
     if (parent().descendantsToResolve == DescendantsToResolve::All)
         return DescendantsToResolve::All;
@@ -296,8 +297,10 @@ auto TreeResolver::computeDescendantsToResolve(const ElementUpdate& update, cons
             }
             return false;
         }();
-        if (customPropertyInStyleContainerQueryChanged)
+        if (customPropertyInStyleContainerQueryChanged) {
+            update.changes.add(Change::Container);
             return DescendantsToResolve::All;
+        }
     }
 
     if (update.changes.containsAny({ Change::Container, Change::Renderer }))
@@ -305,6 +308,11 @@ auto TreeResolver::computeDescendantsToResolve(const ElementUpdate& update, cons
 
     if (update.changes.containsAny(inheritedChanges()))
         return DescendantsToResolve::Children;
+
+    if (update.changes.contains(Change::InheritedCustomProperty)) {
+        ASSERT(!update.changes.containsAny(inheritedChanges()));
+        return DescendantsToResolve::Children;
+    }
 
     if (update.changes.contains(Change::NonInherited))
         return DescendantsToResolve::ChildrenWithExplicitInherit;
@@ -1053,6 +1061,7 @@ std::unique_ptr<Style::ComputedStyle> TreeResolver::resolveAgainInDifferentConte
     };
 
     styleBuilder.applyAllProperties();
+    styleBuilder.commitCustomPropertyState();
 
     if (newStyle->display() == DisplayType::None)
         return nullptr;
@@ -1091,6 +1100,7 @@ HashSet<AnimatableCSSProperty> TreeResolver::applyCascadeAfterAnimation(Style::C
     };
 
     styleBuilder.applyAllProperties();
+    styleBuilder.commitCustomPropertyState();
 
     return styleBuilder.overriddenAnimatedProperties();
 }
@@ -1330,13 +1340,36 @@ void TreeResolver::resolveComposedTree()
 
         auto changes = OptionSet<Change> { };
         auto descendantsToResolve = DescendantsToResolve::None;
+        uint32_t changedCustomPropertyFilter = 0;
 
 #if ENABLE(ACCESSIBILITY_ISOLATED_TREE)
         bool didAXUpdateFontSubtree = parent.didAXUpdateFontSubtree;
         bool didAXUpdateTextColorSubtree = parent.didAXUpdateTextColorSubtree;
 #endif
         auto resolutionType = determineResolutionType(element.get(), style, parent.descendantsToResolve, parent.changes);
-        if (resolutionType) {
+        auto canTakeCustomPropertyFastPath = [&] {
+            return parent.changedCustomPropertyFilter
+                && parent.changes.containsOnly(Change::InheritedCustomProperty)
+                && parent.descendantsToResolve == DescendantsToResolve::Children
+                && style
+                && !style->hasPseudoElementStyles()
+                && !element->needsStyleRecalc()
+                && !element->hasKeyframeEffects({ })
+                && !(style->customPropertyFilter() & parent.changedCustomPropertyFilter)
+                && !style->declaresInheritedCustomProperty();
+        };
+
+        if (canTakeCustomPropertyFastPath()) {
+            auto clonedStyle = Style::ComputedStyle::clonePtr(*style);
+            clonedStyle->setInheritedCustomPropertiesFrom(parent.style);
+            ElementUpdate elementUpdate { WTF::move(clonedStyle), Change::InheritedCustomProperty };
+            changes = Change::InheritedCustomProperty;
+            descendantsToResolve = DescendantsToResolve::Children;
+            changedCustomPropertyFilter = parent.changedCustomPropertyFilter;
+            style = elementUpdate.style.get();
+            m_update->addElement(element.get(), parent.element, WTF::move(elementUpdate));
+            clearNeedsStyleResolution(element.get());
+        } else if (resolutionType) {
             element->resetComputedStyle();
 
             if (*resolutionType == ResolutionType::Full)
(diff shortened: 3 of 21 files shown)


Source/WebCore/style/StyleTreeResolver.cpp
```diff
@@ -1239,6 +1287,22 @@ static std::unique_ptr<RenderStyle> createInheritedDisplayContentsStyleIfNeeded(
     return style;
 }
 
+void TreeResolver::fastPathInheritCustomProperties(Element& element, const RenderStyle*& style, const RenderStyle& parentStyle, const HashSet<AtomString>& changedCustomProperties)
+{
+    if (!style)
+        return;
+    if (auto* renderer = element.renderer()) {
+        renderer->mutableStyle().updateChangedInheritedCustomProperties(parentStyle, changedCustomProperties);
```
anttijk: Style mutations are only allowed by RenderTreeUpdater and must go via Style::Update We should stick to this invariant to not lose control of the architecture.

Source/WebCore/style/StyleBuilder.cpp
```diff
@@ -119,6 +119,11 @@ void Builder::applyAllProperties()
     adjustAfterApplying();
 }
 
+void Builder::addCustomPropertyFilterBit(const AtomString& name)
+{
+    m_customPropertyFilter |= ComputedStyleBase::maskForCustomPropertyName(name);
+}
```
anttijk, line 125: We might want to look if we can fix MatchedDeclarationsCache keying so it is limited to custom properties used by the style only. We should be able to use a cache entry even if inherited custom properties differ if none of the changed properties was used for building the cached style. This sort of infrastructure is needed for that too.

In theory that might bring significant performance gains here and other cases while likely being less invasive.

anttijk (commented):
We might want to do the inheritance fast path too but I feel taking a look at MDC might be fruitful first.

Source/WebCore/rendering/style/RenderStyle.h
```diff
@@ -115,6 +118,11 @@ class RenderStyle final : public RenderStyleProperties {
     inline bool hasExplicitlyInheritedProperties() const;
     inline void setHasExplicitlyInheritedProperties();
 
+    uint32_t customPropertyFilter() const;
+    void setCustomPropertyFilter(uint32_t);
```
anttijk: cool use of a mini bloom filter

Source/WebCore/style/StyleChange.h
```diff
@@ -42,7 +42,8 @@ enum class Change : uint8_t {
     Inherited = 1 << 1,
     FastPathInherited = 1 << 2,
     Container = 1 << 3,
-    Renderer = 1 << 4
+    Renderer = 1 << 4,
+    CustomProperty = 1 << 5,
```
anttijk: `InheritedCustomProperty`?

Source/WebCore/style/StyleTreeResolver.cpp
```diff
@@ -1330,13 +1342,44 @@ void TreeResolver::resolveComposedTree()
...
+                && parent.descendantsToResolve == DescendantsToResolve::Children
+                && style
+                && !element->needsStyleRecalc()
+                && !element->hasKeyframeEffects({ })
+                && !usesChangedCustomProperty();
+        };
+
+        if (canTakeCustomPropertyFastPath()) {
+            auto clonedStyle = RenderStyle::clonePtr(*style);
+            clonedStyle->setInheritedCustomPropertiesFrom(parent.style);
```
anttijk, line 1364: Claude flagged that this will wipe out elements own inherited properties. Test case:

```
<div id="parent" style="--changed: 1px;">
  <div id="mid" style="--mine: 99px;">
   <div id="leaf" style="width: var(--mine);"></div>
  </div>
 </div>
 <script>
  document.body.offsetHeight;
  parent.style.setProperty('--changed', '2px');  // only --changed changes on parent
  document.body.offsetHeight;
  // BUG: getComputedStyle(mid).getPropertyValue('--mine') becomes "" 
  //   and leaf width collapses, because mid took the fast path and had
  //   its --mine clobbered by the parent's inherited custom properties.
 </script>
```

anttijk (changes requested):

Source/WebCore/style/StyleChange.h
NathanSolomon733: Done

Source/WebCore/style/StyleTreeResolver.cpp
NathanSolomon733: I decided to disable the fast path for elements with own inherited properties. The site we're trying to optimize, Github, does not regress in performance if we disable for this case. Most of the elements being style resolved only reference a custom property, not declare them.
I have this example above as a test case to the now added layout test /fast/css/variables/custom-property-fastpath-preserves-own-declarations.html which tests that the fast path does not break correctness for custom properties.

Also, because Github did not get slower if we take the slow path for elements with their own declarations, I also changed the microbenchmark (PerformanceTests/CSS/CustomPropertySubtreeInvalidation.html) to not include elements that do this to make the microbenchmark more accurate to the real world perf issue.

NathanSolomon733: Claude also caught that cloning the style with clonePtr drops the element's cached pseudo-element styles, so the render-tree update tears down ::before/::after generated content when an ancestor's custom property changes I added this case to the layout test in this PR. To keep this PR from being overly complicated I also exclude elements that have pseudo-element styles from the fast path. The microbenchmark perf is unaffected and neither is Github perf.

PR #60548, opened 2026-03-13
[JSC] Implement `Iterator.prototype.includes` (disabled by default)

tetsuharuohzeki:
#### 7fdc018c4dd77266f919db0a9ff495ebafde757e
<pre>
[JSC] Implement `Iterator.prototype.includes` (disabled by default)
<a href="https://bugs.webkit.org/show_bug.cgi?id=309883">https://bugs.webkit.org/show_bug.cgi?id=309883</a>

Reviewed by Yusuke Suzuki.

This implements Iterator Includes proposal <a href="https://tc39.es/proposal-iterator-includes/">https://tc39.es/proposal-iterator-includes/</a>
that refers rev. March 12, 2026.
<a href="https://github.com/tc39/proposal-iterator-includes/tree/8787f1ae4500fc6c7851f4eccb9557f9d74fa694">https://github.com/tc39/proposal-iterator-includes/tree/8787f1ae4500fc6c7851f4eccb9557f9d74fa694</a>

This feature is behind the flag `--useIteratorIncludes`.

Test: JSTests/stress/iterator-prototype-includes.js

* JSTests/stress/iterator-prototype-includes.js: Added.
* Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp:
    This introduces `--useIteratorIncludes` flag.

Canonical link: <a href="https://commits.webkit.org/310769@main">https://commits.webkit.org/310769@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/5618d974f7523081561a6b899b4624bd80e83070

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/154477 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/27735 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/20895 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/163233 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/107946 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/27869 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/27585 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/119487 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/84510 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/157436 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/21765 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/138742 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/100184 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/20852 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/18859 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/11063 "Built successfully") | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/146526 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/130514 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/16586 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/165705 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/15308 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/8912 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/18195 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/127586 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/27281 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/22899 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/127731 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/27205 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/138379 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/83888 "The change is no longer eligible for processing.") | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/23621 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/22625 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/15171 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/186185 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/26896 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/90998 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/47722 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/26476 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/26707 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/26549 "Built successfully") | | |

### DIFF
--- a/JSTests/stress/iterator-prototype-includes.js
+++ b/JSTests/stress/iterator-prototype-includes.js
@@ -1,0 +1,524 @@
+//@ requireOptions("--useIteratorIncludes=1")
+
+function sameValue(a, b, testname) {
+    if (a !== b)
+        throw new Error(`${testname}: Expected ${b} but got ${a}`);
+}
+
+function shouldThrow(caseName, fn, expectedErrorCtor, expectedErrorMessage) {
+    if (!caseName) {
+        throw new Error(`must specify test case name`);
+    }
+
+    try {
+        fn();
+        throw new Error(`${caseName}: Expected to throw, but succeeded`);
+    } catch (e) {
+        if (!(e instanceof expectedErrorCtor))
+            throw new Error(`${caseName}: Expected to throw ${expectedErrorCtor.name} but got ${e.name}`);
+        if (e.message !== expectedErrorMessage)
+            throw new Error(`${caseName}: Expected ${expectedErrorCtor.name} with '${expectedErrorMessage}' but got '${e.message}'`);
+    }
+}
+
+function callTestTargetFunction(iterator, searchElement, skippedElements) {
+    return Iterator.prototype.includes.call(iterator, searchElement, skippedElements);
+}
+
+class TestIterator extends Iterator {
+    value = 0;
+    isClosed = false;
+    isDone = false;
+    constructor(max) {
+        super();
+        if (max < 0) {
+            throw new RangeError('max must be >= 0');
+        }
+        this.max = max;
+    }
+    next() {
+        const value = this.value;
+        if (value > this.max) {
+            this.isDone = true;
+            return {
+                done: true,
+                value: undefined,
+            };
+        }
+
+        this.value += 1;
+        return {
+            done: false,
+            value,
+        };
+    }
+    return() {
+        this.isClosed = true;
+        return {
+            done: true,
+            value: undefined,
+        };
+    }
+}
+
+// without second argument
+
+{
+    const TEST_NAME = `simply include`;
+
+    const gen = function* (num) {
+        for (let i = 0; i <= num; ++i) {
+            yield i;
+        }
+    };
+    const iter = gen(3);
+    sameValue(callTestTargetFunction(iter, 1), true, TEST_NAME);
+}
+
+
+{
+    const TEST_NAME = `simply not include`;
+
+    const gen = function* (len) {
+        for (let i = 0; i < len; ++i) {
+            yield i;
+        }
+    };
+    const iter = gen(3);
+    sameValue(callTestTargetFunction(iter, 4), false, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `iterator is empty`;
+
+    const gen = function* () {};
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, 1), false, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `Use SameValueZero comparison`;
+
+    const gen = function* () {
+        yield +0;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, -0), true, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `Use SameValueZero comparison, search NaN`;
+
+    const gen = function* () {
+        yield NaN;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, NaN), true, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `Use SameValueZero comparison, search Number.NaN`;
+
+    const gen = function* () {
+        yield NaN;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, Number.NaN), true, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `Use SameValueZero comparison, the iteraror yeild NaN but the searching value is anothor one.`;
+
+    const gen = function* () {
+        yield NaN;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, 0), false, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `Use SameValueZero comparison, the iteraror yeild Number.NaN but the searching value is anothor one.`;
+
+    const gen = function* () {
+        yield Number.NaN;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, 0), false, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `iterator should be closed`;
+
+    const gen = function* () {
+        yield 0;
+        yield 1;
+        yield 2;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, 1), true, `${TEST_NAME}: consume once`);
+    sameValue(callTestTargetFunction(iter, 2), false, `${TEST_NAME}: 2nd iteration on consumed iterator would be nothing`);
+}
+
+{
+    const TEST_NAME = `do not iterate all, the actual should be expected`;
+
+    class Unreachable extends Error {
+        constructor(message) {
+            super(message);
+            this.name = new.target.name;
+        }
+    }
+    const gen = function* () {
+        yield 0;
+        throw new Unreachable("do not iterate all after the target was found");
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, 0), true, TEST_NAME);
+}
+
+{
+    const TEST_NAME = 'calling .includes() should close the iterator if the searched was found';
+
+    const iter = new TestIterator(5);
+    sameValue(callTestTargetFunction(iter, 4), true, `${TEST_NAME}: call`);
+    sameValue(iter.isDone, false, `${TEST_NAME}: iterator should not be done`);
+    sameValue(iter.isClosed, true, `${TEST_NAME}: iterator should be closed`);
+}
+
+{
+    const TEST_NAME = 'calling .includes() should not close the iterator if no matched';
+
+    const iter = new TestIterator(5);
+    sameValue(callTestTargetFunction(iter, -1), false, `${TEST_NAME}: call`);
+    sameValue(iter.isDone, true, `${TEST_NAME}: iterator should be done`);
+    sameValue(iter.isClosed, false, `${TEST_NAME}: iterator should not be closed`);
+}
+
+// with second argument
+
+{
+    const TEST_NAME = `call with 2nd arg: simply include`;
+
+    const gen = function* (len) {
+        for (let i = 0; i < len; ++i) {
+            yield i;
+        }
+    };
+    const iter = gen(3);
+    sameValue(callTestTargetFunction(iter, 2, 2), true, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg is 0: simply include`;
+
+    const gen = function* (len) {
+        for (let i = 0; i < len; ++i) {
+            yield i;
+        }
+    };
+    const iter = gen(3);
+    sameValue(callTestTargetFunction(iter, 2, 2), true, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: simply include but the target is in skipped range`;
+
+    const gen = function* (len) {
+        for (let i = 0; i < len; ++i) {
+            yield i;
+        }
+    };
+    const iter = gen(3);
+    sameValue(callTestTargetFunction(iter, 2, 3), false, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: simply include but the skipped range is over than the iterator length`;
+
+    const iter = new TestIterator(3);
+    sameValue(callTestTargetFunction(iter, 2, 100), false, TEST_NAME);
+    sameValue(iter.isDone, true, `${TEST_NAME}: iterator should be done properly`);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: simply not include`;
+
+    const gen = function* (len) {
+        for (let i = 0; i < len; ++i) {
+            yield i;
+        }
+    };
+    const iter = gen(3);
+    sameValue(callTestTargetFunction(iter, 4, 0), false, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: iterator is empty`;
+
+    const gen = function* () {};
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, 1, 1), false, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: Use SameValueZero comparison`;
+
+    const gen = function* () {
+        yield -1;
+        yield +0;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, -0, 1), true, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: Use SameValueZero comparison, search NaN`;
+
+    const gen = function* () {
+        yield -1;
+        yield NaN;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, NaN, 1), true, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: Use SameValueZero comparison, search NaN but the iterator will not return NaN`;
+
+    const gen = function* () {
+        yield -1;
+        yield 0;
+        yield 1;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, NaN, 1), false, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: Use SameValueZero comparison, search Number.NaN`;
+
+    const gen = function* () {
+        yield -1;
+        yield NaN;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, Number.NaN, 1), true, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: Use SameValueZero comparison, search Number.NaN but the iterator will not return NaN`;
+
+    const gen = function* () {
+        yield -1;
+        yield 0;
+        yield 1;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, Number.NaN, 1), false, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: Use SameValueZero comparison, the iterator yield NaN but the searching value is another one`;
+
+    const gen = function* (limit) {
+        for (let i = 0; i < limit; ++i) {
+            yield NaN;
+        }
+    };
+    const iter = gen(10);
+    sameValue(callTestTargetFunction(iter, 0, 1), false, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: Use SameValueZero comparison, the iterator yield Number.NaN but the searching value is another one`;
+
+    const gen = function* (limit) {
+        for (let i = 0; i < limit; ++i) {
+            yield NaN;
+        }
+    };
+    const iter = gen(10);
+    sameValue(callTestTargetFunction(iter, 0, 1), false, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: iterator should be closed`;
+
+    const gen = function* () {
+        yield 0;
+        yield 1;
+        yield 2;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, 1, 1), true, `${TEST_NAME}: consume once`);
+    sameValue(callTestTargetFunction(iter, 2, 1), false, `${TEST_NAME}: 2nd iteration on closed iterator would be nothing`);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg but skipped all: iterator should be closed`;
+
+    const gen = function* () {
+        yield 0;
+        yield 1;
+        yield 2;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, 1, 100), false, `${TEST_NAME}: consume once`);
+    sameValue(callTestTargetFunction(iter, 2, 100), false, `${TEST_NAME}: 2nd iteration on closed iterator would be nothing`);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: do not iterate all`;
+
+    class Unreachable extends Error {
+        constructor(message) {
+            super(`unreachable: ${message}`);
+            this.name = new.target.name;
+        }
+    }
+    const gen = function* () {
+        yield -1;
+        yield 0;
+        throw new Unreachable(`${TEST_NAME}: after the searched was found`);
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, 0, 1), true, `${TEST_NAME}: the searched was found`);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: skipped range throw something`;
+
+    class TestIterator extends Iterator {
+        value = 0;
+        isClosed = false;
+        isDone = false;
+        constructor(max) {
+            super();
+            this.max = max;
+        }
+        next() {
+            const value = this.value;
+            if (value > this.max) {
+                this.isDone = true;
+                return {
+                    done: true,
+                    value: undefined,
+                };
+            }
+
+            throw new Error('this is in skipped range!');
+        }
+        return() {
+            this.isClosed = true;
+            return {
+                done: true,
+                value: undefined,
+            };
+        }
+    }
+
+    const iter = new TestIterator(3);
+    shouldThrow(TEST_NAME, () => {
+        callTestTargetFunction(iter, 1, 100);
+    }, Error, 'this is in skipped range!');
+    sameValue(iter.isDone, false, `${TEST_NAME}: iterator should not be done`);
+    sameValue(iter.isClosed, false, `${TEST_NAME}: iterator should not be closed`);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg: calling .includes() should close the iterator`;
+
+    const iter = new TestIterator(5);
+    sameValue(callTestTargetFunction(iter, 4, 1), true, `${TEST_NAME}: calling .includes()`);
+    sameValue(iter.isDone, false, `${TEST_NAME}: iterator should not be done`);
+    sameValue(iter.isClosed, true, `${TEST_NAME}: iterator should be closed`);
+}
+
+{
+    const TEST_NAME = `call with 2nd arg but skip everything`;
+
+    const iter = new TestIterator(5);
+    sameValue(callTestTargetFunction(iter, 4, 100), false, `${TEST_NAME}: calling .includes()`);
+    sameValue(iter.isDone, true, `${TEST_NAME}: iterator should be done`);
+    sameValue(iter.isClosed, false, `${TEST_NAME}: iterator should not be closed`);
+}
+
+{
+    const valid2ndArgument = [
+        [Number.POSITIVE_INFINITY, `+Infinity`],
+        [Number.MAX_SAFE_INTEGER, `Number.MAX_SAFE_INTEGER`],
+        [Number.MAX_SAFE_INTEGER + 1, `(Number.MAX_SAFE_INTEGER + 1)`],
+        [Number.MAX_VALUE, `Number.MAX_VALUE`],
+    ];
+
+    for (const [skippedElements, label] of valid2ndArgument) {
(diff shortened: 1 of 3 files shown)


Source/JavaScriptCore/builtins/JSIteratorPrototype.js
```diff
@@ -472,6 +472,59 @@ function windows(windowSize /*, undersized */)
...
+    var skipped = 0;
+    var wrapper = { @@iterator: function () { return iterated; }};
+    for (var item of wrapper) {
+        if (skipped < toSkip) {
+            skipped += 1;
+            continue;
+        }
+
+        // Use SameValueZero comparison, rather than just StrictEquals.
+        if (item === searchElement || (item !== searchElement && item !== item))
```
michaelficarra: This will cause `includes` to always return `true` when the iterator yields a `NaN`, regardless of what you're searching for. You want

```suggestion
        if (item === searchElement || (searchElement !== searchElement && item !== item))
```

and should also test for this scenario.

JSTests/stress/iterator-prototype-includes.js
```diff
@@ -0,0 +1,443 @@
...
+{
+    const TEST_NAME = `iterator is empty`;
+
+    const gen = function* () {};
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, 1), false, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `Use SameValueZero comparision`;
```
michaelficarra:
```suggestion
    const TEST_NAME = `Use SameValueZero comparison`;
```

as well as the 5 other occurrences of this typo

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
```diff
@@ -472,6 +472,59 @@ function windows(windowSize /*, undersized */)
...
+            }
+        }
+        toSkip = skippedElements;
+    }
+
+    if (toSkip < 0) {
+        try {
+            @iteratorGenericClose(iterated);
+        } finally {
+            @throwRangeError("Iterator.prototype.includes requires that second argument be a positive integer.");
```
michaelficarra:
```suggestion
            @throwRangeError("Iterator.prototype.includes requires that the second argument is a non-negative integral Number or Infinity.");
```

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
```diff
@@ -472,6 +472,59 @@ function windows(windowSize /*, undersized */)
...
+    var skippedElements = @argument(1);
+    var toSkip = 0;
+    var isIntegerOrInfinity = false;
+    if (skippedElements !== @undefined) {
+        isIntegerOrInfinity = (typeof skippedElements === 'number') && (@isInteger(skippedElements) || !@isFinite(skippedElements));
+        if (!isIntegerOrInfinity) {
+            try {
+                @iteratorGenericClose(iterated);
+            } finally {
+                @throwTypeError("Iterator.prototype.includes requires that second argument be a +/- Infinity or integer.");
```
michaelficarra:
```suggestion
                @throwTypeError("Iterator.prototype.includes requires that the second argument is a non-negative integral Number or Infinity.");
```

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
```diff
@@ -472,6 +472,59 @@ function windows(windowSize /*, undersized */)
...
+        } finally {
+            @throwRangeError("Iterator.prototype.includes requires that second argument be a positive integer.");
+        }
+    }
+
+    var skipped = 0;
+    var wrapper = { @@iterator: function () { return iterated; }};
+    for (var item of wrapper) {
+        if (skipped < toSkip) {
+            skipped += 1;
```
michaelficarra: This won't work for `toSkip` values above `Number.MAX_SAFE_INTEGER` (`9007199254740991`). You can use a bigint instead.

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
```diff
@@ -472,6 +472,59 @@ function windows(windowSize /*, undersized */)
...
+    var iterated = this;
+
+    if (!@isObject(iterated))
+        @throwTypeError("Iterator.prototype.includes requires that |this| be an Object.");
+
+    var skippedElements = @argument(1);
+    var toSkip = 0;
+    var isIntegerOrInfinity = false;
+    if (skippedElements !== @undefined) {
+        isIntegerOrInfinity = (typeof skippedElements === 'number') && (@isInteger(skippedElements) || !@isFinite(skippedElements));
```
justinmichaud: Is there a reason we need @isInteger and we can't use Number.isInteger?

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
tetsuharuohzeki: @michaelficarra 

[By the spec (March 12, 2026)](https://tc39.es/proposal-iterator-includes/), step 5-a reject if _skippedElements_ is bigint. Thus here's _toSkip_ (step 9) would be a [extended mathematical values](https://tc39.es/ecma262/#extended-mathematical-value) but it ensures a not bigint.

In this situation, is it leget to accept some value which is `>= Number.MAX_SAFE_INTEGER` as the 2nd argument (_skippedElements_)?  I doubt it's a actual problem where a value precision under this situation which rejects bigint as the 2nd argument.

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
tetsuharuohzeki: I addressed it by switching to use `@Number.@isInteger()`.

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
tetsuharuohzeki: I addressed it.

JSTests/stress/iterator-prototype-includes.js
tetsuharuohzeki: Thanks! I addressed them.

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
michaelficarra: @tetsuharuohzeki There are plenty of integral Numbers between `Number.MAX_SAFE_INTEGER` and `Number.MAX_VALUE`. My suggestion is that you use a bigint for your internal counter, not that you accept a bigint as input.

JSTests/stress/iterator-prototype-includes.js
```diff
@@ -0,0 +1,467 @@
...
+
+    const gen = function* () {
+        yield +0;
+    };
+    const iter = gen();
+    sameValue(callTestTargetFunction(iter, -0), true, TEST_NAME);
+}
+
+{
+    const TEST_NAME = `Use SameValueZero comparison, seach NaN`;
```
michaelficarra:
```suggestion
    const TEST_NAME = `Use SameValueZero comparison, search NaN`;
```

And 5 other occurrences.

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
tetsuharuohzeki: I addressed it.

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
```diff
@@ -472,6 +472,73 @@ function windows(windowSize /*, undersized */)
     return @iteratorHelperCreate(generator, iterated);
 }
 
+// https://tc39.es/proposal-iterator-includes/#sec-iterator.prototype.includes
+// FIXME: This `Method` suffix is just a workaround to the following compile error:
+//  ```
+//  Source/JavaScriptCore/builtins/BuiltinNames.h:271:55: error: class member cannot be redeclared
+//  ```
+@overriddenName="includes"
+function includesMethod(searchElement/*, skippedElements */)
```
michaelficarra: Why can't it be done like this?

```suggestion
function includes(searchElement /*, skippedElements */)
```

michaelficarra (commented):
LGTM

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
tetsuharuohzeki: It's just a workaround for the compile error:

    ```
    Source/JavaScriptCore/builtins/BuiltinNames.h:271:55: error: class member cannot be redeclared
    271 |     JSC_COMMON_PRIVATE_IDENTIFIERS_EACH_PROPERTY_NAME(DECLARE_BUILTIN_IDENTIFIER_ACCESSOR_IN_JSC)
    ```

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
```diff
@@ -472,6 +472,73 @@ function windows(windowSize /*, undersized */)
...
+    var iterated = this;
+
+    if (!@isObject(iterated))
+        @throwTypeError("Iterator.prototype.includes requires that |this| be an Object.");
+
+    var wrapper = { @@iterator: function () { return iterated; }};
+    var item;
+
+    var skippedElements = @argument(1);
+    var toSkip = 0n;
```
darinadler: I want to be sure we have a high performance implementation of this function; ideally it’s faster than hand-written loops, for example. How are we measuring the performance? What are the performance properties of always using `BigInt` here? Obviously we need this to work well for even super-large search elements, but do we have good enough optimizations for `BigInt` so that this will be quite fast on not such huge ones?

I generally don’t understand our performance strategy; how do we make the right decisions on C++ vs. JavaScript, for example?

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
kmiller68: I think we should change the spec to throw if you pass a toSkip bigger than `Number.MAX_SAFE_INTEGER`. It's effectively a gonna infinite loop your program anyway (It would take 4 months to get through 9007199254740991 elements if each iteration took 1ns). For now we should just throw a `StackOverflow` since we're always free to do that and go back to using a `Number`.

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
kmiller68: > I want to be sure we have a high performance implementation of this function; ideally it’s faster than hand-written loops, for example. How are we measuring the performance? What are the performance properties of always using `BigInt` here? Obviously we need this to work well for even super-large search elements, but do we have good enough optimizations for `BigInt` so that this will be quite fast on not such huge ones?

I commented on this [here](https://github.com/WebKit/WebKit/pull/60548#discussion_r2967134851). BigInt is definitely not the right call here, addition is much slower and each iteration does an allocation. While it implements the spec correctly, I think we should change the spec since passing anything big enough to `toSkip` where a `Number` would break is likely a bug anyway.

> I generally don’t understand our performance strategy; how do we make the right decisions on C++ vs. JavaScript, for example?

We generally use JS when inlining could be beneficial as it's not possible to profile enough information to do any inlining from C++ intrinsics today. We explored making inlining from C++ intrinsics possible but it didn't really pan out enough to justify carrying it over the finish line.

Constellation: Using BigInt raises sharp concern about performance. You need to make things faster until we reach to a point of Int52 overflow, and we should fallback to BigInt (or C++ or whatever) when necessary. Current approach is always using BigInt, and that's not acceptable in terms of performance.

Using C++ is always preferred unless the code is involving callbacks / JS function calls because JS -> JS calls are faster and encouraging inlining in DFG / FTL. But even if it is involving JS calls, the thing is more nuanced and depending on what is used in JS and what is used in C++.

But if we cannot implement spec compliant version of code in JS without using BigInt, then 100% we should go to C++.

Constellation (changes requested):
Anyway, BigInt use is not acceptable. We need a spec update, or making the implementation work without sacrificing the performance.

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
Constellation: After safe integer (53bit), numbers cannot be incremented in an expected way. Like, `++` can increment two sometimes. If we are using numbers which can be incremented / decremented as a counter, we should disallow more than `Number.MAX_SAFE_INTEGER` as now numbers cannot work as expected.

Source/JavaScriptCore/builtins/JSIteratorPrototype.js
Constellation: @kmiller68 filed an issue https://github.com/tc39/proposal-iterator-includes/issues/12

Source/JavaScriptCore/runtime/JSCJSValue.h
```diff
@@ -1130,6 +1131,13 @@ inline double JSValue::toIntegerOrInfinity(JSGlobalObject* globalObject) const
     return isnan(d) ? 0.0 : trunc(d) + 0.0;
 }
 
+inline bool JSValue::isFinite(JSGlobalObject* globalObject) const
+{
+    double d = toNumber(globalObject);
+    bool ok = std::isfinite(d);
+    return ok;
+}
```
darinadler: I don’t think we need to add this, because it’s not any better from writing `std::isfinite(value.toNumber(globalObject))` at the call site. I could imagine adding the function if we wanted to incorporate optimizations for other types, but that’s not how we are using it in the new code, so I suggest not adding it.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
```diff
@@ -196,4 +207,105 @@ JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncForEach, (JSGlobalObject* globalObject
...
+
+        if (skippedElementsArg.asNumber() < 0) {
+            iteratorClose(globalObject, thisValue);
+            TRY_CLEAR_EXCEPTION(scope, { });
+            return throwVMRangeError(globalObject, scope, errorMessage);
+        }
+
+        if (!skippedElementsArg.isFinite(globalObject)) {
+            RETURN_IF_EXCEPTION(scope, { });
+            shouldSkipAsInfinity = true;
```
darinadler: If we keep the code like this, I suggest setting `shouldSkip` to `std::numeric_limits<uint64_t>::max()` here and eliminating the `shouldSkipAsInfinity` global entirely.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
```diff
@@ -196,4 +207,105 @@ JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncForEach, (JSGlobalObject* globalObject
...
+    JSValue thisValue = callFrame->thisValue().toThis(globalObject, ECMAMode::strict());
+    RETURN_IF_EXCEPTION(scope, { });
+    if (!thisValue.isObject()) [[unlikely]]
+        return throwVMTypeError(globalObject, scope, "Iterator.prototype.includes requires that |this| be an Object."_s);
+
+    JSValue skippedElementsArg = callFrame->argument(1);
+    bool shouldSkipAsInfinity = false;
+    uint64_t shouldSkip = 0;
+    if (!skippedElementsArg.isUndefined()) {
+        constexpr const ASCIILiteral errorMessage = "Iterator.prototype.includes requires that the second argument is a non-negative integral Number or Infinity."_s;
```
darinadler: No need for both `constexpr` and `const`, so please omit `const`.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
```diff
@@ -196,4 +207,105 @@ JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncForEach, (JSGlobalObject* globalObject
...
+
+        bool isIntegerOrInfinity = skippedElementsArg.isNumber() && !skippedElementsArg.isBigInt() && (skippedElementsArg.isAnyInt() || !skippedElementsArg.isFinite(globalObject));
```
darinadler: Not quite right: The `isBigInt` check here is unnecessary, since `isNumber` is `false` for `BigInt`. Using `isFinite` is incorrect, since it returns false for `NaN`, not just infinity. Also, a more efficient coding style would be to check for the common cases first and not repeat checks. Something like this might be right:
```
    if (skippedElementsArg.isInt32()) {
        auto skippedAsInt32 = skippedElementsArg.asInt32();
        if (skippedAsInt32 < 0)
            return throwRangeError();
        shouldSkip = skippedAsInt32;
    } else if (skippedElementsArg.isDouble()) {
        auto skippedAsDouble = skippedElementsArg.asDouble();
        if (skippedAsDouble < 0)
            return throwRangeError();
        auto skippedAsInt52 = tryConvertToInt52(skippedAsDouble);
        if (skippedAsInt52 != JSValue::notInt52)
            shouldSkip = skippedAsInt52;
        else if (std::isinf(skippedAsDouble))
            shouldSkip = std::numeric_limits<uint64_t>::max();
        else
            return throwTypeError(); // FIXME: This behavior seems like it might be incorrect for very large integral values; we should add a test for that.
    } else
        return throwTypeError();
```

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
```diff
@@ -196,4 +207,105 @@ JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncForEach, (JSGlobalObject* globalObject
...
+        RETURN_IF_EXCEPTION(scope, { });
```
darinadler: This is wrong, there is nothing here that can raise an exception.

darinadler (changes requested):

Source/JavaScriptCore/runtime/JSCJSValue.h
tetsuharuohzeki: I revered it.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
tetsuharuohzeki: I addressed it.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
tetsuharuohzeki: I addressed it.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
tetsuharuohzeki: I addressed it.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
tetsuharuohzeki: After adding the care about `NaN` and remove `skippedElementsArg.isBigInt()`, the code was unified into simple `value.asNumber()` case. 

In this case, I seem that there is no definitive performance if add some branch for value is int32, double, or others.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
```diff
@@ -196,4 +211,110 @@ JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncForEach, (JSGlobalObject* globalObject
     return JSValue::encode(jsUndefined());
 }
 
+// https://tc39.es/proposal-iterator-includes/#sec-iterator.prototype.includes
+JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncIncludes, (JSGlobalObject* globalObject, CallFrame* callFrame))
+{
+    VM& vm = globalObject->vm();
+    ThrowScope scope = DECLARE_THROW_SCOPE(vm);
```
Constellation: Usually we take `auto scope = ...`

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
```diff
@@ -196,4 +211,110 @@ JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncForEach, (JSGlobalObject* globalObject
...
+                return throwVMRangeError(globalObject, scope, errorMessage);
+            }
+
+            // if the 2nd argument is +Infinity, we should consume the iterator to the end.
+            shouldSkip = std::numeric_limits<uint64_t>::max();
+        } else {
+            iteratorClose(globalObject, thisValue);
+            TRY_CLEAR_EXCEPTION(scope, { });
+            return throwVMTypeError(globalObject, scope, errorMessage);
+        }
```
Constellation: Because this should be int32 most of time, we should have a fast path via `if (skippedElementsArg.isInt32()) [[likely]]`.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
```diff
@@ -196,4 +211,110 @@ JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncForEach, (JSGlobalObject* globalObject
...
+
+    JSValue nextMethod = iterationRecord.nextMethod;
+    CallData callData = getCallDataInline(nextMethod);
+
+    std::optional<CachedCall> cachedCallHolder;
+    CachedCall* cachedCall = nullptr;
+    if (callData.type == CallData::Type::JS) [[likely]] {
+        cachedCallHolder.emplace(globalObject, jsCast<JSFunction*>(nextMethod), 0);
+        if (scope.exception()) [[unlikely]]
+            RETURN_IF_EXCEPTION(scope, { });
```
Constellation: You do not need `scope.exception()`.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
```diff
@@ -196,4 +211,110 @@ JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncForEach, (JSGlobalObject* globalObject
...
+    VM& vm = globalObject->vm();
+    ThrowScope scope = DECLARE_THROW_SCOPE(vm);
+
+    JSValue thisValue = callFrame->thisValue().toThis(globalObject, ECMAMode::strict());
+    RETURN_IF_EXCEPTION(scope, { });
+    if (!thisValue.isObject()) [[unlikely]]
+        return throwVMTypeError(globalObject, scope, "Iterator.prototype.includes requires that |this| be an Object."_s);
+
+    JSValue skippedElementsArg = callFrame->argument(1);
+    uint64_t shouldSkip = 0;
```
Constellation: I think using `toSkip` name is better as it is what the spec is using.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
```diff
@@ -196,4 +211,110 @@ JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncForEach, (JSGlobalObject* globalObject
...
+
+        if (isEqual) {
+            found = true;
+            iteratorClose(globalObject, iterationRecord.iterator);
+            TRY_CLEAR_EXCEPTION(scope, { });
+            break;
+        }
+    }
+
+    scope.release();
```
Constellation: This is unnecessary.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
```diff
@@ -196,4 +211,110 @@ JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncForEach, (JSGlobalObject* globalObject
...
+        JSValue nextValue = iteratorValue(globalObject, next);
+        RETURN_IF_EXCEPTION(scope, { });
+
+        bool isEqual = sameValueZero(globalObject, nextValue, searchElement);
+        RETURN_IF_EXCEPTION(scope, { });
+
+        if (isEqual) {
+            found = true;
+            iteratorClose(globalObject, iterationRecord.iterator);
+            TRY_CLEAR_EXCEPTION(scope, { });
```
Constellation: I think returning `return JSValue::encode(jsBoolean(true));` from this clause is simpler.
Then you do not need to have `found`.

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
```diff
@@ -196,4 +211,110 @@ JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncForEach, (JSGlobalObject* globalObject
...
+            break;
+        }
+    }
+
+    scope.release();
+    return JSValue::encode(jsBoolean(found));
```
Constellation: Here is unreachable if we return from the loop. Thus, just put
```
RELEASE_ASSERT_NOT_REACHED();
return { };
```

Source/JavaScriptCore/runtime/JSIteratorPrototype.cpp
```diff
@@ -196,4 +211,110 @@ JSC_DEFINE_HOST_FUNCTION(iteratorProtoFuncForEach, (JSGlobalObject* globalObject
...
+        if (cachedCall) [[likely]] {
+            cachedCall->clearArguments();
+            next = iteratorStepWithCachedCall(globalObject, iterationRecord, cachedCall);
+        } else
+            next = iteratorStep(globalObject, iterationRecord);
+
+        RETURN_IF_EXCEPTION(scope, { });
+
+        if (next.isFalse())
+            break;
```
Constellation: Just returning `return JSValue::encode(jsBoolean(false));`? That's what the spec is doing.
And it is simpler.

Constellation (changes requested):
Using BigInt in builtin JS is not fine.

Constellation (approved):

PR #60644, opened 2026-03-15
[Linux] Move thread suspension into libpas

justinmichaud:
#### cbd0e10e16c04fafaa7423a1076d8ded570de69a
<pre>
[Linux] Move thread suspension into libpas
<a href="https://bugs.webkit.org/show_bug.cgi?id=309968">https://bugs.webkit.org/show_bug.cgi?id=309968</a>

Reviewed by NOBODY (OOPS!).

Let's move thread suspension down into libpas so that the scavenger
can use it too.

This patch moves Yusuke's signal-handler linux implementation down,
re-implements it with c/pas idioms, and then adds a callback to
let WTF collect platform registers without having to move all of that
code to libpas too.

Finally, we enable partial decommit on linux too.

New tests:
-  Suspends thread and clears TLC
Before, on Linux we would be unable to partially decommit TLC memory.

Source/bmalloc/libpas:
- Add the suspension functionality, plus StackPointer and StackBounds

Source/bmalloc:
- Add the non-pas fallback copies, and generic C++ wrapper API

WTF:
- Make WTF either use the bmalloc APIs now, or assert that the results
  are equal

</pre>
https://github.com/WebKit/WebKit/commit/1a1c9311d4967638677f293644cbe5f5dd39b6bd

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/174360 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/48293 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/42074 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/183936 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/132228 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/176225 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/49585 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/47975 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/135638 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/95699 "1 flakes 4 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/177318 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/39072 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/156434 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/116116 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/37713 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/36076 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/32418 "Built successfully") | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/4943 "Passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/148136 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/35926 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/186362 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/35622 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/39570 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/40273 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/144548 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/47960 "Built successfully") | | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/144406 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/48639 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/32546 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/114181 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/39306 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/120578 "Built successfully") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/211395 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/47145 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/10890 "Passed tests") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/55083 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/46548 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/46779 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/46606 "Built successfully") | | |


Constellation (commented):
Maybe we need to duplicate the implementation since Linux has the configuration which does not enable libpas, but still suspend / resume features are mandatory for concurrent GC

Source/bmalloc/libpas/src/libpas/pas_thread_suspend_signal_handler.c
Constellation: We should have a function `pas_thread_yield`, and Darwin should have a bit different definition.
```
#if PAS_OS(DARWIN)
        const mach_msg_timeout_t timeoutInMS = 1;
        thread_switch(MACH_PORT_NULL, SWITCH_OPTION_DEPRESS, timeoutInMS);
#else
        sched_yield();
#endif
```

Source/bmalloc/libpas/src/libpas/pas_thread_suspend.c
Constellation, line 52: We should use pthread_once for `pas_thread_suspend_signal_handler_install` call.

Source/bmalloc/libpas/src/libpas/pas_thread_suspend_signal_handler.h
Constellation: I think we should not have these things. Let's just copy the implementation in libpas side, as we need both because libpas is not always enabled.

Source/cmake/WebKitCompilerFlags.cmake
```diff
@@ -173,8 +173,11 @@ if (COMPILER_IS_GCC_OR_CLANG)
     endif ()
 
     # Warnings to be enabled
-    WEBKIT_PREPEND_GLOBAL_COMPILER_FLAGS(-Wcast-align
-                                         -Wformat-security
+    # -Wcast-align is too noisy on 32-bit
```
mcatanzaro:
```
       -Wcast-align
           Warn whenever a pointer is cast such that the required alignment of the target is increased.  For example, warn if a "char *" is
           cast to an "int *" on machines where integers can only be accessed at two- or four-byte boundaries.

           Warn whenever a pointer is cast such that the required alignment of the target is increased.  For example, warn if a "char *" is
           cast to an "int *" regardless of the target machine.
```

Isn't this _always_ a bug? Maybe not a good one to turn off?

justinmichaud: The problem is that the natural alignment of many tyes (ex: in glib) is larger than a void or int ptr, so we get this all the time. The alternative is to add about 20 or so fixes in random places (possibly including asserts), which is definitely the better solution. On the other hand, since 32-bit is EOL, I think that most people wouldn't want to deal with the extra burden. Either way, this will go into a separate patch once this passes EWS

Constellation (changes requested):
I think this fallback mechanism adds too many duplicate code, making libpas hard to maintain. So if you would like to make libpas buildable, I think we should reorganize it much better way instead of copying files.
I really think that we should just have duplicate thread-suspend implementation, that's easier to maintain than introducing fallback.

Source/bmalloc/libpas/src/libpas/pas_machine_current_stack_pointer.h
```diff
@@ -23,29 +23,35 @@
...
-    EmbedderArrayLike& embeddedArrayLike() const { return m_embeddedArrayLike; }
+#if PAS_X86_64 || PAS_ARM64 || PAS_ARM
 
-protected:
-    JSEmbedderArrayLike(VM& vm, Structure* structure, Ref<EmbedderArrayLike>&& embeddedArrayLike)
-        : Base(vm, structure)
-        , m_embeddedArrayLike(WTF::move(embeddedArrayLike))
-    {
-    }
+static inline __attribute__((__always_inline__)) void* pas_machine_current_stack_pointer(void)
```
Constellation: Use `PAS_ALWAYS_INLINE`.

Tools/TestWebKitAPI/Tests/WTF/CurrentStackPointer.cpp
```diff
@@ -0,0 +1,216 @@
...
+{
+    volatile int buffer[1];
+    buffer[0] = 1;
+    buffer[0] = buffer[0] + 1;
+    return currentStackPointer();
+}
+
+// Uses __attribute__ directly because the ALWAYS_INLINE macro falls back
+// to plain "inline" in debug builds.
+static inline __attribute__((__always_inline__)) void* getStackPointerFromAlwaysInline()
```
Constellation, line 48: Use `ALWAYS_INLINE`.

Source/bmalloc/libpas/src/libpas/pas_machine_current_stack_pointer.h
```diff
@@ -23,29 +23,35 @@
...
+#if PAS_X86_64
+    __asm__ volatile("movq %%rsp, %0" : "=r"(sp) ::);
+#elif PAS_ARM64 && defined(__ILP32__)
+    uint64_t sp64 = 0;
+    __asm__ volatile("mov %0, sp" : "=r"(sp64) ::);
+    sp = (void*)(uintptr_t)sp64;
+#elif PAS_ARM64 || PAS_ARM
+    __asm__ volatile("mov %0, sp" : "=r"(sp) ::);
+#endif
+    return sp;
```
Constellation, line 52: Looks like a lot of original code in StackPointer.cpp are missing. Can you move all of the code?

Source/bmalloc/libpas/src/libpas/pas_thread_suspend_signal_handler.c
```diff
@@ -0,0 +1,233 @@
...
+#include <stdatomic.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+PAS_BEGIN_EXTERN_C;
+
+static bool pas_thread_suspend_signal_handler_installed = false;
+
+static sem_t global_suspend_semaphore;
+static int pas_signal_number_for_suspend = SIGSTOP;
```
Constellation: I don't think `SIGSTOP` was the right signal used for this implementation before. Shouldn't it be SIGUSR1?

Source/WTF/wtf/Threading.h
```diff
@@ -410,13 +411,15 @@ class WTF_CAPABILITY("is current") Thread : public ThreadSafeRefCountedAndCanMak
...
+    unsigned m_suspendCount { 0 };
 #elif USE(PTHREADS)
 #if OS(LINUX)
     ThreadIdentifier m_id { 0 };
 #endif
-    PlatformRegisters* m_platformRegisters { nullptr };
+    std::unique_ptr<bmalloc::api::ThreadSuspendData> m_suspendData;
     unsigned m_suspendCount { 0 };
-#endif
+#endif // OS(WINDOWS)
```
Constellation: These things makes me think that we should also just move Windows part to libpas as well, as libpas is built on Windows too. That makes code much cleaner.
Windows is having API for suspend / resume, so situation is similar to Darwin, so code is not so complicated.

Source/bmalloc/libpas/src/libpas/pas_machine_current_stack_pointer.h
justinmichaud: The problem is that PAS_ALWAYS_INLINE isn't always inline for dev builds, but this function wouldn't be correct if not inlined

Source/bmalloc/libpas/src/libpas/pas_thread_local_cache.c
Constellation: When we use TLS partial decommission, you need to use asymmetric commit / recommit always. So, you need to change it to false (& you also need to change commit part of TLS to `is_symmetric = false` explicitly).

Source/WTF/wtf/posix/ThreadingPOSIX.cpp
```diff
@@ -472,54 +355,33 @@ bool Thread::signal(int signalNumber)
...
     RELEASE_ASSERT_WITH_MESSAGE(this != &Thread::currentSingleton(), "We do not support suspending the current thread itself.");
-#if OS(DARWIN)
-    kern_return_t result = thread_suspend(m_platformThread);
-    if (result != KERN_SUCCESS)
-        return makeUnexpected(result);
-    return { };
-#else
+    ASSERT(!m_suspendData);
+#if BOS(DARWIN) || BOS(LINUX)
+    m_suspendData = std::make_unique<bmalloc::api::ThreadSuspendData>(m_handle, m_stack.origin(), m_stack.end());
```
Constellation: Why do we need this on Darwin? Also, it is POSIX, so this function must support all POSIX environment (Not only Darwin & Linux).

Source/WTF/wtf/posix/ThreadingPOSIX.cpp
```diff
@@ -82,101 +83,11 @@ namespace WTF {
...
-    globalSemaphoreForSuspendResume->post();
-}
-
-#endif // !OS(DARWIN)
-
 void Thread::initializePlatformThreading()
 {
+#if BOS(DARWIN) || BOS(LINUX)
+    bmalloc::api::threadSuspendSignalHandlerInstall();
+#endif
```
Constellation: Ditto. You need to support all POSIX environment.

Source/WTF/wtf/posix/ThreadingPOSIX.cpp
```diff
@@ -472,54 +355,33 @@ bool Thread::signal(int signalNumber)
...
 }
 
 void Thread::resume(const ThreadSuspendLocker&)
 {
-#if OS(DARWIN)
-    thread_resume(m_platformThread);
-#else
+    ASSERT(m_suspendData);
+
+#if BOS(DARWIN) || BOS(LINUX)
```
Constellation: Ditto.

Source/bmalloc/libpas/src/libpas/pas_thread_local_cache.c
justinmichaud: Does that mean that after this change, we will slowly leak commit charge?

Source/WTF/wtf/posix/ThreadingPOSIX.cpp
justinmichaud: I removed this, although that does mean we need some more #if guards

Source/bmalloc/libpas/src/libpas/pas_thread_local_cache.c
```diff
@@ -122,8 +130,8 @@ static void deallocate(pas_thread_local_cache* thread_local_cache)
         thread_local_cache->allocator_index_capacity);
 
     /* If we're doing symmetric decommit, then we need to commit the memory for the TLC now. */
-    pas_page_malloc_commit_without_mprotect(begin, size, /* is_symmetric */ !!PAS_USE_SYMMETRIC_PAGE_ALLOCATION, pas_page_flags_none);
-    
+    pas_page_malloc_commit_without_mprotect(begin, size, is_symmetric, pas_page_flags_none);
```
Constellation: Given that C puts the variable far from this call, let's do,
`/* is_symmetric */ false`, for readability.

Source/bmalloc/libpas/src/libpas/pas_thread_suspend_signal_handler.c
```diff
@@ -0,0 +1,243 @@
...
+    action.sa_sigaction = &pas_thread_suspend_signal_handler_impl;
+    action.sa_flags = SA_RESTART | SA_SIGINFO;
+
+    // Theoretically, this can have race conditions but currently, there is no way to deal with it,
+    // plus, we do not expect that this initialization is executed concurrently with the other
+    // initialization which also installs specific signals. If this is the problem, applications should
+    // change how to initialize things.
+    struct sigaction oldAction;
+    if (sigaction(pas_signal_number_for_suspend, NULL, &oldAction)) {
+        PAS_ASSERT(false, "failed to get existing signal handler");
```
Constellation: PAS_ASSERT_NOT_REACHED?

Source/bmalloc/libpas/src/libpas/pas_thread_suspend_signal_handler.c
```diff
@@ -0,0 +1,243 @@
...
+        pas_thread_suspend_lock_unlock();
+        return;
+    }
+    // It has signal already.
+    if (oldAction.sa_handler != SIG_DFL || (void*)oldAction.sa_sigaction != (void*)SIG_DFL) {
+        PAS_ASSERT(false, "There is already an existing handler for signal %d. Set JSC_SIGNAL_FOR_GC if you want WebKit to use a different signal", pas_signal_number_for_suspend);
```
Constellation: Ditto.

Source/bmalloc/libpas/src/libpas/pas_thread_suspend_signal_handler.c
```diff
@@ -0,0 +1,243 @@
...
+        pas_thread_suspend_lock_unlock();
+        return;
+    }
+    if (sigaction(pas_signal_number_for_suspend, &action, 0)) {
+        PAS_ASSERT(false, "failed to install signal handler");
```
Constellation: Ditto.

Source/bmalloc/libpas/src/libpas/pas_thread_local_cache.c
```diff
@@ -307,6 +315,21 @@ pas_thread_local_cache* pas_thread_local_cache_create(void)
...
+        HANDLE handle = NULL;
+        BOOL ok = DuplicateHandle(GetCurrentProcess(), GetCurrentThread(),
+                                  GetCurrentProcess(), &handle,
+                                  0, FALSE, DUPLICATE_SAME_ACCESS);
+        PAS_ASSERT(ok);
+        thread_local_cache->native_thread_handle = handle;
+    }
+#endif
+
+#if PAS_HAVE(THREAD_SUSPEND)
```
Constellation: This `PAS_HAVE(THREAD_SUSPEND)` should not exist. WebKit / JSC / libpas do not work without suspension, and all supported platforms must implement it. If we are copying all of WTF things to libpas, we should not have this.

PR #60653, opened 2026-03-15
Blocked website bypasses restriction by redirecting to a different domain

jelee53:
#### db6b2e64a5e963539242ba5de7f96c4a8dc2dc7d
<pre>
Blocked website bypasses restriction by redirecting to a different domain
<a href="https://bugs.webkit.org/show_bug.cgi?id=309979">https://bugs.webkit.org/show_bug.cgi?id=309979</a>
<a href="https://rdar.apple.com/147259482">rdar://147259482</a>

Reviewed by Sihui Liu.

Domains that redirect may not properly be blocked. We should correct this so that
redirected sites still apply web content restrictions. We also refactor the
NetworkResourceLoader to asynchronously process filtering.

Tests: Tools/TestWebKitAPI/TestWebKitAPI.xcodeproj/project.pbxproj
       Tools/TestWebKitAPI/Tests/WebKit/WKWebView/ParentalControlsContentFilteringTests.mm

* Source/WebCore/PlatformMac.cmake:
* Source/WebCore/WebCore_Private.modulemap:
* Source/WebCore/loader/ContentFilter.cpp:
(WebCore::ContentFilter::continueAfterWillSendRequest):
* Source/WebCore/platform/cocoa/ParentalControlsContentFilter.h:
* Source/WebCore/platform/cocoa/ParentalControlsContentFilter.mm:
(WebCore::canHandleResponse):
(WebCore::ParentalControlsContentFilter::willSendRequest):
* Source/WebCore/platform/cocoa/ParentalControlsURLFilter.h:
* Source/WebCore/platform/cocoa/ParentalControlsURLFilter.mm:
(WebCore::ParentalControlsURLFilter::hasGlobalFilter):
(WebCore::ParentalControlsURLFilter::setFilterForTesting):
* Source/WebCore/platform/cocoa/ParentalControlsURLFilterParameters.h:
* Source/WebCore/testing/MockParentalControlsURLFilter.h: Copied from Source/WebCore/platform/cocoa/ParentalControlsURLFilterParameters.h.
* Source/WebCore/testing/MockParentalControlsURLFilter.mm: Added.
(WebCore::MockParentalControlsURLFilter::create):
(WebCore::MockParentalControlsURLFilter::MockParentalControlsURLFilter):
(WebCore::MockParentalControlsURLFilter::isEnabledImpl const):
(WebCore::MockParentalControlsURLFilter::isURLAllowedImpl):
(WebCore::MockParentalControlsURLFilter::allowURL):
* Source/WebKit/NetworkProcess/NetworkProcess.cpp:
(WebKit::NetworkProcess::installMockParentalControlsURLFilterForTesting):
* Source/WebKit/NetworkProcess/NetworkProcess.h:
* Source/WebKit/NetworkProcess/NetworkProcess.messages.in:
* Source/WebKit/NetworkProcess/NetworkResourceLoader.cpp:
(WebKit::NetworkResourceLoader::willSendRedirectedRequestInternal):
(WebKit::NetworkResourceLoader::continueWillSendRedirectedRequestAfterContentFiltering):
* Source/WebKit/NetworkProcess/NetworkResourceLoader.h:
* Source/WebKit/Shared/ios/WebParentalControlsURLFilter.mm:
(WebKit::WebParentalControlsURLFilter::setSharedParentalControlsURLFilterIfNecessary):
* Source/WebKit/UIProcess/API/Cocoa/WKWebsiteDataStore.mm:
(-[WKWebsiteDataStore _installMockParentalControlsURLFilterForTestingWithBlockedURLs:completionHandler:]):
* Source/WebKit/UIProcess/API/Cocoa/WKWebsiteDataStorePrivate.h:
* Source/WebKit/UIProcess/Network/NetworkProcessProxy.cpp:
(WebKit::NetworkProcessProxy::installMockParentalControlsURLFilterForTesting):
* Source/WebKit/UIProcess/Network/NetworkProcessProxy.h:
* Source/WebKit/UIProcess/WebsiteData/WebsiteDataStore.cpp:
(WebKit::WebsiteDataStore::installMockParentalControlsURLFilterForTesting):
* Source/WebKit/UIProcess/WebsiteData/WebsiteDataStore.h:
* Tools/TestWebKitAPI/Tests/WebKit/WKWebView/ParentalControlsContentFilteringTests.mm: Added.
(TEST(ParentalControlsContentFilteringTests, BlockedURLAfterRedirect)):
(TEST(ParentalControlsContentFilteringTests, BlockedURLAfterMultipleRedirections_BlockStart)):
(TEST(ParentalControlsContentFilteringTests, BlockedURLAfterMultipleRedirections_BlockMiddle)):
(TEST(ParentalControlsContentFilteringTests, BlockedURLAfterMultipleRedirections_BlockFinal)):

Canonical link: <a href="https://commits.webkit.org/311880@main">https://commits.webkit.org/311880@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/43b0a64351421711952732d4ca0eea1ee3947190

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/158150 "Passed style check") | [loading 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/31487 "Build is in progress. Recent messages:OS: Tahoe (26.2), Xcode: 26.2; Running apply-patch; Checked out pull request; Compiled WebKit (warnings)") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/24681 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/166979 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/112233 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/160021 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/31624 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/31505 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/122476 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85974 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/161108 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/24744 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/142040 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/103145 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/23800 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/22154 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/14751 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/133484 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/19832 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/169468 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/14822 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/21455 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/130657 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/31233 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/26219 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/130772 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/31171 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/141626 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/89067 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/24062 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/25484 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/18432 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/30723 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/96256 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/30244 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/30474 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/30371 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/loader/ContentFilter.cpp
+++ b/Source/WebCore/loader/ContentFilter.cpp
@@ -192,7 +192,10 @@ void ContentFilter::continueAfterWillSendRequest(ResourceRequest&& request, cons
             contentFilterCallbackAggregator->didReceivePlatformContentFilterDecision(platformContentFilter, WTF::move(urlString));
         };
 
-        ASSERT(platformContentFilter->needsMoreData());
+        if (!platformContentFilter->needsMoreData()) {
+            completion({ });
+            continue;
+        }
         platformContentFilter->willSendRequest(ResourceRequest { request }, redirectResponse, WTF::move(completion));
     }
 }
--- a/Source/WebCore/platform/cocoa/ParentalControlsContentFilter.mm
+++ b/Source/WebCore/platform/cocoa/ParentalControlsContentFilter.mm
@@ -93,14 +93,31 @@ ParentalControlsContentFilter::ParentalControlsContentFilter(const PlatformConte
     UNUSED_PARAM(params);
 }
 
-void ParentalControlsContentFilter::willSendRequest(ResourceRequest&&, const ResourceResponse&, CompletionHandler<void(String&&)>&& completionHandler)
-{
-    completionHandler({ });
-}
-
 static inline bool canHandleResponse(const ResourceResponse& response)
 {
     return response.url().protocolIsInHTTPFamily();
+}
+
+void ParentalControlsContentFilter::willSendRequest(ResourceRequest&&, const ResourceResponse& redirectResponse, CompletionHandler<void(String&&)>&& completionHandler)
+{
+#if HAVE(WEBCONTENTRESTRICTIONS)
+    if (redirectResponse.isNull() || !canHandleResponse(redirectResponse) || !enabled()) {
+        completionHandler({ });
+        return;
+    }
+
+    urlFilter()->isURLAllowed(m_mainDocumentURL, redirectResponse.url(), [this, protectedThis = Ref { *this }, evaluatedURL = redirectResponse.url(), completionHandler = WTF::move(completionHandler)](bool isAllowed, NSData *replacementData) mutable {
+        if (!isAllowed) {
+            m_state = State::Blocked;
+            m_evaluatedURL = evaluatedURL;
+            m_replacementData = replacementData;
+        }
+        completionHandler({ });
+    });
+#else
+    UNUSED_PARAM(redirectResponse);
+    completionHandler({ });
+#endif
 }
 
 void ParentalControlsContentFilter::responseReceived(const ResourceResponse& response)
--- a/Source/WebCore/testing/MockParentalControlsURLFilter.h
+++ b/Source/WebCore/testing/MockParentalControlsURLFilter.h
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+ * THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#if HAVE(WEBCONTENTRESTRICTIONS)
+
+#include <WebCore/ParentalControlsURLFilter.h>
+#include <wtf/HashSet.h>
+#include <wtf/URL.h>
+#include <wtf/URLHash.h>
+
+namespace WebCore {
+
+class MockParentalControlsURLFilter final : public ParentalControlsURLFilter {
+public:
+    WEBCORE_TESTSUPPORT_EXPORT static Ref<MockParentalControlsURLFilter> create(Vector<URL>&& blockedURLs);
+    WEBCORE_TESTSUPPORT_EXPORT ~MockParentalControlsURLFilter();
+
+private:
+    explicit MockParentalControlsURLFilter(Vector<URL>&& blockedURLs);
+
+    bool isEnabledImpl() const final;
+    void isURLAllowedImpl(const URL& mainDocumentURL, const URL&, CompletionHandler<void(bool, NSData *)>&&) final;
+    void allowURL(const URL&, CompletionHandler<void(bool)>&&) final;
+
+    HashSet<URL> m_blockedURLs;
+};
+
+} // namespace WebCore
+
+#endif // HAVE(WEBCONTENTRESTRICTIONS)
+
--- a/Source/WebCore/testing/MockParentalControlsURLFilter.mm
+++ b/Source/WebCore/testing/MockParentalControlsURLFilter.mm
@@ -1,0 +1,74 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+ * THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import "config.h"
+#import "MockParentalControlsURLFilter.h"
+
+#if HAVE(WEBCONTENTRESTRICTIONS)
+
+#import <wtf/CompletionHandler.h>
+#import <wtf/MainThread.h>
+#import <wtf/URL.h>
+#import <wtf/WorkQueue.h>
+
+namespace WebCore {
+
+Ref<MockParentalControlsURLFilter> MockParentalControlsURLFilter::create(Vector<URL>&& blockedURLs)
+{
+    return adoptRef(*new MockParentalControlsURLFilter(WTF::move(blockedURLs)));
+}
+
+MockParentalControlsURLFilter::MockParentalControlsURLFilter(Vector<URL>&& blockedURLs)
+{
+    for (auto& url : blockedURLs)
+        m_blockedURLs.add(WTF::move(url));
+}
+
+MockParentalControlsURLFilter::~MockParentalControlsURLFilter() = default;
+
+bool MockParentalControlsURLFilter::isEnabledImpl() const
+{
+    return true;
+}
+
+void MockParentalControlsURLFilter::isURLAllowedImpl(const URL& mainDocumentURL, const URL& url, CompletionHandler<void(bool, NSData *)>&& completionHandler)
+{
+    ASSERT(isMainThread());
+
+    bool isBlocked = m_blockedURLs.contains(url) || (!mainDocumentURL.isEmpty() && m_blockedURLs.contains(mainDocumentURL));
+
+    workQueueSingleton().dispatch([completionHandler = WTF::move(completionHandler), isAllowed = !isBlocked]() mutable {
+        completionHandler(isAllowed, nullptr);
+    });
+}
+
+void MockParentalControlsURLFilter::allowURL(const URL&, CompletionHandler<void(bool)>&& completionHandler)
+{
+    completionHandler(true);
+}
+
+} // namespace WebCore
+
+#endif // HAVE(WEBCONTENTRESTRICTIONS)
--- a/Source/WebKit/NetworkProcess/NetworkProcess.cpp
+++ b/Source/WebKit/NetworkProcess/NetworkProcess.cpp
@@ -135,6 +135,7 @@
 #endif
 
 #if HAVE(WEBCONTENTRESTRICTIONS)
+#include <WebCore/MockParentalControlsURLFilter.h>
 #include <WebCore/ParentalControlsURLFilter.h>
 #endif
 
@@ -3448,6 +3449,13 @@ void NetworkProcess::allowEvaluatedURL(const WebCore::ParentalControlsURLFilterP
 #else
     filter->allowURL(parameters.urlToAllow, WTF::move(completionHandler));
 #endif
+}
+
+void NetworkProcess::installMockParentalControlsURLFilterForTesting(Vector<URL>&& blockedURLs, CompletionHandler<void()>&& completionHandler)
+{
+    Ref mock = WebCore::MockParentalControlsURLFilter::create(WTF::move(blockedURLs));
+    WebCore::ParentalControlsURLFilter::setFilterForTesting(WTF::move(mock));
+    completionHandler();
 }
 #endif
 
--- a/Source/WebKit/UIProcess/API/Cocoa/WKWebsiteDataStore.mm
+++ b/Source/WebKit/UIProcess/API/Cocoa/WKWebsiteDataStore.mm
@@ -1615,6 +1615,19 @@ struct WKWebsiteData {
     });
 }
 
+- (void)_installMockParentalControlsURLFilterForTestingWithBlockedURLs:(NSArray<NSURL *> *)blockedURLs completionHandler:(void(^)(void))completionHandler
+{
+#if HAVE(WEBCONTENTRESTRICTIONS)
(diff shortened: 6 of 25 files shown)


Source/WebCore/platform/cocoa/ParentalControlsContentFilter.mm
```diff
@@ -206,7 +258,8 @@ static inline bool canHandleResponse(const ResourceResponse& response)
         return;
 
     Locker resultLocker { m_resultLock };
-    ASSERT(m_isAllowdByWebContentRestrictions);
+    if (!m_isAllowdByWebContentRestrictions)
```
cdumez: Typo here (in pre-existing code)

Source/WebCore/platform/cocoa/ParentalControlsContentFilter.mm
```diff
@@ -92,14 +92,66 @@
...
+#if HAVE(WEBCONTENTRESTRICTIONS)
+    if (redirectResponse.isNull() || !canHandleResponse(redirectResponse) || !enabled()) {
+        completionHandler({ });
+        return;
+    }
+
+    urlFilter()->isURLAllowed(m_mainDocumentURL, redirectResponse.url(), *this);
+
+    Locker resultLocker { m_resultLock };
+    while (!m_isAllowdByWebContentRestrictions)
```
cdumez: I'm guessing this is a pre-existing pattern and you're just replicating here?

This looks complicated and error-prone. Surely there must be an easier way to write this?

Source/WebCore/platform/cocoa/ParentalControlsContentFilter.mm
```diff
@@ -92,14 +92,66 @@
...
+        m_resultCondition.wait(m_resultLock);
```
cdumez: My big concern here is that we seem to be blocking the main thread until we have a result even though this function can work asynchronously (it takes a CompletionHandler). This seems unnecessarily inefficient. Why hang the main thread?

Source/WebCore/platform/cocoa/ParentalControlsContentFilter.mm
jelee53: Yes, this was a pre-existing pattern.

cdumez: Well, yes and no. As far as I can tell, the existing pattern (only in one place) was calling `waitFor()`, not `wait()`. It thus had a timeout and would not hang the main thread indefinitely at least 🤷🏻 So I do think the existing pattern wasn't good to start (maybe that existing code HAD to really run synchronously for some reason). Here though, this is an asynchronous function with a completion handler so it's not obvious to me that we need to hang.

cdumez: Nice!

jelee53: Okay it has been pushed - thanks for your patience and looking forward to feedback!

Source/WebCore/loader/ContentFilter.cpp
```diff
@@ -188,7 +188,10 @@ void ContentFilter::continueAfterWillSendRequest(ResourceRequest&& request, cons
             contentFilterCallbackAggregator->didReceivePlatformContentFilterDecision(platformContentFilter, WTF::move(urlString));
         };
 
-        ASSERT(platformContentFilter->needsMoreData());
+        if (!platformContentFilter->needsMoreData()) {
```
cdumez, line 195: In several places in this PR, assertions were replaced with early returns without explanation. This is usually not a great sign when we have to "silence" assertions so it is worth explaining why it's OK when we do.

Source/WebCore/platform/cocoa/ParentalControlsContentFilter.mm
```diff
@@ -92,14 +92,27 @@
...
 
-void ParentalControlsContentFilter::willSendRequest(ResourceRequest&&, const ResourceResponse&, CompletionHandler<void(String&&)>&& completionHandler)
+static inline bool canHandleResponse(const ResourceResponse& response)
 {
-    completionHandler({ });
+    return response.url().protocolIsInHTTPFamily();
 }
 
-static inline bool canHandleResponse(const ResourceResponse& response)
+void ParentalControlsContentFilter::willSendRequest(ResourceRequest&&, const ResourceResponse& redirectResponse, CompletionHandler<void(String&&)>&& completionHandler)
```
cdumez, line 101: I *think* this patch trades "redirect URL not checked" for "final URL not checked," which is a regression. The final URL is the one the user actually sees.

ParentalControlsContentFilter was designed for a single evaluation lifecycle. Its m_state starts at Filtering, transitions to Allowed or Blocked once, and that's it. All subsequent ContentFilter operations (responseReceived, addData, finishedAddingData) use forEachContentFilterUntilBlocked, which skips any platform filter where needsMoreData() is false (i.e., any filter no longer in Filtering state).                                                                

The patch adds a real evaluation in willSendRequest during redirects, which sets m_state to Allowed or Blocked. This "uses up" the filter's single decision. When responseReceived later runs for the final response, forEachContentFilterUntilBlocked sees ParentalControls is already decided, skips it, and counts the stale redirect-source decision as the final answer.

**Concrete example: Site A (allowed) redirects to site B (blocked).**
- Before the patch: willSendRequest is a no-op, state stays Filtering. responseReceived evaluates site B → blocked. Correct.
- After the patch: willSendRequest evaluates site A → state becomes Allowed. responseReceived skips ParentalControls (already decided) → site B loads. Bypass.

Possible approaches to fix it (if I'm correct, please double check):
  1. Reset ParentalControls state after the willSendRequest evaluation completes (reset m_state back to Filtering, clear m_evaluatedURL, clear m_isAllowedByWebContentRestrictions) so the response-phase evaluation can proceed normally.
  2. Use a separate code path for redirect evaluation in willSendRequest that doesn't mutate m_state (e.g., a dedicated m_redirectDecision field), reporting the block/allow decision through the completion handler without changing the filter's lifecycle state.
  3. Adjust ContentFilter (the container) to understand that willSendRequest decisions are transient and shouldn't affect subsequent forEachContentFilterUntilBlocked calls.

cdumez: We should also make sure that if there is more than 1 redirect in a row (common on the Web), then ALL the redirect URLs get checked. It's hard for me to tell for sure whether this is working as expected or not but it is easy to get wrong.

jelee53: For your example, the destination site will still be correctly blocked in this code's current state. `ParentalControlsContentFilter::responseReceived` sets the filtering state back to State::Filtering. 

However, you are right that the current code does not work with more than one redirect. I will fix this.

jelee53:
> > > #### [8173f97](https://github.com/WebKit/WebKit/commit/8173f9743ae3644bbce229cc76373dc584724315)
> > > Blocked website bypasses restriction by redirecting to a different domain
> > > https://bugs.webkit.org/show_bug.cgi?id=309979
> > > [rdar://147259482](https://rdar.apple.com/147259482)
> > > Reviewed by NOBODY (OOPS!).
> > > Domains that redirect may not properly be blocked. We should correct this so that
> > > redirected sites still apply web content restrictions. We also refactor the
> > > NetworkResourceLoader to asynchronously process filtering.
> > > No new tests needed.
> > 
> > 
> > @szewai Is there really no way to write a test for this.
> > No new tests needed is a lie since there is a behavior change. Writing a test may not be possible given our current testing facilities and if so, we should use that as justification.
> 
> There's no way to test the system allowlist; but maybe we could create a `MockParentalControlsURLFilter` (which has a configurable allowlist) for testing; like the `MockContentFilter`.

@szewai @cdumez for now, I have updated the description of the PR to explain why we cannot add a test.

Source/WebCore/platform/cocoa/ParentalControlsContentFilter.mm
jelee53: I have reset the parental controls state after a willSendRequest evaluation. I also have tested multiple redirections with this redirection example that corrects the protocol and domain extension.
http://discord.gg -> https://discord.gg -> https://discord.com 


When https://discord.gg and https://discord.com (the middle redirection and the destination) are added to the blocklist, I confirmed that the page is successfully blocked.

Source/WebCore/loader/ContentFilter.cpp
```diff
@@ -138,9 +138,6 @@ ContentFilter::ContentFilterCallbackAggregator::~ContentFilterCallbackAggregator
         return;
     }
 
-    if (RefPtr contentFilter = m_contentFilter; contentFilter && m_numberOfFiltersAllowed == contentFilter->m_contentFilters.size())
```
jelee53, line 145: To explain why this was removed: we need to remove this because this prematurely decides whether or not the state is allowed based on the redirect URL, not the destination URL.

Source/WebCore/loader/ContentFilter.cpp
```diff
@@ -188,7 +187,10 @@ void ContentFilter::continueAfterWillSendRequest(ResourceRequest&& request, cons
             contentFilterCallbackAggregator->didReceivePlatformContentFilterDecision(platformContentFilter, WTF::move(urlString));
         };
 
-        ASSERT(platformContentFilter->needsMoreData());
+        if (!platformContentFilter->needsMoreData()) {
```
jelee53, line 195: To explain why this was removed: `NetworkResourceLoader` now checks if a URL is allowed asynchronously rather than synchronously, changing the timing so various platformContentFilters may update state without deterministic timing. For example, `NetworkExtensionContentFilter` state may be Blocked/Allowed even if `ParentalControlsContentFilter` is still filtering redirects. If a filter's state has already been determined, it should be allowed to proceed with its completionHandler and return.

Source/WebCore/loader/ContentFilter.cpp
jelee53: Added a comment explaining why it was removed.

Source/WebCore/platform/cocoa/ParentalControlsContentFilter.mm
```diff
@@ -92,16 +92,39 @@
...
+#if HAVE(WEBCONTENTRESTRICTIONS)
+    if (redirectResponse.isNull() || !canHandleResponse(redirectResponse) || !enabled()) {
+        completionHandler({ });
+        return;
+    }
+
+    m_willSendRequestCompletionHandler = WTF::move(completionHandler);
+    m_evaluatedURL = redirectResponse.url();
+    m_state = State::Filtering;
+    urlFilter()->isURLAllowed(m_mainDocumentURL, redirectResponse.url(), *this);
```
szewai: Why not just invoke the async version of `isURLAllowed` which takes a `completionHandler`? You could capture the `completionHandler` parameter in a lambda and change state to blocked if result is blocked. Then many of the changes seem unneeded.

Source/WebKit/NetworkProcess/NetworkProcess.cpp
```diff
@@ -3445,6 +3440,13 @@ void NetworkProcess::allowEvaluatedURL(const WebCore::ParentalControlsURLFilterP
     filter->allowURL(parameters.urlToAllow, WTF::move(completionHandler));
 #endif
 }
+
+void NetworkProcess::installMockParentalControlsURLFilterForTesting(Vector<URL>&& blockedURLs, CompletionHandler<void()>&& completionHandler)
+{
+    auto mock = WebCore::MockParentalControlsURLFilter::create(WTF::move(blockedURLs));
```
szewai: `Ref mock`

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/ParentalControlsContentFilteringTests.mm
```diff
@@ -0,0 +1,154 @@
...
+    __block bool didFail = false;
+    [navigationDelegate setDidFailProvisionalNavigation:^(WKWebView *, WKNavigation *, NSError *error) {
+        didFail = true;
+    }];
+    [webView setNavigationDelegate:navigationDelegate.get()];
+
+    [webView loadRequest:server.request("/start"_s)];
+
+    Util::run(&didFail);
+    EXPECT_TRUE(didFail);
```
szewai: Unnecessary `EXPECT_TRUE` -- if `didFail` is not true test will time out. I think we should actually validate the error is expected (content filter error).

Source/WebKit/UIProcess/API/Cocoa/WKWebsiteDataStore.mm
```diff
@@ -1615,6 +1601,22 @@ - (void)_isStorageSuspendedForTesting:(void(^)(BOOL))completionHandler
     });
 }
 
+- (void)_installMockParentalControlsURLFilterForTestingWithBlockedURLs:(NSArray<NSURL *> *)blockedURLs completionHandler:(void(^)(void))completionHandler
+{
+#if HAVE(WEBCONTENTRESTRICTIONS)
+    Vector<URL> urls;
```
szewai: Maybe just `auto urls = makeVector<URL>(blockedURLs);`

Source/WebKit/UIProcess/API/Cocoa/WKWebsiteDataStore.mm
szewai: Duplicate `append` -- urls already has blocked urls with `makeVector`

Source/WebCore/testing/MockParentalControlsURLFilter.mm
szewai, line 63: `CompletionHandler` must be invoked on the same thread it is created, in this case it's main thread.
If you want to make it async you should use `callOnMainRunLoop`

bnham: This does seem surprising, but the caller here is `ParentalControlsURLFilter::isURLAllowed` and it explicitly creates the CompletionHandler with the `CompletionHandlerCallThread::AnyThread` attribute and is written to be invoked on any thread.

jelee53: The completionHandler itself also does nothing but wrap a `callOnMainRunLoop` from the previous caller `ParentalControlsURLFilter::isURLAllowed`. So the "meat" of the completionHandler does run in the main thread. 

For other filters like ParentalControlsURLFilter and WebParentalControlsURLFilter, they usually execute `isURLAllowedImpl` on the work queue thread because that's where they're executing BEK or WCR that may take some time.

Source/WebCore/testing/MockParentalControlsURLFilter.h
pvollan: 2026

Source/WebCore/testing/MockParentalControlsURLFilter.mm
pvollan: 2026

Source/WebCore/loader/ContentFilter.cpp
pvollan: I am curious if this assert failed because this patch is implementing the async version of ParentalControlsContentFilter::willSendRequest. Would it make sense to still call `platformContentFilter->willSendRequest` instead of returning early?

Source/WebCore/testing/MockParentalControlsURLFilter.mm
jelee53: Once I figure out why the API tests are failing, I will change this. Unfortunately, git-webkit pr only works on one commit at a time :(

Source/WebCore/testing/MockParentalControlsURLFilter.h
jelee53: Once I figure out why the API tests are failing, I will change this. Unfortunately, git-webkit pr only works on one commit at a time :(

Source/WebCore/loader/ContentFilter.cpp
jelee53: This assert is definitely failing because it's now async!

I think we need to return if a particular content filter does not need more data because it's already made a decision about if the URL should be blocked or not.

szewai (approved):

PR #60706, opened 2026-03-16
[audit-spi] Swift symbols can change between SDK builds, breaking allowlists

emw-apple:
#### 8a4f2c1e10436b245de90ca58cb16b4740d43ccc
<pre>
[audit-spi] Swift symbols can change between SDK builds, breaking allowlists
<a href="https://bugs.webkit.org/show_bug.cgi?id=309970">https://bugs.webkit.org/show_bug.cgi?id=309970</a>
<a href="https://rdar.apple.com/172677139">rdar://172677139</a>

Reviewed by Mike Wyrzykowski and Sam Sneddon.

Swift makes it relatively easy for library authors to make
source-compatible but ABI-breaking changes, and for SPI (where there is
no implicit contract to retain binary compatibility) we observe changes
to symbol names in between SDK builds, as OS frameworks submit new
versions. This causes our allowlists to go out of date despite no change
in our SPI footprint, disrupting CI and driving frustration with the
tools.

Work towards improving how we declare Swift SPI in allowlists by
supporting a higher-level syntax for naming the Swift declarations we
use. Instead of needing to declare the exact mangled symbol used,
allowlists recognize a &quot;swift-decl&quot; section that lists nominal types and
method names that we are using. For example, instead of writing:

    symbols = [&quot;$s5UIKit16UITextEffectViewC015ReplacementTextC0C5chunk4view8delegate9fromColorAeA0bcF5ChunkC_AcE8Delegate_pSo7UIColorCtcfC&quot;]

we can write:

    swift-decls = [
        { name = &quot;UIKit.UITextEffectView.PonderingEffect.init(chunk:view:lightConfiguration:)&quot; }
    ]

A &quot;partial name mangler&quot; based on the Swift compiler&apos;s own name mangling
spec &lt;<a href="https://github.com/swiftlang/swift/blob/main/docs/ABI/Mangling.rst">https://github.com/swiftlang/swift/blob/main/docs/ABI/Mangling.rst</a>&gt;
parses these nominal types and produces the expected *prefix* that the
real symbol starts with.

Objects which are fully SPI can be matched by only writing the type
name. This entry matches the SomeSPIClass type and all its member
functions:

    { name = &quot;ModuleName.SomeSPIClass&quot; }

Additionally, function declarations can omit their parameter names to
match functions/initializers with arbitrary parameters:

    { name = &quot;UIKit.UITextEffectView.PonderingEffect.init()&quot; }

Entries support some additional fields which customize the partial
symbol generated. A complete example:

    { name = &quot;UIKit.UITextEffectView.ReplacementTextEffect.Delegate.performAnimatedReplacement(for:effect:animation:)&quot;,

      # Maps components of the nominal type to their declaration kinds (struct, protocol, enum, class).
      type_kinds = { &quot;Delegate&quot; = &quot;protocol&quot; },

      # If the declaration is an extension, the extension is part of
      # this module.
      extension = &quot;UIKit&quot;,

      # How deeply nested in the nominal type the extension occurs. For example, this entry matches a symbol
      # `performAnimatedReplacement` on an `extension UIKitCore.UITextEffectView.ReplacementTextEffect.Delegate { ... }` declaration.
      extension_base_depth = 3 },

The sdkdb engine uses these partial symbols to make prefix queries
against the table of imported symbols. The same rules about unused
allowlist entries still apply; a Swift declaration that does not match
against any imported symbols is flagged.

For performance reasons, refactor SDKDB.audit() and split the query into
multiple steps:

1. First, it prepares the table of active conditions.
2. Then, it subtracts the `exports` from `imports` to find potential SPI
   in imported binaries.
3. Then, it joins allowlist entries to `imports` to find potential SPI
   which are covered by an allowlist.
4. The results of steps 2 and 3 are compared to produce the different
   diagnostic classes.

Canonical link: <a href="https://commits.webkit.org/310820@main">https://commits.webkit.org/310820@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/e205d274f3925c8282f5a05f24569778cc468d39

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/155115 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/28375 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/21534 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/163875 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/108629 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [loading 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/e8835ba4-fef5-480c-94ea-a011d0a0da7d) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/156988 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/28515 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/28223 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/120011 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/108629 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [loading 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/2a3c26ed-f8b6-4b5c-ba83-cf1089e3fa1e) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/158074 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/22264 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/139294 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/100704 "Passed tests") | | ⏳ 🛠 vision-apple 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/154435 "Passed tests") | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/21349 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/19397 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/11701 "Built successfully") | | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/131011 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-wk1~~](https://ews-build.webkit.org/#/builders/167/builds/17136 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/166353 "Built successfully") | | 
| | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/10198 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/18745 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/128112 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/27919 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/23436 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/128250 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/27843 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/138929 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/84552 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/23636 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/23121 "Passed tests") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/15724 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/27536 "Built successfully") | [  ~~🛠 mac-safer-cpp~~](https://ews-build.webkit.org/#/builders/120/builds/91639 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [  ~~🛠 tv-sim~~](https://ews-build.webkit.org/#/builders/157/builds/27114 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [  ~~🛠 watch~~](https://ews-build.webkit.org/#/builders/163/builds/27344 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/27187 "Built successfully") | | | |

### DIFF
--- a/Tools/Scripts/libraries/webkitapipy/webkitapipy/allow.py
+++ b/Tools/Scripts/libraries/webkitapipy/webkitapipy/allow.py
@@ -29,7 +29,7 @@ from dataclasses import dataclass, field
 from decimal import Decimal
 from pathlib import Path
 from enum import Enum
-from typing import Any, NamedTuple, Optional, Union
+from typing import Any, Mapping, NamedTuple, Optional, Union
 
 if sys.version_info < (3, 11):
     from webkitapipy._vendor import tomli as tomllib
@@ -54,6 +54,8 @@ class AllowedSPI:
     symbols: list[str]
     selectors: list[Selector]
     classes: list[str]
+    swift_decls: list[SwiftDecl] = field(default_factory=list)
+
     requires: list[str] = field(default_factory=list)
     requires_os: list[RequiredVersion] = field(default_factory=list)
     requires_sdk: list[RequiredVersion] = field(default_factory=list)
@@ -72,6 +74,12 @@ class AllowedSPI:
         operator: str
         version: str
 
+    @dataclass(frozen=True)
+    class SwiftDecl:
+        name: str
+        type_kinds: Optional[Mapping[str, str]] = None
+        extension: Optional[str] = None
+        extension_base_depth: Optional[int] = None
 
 class AllowedReason(StrEnum):
     LEGACY = 'legacy'
@@ -142,6 +150,11 @@ class AllowList:
                 for sym in entry.pop('symbols', []):
                     syms.append(f'_{sym}')
 
+                swift_decls = []
+                for decl in entry.pop('swift-decls', []):
+                    decl = AllowedSPI.SwiftDecl(**decl)
+                    swift_decls.append(decl)
+
                 bugs = AllowedSPI.Bugs(entry.pop('request', None),
                                        entry.pop('cleanup', None))
                 allow_unused = bool(entry.pop('allow-unused', False))
@@ -167,7 +180,8 @@ class AllowList:
                             AllowedSPI.RequiredVersion(platform, op,
                                                        version))
                 allow = AllowedSPI(reason=reason, bugs=bugs, symbols=syms,
-                                   selectors=sels, classes=clss, requires=reqs,
+                                   selectors=sels, classes=clss,
+                                   swift_decls=swift_decls, requires=reqs,
                                    allow_unused=allow_unused,
                                    requires_os=requires_os,
                                    requires_sdk=requires_sdk)
--- a/Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
+++ b/Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
@@ -26,7 +26,6 @@ from __future__ import annotations
 import json
 import os
 import sqlite3
-import sys
 from enum import Enum
 from fnmatch import fnmatch
 from typing import Any, Callable, Iterable, Mapping, NamedTuple, Optional, Union
@@ -35,10 +34,11 @@ from pathlib import Path
 from .macho import APIReport, objc_fully_qualified_method
 from .tbd import TBD
 from .allow import AllowList
+from .swift_mangle import mangle_partial
 
 # Increment this number to force clients to rebuild from scratch, to
 # accomodate schema changes or fix caching bugs.
-VERSION = 9
+VERSION = 10
 
 
 class DeclarationKind(Enum):
@@ -182,15 +182,16 @@ class SDKDB:
             return
         cur.execute('CREATE TABLE input_file(path PRIMARY KEY, hash)')
         cur.execute('CREATE TABLE exports('
-                    '   name, class_name, kind DeclarationKind, '
+                    '   name TEXT, class_name, kind DeclarationKind, '
                     '   input_file REFERENCES input_file(path) '
                     '              ON DELETE CASCADE)')
         cur.execute('CREATE INDEX export_names ON exports (name, kind)')
         cur.execute('CREATE TABLE allow('
-                    '   name, class_name, allow_unused, kind DeclarationKind, '
+                    '   name TEXT, is_prefix, class_name, allow_unused, '
+                    '   kind DeclarationKind, '
                     '   cond_id, input_file REFERENCES input_file(path) '
                     '                       ON DELETE CASCADE)')
-        cur.execute('CREATE INDEX allow_names ON allow (name, kind)')
+        cur.execute('CREATE INDEX allow_names ON allow (name, kind, is_prefix)')
         cur.execute('CREATE TABLE condition_chain(name, '
                     "   op CHECK (op IN ('==', '!=', '<', '<=', '>', '>=')), "
                     '   operand, nextid, '
@@ -203,10 +204,11 @@ class SDKDB:
         cur = self.con.cursor()
         cur.execute('CREATE TEMPORARY TABLE window(input_file)')
         cur.execute('CREATE INDEX selected_files ON window(input_file)')
-        cur.execute('CREATE TEMPORARY TABLE imports(name, '
+        cur.execute('CREATE TEMPORARY TABLE imports(name TEXT, '
                     '   kind DeclarationKind, input_file, arch)')
         cur.execute('CREATE INDEX import_names ON imports(name, kind)')
         cur.execute('CREATE TEMPORARY TABLE condition(name UNIQUE, value)')
+        cur.execute('CREATE TEMPORARY TABLE active_cond(nextid, name TEXT)')
         self.con.commit()
 
     def __del__(self):
@@ -315,16 +317,23 @@ class SDKDB:
     class InsertionKind(Enum):
         EXPORTS = 1
         ALLOW = 2
+        ALLOW_PREFIX = 3
 
         @property
         def statement(self) -> str:
             if self == self.EXPORTS:
-                return (f'INSERT INTO exports VALUES (:name, :class_name, '
-                        '                             :kind, :file)')
-            else:  # self.ALLOW
-                return (f'INSERT INTO allow VALUES (:name, :class_name, '
+                return ('INSERT INTO exports VALUES (:name, :class_name, '
+                        '                            :kind, :file)')
+            elif self == self.ALLOW:
+                return ('INSERT INTO allow VALUES (:name, 0, :class_name, '
                         '                           :allow_unused, :kind, '
                         '                           :cond, :file)')
+            elif self == self.ALLOW_PREFIX:
+                return ('INSERT INTO allow VALUES (:name, 1, :class_name, '
+                        '                          :allow_unused, :kind, '
+                        '                          :cond, :file)')
+            else:
+                raise RuntimeError('unreachable')
 
     def _add_api_report(self, report: APIReport, binary: Path,
                         dest=InsertionKind.EXPORTS):
@@ -416,6 +425,16 @@ class SDKDB:
                                         dest=self.InsertionKind.ALLOW,
                                         cond_id=cond_id,
                                         allow_unused=entry.allow_unused)
+            for decl in entry.swift_decls:
+                mangled = mangle_partial(decl.name, type_kinds=decl.type_kinds,
+                                         extension_module=decl.extension,
+                                         extension_base_depth=decl.extension_base_depth)
+                assert all(c != '*' and c != '?' for c in mangled), \
+                    'Mangled Swift symbol "{mangled}" contains glob characters'
+                self._add_symbol(mangled, allowlist,
+                                 dest=self.InsertionKind.ALLOW_PREFIX,
+                                 cond_id=cond_id,
+                                 allow_unused=entry.allow_unused)
 
     def add_conditions(self, conditions: Mapping[str, ConditionVariable]):
         cur = self.con.cursor()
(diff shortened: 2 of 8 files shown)


mwyrzykowski (approved):

Tools/Scripts/libraries/webkitapipy/webkitapipy/allow.py
```diff
@@ -72,6 +74,12 @@ class RequiredVersion(NamedTuple):
         operator: str
         version: str
 
+    @dataclass(frozen=True)
+    class SwiftDecl:
+        name: str
+        type_kinds: Optional[dict[str, str]] = None
```
gsnedders: Do you want a `Mapping` here, so you have an immutable data type within the frozen dataclass?

Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
```diff
@@ -96,6 +97,10 @@ def semver_to_int(semver: str) -> int:
         return version[0] * 10000
 
 
+def escape_for_like_sql(value: str) -> str:
+    return f'replace(replace(replace({value}, "\\", "\\\\"), "_", "\\_"), "%", "\\%")'
```
gsnedders: This mangling to produce an SQL expression does kinda scare me, but I guess it does keep it within the prepared statement, rather than having to duplicate it for all callsite.

That said, an option that might be worth considering is using `GLOB` instead of `LIKE` — while Mach-O files can (AFAIK) contain any C string in their symbol table, none of the languages we currently use to my knowledge emit symbols containing `*`, `?`, or `[`.

Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
```diff
@@ -416,6 +430,14 @@ def _add_allowlist(self, config: AllowList, allowlist: Path):
                                         dest=self.InsertionKind.ALLOW,
                                         cond_id=cond_id,
                                         allow_unused=entry.allow_unused)
+            for decl in entry.swift_decls:
+                mangled = mangle_partial(decl.name, type_kinds=decl.type_kinds,
+                                         extension_module=decl.extension,
+                                         extension_base_depth=decl.extension_base_depth)  # TODO: adjust
```
gsnedders: Can you expand on what the TODO is here?

Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
```diff
@@ -439,14 +461,14 @@ def _add_imports(self, report: APIReport):
         # different invocation that reads exports from this binary from
         # inserting to the exports table.
         cur.execute('INSERT INTO window VALUES (?)', (path,))
-        cur.executemany('INSERT INTO imports VALUES (?, ?, ?, ?)',
+        cur.executemany(f'INSERT INTO imports VALUES (?, ?, ?, ?)',
```
gsnedders: This introduces an unneeded f-string.

Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
```diff
@@ -439,14 +461,14 @@ def _add_imports(self, report: APIReport):
...
                         ((sym.removeprefix(_OBJC_CLASS_), OBJC_CLS,
                           path, arch) if sym.startswith(_OBJC_CLASS_) else
                          (sym.removeprefix(_OBJC_METACLASS_), OBJC_CLS,
                           path, arch) if sym.startswith(_OBJC_METACLASS_) else
                          (sym, SYMBOL, path, arch)
                          for sym in report.imports))
-        cur.executemany('INSERT INTO imports VALUES (?, ?, ?, ?)',
+        cur.executemany(f'INSERT INTO imports VALUES (?, ?, ?, ?)',
```
gsnedders: This introduces an unneeded f-string.

Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
```diff
@@ -483,7 +505,8 @@ def audit(self) -> Iterable[Diagnostic]:
                     '           aw.input_file IS NOT NULL) as allow_found '
                     'FROM imports AS i '
                     'LEFT JOIN exports AS e USING (name, kind) '
-                    'FULL JOIN allow AS a USING (name, kind) '
+                    # TODO: check optimization
```
gsnedders: Have you done so?

Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
```diff
@@ -525,6 +549,10 @@ def audit(self) -> Iterable[Diagnostic]:
                     yield UnusedAllowedName(name=allowed_name, file=Path(path),
                                             kind=allowed_kind)
             elif allow_found and export_found:
+                if allowed_kind == SYMBOL and allowed_name.endswith('%'):
```
gsnedders: Do we want an assert that `%` never occurs anywhere else in the string?

Tools/Scripts/libraries/webkitapipy/webkitapipy/swift_mangle.py
```diff
@@ -0,0 +1,664 @@
...
+import re
+from dataclasses import dataclass, field
+
+# ---------------------------------------------------------------------------
+# Standard-type substitution tables (from StandardTypesMangling.def)
+# ---------------------------------------------------------------------------
+
+# Swift module standard types: name -> mangling char
+# Emitted as "S<char>"
+STANDARD_TYPES: dict[str, str] = {
```
gsnedders: `Mapping`, given I doubt we want this to be mutable.

Tools/Scripts/libraries/webkitapipy/webkitapipy/swift_mangle.py
```diff
@@ -0,0 +1,664 @@
...
+    "Strideable": "x",
+    "RawRepresentable": "Y",
+    "StringProtocol": "y",
+    "SignedInteger": "Z",
+    "BinaryInteger": "z",
+}
+
+# Concurrency standard types: name -> mangling char
+# Emitted as "Sc<char>"
+STANDARD_TYPES_CONCURRENCY: dict[str, str] = {
```
gsnedders: `Mapping`, given I doubt we want this to be mutable.

Tools/Scripts/libraries/webkitapipy/webkitapipy/swift_mangle.py
```diff
@@ -0,0 +1,664 @@
...
+This enables prefix-matching against real Swift symbols.
+"""
+
+from __future__ import annotations
+
+import re
+from dataclasses import dataclass, field
+
+# ---------------------------------------------------------------------------
+# Standard-type substitution tables (from StandardTypesMangling.def)
```
gsnedders, line 28: If you're going to reference other filenames, you should specify where (the Swift compiler, I presume, rather than LLVM generally).

Tools/Scripts/libraries/webkitapipy/webkitapipy/swift_mangle.py
```diff
@@ -0,0 +1,664 @@
...
+    "MainActor": "M",
+    "TaskPriority": "P",
+    "AsyncStream": "S",
+    "AsyncThrowingStream": "s",
+    "Task": "T",
+    "UnsafeCurrentTask": "t",
+}
+
+# Kind operators for nominal types
+KIND_OPERATORS: dict[str, str] = {
```
gsnedders, line 112: `Mapping`, given I doubt we want this to be mutable.

Tools/Scripts/libraries/webkitapipy/webkitapipy/swift_mangle.py
```diff
@@ -0,0 +1,664 @@
...
+}
+
+# ---------------------------------------------------------------------------
+# Word-boundary helpers (ports isWordStart / isWordEnd from ManglingUtils.h)
+# ---------------------------------------------------------------------------
+
+
+def _is_word_start(ch: str) -> bool:
+    """A word starts at a non-digit, non-underscore, non-NUL character."""
+    return ch != "\0" and ch != "_" and not ch.isdigit()
```
gsnedders, line 126: Maybe add an `assert` that `len(ch) == 1`?

How are we going from Mach-O symbol (a bunch of bytes) to a `str`? Might `str.isdigit()` allow too much, if we the string might contain non-ASCII digits?

Tools/Scripts/libraries/webkitapipy/webkitapipy/swift_mangle.py
```diff
@@ -0,0 +1,664 @@
...
+
+
+def _is_word_end(ch: str, prev: str) -> bool:
+    """A word ends when we hit underscore, NUL, or a lowercase->uppercase transition."""
+    if ch == "_" or ch == "\0":
+        return True
+    if not prev.isupper() and ch.isupper():
+        return True
+    return False
```
gsnedders, line 135: \[Similar comment to `_is_word_start`]

Tools/Scripts/libraries/webkitapipy/webkitapipy/swift_mangle_unittest.py
```diff
@@ -0,0 +1,439 @@
...
+import subprocess
+import unittest
+
+from .swift_mangle import Mangler, mangle_partial
+
+
+class TestIdentifierEncoding(unittest.TestCase):
+    """Test the word-substitution identifier encoding algorithm."""
+
+    def _encode(self, ident: str) -> str:
```
gsnedders: I might expect this to be `_mangle` or something rather than `_encode` (a vastly overloaded term!).

Tools/Scripts/libraries/webkitapipy/webkitapipy/swift_mangle_unittest.py
```diff
@@ -0,0 +1,439 @@
...
+        result = "".join(m.buffer)
+        # "Abc" at pos 0 (new, word 0), "Abc" at pos 3 (matches word 0)
+        # Substitutions: [(3, 0)], prefix '0', then literal "3Abc" + last subst 'A0'
+        self.assertEqual(result, "03AbcA0")
+
+    def test_word_substitution_across_identifiers(self):
+        # Mangling "ModuleName" then "TypeName" should reuse "Name"
+        m = Mangler()
+        m._mangle_identifier("ModuleName")
+        buf_after_first = "".join(m.buffer)
```
gsnedders, line 43: Why doesn't this use `_encode` above?

Tools/Scripts/libraries/webkitapipy/webkitapipy/swift_mangle_unittest.py
```diff
@@ -0,0 +1,439 @@
...
+        buf_after_second = "".join(m.buffer)
+        # "TypeName": Type is new (word 2), Name matches word 1 from buffer
+        # Substitutions: [(4, 1)] -> prefix '0', literal "4Type", last subst 'B0'
+        self.assertEqual(buf_after_second, "10ModuleName04TypeB0")
+
+    def test_underscore_prefix(self):
+        # Identifiers starting with _ should work
+        m = Mangler()
+        m._mangle_identifier("_arg")
+        result = "".join(m.buffer)
```
gsnedders: Why doesn't this use `_encode` above?

Tools/Scripts/libraries/webkitapipy/webkitapipy/swift_mangle_unittest.py
```diff
@@ -0,0 +1,439 @@
...
+    def test_digit_first_literal(self):
+        # If first char of a literal substring is a digit, emit 'X' instead
+        # This is for error recovery
+        m = Mangler()
+        m._mangle_identifier("a1b")
+        # "a" not word (len < 2), "1b" not word start (digit)
+        # Actually: word scan: a=word start, 1=word end -> word "a" (len 1, too short)
+        # Then 1 is not word start (digit), b is word start, \0 is word end -> word "b" (len 1, too short)
+        # No substitutions -> just emit "3a1b"
+        result = "".join(m.buffer)
```
gsnedders: Why doesn't this use `_encode` above?

Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
emw-apple: Ah yes, sorry, forgot to remove. I haven't profiled the difference yet, but query performance is no longer perceptibly slower. I've set `case_sensitive_like` on the connection and believe [the LIKE optimization](https://sqlite.org/optoverview.html#the_like_optimization) now applies; it's certainly faster than when I first started working on this and case sensitivity was off.

Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
emw-apple: I don't think it's necessary—if a symbol in an allowlist _actually_ ended with `%`, then at this point in the code it it would be confused with a swift-decl symbol prefix, because `allowed_name` has been unescaped at this point. But it's fine to put such a symbol into the `seen_partial_symbols` set, because it's only going to be reported on once anyways.

I'll add a comment to clarify what `seen_partial_symbols` is for, in case that helps here.

Tools/Scripts/libraries/webkitapipy/webkitapipy/swift_mangle.py
emw-apple: Well, we're going in the opposite direction — the mangler is giving a human-authored declaration as a Unicode `str`, and translates it to the characters of a Mach-O symbol. If a user provided a non-ASCII declaration, I'm not sure what would happen. It would probably depend on whether the normalized form matches the ASCII representation that we would get from scanning binaries. 

AFAIK, valid Swift symbols never use non-ASCII characters. The language _does_ support non-ASCII declarations, but they are translated to a punycode in the symbol name. This mangler does not attempt to support punycode, and in practice, we don't need it for representing API/SPI on Apple platforms.

Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
emw-apple: At one point I was intending on coming up with a way to infer the `extension_base_depth` from some other part of the declaration. But that's not possible with just the nominal type. So I gave up, and for now I think it's fine to expose this ABI detail to the surface.

Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
emw-apple: GLOB might be better indeed (afaik it's subject to the same optimizations as LIKE), but I think we would still need to do escaping since those characters are allowable in symbol names.

> This mangling to produce an SQL expression does kinda scare me

If it's any consolation, this code is only used when inserting from allowlists, not for arbitrary symbols from binaries or the SDK.

Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
gsnedders: I think per Slack this is unresolved?

Tools/Scripts/libraries/webkitapipy/webkitapipy/sdkdb.py
gsnedders: Ah, and the allowlists of course have unmangled stuff, yeah, that makes sense.

Leaving this unresolved for the GLOB vs. LIKE question, in case you want to reconsider that.

gsnedders (approved):
This looks fine, modulo the performance question which I think might be re-open?

PR #60801, opened 2026-03-17
Move RenderBlock overflow relayout logic out of RenderLayerScrollableArea.

sammygill:
#### d801bd01eb61cf3c6404099d4ba07a56eb7daf26
<pre>
Move RenderBlock overflow relayout logic out of RenderLayerScrollableArea.
<a href="https://bugs.webkit.org/show_bug.cgi?id=312392">https://bugs.webkit.org/show_bug.cgi?id=312392</a>
<a href="https://rdar.apple.com/problem/174850591">rdar://problem/174850591</a>

Reviewed by Alan Baradlay.

The way that RenderLayerScrollableArea::updateScrollInfoAfterLayout is
roughly organized, for the purposes of this change, is as follows:

1.) A preamble that computes the scroll dimensions and origin based off
the renderer&apos;s overflow rect and some styling information (e.g. writing
mode) along with some other state updates. This will have the side
effect of creating or destroying scrollbars.

2.) If there is a change in whether an automatic scrollbar is gained or
lost then we will rerun layout on the renderer. This is because a change
in the scrollbar may affect the available space for the content which
would result in a different layout.

3.) Some final updating like updating the scrollbar steps and notifying
the scroll anchoring controller.

All of this is currently done towards the end of a renderer&apos;s layout,
via the LayoutScope destructor, but this is a bit misleading since as
part of step two we may layout the renderer and its contents again. From
the perspective of the layout code this is pretty surprising since this
is very far away and hidden from the rest of layout. In this patch we
attempt to reconcile this by trying to move the logic contained within
step two outside of RenderLayerScrollableArea a bit.

This should be just a mechanical change with no changes in functional
behavior since it should be some shuffling of code that runs in the same
order but just via a couple of new mechanisms. Those two new mechanisms
are RelayoutScopeForScrollbarChange and ScrollbarUpdateScope.
The idea is that the former will hold the logic related to layout while
the latter does the work needed for RenderLayerScrollableArea. So, now
what happens is:

1.) Instead of calls to RenderBlock::updateScrollInfoAfterLayout, which
finds its way into the RenderLayerScrollableArea variant, we replace
those with a construction of RelayoutScopeForScrollbarChange.

2.) We will perform the preamble and return a ScrollbarUpdateScope
and will hold onto it in RelayoutScopeForScrollbarChange.

3.) When RelayoutScopeForScrollbarChange is destroyed we will do the work
that was included in step two of the original flow.

4.) When ScrollbarUpdateScope is destroyed it
will run the remaining work that is in step three of the original flow.

There are likely a number of ways we can improve on this but this should
hopefully be a good starting point to build off of.

(WebCore::RenderLayerScrollableArea::updateScrollInfoAfterLayout):
1.) This function, including the variants that seem to be mostly
plumbing into this, needs to return an optional to match the various
early returns that could have been hit before. This will serve as the
indicator that we returned early somewhere in this stack so we should
not try to do any of the work that got moved into RelayoutScopeForScrollbarChange.

2.) Move the logic that was in updateScrollbarsAfterLayout into this
function. So now this function should contain all of the previous code
that ran up to the point we started to perform layout on the renderer
again.

* Source/WebCore/rendering/LayoutScope.cpp:
(WebCore::LayoutScope::LayoutScope):
RelayoutScopeForScrollbarChange will need to know if we are already in an
overflow relayout so that we avoid recursion. Previously, this was
tracked via a SetForScope when we entered the RenderLayerScrollableArea
code. Instead when we create a LayoutScope there we will let it know
that we are in overflow relayout, which will pass it into the
RelayoutScopeForScrollbarChange it creates in the destructor.

* Source/WebCore/rendering/RelayoutScopeForScrollbarChange.cpp: Added.
(WebCore::RelayoutScopeForScrollbarChange::RelayoutScopeForScrollbarChange):
In the constructor we will call updateScrollInfoAfterLayout which will
eventually reach RenderLayerScrollableArea. We hold onto the return
value because we want to make sure that the destructor for
ScrollbarUpdateScope only runs after we run the
code that potentially runs layout on the renderer again to make sure we
keep the same flow.

(WebCore::RelayoutScopeForScrollbarChange::~RelayoutScopeForScrollbarChange):
Contains the logic that was ripped out of RenderLayerScrollableArea
related to performing layout on the renderer. Note that the original
code contained an else layout() when the renderer was not a RenderBlock,
but that seems to have been dead code since we never called
updateScrollInfoAfterLayout on a renderer that was not a RenderBlock.

* Source/WebCore/rendering/RelayoutScopeForScrollbarChange.h: Added.
I deleted the copy constructor and assignment operator since it does not
seem like it would make sense to have copies of this object. That would
be like calling updateScrollInfoAfterLayout multiple times which I think
is a bug so this should hopefully make it harder to do accidentally.

* Source/WebCore/rendering/ScrollbarUpdateScope.cpp: Added.
(WebCore::ScrollbarUpdateScope::ScrollbarUpdateScope):
We create this at the end of RenderLayerScrollableArea::updateScrollInfoAfterLayout
right before we would run the code to layout the renderer again. This is
used mostly to hold onto some state that is needed after the renderer
has gone through layout.

(WebCore::ScrollbarUpdateScope::~ScrollbarUpdateScope):
Should just be the remaining code in updateScrollInfoAfterLayout after we
run layout on the renderer.

* Source/WebCore/rendering/ScrollbarUpdateScope.h: Added.
Same idea with the copy constructor and assignment as RelayoutScopeForScrollbarChange.

Canonical link: <a href="https://commits.webkit.org/311471@main">https://commits.webkit.org/311471@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/7c636c44f041172afc8067c7f83426d23d50a426

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/157081 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/30417 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/23608 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/165904 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/111163 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/158952 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/30553 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/30420 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/121656 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/111163 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/160039 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/23908 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/141061 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/102324 "Passed tests") | | [loading 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/542c78ac-c3ce-4c20-98fa-24d7301db2f0) 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/22963 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/21188 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/13676 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/132643 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/18889 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/168389 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/12548 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/20509 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/129782 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/30019 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/25266 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/129890 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35180 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/29942 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/140683 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/87763 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/24724 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/17487 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/29653 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/93667 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/29175 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/29405 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/29302 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/rendering/RelayoutScopeForScrollbarChange.h
+++ b/Source/WebCore/rendering/RelayoutScopeForScrollbarChange.h
@@ -1,0 +1,59 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+ * THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <WebCore/ScrollbarUpdateScope.h>
+#include <wtf/CheckedPtr.h>
+
+namespace WebCore {
+
+class RenderBlock;
+
+enum class InOverflowRelayout : bool;
+
+// When a renderer goes through layout it may gain or lose a scrollbar which may
+// affect the layout of its contents. We learn from ScrollbarUpdaterScope if this
+// happens and use this scope to potentially relayout the renderer in response.
+class RelayoutScopeForScrollbarChange {
+public:
+    RelayoutScopeForScrollbarChange(RenderBlock&, InOverflowRelayout);
+    ~RelayoutScopeForScrollbarChange();
+
+    RelayoutScopeForScrollbarChange(const RelayoutScopeForScrollbarChange&) = delete;
+    RelayoutScopeForScrollbarChange& operator=(const RelayoutScopeForScrollbarChange&) = delete;
+    RelayoutScopeForScrollbarChange(RelayoutScopeForScrollbarChange&&) = default;
+
+private:
+    const CheckedRef<RenderBlock> m_renderBlock;
+    InOverflowRelayout m_inOverflowRelayout;
+
+    // When |this| is being destroyed, whichs means we may have ran layout on
+    // the renderer again, ScrollbarUpdateScope's destructor will know that
+    // the associated renderer's geometry and content's geometry may have changed.
+    std::optional<ScrollbarUpdateScope> m_scrollbarUpdateScope;
+};
+
+}
--- a/Source/WebCore/rendering/RenderLayer.h
+++ b/Source/WebCore/rendering/RenderLayer.h
@@ -510,7 +510,7 @@ public:
     bool isPointInResizeControl(IntPoint localPoint) const;
     IntSize offsetFromResizeCorner(const IntPoint& localPoint) const;
 
-    void updateScrollInfoAfterLayout();
+    std::optional<ScrollbarUpdateScope> updateScrollInfoAfterLayout();
     void updateScrollbarSteps();
 
     // Returns true if this RenderLayer is a candidate for scrolling during scrollIntoView operations.
--- a/Source/WebCore/rendering/RenderTheme.cpp
+++ b/Source/WebCore/rendering/RenderTheme.cpp
@@ -96,6 +96,10 @@
 #include <wtf/Language.h>
 #include <wtf/NeverDestroyed.h>
 
+#if PLATFORM(COCOA)
+#include <wtf/cocoa/RuntimeApplicationChecksCocoa.h>
+#endif
+
 #if ENABLE(SERVICE_CONTROLS)
 #include "ImageControlsMac.h"
 #endif
--- a/Source/WebCore/rendering/ScrollbarUpdateScope.h
+++ b/Source/WebCore/rendering/ScrollbarUpdateScope.h
@@ -1,0 +1,66 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+ * THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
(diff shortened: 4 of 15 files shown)


Source/WebCore/rendering/RenderTheme.cpp
yulunwu2, line 102: Needed?

Source/WebCore/rendering/RenderLayer.h
yulunwu2: order swap intended?

Source/WebCore/rendering/RenderBlockScrollbarUpdater.cpp
yulunwu2, line 68: Do we want the core logic for this class in the destructor?

yulunwu2: Thoughts on having a function in a util file instead?

yulunwu2 (commented):
Added some comments

Source/WebCore/rendering/RenderBlockScrollbarUpdater.cpp
smfr, line 65: Maybe this is the right time to find a different home for this.

Source/WebCore/rendering/RenderTheme.cpp
sammygill: Unfortunately :/

Source/WebCore/rendering/RenderBlockScrollbarUpdater.cpp
sammygill: Yeah I wasn't sure whether I wanted to do this or have a member function like `relayoutIfScrollbarChanged()` or something like that. I went with this since at all sites we currently want to do this but I don't think I feel too strongly if you or anyone else prefers one way vs another!

Source/WebCore/rendering/RenderLayer.h
sammygill: Nope! Changed it back

Source/WebCore/rendering/RenderBlockScrollbarUpdater.cpp
yulunwu2: I'd defer to @smfr or @alanbaradlay on this, I'm not sure what the convention is at webkit for having this kind of logic in the destructor.

sammygill: I think we are definitely closer but I can't think of the right place/way to do it without adding any unnecessary risk for this patch :/

Source/WebCore/rendering/RenderLayerScrollableAreaScrollbarUpdateScope.cpp
alanbaradlay: This is a bit verbose.  Not sure if `RenderLayerScrollableArea` adds any value here.

Source/WebCore/rendering/RenderBlockScrollbarUpdater.cpp
alanbaradlay: `RenderBlockScrollbarUpdater` not sure if `RenderBlock` adds any value here..

Source/WebCore/rendering/RenderLayerScrollableAreaScrollbarUpdateScope.cpp
sammygill: Changed to ScrollbarUpdateScope

Source/WebCore/rendering/RenderBlockScrollbarUpdater.cpp
sammygill: Changed to ScrollbarRelayoutScope

sammygill: actually changed to RelayoutScopeForScrollbarChange

Source/WebCore/rendering/ScrollbarUpdateScope.h
```diff
@@ -0,0 +1,63 @@
...
+#include <wtf/OptionSet.h>
+
+namespace WebCore {
+
+class RenderLayerScrollableArea;
+
+enum class HasHorizontalOverflow : bool { No, Yes };
+enum class HasVerticalOverflow : bool { No, Yes };
+
+class ScrollbarUpdateScope {
```
alanbaradlay, line 42: Could you just add a sentence or 2 to explain what this is about?

Source/WebCore/rendering/RelayoutScopeForScrollbarChange.h
```diff
@@ -0,0 +1,52 @@
...
+#include <WebCore/ScrollbarUpdateScope.h>
+#include <wtf/CheckedPtr.h>
+
+namespace WebCore {
+
+class RenderBlock;
+
+enum class InOverflowRelayout : bool;
+
+class RelayoutScopeForScrollbarChange {
```
alanbaradlay, line 40: Could you just add a sentence or 2 to explain what this is about?

alanbaradlay (approved):

PR #60824, opened 2026-03-18
Fix HTML map element not matching image using id attribute

Ahmad-S792:
#### f8d6173df229a9ed92ba8175d23789e4bfe75bff
<pre>
Fix HTML map element not matching image using id attribute
<a href="https://bugs.webkit.org/show_bug.cgi?id=91228">https://bugs.webkit.org/show_bug.cgi?id=91228</a>
<a href="https://rdar.apple.com/12359382">rdar://12359382</a>

Reviewed by Ryosuke Niwa.

The spec says that parsing a hash-name reference to a map element [1]
should find the first map element in tree order whose name or id
attribute matches the fragment, regardless of HTML versus XML syntax.

WebKit was only using the name attribute for HTML documents (ignoring
id entirely), and for XML documents it used whichever of name/id was
parsed last, ignoring the other.

This patch also fixes parseHTMLHashNameReference to handle the case
where the &apos;#&apos; is the last character in the string, which per spec [2]
should return null, fixing usemap=&quot;#&quot; and usemap=&quot;hash-last#&quot; cases.

Lastly, this also fixes the code that considers both id and name to pass
old attribute values explicitly to removeImageMap, since by the time
attributeChanged is called the element&apos;s attribute is already updated
to the new value.

[1] <a href="https://html.spec.whatwg.org/multipage/image-maps.html#image-map-processing-model">https://html.spec.whatwg.org/multipage/image-maps.html#image-map-processing-model</a>
[2] <a href="https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#syntax-references">https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#syntax-references</a>

* Source/WebCore/html/parser/HTMLParserIdioms.cpp:
(WebCore::parseHTMLHashNameReference): Return nullAtom() for empty fragments.
* Source/WebCore/html/HTMLMapElement.cpp:
(WebCore::HTMLMapElement::attributeChanged): Remove HTML-only restriction;
handle both name and id attributes uniformly; remove redundant base class call.
* Source/WebCore/dom/TreeScope.cpp:
(WebCore::TreeScope::addImageMap): Register by both name and id.
(WebCore::TreeScope::removeImageMap): Unregister by both name and id.
* Source/WebCore/dom/TreeScopeOrderedMap.cpp:
(WebCore::TreeScopeOrderedMap::getElementByMapName): Match by either name or id.
* LayoutTests/imported/w3c/web-platform-tests/html/semantics/embedded-content/image-maps/image-map-processing-model/hash-name-reference-dynamic-expected.txt: Added.
* LayoutTests/imported/w3c/web-platform-tests/html/semantics/embedded-content/image-maps/image-map-processing-model/hash-name-reference-dynamic.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/html/semantics/embedded-content/image-maps/image-map-processing-model/hash-name-reference-id-update-expected.txt: Added.
* LayoutTests/imported/w3c/web-platform-tests/html/semantics/embedded-content/image-maps/image-map-processing-model/hash-name-reference-id-update.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/html/semantics/embedded-content/image-maps/image-map-processing-model/hash-name-reference-expected.txt: Progressions
* LayoutTests/imported/w3c/web-platform-tests/uievents/mouse/mouse_boundary_events_on_image_map-expected.txt: Ditto

Canonical link: <a href="https://commits.webkit.org/309490@main">https://commits.webkit.org/309490@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/816c86702784aa90868931c2646f2cd63068d442

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/150809 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/23567 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/17138 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/159532 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/104243 "Built successfully") | [❌ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/1b981717-d291-4102-9b01-e64a796d72b5) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/152682 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/23999 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/23774 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/116403 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/82656 "Passed tests") | [loading 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/83a78070-f4ea-4455-8a16-8975ae199fc5) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/153769 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/18513 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/135295 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/97131 "Passed tests") | | [❌ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/9666c9f5-30d2-48bd-b622-8e2c368c3fa6) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/17610 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/15561 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/7379 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/127224 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/13219 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/162006 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/5126 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/14776 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/124405 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/23371 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/19608 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/124601 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/33819 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/23360 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/135014 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/79746 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/19671 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/11776 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/22971 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/86771 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/22683 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/22835 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/22737 "Built successfully") | | | |


Source/WebCore/html/HTMLMapElement.cpp
```diff
@@ -86,27 +86,20 @@ RefPtr<HTMLImageElement> HTMLMapElement::imageElement()
 
 void HTMLMapElement::attributeChanged(const QualifiedName& name, const AtomString& oldValue, const AtomString& newValue, AttributeModificationReason attributeModificationReason)
 {
+    if ((name == HTMLNames::idAttr || name == HTMLNames::nameAttr) && isInTreeScope())
+        treeScope().removeImageMap(*this);
```
rniwa: This doesn't do what you think it does. By the time `attributeChanged` is called, the attribute value has already been updated so this will try to remove the element using new ID & name attribute value, which would not work. We should make `addImageMap` and `removeImageMap` take name in addition to the element so that we can pass in `oldValue` here and `newValue` down below. Also, it seems that we don't have a test for this?

rniwa (changes requested):
I don't think we can land this change because of the bug I just pointed out.

Source/WebCore/html/HTMLMapElement.cpp
Ahmad-S792: @rniwa - Good spot, I did bit more testing and definitely we missed the case. Based on testing, I think Chrome might also have small bug and Gecko as well. I have added test now.

Source/WebCore/html/HTMLMapElement.cpp
```diff
@@ -86,27 +86,29 @@ RefPtr<HTMLImageElement> HTMLMapElement::imageElement()
 
 void HTMLMapElement::attributeChanged(const QualifiedName& name, const AtomString& oldValue, const AtomString& newValue, AttributeModificationReason attributeModificationReason)
 {
-    HTMLElement::attributeChanged(name, oldValue, newValue, attributeModificationReason);
+    if (name == HTMLNames::idAttr || name == HTMLNames::nameAttr) {
+        // By this point the attribute value on the element is already newValue,
+        // so we must use oldValue for the changing attribute and read the other
+        // (unchanged) attribute directly from the element.
+        auto oldMapName = m_name;
+        auto oldId = (name == HTMLNames::idAttr) ? oldValue : getIdAttribute();
```
rniwa: Parenthesis is not needed here.

Source/WebCore/html/HTMLMapElement.cpp
```diff
@@ -86,27 +86,29 @@ RefPtr<HTMLImageElement> HTMLMapElement::imageElement()
 
 void HTMLMapElement::attributeChanged(const QualifiedName& name, const AtomString& oldValue, const AtomString& newValue, AttributeModificationReason attributeModificationReason)
 {
-    HTMLElement::attributeChanged(name, oldValue, newValue, attributeModificationReason);
```
rniwa, line 89: We need to keep this call here.

Source/WebCore/html/HTMLMapElement.cpp
```diff
@@ -86,27 +86,29 @@ RefPtr<HTMLImageElement> HTMLMapElement::imageElement()
...
-        AtomString mapName = newValue;
-        if (mapName[0] == '#')
-            mapName = StringView(mapName).substring(1).toAtomString();
-        m_name = WTF::move(mapName);
-        if (isInTreeScope())
-            treeScope().addImageMap(*this);
-    }
+            treeScope().addImageMap(*this, m_name, getIdAttribute());
+    } else
+        HTMLElement::attributeChanged(name, oldValue, newValue, attributeModificationReason);
```
rniwa: No, this isn't right. `HTMLElement::attributeChanged` must be called for both `idAttr` and `nameAttr`.

rniwa (changes requested):
I don't think we can land the PR as is since not calling `HTMLElement::attributeChanged` for id/name change will result in a security bug.

Source/WebCore/html/HTMLMapElement.cpp
```diff
@@ -88,24 +88,25 @@ void HTMLMapElement::attributeChanged(const QualifiedName& name, const AtomStrin
...
-    // Either the id or name will be used depending on the order the attributes are parsed.
-
     if (name == HTMLNames::idAttr || name == HTMLNames::nameAttr) {
-        if (name == HTMLNames::idAttr) {
-            // Call base class so that hasID bit gets set.
-            HTMLElement::attributeChanged(name, oldValue, newValue, attributeModificationReason);
-            if (document().isHTMLDocument())
-                return;
-        }
+        // By this point the attribute value on the element is already newValue,
```
rniwa: I don't think this. comment is necessary.

LayoutTests/imported/w3c/web-platform-tests/html/semantics/embedded-content/image-maps/image-map-processing-model/hash-name-reference-dynamic.html
```diff
@@ -0,0 +1,85 @@
...
+test(() => {
+    const newMap = document.createElement("map");
+    newMap.id = "foo";
+    document.body.appendChild(newMap);
+    assert_equals(document.getElementById("foo"), newMap);
+    newMap.id = "bar";
+    assert_equals(document.getElementById("foo"), null);
+    assert_equals(document.getElementById("bar"), newMap);
+    newMap.remove();
+}, "Changing a map element's id properly updates getElementById");
```
rniwa: I don't think this test case belongs to this test. We should probably put into a separate file.

Source/WebCore/dom/TreeScope.cpp
```diff
@@ -277,24 +277,28 @@ Element* TreeScope::ancestorElementInThisScope(Element* element) const
...
-    m_imageMapsByName->add(name, imageMap, *this);
+
+    if (!name.isNull())
+        m_imageMapsByName->add(name, imageMap, *this);
+    if (!id.isNull() && id != name)
+        m_imageMapsByName->add(id, imageMap, *this);
 }
 
-void TreeScope::removeImageMap(HTMLMapElement& imageMap)
+void TreeScope::removeImageMap(HTMLMapElement& imageMap, const AtomString& name, const AtomString& id)
```
jelee53: Is there a reason that you chose to refactor `name` to be a parameter rather than just retrieving it from the imageMap?

Ahmad-S792, line 293: No particular reason.

Source/WebCore/html/parser/HTMLParserIdioms.cpp
```diff
@@ -454,7 +454,10 @@ AtomString parseHTMLHashNameReference(StringView usemap)
     size_t numberSignIndex = usemap.find('#');
     if (numberSignIndex == notFound)
         return nullAtom();
-    return usemap.substring(numberSignIndex + 1).toAtomString();
+    auto result = usemap.substring(numberSignIndex + 1);
+    if (result.isEmpty())
+        return nullAtom();
```
jelee53, line 459: If result is empty, doesn't `result.toAtomString()` produce a `nullAtom()`?

Ahmad-S792: an empty StringView::toAtomString() produces emptyAtom() (""), not nullAtom().

Source/WebCore/dom/TreeScope.cpp
Ahmad-S792: Do you want me to change this because I am about to push change suggested by Ryosuke for test.

jelee53:
I had a comment asking why we had refactored this so that `name` became a parameter rather than being retrieved from the image-map like normal. 


@Ahmad-S792 - Doesn't matter. I'd prefer to have `name` retrieved from image map like previously because it seems simpler. But if there's a use-case where it doesn't come from the imageMap, then it can be left as-is.

rniwa:
Also, retrieving an attribute value out of an element is O(k) operation where k is the number of attributes on the element since we use Vector to store attributes.

LayoutTests/imported/w3c/web-platform-tests/html/semantics/embedded-content/image-maps/image-map-processing-model/hash-name-reference-dynamic.html
Ahmad-S792: Added as separate file.

rniwa (approved):

PR #60840, opened 2026-03-18
[Site Isolation] Web Inspector: extract frame/loader ID tracking from InspectorPageAgent into IdentifierRegistry

burg:
#### 50fdc88f3dde36d88bcdd1776e70014ad9bdc3b0
<pre>
[Site Isolation] Web Inspector: extract frame/loader ID tracking from InspectorPageAgent into IdentifierRegistry
<a href="https://bugs.webkit.org/show_bug.cgi?id=310162">https://bugs.webkit.org/show_bug.cgi?id=310162</a>

Reviewed by Qianlang Chen.

Introduce an abstract IdentifierRegistry interface and LegacyIdentifierRegistry
implementation to own the frame-to-identifier and loader-to-identifier maps that
were previously private to InspectorPageAgent. PageInspectorController creates
and owns the registry (Ref&lt;&gt;); agents that need frame lookups hold a WeakRef&lt;&gt;.

This is a pure refactoring with no behavioral change for non-SI configurations.
LegacyIdentifierRegistry preserves the existing IdentifiersFactory sequential
counter behavior (&quot;0.1&quot;, &quot;0.2&quot;, ...) identically. Under Site Isolation, a
BackendIdentifierRegistry subclass (added in a follow-up) produces deterministic
type-prefixed IDs derived from FrameIdentifier so UIProcess and WebContent
processes can independently compute matching IDs.

The frontend treats all protocol frameId and loaderId values as opaque strings.
They flow through NetworkManager._frameIdentifierMap (keyed by frameId),
Frame.loaderIdentifier, and are passed to backend commands like
Page.getResourceTree and Network.getResponseBody. No frontend changes are
needed because the string format is never parsed by the frontend -- only
compared for equality and used as map keys.

Agents that previously required InspectorPageAgent (and thus the Page domain to
be enabled) now access frame/loader IDs through the always-available registry
on PageInspectorController. This means PageRuntimeAgent::didClearWindowObjectInWorld
can now resolve frameIds even when the Page domain is disabled. This is safe
because the Runtime agent only dispatches executionContextCreated if its own
domain is enabled and a frontend dispatcher is connected. The frameId lookup
is a prerequisite for the notification, not a trigger.

No new tests: this is a mechanical extraction. All existing inspector/ layout
tests validate that the refactored ID paths produce identical behavior.

* Source/WebCore/inspector/InspectorIdentifierRegistry.h: Added.
(Inspector::IdentifierRegistry): Abstract base with frameId/loaderId/
  frameForId/assertFrame/takeFrame/takeLoader virtual interface.
(Inspector::LegacyIdentifierRegistry): Concrete impl using
  IdentifiersFactory sequential counters and bidirectional hash maps.
* Source/WebCore/inspector/InspectorIdentifierRegistry.cpp: Added.
* Source/WebCore/inspector/PageInspectorController.h:
(WebCore::PageInspectorController::identifierRegistry): Added.
* Source/WebCore/inspector/PageInspectorController.cpp:
  Create LegacyIdentifierRegistry in constructor.
  Delegate frameId/loaderId/frameForId/assertFrame/takeFrame/takeLoader
  to identifierRegistry(). Remove private hash maps.
* Source/WebCore/inspector/InspectorStyleSheet.h:
  Change from CheckedPtr&lt;InspectorPageAgent&gt; to WeakRef&lt;IdentifierRegistry&gt;
  for frame lookups (decouples from Page agent lifecycle).
* Source/WebCore/inspector/InspectorShaderProgram.h:
  Add missing #include &lt;wtf/WeakRef.h&gt;.
* Source/WebCore/Sources.txt:

Canonical link: <a href="https://commits.webkit.org/311409@main">https://commits.webkit.org/311409@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/304d93ee1179e73c9628a3d5c0243f2d94d9aab8

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/156780 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/30116 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/23299 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/165603 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/110862 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/158651 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/30252 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/30119 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/121436 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85290 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/159738 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/23656 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/140785 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/102104 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/22712 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/20923 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/13375 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/132391 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/18614 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/168086 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/12247 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/20234 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/129551 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/29718 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/25001 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/129660 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35141 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/29641 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/140408 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/87445 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/24483 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/17212 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/29350 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/93366 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/28874 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/29104 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/29000 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/inspector/InspectorIdentifierRegistry.cpp
+++ b/Source/WebCore/inspector/InspectorIdentifierRegistry.cpp
@@ -1,0 +1,88 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+ * THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "InspectorIdentifierRegistry.h"
+
+#include <JavaScriptCore/IdentifiersFactory.h>
+#include <WebCore/DocumentLoader.h>
+#include <WebCore/LocalFrame.h>
+#include <wtf/TZoneMallocInlines.h>
+
+namespace Inspector {
+
+WTF_MAKE_TZONE_ALLOCATED_IMPL(IdentifierRegistry);
+WTF_MAKE_TZONE_ALLOCATED_IMPL(LegacyIdentifierRegistry);
+
+LegacyIdentifierRegistry::LegacyIdentifierRegistry() = default;
+LegacyIdentifierRegistry::~LegacyIdentifierRegistry() = default;
+
+WebCore::Frame* LegacyIdentifierRegistry::frameForId(const Protocol::Network::FrameId& frameId)
+{
+    return frameId.isEmpty() ? nullptr : m_identifierToFrame.get(frameId);
+}
+
+Protocol::Network::FrameId LegacyIdentifierRegistry::frameId(const WebCore::Frame* frame)
+{
+    if (!frame)
+        return emptyString();
+    return m_frameToIdentifier.ensure(*frame, [this, frame] {
+        auto identifier = IdentifiersFactory::createIdentifier();
+        m_identifierToFrame.set(identifier, frame);
+        return identifier;
+    }).iterator->value;
+}
+
+Protocol::Network::LoaderId LegacyIdentifierRegistry::loaderId(WebCore::DocumentLoader* loader)
+{
+    if (!loader)
+        return emptyString();
+    return m_loaderToIdentifier.ensure(loader, [] {
+        return IdentifiersFactory::createIdentifier();
+    }).iterator->value;
+}
+
+WebCore::LocalFrame* LegacyIdentifierRegistry::assertFrame(Protocol::ErrorString& errorString, const Protocol::Network::FrameId& frameId)
+{
+    auto* frame = dynamicDowncast<WebCore::LocalFrame>(frameForId(frameId));
+    if (!frame)
+        errorString = "Missing frame for given frameId"_s;
+    return frame;
+}
+
+Protocol::Network::FrameId LegacyIdentifierRegistry::takeFrame(const WebCore::Frame& frame)
+{
+    auto identifier = m_frameToIdentifier.take(frame);
+    if (!identifier.isNull())
+        m_identifierToFrame.remove(identifier);
+    return identifier;
+}
+
+Protocol::Network::LoaderId LegacyIdentifierRegistry::takeLoader(WebCore::DocumentLoader& loader)
+{
+    return m_loaderToIdentifier.take(&loader);
+}
+
+} // namespace Inspector
--- a/Source/WebCore/inspector/InspectorIdentifierRegistry.h
+++ b/Source/WebCore/inspector/InspectorIdentifierRegistry.h
@@ -1,0 +1,122 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+ * THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#include <JavaScriptCore/InspectorProtocolObjects.h>
+#include <wtf/Forward.h>
+#include <wtf/RefCountedAndCanMakeWeakPtr.h>
+#include <wtf/RobinHoodHashMap.h>
+#include <wtf/TZoneMalloc.h>
+#include <wtf/WeakHashMap.h>
+#include <wtf/WeakPtr.h>
+#include <wtf/text/WTFString.h>
+
+namespace WebCore {
+class DocumentLoader;
+class Frame;
+class LocalFrame;
+}
+
+namespace Inspector {
+
+// Abstract interface for mapping WebCore objects to inspector protocol identifier strings.
+// InspectorPageAgent, NetworkAgent, and other agents use these to produce consistent
+// frameId/loaderId values. Under Site Isolation, the subclass can produce deterministic
+// IDs from FrameIdentifier so that UIProcess and WebProcess agree without syncing state.
+class IdentifierRegistry : public RefCountedAndCanMakeWeakPtr<IdentifierRegistry> {
+    WTF_MAKE_TZONE_ALLOCATED(IdentifierRegistry);
+public:
+    virtual ~IdentifierRegistry() = default;
+
+    virtual WebCore::Frame* frameForId(const Protocol::Network::FrameId&) = 0;
+    WEBCORE_EXPORT virtual Protocol::Network::FrameId frameId(const WebCore::Frame*) = 0;
+    virtual Protocol::Network::LoaderId loaderId(WebCore::DocumentLoader*) = 0;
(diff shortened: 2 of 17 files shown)


Source/WebCore/inspector/InspectorIdentifierRegistry.cpp
```diff
@@ -0,0 +1,87 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "InspectorIdentifierRegistry.h"
+
+#include <JavaScriptCore/IdentifiersFactory.h>
+#include <WebCore/DocumentLoader.h>
```
AnujPanta1, line 30: Are we able to have quote-style includes for these ?

Source/WebCore/inspector/InspectorIdentifierRegistry.h
```diff
@@ -0,0 +1,88 @@
...
+    WebCore::Frame* frameForId(const Protocol::Network::FrameId&) final;
+    WEBCORE_EXPORT Protocol::Network::FrameId frameId(WebCore::Frame*) final;
+    Protocol::Network::LoaderId loaderId(WebCore::DocumentLoader*) final;
+    WebCore::LocalFrame* assertFrame(Protocol::ErrorString&, const Protocol::Network::FrameId&) final;
+    Protocol::Network::FrameId takeFrame(WebCore::Frame&) final;
+    Protocol::Network::LoaderId takeLoader(WebCore::DocumentLoader&) final;
+
+private:
+    WeakHashMap<WebCore::Frame, String> m_frameToIdentifier;
+    MemoryCompactRobinHoodHashMap<String, WeakPtr<WebCore::Frame>> m_identifierToFrame;
```
AnujPanta1, line 88: I'm curious, it seems there might be some asymmetry here. It seems only `takeFrame` explicitly cleans `m_frameToIdentifier` and `m_identifierToFrame`. If there are no possible scenarios frame is destroyed without `frameDetached()` then it seems fine, but it possible it might lead to garbage entries ? Maybe a comment saying to explicitly call `frameDetached`

Source/WebCore/inspector/InspectorStyleSheet.cpp
```diff
@@ -1279,7 +1279,7 @@ RefPtr<Inspector::Protocol::CSS::CSSStyleSheetHeader> InspectorStyleSheet::build
         .setDisabled(styleSheet->disabled())
         .setSourceURL(finalURL())
         .setTitle(styleSheet->title())
-        .setFrameId(m_pageAgent->frameId(frame.get()))
+        .setFrameId(m_identifierRegistry->frameId(frame.get()))
```
AnujPanta1, line 1282: if controller destroyed while InspectorStyleSheet still exists would we have issues, maybe changing this to a CheckedPtr would help ?

Source/WebCore/inspector/InspectorIdentifierRegistry.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+WebCore::LocalFrame* LegacyIdentifierRegistry::assertFrame(Protocol::ErrorString& errorString, const Protocol::Network::FrameId& frameId)
+{
+    auto* frame = dynamicDowncast<WebCore::LocalFrame>(frameForId(frameId));
+    if (!frame)
+        errorString = "Missing frame for given frameId"_s;
+    return frame;
+}
+
+Protocol::Network::FrameId LegacyIdentifierRegistry::takeFrame(WebCore::Frame& frame)
```
the-chenergy: Can the `frame` argument and the `loader` argument in the function below be `const`?

Source/WebCore/inspector/InspectorIdentifierRegistry.h
```diff
@@ -0,0 +1,88 @@
...
+    HashMap<WebCore::DocumentLoader*, String> m_loaderToIdentifier;
```
the-chenergy, line 92: I think we should make `DocumentLoader` either a `CanMakeCheckedPtr` or `RefCountedAndCanMakeWeakPtr`, and use a smart key instead. Or add a `FIXME` and address that later, if you wanted to preserve the existing legacy behavior for as much as possible in this patch

Source/WebCore/inspector/agents/InspectorCSSAgent.cpp
```diff
@@ -1251,12 +1249,14 @@ void InspectorCSSAgent::nodesWithPendingLayoutFlagsChangeDispatchTimerFired()
...
-        String newStyleSheetId = String::number(m_lastStyleSheetId++);
-        auto inspectorStyleSheet = InspectorStyleSheetForInlineStyle::create(Ref { m_instrumentingAgents.get() }->enabledPageAgent(), newStyleSheetId, element, Inspector::Protocol::CSS::StyleSheetOrigin::Author, this);
-        m_idToInspectorStyleSheet.set(newStyleSheetId, inspectorStyleSheet.copyRef());
-        return inspectorStyleSheet;
-    }).iterator->value;
+    auto it = m_nodeToInspectorStyleSheet.find(&element);
+    if (it != m_nodeToInspectorStyleSheet.end())
+        return it->value;
+
+    String newStyleSheetId = String::number(m_lastStyleSheetId++);
```
the-chenergy: I think we can take the chance to make this more readable:
```suggestion
    String newStyleSheetId = String::number(m_nextStyleSheetId);
    ++m_nextStyleSheetId;
```
From this usage it seems like the member variable records the next id to use, rather than the last one (if it were the last one, the code would've been `String::number(++m_lastStyleSheetId)`

Source/WebCore/inspector/InspectorStyleSheet.cpp
burg: Good question! The old `m_pageAgent` was a `CheckedPtr<InspectorPageAgent>`, and the new m_identifierRegistry is a `WeakRef<Inspector::IdentifierRegistry>`, so it's actually safer than before -- `WeakRef` will null out if the registry is destroyed while `InspectorStyleSheet` is still alive, rather than trapping on access like `CheckedPtr` would.

Source/WebCore/inspector/InspectorIdentifierRegistry.cpp
burg: WebKit style uses angle brackets for cross-framework includes (e.g., `<JavaScriptCore/IdentifiersFactory.h>`) and quotes for same-project includes.

Source/WebCore/inspector/InspectorIdentifierRegistry.h
burg: `takeFrame` cleans both maps because the frame-to-identifier mapping is bidirectional (`m_frameToIdentifier` + `m_identifierToFrame`). `takeLoader` only has a one-directional map (`m_loaderToIdentifier`), so there's no reverse map to clean. The `WeakHashMap` for frames handles the case where a frame is destroyed without `takeFrame` being called (the weak reference auto-nulls), but `takeFrame` is still needed to clean the reverse `m_identifierToFrame` map eagerly. This asymmetry is intentional.

Source/WebCore/inspector/InspectorIdentifierRegistry.cpp
burg: frame, yes.

For `loaderId()`, the `DocumentLoader*` can't be const because the `HashMap` key is a raw `DocumentLoader*`. Making the parameter const would require casting away const at the `ensure()` call. There's a FIXME in the header about moving `DocumentLoader` to a smart pointer key, which would let us fix this properly.

Source/WebCore/inspector/InspectorIdentifierRegistry.h
burg: Making `DocumentLoader` a `CanMakeCheckedPtr` is a larger change that touches a lot of code outside this patch. Prefer to address it as a follow-up. This is in the header:

```
// FIXME: DocumentLoader should use a smart pointer key (CheckedPtr or RefPtr).
// It currently holds raw DocumentLoader* which prevents making loaderId/takeLoader
// parameters const. See webkit.org/b/310162 for follow-up.
```

Source/WebCore/inspector/agents/InspectorCSSAgent.cpp
burg: I'd rather not make any further drivebys, this code will require further changes for SI to fully work.

Source/WebCore/inspector/InspectorStyleSheet.cpp
```diff
@@ -1279,7 +1279,7 @@ RefPtr<Inspector::Protocol::CSS::CSSStyleSheetHeader> InspectorStyleSheet::build
         .setDisabled(styleSheet->disabled())
         .setSourceURL(finalURL())
         .setTitle(styleSheet->title())
-        .setFrameId(m_pageAgent->frameId(frame.get()))
+        .setFrameId(m_identifierRegistry.ptr() ? m_identifierRegistry->frameId(frame.get()) : emptyString())
```
basuke: `m_identifierRegistry` is declared as `WeakRef<IdentifierRegistry>`, which by definition is never null. The `ptr()` null check here is contradictory — if this could actually be null at this point, it should be a `WeakPtr`; if not (and I believe it can't be, since the registry is owned by `PageInspectorController` which outlives all its agents and their style sheets), then just call `m_identifierRegistry->frameId(frame.get())` unconditionally.

burg: Yeah, this isn't necessary or intentional, I'll remove it. (I think it's leftover from trying Anuj's CheckedPtr suggestion)

Source/WebCore/inspector/InspectorIdentifierRegistry.cpp
the-chenergy: The two includes in WebCore are in the same project, and we should be able to use quotes for those
```cpp
#include "DocumentLoader.h"
#include "InspectorIdentifierRegistry.h" // From above
#include "LocalFrame.h"
```

Source/WebCore/inspector/InspectorResourceUtilities.h
```diff
@@ -39,6 +40,10 @@ class TextResourceDecoder;
 
 namespace Inspector {
 namespace Protocol {
+typedef String ErrorString;
```
the-chenergy, line 43: Can we use `using` here?
```suggestion
using ErrorString = String;
```

the-chenergy (approved):

burg:
I am going to begin merging this PR today.

PR #60841, opened 2026-03-18
[Site Isolation] Web Inspector: add ProxyingNetworkAgent for cross-process network instrumentation

burg:
#### 73d97f544b965a58c49b45851423103e3a957e1c
<pre>
[Site Isolation] Web Inspector: add ProxyingNetworkAgent for cross-process network instrumentation
<a href="https://bugs.webkit.org/show_bug.cgi?id=310163">https://bugs.webkit.org/show_bug.cgi?id=310163</a>

Reviewed by Qianlang Chen.

Under Site Isolation, cross-origin iframes run in separate WebContent processes.
The existing InspectorNetworkAgent only sees network activity within its own
process. This patch adds the cross-process plumbing so Web Inspector&apos;s Network
panel observes all network activity regardless of which process handles it.

On the WebProcess side, WebInspectorBackend is extended with
enableNetworkInstrumentation / disableNetworkInstrumentation IPC, which create
per-frame FrameNetworkAgentProxy instances and call
connectRemoteInstrumentation() on PageInspectorController to register the
frame&apos;s InstrumentingAgents.

NetworkAgentInstrumentation.h defines the abstract interface that both
InspectorNetworkAgent and FrameNetworkAgentProxy implement, allowing
InstrumentingAgents to hold a separate slot for the proxy agent.

Two new classes implement the &quot;octopus agent&quot; pattern used by SI Inspector:

FrameNetworkAgentProxy (WebContent process):
  Implements the NetworkAgentInstrumentation interface -- the same hooks that
  InspectorNetworkAgent uses (willSendRequest, didReceiveResponse, etc.).
  Registered on a per-frame InstrumentingAgents slot so both the in-process
  InspectorNetworkAgent and the cross-process proxy can coexist. Serializes
  each hook&apos;s parameters into IPC messages sent to the UIProcess.

ProxyingNetworkAgent (UIProcess):
  Receives IPC from all WebContent processes, translates parameters back into
  protocol events, and dispatches them to the frontend via the existing Network
  domain dispatcher. Implements NetworkBackendDispatcherHandler for frontend
  commands (enable, disable, setExtraHTTPHeaders, getResponseBody).

When Site Isolation is not enabled, ProxyingNetworkAgent::enable() returns
early (siteIsolationEnabled() check). The existing single-process
InspectorNetworkAgent handles everything as before. No IPC is sent, no
proxies are created. WebKitLegacy keeps using InspectorNetworkAgent.

== Process instrumentation lifecycle ==

ProxyingNetworkAgent tracks instrumented (processID, pageID) pairs with a
reference count in m_instrumentedProcessPageCounts. Multiple frames in the
same process share one EnableNetworkInstrumentation IPC; the last
disableInstrumentationForProcess() call tears down the receiver. The
disable() command is a force-teardown path that bypasses refcounting --
it sends DisableNetworkInstrumentation to all processes and clears all
state unconditionally, because disable() means the Network domain is being
torn down entirely.

== Test plan ==

Four new layout tests cover cross-origin iframe network instrumentation.
Three of four tests are expected to fail in this patch because they depend
on deterministic protocol IDs and frontend routing fixes added in the
follow-up patch (<a href="https://bugs.webkit.org/show_bug.cgi?id=310164).">https://bugs.webkit.org/show_bug.cgi?id=310164).</a> The
fourth test (request-id-uniqueness) passes because it only checks ID
collision, not frame routing.

* Source/WebCore/inspector/hooks/NetworkAgentInstrumentation.h: Added.
  Abstract interface for network instrumentation hooks, parallel to
  InspectorNetworkAgent&apos;s hook surface. Enables InstrumentingAgents to
  hold both the in-process agent and the cross-process proxy simultaneously.
* Source/WebCore/inspector/InstrumentingAgents.h:
  Add networkAgentProxy slot alongside existing networkAgent. Since Site
  Isolation is runtime-enabled, both agent implementations can be active
  simultaneously. This is safe because they use separate resource data
  storage systems (NetworkResourcesData vs BackendResourceDataStore).
* Source/WebKit/UIProcess/Inspector/Agents/ProxyingNetworkAgent.h: Added.
* Source/WebKit/UIProcess/Inspector/Agents/ProxyingNetworkAgent.cpp: Added.
  UIProcess aggregator agent. Handles enable/disable lifecycle, IPC
  receiver registration, and Network protocol event dispatch.
* Source/WebKit/UIProcess/Inspector/Agents/ProxyingNetworkAgent.messages.in: Added.
  IPC messages from FrameNetworkAgentProxy to ProxyingNetworkAgent:
  RequestWillBeSent, ResponseReceived, DataReceived, LoadingFinished,
  LoadingFailed, RequestServedFromMemoryCache.
* Source/WebKit/WebProcess/Inspector/FrameNetworkAgentProxy.h: Added.
* Source/WebKit/WebProcess/Inspector/FrameNetworkAgentProxy.cpp: Added.
  WebContent-side proxy. Implements NetworkAgentInstrumentation hooks and
  forwards each event via IPC to the UIProcess ProxyingNetworkAgent.
* Source/WebKit/WebProcess/Inspector/WebInspectorBackend.h:
* Source/WebKit/WebProcess/Inspector/WebInspectorBackend.cpp:
(WebKit::WebInspectorBackend::enableNetworkInstrumentation):
  Create FrameNetworkAgentProxy per local frame, call
  connectRemoteInstrumentation() to register InstrumentingAgents.
(WebKit::WebInspectorBackend::disableNetworkInstrumentation):
  Tear down all proxies, call disconnectRemoteInstrumentation().
(WebKit::WebInspectorBackend::ensureInstrumentationForFrame):
  Create proxy for a specific frame with correct WebAgentContext.
* Source/WebKit/WebProcess/Inspector/WebInspectorBackend.messages.in:
  Add EnableNetworkInstrumentation, DisableNetworkInstrumentation.
* Source/WebKit/UIProcess/Inspector/WebPageInspectorController.h:
* Source/WebKit/UIProcess/Inspector/WebPageInspectorController.cpp:
  Own ProxyingNetworkAgent, wire enable/disable to forEachWebContentProcess.
* Source/WebKit/Sources.txt:
* Source/WebKit/SourcesCocoa.txt:
* LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-resource-appears-in-network-domain.html: Added.
  Verifies a fetch() in a cross-origin iframe produces a Resource in the
  Network panel via WI.Frame.Event.ResourceWasAdded. Expected to fail
  until deterministic IDs land (bug 310164).
* LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-resource-appears-in-network-domain-expected.txt: Added.
* LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-network-loading-finished.html: Added.
  Verifies requestWillBeSent -&gt; responseReceived -&gt; loadingFinished event
  sequence for a successful fetch, and requestWillBeSent -&gt; loadingFailed
  for an aborted fetch. Expected to fail until deterministic IDs land.
* LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-network-loading-finished-expected.txt: Added.
* LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-network-events-have-correct-frame-id.html: Added.
  Verifies frameId on cross-origin network events differs from the main
  frame&apos;s ID. Expected to fail until deterministic IDs land.
* LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-network-events-have-correct-frame-id-expected.txt: Added.
* LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-request-id-uniqueness.html: Added.
  Verifies requestIds from different WebContent processes don&apos;t collide.
* LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-request-id-uniqueness-expected.txt: Added.
* LayoutTests/http/tests/site-isolation/inspector/network/resources/iframe-with-fetch.html: Added.
  Helper page that triggers a fetch() for cross-origin network tests.
* LayoutTests/http/tests/site-isolation/inspector/network/resources/iframe-with-abort.html: Added.
  Helper page that triggers an aborted fetch() for loading-finished test.

Canonical link: <a href="https://commits.webkit.org/313207@main">https://commits.webkit.org/313207@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/f101a4c31c637f99d42e0bad19b3e0378daaf0cb

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/162400 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/35876 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/28992 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/171338 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/118845 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/164269 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/35980 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/35878 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/126024 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/118845 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/165359 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/28370 "Passed tests") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/145893 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/106602 "Passed tests") | | ⏳ 🛠 vision-apple 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/161720 "Passed tests") | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/27417 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/25942 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/19038 "Built successfully") | | 
| [⏳ 🛠 🧪 jsc ](https://ews-build.webkit.org/#/builders/JSC-Tests-EWS "Waiting in queue, processing has not started yet") | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/137120 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/23624 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/173842 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/21155 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/25268 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/134332 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/35550 "Built successfully") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/30032 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/134332 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36264 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/35534 "Built successfully") | [❌ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/145420 "") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/97789 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/28866 "Passed tests") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/22251 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/35043 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/102795 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/34545 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/34790 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/34693 "Built successfully") | | | |


Source/WebCore/inspector/WebInjectedScriptManager.h
```diff
@@ -25,8 +25,8 @@
 
 #pragma once
 
-#include "CommandLineAPIHost.h"
 #include <JavaScriptCore/InjectedScriptManager.h>
+#include <WebCore/CommandLineAPIHost.h>
```
the-chenergy: Was this change necessary?

Source/WebCore/inspector/agents/InspectorNetworkAgent.h
```diff
@@ -51,7 +51,7 @@ class ConsoleMessage;
 class InjectedScriptManager;
 class PendingInterceptRequest;
 class PendingInterceptResponse;
-enum class ResourceType;
+enum class ResourceType : uint8_t;
 struct Intercept;
 }
```
the-chenergy, line 58: Did you mean to make InspectorNetworkAgent inherit from NetworkAgentInstrumentation, like the commit message suggested?

Also I notice there's no changes to the protocol—Network.json still defines the Network domain to be provided by the page target only, not web-page target. Is this to be done in a follow-up? (Should we clarify this in the commit message?)

Source/WebCore/inspector/hooks/NetworkAgentInstrumentation.h
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright (C) 2025 Apple Inc. All rights reserved.
```
the-chenergy: 2025 -> 2026. Also check these files:
```
Source/WebKit/Shared/InspectorNetworkTypes.serialization.in
Source/WebKit/UIProcess/Inspector/Agents/ProxyingNetworkAgent.h
Source/WebKit/UIProcess/Inspector/Agents/ProxyingNetworkAgent.messages.in
Source/WebKit/WebProcess/Inspector/FrameNetworkAgentProxy.cpp
Source/WebKit/WebProcess/Inspector/FrameNetworkAgentProxy.h
```

Source/WebCore/inspector/hooks/NetworkAgentInstrumentation.h
```diff
@@ -0,0 +1,85 @@
...
+
+// NetworkAgentInstrumentation is the abstract interface for network instrumentation hooks.
+// Both InspectorNetworkAgent (WebCore) and NetworkAgentProxy (WebKit) implement this
+// interface. InstrumentingAgents has a separate slot for the proxy so that both the
+// in-process agent and the cross-process proxy can coexist.
+//
+// It inherits InspectorWebAgentBase so subclasses can register with AgentRegistry,
+// and AbstractCanMakeCheckedPtr so InstrumentingAgents can hold CheckedPtr to it.
+// Concrete subclasses must use CanMakeCheckedPtr and WTF_OVERRIDE_DELETE_FOR_CHECKED_PTR.
+class NetworkAgentInstrumentation : public WebCore::InspectorAgentBase, public AbstractCanMakeCheckedPtr {
```
the-chenergy, line 63: If the idea is to make `InspectorNetworkAgent` and `ProxyingNetworkAgent` both inherit from this, I suppose it'd make sense to name this class `InspectorNetworkAgentBase`? The "instrumentation" part makes it sound like it relates to `InspectorInstrumentation`, but this class looks more like an agent base

burg: This class is being used like a protocol in Swift- the concrete subclasses conform by implementing all the required hooks. Basically, I want to know the instrumentation hooks for a specific agent / its methods, without digging through InspectorInstrumentation.cpp every time. Similarly, InspectorInstrumentation can't know about all of its consumers since NetworkAgentProxy is defined in WebKit framework.

InspectorInstrumentation is a monolithic WebCore concept which we have already tried splitting up in various ways to make it less annoying to research and change instrumentation hooks. I think grouping hooks by agent is way nicer than the existing InspectorInstrumentationPublic.h and similar sub-headers.

The actual subtype relationship was also necessary to be able to store either implementation in AgentRegistry. Eventually we will have just one slot per instrumentation type, but in the meantime, the interface is clearer than a base class with hook definitions. I tried making it more agent-like-- as you suggested-- and it was challenging to understand which state and methods are actually shared, and which are generated protocol command interfaces. The proxy objects on the WebProcess side don't actually implement protocol commands. My intent is that the protocol-naming will discourage sharing of state in the base class between the legacy NetworkAgent and new NetworkAgentProxy.

LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-network-events-have-correct-frame-id.html
```diff
@@ -0,0 +1,58 @@
...
+            let resourceEvent = await resourceAddedPromise;
+            let resource = resourceEvent.data.resource;
+
+            InspectorTest.expectThat(resource instanceof WI.Resource, "Resource should be created.");
+
+            let resourceFrameId = resource.parentFrame.id;
+            InspectorTest.expectNotEqual(resourceFrameId, mainFrameId, "Resource frame ID should not equal main frame ID.");
+            InspectorTest.expectThat(resourceFrameId !== null && resourceFrameId !== undefined, "Resource frame ID should not be null.");
+            InspectorTest.log(`Main frame ID: ${frameIdMap.get(mainFrameId)}`);
+            InspectorTest.log(`Resource frame ID: ${frameIdMap.get(resourceFrameId)}`);
```
the-chenergy, line 49: We can just directly check that the resource's frame id matches the added iframe's id, to be more self-documenting and also matches the test case's description:
```suggestion
            InspectorTest.expectEqual(resourceFrameId, frameTarget.identifier, "Resource frame ID should match the iframe's identifier.");
            InspectorTest.expectNotEqual(resourceFrameId, mainFrameId, "Resource frame ID should not equal main frame ID.");
            InspectorTest.expectThat(resourceFrameId !== null && resourceFrameId !== undefined, "Resource frame ID should not be null.");
```

LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-resource-appears-in-network-domain.html
```diff
@@ -0,0 +1,52 @@
...
+
+            // Trigger the fetch inside the cross-origin iframe via RuntimeAgent.
+            await frameTarget.RuntimeAgent.evaluate.invoke({expression: "triggerFetch()", objectGroup: "test"});
+
+            let resourceEvent = await resourceAddedPromise;
+            let resource = resourceEvent.data.resource;
+
+            InspectorTest.expectThat(resource instanceof WI.Resource, "Resource should be created.");
+            InspectorTest.expectThat(resource.url.includes("data.json"), "Resource URL should contain 'data.json'.");
+            InspectorTest.expectNotEqual(resource.parentFrame, WI.networkManager.mainFrame, "Resource parent frame should not be the main frame.");
```
the-chenergy, line 43: Similarly here, also check the resource's frame id is the added frame id:
```suggestion
            InspectorTest.expectEqual(resource.parentFrame.id, frameTarget.identifier, "Resource's parent frame should be the added iframe.");
            InspectorTest.expectNotEqual(resource.parentFrame, WI.networkManager.mainFrame, "Resource parent frame should not be the main frame.");
```

Source/WebKit/UIProcess/Inspector/Agents/ProxyingNetworkAgent.cpp
```diff
@@ -0,0 +1,428 @@
...
+    m_instrumentedProcessPageCounts.remove(it);
+    webProcess.send(Messages::WebInspectorBackend::DisableNetworkInstrumentation { }, pageID);
+    webProcess.removeMessageReceiver(Messages::ProxyingNetworkAgent::messageReceiverName(), pageID);
+}
+
+CommandResult<void> ProxyingNetworkAgent::enable()
+{
+    // FIXME: <https://webkit.org/b/308890> Only needed under Site Isolation; without it,
+    // InspectorNetworkAgent in the single WebContent process handles network events.
+    Ref<WebKit::WebPageProxy> inspectedPage = m_inspectedPage.get();
```
the-chenergy:
```suggestion
    Ref inspectedPage = m_inspectedPage.get();
```

Source/WebKit/UIProcess/Inspector/Agents/ProxyingNetworkAgent.cpp
```diff
@@ -0,0 +1,428 @@
...
+    if (!m_enabled)
+        return { };
+
+    m_enabled = false;
+
+    // Force-teardown: disable all processes unconditionally, bypassing refcount
+    // discipline in disableInstrumentationForProcess(). This is correct because
+    // disable() is called when the Network domain is torn down entirely --
+    // no per-frame refcount preservation is needed.
+    Ref<WebKit::WebPageProxy> inspectedPage = m_inspectedPage.get();
```
the-chenergy:
```suggestion
    Ref inspectedPage = m_inspectedPage.get();
```

Source/WebKit/WebProcess/Inspector/FrameNetworkAgentProxy.cpp
```diff
@@ -0,0 +1,314 @@
...
+
+    auto timestamp = MonotonicTime::now().secondsSinceEpoch().value();
+    auto walltime = WallTime::now().secondsSinceEpoch().value();
+    auto documentURL = loader->url().string();
+
+    protect(WebProcess::singleton().parentProcessConnection())->send(
+        Messages::ProxyingNetworkAgent::RequestWillBeSent(
+            resourceID, *frameID, *contextID, String(), documentURL, request,
+            std::nullopt, ResourceType::Other, timestamp, walltime),
+        page->identifier());
```
the-chenergy, line 202: Why are we always sending `ResourceType::Other` rather than mapping from the `UncachedLoadType` passed in? Is this to be updated in a follow-up?

Source/WebKit/WebProcess/WebPage/WebPage.cpp
```diff
@@ -961,6 +961,11 @@ WebPage::WebPage(PageIdentifier pageID, WebPageCreationParameters&& parameters)
             frame->inspectorController().siteIsolationFirstEnabled();
     }
 
+    if (parameters.shouldEnableNetworkInstrumentation) {
+        if (auto* backend = inspector(LazyCreationPolicy::CreateIfNeeded))
+            backend->enableNetworkInstrumentation();
+    }
```
the-chenergy, line 985: This runs during `WebPage` construction and before the inspector target's connection (i.e. before `PageInspectorController::connectFrontend` is called). `enableNetworkInstrumentation` calls into `connectRemoteInstrumentation` and into `InspectorInstrumentation` hooks, which seems like it's done at a time when the frontend is not connected yet. Why is that OK, given:
1. The lazy agents in that target hasn't been created yet, so the page's `m_instrumentingAgents` aren't fully populated yet;
2. There will be double registration in the instrumentation hooks: `frontendCreated` and `registerInstrumentingAgents` each get called here, and then again by `connectFrontend`?

Should we defer this work, or only run this if the target has been connected?

Source/WebKit/WebProcess/Inspector/WebInspectorBackend.cpp
```diff
@@ -303,4 +306,64 @@ void WebInspectorBackend::updateDockingAvailability()
...
+}
+
+void WebInspectorBackend::enableNetworkInstrumentation()
+{
+    if (!m_page || !m_page->corePage())
+        return;
+
+    if (!m_networkInstrumentationEnabled) {
+        m_networkInstrumentationEnabled = true;
+        m_page->corePage()->settings().setDeveloperExtrasEnabled(true);
```
the-chenergy: Tying to the ordering issue I brought up [here,](https://github.com/WebKit/WebKit/pull/60841/changes/2fb78425113f99c9255c858319089220c3619d05#r3083254809) we may not need to do this here if we let the page connects to the target first, right?

Source/WebKit/UIProcess/Inspector/WebPageInspectorController.cpp
```diff
@@ -107,8 +108,11 @@ void WebPageInspectorController::connectFrontend(Inspector::FrontendChannel& fro
 
     m_frontendRouter->connectFrontend(frontendChannel);
 
-    if (connectingFirstFrontend)
+    if (connectingFirstFrontend) {
         m_agents.didCreateFrontendAndBackend();
+        if (m_networkAgent)
+            m_networkAgent->didCreateFrontendAndBackend();
```
the-chenergy: Why don't we add the network agent to `m_agents`, to have this called automatically (and also in other functions, `disconnectFrontend` and `disconnectAllFrontends`)?

Source/WebKit/UIProcess/Inspector/WebPageInspectorController.cpp
```diff
@@ -279,14 +288,25 @@ void WebPageInspectorController::didCreateFrame(WebFrameProxy& frame)
...
 }
 
 void WebPageInspectorController::willDestroyFrame(const WebFrameProxy& frame)
 {
     if (!shouldManageFrameTargets())
         return;
 
+    if (m_networkAgent && m_networkAgent->isEnabled()) {
+        // const_cast is safe; webPageIDInCurrentProcess() is a non-mutating lookup.
+        if (auto pageID = const_cast<WebFrameProxy&>(frame).webPageIDInCurrentProcess())
```
the-chenergy: Can we make `WebFrameProxy::webPageIDInCurrentProcess` `const` to really enforce this and benefit future callers?

Source/WebKit/WebProcess/Inspector/FrameNetworkAgentProxy.cpp
burg: It's complicated and related to the BackendResourceDataStore where we save the record of the type. `UncachedLoadType` doesn't map cleanly to protocol `ResourceType`. Adding a FIXME to revisit once the proxy has richer context.

Source/WebKit/UIProcess/Inspector/WebPageInspectorController.cpp
burg: Applied

Source/WebKit/WebProcess/WebPage/WebPage.cpp
burg: This provides one exception to the no-frontends no-instrumentation rule. If `WebPageCreationParameters` ->`parameters.shouldEnableNetworkInstrumentation` is set, then there is already a frontend connection to UIProcess/`web-page` target, we are preloading that fact into the WebPage in WebProcess so that we don't miss early Network instrumentations. These would be missed otherwise, precisely because of the no-frontends rule.

The alternative to this design that I considered would be to enable/disable instrumentation for the entire process, which seems worse. If the initial willSendRequest is not instrumented in a cross-origin frame, it is very hard to track the lifecycle correctly. Down the line we'll need to support request interceptions too.

Source/WebKit/WebProcess/Inspector/WebInspectorBackend.cpp
burg: This is required with the WebPage constructor change to catch early willSendRequest.

Source/WebCore/inspector/agents/InspectorNetworkAgent.h
burg: It was supposed to inherit, i'll fix this.

The frontend isn't ready for this change yet, so the Network.json change to add `web-page` won't be until the next PR.

Source/WebKit/UIProcess/Inspector/WebPageInspectorController.cpp
burg: The main reason is because it needs to be RefCounted to be a message receiver. To me it seemed easier to special case two agents rather than introduce a refcounted bridge object to be a message receiver, or try to make all UIProcess agents refcounted.
```
void WebPageInspectorController::createLazyAgents()
{
    if (m_didCreateLazyAgents)
        return;

    m_didCreateLazyAgents = true;
    auto webPageContext = webPageAgentContext();
    m_agents.append(makeUniqueRef<InspectorBrowserAgent>(webPageContext));

    if (protect(protect(m_inspectedPage)->preferences())->siteIsolationEnabled()) {
        // ProxyingNetworkAgent and ProxyingPageAgent are RefCounted (for IPC MessageReceiver)
        // so they can't be stored in AgentRegistry which expects UniqueRef ownership.
        // Their lifecycle (didCreateFrontendAndBackend / willDestroyFrontendAndBackend) is
        // managed explicitly in connectFrontend / disconnectFrontend / disconnectAllFrontends.
        m_networkAgent = adoptRef(*new Inspector::ProxyingNetworkAgent(webPageContext));
        m_pageAgent = adoptRef(*new Inspector::ProxyingPageAgent(webPageContext));
    }
}
```

Source/WebKit/UIProcess/Inspector/WebPageInspectorController.cpp
```diff
@@ -324,6 +343,17 @@ void WebPageInspectorController::didCommitProvisionalFrame(WebFrameProxy& frame,
 
     if (auto oldTarget = m_targets.take(oldTargetID))
         targetAgent->targetDestroyed(protect(*oldTarget));
+
+    // Instrument the new process for network events now that the
+    // frame has committed in its final process.
+    // FIXME: We should also disable instrumentation for the old process that was
+    // hosting this frame before the process swap. Currently the old (processID, pageID)
+    // pair retains a refcount in m_instrumentedProcessPageCounts and the IPC message
+    // receiver stays registered until the old process is torn down.
```
the-chenergy, line 405: Can you file a Bugzilla and include a link for that?

Source/WebKit/WebProcess/WebPage/WebPage.cpp
the-chenergy: Can you add a comment in `PageInspectorController::connectRemoteInstrumentation` to clarify that we're calling `frontendConnected`, despite there being no frontends yet, specifically to make sure `InspectorInstrumentation` stop ignoring and start collecting network events?

the-chenergy (approved):

Source/WebKit/UIProcess/Inspector/WebPageInspectorController.cpp
burg: Filed webkit.org/b/313015

burg:
Latest push addresses all outstanding feedback. So close!

PR #60842, opened 2026-03-18
[Site Isolation] Web Inspector: deterministic Network IDs, event routing, and frontend fixes for cross-origin iframes

burg:
#### 2bfe8aece710c090c98d8fe093c106b405f97f83
<pre>
[Site Isolation] Web Inspector: deterministic Network IDs, event routing, and frontend fixes for cross-origin iframes
<a href="https://bugs.webkit.org/show_bug.cgi?id=310164">https://bugs.webkit.org/show_bug.cgi?id=310164</a>

Reviewed by Qianlang Chen.

This is the companion to bug 310163 (ProxyingNetworkAgent IPC plumbing). That
patch adds the cross-process forwarding infrastructure; this patch makes the
protocol IDs deterministic and fixes the frontend routing so Network events
from remote processes actually appear in the Network panel. The split keeps
IPC/agent structure separate from ID scheme and frontend integration for
reviewability.

Under the SI Inspector architecture, some domains are per-frame (Console,
Runtime, Debugger -- handled via FrameInspectorController in the WebContent
process and routed through the target system) while others are page-level
&quot;octopus&quot; agents (Network, Page -- aggregated in UIProcess via
ProxyingNetworkAgent/ProxyingPageAgent on the MultiplexingBackendTarget).
Network is an octopus domain because the frontend&apos;s NetworkManager maintains
a single unified resource list and frame tree, so all processes&apos; events must
be merged into one coherent stream with non-colliding IDs.

* Source/WebCore/inspector/InspectorIdentifierRegistry.h:
(Inspector::IdentifierRegistry::deterministicFrameId): Added.
  Produce &quot;frame-PID.OID&quot; from FrameIdentifier bit-packing so UIProcess
  and WebContent independently compute matching frame IDs.
(Inspector::IdentifierRegistry::deterministicRequestId): Added.
  Produce &quot;request-PID.OID&quot; from ProcessIdentifier + ResourceLoaderIdentifier.
(Inspector::IdentifierRegistry::deterministicLoaderId): Added.
  Produce &quot;loader-PID.OID&quot; from ScriptExecutionContextIdentifier.
* Source/WebCore/inspector/InspectorIdentifierRegistry.cpp:
  Add BackendIdentifierRegistry implementation. Uses deterministic IDs for
  frameId/requestId, falls back to IdentifiersFactory for loaderId when
  the document is not yet available (FIXME <a href="https://rdar.apple.com/170087346">rdar://170087346</a>).

* Source/JavaScriptCore/inspector/protocol/Network.json:
  Add &quot;web-page&quot; to Network domain targetTypes. This enables the frontend
  to initialize NetworkManager on the MultiplexingBackendTarget, which is
  the page-level target that octopus agents like ProxyingNetworkAgent use
  to dispatch events aggregated from all WebContent processes.

(WebCore::InspectorPageAgent::buildObjectForFrame):
  Emit RemoteFrame stubs in Page.getResourceTree response so the frontend
  creates placeholder frames for cross-origin iframes before Network
  events arrive.
  Add buildObjectForRemoteFrame() helper.

* Source/WebCore/inspector/agents/page/PageNetworkAgent.cpp:
(WebCore::PageNetworkAgent::enable):
  Return early under Site Isolation. When SI is active,
  ProxyingNetworkAgent handles all network instrumentation across
  processes; enabling PageNetworkAgent would produce duplicate events
  for the main frame&apos;s process.
* Source/WebCore/inspector/agents/page/PageNetworkAgent.h:
  Add siteIsolationFirstEnabled() check.

* Source/WebCore/inspector/PageInspectorController.cpp:
(WebCore::PageInspectorController::siteIsolationFirstEnabled):
  Track first SI activation to gate one-time initialization paths.
* Source/WebCore/inspector/PageInspectorController.h:
  Add m_siteIsolationFirstEnabled flag, accessor.

* Source/WebCore/loader/ResourceLoaderIdentifier.h:
  Add toRawValue() for deterministic ID bit-extraction.

* Source/WebKit/Shared/ProcessQualified.serialization.in:
  Add IPC serialization for ProcessQualified&lt;ResourceLoaderIdentifier&gt;
  (ScopedResourceLoaderIdentifier).

* Source/WebKit/UIProcess/Inspector/Agents/ProxyingNetworkAgent.cpp:
(Inspector::ProxyingNetworkAgent::requestWillBeSent):
(Inspector::ProxyingNetworkAgent::responseReceived):
(Inspector::ProxyingNetworkAgent::dataReceived):
(Inspector::ProxyingNetworkAgent::loadingFinished):
(Inspector::ProxyingNetworkAgent::loadingFailed):
(Inspector::ProxyingNetworkAgent::requestServedFromMemoryCache):
  Use deterministicFrameId/RequestId/LoaderId for all protocol events
  dispatched to the frontend. Extract ProcessIdentifier from
  ScopedResourceLoaderIdentifier for request ID qualification.
* Source/WebKit/UIProcess/Inspector/Agents/ProxyingNetworkAgent.messages.in:
  Change ResourceLoaderIdentifier to ScopedResourceLoaderIdentifier in
  all IPC messages for process-qualified resource IDs.

* Source/WebKit/WebProcess/Inspector/FrameNetworkAgentProxy.cpp:
  Wrap ResourceLoaderIdentifier with ProcessIdentifier to produce
  ScopedResourceLoaderIdentifier for IPC transport.

  Register new frame with WebInspectorBackend when a local frame is
  created under SI, triggering FrameNetworkAgentProxy setup.

  Add ProcessQualified import for IPC code generation.

(WI.NetworkManager.prototype._addNewResourceToFrameOrTarget):
  Lazily create stub WI.Frame objects when Network events arrive for
  deterministic &quot;frame-*&quot; IDs not yet in _frameIdentifierMap. This
  handles the case where Network events arrive before Page.getResourceTree
  completes for dynamically-added cross-origin iframes.
(WI.NetworkManager.prototype.frameDidNavigate):
  Handle null loaderIdentifier on RemoteFrame stubs by adopting the
  first resource&apos;s loaderIdentifier.
  Defer Network agent initialization on MultiplexingBackendTarget until
  the first FrameTarget signals SI is active, preventing premature
  enable on non-SI pages.
* Source/WebInspectorUI/UserInterface/Protocol/MultiplexingBackendTarget.js:
(WI.MultiplexingBackendTarget):
  Add Network to the set of agents initialized on the web-page target.

* LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-request-id-uniqueness.html:
  Add waitForCommittedFrameTarget() helper that handles the provisional
  target lifecycle (initial target -&gt; provisional commit -&gt; committed
  target with execution context).

Canonical link: <a href="https://commits.webkit.org/313621@main">https://commits.webkit.org/313621@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/35ef144aec90bc1971eb051738334fdfef0049bf

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/163600 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/37084 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/30303 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/172540 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/120049 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/165469 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/37415 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/37083 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/127075 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/89583 "layout-tests (failure)") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/166559 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/29348 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/147086 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/107629 "Passed tests") | | ⏳ 🛠 vision-apple 
| [❌ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/162921 "Found 1 webkitpy test failure: webkitpy.scripts.measure_build_time_unittest.MeasureBuildTimeTest.test_clean_and_null_build") | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/28397 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/27029 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/20243 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/20/builds/155751 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/138295 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/24803 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/175045 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/24491 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/27976 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/26466 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/135381 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/36754 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/31131 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/135363 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36490 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/37539 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/146598 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/99600 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/30670 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/23450 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/196002 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/36249 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/109395 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/51100 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/35747 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/1476 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/35997 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/35894 "Built successfully") | | | |


LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-request-id-uniqueness.html
```diff
@@ -0,0 +1,129 @@
...
+            InspectorTest.assert(frameTargetA instanceof WI.FrameTarget, "First frame target should be created.");
+            InspectorTest.assert(frameTargetA.executionContext, "First frame target should have an execution context.");
+
+            // Add second cross-origin iframe and wait for its committed frame target.
+            InspectorTest.evaluateInPage("addIFrameB()");
+            let frameTargetB = await waitForCommittedFrameTarget();
+            InspectorTest.assert(frameTargetB instanceof WI.FrameTarget, "Second frame target should be created.");
+            InspectorTest.assert(frameTargetB.executionContext, "Second frame target should have an execution context.");
+
+            // Trigger fetches in both iframes and collect the resources.
```
the-chenergy: I feel like with the Network domain getting close to completed, we should be able to make use of certain properties of the Network agent to identify the corresponding frame target for each added iframe. (I'm recalling this idea of using Network agent's features to identify the main frame in the second paragraph of [this commit message](https://github.com/WebKit/WebKit/commit/4df7fa78e8e3db4c9b95b0c5eb363c4582fea512))

I'm assuming the reason we're currently always having to (1) create the iframes and (2) wait for its `targetAdded` event, is to be able to identify the targets with the iframes. Ideally we shouldn't need to do all this in every test case unrelated to targets' lifecycle events. We should be able to include the iframes in the HTML in `<body>`, and then be able to identify their targets with source URLs in the tests which should already have agents enabled and ready to be used

All this to say, I'm curious whether we can use that idea to simplify these tests and all future test cases we write. If not, maybe that's something worth thinking about and coming back to, once we implement more of the Network domain 😊

the-chenergy: I added the comment on this line with the thought that ideally this is the place the test case should begin, without having to do all the work before this point

Source/WebCore/inspector/agents/InspectorPageAgent.cpp
```diff
@@ -877,14 +878,41 @@ Ref<Inspector::Protocol::Page::FrameResourceTree> InspectorPageAgent::buildObjec
...
+        if (auto* localChild = dynamicDowncast<LocalFrame>(child)) {
+            childrenArray->addItem(buildObjectForFrameTree(localChild));
             continue;
-        childrenArray->addItem(buildObjectForFrameTree(localChild));
+        }
+        if (auto* remoteChild = dynamicDowncast<RemoteFrame>(child)) {
+            auto stubTree = Inspector::Protocol::Page::FrameResourceTree::create()
+                .setFrame(buildObjectForRemoteFrame(*remoteChild, *frame))
+                .setResources(JSON::ArrayOf<Inspector::Protocol::Page::FrameResource>::create())
+                .release();
```
the-chenergy: Should we make `buildObjectForFrameTree` (this function) take a `Frame*` instead, to do this work if it downcasts to a remote frame? That'll keep the code in this loop simple

Source/WebCore/inspector/InspectorIdentifierRegistry.h
```diff
@@ -59,6 +64,25 @@ class IdentifierRegistry : public RefCountedAndCanMakeWeakPtr<IdentifierRegistry
     virtual Protocol::Network::FrameId takeFrame(WebCore::Frame&) = 0;
     // Called when a document loader is detached; returns the protocol ID that was assigned.
     virtual Protocol::Network::LoaderId takeLoader(WebCore::DocumentLoader&) = 0;
+
+    // Deterministic protocol ID helpers. These produce type-prefixed strings that
+    // encode the ProcessIdentifier, so both UIProcess and WebContent process can
+    // independently compute the same ID from the same typed identifier.
+    // Format: "type-PID.objectID" (e.g., "frame-12345.3", "request-12345.17")
```
the-chenergy: > `Format: "type-PID.objectID"`

Is the part before the dot truly the PID? Based on how I had to implement provisional frame targets in `WebPageInspectorController`, the `FrameIdentifier` doesn't actually encode which process the frame comes from, which is why I had to append the process id to the target id so that the provisional frame can be identified as different by the frontend 🤔

Source/WebCore/inspector/InspectorIdentifierRegistry.h
```diff
@@ -59,6 +64,25 @@ class IdentifierRegistry : public RefCountedAndCanMakeWeakPtr<IdentifierRegistry
...
+    static inline String deterministicFrameId(WebCore::FrameIdentifier frameID)
```
the-chenergy: It feels to me that name "deterministic" isn't adding enough value really, if it's still based on the `FrameIdentifier` in the end (which implies it had to be deterministic to begin with already). Do you think we should rename it simply to `protocolFrameId` (and similarly for the ids below)? That'd match its usage quite nicely, especially if we could find an answer to my comment [above](https://github.com/WebKit/WebKit/pull/60842/changes/6b56851cde284c081a3b2374745e9815a5914e92#r3003370851) and let this `protocolFrameId` be process-wise unique as well, we could use it as the target id as-is too. Having `FrameInspectorTarget(Proxy)` be using this id as the target id should be real nice, given the main purpose of this id registry

Source/WebCore/inspector/PageInspectorController.cpp
```diff
@@ -100,7 +100,7 @@ PageInspectorController::PageInspectorController(Page& page, std::unique_ptr<Ins
     , m_overlay(makeUniqueRefWithoutRefCountedCheck<InspectorOverlay>(*this, inspectorBackendClient.get()))
     , m_executionStopwatch(Stopwatch::create())
     , m_inspectorBackendClient(WTF::move(inspectorBackendClient))
-    , m_identifierRegistry(adoptRef(*new Inspector::LegacyIdentifierRegistry()))
+    , m_identifierRegistry(adoptRef<Inspector::IdentifierRegistry>(*new Inspector::LegacyIdentifierRegistry()))
```
the-chenergy: I think at this point we should add the `create` static method for all these `IdentifierRegistry` classes, to simplify usages like this and also enforce smart pointer usage

Source/WebInspectorUI/UserInterface/Protocol/MultiplexingBackendTarget.js
```diff
@@ -33,14 +33,17 @@ WI.MultiplexingBackendTarget = class MultiplexingBackendTarget extends WI.Target
         const targetId = "multi";
         super(parentTarget, targetId, WI.UIString("Web Page"), WI.TargetType.WebPage, InspectorBackend.backendConnection);
 
-        console.assert(Array.shallowEqual(Object.keys(this._agents), ["Browser", "Target"]));
+        console.assert(Array.shallowEqual(Object.keys(this._agents), ["Browser", "Network", "Target"]));
```
the-chenergy: Let's take the chance to make this more robust by sorting the keys:
```suggestion
        console.assert(Array.shallowEqual(Object.keys(this._agents).sort(), ["Browser", "Network", "Target"]));
```
Since `Object.keys` returns the keys in insertion order, right now this is basically relying on `InspectorBackendCommands.js` defining the domains in that order, but we don't have to given that's so many layers away

Source/WebInspectorUI/UserInterface/Controllers/NetworkManager.js
```diff
@@ -1200,6 +1209,23 @@ WI.NetworkManager = class NetworkManager extends WI.Object
...
+        // treated as the main resource of a brand-new frame (firing FrameWasAdded).
+        if (!frame && frameIdentifier.startsWith("frame-")) {
+            let mainResource = new WI.Resource("about:blank");
+            frame = new WI.Frame(frameIdentifier, frameOptions.name, frameOptions.securityOrigin, null, mainResource);
+            this._frameIdentifierMap.set(frame.id, frame);
+            mainResource.markAsFinished();
+            if (this._mainFrame)
+                this._mainFrame.addChildFrame(frame);
+            this._dispatchFrameWasAddedEvent(frame);
+        }
```
the-chenergy: I see the purpose of this explained in the commit message:
> `Lazily create stub WI.Frame objects when Network events arrive for
deterministic "frame-*" IDs not yet in _frameIdentifierMap. This
handles the case where Network events arrive before Page.getResourceTree
completes for dynamically-added cross-origin iframes.`

Will this be addressed by eventually implementing `Page.getResourceTree` for site isolation (by moving the command to the web-page target too) [(bug 308896)](https://bugs.webkit.org/show_bug.cgi?id=308896)? If so, would that be worth noting as a `FIXME`?

Source/WebInspectorUI/UserInterface/Controllers/NetworkManager.js
```diff
@@ -213,7 +213,6 @@ WI.NetworkManager = class NetworkManager extends WI.Object
 
         if (target.hasDomain("Network")) {
             target.NetworkAgent.enable();
-            target.NetworkAgent.setResourceCachingDisabled(WI.settings.resourceCachingDisabled.value);
```
the-chenergy, line 216: Did you mean to move this somewhere else; why is this dropped? (I'm not familiar with why it's originally here)

Source/WebCore/inspector/InspectorIdentifierRegistry.h
burg: Good point, I had used getpid earlier and cleaned it up to use WebKit types. will fix comment.

burg: Just to clarify the reasoning its fallible, in the provisional case, FrameIdentifier is kind of dicey because the frame could get relocated and therefore need a new concrete value?

LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-request-id-uniqueness.html
burg: Good point, if the machinery is sufficient then I'll switch over. But we may need to wait. If so, I'll insert a FIXME followup.

Source/WebCore/inspector/InspectorIdentifierRegistry.h
the-chenergy: `FrameIdentifier` is dicey because that itself does not uniquely identify a `WebCore::Frame` in the web process, when provisional loading is involved

There are now three cases a provisional frame target can surface to the frontend, as of https://github.com/WebKit/WebKit/pull/61055:
1. During a frame navigation, when a `ProvisionalFrameProxy` is created
2. During a page navigation, when `ProvisionalPageProxy` reuses the main frame's `WebFrameProxy` (but still creates a new `WebCore::Frame` in a new web process, of course)
3. During a page navigation, when `ProvisionalPageProxy` creates a new `WebFrameProxy`

In the first two cases, `FrameIdentifier` gets preserved. `FrameInspectorTarget` therefore had to also include the process id to appear as unique to the frontend

Case 3 is where the new frame does get a newly generated `FrameIdentifier`, and that's basically what this current code is assuming always happens

Did this help clear it up a bit?

Source/WebCore/inspector/InspectorIdentifierRegistry.h
```diff
@@ -63,6 +68,25 @@ class IdentifierRegistry : public RefCountedAndCanMakeWeakPtr<IdentifierRegistry
...
     // Called when a document loader is detached; returns the protocol ID that was assigned.
     virtual Protocol::Network::LoaderId takeLoader(WebCore::DocumentLoader&) = 0;
+
+    // Protocol ID helpers for Site Isolation. These produce type-prefixed strings that
+    // encode the ProcessIdentifier, so both UIProcess and WebContent process can
+    // independently compute the same ID from the same typed identifier.
+    // Format: "type-processID.objectID" (e.g., "frame-12345.3", "request-12345.17")
+    static inline String protocolFrameId(WebCore::FrameIdentifier frameID)
+    {
+        return makeString("frame-"_s, frameID.toRawValue() >> 32, '.', static_cast<uint32_t>(frameID.toRawValue()));
```
the-chenergy, line 92: This is still not unique enough, according to our discussion on [why `FrameIdentifier` is dicey](https://github.com/WebKit/WebKit/pull/60842#discussion_r3030878750)

the-chenergy (changes requested):

Source/WebCore/inspector/InspectorIdentifierRegistry.h
the-chenergy: I should clarify a bit: I acknowledge that the upper bits of `FrameIdentifier` does encode the process id, but they reflect the original creating process, not always the current hosting process. During provisional frame loading, the id may be preserved across process swaps, so the process id extracted from the destination frame would contain incorrect process id info, breaking the promise in the comment above

Besides, if we really want this function to be the general protocol id generation solution, the expectation is it should handle the provisional frame case, so any protocol domain referencing frames (not just Network) never encounter the collision issue

Is it possible and still reasonable to make this work by also requiring a true `ProcessIdentifier` parameter? If not, perhaps this helper needs to be scoped for a slightly narrower use case, sadly

LayoutTests/http/tests/site-isolation/inspector/network/cross-origin-iframe-request-id-uniqueness.html
burg: This will be possible to make a lot simpler once `Page.getResourceTree` is in place.

Source/WebCore/inspector/InspectorIdentifierRegistry.h
```diff
@@ -63,6 +68,39 @@ class IdentifierRegistry : public RefCountedAndCanMakeWeakPtr<IdentifierRegistry
...
+    // Prefer the 2-arg protocolFrameId(FrameIdentifier, ProcessIdentifier) which uses
+    // an explicit hosting process. The 1-arg version extracts the process from the
+    // FrameIdentifier's upper bits, which is incorrect for provisional frames where
+    // the identifier is preserved across process swaps. See webkit.org/b/310164.
+    static inline String protocolFrameId(WebCore::FrameIdentifier frameID, WebCore::ProcessIdentifier processID)
+    {
+        return makeString("frame-"_s, processID.toUInt64(), '.', static_cast<uint32_t>(frameID.toRawValue()));
+    }
+
+    // FIXME: <https://webkit.org/b/310164> Callers that receive FrameIdentifier via IPC
```
the-chenergy, line 86: Can we create a new Bugzilla and link? This link is to the bug this PR is fixing

the-chenergy (approved):

PR #60866, opened 2026-03-18
[WebDriver BiDi] Implement browsingContext.locateNodes with CSS selector support

sergedeh:
#### bb48b7c44ce67a4f1bf5aedab6e08cef784dbcc4
<pre>
[WebDriver BiDi] Implement browsingContext.locateNodes with CSS selector support
<a href="https://bugs.webkit.org/show_bug.cgi?id=288329">https://bugs.webkit.org/show_bug.cgi?id=288329</a>

Reviewed by BJ Burg.

This patch implements the browsingContext.locateNodes command with initial
support for CSS selector-based node location. The implementation includes
protocol definitions, parameter validation, DOM query execution, and
RemoteValue serialization for found nodes.

This patch also extracts shared serializationOptions validation to a reusable
helper and fixes a critical security issue where user-provided functions could
access WebAutomationSessionProxy internals via &apos;this&apos; binding.

Key implementation details:
- Uses extractBrowsingContextHandles() to support both top-level and iframe contexts
- Inlines CSS selector logic directly in C++ string since _execute uses apply(null, ...)
- Returns actual DOM nodes from JavaScript for proper handle generation via WebKit&apos;s
  node serialization (_createNodeHandle/_identifierForNode in WebAutomationSessionProxy)
- Throws InvalidSelector error (not generic Error) for proper BiDi error mapping
- Accepts serializationOptions parameter per spec (validation only, not yet applied)
- Shared validation helper enables reuse across multiple BiDi commands

* Source/WebKit/UIProcess/Automation/protocol/BidiBrowsingContext.json:
Added Locator and LocatorType type definitions, and the locateNodes
command definition. Currently supports CSS selectors, with additional
locator types (xpath, innerText, accessibility) to be added in follow-up patches.
Added serializationOptions parameter to match BiDi spec.

* Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.h:
Added locateNodes method declaration with serializationOptions parameter.

* Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp:
(WebKit::extractNodeFromLocateResult): Helper to parse [node, metadata] tuples.
(WebKit::parseLocateNodesResult): Helper to parse JavaScript query results.
(WebKit::BidiBrowsingContextAgent::locateNodes):
Implemented locateNodes command with CSS selector support. Uses shared
validateSerializationOptions() helper and inlines JavaScript logic since
wrapper functions have this=null and cannot access proxy methods.

* Source/WebKit/UIProcess/Automation/WebDriverBidiProcessor.h:
(WebKit::WebDriverBidiProcessor::validateSerializationOptions):
Added static helper to validate BiDi serializationOptions parameter (maxDomDepth,
maxObjectDepth, includeShadowTree). Returns CommandResult&lt;void&gt; for consistency.
Shared across multiple commands: browsingContext.locateNodes, script.callFunction,
script.evaluate.

* Source/WebKit/WebProcess/Automation/WebAutomationSessionProxy.js:
(AutomationSessionProxy.prototype._execute):
Changed functionValue.apply(this, ...) to apply(null, ...) to prevent
user-provided functions from accessing proxy internals via &apos;this&apos; binding.
This secures ALL user function execution (script.evaluate, script.callFunction).

Enabled browsingContext.locateNodes tests for CSS selector locator type.

Support for startNodes parameter, serializationOptions application, and additional
locator types will be added in subsequent patches.

Canonical link: <a href="https://commits.webkit.org/316749@main">https://commits.webkit.org/316749@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/2d72730a74fbbeb3e763e3316234670f6ac63d3c

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/152383 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/25165 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/18764 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/161126 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/105840 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/25693 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/25471 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/117741 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/83470 "3 flakes 2 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/155343 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/19953 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/136805 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/98454 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/19030 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/16963 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/8960 "Built successfully") | 
| [⏳ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/JSC-Tests-x86-64-EWS "Waiting to run tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/128680 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/16273 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/163596 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/6738 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/21002 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/125777 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/24963 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/136475 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/125948 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/38729 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/24965 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/13254 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/81565 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/20943 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/88867 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/24581 "Built successfully") | [⏳ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/macOS-Sequoia-Release-Site-Isolation-Tests-EWS "Waiting to run tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/24272 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/24432 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/24333 "Built successfully") | | |


Source/WebKit/UIProcess/Automation/protocol/BidiBrowsingContext.json
```diff
@@ -65,6 +65,23 @@
...
+            "spec": "https://w3c.github.io/webdriver-bidi/#type-browsingContext-Locator",
+            "type": "object",
+            "properties": [
+                { "name": "type", "$ref": "BidiBrowsingContext.LocatorType" },
+                { "name": "value", "type": "any" }
+            ]
+        },
+        {
+            "id": "LocatorType",
+            "description": "The type of locator strategy to use for finding nodes.",
```
burg, line 81: Please add link to the bugzilla that tracks adding the rest of the values. you can add this as a "bug" property (any unrecognized properties are ignored).

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -274,6 +275,150 @@ void BidiBrowsingContextAgent::handleUserPrompt(const BrowsingContext& browsingC
...
+
+    // Parse and validate the locator type
+    String locatorType;
+    ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(!locator->getString("type"_s, locatorType), InvalidParameter);
+
+    // Get the locator value
+    RefPtr<JSON::Value> locatorValue;
+    ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(!locator->getValue("value"_s, locatorValue), InvalidParameter);
+
+    // Validate maxNodeCount if present
```
burg: Nit: i think these comments could be dropped, it's obvious from the good local variable names :)

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -274,6 +275,150 @@ void BidiBrowsingContextAgent::handleUserPrompt(const BrowsingContext& browsingC
...
+    std::optional<uint64_t> maxNodeCount;
+    if (optionalMaxNodeCount) {
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(*optionalMaxNodeCount < 0, InvalidParameter);
```
burg: Should this be <=0? What does the spec say about maxNodeCount = 0?

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -274,6 +275,150 @@ void BidiBrowsingContextAgent::handleUserPrompt(const BrowsingContext& browsingC
...
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(*optionalMaxNodeCount > JSC::maxSafeInteger(), InvalidParameter);
+        maxNodeCount = static_cast<uint64_t>(*optionalMaxNodeCount);
+    }
+
+    // Validate serializationOptions if present (shared validation used across multiple BiDi commands)
+    if (optionalSerializationOptions) {
+        if (auto error = WebDriverBidiProcessor::validateSerializationOptions(*optionalSerializationOptions))
+            ASYNC_FAIL_WITH_PREDEFINED_ERROR_AND_DETAILS(InvalidParameter, *error);
+
+        // FIXME: Implement serializationOptions support. Currently these options are accepted but not applied.
```
burg: Similarly, please ensure this bugzilla is mentioned in the protocol type definition.

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -274,6 +275,150 @@ void BidiBrowsingContextAgent::handleUserPrompt(const BrowsingContext& browsingC
...
+        // https://bugs.webkit.org/show_bug.cgi?id=288329
+    }
+
+    // Currently only CSS selector is supported
+    if (locatorType != "css"_s) {
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_AND_DETAILS(InvalidParameter,
```
burg: This should be NotImplemented or similar.

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -274,6 +275,150 @@ void BidiBrowsingContextAgent::handleUserPrompt(const BrowsingContext& browsingC
...
+        selectorString = stringValue;
+    else
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_AND_DETAILS(InvalidParameter, "CSS locator value must be a string"_s);
+
+    ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(selectorString.isEmpty(), InvalidParameter);
+
+    // FIXME: Implement startNodes support. https://bugs.webkit.org/show_bug.cgi?id=288329
+
+    // Call locateNodesByCSS method on AutomationSessionProxy
+    // Returns array of [node, metadata] tuples where:
```
burg: This comment is redundant with the same in the callee.

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -274,6 +275,150 @@ void BidiBrowsingContextAgent::handleUserPrompt(const BrowsingContext& browsingC
...
+
+    // Build arguments for locateNodesByCSS(selector, maxCount)
+    // Arguments must be JSON-encoded strings that will be parsed by _jsonParse in WebAutomationSessionProxy.js
+    Ref arguments = JSON::Array::create();
+    arguments->pushString(JSON::Value::create(selectorString)->toJSONString());
+    if (maxNodeCount)
+        arguments->pushString(JSON::Value::create(static_cast<double>(*maxNodeCount))->toJSONString());
+
+    // Call locateNodesByCSS method on AutomationSessionProxy (WebAutomationSessionProxy.js)
+    // The function is applied with `this` bound to the proxy instance, so this.locateNodesByCSS() works
```
burg: We can't do this, for both correctness and security reasons. It would expose WebAutomationSessionProxy internals as `this` for all user function calls. User functions need to be invoked with `this` bound to `null`.

Better solution: pass the proxy as an argument and do `proxy.locateNodesByCSS(...)`

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -274,6 +275,150 @@ void BidiBrowsingContextAgent::handleUserPrompt(const BrowsingContext& browsingC
...
+    session->evaluateJavaScriptFunction(topLevelContextHandle, frameHandle, "function(selector, maxCount) { return this.locateNodesByCSS(selector, maxCount); }"_s, WTF::move(arguments), false, false, std::nullopt, [callback = WTF::move(callback)](Inspector::CommandResult<String>&& result) mutable {
+        if (!result) {
+            callback(makeUnexpected(result.error()));
+            return;
+        }
+
+        // Parse the JSON result - array of [node, metadata] tuples
```
burg: I'd prefer one description of the result shape near the top, then the rest can be shortened significantly using ASYNC_FAIL_WITH_PREDEFINED_ERROR macro family.

```
auto parsedValue = ...;
ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(!parsedValue, InternalError);
...
auto resultArray = ...;
ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(!resultArray, InternalError);

File: Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp [hunk 9]
```diff
@@ -274,6 +275,150 @@ void BidiBrowsingContextAgent::handleUserPrompt(const BrowsingContext& browsingC
...
+        auto resultArray = parsedValue->asArray();
+        if (!resultArray) {
+            callback(makeUnexpected(STRING_FOR_PREDEFINED_ERROR_MESSAGE_AND_DETAILS("InternalError"_s, "Query results are not an array"_s)));
+            return;
+        }
+
+        // Build the RemoteValue array from parsed results
+        auto nodesArray = JSON::ArrayOf<Inspector::Protocol::BidiScript::RemoteValue>::create();
+
+        for (unsigned i = 0; i < resultArray->length(); i++) {
```
burg: What's the argument for/against silently ignoring problems in the resultArray, which would indicate a programming error? Should these be RELEASE_ASSERT or at least fail out the command?

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -274,6 +275,150 @@ void BidiBrowsingContextAgent::handleUserPrompt(const BrowsingContext& browsingC
...
+            for (auto& key : nodeObject->keys()) {
+                if (key.startsWith("session-node-"_s)) {
+                    if (nodeObject->getValue(key, sessionNodeValue) && sessionNodeValue) {
+                        nodeHandle = sessionNodeValue->asString();
+                        break;
+                    }
+                }
+            }
+
+            // Extract metadata (index 1)
```
burg: There's a lot of stuff going on here in the lambda passed to evaluateJavaScriptFunction. It's going to be easier to understand and maintain with some helper functions to validate and extract the key data from the possibly messed up `JSON::Value`.

Source/WebKit/UIProcess/Automation/WebDriverBidiProcessor.cpp
```diff
@@ -83,6 +84,44 @@ void WebDriverBidiProcessor::processBidiMessage(const String& message)
     m_backendDispatcher->dispatch(message);
 }
 
+std::optional<String> WebDriverBidiProcessor::validateSerializationOptions(const JSON::Object& serializationOptions)
+{
+    // Validate maxDomDepth if present
```
burg: Nit: these comments not adding anything, can be omitted

Source/WebKit/UIProcess/Automation/WebDriverBidiProcessor.cpp
```diff
@@ -83,6 +84,44 @@ void WebDriverBidiProcessor::processBidiMessage(const String& message)
...
+    if (serializationOptions.getValue("maxObjectDepth"_s, maxObjectDepthValue)) {
+        if (auto maxObjectDepth = maxObjectDepthValue->asDouble()) {
+            if (*maxObjectDepth < 0)
+                return "serializationOptions.maxObjectDepth must be non-negative"_s;
+            if (std::floor(*maxObjectDepth) != *maxObjectDepth)
+                return "serializationOptions.maxObjectDepth must be an integer"_s;
+            if (*maxObjectDepth > JSC::maxSafeInteger())
+                return "serializationOptions.maxObjectDepth exceeds maximum safe integer"_s;
+        }
+    }
```
burg, line 116: Hmmm... spec says

> If serialization options["maxDomDepth"] is equal to 0, or if value implements [ShadowRoot](https://dom.spec.whatwg.org/#shadowroot) and serialization options["includeShadowTree"] is "none", or if serialization options["includeShadowTree"] is "open" and value’s [mode](https://dom.spec.whatwg.org/#shadowroot-mode) is "closed", let children be null.

so I think these checks are OK, but at the use-site, needs to special case 0 as null rather than [].

Source/WebKit/UIProcess/Automation/WebDriverBidiProcessor.cpp
```diff
@@ -83,6 +84,44 @@ void WebDriverBidiProcessor::processBidiMessage(const String& message)
     m_backendDispatcher->dispatch(message);
 }
 
+std::optional<String> WebDriverBidiProcessor::validateSerializationOptions(const JSON::Object& serializationOptions)
```
burg: This should return a more specific type. `Inspector::CommandResult<void>`, which allows returning ErrorString when an error happens, seems like the solution you need here. If everything passes validation then the result is void.

Source/WebKit/UIProcess/Automation/WebDriverBidiProcessor.cpp
```diff
@@ -83,6 +84,44 @@ void WebDriverBidiProcessor::processBidiMessage(const String& message)
...
+                return "serializationOptions.maxObjectDepth must be an integer"_s;
+            if (*maxObjectDepth > JSC::maxSafeInteger())
+                return "serializationOptions.maxObjectDepth exceeds maximum safe integer"_s;
+        }
+    }
+
+    // Validate includesShadowTree if present
+    String includesShadowTree;
+    if (serializationOptions.getString("includesShadowTree"_s, includesShadowTree)) {
+        if (includesShadowTree != "none"_s && includesShadowTree != "open"_s && includesShadowTree != "all"_s)
```
burg: includesShadowTree values should be an enum in the protocol, so we don't have to hand-write these validation methods.

Source/WebKit/WebProcess/Automation/WebAutomationSessionProxy.js
```diff
@@ -63,6 +63,43 @@ let AutomationSessionProxy = class AutomationSessionProxy
...
+                        metadata.attributes[attr.name] = attr.value;
+                    }
+                }
+
+                results.push([elem, metadata]);
+            }
+
+            return results;
+        } catch (e) {
+            throw { name: 'InvalidSelector', message: 'Invalid CSS selector: ' + e.message };
```
burg: This will throw an object rather than throwing an Error subclass. Pls wrap in Error so that the properties you already use are well-defined.

Source/WebKit/WebProcess/Automation/WebAutomationSessionProxy.js
```diff
@@ -100,7 +137,7 @@ let AutomationSessionProxy = class AutomationSessionProxy
             let argumentValues = argumentStrings.map(this._jsonParse, this);
             if (expectsImplicitCallbackArgument)
                 argumentValues.push(resolve);
-            let resultPromise = functionValue.apply(null, argumentValues);
```
burg, line 103: r- until we figure a way around this

burg (changes requested):
This is great progress, I think the remaining issues are:

- Find a way to keep `this` bound to `null` when invoking user JS
- tighten up typing, errors, style in locateNodesByCSS() and related
- decompose the evaluateJavaScriptFunction callback to keep the overall method readable
- other minor issues as noted.

Source/WebKit/WebProcess/Automation/WebAutomationSessionProxy.js
sergedeh: I reverted the general execution path to bind user JS with this = null, and inlined the locateNodes helper at the call site so this command no longer depends on proxy-bound this. It’s not as nice as keeping the helper in the JS file, but it keeps the behavior correct and avoids exposing proxy internals. I also considered the alternative of passing the proxy explicitly, but with the current execution model this inline version was the most straightforward way to keep the behavior correct for this patch.

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
sergedeh: I changed this to fail the command with InternalError on malformed result entries rather than ignoring them.

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -57,6 +58,85 @@ using UserPromptHandlerType = Inspector::Protocol::BidiSession::UserPromptHandle
...
+    RefPtr<JSON::Value> metadata = tuple->get(1);
+    if (!metadata)
+        return std::nullopt;
+
+    // Create RemoteValue for this node
+    auto remoteValue = Inspector::Protocol::BidiScript::RemoteValue::create()
+        .setType(Inspector::Protocol::BidiScript::RemoteValueType::Node)
+        .release();
+
+    // Set the handle (sharedId in BiDi terminology)
```
burg: TIL!

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -57,6 +58,85 @@ using UserPromptHandlerType = Inspector::Protocol::BidiSession::UserPromptHandle
...
+    if (!nodeHandle.isEmpty())
+        remoteValue->setHandle(nodeHandle);
+
+    // Set the metadata as the value object
+    remoteValue->setValue(metadata.releaseNonNull());
+
+    return remoteValue;
+}
+
+static CommandResult<Ref<JSON::ArrayOf<Inspector::Protocol::BidiScript::RemoteValue>>> parseLocateNodesResult(const String& jsonResult)
```
burg: Naming nit: `resultString`

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -274,6 +354,120 @@ void BidiBrowsingContextAgent::handleUserPrompt(const BrowsingContext& browsingC
...
+    auto& [topLevelContextHandle, frameHandle] = pageAndFrameHandles.value();
+
+    Ref arguments = JSON::Array::create();
+    arguments->pushString(JSON::Value::create(selectorString)->toJSONString());
+    if (maxNodeCount)
+        arguments->pushString(JSON::Value::create(static_cast<double>(*maxNodeCount))->toJSONString());
+
+    // FIXME: Implement startNodes support. https://bugs.webkit.org/show_bug.cgi?id=288329
+    // Inline JavaScript for CSS selector logic. Since _execute uses .apply(null, ...), this wrapper has this=null,
+    // so we cannot access proxy methods. Instead, we inline the complete logic here.
```
burg, line 430: Might be possible to pass `proxy` as an argument? Anyway, this is acceptable for now.

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
```diff
@@ -274,6 +354,120 @@ void BidiBrowsingContextAgent::handleUserPrompt(const BrowsingContext& browsingC
...
+    session->evaluateJavaScriptFunction(topLevelContextHandle, frameHandle,
+        "function(selector, maxCount) { \
```
burg, line 432: The traditional way to do this (i.e., FindNodes.js) is to turn the text into binary using `xxd` and embed it into a function that returns the string. This is performed at build time by Derived Sources. We should probably continue that approach here so that the helper method remains editable.

But if the code was in the proxy js in the first place, we should just try passing it as an argument and calling to it.

WebDriverTests/TestExpectations.json
```diff
@@ -2473,7 +2473,27 @@
         "expected": { "all": { "status": ["FAIL"], "bug": "webkit.org/b/288329"}}
     },
     "imported/w3c/webdriver/tests/bidi/browsing_context/locate_nodes/invalid.py": {
-        "expected": { "all": { "status": ["FAIL"], "bug": "webkit.org/b/288329"}}
+        "expected": { "all": { "status": ["PASS"] }},
```
burg, line 3153: rubberstamp approving these progressions (I assume it's a net progression?)

Source/WebKit/UIProcess/Automation/BidiBrowsingContextAgent.cpp
burg: Let's address this in a followup.

burg (approved):
Nice work! Please update the changelog, it is slightly outdated now.

WebDriverTests/TestExpectations.json
sergedeh: yes it is, we have now 59/77 PASS for invalid.py against 0/77 before update.

burg (approved):

PR #60893, opened 2026-03-18
[LBSE] Move SVG-specific data out of RenderLayer into a lazily-allocated SVGData struct

nikolaszimmermann:
#### 967dba77d23fa5c99511879e4f6b9b520789cc05
<pre>
[LBSE] Move SVG-specific data out of RenderLayer into a lazily-allocated SVGData struct
<a href="https://bugs.webkit.org/show_bug.cgi?id=310224">https://bugs.webkit.org/show_bug.cgi?id=310224</a>

Reviewed by Simon Fraser.

Move SVG-specific state into a lazily-allocated SVGData struct, keeping RenderLayer
final with no virtual dispatch. SVG logic lives in RenderLayerSVGAdditions.cpp and
RenderLayerSVGAdditionsInlines.h, called from RenderLayer.

We chose not to introduce a RenderLayerSVG subclass because making RenderLayer
non-final would prevent the compiler from devirtualizing calls, adding vtable
pointer overhead to every layer and indirect dispatch on hot paths like
hasVisibleContentForPainting() and setupClipPath(), which is a 0.25 - 0.7%
regression on SpeedoMeter 3 -- avoid that.

Covered by existing tests.

* Source/WebCore/Sources.txt:
(WebCore::RenderElement::isInsideEntirelyHiddenLayer const):
* Source/WebCore/rendering/RenderLayer.cpp:
(WebCore::RenderLayer::RenderLayer):
(WebCore::RenderLayer::recursiveUpdateLayerPositions):
(WebCore::RenderLayer::referenceBoxRectForClipPath const):
(WebCore::RenderLayer::updateAncestorDependentState):
(WebCore::RenderLayer::computeHasVisibleContent const):
(WebCore::RenderLayer::shouldRepaintAfterLayout const):
(WebCore::RenderLayer::convertToLayerCoords const):
(WebCore::RenderLayer::setupClipPath):
(WebCore::RenderLayer::paintLayerContents):
(WebCore::RenderLayer::paintForegroundForFragments):
(WebCore::RenderLayer::hitTestLayer):
(WebCore::RenderLayer::calculateClipRects const):
(WebCore::RenderLayer::paintSVGResourceLayer): Deleted.
* Source/WebCore/rendering/RenderLayer.h:
* Source/WebCore/rendering/RenderLayerInlines.h:
(WebCore::RenderLayer::enclosingSVGHiddenOrResourceContainer const): Deleted.
* Source/WebCore/rendering/RenderLayerSVGAdditions.cpp: Added.
(WebCore::RenderLayer::hasVisibleContentForPaintingForSVG const):
(WebCore::RenderLayer::paintResourceLayerForSVG):
(WebCore::RenderLayer::setupClipPathIfNeededForSVG):
(WebCore::RenderLayer::paintForegroundForFragmentsForSVG):
(WebCore::RenderLayer::shouldSkipRepaintAfterLayoutForSVG const):
(WebCore::RenderLayer::shouldSkipHitTestForSVG const):
(WebCore::RenderLayer::hasFailedFilterForSVG const):
(WebCore::RenderLayer::updateAncestorDependentStateForSVG):
* Source/WebCore/rendering/RenderLayerSVGAdditionsInlines.h: Added.
(WebCore::RenderLayer::isPaintingResourceLayerForSVG const):
(WebCore::RenderLayer::enclosingHiddenOrResourceContainerForSVG const):
* Source/WebCore/rendering/RenderSearchField.h:
* Source/WebCore/rendering/RenderTheme.cpp:
* Source/WebCore/rendering/shapes/LayoutShape.cpp:
* Source/WebCore/rendering/svg/RenderSVGBlock.cpp:
* Source/WebCore/rendering/svg/RenderSVGPath.cpp:
(WebCore::RenderSVGPath::drawMarkers):
* Source/WebCore/rendering/svg/RenderSVGResourceClipper.cpp:
(WebCore::RenderSVGResourceClipper::applyMaskClipping):
* Source/WebCore/rendering/svg/RenderSVGResourceMasker.cpp:
(WebCore::RenderSVGResourceMasker::drawContentIntoContext):
* Source/WebCore/rendering/svg/RenderSVGResourcePattern.cpp:
(WebCore::RenderSVGResourcePattern::createTileImage const):
* Source/WebCore/rendering/svg/SVGTextQuery.cpp:
* Source/WebCore/svg/SVGGraphicsElement.cpp:
(WebCore::SVGGraphicsElement::invalidateResourceImageBuffersIfNeeded):

Canonical link: <a href="https://commits.webkit.org/311493@main">https://commits.webkit.org/311493@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/a82c710cea5efd758b9d97631186abf79879c9be

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/157139 "Passed style check") | [loading 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/30475 "Build is in progress. Recent messages:OS: Tahoe (26.2), Xcode: 26.2; Skipping applying patch since patch_id isn't provided; Checked out pull request; Compiled WebKit (warnings)") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/23666 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/165962 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/111221 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/159010 "Passed tests") | [  ~~🛠 ios-sim~~](https://ews-build.webkit.org/#/builders/155/builds/30611 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/30478 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/121691 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/111221 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/160097 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/155/builds/30611 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/141099 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/102359 "Passed tests") | 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/155/builds/30611 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/21228 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/13734 "Built successfully") | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/155/builds/30611 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/18927 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/168447 "Built successfully") | 
| | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/12606 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/20547 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/129822 "Passed tests") | 
| | [  ~~🛠 vision~~](https://ews-build.webkit.org/#/builders/153/builds/30077 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/25306 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/129930 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35196 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/30000 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/140721 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/87821 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | 
| | [  ~~🧪 vision-wk2~~](https://ews-build.webkit.org/#/builders/164/builds/24754 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/17525 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/29711 "Built successfully") | [  ~~🛠 mac-safer-cpp~~](https://ews-build.webkit.org/#/builders/120/builds/93725 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [  ~~🛠 tv-sim~~](https://ews-build.webkit.org/#/builders/157/builds/29233 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [  ~~🛠 watch~~](https://ews-build.webkit.org/#/builders/163/builds/29463 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [  ~~🛠 watch-sim~~](https://ews-build.webkit.org/#/builders/152/builds/29360 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | |

### DIFF
--- a/Source/WebCore/rendering/RenderLayer.cpp
+++ b/Source/WebCore/rendering/RenderLayer.cpp
@@ -2,7 +2,7 @@
  * Copyright (C) 2006-2024 Apple Inc. All rights reserved.
  * Copyright (C) 2013-2014 Google Inc. All rights reserved.
  * Copyright (C) 2019 Adobe. All rights reserved.
- * Copyright (c) 2020, 2021, 2022 Igalia S.L.
+ * Copyright (c) 2020, 2021, 2022, 2026 Igalia S.L.
  *
  * Portions are Copyright (C) 1998 Netscape Communications Corporation.
  *
@@ -114,6 +114,7 @@
 #include "RenderLayerFilters.h"
 #include "RenderLayerInlines.h"
 #include "RenderLayerModelObjectInlines.h"
+#include "RenderLayerSVGAdditionsInlines.h"
 #include "RenderLayerScrollableArea.h"
 #include "RenderMarquee.h"
 #include "RenderMultiColumnFlow.h"
@@ -124,6 +125,7 @@
 #include "RenderSVGInline.h"
 #include "RenderSVGModelObject.h"
 #include "RenderSVGResourceClipper.h"
+#include "RenderSVGResourceContainer.h"
 #include "RenderSVGRoot.h"
 #include "RenderSVGText.h"
 #include "RenderSVGViewportContainer.h"
@@ -363,6 +365,9 @@ RenderLayer::RenderLayer(RenderLayerModelObject& renderer)
     , m_hasNotIsolatedBlendingDescendantsStatusDirty(false)
     , m_renderer(renderer)
 {
+    if (renderer.isSVGLayerAwareRenderer() && renderer.document().settings().layerBasedSVGEngineEnabled())
+        m_svgData = makeUnique<SVGData>();
+
     setIsNormalFlowOnly(shouldBeNormalFlowOnly());
     setIsCSSStackingContext(shouldBeCSSStackingContext());
     setCanBeBackdropRoot(computeCanBeBackdropRoot());
@@ -380,7 +385,7 @@ RenderLayer::RenderLayer(RenderLayerModelObject& renderer)
         // Leave m_visibleContentStatusDirty = true in any case. The associated renderer needs to be inserted into the
         // render tree, before we can determine the visible content status. The visible content status of a SVG renderer
         // depends on its ancestors (all children of RenderSVGHiddenContainer are recursively invisible, no matter what).
-        if (renderer.isSVGLayerAwareRenderer() && renderer.document().settings().layerBasedSVGEngineEnabled())
+        if (m_svgData)
             return false;
 
         //  We need the parent to know if we have skipped content or content-visibility root.
@@ -1291,7 +1296,7 @@ void RenderLayer::recursiveUpdateLayerPositions(OptionSet<UpdateLayerPositionsFl
         m_enclosingPaginationLayer = nullptr;
     }
 
-    if (renderer().isSVGLayerAwareRenderer() && renderer().document().settings().layerBasedSVGEngineEnabled()) {
+    if (m_svgData) {
         if (!is<RenderSVGRoot>(renderer())) {
             ASSERT(!renderer().isFixedPositioned());
             if (mode == Write)
@@ -1681,12 +1686,7 @@ void RenderLayer::dirtyAncestorChainHasAlwaysIncludedInZOrderListsDescendants()
 
 FloatRect RenderLayer::referenceBoxRectForClipPath(CSSBoxType boxType, const LayoutSize& offsetFromRoot, const LayoutRect& rootRelativeBounds) const
 {
-    bool isReferenceBox = false;
-
-    if (renderer().document().settings().layerBasedSVGEngineEnabled() && renderer().isSVGLayerAwareRenderer())
-        isReferenceBox = true;
-    else
-        isReferenceBox = renderer().isRenderBox();
+    bool isReferenceBox = m_svgData ? true : renderer().isRenderBox();
 
     // FIXME: Support different reference boxes for inline content.
     // https://bugs.webkit.org/show_bug.cgi?id=129047
@@ -1946,23 +1946,14 @@ void RenderLayer::dirtyAncestorChainVisibleDescendantStatus()
 
 void RenderLayer::updateAncestorDependentState()
 {
-    m_enclosingSVGHiddenOrResourceContainer = nullptr;
-    auto determineSVGAncestors = [&] (const RenderElement& renderer) {
-        for (auto* ancestor = renderer.parent(); ancestor; ancestor = ancestor->parent()) {
-            if (auto* container = dynamicDowncast<RenderSVGHiddenContainer>(ancestor)) {
-                m_enclosingSVGHiddenOrResourceContainer = container;
-                return;
-            }
-        }
-    };
-    if (renderer().document().settings().layerBasedSVGEngineEnabled())
-        determineSVGAncestors(renderer());
+    if (m_svgData)
+        updateAncestorDependentStateForSVG();
 
     bool insideSVGForeignObject = false;
     if (renderer().document().mayHaveRenderedSVGForeignObjects()) {
         if (ancestorsOfType<LegacyRenderSVGForeignObject>(renderer()).first())
             insideSVGForeignObject = true;
-        else if (renderer().document().settings().layerBasedSVGEngineEnabled() && ancestorsOfType<RenderSVGForeignObject>(renderer()).first())
+        else if (m_svgData && ancestorsOfType<RenderSVGForeignObject>(renderer()).first())
             insideSVGForeignObject = true;
     }
 
@@ -2047,7 +2038,7 @@ bool RenderLayer::computeHasVisibleContent() const
     if (renderer().style().usedVisibility() == Visibility::Visible)
         return true;
 
-    if (!renderer().style().filter().isNone() && renderer().isSVGLayerAwareRenderer())
+    if (m_svgData && !renderer().style().filter().isNone())
         return true;
 
     // Layer's renderer has visibility:hidden, but some non-layer child may have visibility:visible.
@@ -2407,11 +2398,7 @@ RenderLayer* RenderLayer::enclosingTransformedAncestor() const
 
 bool RenderLayer::shouldRepaintAfterLayout() const
 {
-    // The SVG containers themselves never trigger repaints, only their contents are allowed to.
-    // SVG container sizes/positions are only ever determined by their children, so they will
-    // change as a reaction on a re-position/re-sizing of the children - which already properly
-    // trigger repaints.
-    if (is<RenderSVGContainer>(renderer()) && !shouldPaintWithFilters())
+    if (m_svgData && shouldSkipRepaintAfterLayoutForSVG())
         return false;
 
     if (m_repaintStatus == RepaintStatus::NeedsNormalRepaint || m_repaintStatus == RepaintStatus::NeedsFullRepaint)
@@ -2926,7 +2913,7 @@ LayoutPoint RenderLayer::convertToLayerCoords(const RenderLayer* ancestorLayer, 
         currLayer = accumulateOffsetTowardsAncestor(currLayer, ancestorLayer, locationInLayerCoords, adjustForColumns);
 
     // Pixel snap the whole SVG subtree as one "block" -- not individual layers down the SVG render tree.
-    if (renderer().isRenderSVGRoot())
+    if (m_svgData && renderer().isRenderSVGRoot())
         return LayoutPoint(roundPointToDevicePixels(locationInLayerCoords, renderer().document().deviceScaleFactor()));
 
     return locationInLayerCoords;
@@ -3302,33 +3289,6 @@ static inline bool NODELETE shouldSuppressPaintingLayer(RenderLayer* layer)
     return false;
 }
 
-void RenderLayer::paintSVGResourceLayer(GraphicsContext& context, const AffineTransform& layerContentTransform)
-{
-    bool wasPaintingSVGResourceLayer = m_isPaintingSVGResourceLayer;
-    m_isPaintingSVGResourceLayer = true;
-    context.concatCTM(layerContentTransform);
-
-    auto localPaintDirtyRect = LayoutRect::infiniteRect();
-
-    auto* rootPaintingLayer = [&] () {
-        auto* curr = parent();
-        while (curr && !(curr->renderer().isAnonymous() && is<RenderSVGViewportContainer>(curr->renderer())))
-            curr = curr->parent();
-        return curr;
-    }();
-    ASSERT(rootPaintingLayer);
-
-    LayerPaintingInfo paintingInfo(rootPaintingLayer, localPaintDirtyRect, PaintBehavior::Normal, LayoutSize());
-
-    OptionSet<PaintLayerFlag> flags { PaintLayerFlag::TemporaryClipRects };
-    if (!renderer().hasNonVisibleOverflow())
-        flags.add({ PaintLayerFlag::PaintingOverflowContents, PaintLayerFlag::PaintingOverflowContentsRoot });
-
-    paintLayer(context, paintingInfo, flags);
-
-    m_isPaintingSVGResourceLayer = wasPaintingSVGResourceLayer;
-}
-
 static inline bool NODELETE paintForFixedRootBackground(const RenderLayer* layer, OptionSet<RenderLayer::PaintLayerFlag> paintFlags)
 {
     return layer->renderer().isDocumentElementRenderer() && (paintFlags & RenderLayer::PaintLayerFlag::PaintingRootBackgroundOnly);
@@ -3539,14 +3499,8 @@ void RenderLayer::setupClipPath(GraphicsContext& context, GraphicsContextStateSa
     if (!renderer().hasClipPath() || (context.paintingDisabled() && !isCollectingEventRegion) || paintingInfo.paintDirtyRect.isEmpty())
         return;
 
-    // Applying clip-path on <clipPath> enforces us to use mask based clipping, so return false here to disable path based clipping.
-    // Furthermore if we're the child of a resource container (<clipPath> / <mask> / ...) disabled path based clipping.
-    if (is<RenderSVGResourceClipper>(m_enclosingSVGHiddenOrResourceContainer)) {
-        // If m_isPaintingSVGResourceLayer is true, this function was invoked via paintSVGResourceLayer() -- clipping on <clipPath> is already
-        // handled in RenderSVGResourceClipper::applyMaskClipping(), so do not set paintSVGClippingMask to true here.
-        paintFlags.set(PaintLayerFlag::PaintingSVGClippingMask, !m_isPaintingSVGResourceLayer);
+    if (m_svgData && setupClipPathIfNeededForSVG(paintFlags))
         return;
-    }
 
     auto clippedContentBounds = calculateLayerBounds(paintingInfo.rootLayer, offsetFromRoot, { UseLocalClipRectIfPossible });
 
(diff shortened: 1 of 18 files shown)


Source/WebCore/rendering/RenderLayerHTML.h
```diff
@@ -0,0 +1,39 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+ */
+
+#pragma once
+
+#include "RenderLayer.h"
+
+namespace WebCore {
+
+class RenderLayerHTML final : public RenderLayer {
```
shallawa: Why do we need this empty class? I think we can have only `RenderLayer` and `RenderLayerSVG`. We can look at `RenderLayerSVG` as a special case for the general case `RenderLayer`.

Source/WebCore/rendering/RenderLayerModelObject.cpp
```diff
@@ -110,7 +112,10 @@ void RenderLayerModelObject::destroyLayer()
 void RenderLayerModelObject::createLayer()
 {
     ASSERT(!m_layer);
-    m_layer = RenderLayer::create(*this);
+    if (isSVGLayerAwareRenderer() && document().settings().layerBasedSVGEngineEnabled())
+        m_layer = RenderLayerSVG::create(*this);
+    else
+        m_layer = RenderLayerHTML::create(*this);
```
shallawa: Maybe we can move this logic to `RenderLayer::create()` and make it decide which `RenderLayer` type to create.

Source/WebCore/rendering/RenderLayerSVG.h
```diff
@@ -0,0 +1,53 @@
...
+
+    WEBCORE_EXPORT ~RenderLayerSVG() final;
+
+    bool isPaintingSVGResourceLayer() const final { return m_isPaintingSVGResourceLayer; }
+    RenderSVGHiddenContainer* enclosingSVGHiddenOrResourceContainer() const final { return m_enclosingSVGHiddenOrResourceContainer.get(); }
+    void paintSVGResourceLayer(GraphicsContext&, const AffineTransform&) final;
+
+protected:
+    bool hasVisibleContentForPainting() const final;
+    void updateSVGSpecificAncestorState() final;
```
shallawa: This class is marked final. So these methods should be private.

Source/WebCore/rendering/RenderLayer.h
```diff
@@ -164,22 +166,23 @@ enum class UpdateBackingSharingFlags {
...
+class RenderLayer : public UniquelyOwned<RenderLayer> {
     WTF_MAKE_PREFERABLY_COMPACT_TZONE_ALLOCATED_EXPORT(RenderLayer, WEBCORE_EXPORT);
 public:
     friend class RenderReplica;
     friend class RenderLayerFilters;
     friend class RenderLayerBacking;
     friend class RenderLayerCompositor;
     friend class RenderLayerScrollableArea;
+    friend class RenderLayerHTML;
+    friend class RenderLayerSVG;
```
shallawa: Why does the base class need to be a friend to its sub-classes? I think any `private` method or member in `RenderLayer` should be `protected` if it is needed by its sub-classes.

Source/WebCore/rendering/RenderLayer.h
```diff
@@ -164,22 +166,23 @@ enum class UpdateBackingSharingFlags {
...
     friend void ::outputLayerPositionTreeRecursive(TextStream&, const WebCore::RenderLayer&, unsigned, const WebCore::RenderLayer*);
 
-    static UniquelyOwnedPtr<RenderLayer> create(RenderLayerModelObject& modelObject)
-    {
-        return adoptUniquelyOwned(new RenderLayer(modelObject));
-    }
+    WEBCORE_EXPORT virtual ~RenderLayer();
 
-    WEBCORE_EXPORT ~RenderLayer();
+    RenderLayerType layerType() const { return m_layerType; }
```
shallawa: Why do you need a layerType? I do not see any dynamic casting to `RenderLayer`.

Source/WebCore/rendering/RenderLayer.cpp
```diff
@@ -3529,10 +3494,10 @@ void RenderLayer::setupClipPath(GraphicsContext& context, GraphicsContextStateSa
...
     // Furthermore if we're the child of a resource container (<clipPath> / <mask> / ...) disabled path based clipping.
-    if (is<RenderSVGResourceClipper>(m_enclosingSVGHiddenOrResourceContainer)) {
-        // If m_isPaintingSVGResourceLayer is true, this function was invoked via paintSVGResourceLayer() -- clipping on <clipPath> is already
+    if (is<RenderSVGResourceClipper>(enclosingSVGHiddenOrResourceContainer())) {
+        // If isPaintingSVGResourceLayer() is true, this function was invoked via paintSVGResourceLayer() -- clipping on <clipPath> is already
         // handled in RenderSVGResourceClipper::applyMaskClipping(), so do not set paintSVGClippingMask to true here.
-        paintFlags.set(PaintLayerFlag::PaintingSVGClippingMask, !m_isPaintingSVGResourceLayer);
+        paintFlags.set(PaintLayerFlag::PaintingSVGClippingMask, !isPaintingSVGResourceLayer());
         return;
     }
```
shallawa: This logic is all about SVG. It should all be moved to `RenderLayerSVG`:

Source/WebCore/rendering/RenderLayerSVG.cpp
```diff
@@ -0,0 +1,101 @@
...
+            return;
+        }
+    }
+
+    m_enclosingSVGHiddenOrResourceContainer = nullptr;
+}
+
+bool RenderLayerSVG::hasVisibleContentForPainting() const
+{
+    if (!hasVisibleContent())
```
shallawa: Maybe `if (!RenderLayer::hasVisibleContentForPainting())`.

Source/WebCore/rendering/RenderLayerSVG.cpp
```diff
@@ -0,0 +1,101 @@
...
+    // SVG resource layers and their children are only painted indirectly, via paintSVGResourceLayer().
+    ASSERT(container->hasLayer());
+    CheckedPtr containerLayer = container->layer();
+    return containerLayer->isPaintingSVGResourceLayer();
+}
+
+void RenderLayerSVG::paintSVGResourceLayer(GraphicsContext& context, const AffineTransform& layerContentTransform)
+{
+    bool wasPaintingSVGResourceLayer = m_isPaintingSVGResourceLayer;
+    m_isPaintingSVGResourceLayer = true;
```
shallawa: Not new but this logic is weird. What does it actually mean to have `wasPaintingSVGResourceLayer` equal to true? Are we painting the same `RenderLayer` recursively? What prevents infinite loop from happening?

shallawa (commented):
This PR moves some members and methods from `RenderLayer` to `RenderLayerSVG`. Instead of using them directly, `RenderLayer` are using them through virtual methods.

I think the problem is `RenderLayer` is not using them in abstract way. But rather `RenderLayer`  still refers to them as SVG related stuff. I am afraid that `RenderLayerSVG` will be just a way to split the code rather than refactoring it or rearchitecting to work seamlessly for SVG and not SVG elements.

nikolaszimmermann:
> This PR moves some members and methods from `RenderLayer` to `RenderLayerSVG`. Instead of using them directly, `RenderLayer` are using them through virtual methods.

Thanks @shallawa for checking the PR. `RenderLayerSVG` is intentionally trivial at the moment, to keep this first patch small and evaluate whether introducing virtual dispatch causes any non-acceptable overhead. PR-59346 contains many more overriddden methods in RenderLayerSVG - the whole painting / hit-testing logic needs to be adapted to support painting a potentially transformed SVG subtree, without requiring the SVG renderers to create layers. The legacy SVG engine initiated painting from `RenderSVGRoot`, transforming the GraphicsContext along the way to the leaf renderers. The "conditional-layer LBSE" variant still paints from the outermost, enclosing RenderLayer(SVG), taking care of handling transformations, etc. without shifting the responsibility into the renderers. This PR is only a preparation for that.
 
> I think the problem is `RenderLayer` is not using them in abstract way. But rather `RenderLayer` still refers to them as SVG related stuff. I am afraid that `RenderLayerSVG` will be just a way to split the code rather than refactoring it or rearchitecting to work seamlessly for SVG and not SVG elements.
I fully agree - the change of the architecture is deliberately left for follow-up patches. I will have to introduce more SVG specific members that would otherwise bloat the RenderLayer size for non-SVG content, which is something I want to avoid, as well as adding a few hundred lines of code to RenderLayer that is purely SVG specific.

Source/WebCore/rendering/RenderLayerHTML.h
nikolaszimmermann: Fair enough we can do that -- however I planned to move more stuff out of RenderLayer that is not needed for SVG. But if you prefer, we can introduce RenderLayerHTML at a later point in time, when that move actually happens.

Source/WebCore/rendering/RenderLayerModelObject.cpp
nikolaszimmermann: Sounds good.

Source/WebCore/rendering/RenderLayerSVG.h
nikolaszimmermann: Correct. Will adapt.

Source/WebCore/rendering/RenderLayer.h
nikolaszimmermann: Likely a left-over from earlier experimentation, will get rid of it.

Source/WebCore/rendering/RenderLayer.h
nikolaszimmermann: Correct, will get rid of that.

Source/WebCore/rendering/RenderLayer.cpp
nikolaszimmermann: Ok, I can do that as well in this PR.

Source/WebCore/rendering/RenderLayerSVG.cpp
nikolaszimmermann: Good point.

Source/WebCore/rendering/RenderLayerSVG.cpp
nikolaszimmermann: Will check that again, it's old LBSE code, that will vanish as we stop to create layers for resources.

rwlbuis: Does that mean paintSVGResourceLayer will be removed eventually? If so we could add a comment for that.

Source/WebCore/rendering/RenderLayerHTML.h
rwlbuis: I think not adding RenderLayerHTML right now or keeping it in RenderLayer.cpp could improve the ugly build system fixes, but I guess once we introduce RenderLayerHTML properly we'd run into the build fixes anyway... It's mostly a review only problem though.

nikolaszimmermann: Done.

Source/WebCore/rendering/RenderLayerModelObject.cpp
nikolaszimmermann: Done.

Source/WebCore/rendering/RenderLayer.h
nikolaszimmermann: Done.

Source/WebCore/rendering/RenderLayerSVG.cpp
nikolaszimmermann: It's about nesting resources in LBSE - e.g. a mask that has a child which is clipped -- in that case wasPaintingSVGResourceLayer is true when you are painting the mask elements, and then reach a child that needs to apply a clip-path -- for now I just want to move this code around, we'll revisit resources soon, to get rid of the confusing resource layer logic altogether.

nikolaszimmermann:
> Hmm that’s unfortunate - since this blocks all further LBSE upstreaming. Can we measure this with sub percent accuracy? What I mean is the deviation 0.25 +- 0.50% or is it 0.25 +- 0.05% - the latter would be a statistical significant regression, the former not.
I missed your previous comment before posting this. It's clear it's significant.

I came up with a new approach, @smfr / @shallawa - could you give it a review and restart the AB testing please? Need to get the base perf-neutral so that we can continue evolve LBSE upstream based on a non-impactful design.

rwlbuis:
This looks good to me and I somehow prefer this design.

Source/WebCore/rendering/RenderLayer.h
```diff
@@ -455,9 +464,14 @@ class RenderLayer final : public UniquelyOwned<RenderLayer> {
             || m_hasAlwaysIncludedInZOrderListsDescendantsStatusDirty;
     }
 
-    bool isPaintingSVGResourceLayer() const { return m_isPaintingSVGResourceLayer; }
-
+    // SVG-specific methods -- defined in RenderLayerSVGInlines.h / RenderLayerSVG.cpp.
+    inline bool isPaintingSVGResourceLayer() const;
     inline RenderSVGHiddenContainer* enclosingSVGHiddenOrResourceContainer() const;
+    void paintSVGResourceLayer(GraphicsContext&, const AffineTransform&);
```
nikolaszimmermann: I will propose a unified naming scheme for those methods e.g. all having a ForSVG suffix in another PR. I did not want to enlarge this one further.

Source/WebCore/rendering/RenderLayerSVG.cpp
```diff
@@ -0,0 +1,175 @@
+/*
```
smfr, line 1: The filename makes it sound like RenderLayerSVG is a class. A good alternative name might be `RenderLayer+SVG.*` or `RenderLayerSVGAdditions.*`.

nikolaszimmermann: Done, went for RenderLayerSVGAdditions

Source/WebCore/rendering/RenderLayer.h
nikolaszimmermann: Ok, it makes sense to to it now - it's only a few additional places and leads to a unified naming scheme.

Source/WebCore/rendering/RenderLayer.h
nikolaszimmermann: Forgot to rename those to have the 'ForSVG' suffix - fixing.

Source/WebCore/rendering/RenderLayer.cpp
smfr: I don't think the impact of these `[[unlikely]]` will be significant, so let's not use them unless perf testing shows they make a difference.

smfr (approved):

Source/WebCore/rendering/RenderLayer.cpp
nikolaszimmermann: Ok, let's remove all of them and see the impact, if any. Will land ASAP. Thanks again for the review!

PR #60910, opened 2026-03-19
Back button unexpectedly enabled in newly opened tab

beidson:
#### 16dc6bfd11c1dc99d785ada5b4b5f0af08dd63fb
<pre>
Back button unexpectedly enabled in newly opened tab
<a href="https://rdar.apple.com/174798415">rdar://174798415</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=310243">https://bugs.webkit.org/show_bug.cgi?id=310243</a>

Reviewed by Chris Dumez.

Normally, navigating a page to a new URL will push a new back/forward history entry.

Browsers detect client redirects that occur during the loading process (i.e. before the load event)
and update the existing back/forward item instead of creating new ones for the location change.

Some sites do something like this &quot;quick client redirect&quot; but subtly asynchronous. For example, via
a promise chain that sometimes can resolve synchronously, but often does not, even if it resolves
very shortly after the load event. This causes the location change to actually happen after the load
event, and therefore its treated as a new navigation that warrants creating a new back/forward item.

This leads to an annoying user experience for users when, for example, a new window or tab ends up
with multiple items in the back/forward list, leading to an enabled back button, confusing the user.

In <a href="https://commits.webkit.org/259437@main">https://commits.webkit.org/259437@main</a> we added logic that skips some items created by JS when
going back or forward. This was a fix to guard against malicious JS trying to hijack the list.
But the same logic can be applied here - Tracking items that are created without user interaction
and treating them differently.

The fix in 259437@main was incomplete as it only applied to &quot;goBack&quot; and &quot;goForward&quot; to protect
against hijacking swipe gestures. It did not apply to getting the items in the lists or to the
&quot;goToItem:&quot; API.

This change expands that fix to all API access of the back/forward list, presenting an entirely
different view of the list to the client application then JavaScript will see in the `history`
interface. Additionally, it sets the flag in one more case when one of these &quot;quick redirects&quot;
occurs but after the load event.

This type of &quot;presenting a different back/forward list to the app than what JavaScript sees&quot; has
precedent in other browser engines and is probably a good mechanism to have going forward.

Finally, it also tests the new behavior both via JavaScript and the WebKit API to make sure that
the back/forward list is represented differently to each.

Test: Tools/TestWebKitAPI/Tests/WebKit/WKWebView/WKBackForwardListTests.mm

* Source/WTF/wtf/cocoa/RuntimeApplicationChecksCocoa.h:
* Source/WebCore/history/HistoryItem.cpp:
(WebCore::HistoryItem::setWasCreatedByJSWithoutUserInteraction):
* Source/WebCore/history/HistoryItem.h:
(WebCore::HistoryItem::setWasCreatedByJSWithoutUserInteraction): Deleted.
* Source/WebCore/loader/HistoryController.cpp:
(WebCore::HistoryController::updateForStandardLoad):
* Source/WebKit/UIProcess/API/C/WKBackForwardListRef.cpp:
(WKBackForwardListGetBackItem):
(WKBackForwardListGetForwardItem):
* Source/WebKit/UIProcess/API/Cocoa/WKBackForwardList.mm:
(-[WKBackForwardList backItem]):
(-[WKBackForwardList forwardItem]):
* Source/WebKit/UIProcess/WebBackForwardList.cpp:
(shouldSkipItemsForWebKitAPI):
(WebKit::WebBackForwardList::backItem const):
(WebKit::WebBackForwardList::forwardItem const):
(WebKit::WebBackForwardList::itemAtDeltaFromCurrentIndex const):
(WebKit::WebBackForwardList::backListAsAPIArrayWithLimit const):
(WebKit::WebBackForwardList::forwardListAsAPIArrayWithLimit const):
(WebKit::WebBackForwardList::itemStartingAtIndexSkippingItemsAddedByJSWithoutUserGesture const):
(WebKit::WebBackForwardList::backForwardUpdateItem):
(WebKit::WebBackForwardList::backForwardItemAtIndexForWebContent):
(WebKit::WebBackForwardListWrapper::backItem const):
(WebKit::WebBackForwardListWrapper::forwardItem const):
(WebKit::WebBackForwardListWrapper::itemAtDeltaFromCurrentIndex const):
* Source/WebKit/UIProcess/WebBackForwardList.h:
(WebKit::WebPageProxy::didChangeBackForwardList):
(WebKit::WebPageProxy::updateCanGoBackAndForward):
* Tools/TestWebKitAPI/Tests/WebKit/WKWebView/WKBackForwardListTests.mm:
(TEST(WKBackForwardList, BackForwardNavigationSkipsPromiseRedirectWithoutUserInteraction)):

Canonical link: <a href="https://commits.webkit.org/311615@main">https://commits.webkit.org/311615@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/d994c8adb07d2da5c3387a20e88186024cb0acab

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/157343 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/30680 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/23873 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/166167 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/111425 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/48d3e7ba-bc7e-4564-b21c-15fc9668436c) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/159214 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/30815 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/30682 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/121864 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85575 "Passed tests") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/c73d123e-0991-4ffc-b025-85211e194de7) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/160301 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/24108 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/141286 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/102532 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/d598e15d-a26a-412a-abee-7b99583e4ce7) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/23164 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/21412 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/13938 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/149394 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/132843 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/19114 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/168652 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/18178 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/12810 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/20734 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/129998 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/30281 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/25490 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/130105 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35271 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/30204 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/140908 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/88132 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/24928 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/17713 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/189362 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/29915 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/93929 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/48585 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/29437 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/29667 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/29564 "Built successfully") | | | |

### DIFF
--- a/Source/WebKit/UIProcess/WebBackForwardList.cpp
+++ b/Source/WebKit/UIProcess/WebBackForwardList.cpp
@@ -65,6 +65,14 @@ static inline void setBackForwardItemIdentifiers(FrameState& frameState, BackFor
 
 #if !ENABLE(BACK_FORWARD_LIST_SWIFT)
 
+static bool shouldSkipItemsWithoutUserGestureForWebKitAPI()
+{
+#if PLATFORM(COCOA)
+    return linkedOnOrAfterSDKWithBehavior(SDKAlignedBehavior::AllBackForwardItemsWithoutUserGestureInvisibleToUI);
+#endif
+    return false;
+}
+
 static const unsigned DefaultCapacity = 100;
 
 WebBackForwardList::WebBackForwardList(WebPageProxy& page)
@@ -265,26 +273,53 @@ WebBackForwardListItem* WebBackForwardList::currentItem() const
     return m_page && m_currentIndex ? m_entries[*m_currentIndex].ptr() : nullptr;
 }
 
-WebBackForwardListItem* WebBackForwardList::backItem() const
+RefPtr<WebBackForwardListItem> WebBackForwardList::backItem() const
 {
     ASSERT(!m_currentIndex || *m_currentIndex < m_entries.size());
+    if (!m_page || !m_currentIndex)
+        return nullptr;
 
-    return m_page && m_currentIndex && *m_currentIndex ? m_entries[*m_currentIndex - 1].ptr() : nullptr;
+    if (shouldSkipItemsWithoutUserGestureForWebKitAPI())
+        return itemStartingAtIndexSkippingItemsAddedByJSWithoutUserGesture(NavigationDirection::Backward, *m_currentIndex).first;
+
+    return *m_currentIndex ? m_entries[*m_currentIndex - 1].ptr() : nullptr;
 }
 
-WebBackForwardListItem* WebBackForwardList::forwardItem() const
+RefPtr<WebBackForwardListItem> WebBackForwardList::forwardItem() const
 {
     ASSERT(!m_currentIndex || *m_currentIndex < m_entries.size());
+    if (!m_page || !m_currentIndex)
+        return nullptr;
 
-    return m_page && m_currentIndex && m_entries.size() && *m_currentIndex < m_entries.size() - 1 ? m_entries[*m_currentIndex + 1].ptr() : nullptr;
+    if (shouldSkipItemsWithoutUserGestureForWebKitAPI())
+        return itemStartingAtIndexSkippingItemsAddedByJSWithoutUserGesture(NavigationDirection::Forward, *m_currentIndex).first;
+
+    return m_entries.size() && *m_currentIndex < m_entries.size() - 1 ? m_entries[*m_currentIndex + 1].ptr() : nullptr;
 }
 
-RefPtr<WebBackForwardListItem> WebBackForwardList::itemAtDeltaFromCurrentIndex(int delta) const
+RefPtr<WebBackForwardListItem> WebBackForwardList::itemAtDeltaFromCurrentIndex(int delta, AllowSkippingBackForwardItems allowSkippingBackForwardItems) const
 {
     if (!m_currentIndex || (int)*m_currentIndex + delta < 0)
         return nullptr;
 
-    return itemAtIndexWithoutSkipping(*m_currentIndex + delta).first;
+    // API requests to get the current item will always get the current item without any skipping logic.
+    if (!delta)
+        return itemAtIndexWithoutSkipping(*m_currentIndex).first;
+
+    if (allowSkippingBackForwardItems == AllowSkippingBackForwardItems::No || !shouldSkipItemsWithoutUserGestureForWebKitAPI())
+        return itemAtIndexWithoutSkipping(*m_currentIndex + delta).first;
+
+    auto direction = delta < 0 ? NavigationDirection::Backward : NavigationDirection::Forward;
+    size_t stepsLeft = abs(delta);
+    size_t nextIndex = *m_currentIndex;
+    while (stepsLeft) {
+        auto item = itemStartingAtIndexSkippingItemsAddedByJSWithoutUserGesture(direction, nextIndex);
+        if (!item.first || !--stepsLeft)
+            return item.first;
+        nextIndex = item.second;
+    }
+
+    return nullptr;
 }
 
 std::pair<RefPtr<WebBackForwardListItem>, size_t> WebBackForwardList::itemAtIndexWithoutSkipping(size_t index) const
@@ -340,6 +375,25 @@ Ref<API::Array> WebBackForwardList::backListAsAPIArrayWithLimit(unsigned limit) 
         return API::Array::create();
 
     ASSERT(backListSize >= size);
+
+    if (shouldSkipItemsWithoutUserGestureForWebKitAPI()) {
+        Vector<RefPtr<API::Object>> vector;
+
+        size_t nextStartingIndex = *m_currentIndex;
+        while (backListSize) {
+            auto item = itemStartingAtIndexSkippingItemsAddedByJSWithoutUserGesture(NavigationDirection::Backward, nextStartingIndex);
+            if (item.first)
+                vector.append(item.first);
+
+            if (!item.first || !--backListSize || !item.second)
+                break;
+            nextStartingIndex = item.second;
+        }
+        vector.reverse();
+
+        return API::Array::create(WTF::move(vector));
+    }
+
     size_t startIndex = backListSize - size;
     Vector<RefPtr<API::Object>> vector(size, [&](size_t i) -> RefPtr<API::Object> {
         // FIXME: Remove SUPPRESS_UNCOUNTED_LAMBDA_CAPTURE when the false positive
@@ -360,6 +414,23 @@ Ref<API::Array> WebBackForwardList::forwardListAsAPIArrayWithLimit(unsigned limi
     unsigned size = std::min(static_cast<unsigned>(forwardListCount()), limit);
     if (!size)
         return API::Array::create();
+
+    if (shouldSkipItemsWithoutUserGestureForWebKitAPI()) {
+        Vector<RefPtr<API::Object>> vector;
+
+        size_t nextStartingIndex = *m_currentIndex;
+        while (size) {
+            auto item = itemStartingAtIndexSkippingItemsAddedByJSWithoutUserGesture(NavigationDirection::Forward, nextStartingIndex);
+            if (item.first)
+                vector.append(item.first);
+
+            if (!item.first || !--size || !item.second)
+                break;
+            nextStartingIndex = item.second;
+        }
+
+        return API::Array::create(WTF::move(vector));
+    }
 
     size_t startIndex = *m_currentIndex + 1;
     Vector<RefPtr<API::Object>> vector(size, [&](size_t i) -> RefPtr<API::Object> {
@@ -517,19 +588,44 @@ std::pair<RefPtr<WebBackForwardListItem>, size_t> WebBackForwardList::itemStarti
         return { nullptr, 0 };
 
     auto item = itemAtIndexWithoutSkipping(itemIndex);
-    ASSERT(item.first);
 
 #if PLATFORM(COCOA)
     if (!linkedOnOrAfterSDKWithBehavior(SDKAlignedBehavior::UIBackForwardSkipsHistoryItemsWithoutUserGesture))
         return item;
 #endif
 
-    // For example:
-    // Yahoo -> Yahoo#a (no userInteraction) -> Google -> Google#a (no user interaction) -> Google#b (no user interaction)
-    // If we're on Google and navigate back, we don't want to skip anything and load Yahoo#a.
-    // However, if we're on Yahoo and navigate forward, we do want to skip items and end up on Google#b.
-    if (direction == NavigationDirection::Backward && !currentItem()->wasCreatedByJSWithoutUserInteraction())
+    if (!item.first)
         return item;
+
+    // For example:
+    // A -> A#a (no userInteraction) -> B -> B#a (no user interaction) -> B#b (no user interaction)
+    // If we're on B and navigate back, we don't want to skip anything and load A#a.
+    // However, if we're on A and navigate forward, we do want to skip items and end up on B#b.
+    if (direction == NavigationDirection::Backward && !currentItem()->wasCreatedByJSWithoutUserInteraction()) {
+        // The exception to the above is that if the entire list of back items is missing user interaction,
+        // they should all be ignored. For example:
+        // A (no userInteraction) -> B (no userInteraction) -> C*
+        // From the API perspective, C should be item 0, and going back is not an option.
+        // This happens e.g. with new windows that undergo a series of JS driven client redirects.
+        // See https://bugs.webkit.org/show_bug.cgi?id=310243
+
+        auto itemToReturn = item;
+        do {
+            if (item.first->wasCreatedByJSWithoutUserInteraction()) {
+                if (item.second)
+                    item = itemAtIndexWithoutSkipping(item.second - 1);
+                else {
+                    // We reached the beginning of the list and still didn't find an item with user interaction,
+                    // therefore we are returning nothing.
+                    itemToReturn = { };
+                    break;
+                }
+            } else
+                break;
+        } while (true);
+
+        return itemToReturn;
+    }
 
     // For example:
     // Yahoo -> Yahoo#a (no userInteraction) -> Google -> Google#a (no user interaction) -> Google#b (no user interaction)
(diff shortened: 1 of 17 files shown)


beidson:
This intel-wk2 failure with imported/w3c/web-platform-tests/webrtc/RTCPeerConnection-getStats.https.html?rest seems extraordinarily unlikely to be from this back/forward focused patch.

> FAIL RTCStats.timestamp increases with time passing assert_less_than_equal: expected a number less than or equal to 503 but got 503.000244140625

So I'm not gonna sweat that.

Source/WebKit/UIProcess/WebBackForwardList.cpp
```diff
@@ -824,19 +933,19 @@ WebBackForwardListItem* WebBackForwardListWrapper::currentItem() const
     return m_impl->currentItem();
 }
 
-WebBackForwardListItem* WebBackForwardListWrapper::backItem() const
+RefPtr<WebBackForwardListItem>  WebBackForwardListWrapper::backItem() const
```
cdumez: extra space after return type here.

Source/WebKit/UIProcess/WebBackForwardList.cpp
```diff
@@ -824,19 +933,19 @@ WebBackForwardListItem* WebBackForwardListWrapper::currentItem() const
...
 {
     return m_impl->backItem();
 }
 
-WebBackForwardListItem* WebBackForwardListWrapper::forwardItem() const
+RefPtr<WebBackForwardListItem>  WebBackForwardListWrapper::forwardItem() const
```
cdumez: extra space after return type here.

Source/WebKit/UIProcess/WebBackForwardList.cpp
```diff
@@ -52,6 +52,19 @@
 // FIXME: https://bugs.webkit.org/show_bug.cgi?id=306415
 #include "WebKit-Swift.h"
 
+static bool shouldSkipItemsForWebKitAPI()
```
cdumez: The name is too generic here. It should be something like `shouldSkipItemsWithoutUserGestureForWebKitAPI()`

Source/WebKit/UIProcess/WebBackForwardList.cpp
```diff
@@ -52,6 +52,19 @@
...
+{
+#if PLATFORM(COCOA)
+    static bool shouldSkip;
```
cdumez: There is not really much use for this static variable. `linkedOnOrAfterSDKWithBehavior(SDKAlignedBehavior::AllBackForwardItemsWithoutUserGestureInvisibleToUI)` is cheap and already gets its result from a static BitSet. So your static is only avoiding a bit operation here, the code is not hot enough to warrant the extra code.

Source/WebKit/UIProcess/WebBackForwardList.cpp
```diff
@@ -517,20 +593,45 @@ std::pair<RefPtr<WebBackForwardListItem>, size_t> WebBackForwardList::itemStarti
...
-    // However, if we're on Yahoo and navigate forward, we do want to skip items and end up on Google#b.
-    if (direction == NavigationDirection::Backward && !currentItem()->wasCreatedByJSWithoutUserInteraction())
+    if (!item.first)
         return item;
 
+    // For example:
+    // A -> A#a (no userInteraction) -> B -> B#a (no user interaction) -> B#b (no user interaction)
+    // If we're on B and navigate back, we don't want to skip anything and load A#a.
+    // However, if we're on A and navigate forward, we do want to skip items and end up on B#b.
+    if (direction == NavigationDirection::Backward && !protect(currentItem())->wasCreatedByJSWithoutUserInteraction()) {
```
cdumez: `wasCreatedByJSWithoutUserInteraction()` is marked as `NODELETE`, the `protect()` is not needed here. Note that the previous code did not have the `protect()` and WebKit/ is clean from a safer cpp standpoint.

Source/WebKit/UIProcess/WebBackForwardList.cpp
```diff
@@ -517,20 +593,45 @@ std::pair<RefPtr<WebBackForwardListItem>, size_t> WebBackForwardList::itemStarti
...
+
+        auto itemToReturn = item;
+        do {
+            if (item.first->wasCreatedByJSWithoutUserInteraction()) {
+                if (item.second)
+                    item = itemAtIndexWithoutSkipping(item.second - 1);
+                else {
+                    // We reached the beginning of the list and still didn't find an item with user interaction,
+                    // therefore we are returning nothing.
+                    itemToReturn = { nullptr, 0 };
```
cdumez: This can simplify be `itemToReturn = { };` I believe.

Source/WebKit/UIProcess/WebBackForwardList.cpp
```diff
@@ -540,8 +641,14 @@ std::pair<RefPtr<WebBackForwardListItem>, size_t> WebBackForwardList::itemStarti
...
         itemIndex += delta;
         item = itemAtIndexWithoutSkipping(itemIndex);
-        if (!item.first)
-            return originalitem;
+        if (!item.first) {
+            // If there are no more back items that ever had a user gesture, then we should not enable going back.
+            // This happens when e.g. a new window is created by JavaScript then client redirects occur that create
+            // a sequence of history items, each without user interaction.
+            RELEASE_LOG(Loading, "UI Navigation is disabling going back because no more WebBackForwardListItem items in the back list had user interaction");
+            return { nullptr, 0 };
```
cdumez: This could be `return { };`

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -2253,6 +2253,8 @@ void WebPageProxy::loadRequestWithNavigationShared(Ref<WebProcessProxy>&& proces
     loadParameters.existingNetworkResourceLoadIdentifierToResume = existingNetworkResourceLoadIdentifierToResume;
     loadParameters.advancedPrivacyProtections = navigation.originatorAdvancedPrivacyProtections();
     loadParameters.isRequestFromClientOrUserInput = navigation.isRequestFromClientOrUserInput();
+    if (auto& action = navigation.lastNavigationAction())
```
cdumez: There is already such an `if` check below for `loadParameters.requester = action->requester;`, let's move our assignment to the other `if` scope to avoid the null check twice.

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -5778,6 +5786,8 @@ void WebPageProxy::continueNavigationInNewProcess(API::Navigation& navigation, W
         loadParameters.shouldTreatAsContinuingLoad = shouldTreatAsContinuingLoad;
         loadParameters.frameIdentifier = frame.frameID();
         loadParameters.isRequestFromClientOrUserInput = navigation.isRequestFromClientOrUserInput();
+        if (auto& action = navigation.lastNavigationAction())
```
cdumez: Let's reuse this scope below:
```
if (auto& action = navigation.lastNavigationAction())
    loadParameters.requester = action->requester;
```

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/ProcessSwapOnNavigation.mm
```diff
@@ -2457,21 +2457,10 @@ static void runNavigationWithLockedHistoryTest(ShouldEnablePSON shouldEnablePSON
     EXPECT_EQ(webkitPID, [webView _webProcessIdentifier]);
     EXPECT_WK_STREQ(@"pson://www.webkit.org/main.html", [backForwardList.currentItem.URL absoluteString]);
     EXPECT_TRUE(!backForwardList.backItem);
-    EXPECT_EQ(1U, backForwardList.forwardList.count);
```
cdumez, line 2460: We're losing test coverage here. It would be better to generate navigations with a user gesture in those tests rather than making such changes.

cdumez: The same comment applies to other tests below.

cdumez (changes requested):

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/ProcessSwapOnNavigation.mm
cdumez: Not approving PR because of API test changes.

beidson: I'd looked at it as the opposite of "we're gaining test coverage for the new behavior, nice"

But we're both right.

I'll see about refactoring these tests to run both ways.

beidson: Nevermind, the PSON tests should all be covering PSON still exclusively.
I resolved all the PSON tests to have user gestures, and went over the back/forward list tests closely to make sure all the changes were expected due to the new behavior.

Source/WebKit/UIProcess/WebBackForwardList.cpp
```diff
@@ -340,6 +375,25 @@ Ref<API::Array> WebBackForwardList::backListAsAPIArrayWithLimit(unsigned limit)
         return API::Array::create();
 
     ASSERT(backListSize >= size);
+
+    if (shouldSkipItemsWithoutUserGestureForWebKitAPI()) {
+        Vector<RefPtr<API::Object>> vector;
+
+        int nextStartingIndex = *m_currentIndex;
```
jelee53: Nit: `size_t` or `unsigned`for indices maybe?

Source/WebKit/UIProcess/WebBackForwardList.cpp
```diff
@@ -265,26 +273,53 @@ WebBackForwardListItem* WebBackForwardList::currentItem() const
...
+    // API requests to get the current item will always get the current item without any skipping logic.
+    if (!delta)
+        return itemAtIndexWithoutSkipping(*m_currentIndex).first;
+
+    if (allowSkippingIfSupported == AllowSkippingIfSupported::No || !shouldSkipItemsWithoutUserGestureForWebKitAPI())
+        return itemAtIndexWithoutSkipping(*m_currentIndex + delta).first;
+
+    auto direction = delta < 0 ? NavigationDirection::Backward : NavigationDirection::Forward;
+    int stepsLeft = abs(delta);
+    int nextIndex = *m_currentIndex;
```
jelee53: Nit: `size_t` or `unsigned` for `stepsLeft` and `nextIndex` maybe?

cdumez: I agree we should be using `size_t` since `m_currentIndex` is a size_t and `itemStartingAtIndexSkippingItemsAddedByJSWithoutUserGesture()` also returns a size_t.

Source/WebKit/UIProcess/WebBackForwardList.cpp
cdumez: `size_t` indeed.

Source/WebKit/UIProcess/WebBackForwardList.cpp
```diff
@@ -361,6 +415,23 @@ Ref<API::Array> WebBackForwardList::forwardListAsAPIArrayWithLimit(unsigned limi
     if (!size)
         return API::Array::create();
 
+    if (shouldSkipItemsWithoutUserGestureForWebKitAPI()) {
+        Vector<RefPtr<API::Object>> vector;
+
+        int nextStartingIndex = *m_currentIndex;
```
cdumez: `size_t`

cdumez (approved):

PR #61055, opened 2026-03-21
[Site Isolation] Web Inspector: Consider involving a provisional frame target on page-level cross-origin navigation

the-chenergy:
#### 0b243f712b46a34f96911dd160dc0236266c3361
<pre>
[Site Isolation] Web Inspector: Consider involving a provisional frame target on page-level cross-origin navigation
<a href="https://bugs.webkit.org/show_bug.cgi?id=310143">https://bugs.webkit.org/show_bug.cgi?id=310143</a>

Reviewed by BJ Burg.

In 308662@main, we introduced the provisional frame target which
surfaces during a cross-origin navigation of an iframe. A provisional
target helps maintain inspector continuity across process swaps: it
allows the inspector to connect to the new page or frame element before
the element e.g. begins script evaluation.

However, that implementation does not address the case of a page-level
cross-origin navigation. A ProvisionalPageProxy creates a regular main
frame (WebFrameProxy, rather than ProvisionalFrameProxy), which surfaces
as a non-provisional frame target in the frontend. This means we are
still relying on the provisional page target for all the provisional
loading support, which is unreliable especially as more features migrate
to the frame target. At minimum, in the far future when we deprecate the
page target altogether, we will need a provisional frame target
representation during a page-level navigation as well.

This patch makes the regular main frame created by a provisional page
show up as a provisional frame target as well. From frontend&apos;s
perspective, a page-level cross-origin navigation is the same as a
main frame cross-origin navigation, and the same procedure of a
provisional target applies. Now:
- Normally, there&apos;s one page target and some frame targets, one of the
  frame targets being the main frame target.
- When a page navigation begins, a provisional page target and a
  provisional main frame target are created.
- When provisional loading commits, the two provisional targets commit,
  while all other targets get destroyed.
     - Any other frames that the new page ends up loading will be
       reported as new provisional (if cross-origin) or non-provisional
       (if same-origin) frame targets after this point.
- When provisional loading is cancelled, the two provisional targets
  get destroyed.

I made a mistake while composing the mentioned patch 308662@main: I
said WebFrameProxies were &quot;preserved&quot; during a page navigation, which
was false. In reality, all non-main frames belonging to the new page are
newly created. For the main frame, the ProvisionalPageProxy may either
create a new WebFrameProxy or reuse it while setting it to a new
web process. In either case, we create a provisional main frame target
proxying to the resulting WebFrame in the appropriate web process,
identified by the frame id and the process reference. In
didCommitProvisionalPage, we don&apos;t need to recreate any frame targets:
the new non-main targets will be created with their new WebFrameProxies,
after document parsing begins.

* LayoutTests/http/tests/site-isolation/inspector/debugger/resources/site-isolation-debugger-test-utilities.js:
* LayoutTests/http/tests/site-isolation/inspector/target/target-cross-origin-page-navigation-expected.txt:
* LayoutTests/http/tests/site-isolation/inspector/target/target-cross-origin-page-navigation.html:
   Updated to verify the provisional main frame target&apos;s presence.
   The unexpected console error noise no longer shows up apparently as
   a result of this change.

* LayoutTests/platform/mac-wk2/TestExpectations:
   - As a drive-by, fix the formatting of nearby related inspector test
     expectations to include the correct bug ID and the correct build
     marker ([ Debug ]).

* Source/WebKit/UIProcess/Inspector/WebPageInspectorController.cpp:
(WebKit::WebPageInspectorController::didCreateProvisionalPage):
(WebKit::WebPageInspectorController::willDestroyProvisionalPage):
(WebKit::WebPageInspectorController::didCommitProvisionalPage):
* Source/WebKit/UIProcess/Inspector/WebPageInspectorController.h:
(WebKit::ProvisionalPageProxy::~ProvisionalPageProxy):
(WebKit::ProvisionalPageProxy::initializeWebPage):
(WebKit::WebFrameProxy::WebFrameProxy):
(WebKit::WebFrameProxy::didCreateSubframe):
(WebKit::WebPageProxy::initializeWebPage):

Canonical link: <a href="https://commits.webkit.org/310448@main">https://commits.webkit.org/310448@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/482279ac3ba9347ffcc7254f673bbe8211f50e09

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/153772 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/26556 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/20173 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/162523 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/107233 "Built successfully") | [loading 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/cff3a446-1090-45f7-9f60-8d1280936aaa) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/27084 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/26878 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/118887 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/84069 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/156731 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/21153 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/138073 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/99597 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b396004b-ebfb-446a-951b-693ac83762af) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/20233 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/18190 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/10355 "Built successfully") | | 
| | [loading 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/129881 "Build is in progress. Recent messages:OS: Tahoe (26.2), Xcode: 26.2; Running apply-patch; Checked out pull request; run-api-tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/15932 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/164993 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/8127 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/17526 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/126972 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/26353 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/22221 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/127139 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/26355 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/137727 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/83033 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/23506 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/22042 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/14510 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/25972 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/90260 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/25663 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/25823 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/25723 "Built successfully") | | | |

### DIFF
--- a/LayoutTests/http/tests/site-isolation/inspector/target/target-cross-origin-page-navigation.html
+++ b/LayoutTests/http/tests/site-isolation/inspector/target/target-cross-origin-page-navigation.html
@@ -1,4 +1,5 @@
 <script src="../../../inspector/resources/inspector-test.js"></script>
+<script src="../debugger/resources/site-isolation-debugger-test-utilities.js"></script>
 
 <script>
     function test() {
@@ -7,37 +8,55 @@
         suite.addTestCase({
             name: "SiteIsolation.Target.CrossOriginPageNavigation.FrameTargetLifetime",
             description: "On cross-origin page navigation, frame targets are destroyed and created as appropriate.",
-            async test() {
-                async function logDocumentURLs(targets) {
-                    let urls = await Promise.all(targets.map(async function (target) {
-                        let { result } = await target.RuntimeAgent.evaluate.invoke({
-                            expression: "document.location.href",
-                            objectGroup: "test",
-                            returnByValue: true,
-                        });
-                        return result.value;
-                    }));
-                    InspectorTest.log(urls.sort().join("\n"));
-                }
-
+            async test(script) {
                 let oldFrameTargets = WI.targets.filter((t) => t.type === WI.TargetType.Frame);
                 InspectorTest.expectEqual(oldFrameTargets.length, 3, "Should have main, middle, and leaf frame targets before navigation.");
+                let urlTargetMap = await fetchDocumentURLsForTargets(oldFrameTargets);
                 InspectorTest.log("Frame URLs before navigation:");
-                await logDocumentURLs(oldFrameTargets);
+                InspectorTest.log(Array.from(urlTargetMap.keys()).sort().join("\n"));
+
+                let oldMainFrameTargetId = urlTargetMap.get(WI.networkManager.mainFrame.url).identifier;
+
+                // Promises that wait for the next TargetAdded and DidCommitProvisionalTarget events for a frame.
+                let provisionalFrameTargetAdded = new Promise(function (resolve) {
+                    function handle(event) {
+                        let { target } = event.data;
+                        if (target.type === WI.TargetType.Frame) {
+                            WI.targetManager.removeEventListener(WI.TargetManager.Event.TargetAdded, handle);
+                            resolve(target.identifier);
+                        }
+                    }
+                    WI.targetManager.addEventListener(WI.TargetManager.Event.TargetAdded, handle);
+                });
+                let provisionalFrameTargetCommitted = new Promise(function (resolve) {
+                    function handle(event) {
+                        let { previousTargetId, target } = event.data;
+                        if (target.type === WI.TargetType.Frame) {
+                            WI.targetManager.removeEventListener(WI.TargetManager.Event.DidCommitProvisionalTarget, handle);
+                            resolve({ previousTargetId, newTargetId: target.identifier });
+                        }
+                    }
+                    WI.targetManager.addEventListener(WI.TargetManager.Event.DidCommitProvisionalTarget, handle);
+                });
 
                 InspectorTest.log("Navigating cross-origin...");
                 InspectorTest.deferOutputUntilTestPageIsReloaded();
-                // FIXME <https://webkit.org/b/309922>: Address the unrelated console error that shows up
-                // during this navigation.
                 InspectorTest.evaluateInPage(`location.hostname = "localhost";`);
 
                 // Wait for the log from leaf-iframe to ensure all loading has completed.
                 await WI.consoleManager.awaitEvent(WI.ConsoleManager.Event.MessageAdded);
 
+                let provisionalFrameTargetId = await provisionalFrameTargetAdded;
+                InspectorTest.pass("Provisional main frame target gets created.");
+
+                let { previousTargetId, newTargetId } = await provisionalFrameTargetCommitted;
+                InspectorTest.expectEqual(newTargetId, provisionalFrameTargetId, "Main frame target is committed.");
+                InspectorTest.expectEqual(previousTargetId, oldMainFrameTargetId, "Previous id reported by the commit event should be the old main frame target's id.");
+
                 let newFrameTargets = WI.targets.filter((t) => t.type === WI.TargetType.Frame);
                 InspectorTest.expectEqual(newFrameTargets.length, 3, "Should have main, middle, and leaf frame targets after navigation.");
                 InspectorTest.log("Frame URLs after navigation:");
-                await logDocumentURLs(newFrameTargets);
+                InspectorTest.log(Array.from((await fetchDocumentURLsForTargets(newFrameTargets)).keys()).sort().join("\n"));
 
                 let oldIds = oldFrameTargets.map((t) => t.identifier);
                 let targetsHaveNewIds = newFrameTargets.every((t) => !oldIds.includes(t.identifier));
--- a/Source/WebKit/UIProcess/Inspector/WebPageInspectorController.cpp
+++ b/Source/WebKit/UIProcess/Inspector/WebPageInspectorController.cpp
@@ -208,41 +208,69 @@ void WebPageInspectorController::setContinueLoadingCallback(const ProvisionalPag
     target->setResumeCallback(WTF::move(callback));
 }
 
-void WebPageInspectorController::didCreateProvisionalPage(ProvisionalPageProxy& provisionalPage)
+void WebPageInspectorController::didCreateProvisionalPage(ProvisionalPageProxy& provisionalPage, WebCore::FrameIdentifier mainFrameID, WebProcessProxy& mainFrameProcess)
 {
     addTarget(PageInspectorTargetProxy::create(provisionalPage, getTargetID(provisionalPage), Inspector::InspectorTargetType::Page));
-}
-
-void WebPageInspectorController::willDestroyProvisionalPage(const ProvisionalPageProxy& provisionalPage)
-{
-    removeTarget(getTargetID(provisionalPage));
-}
-
-void WebPageInspectorController::didCommitProvisionalPage(WebCore::PageIdentifier oldWebPageID, WebCore::PageIdentifier newWebPageID)
-{
-    String oldID = PageInspectorTarget::toTargetID(oldWebPageID);
-    String newID = PageInspectorTarget::toTargetID(newWebPageID);
-    auto newTarget = m_targets.take(newID);
-    CheckedPtr targetAgent = m_targetAgent;
-    ASSERT(newTarget);
-    newTarget->didCommitProvisionalTarget();
-    targetAgent->didCommitProvisionalTarget(oldID, newID);
-
-    // Update target list to only include targets from the committed page.
-    for (auto& target : m_targets.values())
-        targetAgent->targetDestroyed(*target);
-    m_targets.clear();
-    m_targets.set(newTarget->identifier(), WTF::move(newTarget));
 
     if (shouldManageFrameTargets()) {
-        // WebFrameProxies are structurally preserved across page navigation rather than destroyed and recreated,
-        // so no didCreateFrame/willDestroyFrame callbacks fire for them. Recreate the frame targets here.
-        //
-        // (Frame target ids include the process id, which changes across a cross-origin navigation,
-        // so surviving frames need to surface as new targets to the frontend.)
-        for (RefPtr frame = m_inspectedPage->mainFrame(); frame; frame = frame->traverseNext().frame)
-            didCreateFrame(*frame);
+        constexpr bool isProvisional = true;
+        addTarget(makeUnique<FrameInspectorTargetProxy>(mainFrameID, mainFrameProcess, isProvisional));
     }
+}
+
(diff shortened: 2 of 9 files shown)


basuke (approved):
Nice cleanup!

One small concern: `didCreateProvisionalPage` explicitly calls `addTarget` for the provisional frame target, but `willDestroyProvisionalPage` relies on the `~WebFrameProxy -> willDestroyFrame` path to remove it. This implicit dependency feels a bit fragile — if the destruction order ever changes, the target could leak. Would it be worth adding an explicit `removeTarget` here (or at least an `ASSERT`) for symmetry?

Also the test covers the happy path nicely. Do we have (or plan to have) a test for the cancellation case — where provisional loading is cancelled and both provisional targets get destroyed? Seems like a good thing to cover to prevent regressions.

Neither of these should hinder the commit, so please keep them in mind for future reference.

basuke (changes requested):
Please fix this. This causes many crashes!

Source/WebKit/UIProcess/WebPageProxy.cpp
basuke, line 5647: `mainFrameInPreviousProcess` is already moved. We need to save `oldMainFrameID` before here.

basuke (changes requested):

LayoutTests/http/tests/site-isolation/inspector/target/target-cross-origin-page-navigation.html
```diff
@@ -7,37 +7,68 @@
         suite.addTestCase({
             name: "SiteIsolation.Target.CrossOriginPageNavigation.FrameTargetLifetime",
             description: "On cross-origin page navigation, frame targets are destroyed and created as appropriate.",
-            async test() {
-                async function logDocumentURLs(targets) {
-                    let urls = await Promise.all(targets.map(async function (target) {
+            async test(script) {
+                // Returns a Map of {documentURL => target}.
```
burg: This might be useful to re-use, i.e., in Network and Page tests with complicated frame trees.

burg: I prefer fetch over get for naming async methods, but either is acceptable.

LayoutTests/http/tests/site-isolation/inspector/target/target-cross-origin-page-navigation.html
```diff
@@ -7,37 +7,68 @@
...
+                async function getDocumentURLs(targets) {
+                    let entries = await Promise.all(targets.map(async function (target) {
```
burg: Just to clarify: what happens if the page is suspended or paused in debugger, will this still actually eval on the frame target?

What if the page has somehow overridden `document.location.href`? The snippet could be evaluated in a separate world to avoid those consequences.

I guess the reason I asked if it's worth reusing is, all these questions would be answered by utility class test cases.

Source/WebKit/UIProcess/Inspector/WebPageInspectorController.cpp
```diff
@@ -208,49 +208,72 @@ void WebPageInspectorController::setContinueLoadingCallback(const ProvisionalPag
...
+        CheckedPtr mainFrameTarget = m_targets.get(mainFrameTargetID);
+        ASSERT(mainFrameTarget && mainFrameTarget->isProvisional());
+        mainFrameTarget->didCommitProvisionalTarget();
+
+        ASSERT(oldMainFrameID);
+        String oldMainFrameTargetID = FrameInspectorTarget::toTargetID(*oldMainFrameID, oldProcessID);
+        targetAgent->didCommitProvisionalTarget(oldMainFrameTargetID, mainFrameTargetID);
     }
+
+    // Update target list to only include targets belonging to the committed page.
```
burg, line 257: Due to the function calls this looks suspiciously like mutation during iteration. I'd prefer we compute the delete-list, send notifications, then actually destroy the targets. (targetDestroyed might be better as `will/didDestroyTarget`, but that's existing code.)

When we do notifications as a side-effect of removeIf callbacks, it least to the potential for corrupting m_targets-- say if `targetAgent->targetDestroyed` indirectly led some other code to modify `m_targets`.

Source/WebKit/UIProcess/ProvisionalPageProxy.cpp
burg, line 292: Side quest: is m`_shouldReuseMainFrame` something that should factor into our decision whether to recreate the main frame or page target? I imagine if the processIdentifier doesn't actually change, then it might ask for that?

Source/WebKit/UIProcess/WebPageProxy.cpp
burg: Nit: you can adopt `oldMainFrameID` here.

burg: Ditto.

burg (commented):
I think the overall design is correct, but I did have some questions and suggestions to improve the code a tiny bit.

LayoutTests/http/tests/site-isolation/inspector/target/target-cross-origin-page-navigation.html
the-chenergy: Good points. I'll extract that to a utility function

For now since this is only used during test setups on environments we control, I think evaluating in the main world is fine. When we later need this utility in a more sensitive context, we can expand it and add test coverage. I don't think it's worth blocking on that now

Source/WebKit/UIProcess/ProvisionalPageProxy.cpp
the-chenergy: `m_shouldUseMainFrame` is a property of `ProvisionalPageProxy`, meaning that a cross-origin provisional loading was determined regardless of whether the main frame was reused. A reused `WebFrameProxy` will still be updated to point to the new web process when loading commits [(here)](https://github.com/WebKit/WebKit/blob/4fbffbbff74b460d601bf42dbfb3ffd31d2062e4/Source/WebKit/UIProcess/WebPageProxy.cpp#L5414), so the main frame target's id will change, and it's still correct to surface a new frame target to the frontend given the new id

In the reusing case, before loading commits, our new provisional frame target will point to the `FrameInspectorTarget` in the web process that doesn't yet have a UI process representation (`WebFrameProxy`). This adds to the story that `FrameInspectorTargetProxy` proxies the real target in the web process with or without a UI process frame proxy representation, identified by a frame id + web process reference

Does this help make the reusing case more clear?

Source/WebKit/UIProcess/Inspector/FrameInspectorTargetProxy.cpp
```diff
@@ -42,71 +43,42 @@ using namespace Inspector;
...
         resume();
 
-    if (RefPtr provisionalFrame = m_provisionalFrame.get()) {
-        protect(provisionalFrame->process())->send(Messages::WebFrame::DisconnectInspector(), protect(provisionalFrame->frame())->frameID());
-        return;
-    }
-
-    Ref frame = m_frame.get();
-    protect(frame->process())->send(Messages::WebFrame::DisconnectInspector(), frame->frameID());
+    protect(m_targetProcess.get())->send(Messages::WebFrame::DisconnectInspector(), m_frameID);
```
burg, line 66: so much cleaner, yay!

burg (approved):

PR #61184, opened 2026-03-23
Web Inspector: Remove obsolete per-resource content view settings after 308142@main

rcaliman-apple:
#### a6ff43363a20fa551dfb73dfe3c2d5b1418c91e3
<pre>
Web Inspector: Remove obsolete per-resource content view settings after <a href="https://commits.webkit.org/308142@main">308142@main</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=310558">https://bugs.webkit.org/show_bug.cgi?id=310558</a>
<a href="https://rdar.apple.com/173171572">rdar://173171572</a>

Reviewed by NOBODY (OOPS!).

Per-resource content view settings were removed in <a href="https://commits.webkit.org/308142@main">https://commits.webkit.org/308142@main</a>
in favor of per-MIME type settings. This patch removes the obsolete settings from Local Storage.

(WI.Setting.prototype.reset):
(WI.Setting._removeObsoleteResourceCurrentViewSettings):
(WI.Setting.reset): Deleted.
</pre>
https://github.com/WebKit/WebKit/commit/a6ff43363a20fa551dfb73dfe3c2d5b1418c91e3

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/151722 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/24503 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/18073 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/160462 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/105179 "Built successfully") | [loading 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/2352bc6f-2fad-4549-9a95-31947bd81322) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/24976 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/24796 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/117188 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/83172 "Passed tests") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/46a1b40d-7c33-4d8f-bb68-13f8d65d1f01) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/154682 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/19328 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/136143 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/97903 "Passed tests") | | [loading 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/193e2b13-9a2b-401e-abe8-ddeb017a3e56) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/18414 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/16368 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/8299 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/128044 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/14047 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/162928 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/6077 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/15638 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/125210 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/24302 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/20427 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/125386 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/24303 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/135843 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/80878 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/20417 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/12618 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/23919 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/88204 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/23611 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/23771 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/23671 "Built successfully") | | | |

### DIFF
--- a/Source/WebInspectorUI/UserInterface/Base/Setting.js
+++ b/Source/WebInspectorUI/UserInterface/Base/Setting.js
@@ -60,9 +60,9 @@ WI.Setting = class Setting extends WI.Object
         return value;
     }
 
-    static reset()
+    static reset(options = {})
     {
-        let prefix = WI.Setting._localStorageKeyPrefix;
+        let prefix = WI.Setting._localStorageKeyPrefix + (options.keyPrefix || "");
 
         let keysToRemove = [];
         for (let i = 0; i < window.localStorage.length; ++i) {
@@ -148,6 +148,20 @@ WI.Setting.isFirstLaunch = !!window.InspectorTest || (window.localStorage && Obj
 WI.Setting.Event = {
     Changed: "setting-changed"
 };
+
+// COMPATIBILITY (macOS X.X, iOS X.X): Per-resource content view settings were removed in favor of per-mimeType settings. See <https://webkit.org/b/308310>
+WI.Setting._removeObsoleteResourceCurrentViewSettings = (function() {
+    if (window.InspectorTest || !window.localStorage)
+        return;
+
+    let didCleanupSetting = new WI.Setting("did-remove-obsolete-resource-current-view-settings", false);
+    if (didCleanupSetting.value)
+        return;
+
+    WI.Setting.reset({keyPrefix: "resource-current-view-"});
+
+    didCleanupSetting.value = true;
+})();
 
 WI.EngineeringSetting = class EngineeringSetting extends WI.Setting
 {


Source/WebInspectorUI/UserInterface/Base/Setting.js
the-chenergy, line 153:
```suggestion
(function() {
```
Given this function doesn't return anything, maybe it's not worth storing that `undefined` return value? The comment already gives a good summary of what this function is for

dcrousso, line 161: hmm we've never bothered to cleanup old settings in the past (except for when migrating to a new key) cause it doesn't really cost very much to just have it (and it's not like it gets migrated to a new machine via iCloud or whatnot)

is there any evidence that this is actually necessary?

if so, i feel like there's a ton of other storage we also should probably be cleaning up too 
- all uses of `WI.Setting.migrateValue`
- collapsed state for each `WI.DetailsSection`/`WI.FolderTreeElement` since that's currently unique for each
- adding some sort of LRU eviction system for `WI.objectStores.breakpoints`/`WI.objectStores.domBreakpoints` since that will currently endlessly grow
- etc.

rcaliman-apple: > is there any evidence that this is actually necessary?

I don't think it's _necessary_, but we also don't have a way to measure how much stale date there is in WI local storage.

Purely anecdotal, I had 27 entries, and I almost never used the feature. Other developers might have a lot more.
Screenshot 2026-03-23 at 19 49 22

It would be nice to clean this up particularly because most of it was probably not even working due to the timing issue resolved by https://commits.webkit.org/308142@main where custom content views saved as a preference would load after rendering and therefore never get used, so the settings were useless.

It's not necessary, but a nice to have.

rcaliman-apple: > if so, i feel like there's a ton of other storage we also should probably be cleaning up too 
> - all uses of `WI.Setting.migrateValue`

`WI.Setting.migrateValue` already removes the local storage entry for the given key. Are you referring to removing the call to `WI.Setting.migrateValue()` itself?

> - adding some sort of LRU eviction system for `WI.objectStores.breakpoints`/`WI.objectStores.domBreakpoints` since that will currently endlessly grow

Yes, that would be nice. There is a lot of stale breakpoint data.

> - collapsed state for each `WI.DetailsSection`/`WI.FolderTreeElement` since that's currently unique for each

Perhaps the same approach as for breakpoints: a type of setting that has an expiration date of ~1 year since last accessed.

All this would be a separate patch anyway.

burg, line 152: Please fill in the placeholders before landing. But, I am not sure that this has anything to do with the operating system of the inspection target. Maybe just leave a comment without COMPATIBILITY?

burg (changes requested):
r=me on code changes, but this needs a test.

---

While Web Inspector doesn't have many settings migration paths, this is common in other parts of the browser. I support this change.

_Except_, we need some unit test coverage for the migration path. Can you please add a layout test?

Source/WebInspectorUI/UserInterface/Base/Setting.js
burg: These are all good ideas, it is also possible to add some telemetry to dump out the size of our indexeddb and other storage, but this hasn't been a priority.

dcrousso: this would only need a comment for macOS since it has nothing to do with iOS (i.e. it only affects the macOS host machine)

it's also not really a compatibility thing since it's purely about the frontend (i.e. it's irrelevant to what "version" of the protocol the backend supports)

as such maybe we just have a FIXME comment with a bug to remove this in a year or two?

dcrousso: yeah we don't need to save this to a variable since the IIFE will just return `undefined`

dcrousso, line 153: startup is a pretty sensitive codepath, so can we delay this until a less sensitive time?

perhaps we could put this inside a `requestIdleCallback` for few seconds at the end of `WI.contentLoaded`?

dcrousso: yeah i meant to remove all `WI.Setting.migrateValue` callers since i think most of them are many years old at this point

the difficulty with having any sort of timeout is that we'd need some other code that knows about all of these various random settings that need removal in order to check whether too much time has passed

we probably could do this in the same `requestIdleCallback` i suggested above, but even then we'd still need a registry for what settings can expire and when each was last accessed

dcrousso (changes requested):
overall im not against doing a cleanup but id prefer if it was more generalized to other things that also need it and overall it must be done in a performance sensitive way

PR #61301, opened 2026-03-25
ViewTransition snapshots are incorrectly stored in sRGB.

mattwoodrow:
#### df818e9231ffeaecd0156bd769c58c147ff66b54
<pre>
ViewTransition snapshots are incorrectly stored in sRGB.
<a href="https://bugs.webkit.org/show_bug.cgi?id=304981">https://bugs.webkit.org/show_bug.cgi?id=304981</a>
&lt;<a href="https://rdar.apple.com/problem/167634138">rdar://problem/167634138</a>&gt;

Reviewed by NOBODY (OOPS!).

ViewTransition snapshot were using sRGB, which can represent colors differently to how the live content displays them (in device-specific colorspace).

Adds a new option to ImageBuffer::create to get a snapshot buffer using the device colorspace.

* Source/WebCore/dom/ViewTransition.cpp:
* Source/WebCore/page/Chrome.cpp:
(WebCore::Chrome::createCompatibleImageBuffer const):
* Source/WebCore/page/Chrome.h:
(WebCore::ChromeClient::createCompatibleImageBuffer const):
* Source/WebCore/platform/GraphicsClient.h:
* Source/WebCore/platform/graphics/ImageBuffer.cpp:
(WebCore::ImageBuffer::create):
(WebCore::ImageBuffer::createCompatible):
* Source/WebCore/platform/graphics/ImageBuffer.h:
* Source/WebKit/WebProcess/GPU/graphics/RemoteRenderingBackendProxy.cpp:
(WebKit::RemoteRenderingBackendProxy::createCompatibleImageBuffer):
* Source/WebKit/WebProcess/GPU/graphics/RemoteRenderingBackendProxy.h:
(WebKit::RemoteRenderingBackendProxy::setColorSpace):
(WebKit::WebChromeClient::createCompatibleImageBuffer const):
* Source/WebKit/WebProcess/WebCoreSupport/WebWorkerClient.cpp:
(WebKit::GPUProcessWebWorkerClient::createCompatibleImageBuffer const):
(WebKit::WebWorkerClient::createCompatibleImageBuffer const):
* Source/WebKit/WebProcess/WebCoreSupport/WebWorkerClient.h:
(WebKit::WebPage::colorspaceChanged):
(WebKit::WebPage::ensureRemoteRenderingBackendProxy):
</pre>
https://github.com/WebKit/WebKit/commit/df818e9231ffeaecd0156bd769c58c147ff66b54

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/152060 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/24842 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/18497 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/160803 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/105517 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/153934 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/25334 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/25148 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/117460 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/83313 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/155020 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/19643 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/136554 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/98174 "Passed tests") | | [loading 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/982da4ef-8a61-4970-8226-37ec1b87db52) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/18721 "Passed tests") | [❌ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/16670 "Found 10 new API test failures: TestWebKitAPI.WKWebExtensionAPIExtension.GetViewsExcludesServiceWorkerBackground, TestWebKitAPI.WKWebExtensionAPIExtension.GetBackgroundPageFromTab, TestWebKitAPI.LockdownMode.NotSupportedFontLoadingAPI, TestWebKitAPI.LockdownMode.NotAllowedFont, TestWebKitAPI.JSHandle.WebpagePreferences, TestWebKitAPI.WKWebExtension.LoadFromDirectoryWithoutTrailingSlash, TestWebKitAPI.KeyboardEventTests.SmoothKeyboardScrolling, TestWebKitAPI.ImmediateActionTests.ImmediateActionOverText, TestWebKitAPI.EditorStateTests.TypingAttributesTextAlignmentStartEnd, TestWebKitAPI.WEBKIT.NavigationActionHasOpener14 (failure)") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/8637 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/128363 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/14433 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/163269 "Built successfully") | | 
| | [❌ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/6415 "Found 6 new failures in WebProcess/WebPage/WebPage.cpp") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/16025 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/125487 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/24640 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/20719 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/125663 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/24641 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/136229 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/81223 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/20673 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/13006 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/24258 "Built successfully") | [❌ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/88543 "Found 6 new failures in WebProcess/WebPage/WebPage.cpp") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/23949 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/24109 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/24010 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/dom/ViewTransition.cpp
+++ b/Source/WebCore/dom/ViewTransition.cpp
@@ -465,7 +465,9 @@ static RefPtr<ImageBuffer> snapshotElementVisualOverflowClippedToViewport(LocalF
         return nullptr;
     auto hostWindow = frameView->root() ? protect(frameView->root())->hostWindow() : nullptr;
 
-    auto buffer = ImageBuffer::create(paintRect.size(), RenderingMode::Accelerated, RenderingPurpose::Snapshot, scaleFactor, DestinationColorSpace::SRGB(), PixelFormat::BGRA8, hostWindow);
+    auto buffer = hostWindow
+        ? ImageBuffer::createCompatible(*hostWindow, paintRect.size(), RenderingMode::Accelerated, RenderingPurpose::Snapshot, scaleFactor, ImageBufferFormat { PixelFormat::BGRA8 })
+        : ImageBuffer::create(paintRect.size(), RenderingMode::Accelerated, RenderingPurpose::Snapshot, scaleFactor, DestinationColorSpace::SRGB(), ImageBufferFormat { PixelFormat::BGRA8 });
     if (!buffer)
         return nullptr;
 
--- a/Source/WebCore/page/Chrome.h
+++ b/Source/WebCore/page/Chrome.h
@@ -131,6 +131,7 @@ public:
     void setCursorHiddenUntilMouseMoves(bool) override;
 
     RefPtr<ImageBuffer> createImageBuffer(const FloatSize&, RenderingMode, RenderingPurpose, float resolutionScale, const DestinationColorSpace&, ImageBufferFormat) const override;
+    RefPtr<ImageBuffer> createCompatibleImageBuffer(const FloatSize&, RenderingMode, RenderingPurpose, float resolutionScale, ImageBufferFormat) const override;
     RefPtr<WebCore::ImageBuffer> sinkIntoImageBuffer(std::unique_ptr<WebCore::SerializedImageBuffer>) override;
 
 #if ENABLE(WEBGL)
--- a/Source/WebCore/platform/graphics/ImageBuffer.cpp
+++ b/Source/WebCore/platform/graphics/ImageBuffer.cpp
@@ -74,11 +74,8 @@ WTF_MAKE_TZONE_ALLOCATED_IMPL(SerializedImageBuffer);
 
 static const float MaxClampedLength = 4096;
 static const float MaxClampedArea = MaxClampedLength * MaxClampedLength;
-
 RefPtr<ImageBuffer> ImageBuffer::create(const FloatSize& size, RenderingMode renderingMode, RenderingPurpose purpose, float resolutionScale, const DestinationColorSpace& colorSpace, ImageBufferFormat pixelFormat, GraphicsClient* graphicsClient)
 {
-    RefPtr<ImageBuffer> imageBuffer;
-
     if (graphicsClient) {
         if (auto imageBuffer = graphicsClient->createImageBuffer(size, renderingMode, purpose, resolutionScale, colorSpace, pixelFormat))
             return imageBuffer;
@@ -116,6 +113,14 @@ RefPtr<ImageBuffer> ImageBuffer::create(const FloatSize& size, RenderingMode ren
 
     ASSERT_NOT_REACHED();
     return nullptr;
+}
+
+RefPtr<ImageBuffer> ImageBuffer::createCompatible(GraphicsClient& graphicsClient, const FloatSize& size, RenderingMode renderingMode, RenderingPurpose purpose, float resolutionScale, ImageBufferFormat pixelFormat)
+{
+    if (auto imageBuffer = graphicsClient.createCompatibleImageBuffer(size, renderingMode, purpose, resolutionScale, pixelFormat))
+        return imageBuffer;
+
+    return create(size, renderingMode, purpose, resolutionScale, DestinationColorSpace::SRGB(), pixelFormat, &graphicsClient);
 }
 
 ImageBuffer::ImageBuffer(Parameters parameters, const ImageBufferBackend::Info& backendInfo, const WebCore::ImageBufferCreationContext&, std::unique_ptr<ImageBufferBackend>&& backend, RenderingResourceIdentifier renderingResourceIdentifier)
--- a/Source/WebCore/platform/graphics/ImageBuffer.h
+++ b/Source/WebCore/platform/graphics/ImageBuffer.h
@@ -99,6 +99,7 @@ public:
     }
 
     WEBCORE_EXPORT static RefPtr<ImageBuffer> create(const FloatSize&, RenderingMode, RenderingPurpose, float resolutionScale, const DestinationColorSpace&, ImageBufferFormat, GraphicsClient* = nullptr);
+    WEBCORE_EXPORT static RefPtr<ImageBuffer> createCompatible(GraphicsClient&, const FloatSize&, RenderingMode, RenderingPurpose, float resolutionScale, ImageBufferFormat);
 
     template<typename BackendType, typename ImageBufferType = ImageBuffer, typename... Arguments>
     static RefPtr<ImageBufferType> create(const FloatSize& size, float resolutionScale, const DestinationColorSpace& colorSpace, ImageBufferFormat bufferFormat, RenderingPurpose purpose, const ImageBufferCreationContext& creationContext, Arguments&&... arguments)
--- a/Source/WebCore/page/Chrome.cpp
+++ b/Source/WebCore/page/Chrome.cpp
@@ -522,6 +522,11 @@ RefPtr<ImageBuffer> Chrome::createImageBuffer(const FloatSize& size, RenderingMo
     return m_client->createImageBuffer(size, renderingMode, purpose, resolutionScale, colorSpace, pixelFormat);
 }
 
+RefPtr<ImageBuffer> Chrome::createCompatibleImageBuffer(const FloatSize& size, RenderingMode renderingMode, RenderingPurpose purpose, float resolutionScale, ImageBufferFormat pixelFormat) const
+{
+    return m_client->createCompatibleImageBuffer(size, renderingMode, purpose, resolutionScale, pixelFormat);
+}
+
 RefPtr<ImageBuffer> Chrome::sinkIntoImageBuffer(std::unique_ptr<SerializedImageBuffer> imageBuffer)
 {
     return m_client->sinkIntoImageBuffer(WTF::move(imageBuffer));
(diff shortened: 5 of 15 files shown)


Source/WebCore/dom/ViewTransition.cpp
```diff
@@ -465,7 +465,7 @@ static RefPtr<ImageBuffer> snapshotElementVisualOverflowClippedToViewport(LocalF
         return nullptr;
     auto hostWindow = frameView->root() ? protect(frameView->root())->hostWindow() : nullptr;
 
-    auto buffer = ImageBuffer::create(paintRect.size(), RenderingMode::Accelerated, RenderingPurpose::Snapshot, scaleFactor, DestinationColorSpace::SRGB(), PixelFormat::BGRA8, hostWindow);
+    auto buffer = ImageBuffer::create(paintRect.size(), RenderingMode::Accelerated, RenderingPurpose::Snapshot, scaleFactor, std::nullopt, ImageBufferFormat { PixelFormat::BGRA8 }, hostWindow);
```
smfr: We know the display colorspace in the web process; can't we just use that?

Source/WebCore/platform/graphics/ImageBuffer.h
```diff
@@ -98,7 +98,7 @@ class ImageBuffer : public ThreadSafeRefCountedAndCanMakeThreadSafeWeakPtr<Image
         return create(size, mode, purpose, resolutionScale, colorSpace, ImageBufferFormat { bufferFormat }, client);
     }
 
-    WEBCORE_EXPORT static RefPtr<ImageBuffer> create(const FloatSize&, RenderingMode, RenderingPurpose, float resolutionScale, const DestinationColorSpace&, ImageBufferFormat, GraphicsClient* = nullptr);
+    WEBCORE_EXPORT static RefPtr<ImageBuffer> create(const FloatSize&, RenderingMode, RenderingPurpose, float resolutionScale, const std::optional<DestinationColorSpace>&, ImageBufferFormat, GraphicsClient* = nullptr);
```
weinig: Using an optional color space to denote seems more confusing than necessary.

If what you are trying to achieve is to make a ImageBuffer that is compatible with host window, than an overload of ImageBuffer::create, or something even more clear, like createUsingHostWindowColorSpace(), seems more clear (I think we had something like this in the past, maybe called createCompatible()?).

I have made an effort over the years to remove all the optional color space arguments, so please don't add more back :).

weinig (changes requested):

Source/WebCore/dom/ViewTransition.cpp
mattwoodrow: This is the web process. It's just that we know the display colorspace in WebKit, and this is WebCore code. So the PR is just plumbing to get that applied across the module boundary.

Source/WebCore/platform/graphics/ImageBuffer.h
mattwoodrow: This is already the overload that takes a HostWindow (`GraphicsClient` == `HostWindow`).

I could make an explicitly named overload (like `createCompatible`) which doesn't allow a null GraphicsClient, and omits the colorspace parameter?

weinig: I've also been trying to cull the places where DestinationColorSpace::SRGB() is passed, either explicitly, like here, or due to it being the default value for a parameter. Almost always these are wrong (except in the filter code, where it explicitly wants sRGB).

weinig: > I could make an explicitly named overload (like createCompatible) which doesn't allow a null GraphicsClient, and omits the colorspace parameter?

I think that would be much more clear.

Source/WebCore/dom/ViewTransition.cpp
weinig: > This is the web process. It's just that we know the display colorspace in WebKit, and this is WebCore code. So the PR is just plumbing to get that applied across the module boundary.

I believe we do actually have this info in WebCore in ScreenData, but trying to use that is probably not the best way to go about creating a compatible image buffer for cases like this.

Source/WebCore/dom/ViewTransition.cpp
```diff
@@ -465,7 +465,9 @@ static RefPtr<ImageBuffer> snapshotElementVisualOverflowClippedToViewport(LocalF
         return nullptr;
     auto hostWindow = frameView->root() ? protect(frameView->root())->hostWindow() : nullptr;
 
-    auto buffer = ImageBuffer::create(paintRect.size(), RenderingMode::Accelerated, RenderingPurpose::Snapshot, scaleFactor, DestinationColorSpace::SRGB(), PixelFormat::BGRA8, hostWindow);
+    auto buffer = hostWindow
```
weinig, line 468: Does it make sense to do anything when there is no host window? We already return nullptr when there is no frameView, so could we do the same for this condition?

Source/WebCore/page/Chrome.h
```diff
@@ -131,6 +131,7 @@ class Chrome : public HostWindow {
     void setCursorHiddenUntilMouseMoves(bool) override;
 
     RefPtr<ImageBuffer> createImageBuffer(const FloatSize&, RenderingMode, RenderingPurpose, float resolutionScale, const DestinationColorSpace&, ImageBufferFormat) const override;
+    RefPtr<ImageBuffer> createCompatibleImageBuffer(const FloatSize&, RenderingMode, RenderingPurpose, float resolutionScale, ImageBufferFormat) const override;
```
weinig, line 134: Since this is only color space compatible (the rendering mode, scale and format are potentially different), I think an even more specific name is warranted. Alternatively, removing all the parameters except the size and purpose?

Source/WebCore/platform/graphics/ImageBuffer.cpp
```diff
@@ -118,6 +115,14 @@ RefPtr<ImageBuffer> ImageBuffer::create(const FloatSize& size, RenderingMode ren
     return nullptr;
 }
 
+RefPtr<ImageBuffer> ImageBuffer::createCompatible(GraphicsClient& graphicsClient, const FloatSize& size, RenderingMode renderingMode, RenderingPurpose purpose, float resolutionScale, ImageBufferFormat pixelFormat)
+{
+    if (auto imageBuffer = graphicsClient.createCompatibleImageBuffer(size, renderingMode, purpose, resolutionScale, pixelFormat))
+        return imageBuffer;
+
+    return create(size, renderingMode, purpose, resolutionScale, DestinationColorSpace::SRGB(), pixelFormat, &graphicsClient);
```
weinig, line 123: Why is SRGB the correct choice here? If creating a compatible buffer can fail, I think we probably want to let the caller know rather than silently doing something else.

mattwoodrow: I don't have a good answer for this in the general case. I think ViewTransition would just use sRGB, because what else would we do?

This is making me think it might be better to just pull the colorspace we want from the ScreenData upfront, and use the normal allocation path.

PR #61338, opened 2026-03-25
Add support for `Secure` cookies for loopback hostnames.

xbhatnag:
#### aa297f89b9a60b60a28828a20bc1ea5c36da6840
<pre>
Add support for `Secure` cookies for loopback hostnames.
<a href="https://bugs.webkit.org/show_bug.cgi?id=281149">https://bugs.webkit.org/show_bug.cgi?id=281149</a>
<a href="https://rdar.apple.com/137604100">rdar://137604100</a>, <a href="https://rdar.apple.com/84509597">rdar://84509597</a>

Reviewed by Blaze Burg, Matthew Finkel, Anne Van Kesteren.

WebKit will now allow retrieval of `Secure` cookies created for loopback hostnames
on Cocoa builds only.

For macOS builds, this change only takes effect on &gt;= 26.0. Those builds
enforce that `localhost` and subdomains always resolve to loopback addresses.

Web Inspector will still not show `Secure` cookies for loopback hostnames with
insecure schemes (<a href="https://bugs.webkit.org/show_bug.cgi?id=311205).">https://bugs.webkit.org/show_bug.cgi?id=311205).</a>

This change will only work correctly once the underlying cookie storage starts
accepting `Secure` cookies for loopback hostnames. Hence, this change, in isolation
does not make `Secure` cookies work correctly for loopback hostnames.

These are the loopback hostnames that are detected by WebKit:
* localhost and *.localhost
* IPv4 address in 127.0.0.0/8 subnet

* LayoutTests/TestExpectations: get-cookies layout tests are no longer skipped, but expected to fail for now
* LayoutTests/http/tests/inspector/page/get-cookies-expected.txt: Expectations for non-HTTPS loopback URLs with Secure cookies
* LayoutTests/http/tests/inspector/page/get-cookies.html: New test case for non-HTTPS loopback URLs with Secure cookies
* LayoutTests/http/tests/inspector/page/resources/set-cookie.py: Updated Python server to support adding Secure attribute to cookie
* Source/WTF/wtf/PlatformHave.h: Added LOCALHOST_TIED_TO_LOOPBACK build constant
* Source/WebCore/loader/CookieJar.cpp:
(WebCore::CookieJar::shouldIncludeSecureCookies): Modified to now return `YES` for loopback hostnames, irrespective of scheme
* Source/WebCore/loader/CookieJar.h:
* Source/WebCore/platform/network/CacheValidation.cpp: Updated to include loopback hostnames when evaluating if Secure cookies should be included
(WebCore::cookieRequestHeaderFieldValue):
* Tools/TestWebKitAPI/CMakeLists.txt: Updated to include CookieJar.cpp
* Tools/TestWebKitAPI/TestWebKitAPI.xcodeproj/project.pbxproj: Updated to include CookieJar.cpp
* Tools/TestWebKitAPI/Tests/WebCore/CookieJar.cpp: Added.
(TestWebKitAPI::TEST(CookieJar, ShouldIncludeSecureCookiesForHTTPS)): Checks that HTTPS scheme means Secure cookies are allowed.
(TestWebKitAPI::TEST(CookieJar, ShouldNotIncludeSecureCookiesForPlainHTTP)): Checks that other HTTP addresses means Secure cookies are not allowed.
(TestWebKitAPI::TEST(CookieJar, ShouldNotIncludeSecureCookiesForNonLocalHostnames)): Checks that other variations on localhost means Secure cookies are not allowed.
(TestWebKitAPI::TEST(CookieJar, ShouldNotIncludeSecureCookiesForNonLoopbackIPv4)): Checks that other variations on IPv4 addresses means Secure cookies are not allowed.
(TestWebKitAPI::TEST(CookieJar, ShouldIncludeSecureCookiesForIPv6Loopback)): Checks that IPv6 loopback addresses means Secure cookies are allowed.
(TestWebKitAPI::TEST(CookieJar, ShouldIncludeSecureCookiesForIPv4Loopback)): Checks that IPv4 loopback addresses means Secure cookies are allowed.
(TestWebKitAPI::TEST(CookieJar, ShouldIncludeSecureCookiesForLocalhost)): Checks that localhost means Secure cookies are allowed.
(TestWebKitAPI::TEST(CookieJar, ShouldIncludeSecureCookiesForLocalhostSubdomains)): Checks that localhost subdomains means Secure cookies are allowed.
(TestWebKitAPI::TEST(CookieJar, ShouldIncludeSecureCookiesForNormalizedIPv4Loopback)): Checks that WHATWG normalized IPv4 loopback address means Secure cookies are allowed.

Canonical link: <a href="https://commits.webkit.org/310542@main">https://commits.webkit.org/310542@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/878bc4ac9abab2f3bc181dc4bb8c134bece376e0

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/154108 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/27365 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/20523 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/162860 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/107576 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/155981 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/27497 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/27216 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/119185 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/84262 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/157067 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/21438 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/138397 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/99885 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/20524 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/18520 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/10693 "Built successfully") | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/146157 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/130186 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/16241 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/165333 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/14950 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/8543 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/17850 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/127277 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/26911 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/22564 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/127428 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/34579 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/26834 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/138034 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/83421 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/22299 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/14822 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/185743 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/26526 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/90629 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/47638 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/26106 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/26350 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/26178 "Built successfully") | | |


Source/WebCore/loader/CookieJar.cpp
```diff
@@ -54,14 +54,55 @@ static ShouldRelaxThirdPartyCookieBlocking shouldRelaxThirdPartyCookieBlocking(c
     return ShouldRelaxThirdPartyCookieBlocking::No;
 }
 
+static bool isLoopbackIPAddress(StringView host)
```
xbhatnag: These static functions are borrowed from `SecurityOrigin.cpp`: https://github.com/WebKit/WebKit/blob/main/Source/WebCore/page/SecurityOrigin.cpp#L633

Source/WebCore/loader/CookieJar.cpp
```diff
@@ -54,14 +54,55 @@ static ShouldRelaxThirdPartyCookieBlocking shouldRelaxThirdPartyCookieBlocking(c
...
     return adoptRef(*new CookieJar(WTF::move(storageSessionProvider)));
 }
 
+
 IncludeSecureCookies CookieJar::shouldIncludeSecureCookies(const URL& url)
 {
-    return url.protocolIs("https"_s) ? IncludeSecureCookies::Yes : IncludeSecureCookies::No;
+    bool local = isLocalHostOrLoopbackIPAddress(url.host());
+    bool https = url.protocolIs("https"_s);
+    return local || https ? IncludeSecureCookies::Yes : IncludeSecureCookies::No;
```
burg: Nit: please include parentheses to make this less ambiguous.

Source/WebCore/loader/CookieJar.cpp
burg: Is there a reason to duplicate this? It'd be more typical to export as static functions from one or the other file. In fact it already appears to be WEBCORE_EXPORT.

burg (changes requested):
The code changes and tests look good, but i'm concerned we are duplicating code (and tests). Can you find a place to share it?

For Web Inspector we have resorted to files like InspectorResourceUtilities.h to hold all these free helper functions.

Source/WebCore/loader/CookieJar.cpp
```diff
@@ -54,14 +54,55 @@ static ShouldRelaxThirdPartyCookieBlocking shouldRelaxThirdPartyCookieBlocking(c
...
+
+    return false;
+}
+
 Ref<CookieJar> CookieJar::create(Ref<StorageSessionProvider>&& storageSessionProvider)
 {
     return adoptRef(*new CookieJar(WTF::move(storageSessionProvider)));
 }
 
+
```
burg: Nit: extra newline

Source/WebInspectorUI/UserInterface/Models/CookieStorageObject.js
```diff
@@ -39,8 +39,7 @@ WI.CookieStorageObject = class CookieStorageObject
             return false;
 
         return parsedURL.path.startsWith(cookie.path)
-            && (!cookie.port || parsedURL.port === cookie.port)
-            && (!cookie.secure || parsedURL.scheme === "https");
+            && (!cookie.port || parsedURL.port === cookie.port);
```
burg: We don't have great test coverage for this part. But does this PR materially affect the result of `Page.getCookies`, which is used by the frontend to grab cookies? In that case, I think we should have an inspector layout test, maybe add some missing loopback test cases to `LayoutTests/inspector/page/getCookies.html`. Its OK if some of the test cases fail until the blocker is resolved.

cc @stwrt who's also working on Page domain related things for Site Isolation.

burg:
Also, please include links to bugs for the next step(s).

`CacheValidation.cpp` lines 352 and 360 independently compute `IncludeSecureCookies` via `protocolIs("https")`, bypassing `CookieJar::shouldIncludeSecureCookies()`. The Curl backend has similar patterns in `NetworkDataTaskCurl.cpp` and `WebSocketTaskCurl.cpp`. Can you file a followup bug to address this? I'm guessing they should all be the same behavior...

Source/WebInspectorUI/UserInterface/Models/CookieStorageObject.js
burg: This affects ALL URLs, not just loopback. Previously, viewing `http://example.com` would hide Secure cookies in Storage tab; now they appear. Right?

Would be worth clarifying. cc @dcrousso

xbhatnag: As I understand it (and feel free to correct me please), by removing this logic from the inspector, the determination of whether a `Secure` cookie is shown or not comes down to the individual `NetworkStorageSession` implementations.

Here's how that logic is traced:
1. [The refresh of the Inspector page causes the `PageAgent` to get cookies](https://github.com/WebKit/WebKit/blob/main/Source/WebInspectorUI/UserInterface/Views/CookieStorageContentView.js#L515)
2. [`InspectorPageAgent` asks `CookieJar` for the cookies](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/inspector/agents/InspectorPageAgent.cpp#L397)
3. [`CookieJar` asks the `NetworkStorageSession` for the cookies](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/loader/CookieJar.cpp#L190)

There are 3 `NetworkStorageSession`s:
* [`NetworkStorageSessionCocoa` will call SPI to retrieve cookies.](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/platform/network/cocoa/NetworkStorageSessionCocoa.mm#L308) This SPI will include `Secure` cookies only for `HTTPS` schemes.
* [`NetworkStorageSessionCurl` will use `CookieJarDB` to retrieve cookies](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/platform/network/curl/NetworkStorageSessionCurl.cpp#L239), but [this database doesn't seem to care about enforcing `secure` property based on the input URL.](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/platform/network/curl/CookieJarDB.cpp#L425)
* [`NetworkStoragesSessionSoup` will include `Secure` cookies only for `HTTPS` schemes.](https://github.com/WebKit/WebKit/blob/1b2711ed3412323ea22275360a7404db46fb65da/Source/WebCore/platform/network/soup/NetworkStorageSessionSoup.cpp#L685)

Source/WebCore/loader/CookieJar.cpp
xbhatnag: Removed duplicated logic.

Source/WebCore/loader/CookieJar.cpp
xbhatnag: Fixed

Source/WebCore/loader/CookieJar.cpp
xbhatnag: Fixed

Source/WebInspectorUI/UserInterface/Models/CookieStorageObject.js
dcrousso: so will this (or any of the other changes in this PR) actually modify the behavior of Web Inspector?

if so, we may need to keep this around to preserve existing behavior when inspecting an older backend (e.g. an iOS device) as that wont have any of these changes

xbhatnag: Given my analysis above, this will have an impact when WebKit uses the `curl` backend because `NetworkStorageSessionCurl::getRawCookies` will not filter out `Secure` cookies when given an insecure HTTP URL.

Does that help determine whether we need to preserve existing behavior?

My current plan is to land this change (including the inspector modifications) and then file a bug to fix the curl backend such that:
* It filters `Secure` cookies for calls to `getRawCookies`
* It allows `Secure` cookies for loopback hostnames in `NetworkDataTaskCurl` and `WebSocketTaskCurl`

xbhatnag: Given that we now allow `Secure` cookies for loopback hostnames, I believe this would materially affect the result of `Page.getCookies` in some contexts.

Let me look into adding an inspector layout test for these cases.

xbhatnag: Added new Inspector layout test cases. Re-enabled the `get-cookies.html` test that was disabled.

xbhatnag: https://bugs.webkit.org/show_bug.cgi?id=310830 has been filed for cURL specific changes

dcrousso: normally, if there's a change in the backend we would leave the code in `Source/WebInspectorUI` as-is with a `// COMPATIBILITY` comment so that we don't affect the behavior when a newer frontend is inspecting an older backend, but im not sure if platforms using the `curl` backend support that concept (i.e. as far as i know only iOS and macOS have this concept)

id double check with someone working on the `curl` backend to make sure

xbhatnag: To make forward progress on this PR, I'm undoing the Web Inspector front-end filtering changes.

I've filed a bug to record what changes would need to be made to Web Inspector: https://bugs.webkit.org/show_bug.cgi?id=311205

Source/WebCore/loader/CookieJar.cpp
```diff
@@ -40,6 +40,9 @@
 #include "NetworkingContext.h"
 #include "Page.h"
 #include "PlatformStrategies.h"
+#if HAVE(LOCALHOST_TIED_TO_LOOPBACK)
```
burg: Nit: typically, #if guarded includes go after the unconditional includes

burg (approved):

Source/WebCore/loader/CookieJar.cpp
xbhatnag: Done!

Source/WebCore/loader/CookieJar.cpp
```diff
@@ -61,7 +65,13 @@ Ref<CookieJar> CookieJar::create(Ref<StorageSessionProvider>&& storageSessionPro
 
 IncludeSecureCookies CookieJar::shouldIncludeSecureCookies(const URL& url)
 {
+#if HAVE(LOCALHOST_TIED_TO_LOOPBACK)
+    bool loopback = SecurityOrigin::isLocalHostOrLoopbackIPAddress(url.host());
+    bool https = url.protocolIs("https"_s);
```
sysrqb: I think we should use `SecurityOrigin::isPotentiallyTrustworthy` instead, so something like
```
return SecurityOrigin::create(url)->isPotentiallyTrustworthy() ? IncludeSecureCookies::Yes : IncludeSecureCookies::No;
```

It covers a few more cases, which are debatable, but I think being consistent with what we consider "secure" is better.

Source/WebCore/loader/CookieJar.h
```diff
@@ -57,6 +57,11 @@ class WEBCORE_EXPORT CookieJar : public RefCountedAndCanMakeWeakPtr<CookieJar> {
 
     String cookieRequestHeaderFieldValue(Document&, const URL&) const;
 
+    // Returns `Yes` if the URL would allow `Secure` cookies to be retrieved.
+    // This is usually determined by checking if the URL scheme is `HTTPS`
+    // or if the hostname is a loopback address.
```
sysrqb: WebKit has a style preference against adding comments if they aren't absolutely necessary.

Source/WTF/wtf/PlatformHave.h
```diff
@@ -342,6 +342,16 @@
 #define HAVE_MEDIA_ACCESSIBILITY_PROFILES_API 1
 #endif
 
+// `localhost` is tied to loopback addresses at the DNS resolver level
+// for all non-Darwin platforms and for Darwin platforms >= 26.0
```
sysrqb: ditto

Tools/TestWebKitAPI/Tests/WebCore/CookieJar.cpp
```diff
@@ -0,0 +1,107 @@
...
+    // "localhost" must appear exactly or as the rightmost label.
+    EXPECT_EQ(CookieJarForTesting::shouldIncludeSecureCookies(URL("http://notlocalhost/"_s)), WebCore::IncludeSecureCookies::No);
+    EXPECT_EQ(CookieJarForTesting::shouldIncludeSecureCookies(URL("http://localhost.example.com/"_s)), WebCore::IncludeSecureCookies::No);
+}
+
+TEST(CookieJar, ShouldNotIncludeSecureCookiesForNonLoopbackIPv4)
+{
+    // Starts with "127." but has non-digit characters.
+    EXPECT_EQ(CookieJarForTesting::shouldIncludeSecureCookies(URL("http://127.example.com/"_s)), WebCore::IncludeSecureCookies::No);
+
```
sysrqb: nit: blank line

sysrqb (approved):

Source/WTF/wtf/PlatformHave.h
```diff
@@ -342,6 +342,16 @@
...
+#if !PLATFORM(COCOA) \
+    || ((PLATFORM(MAC) && __MAC_OS_X_VERSION_MIN_REQUIRED >= 260000) \
+    || ((PLATFORM(MACCATALYST) || PLATFORM(IOS)) && __IPHONE_OS_VERSION_MIN_REQUIRED >= 260000) \
```
annevk: For iOS family apart from maybe Mac Catalyst you can just enable it unconditionally. We’ll also cover watchOS that way.

Source/WTF/wtf/PlatformHave.h
xbhatnag: done

Source/WTF/wtf/PlatformHave.h
xbhatnag: done

Tools/TestWebKitAPI/Tests/WebCore/CookieJar.cpp
xbhatnag: done

Source/WebCore/loader/CookieJar.h
xbhatnag: done

Source/WebCore/loader/CookieJar.cpp
xbhatnag: done

Source/WTF/wtf/PlatformHave.h
```diff
@@ -342,6 +342,13 @@
 #define HAVE_MEDIA_ACCESSIBILITY_PROFILES_API 1
 #endif
 
+#if PLATFORM(COCOA) \
+    && ((PLATFORM(MAC) && __MAC_OS_X_VERSION_MIN_REQUIRED >= 260000) \
+    || (PLATFORM(MACCATALYST) && __IPHONE_OS_VERSION_MIN_REQUIRED >= 260000) \
+    || PLATFORM(APPLETV) || PLATFORM(IOS) || PLATFORM(VISION))
```
annevk: I think we should use PLATFORM(IOS_FAMILY) here so we don't exclude watchOS.

xbhatnag: Done

PR #59257, opened 2026-02-23
[EvaluationTimeZoom] Convert border, outline and column-rule-width properties to evaluation time zoom

weinig:
#### 6ea7836374f5a9debb460a109f0fa7ea24cc84d2
<pre>
[EvaluationTimeZoom] Convert border, outline and column-rule-width properties to evaluation time zoom
<a href="https://bugs.webkit.org/show_bug.cgi?id=308486">https://bugs.webkit.org/show_bug.cgi?id=308486</a>

Reviewed by Darin Adler and Antti Koivisto.

Converts the &lt;line-width&gt; value properties, `border-*-width`, `outline-width` and
`column-rule-width`, as well as the `outline-offset` to evaluation time zoom.

Special thanks to Sammy Gill for all the work need to make this perf-neutral.

Tests: imported/w3c/web-platform-tests/css/css-viewport/zoom/column-rule-width.html
       imported/w3c/web-platform-tests/css/css-viewport/zoom/outline-offset.html
       imported/w3c/web-platform-tests/css/css-viewport/zoom/outline-width.html
       imported/w3c/web-platform-tests/css/css-viewport/zoom/reference/column-rule-width-ref.html
       imported/w3c/web-platform-tests/css/css-viewport/zoom/reference/outline-offset-ref.html
       imported/w3c/web-platform-tests/css/css-viewport/zoom/reference/outline-width-ref.html
* LayoutTests/TestExpectations:
* LayoutTests/imported/w3c/web-platform-tests/css/css-viewport/zoom/column-rule-width-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-viewport/zoom/column-rule-width.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-viewport/zoom/outline-offset-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-viewport/zoom/outline-offset.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-viewport/zoom/outline-width-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-viewport/zoom/outline-width.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-viewport/zoom/reference/column-rule-width-ref.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-viewport/zoom/reference/outline-offset-ref.html: Added.
[...]

PR #61340, opened 2026-03-25
Enhanced <select>: correct <option> outline

annevk:
#### 32c2254f39e0d5e781893f478bbc0761b7a1cd29
<pre>
Enhanced &lt;select&gt;: correct &lt;option&gt; outline
<a href="https://bugs.webkit.org/show_bug.cgi?id=310721">https://bugs.webkit.org/show_bug.cgi?id=310721</a>

Reviewed by Antti Koivisto.

Since ::picker(select) has overflow: auto, we need to draw the &lt;option&gt;
outline on the inside. To do this we introduce an -internal-inset value
for outline-offset which we hope to standardize down the line.

We do this by adding a struct for OutlineOffset using the
ValueOrKeyword&lt;&gt; template and add the necessary Blending support.

Not tested via web-platform-tests because this is user-agent-specific.

Canonical link: <a href="https://commits.webkit.org/310364@main">https://commits.webkit.org/310364@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/77f50e3150b504be9416e6dd95a09cc550a399f8

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/153621 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/26405 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/20022 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/162371 "Built successfully") | [💥 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/107079 "An unexpected error occured. Recent messages:Printed configuration") | [loading 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/cb383e9e-163a-4c83-aecb-765da0320c5f) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/155494 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/26933 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/26727 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/118773 "Passed tests") | [💥 🧪 win-tests](https://ews-build.webkit.org/#/builders/59/builds/107079 "An unexpected error occured. Recent messages:Printed configuration") | [loading 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/46c526a9-9045-466e-a28e-e60dca635036) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/156580 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/21024 "Passed tests") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/137930 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/99484 "Passed tests") | | [loading 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b4b58afc-24fe-4aea-83eb-d4c3ad5a0653) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/20103 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/18048 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/10204 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/129750 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/15790 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/164842 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/7976 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/17384 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/126848 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/26202 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/22082 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/127012 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/26204 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/137585 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/82872 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/23483 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/21925 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/14366 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/25821 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/90108 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/25512 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/25672 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/25572 "Built successfully") | | | |

### DIFF
--- a/LayoutTests/platform/ios/fast/forms/select/base/option-focus-visible-outline-offset-expected.txt
+++ b/LayoutTests/platform/ios/fast/forms/select/base/option-focus-visible-outline-offset-expected.txt
@@ -1,0 +1,17 @@
+Tests that option elements in a base-select have a negative outline-offset equal to the outline width when focus-visible.
+
+On success, you will see a series of "PASS" messages, followed by "TEST COMPLETE".
+
+
+FAIL outlineOffset == "-3px" || outlineOffset == "-2px" should be true. Was false.
+PASS getComputedStyle(opt1).outlineOffset is "0px"
+PASS getComputedStyle(opt1).outlineOffset is "0px"
+PASS successfullyParsed is true
+Some tests failed.
+
+TEST COMPLETE
+
+Option 1
+Option 2
+Option 3
+
--- a/Source/WebCore/css/CSSProperties.json
+++ b/Source/WebCore/css/CSSProperties.json
@@ -7037,14 +7037,20 @@
         "outline-offset": {
             "animation-type": "by computed value",
             "initial": "0px",
+            "values": [
+                {
+                	"value": "-internal-inset",
+                	"status": "internal"
+                }
+            ],
             "codegen-properties": {
-                "computed-style-initial-constexpr": true,
                 "computed-style-storage-path": ["m_nonInheritedData", "backgroundData", "outline"],
                 "computed-style-storage-container": "struct",
                 "computed-style-storage-kind": "reference",
-                "computed-style-type": "Style::Length<>",
+                "computed-style-type": "Style::OutlineOffset",
                 "skip-render-style-getter": true,
-                "parser-grammar": "<length>"
+                "style-extractor-custom": true,
+                "parser-grammar": "<length> | <<values>>"
             },
             "specification": {
                 "category": "css-ui",
--- a/Source/WebCore/css/html.css
+++ b/Source/WebCore/css/html.css
@@ -1185,6 +1185,10 @@ select option:disabled {
     color: color-mix(currentColor 50%, transparent);
 }
 
+select option:focus-visible {
+    outline-offset: -internal-inset;
+}
+
 select option::checkmark {
     content: "\2713" / "";
 }
--- a/Source/WebCore/rendering/style/RenderStyle.cpp
+++ b/Source/WebCore/rendering/style/RenderStyle.cpp
@@ -402,7 +402,10 @@ Style::LineWidth RenderStyle::usedColumnRuleWidth() const
 
 Style::Length<> RenderStyle::usedOutlineOffset() const
 {
-    return m_computedStyle.outline().outlineOffset;
+    auto& outline = m_computedStyle.outline();
+    if (outline.outlineOffset.isInternalInset())
+        return Style::Length<> { -Style::evaluate<float>(usedOutlineWidth(), Style::ZoomNeeded { }) };
+    return *outline.outlineOffset.tryLength();
 }
 
 Style::LineWidth RenderStyle::usedOutlineWidth() const
--- a/Source/WebCore/style/values/StyleValueTypes.h
+++ b/Source/WebCore/style/values/StyleValueTypes.h
@@ -1240,6 +1240,24 @@ template<typename... StyleTypes> struct Blending<Variant<StyleTypes...>> {
     }
 };
 
+// Specialization for `ValueOrKeyword`.
+template<ValueOrKeywordDerived T> struct Blending<T> {
+    auto canBlend(const T& a, const T& b) -> bool
+    {
+        if (a.isKeyword() != b.isKeyword())
+            return false;
+        if (a.isKeyword())
+            return true;
+        return WebCore::Style::canBlend(*a.tryValue(), *b.tryValue());
+    }
+    auto blend(const T& a, const T& b, const auto& context) -> T
+    {
+        if (a.isKeyword())
+            return context.progress < 0.5 ? a : b;
+        return T { WebCore::Style::blend(*a.tryValue(), *b.tryValue(), context) };
+    }
+};
+
 // Specialization for `SpaceSeparatedVector`.
 template<typename StyleType, size_t inlineCapacity> struct Blending<SpaceSeparatedVector<StyleType, inlineCapacity>> {
     auto equals(const SpaceSeparatedVector<StyleType, inlineCapacity>& a, const SpaceSeparatedVector<StyleType, inlineCapacity>& b) -> bool
--- a/Source/WebCore/style/values/outline/StyleOutlineOffset.h
+++ b/Source/WebCore/style/values/outline/StyleOutlineOffset.h
@@ -1,0 +1,44 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL APPLE INC. OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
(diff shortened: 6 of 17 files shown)


Source/WebCore/css/html.css
```diff
@@ -1185,6 +1185,10 @@ select option:disabled {
     color: color-mix(currentColor 50%, transparent);
 }
 
+select option:focus-visible {
+    outline-offset: -3px;
```
nt1m: Would be nice if we had a -internal- keyword to represent this value instead of repeating it.

Better yet, maybe we could standardize a keyword like `outline-offset: inset` to allow authors to replicate this effect without using magic values

annevk: Sure

LayoutTests/platform/ios/fast/forms/select/base/option-focus-visible-outline-offset-expected.txt
```diff
@@ -0,0 +1,16 @@
+Tests that option elements in a base-select have a negative outline-offset equal to the outline width when focus-visible.
+
+On success, you will see a series of "PASS" messages, followed by "TEST COMPLETE".
+
+
+FAIL outlineOffset == "-3px" || outlineOffset == "-2px" should be true. Was false.
```
nt1m, line 6: Could you change the assert so this passes?

Ideally this should compute to `inset` once it's standardized

nt1m (commented):
Would love for @weinig to take a look at this

Source/WebCore/rendering/style/RenderStyle.cpp
```diff
@@ -403,6 +403,11 @@ Style::LineWidth RenderStyle::usedColumnRuleWidth() const
 Style::Length<> RenderStyle::usedOutlineOffset() const
 {
     auto& outline = m_computedStyle.outline();
+    if (outline.outlineOffsetIsInternalInset) {
+        if (static_cast<OutlineStyle>(outline.outlineStyle) == OutlineStyle::Auto)
+            return Style::Length<> { -RenderTheme::singleton().platformFocusRingWidth() * usedZoom() };
+        return Style::Length<> { -Style::evaluate<float>(outline.outlineWidth, Style::ZoomNeeded { }) };
```
nt1m: Could we do various reftests combining this new value, various outline width/styles & CSS Zoom ?

LayoutTests/platform/ios/fast/forms/select/base/option-focus-visible-outline-offset-expected.txt
annevk: This is a problem with the select not opening due to lack of touch support on OpenSource.

Source/WebCore/rendering/style/RenderStyle.cpp
annevk: No, because this value is limited to the user agent style sheet and I don't think it warrants introducing even more infrastructure before it's standardized. But I added another test. There's also an existing issue with zoom that @weinig is fixing here: https://github.com/WebKit/WebKit/pull/59257

Source/WebCore/css/CSSProperties.json
```diff
@@ -7037,14 +7037,22 @@
...
                 "computed-style-initial-constexpr": true,
                 "computed-style-storage-path": ["m_nonInheritedData", "backgroundData", "outline"],
                 "computed-style-storage-container": "struct",
                 "computed-style-storage-kind": "reference",
                 "computed-style-type": "Style::Length<>",
                 "skip-render-style-getter": true,
-                "parser-grammar": "<length>"
+                "style-builder-custom": "All",
+                "style-extractor-custom": true,
+                "parser-grammar": "<length> | <<values>>"
```
anttijk, line 7053: We should not need custom code for this. `<length> | keyword` case should be handled by standard infrastructure. There are plenty of similar properties already.

But do we really need to expand this standard property with a non-standard keyword like this? Would a separate internal property be simpler and cleaner?

anttijk (changes requested):

Source/WebCore/rendering/style/RenderStyle.cpp
anttijk: Said added cssInternalAutoBaseParsingEnabled internal setting which allowed testing a UA-sheet only value with regular layout tests and  `<!-- webkit-test-runner [ CSSInternalAutoBaseParsingEnabled=true ] -->`. Similar approach would work here. (or maybe we could have a setting that enables all UA sheet only values)

anttijk (approved):

Source/WebCore/style/values/outline/StyleOutlineOffset.h
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+#pragma once
+
+#include <WebCore/StyleLengthWrapper.h>
+
+namespace WebCore::Style {
+
+// <'outline-offset'> = <length> | -internal-inset
+struct OutlineOffset : LengthWrapperBase<LengthPercentage<>, CSS::Keyword::InternalInset> {
```
anttijk: should this be just `LengthWrapperBase<Length<>, CSS::Keyword::InternalInset> `?

anttijk: I see for example
```
// <clip-edge> = <length> | auto
struct ClipEdge : ValueOrKeyword<Length<>, CSS::Keyword::Auto> {
```

anttijk: cc @weinig

anttijk (approved):

Source/WebCore/style/values/StyleValueTypes.h
```diff
@@ -1240,6 +1240,24 @@ template<typename... StyleTypes> struct Blending<Variant<StyleTypes...>> {
     }
 };
 
+// Specialization for `ValueOrKeyword`.
+template<ValueOrKeywordDerived T> struct Blending<T> {
```
nullhook, line 1244: I'm guessing this generic specialization means every other ValueOrKeyword-derived type now gets blending support. I'm not sure what side effects that might have.

shouldn't `Blending<T>` be specialized to StyleOutlineOffset type instead of being generic?

anttijk: Having a default blending behavior makes sense to me (like LengthWrapperBase has) though maybe this would need some adjustments to only apply when the value type is blendable.

annevk: It shouldn't have any bad side effects on the set of values we currently support. However, making it more tightly scoped is probably a good idea. I created https://github.com/WebKit/WebKit/pull/61897.

PR #61427, opened 2026-03-26
Web Inspector: Styles: Newly added CSS property is intermittently lost when editing a rule after changing its selector

rcaliman-apple:
#### 34544098f193a69078e4b6458d8bdfd2e17b65a3
<pre>
Web Inspector: Styles: Newly added CSS property is intermittently lost when editing a rule after changing its selector
<a href="https://bugs.webkit.org/show_bug.cgi?id=310810">https://bugs.webkit.org/show_bug.cgi?id=310810</a>
<a href="https://rdar.apple.com/164971557">rdar://164971557</a>

Reviewed by BJ Burg.

There is a race condition when editing the selector of an empty CSS rule
and inserting a new CSS property which causes the eviction of the new property
from `WI.CSSStyleDeclaration.prototype._properties` causing all subsequent input
to be discarded.

After adding a new CSS rule and changing its selector, the style body text on
both the frontend and backend is empty. The next style refresh event
(`WI.DOMNodeStyles.Event.Refreshed`) caused by committing the selector text is explicitly
ignored (`WI.DOMNodeStyles.prototype._ignoreNextContentDidChangeForStyleSheet`).

Buf if another event occurs before that and `WI.CSSStyleDeclaration.prototype.update()`
reacts to it, the early return guard that checks for the active editing state (`_locked`)
and the two empty style body texts is not satisfied.

This replaces the list of properties (`WI.CSSStyleDeclaration.prototype._properties`) with the backend&apos;s empty array, moving the locally created blank CSS property to `WI.CSSStyleDeclaration._pendingProperties`. After this, `WI.CSSStyleDeclaration.generateFormattedText()` produces an empty string
(no properties to format), so the `WI.CSSStyleDeclaration.text` setter early returns
on every keystroke (&quot;&quot; === &quot;&quot;), and `WI.DOMNodeStyles.prototype.changeStyleText()`
is never called.

This patch makes the guard in `WI.CSSStyleDeclaration.prototype.update()` uncondtional
to prevent updates while the style is being edited.

The stated reason for having a style body text check in that guard was to allow for updating
any property warnings, but that&apos;s not necessary in this step. The validity and warnings are
updated in `WI.SpreadSheetStyleProperty.prototype.updateStatus()` in response to
`WI.CSSProperty.Event.Changed` triggered by `WI.CSSProperty.update()` whenever the styles get
refreshed (`WI.DOMNodeStyles.prorotype._parseStylePropertyPayload`), regardless of the style
body text check set `WI.CSSStyleDeclaration`.

Canonical link: <a href="https://commits.webkit.org/310921@main">https://commits.webkit.org/310921@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/59e3e3cbdc10c87075ea1c49058d173dcd1527f8

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/152449 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/25231 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/18830 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/161192 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/105906 "Built successfully") | [loading 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/85300003-78f5-4402-9525-559a5bfa21c5) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/25759 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/25537 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/117802 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/83495 "3 flakes 2 failures") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/155409 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/20019 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/136864 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/98516 "Passed tests") | | [loading 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b3c1671b-c801-4045-96bf-5afb44e56230) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/19095 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/17022 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/9028 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/128745 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/14738 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/163662 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/6804 "Built successfully") | [❌ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/16332 "Found 1 new test failure: fast/images/mac/play-pause-individual-animation-context-menu-items.html (failure)") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/125838 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/25029 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/21061 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/126009 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/34853 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/25031 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/136534 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/81631 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/21005 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/13313 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/24647 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/88933 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/24338 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/24498 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/24399 "Built successfully") | | | |

### DIFF
--- a/Source/WebInspectorUI/UserInterface/Models/CSSStyleDeclaration.js
+++ b/Source/WebInspectorUI/UserInterface/Models/CSSStyleDeclaration.js
@@ -198,8 +198,7 @@ WI.CSSStyleDeclaration = class CSSStyleDeclaration extends WI.Object
             return;
         }
 
-        // Allow updates from the backend when text matches because `properties` may contain warnings that need to be shown.
-        if (this._locked && !options.forceUpdate && text !== this._text)
+        if (this._locked && !options.forceUpdate)
             return;
 
         text = text || "";


Source/WebInspectorUI/UserInterface/Models/CSSStyleDeclaration.js
dcrousso, line 202: as far as i can tell it seems like this was added in [197790@main](https://commits.webkit.org/197790@main)

can you please verify that this doesnt break any of the scenarios described in that bug?

also, this absolutely needs some tests

burg (changes requested):
Seconding the question of how do we test this?

Source/WebInspectorUI/UserInterface/Models/CSSStyleDeclaration.js
rcaliman-apple: > as far as i can tell it seems like this was added in [197790@main](https://commits.webkit.org/197790@main)
> 
> can you please verify that this doesnt break any of the scenarios described in that bug?

Yes, that was the source of the change, but it's not addressed in review comments why it was needed in the first place.

I don't believe the guard applies anymore for the reason stated in the code comment. As described in the commit message:

>> The validity and warnings are updated in `WI.SpreadSheetStyleProperty.prototype.updateStatus()` in response to `WI.CSSProperty.Event.Changed` triggered by `WI.CSSProperty.update()`

I did a manual test of adding/editing properties and couldn't see any obvious regression.

> also, this absolutely needs some tests

I tried a lot of ways to test this and hit a dead end. 

This is the best test I could come up with:

```diff
diff --git a/LayoutTests/inspector/css/add-css-property.html b/LayoutTests/inspector/css/add-css-property.html
index 24ef0b63f0b6..812663fb99b0 100644
--- a/LayoutTests/inspector/css/add-css-property.html
+++ b/LayoutTests/inspector/css/add-css-property.html
@@ -110,6 +110,35 @@ function test() {
         }
     });
 
+    suite.addTestCase({
+        name: "AddCSSProperty.NewCSSRuleChangingSelector",
+        description: "Ensure that a newly added blank property in an inspector stylesheet rule is not evicted by a backend style update after changing the rule's selector.",
+        async test() {
+            await Promise.all([nodeStyles.addRule("#x"), nodeStyles.awaitEvent(WI.DOMNodeStyles.Event.Refreshed)]);
+            await nodeStyles.refresh(); // <-- gets styles in sync, done by a view call: WI.GeneralStyleDetailsSidebarPanel.prototype._styleSheetAddedOrRemoved()
+
+            let style = getMatchedStyle("#x");
+
+            await nodeStyles.changeRuleSelector(style.ownerRule, "div#x");
+            await nodeStyles.changeRuleSelector(style.ownerRule, "div#x"); // <-- for the life of me, I can't figure out why only this duplicate call makes `div#x` be the rule's new selector
+
+            style = getMatchedStyle("div#x");
+
+            let newProperty = style.newBlankProperty(0);
+            style.locked = true;
+            newProperty.name = "padding";
+            newProperty.rawValue = "1em";
+
+            InspectorTest.expectThat(style.text.includes("padding: 1em;"), "Style text should contain `padding: 1em` before refresh.");
+
+            await nodeStyles.refresh();
+
+            InspectorTest.expectThat(style.text.includes("padding: 1em;"), "Style text should still contain `padding: 1em` after refresh.");
+
+            style.locked = false;
+        },
+    });
+
     function addTestCaseForCommentingOutAndAppendingProperty({subtestName, description, ruleSelector, newPropertyName, newPropertyValue})
     {
         suite.addTestCase({
```

It turns out the some models in Web Inspector **rely** on the view layer to actually get into the correct state. 

I couldn't write a model-only test. It required manual calls to `WI.DOMNodeStyles.prototype.refresh()` that normally get triggered by views in response to events to get the styles and models into the correct state (for example: `WI.GeneralStyleDetailsSidebarPanel ._styleSheetAddedOrRemoved()`), which I found defeats the purpose of a test since it doesn't reflect the workflow in the view layer. It approximates it, for now, until the view changes.

But, more perplexingly, I just can't get the test, using models only, to change the selector text and actually get it back from the models without a confusing hack. I couldn't figure it out. But, again, I suspect the view layer triggers some events / refreshes that help the models get into the correct state.

Looking at other tests in this area, there's a mix of using model objects while calling the protocol commands directly. So these tests aren't necessarily exercising the same codepaths and triggering the same side effects the actual product does. 

For example: https://github.com/WebKit/WebKit/blob/586a96e75ccf13e024033723ed68b2dd226d75d7/LayoutTests/inspector/css/modify-rule-selector.html#L26-L39

And:
https://github.com/WebKit/WebKit/blob/586a96e75ccf13e024033723ed68b2dd226d75d7/LayoutTests/inspector/css/createStyleSheet.html#L27-L39


I think we need to spend some time auditing our styles-related models to ensure they can be used standalone, in absence of a view layer, and write tests for them without directly calling the protocol commands. But that's way bigger in scope than this patch

I believe this patch is safe enough for fixing the issue after manual testing. But I can't come up with a viable _real_ test of its behavior as-is :(

burg (approved):

PR #61458, opened 2026-03-27
Relax memory ordering for CanMakeThreadSafeCheckedPtr atomic operations

cdumez:
#### bca037abbb7d0977175b2cc6207e7c183a43d967
<pre>
Relax memory ordering for CanMakeThreadSafeCheckedPtr atomic operations
<a href="https://bugs.webkit.org/show_bug.cgi?id=310854">https://bugs.webkit.org/show_bug.cgi?id=310854</a>

Reviewed by Geoffrey Garen.

CanMakeCheckedPtrBase used seq_cst (the default for std::atomic operators)
for all checked pointer count operations. This is stronger than necessary.
The checked pointer count is a diagnostic counter for zombie detection, not
a reference count that guards data access or triggers deallocation.

Use the same memory ordering strategy proven by std::shared_ptr and
Boost.Atomic for reference counting:

- Increment: memory_order_relaxed. A new CheckedPtr/CheckedRef can only be
  created from an existing reference, and communicating that reference across
  threads already provides the necessary synchronization. This is the same
  argument that makes relaxed safe for std::shared_ptr increments.

- Decrement: memory_order_release. The release ensures that all prior
  accesses through the CheckedPtr happen-before the count drops. Unlike
  std::shared_ptr, the decrement never triggers destruction, so the acquire
  side is not needed on the decrement itself. The destroying delete path
  reads the count via checkedPtrCountWithoutThreadCheck() which uses
  memory_order_acquire, forming a valid release/acquire pair.

- Read (checkedPtrCountWithoutThreadCheck): memory_order_acquire. This
  synchronizes with release decrements so the destroying delete path
  reliably sees all outstanding references when deciding whether to enter
  the zombie path.

On ARM64, this replaces full memory barriers (dmb ish) with lighter
instructions: relaxed increment becomes a plain ldadd, release decrement
becomes stlr/ldaddl, and acquire load becomes ldar.

Also introduces an AtomicLike concept to dispatch between atomic and
non-atomic storage, replacing the previous `std::is_same_v&lt;StorageType,
std::atomic&lt;uint32_t&gt;&gt;` check. This is more expressive (describes the
required interface rather than a specific type) and would work with any
atomic-like wrapper, not just `std::atomic&lt;uint32_t&gt;`.

No behavior change for the single-threaded CanMakeCheckedPtr path, which
continues to use plain integer operations with thread assertions.

* Source/WTF/wtf/CheckedRef.h:
(WTF::requires):
(WTF::CanMakeCheckedPtrBase::incrementCheckedPtrCount const):
(WTF::CanMakeCheckedPtrBase::decrementCheckedPtrCount const):
(WTF::CanMakeCheckedPtrBase::checkedPtrCountWithoutThreadCheck const):

Canonical link: <a href="https://commits.webkit.org/310266@main">https://commits.webkit.org/310266@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/20e55e8edcf4627b9beb01311b01b749981f9f02

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/152669 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/25450 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/19049 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/161413 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/106125 "Built successfully") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/25978 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/25756 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/117979 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/83588 "5 flakes 2 failures") | [loading 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/f671da49-a23b-4f2f-931e-4b8445fc0d39) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/155628 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/20203 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/137064 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/98692 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/459b4f59-9597-48cc-ae64-5b9668656741) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/19278 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/17220 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/9248 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/144680 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/128929 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/14938 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/163884 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/13474 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/7023 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/16532 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/126036 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/25248 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/21259 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/126195 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/34376 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/25250 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/136734 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/81854 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/21158 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/13513 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/184301 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/24866 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/89152 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/47047 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/24558 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/24717 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/24618 "Built successfully") | | | |

### DIFF
--- a/Source/WTF/wtf/CheckedRef.h
+++ b/Source/WTF/wtf/CheckedRef.h
@@ -350,6 +350,13 @@ enum class DefaultedOperatorEqual : bool { No, Yes };
 // DO NOT make use of this enum in new code. An object which supports CanMakeCheckedPtr must be heap allocated on its own.
 enum class CheckedPtrDeleteCheckException : bool { No, Yes };
 
+template<typename T>
+concept AtomicLike = requires(T t) {
+    t.load(std::memory_order_relaxed);
+    t.fetch_add(1, std::memory_order_relaxed);
+    t.fetch_sub(1, std::memory_order_relaxed);
+};
+
 template <typename StorageType, typename PtrCounterType, typename DeletionFlagType, CheckedPtrDeleteCheckException deleteException> class CanMakeCheckedPtrBase {
 public:
     CanMakeCheckedPtrBase() = default;
@@ -364,7 +371,13 @@ public:
     }
 
     PtrCounterType checkedPtrCount() const { return m_checkedPtrCount; }
-    void incrementCheckedPtrCount() const { ++m_checkedPtrCount; }
+    void incrementCheckedPtrCount() const
+    {
+        if constexpr (AtomicLike<StorageType>)
+            m_checkedPtrCount.fetch_add(1, std::memory_order_relaxed);
+        else
+            ++m_checkedPtrCount;
+    }
     ALWAYS_INLINE void decrementCheckedPtrCount() const
     {
         // In normal execution, a CheckedPtr always points to an object with a non-zero checkedPtrCount().
@@ -372,13 +385,16 @@ public:
         // When we check checkedPtrCountWithoutThreadCheck() here, we're checking for a scribbled object.
         if (!checkedPtrCountWithoutThreadCheck()) [[unlikely]]
             crashDueToCheckedPtrToDeadObject();
-        --m_checkedPtrCount;
+        if constexpr (AtomicLike<StorageType>)
+            m_checkedPtrCount.fetch_sub(1, std::memory_order_release);
+        else
+            --m_checkedPtrCount;
     }
 
     ALWAYS_INLINE PtrCounterType checkedPtrCountWithoutThreadCheck() const
     {
-        if constexpr (std::is_same_v<StorageType, std::atomic<uint32_t>>)
-            return m_checkedPtrCount;
+        if constexpr (AtomicLike<StorageType>)
+            return m_checkedPtrCount.load(std::memory_order_acquire);
         else
             return m_checkedPtrCount.valueWithoutThreadCheck();
     }


Source/WTF/wtf/CheckedRef.h
```diff
@@ -364,21 +371,30 @@ template <typename StorageType, typename PtrCounterType, typename DeletionFlagTy
...
     ALWAYS_INLINE void decrementCheckedPtrCount() const
     {
         // In normal execution, a CheckedPtr always points to an object with a non-zero checkedPtrCount().
         // When it detects a dangling pointer, WTF_OVERRIDE_DELETE_FOR_CHECKED_PTR scribbles an object with zeroes and then leaks it.
         // When we check checkedPtrCountWithoutThreadCheck() here, we're checking for a scribbled object.
         if (!checkedPtrCountWithoutThreadCheck()) [[unlikely]]
             crashDueToCheckedPtrToDeadObject();
-        --m_checkedPtrCount;
+        if constexpr (AtomicLike<StorageType>)
+            m_checkedPtrCount.fetch_sub(1, std::memory_order_acq_rel);
```
kmiller68: Isn't the underlying value ThreadSafeRefCounted? If so it should have an acquire fence to make sure it doesn't reach above this decrement. So I think this only needs to be `std::memory_order_release`

Source/WTF/wtf/CheckedRef.h
```diff
@@ -364,21 +371,30 @@ template <typename StorageType, typename PtrCounterType, typename DeletionFlagTy
...
+        else
+            --m_checkedPtrCount;
     }
 
     ALWAYS_INLINE PtrCounterType checkedPtrCountWithoutThreadCheck() const
     {
-        if constexpr (std::is_same_v<StorageType, std::atomic<uint32_t>>)
-            return m_checkedPtrCount;
+        if constexpr (AtomicLike<StorageType>)
+            return m_checkedPtrCount.load(std::memory_order_acquire);
```
kmiller68, line 397: I think you only need the acquire to synchronize with other memory accesses not the `m_checkedPtrCount`. At least, assuming I understand the usage/commit message correctly.

kmiller68 (commented):
Some questions

Source/WTF/wtf/CheckedRef.h
cdumez: Objects that are `CanMakeThreadCheckedPtr` and are not necessarily ThreadSafeRefCounted. `IDBBackingStore` is one such example.

kmiller68: I see. So the use case is having a weak reference to something that's intended to be ref/derefed under a lock?

Either way, I'm still not sure I follow. What's an interleaving where this acquire (or the one in `checkedPtrCountWithoutThreadCheck`) would change the semantics to ensure a crash?

cdumez: > that's intended to be ref/derefed under a lock

I do not think this is a requirement. The object doesn't even need to be ref-counted, as evidenced by `IDBBackingStore`.

`IDBBackingStore` subclasses CanMakeThreadSafeCheckedPtr, which means it holds an atomic counter. This counter represents how many CheckedPtrs are currently alive are pointing to this object. So the counter increments when a CheckedPtr is created and decrements this CheckedPtr goes out of scope.

Presumably, the reason `IDBBackingStore` needs the thread-safe version of `CanMakeCheckedPtr` is that we may be creating CheckedPtrs pointing to the same `IDBBackingStore` instance from concurrent threads.

Note that the whole point of CanMakeCheckedPtr/CanMakeThreadSafeCheckedPtr is memory safety (preventing use-after-frees). This is achieved by `IDBBackingStore` overriding `operator delete` and checking the `checkedPtrCount`, if the `checkedPtrCount` is not zero, then we know someone has a CheckedPtr pointing to this object that is about to go stale. The operator delete will thus `memset(0)` the object (which will set `checkedPtrCount` to zero among other things) and leak the object (the object is now a "zombie"). When a CheckedPtr gets destroyed, we try to decrement the pointed-to object's `checkedPtrCount`. However, if this `checkedPtrCount` is zero, we crash safely.

Sorry if I provided a lot of unnecessary details about CheckedPtrs or things you already know about. I want to make sure you have the full context when determining whether this PR is correct or not.

cdumez: I think you're right. `std::memory_order_release` should suffice here since the decrement never triggers destruction. The object's lifetime is managed independently.

Source/WTF/wtf/CheckedRef.h
cdumez: If you look at `WTF_OVERRIDE_DELETE_FOR_CHECKED_PTR_IMPL` in FastMalloc.h. After the destructor runs (line 626), it checks if any CheckedPtrs still exist (line 627). If so, it zero-fills the memory instead of freeing it (so dangling CheckedPtrs crash on access rather than use-after-free).
                                                                                                                                                               
This destructor path runs on the owning thread, after `~T()` has completed. Any other thread that held a CheckedPtr must have already decremented (with release). For this load to correctly see all those decrements, I think it needs `acquire` to pair with those releases.

Does this make sense?

Source/WTF/wtf/CheckedRef.h
cdumez: I updated the PR to use `std::memory_order_release` here and fixed the commit log message.

Source/WTF/wtf/CheckedRef.h
cdumez: I kept this as-is for now as this still looks needed to me but please let me know if you think this is wrong.

Source/WTF/wtf/CheckedRef.h
```diff
@@ -364,21 +371,30 @@ template <typename StorageType, typename PtrCounterType, typename DeletionFlagTy
...
     ALWAYS_INLINE void decrementCheckedPtrCount() const
     {
         // In normal execution, a CheckedPtr always points to an object with a non-zero checkedPtrCount().
         // When it detects a dangling pointer, WTF_OVERRIDE_DELETE_FOR_CHECKED_PTR scribbles an object with zeroes and then leaks it.
         // When we check checkedPtrCountWithoutThreadCheck() here, we're checking for a scribbled object.
         if (!checkedPtrCountWithoutThreadCheck()) [[unlikely]]
             crashDueToCheckedPtrToDeadObject();
-        --m_checkedPtrCount;
+        if constexpr (AtomicLike<StorageType>)
+            m_checkedPtrCount.fetch_sub(1, std::memory_order_release);
```
geoffreygaren, line 389: For extra optimization, you could collapse the acquire in `checkedPtrCountWithoutThreadCheck` with the release in `m_checkedPtrCount.fetch_sub`, as an `acq_rel`, which is more efficient. The `fetch_sub` can return the old value, so you could do a single `fetch_sub` with an `acq_rel`, and then check the return value.

I'm not sure you need to make this change if it's complicated, though. Would require benchmarking to justify.

geoffreygaren (approved):
There's a related issue not caused by this patch: `zeroBytes` overwrites m_checkedPtrCount but does not do its write atomically. 

And a related related issue: The compiler is technically not obligated to honor the zeroing.

I think you could fix both issues by switching from `memset` to `memset_s`. But that requires benchmarking.

Source/WTF/wtf/CheckedRef.h
cdumez: Exploring this via https://github.com/WebKit/WebKit/pull/61734. Assuming I understood the assignment correctly, the result doesn't look complicated.

PR #61528, opened 2026-03-27
[WPE][GTK] Internal error fired from WebLoaderStrategy.cpp(559) : internallyFailedLoadTimerFired

spenap:
#### 0831c81f23e301efb4c38d4fc330746ab77b4160
<pre>
[WPE][GTK] Internal error fired from WebLoaderStrategy.cpp(559) : internallyFailedLoadTimerFired
<a href="https://bugs.webkit.org/show_bug.cgi?id=276312">https://bugs.webkit.org/show_bug.cgi?id=276312</a>

It seems that under certain circumstances, NetworkProcess real-time
threads (promoted by RealtimeKit) will exceed their maximum 200ms hard
limit (rlim_max), getting killed by the kernel.

Although the code has provisions in place to demote the threads back
from real time, the fact the soft limit was set at the same value as the
hard limit meant it would never have a chance to run.

This commit adds two changes:
 - It sets the real time soft limit to 0.8 of the hard one, giving a
   chance to the SIGXCPU handler to run before the hard limit triggers a
   kill.
 - It makes the SIGXCPU signal handler async-signal-safe. The previous
   one needed to acquire a mutex and called sched_setscheduler, the
   current one uses eventfd.

The new implementation uses the handler installed by sigaction to write
to the event fd. This is bridged to the GLib main loop by g_unix_fd_add,
so that the callback can be queued and run safely, demoting all threads.

As the signal handler is installed, we query if a previous handler was
previously defined -- as done in ThreadingPOSIX --, logging it
unconditionally.

This bug can be reproduced consistently on debug builds under heavy load
(using e.g. stress-ng) without this fix. Commenting out setrlimit means
the process attempts to set an unlimited max, which RealtimeKit rejects:
in that case, the NetworkProcess runs normally and is not killed. After
this fix is applied, the demotion runs and NetworkProcess can survive.

* Source/WTF/wtf/linux/RealTimeThreads.cpp:
(WTF::RealTimeThreads::RealTimeThreads):
(WTF::RealTimeThreads::demoteThreadFromRealTime):
(WTF::sigxcpuHandler):
(WTF::RealTimeThreads::setupSignalHandler):
(WTF::RealTimeThreads::signalCallback):
(WTF::RealTimeThreads::realTimeKitMakeThreadRealTime):
* Source/WTF/wtf/linux/RealTimeThreads.h:

Canonical link: <a href="https://commits.webkit.org/310907@main">https://commits.webkit.org/310907@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/1520430f344b7195f91de9c44a593c0d7e9ce243

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/155175 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/28435 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/21594 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/163935 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/108774 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/28575 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/28283 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/120065 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/84811 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/158134 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/22318 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/139348 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/100760 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/21404 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/19456 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/11761 "Built successfully") | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/147225 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/131067 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/17188 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/166413 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/16006 "Built successfully and passed tests") | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/10832 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/18797 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/128169 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/27979 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/23492 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/128306 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/34845 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/27903 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/138981 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/84612 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/23170 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/15777 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/187060 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/27596 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/91700 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/47975 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/27174 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/27404 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/27247 "Built successfully") | | |

### DIFF
--- a/Source/WTF/wtf/linux/RealTimeThreads.cpp
+++ b/Source/WTF/wtf/linux/RealTimeThreads.cpp
@@ -29,12 +29,15 @@
 #include <sched.h>
 #include <signal.h>
 #include <string.h>
+#include <wtf/Logging.h>
 #include <wtf/MainThread.h>
 #include <wtf/NeverDestroyed.h>
 #include <wtf/SafeStrerror.h>
 
 #if USE(GLIB)
 #include <gio/gio.h>
+#include <glib-unix.h>
+#include <sys/eventfd.h>
 #include <sys/resource.h>
 #include <sys/time.h>
 #include <wtf/Seconds.h>
@@ -69,19 +72,11 @@ RealTimeThreads::RealTimeThreads()
 {
 #if USE(GLIB)
     m_discardRealTimeKitProxyTimer.setPriority(RunLoopSourcePriority::ReleaseUnusedResourcesTimer);
-#endif
 
-    callOnMainThread([] {
-        struct sigaction action;
-        sigemptyset(&action.sa_mask);
-        action.sa_sigaction = +[](int, siginfo_t*, void*) {
-            // We don't know which thread caused the limit to be reached,
-            // so we demote all real time threads to avoid SIGKILL.
-            RealTimeThreads::singleton().demoteAllThreadsFromRealTime();
-        };
-        action.sa_flags = SA_SIGINFO;
-        sigaction(SIGXCPU, &action, nullptr);
+    callOnMainThread([this] {
+        setupSignalHandler();
     });
+#endif
 }
 
 void RealTimeThreads::registerThread(Thread& thread)
@@ -140,8 +135,20 @@ void RealTimeThreads::demoteThreadFromRealTime(const Thread& thread)
 {
     ASSERT(isMainThread());
 
+    int previousPolicy = sched_getscheduler(thread.id());
+    struct sched_param prevParam = { };
+    sched_getparam(thread.id(), &prevParam);
+    if (previousPolicy == SCHED_OTHER && !prevParam.sched_priority) {
+        // Skipping thread not running in real-time.
+        return;
+    }
+
     struct sched_param param = { };
-    sched_setscheduler(thread.id(), SCHED_OTHER | SCHED_RESET_ON_FORK, &param);
+    auto ret = sched_setscheduler(thread.id(), SCHED_OTHER | SCHED_RESET_ON_FORK, &param);
+    if (ret)
+        LOG_ERROR("Demote p%d, t%d: sched_setscheduler failed: %s", getpid(), thread.id(), safeStrerror(errno).data());
+    else
+        LOG(Process, "Demote p%d, t%d: sched_setscheduler suceeded", getpid(), thread.id());
 }
 
 void RealTimeThreads::demoteAllThreadsFromRealTime()
@@ -153,6 +160,60 @@ void RealTimeThreads::demoteAllThreadsFromRealTime()
 
 #if USE(GLIB)
 static const Seconds s_dbusCallTimeout = 20_ms;
+
+static int s_eventFd = -1;
+
+static void sigxcpuHandler(int)
+{
+    uint64_t value = 1;
+    auto ret = write(s_eventFd, &value, sizeof(value));
+    UNUSED_VARIABLE(ret);
+}
+
+void RealTimeThreads::setupSignalHandler()
+{
+    ASSERT(isMainThread());
+
+    if ((s_eventFd = eventfd(0, EFD_NONBLOCK | EFD_CLOEXEC)) < 0) {
+        LOG_ERROR("Failed to create eventfd for SIGXCPU: %s", safeStrerror(errno).data());
+        return;
+    }
+
+    struct sigaction action;
+    sigemptyset(&action.sa_mask);
+    action.sa_handler = sigxcpuHandler;
+    action.sa_flags = SA_RESTART;
+
+    struct sigaction oldAction;
+    if (sigaction(SIGXCPU, &action, &oldAction))
+        LOG_ERROR("Failed to install SIGXCPU handler: %s", safeStrerror(errno).data());
+    else if (oldAction.sa_handler != SIG_DFL)
+        LOG_ERROR("Overriding existing handler for signal SIGXCPU");
+
+    g_unix_fd_add(s_eventFd, G_IO_IN, signalCallback, this);
+}
+
+gboolean RealTimeThreads::signalCallback(gint fd, GIOCondition condition, gpointer userData)
+{
+    ASSERT(s_eventFd == fd);
+
+    if (condition & (G_IO_ERR | G_IO_HUP)) {
+        LOG_ERROR("Removing event source as it has errored or disconnected. SIGXCPU is no longer handled.");
+        return G_SOURCE_REMOVE;
+    }
+
+    if (condition & G_IO_IN) {
+        uint64_t value;
+        while (read(fd, &value, sizeof(value)) == sizeof(value)) {
+        }
+
+        // We don't know which thread caused the limit to be reached,
+        // so we demote all real time threads to avoid SIGKILL.
+        static_cast<RealTimeThreads*>(userData)->demoteAllThreadsFromRealTime();
+    }
+
+    return G_SOURCE_CONTINUE;
+}
 
 #ifdef RLIMIT_RTTIME
 static int64_t realTimeKitGetProperty(GDBusProxy* proxy, const char* propertyName, GError** error)
@@ -214,7 +275,8 @@ void RealTimeThreads::realTimeKitMakeThreadRealTime(uint64_t processID, uint64_t
         }
 
         if (rl.rlim_max > static_cast<uint64_t>(rttimeMax)) {
-            rl.rlim_cur = rl.rlim_max = rttimeMax;
+            rl.rlim_cur = static_cast<uint64_t>(0.8 * rttimeMax);
+            rl.rlim_max = rttimeMax;
             setrlimit(RLIMIT_RTTIME, &rl);
         }
     }
--- a/Source/WTF/wtf/linux/RealTimeThreads.h
+++ b/Source/WTF/wtf/linux/RealTimeThreads.h
@@ -58,6 +58,8 @@ private:
     void realTimeKitMakeThreadRealTime(uint64_t processID, uint64_t threadID, uint32_t priority);
     void scheduleDiscardRealTimeKitProxy();
     void discardRealTimeKitProxyTimerFired();
+    void setupSignalHandler();
+    static gboolean signalCallback(gint, GIOCondition, gpointer);
 #endif
 
     Ref<ThreadGroup> m_threadGroup;


nikolaszimmermann (commented):
Nice finding! You should try to update the TestExpectations for glib (maybe also gtk/ and wpe/) since we had to mark quite a few tests are failures because of that.

spenap:
> I would like to understand why the network process needs RT threads.

I think there are a number of follow-up things to do:

- [X] - Check why certain threads are RT, and verify that we're promoting the right ones. See https://github.com/WebKit/WebKit/pull/62481
- [X] - Check all the D-Bus errors we get. I think the first RT promotion we request fails, because the service is not yet up (the "no such file or directory" error we see). That was not the case, but nevertheless see https://github.com/WebKit/WebKit/pull/61830
- [X] - Better error logging when running inside bwrap or outside -- there is one about being unable to map to process ids. Same as above: see https://github.com/WebKit/WebKit/pull/61830
- [X] - Check why the NetworkProcess hits the 200ms limit while other threads don't. It's a thread that does I/O: https://github.com/WebKit/WebKit/pull/62481 lowered its priority.

But I don't think we should postpone landing this PR. This is fixing two important things:
 - We thought we had a way, using the soft limit, that allowed demoting threads. This doesn't work if the soft limit and the hard limit are the same, because then there's no time to do the demotion, and the kernel kills the process
 - Even if we lower the soft limit, the main thread assertion fails, because signals are not guaranteed to arrive in specific threads -- and the demotion is handled with a signal action handler.

mcatanzaro (approved):
These changes are good.

If you're keeping track of the follow-up problems, here is one more: this signal handler is not async-signal-safe; taking a mutex in a signal handler seems especially bold. Easiest solution would be to use `g_unix_signal_add()` instead of `sigaction()`, and then not have to worry about safety. That would require a hard dependency on GLib, though. I'm not sure what else we could do here.

Source/WTF/wtf/linux/RealTimeThreads.cpp
```diff
@@ -136,8 +136,6 @@ void RealTimeThreads::promoteThreadToRealTime(const Thread& thread)
 
 void RealTimeThreads::demoteThreadFromRealTime(const Thread& thread)
 {
-    ASSERT(isMainThread());
```
mcatanzaro, line 141: I was thinking it's OK to remove this assertion because the function doesn't seem to need it.

I see we have parallel assertions in `RealTimeThreads::setEnabled` and `RealTimeThreads::promoteThreadToRealTime`, though. I wonder if I am missing some reason the assert needs to exist. But I think we need to remove it. An alternative to removing the assert would be to use `callOnMainThread` in the signal handler, except we should not do that because it's not async-signal-safe.

Source/WTF/wtf/linux/RealTimeThreads.cpp
```diff
@@ -56,6 +59,14 @@ RealTimeThreads& RealTimeThreads::singleton()
     static LazyNeverDestroyed<RealTimeThreads> realTimeThreads;
     static std::once_flag onceFlag;
     std::call_once(onceFlag, [&] {
+#if USE(GLIB)
+        // Block SIGXCPU before any threads are spawned so all threads
+        // inherit the mask. This ensures signalfd is the sole consumer.
```
mcatanzaro: Potentially difficult problem: this class does not control when the threads are spawned. It only controls when they become realtime. It might need to go in `Thread::initializePlatformThreading`?

The other potentially-good place, `AuxiliaryProcess::platformInitialize`, does not work because we need this in the UI process too.

mcatanzaro: ~~Oh, but I see: it's fine because if the threads are not yet real time, then they'll never receive SIGXCPU. Of course. Oops.~~

mcatanzaro: Oh no, I was right the first time. This really is a problem.

Programming is confusing.

Source/WTF/wtf/linux/RealTimeThreads.cpp
```diff
@@ -154,6 +156,35 @@ void RealTimeThreads::demoteAllThreadsFromRealTime()
...
+    sigemptyset(&mask);
+    sigaddset(&mask, SIGXCPU);
+
+    m_signalFd = signalfd(-1, &mask, SFD_NONBLOCK | SFD_CLOEXEC);
+    if (m_signalFd < 0) {
+        LOG_ERROR("Failed to create signalfd for SIGXCPU: %s", safeStrerror(errno).data());
+        return;
+    }
+
+    g_unix_fd_add(m_signalFd, G_IO_IN, signalfdCallback, this);
```
mcatanzaro: If you're going to depend on GLib, then there's no need to mess with signalfd or blocking SIGXCPU at all. Just use `g_unix_signal_add()` and move on with life. Couldn't be easier. I had thought you wanted to use signalfd to make the _non-GLib_ codepath work.

Source/WTF/wtf/linux/RealTimeThreads.cpp
```diff
@@ -214,7 +245,8 @@ void RealTimeThreads::realTimeKitMakeThreadRealTime(uint64_t processID, uint64_t
         }
 
         if (rl.rlim_max > static_cast<uint64_t>(rttimeMax)) {
-            rl.rlim_cur = rl.rlim_max = rttimeMax;
+            rl.rlim_cur = static_cast<uint64_t>(0.8 * rttimeMax);
```
mcatanzaro, line 278: Honestly I have no clue what good limits are, so I guess this is fine?

Your original choice `rttimeMax / 4` seemed pretty nice to me. Reacting in time to prevent the kill is all luck-based, so might as well give the best chance possible of doing that? `rttimeMax` is already very high at 200 ms; since real time threads preempt everything else, I think they shouldn't get anywhere near 50 ms.

Source/WTF/wtf/linux/RealTimeThreads.cpp
spenap: I understand it the same way you did in your second comment: if a thread _never_ needs to become real time, then it never goes through this code path, and so it doesn't have to worry about real time limits, or a handler when the limits are hit.

Source/WTF/wtf/linux/RealTimeThreads.cpp
spenap: This is a good point.

Source/WTF/wtf/linux/RealTimeThreads.cpp
spenap: This part is a mess. 

Remember how all this started, with the Network Process getting killed? That was my Lenovo ThinkStation, with 64 threads running under debug, while stress-ng was artificially constraining things. In that one, I really thought 1/4 was the right balance. But then, as I started reviewing each segment of the patch, I realised that as long as we didn't set soft = hard, the signal handler was consistently invoked. 

I tried 1/4, 1/2, 3/4 and 0.8. Didn't keep trying, because I could see the thread getting demoted every time. 

As I used the laptop in the evenings to verify things, what I see here is different. The Network Process is not demoted here: a WebProcess thread is. And I know because of the new logs I added: when I run under bwrap, I can see the "failed to map using portal" message, for such process and thread. Then (or before!) I see the same thread being demoted -- which is odd. How can you demote something you haven't been able to promote?

If I run with `WEBKIT_DISABLE_SANDBOX_THIS_IS_DANGEROUS=1`, then I don't see the mapping log, but I still see the demotion. 

So... my conclusions are
1. as long as soft != hard, we're good, and 1/4 vs 0.8 doesn't seem to matter. Happy to set whatever.
2. we need more logs. After this lands, I want to have `g_child_watch_add` setup so we can tell why a certain process has finished (i.e. exited fine, signaled, etc.)
3. we might need to modify a bit the Thread public API so we can print the thread name instead / besides the thread id, as that can surely help, too

(Once we can tell which threads consistently get demoted, we can search what they do and understand whether they are slow or blocking, and address that)

Source/WTF/wtf/linux/RealTimeThreads.cpp
spenap: (If we use g_unix_signal_add, this point becomes irrelevant)

Source/WTF/wtf/linux/RealTimeThreads.cpp
spenap: Things I discovered after a long compilation:

https://github.com/GNOME/glib/blob/9102d86b7506710e7c03c2475829be584eee33ae/glib/glib-unix.c#L230

It uses https://docs.gtk.org/glib-unix/func.signal_source_new.html, which only supports a few signals, not including SIGXCPU. So we can't use this.

Source/WTF/wtf/linux/RealTimeThreads.cpp
mcatanzaro: OK then!

Source/WTF/wtf/linux/RealTimeThreads.cpp
mcatanzaro: But if the thread was created before this is called, then it won't have SIGXCPU blocked.

Source/WTF/wtf/linux/RealTimeThreads.cpp
```diff
@@ -69,19 +80,11 @@ RealTimeThreads::RealTimeThreads()
...
 #if USE(GLIB)
     m_discardRealTimeKitProxyTimer.setPriority(RunLoopSourcePriority::ReleaseUnusedResourcesTimer);
-#endif
 
-    callOnMainThread([] {
-        struct sigaction action;
-        sigemptyset(&action.sa_mask);
-        action.sa_sigaction = +[](int, siginfo_t*, void*) {
-            // We don't know which thread caused the limit to be reached,
-            // so we demote all real time threads to avoid SIGKILL.
```
mcatanzaro, line 79: I think this little comment is useful, so I would copy it into `signalfdCallback`.

Source/WTF/wtf/linux/RealTimeThreads.cpp
mcatanzaro: I'm sorry. :(

What's especially strange is: I don't know why this limitation exists. The actual implementation looks like it supports every imaginable signal, except for the precondition check that fails if you try to use one of the unsupported signals. Oh well.

Source/WTF/wtf/linux/RealTimeThreads.cpp
spenap: Right, will address

Source/WTF/wtf/linux/RealTimeThreads.cpp
spenap: My bad here. It printed about a non-demotion, the log signals the intent, not the fact. Will check to avoid misleading logs.

Source/WTF/wtf/linux/RealTimeThreads.h
```diff
@@ -58,13 +58,16 @@ class RealTimeThreads : public CanMakeWeakPtr<RealTimeThreads> {
...
+    void setupSignalfd();
+    static gboolean signalfdCallback(gint fd, GIOCondition, gpointer userData);
 #endif
 
     Ref<ThreadGroup> m_threadGroup;
     bool m_enabled { true };
 #if USE(GLIB)
     std::optional<GRefPtr<GDBusProxy>> m_realTimeKitProxy;
     RunLoop::Timer m_discardRealTimeKitProxyTimer;
+    int m_signalFd { -1 };
```
aperezdc: Any reason to not use `UnixFileDescriptor`? That way the file descriptor will be automatically closed when the `RealTimeThreads` instance is destroyed, instead of leaking the file descriptor... Although this is not a big deal, because destruction will happen when terminating the process, and the kernel will anyway reclaim the file descriptor 🤔

aperezdc (approved):
LGTM, I left a minor comment but it's not that important and shouldn't be a blocker for landing.

Source/WTF/wtf/linux/RealTimeThreads.h
spenap: No, no reason, and this is a good point!!

Source/WTF/wtf/linux/RealTimeThreads.cpp
```diff
@@ -56,6 +59,16 @@ RealTimeThreads& RealTimeThreads::singleton()
...
+#if USE(GLIB)
+        // Block SIGXCPU in the calling thread. In auxiliary processes
+        // the main thread is already masked via AuxiliaryProcessMainCommon,
+        // but in the UI process this is needed to mask the main thread
+        // before any thread is promoted to real-time.
+        sigset_t mask;
+        sigemptyset(&mask);
+        sigaddset(&mask, SIGXCPU);
+        pthread_sigmask(SIG_BLOCK, &mask, nullptr);
+#endif
```
mcatanzaro: Wait, why are you blocking the signal on the main thread? That's the thread where you _want_ to receive the signal, right?

Regardless, you previously noticed that the signal could be delivered to _any_ thread, not only the real time thread. For this to actually work, we need to mask before any threads are created, not only before any real time threads are created. And since we have no control of what the UI process does before initializing WebKit, I'm afraid the only solution I can think of for properly blocking the signal is to use a library constructor.

It would probably be better to give up and just accept that the signal might be delivered to any thread. Which might be OK, because I see signalfd(2) says:

```
   Thread semantics
       The semantics of signalfd file descriptors in a multithreaded program mirror the standard semantics for signals.   In  other  words,
       when a thread reads from a signalfd file descriptor, it will read the signals that are directed to the thread itself and the signals
       that  are  directed to the process (i.e., the entire thread group).  (A thread will not be able to read signals that are directed to
       other threads in the process.
```

So the question is: is SIGXCPU directed to the entire process? Presumably? If so, then there's no problem: you're guaranteed that it's going to be handled on the thread running the default main context, regardless of which thread actually receives the signal (unless the application interferes by registering its own SIGXCPU handler that might compete with your signalfd).

mcatanzaro (changes requested):

Source/WTF/wtf/linux/RealTimeThreads.cpp
spenap: > Wait, why are you blocking the signal on the main thread? That's the thread where you want to receive the signal, right?

That blocks the default signal delivery mechanism which, without a signal handler, would be to terminate the program: we want to avoid that behaviour.

We want to block `SIGXCPU` for all processes, before any threads are spawned. The strategy I followed was:

1. For auxiliary processes, we do it in the AuxiliaryMainCommon initialisation bits. That covers Web, Network, GPU process -- and anything that might appear in the future
2. For the UI process, we don't have an AuxiliaryMainCommon we can rely on. So in that case, what I did was to patch ThreadingPOSIX. You will notice that any threads created there (there are WTF and non-WTF) will go through an initialisation where we block `SIGXCPU`. 

The above covers non-main threads in the UI process. 

As for the UI process main thread, we land on the code you quoted: we block `SIGXCPU`, and we setup `signalfd`.

mcatanzaro: I think  you've done the best you can without resorting to a library constructor, but `Thread::initializePlatformThreading` is going to be called when initializing WebKit, and applications will undoubtedly create threads before that point. So I think it really needs to be done in a library constructor.

The good news is you can put the library constructor in RealTimeThreads.cpp, so you only need it in one place rather than three.

Source/WTF/wtf/posix/ThreadingPOSIX.cpp
```diff
@@ -176,6 +179,18 @@ void Thread::signalHandlerSuspendResume(int, siginfo_t*, void* ucontext)
...
 
+#if USE(GLIB)
+static void blockSIGXCPU()
+{
+    // Block SIGXCPU so that signalfd in RealTimeThreads is the sole
+    // consumer when real-time threads exceed their CPU time limit.
+    sigset_t mask;
+    sigemptyset(&mask);
+    sigaddset(&mask, SIGXCPU);
+    pthread_sigmask(SIG_BLOCK, &mask, nullptr);
```
mcatanzaro: Hypothetically, let's say we decide to get rid of RealTimeThreads in the future. We'll probably forget about this and fail to remove it, right? If we do the same work in a function of RealTimeThreads, and just call it here, then we slightly reduce code duplication, and won't forget to remove it. Same comment applies to AuxiliaryProcessMain.cpp.

(But note this comment is not actually relevant if you follow my suggestion to use a library constructor.)

Source/WTF/wtf/linux/RealTimeThreads.cpp
mcatanzaro: Simon has pointed out that a library constructor is not good enough, because it may be too late if WebKit is dlopened.

I think that, in practice, applications can and will create threads before dlopening WebKit. Especially likely if the application is written in Python or JavaScript or similar interpreted languages.

spenap: Reworked to use eventfd

Source/WTF/wtf/linux/RealTimeThreads.cpp
```diff
@@ -154,6 +161,54 @@ void RealTimeThreads::demoteAllThreadsFromRealTime()
 #if USE(GLIB)
 static const Seconds s_dbusCallTimeout = 20_ms;
 
+static int s_eventFd = -1;
+
+static void sigxcpuHandler(int)
+{
+    uint64_t value = 1;
+    auto ret = write(s_eventFd, &value, sizeof(value));
+    UNUSED_VARIABLE(ret);
```
aperezdc, line 170: The eventfd is non-blocking so writing may fail with `EINVAL`, so this should retry until success:


```suggestion
    do {
        const uint64_t value = 1;
        auto ret = write(s_eventFd, &value, sizeof(value));
    } while (ret < 0 && errno == EAGAIN);
```

Source/WTF/wtf/linux/RealTimeThreads.cpp
```diff
@@ -154,6 +161,54 @@ void RealTimeThreads::demoteAllThreadsFromRealTime()
...
+{
+    ASSERT(isMainThread());
+
+    m_signalFd = UnixFileDescriptor(eventfd(0, EFD_NONBLOCK | EFD_CLOEXEC), UnixFileDescriptor::Adopt);
+    if (m_signalFd.value() < 0) {
+        LOG_ERROR("Failed to create eventfd for SIGXCPU: %s", safeStrerror(errno).data());
+        return;
+    }
+
+    s_eventFd = m_signalFd.value();
```
aperezdc: If we are going to need a global, let's assign to it directly to avoid it getting `s_eventFd` and `m_signalFd` out of sync:


```suggestion
    if ((m_eventFd = eventfd(0, EFD_NONBLOCK | EFD_CLOEXEC) < 0) {
        LOG_ERROR("Failed to create eventfd for SIGXCPU: %s", safeStrerror(errno).data());
        return;
    }
```

Source/WTF/wtf/linux/RealTimeThreads.cpp
```diff
@@ -154,6 +161,54 @@ void RealTimeThreads::demoteAllThreadsFromRealTime()
...
+    uint64_t value;
+    while (read(fd, &value, sizeof(value)) == sizeof(value)) {
+    }
+
+    // We don't know which thread caused the limit to be reached,
+    // so we demote all real time threads to avoid SIGKILL.
+    static_cast<RealTimeThreads*>(userData)->demoteAllThreadsFromRealTime();
+
+    return G_SOURCE_CONTINUE;
+}
```
aperezdc: Re-reading this I remembered that it is good practice to handle `G_IO_ERR` and specially `G_IO_HUP` to detect when the file descriptor has errors or becomes invalid:


```suggestion
gboolean RealTimeThreads::signalCallback(gint fd, GIOCondition condition, gpointer userData)
{
    ASSERT(s_eventFd == fd);
    
    if (condition & (G_IO_ERR | G_IO_HUP))
        return G_SOURCE_REMOVE;

    if (condition & G_IO_IN) {
        uint64_t value;
        while (read(fd, &value, sizeof(value)) == sizeof(value)) /* no-op */;

        // We don't know which thread caused the limit to be reached,
        // so we demote all real time threads to avoid SIGKILL.
        static_cast<RealTimeThreads*>(userData)->demoteAllThreadsFromRealTime();
    }
    return G_SOURCE_CONTINUE;
}
```

Source/WTF/wtf/linux/RealTimeThreads.cpp
aperezdc: Scratch that. This is not a good idea, the blocking will only happen...

>  If the addition would  cause  the  counter's  value to exceed the maximum, then the write(2) either blocks until a read(2) is performed on the file descriptor, or fails with the error EAGAIN if the file descriptor has been made nonblocking.

The maximum value being `UINT64_MAX-1`, this is unlikely to happen, as noted in the [eventfd(2)](https://www.man7.org/linux/man-pages/man2/eventfd.2.html) manual page as well (emphasis mine):

> As noted above, [write(2)](https://www.man7.org/linux/man-pages/man2/write.2.html) can never overflow the counter.  However an overflow can occur if 2^64 eventfd "signal posts" were performed by the KAIO subsystem (**theoretically possible, but practically unlikely**)

OTOH, if we retry the `write()`, we may be stuck forever because this being the signal handler for `SIGXCPU` means that we are technically in the same thread that might to the corresponding `read()` *but* for that we have to return from the signal handler&mdash;which won't happen if looing on `EAGAIN`!

aperezdc (approved):

spenap:
@mcatanzaro, I have addressed your comments and Adrian has reviewed positively, so I am going to land. If you have any follow-up ideas, let me know! :pray: Thanks a lot (to you and all the others) for steering this with all your comments!

aperezdc:
Backported into `webkitglib/2.52` as commit 1815b508d98d7c905e3368477afdd1008e9f36f4

Source/WTF/wtf/linux/RealTimeThreads.cpp
```diff
@@ -154,6 +161,60 @@ void RealTimeThreads::demoteAllThreadsFromRealTime()
...
+
+    if (condition & (G_IO_ERR | G_IO_HUP)) {
+        LOG_ERROR("Removing event source as it has errored or disconnected. SIGXCPU is no longer handled.");
+        return G_SOURCE_REMOVE;
+    }
+
+    if (condition & G_IO_IN) {
+        uint64_t value;
+        while (read(fd, &value, sizeof(value)) == sizeof(value)) {
+        }
```
mcatanzaro, line 208: Nit: you could have written this as:

```suggestion
        while (read(fd, &value, sizeof(value)) == sizeof(value));
```

which I think is slightly nicer. Probably not worth the effort to change it now that it has already landed, though.

Source/WTF/wtf/linux/RealTimeThreads.cpp
mcatanzaro: Regardless, we do still need to loop in order to check `EINTR`, though, right? Don't we always need to do that?

mcatanzaro (commented):
Very nice. Good job. This was hard, but I think this works!

I only see one fixable problem.

mcatanzaro:
> I only see one fixable problem.

(Regarding `EINTR`.)

Source/WTF/wtf/linux/RealTimeThreads.cpp
spenap: I think I tried it and check-webkit-style complained, or maybe it was something slightly similar. I tried plenty of things :-D

spenap:
> > I only see one fixable problem.
> 
> (Regarding `EINTR`.)

I am not sure. I spoke with Adrian about this, and realised that the time I spent thinking about this error case was not reflected in the patch at all -- not even with a comment in the code or in the PR.

I didn't want to loop because this is a signal handler, I wanted it done ASAP. If EINTR is received, it means a signal was sent while we were handling one, which sounds like (low probability) trouble.

But on the other hand, if we look at Source/WebKit/Platform/IPC/unix/IPCSemaphoreUnix.cpp, we see the same pattern and they do handle `EINTR`, so I guess there's no harm in doing it ourselves. I'll do a follow-up.

```cpp
void Semaphore::signal()
{
#if OS(LINUX)
    ASSERT_WITH_MESSAGE(m_fd.value() >= 0, "Signalling on an invalid semaphore object");

    // Matching waitImpl() and EFD_SEMAPHORE semantics, increment the semaphore value by 1.
    uint64_t value = 1;
    while (true) {
        int ret = write(m_fd.value(), &value, sizeof(uint64_t));
        if (ret != -1 || errno != EINTR) [[likely]]
            break;
    }
#endif
}
```

spenap:
> I would like to understand why the network process needs RT threads.

I've submitted https://github.com/WebKit/WebKit/pull/62481 for this

spenap:
> > I only see one fixable problem.
> 
> (Regarding `EINTR`.)

I have submitted https://github.com/WebKit/WebKit/pull/62622

PR #61635, opened 2026-03-30
[Swift in WebKit] Convert the TestWebKitAPI executable entry points to Swift

rr-codes:
#### 2aaca4d468e9079c7c96ff074942aa7c2d105546
<pre>
[Swift in WebKit] Convert the TestWebKitAPI executable entry points to Swift
<a href="https://bugs.webkit.org/show_bug.cgi?id=311041">https://bugs.webkit.org/show_bug.cgi?id=311041</a>
<a href="https://rdar.apple.com/173648959">rdar://173648959</a>

Reviewed by Elliott Williams.

To support using Swift Testing from TestWebKitAPI, the executable needs to have
a Swift entrypoint to properly support async/await and to call the testing entry point.

To facilitate this, convert the existing entry point logic to Swift, and unify iOS and macOS
logic into one place.

* Tools/TestWebKitAPI/Configurations/Base.xcconfig:
* Tools/TestWebKitAPI/Configurations/TestIPC.xcconfig:
* Tools/TestWebKitAPI/Configurations/TestWGSL.xcconfig:
* Tools/TestWebKitAPI/Configurations/TestWTF.xcconfig:
* Tools/TestWebKitAPI/Configurations/TestWebKitAPI.xcconfig:
* Tools/TestWebKitAPI/Configurations/TestWebKitAPILibrary.xcconfig:
* Tools/TestWebKitAPI/TestWebKitAPI-Bridging-Header.h: Added.
* Tools/TestWebKitAPI/TestWebKitAPI.swift: Added.
(TestWebKitAPI.forceSiteIsolationForTesting):
(TestWebKitAPI.handleArguments(_:)):
(TestWebKitAPI.run):
(TestWebKitAPI.main):
* Tools/TestWebKitAPI/TestWebKitAPISupport.h: Added.
* Tools/TestWebKitAPI/TestWebKitAPISupport.mm: Renamed from Tools/TestWebKitAPI/ios/mainIOS.mm.
(TestWebKitAPIEnableAllSDKAlignedBehaviors):
(TestWebKitAPIRunTests):
* Tools/TestWebKitAPI/ios/UIKitMacHelperSPI.h:
* Tools/TestWebKitAPI/mac/mainMac.mm:

Canonical link: <a href="https://commits.webkit.org/310524@main">https://commits.webkit.org/310524@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/2aaede58298f3cbd5852b2915fb051950cb29f92

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/154113 "3 style errors") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/27370 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/20528 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/162865 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/107581 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/155986 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/27502 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/27221 "Built successfully") | [  ~~🧪 wpe-wk2~~](https://ews-build.webkit.org/#/builders/34/builds/119189 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/107581 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/60202231-fdc4-4045-8d5a-52e47909b535) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/157072 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/162/builds/21439 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/138398 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/99889 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/20525 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/18523 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/10698 "Built successfully") | | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/130187 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-wk1~~](https://ews-build.webkit.org/#/builders/167/builds/16242 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/165338 "Built successfully") | | 
| | | [  ~~🧪 mac-wk2~~](https://ews-build.webkit.org/#/builders/169/builds/17851 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/1/builds/127280 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/26916 "Built successfully") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/22567 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/127431 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/34568 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/26839 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/138035 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/83431 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/22300 "Passed tests") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/14828 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/26531 "Built successfully") | | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/26111 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/26341 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/26183 "Built successfully") | | | |


emw-apple (changes requested):
High level feedback (that I already shared with you, but including for posterity here): Let's please not add a package dependency on ArgumentParser. It will add a package resolution step to everyone's build (which requires fetching code from public github, which likely won't work on all Apple internal build systems). And it will slow everyone else down. ArgumentParser is great, but not at that cost.

I recommend handwriting an argument parser (keep in mind that TestWebKitAPI isn't really intended to be run interactively, that's what `run-api-tests` is for, so this parser can be quite simple), or being old school and using getopt/getopt_long.

Tools/TestWebKitAPI/TestWebKitAPISupport.mm
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2013 Apple Inc. All rights reserved.
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
```
emw-apple: nit: I think this should be `2013-2026`

Tools/TestWebKitAPI/TestWebKitAPISupport.mm
emw-apple, line 1: Did you consider enabling Swift-C++ interop and calling these directly from `TestWebKitAPI.run()`? Seems like you're so close to doing it all with no need for top-level c++ code.

rr-codes: I did; but we don’t have support for Cxx interop in OS 26.2 compilers so I could not unfortunately

emw-apple: Ack. We will be able to drop support for 26.2 toolchains soon, so you might be able to do it in the near future!

Tools/TestWebKitAPI/Tests/mac/WKWebViewMacEditingTests.mm
```diff
@@ -265,8 +265,13 @@ - (NSRange)range
     [webView removeFromSuperview];
 
     __unsafe_unretained id <NSTextInputClient_Async> inputClient = (id <NSTextInputClient_Async>)webView.get();
+    __weak id <NSTextInputClient_Async> weakInputClient = (id <NSTextInputClient_Async>)webView.get();
     [inputClient hasMarkedTextWithCompletionHandler:^(BOOL) {
+        if (!weakInputClient)
+            return;
         [inputClient selectedRangeWithCompletionHandler:^(NSRange) {
+            if (!weakInputClient)
+                return;
```
emw-apple: Curious what this is about. Why does the change in entrypoint affect when this web view is deallocated?

Tools/TestWebKitAPI/TestWebKitAPI.swift
```diff
@@ -0,0 +1,131 @@
...
+            // These defaults are not propagated manually but are only consulted in the UI process.
+            switch argument {
+            case "--remote-layer-tree":
+                argumentDefaults["WebKit2UseRemoteLayerTreeDrawingArea"] = true
+            case "--no-remote-layer-tree":
+                argumentDefaults["WebKit2UseRemoteLayerTreeDrawingArea"] = false
+            case "--use-gpu-process":
+                argumentDefaults["WebKit2GPUProcessForDOMRendering"] = true
+            case "--no-use-gpu-process":
+                argumentDefaults["WebKit2GPUProcessForDOMRendering"] = false
```
emw-apple, line 59: I know you're porting existing logic, but: Since all these flags do is set defaults in the argument domain, is it really worth parsing them as arguments at all? A caller could run `./TestWebKitAPI -WebKit2UseRemoteLayerTreeDrawingArea=YES` instead of `./TestWebKitAPI --remote-layer-tree` and it would set the same defaults key.

Tools/TestWebKitAPI/TestWebKitAPI.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    @MainActor
+    private mutating func run() async {
+        #if targetEnvironment(macCatalyst)
+        UINSApplicationInstantiate()
+        #endif
+
+        UserDefaults.standard.removePersistentDomain(forName: "TestWebKitAPI")
+
+        // Set a user default for TestWebKitAPI to bypass all linked-on-or-after checks in WebKit
```
emw-apple: Does this comment need to move down to L97?

Tools/TestWebKitAPI/TestWebKitAPI.swift
```diff
@@ -0,0 +1,131 @@
...
+                    "ForceLegacyHostingRemoteViewControllerForService": "com.apple.ScreenTime.ScreenTimeWebExtension"
+                ]
+            )
+        #endif // !os(macOS) || targetEnvironment(macCatalyst)
+
+        TestWebKitAPIEnableAllSDKAlignedBehaviors()
+
+        #if os(macOS) && !targetEnvironment(macCatalyst)
+        _ = NSApplication.shared
+        #endif
```
emw-apple, line 100: What's going on here?

Tools/TestWebKitAPI/TestWebKitAPI.swift
```diff
@@ -0,0 +1,131 @@
...
+        // queue normally.
+        let passed = await withCheckedContinuation { continuation in
+            RunLoop.main.perform(inModes: [.default]) {
+                let result = unsafe TestWebKitAPIRunTests(CommandLine.argc, CommandLine.unsafeArgv)
+                CFRunLoopStop(RunLoop.main.getCFRunLoop())
+                continuation.resume(returning: result)
+            }
+
+            CFRunLoopWakeUp(RunLoop.main.getCFRunLoop())
+        }
```
emw-apple: What does using the `await` accomplish here? We're not calling any async code.

It seems like we would get the same behavior by entering the run loop directly, without hopping through  the concurrency runtime.

emw-apple (commented):
> This uses the de-facto standard SPM ArgumentParser package for type-safe argument parsing.

This is outdated now.

Tools/TestWebKitAPI/Tests/mac/WKWebViewMacEditingTests.mm
rr-codes: Because this test was incorrect/broken to begin with, and was relying on racy behavior of the test runner itself exiting before there was a chance for it to actually crash, and the changes to the way the entry point handles the run loop now reveals this longstanding issue. (I discussed this at length with the test author and we arrived at this conclusion)

Tools/TestWebKitAPI/TestWebKitAPI.swift
rr-codes: I decided to leave it just to reduce the scope of changes in this PR

Tools/TestWebKitAPI/TestWebKitAPI.swift
rr-codes: this is existing logic ported over. NSApplication.shared has critical side effects when called

Tools/TestWebKitAPI/Tests/mac/WKWebViewMacEditingTests.mm
pxlcoder: Can we make this change separately?

Tools/TestWebKitAPI/TestWebKitAPI.swift
rr-codes: `withCheckedContinuation` is an `async` function, which finishes once its `resume` method gets called. My comment above this tries to explain why it’s needed. Without this, all tests that spin the run loop hang.

Tools/TestWebKitAPI/TestWebKitAPI.swift
rr-codes: this is directly ported over from the exiting code

Tools/TestWebKitAPI/Tests/mac/WKWebViewMacEditingTests.mm
emw-apple: Ack. As a nit, maybe you could just use the weak pointer (replacing the unretained one)? Since all we do with it is send messages, it will be a no-op when the pointer gets cleared.

Tools/TestWebKitAPI/TestWebKitAPI.swift
```diff
@@ -0,0 +1,131 @@
...
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+// SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+// INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+// CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+// ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+// THE POSSIBILITY OF SUCH DAMAGE.
+
+import Foundation
+
+#if os(macOS) && !targetEnvironment(macCatalyst)
```
emw-apple, line 26: Can we use the WTF platform conditions to avoid this awkward Mac-but-not-catalyst dance? Here and below:

```suggestion
#if PLATFORM_MAC
```

rr-codes: I don’t think the Test{WebKitAPI, WTF, …} binaries are configured to support the WTF conditions iirc, so that’s why I didn’t (I also don’t mind the normal conditions instead of the WTF ones so I didn’t try too hard admittedly)

Tools/TestWebKitAPI/Tests/mac/WKWebViewMacEditingTests.mm
pxlcoder: The `__unsafe_unretained` is intentional. This is a crash test, if weak were used the crash would never occur as the pointer would be zeroed out.

rr-codes: moved this to its own PR @pxlcoder  https://github.com/WebKit/WebKit/pull/61930

pxlcoder: After thinking about this more, and confirming with Wenson, the test as written already stops testing what it was meant to...

Tools/TestWebKitAPI/TestWebKitAPI.swift
emw-apple: It's set up in TestWebKitAPIBundle and TestWebKitAPILibrary:
https://github.com/WebKit/WebKit/blob/93f532da766625d2aa5fe6804085bbc89bd8e8aa/Tools/TestWebKitAPI/Configurations/TestWebKitAPIBundle.xcconfig#L52
https://github.com/WebKit/WebKit/blob/93f532da766625d2aa5fe6804085bbc89bd8e8aa/Tools/TestWebKitAPI/Configurations/TestWebKitAPILibrary.xcconfig#L43

Perhaps we could move these up to Base?

Tools/TestWebKitAPI/TestWebKitAPI.swift
emw-apple: I'm just very confused what it is doing. The statement below it doesn't set any defaults at all. The defaults we set further down are not related to linked-on-or-after checks. The call to `TestWebKitAPIEnableAllSDKAlignedBehaviors` _is_ related to linked-on-or-after checks, but it doesn't set any user defaults.

Unless I'm missing something, maybe you can just remove the comment. I don't think it is contributing to making the code easier to understand.

rr-codes: agreed; removed

Tools/TestWebKitAPI/TestWebKitAPI.swift
rr-codes: somehow removing all the weird run loop stuff entirely just Made it work, weird but good!

Tools/TestWebKitAPI/TestWebKitAPI.swift
rr-codes: It looks like this is not enough; we would need to add these to TestWTF, TestIPC, and TestWGSL, and also add the generation build phase to all of these too. Maybe we could consider doing so in a different PR to reduce the scope and size of this one?

emw-apple: Sure, do it as a follow-up. It doesn't need to be duplicated per-target, we just need to move the existing script phase out of TestWebKitAPILibrary into a target that _all_ the test targets can depend on. I'd recommend a Derived Sources aggregate target like our other projects have.

Tools/TestWebKitAPI/TestWebKitAPI.swift
```diff
@@ -0,0 +1,111 @@
...
+            case "--site-isolation":
+                Self.forceSiteIsolationForTesting()
+            default:
+                break
+            }
+        }
+    }
+
+    @MainActor
+    private mutating func run() async {
```
emw-apple: Out of curiosity, why is this `mutating`? The TestWebKitAPI structure has no data

emw-apple (approved):

PR #61704, opened 2026-03-30
AX: ASSERTION FAILED on AXIsolatedTree.cpp:249 when running http/tests/site-isolation/accessibility/client/simple-iframe.html

minorninth:
#### 8c8159f34720c0fbf31c175d57b843a38a33f1d7
<pre>
AX: ASSERTION FAILED on AXIsolatedTree.cpp:249 when running http/tests/site-isolation/accessibility/client/simple-iframe.html
<a href="https://rdar.apple.com/173698630">rdar://173698630</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=311114">https://bugs.webkit.org/show_bug.cgi?id=311114</a>

Reviewed by Tyler Wilcock.

I intermittently hit this assertion failure when running
http/tests/site-isolation/accessibility/client/simple-iframe.html,
which tests accessibility with site isolation. By fiddling with the
test timing I was able to reproduce it reliably.

ASSERTION FAILED: AX: After applying pending root node, 15 reachable nodes but 17 are in the node map
reachableNodes.size() == m_readerThreadNodeMap.size()
/Volumes/Code/safari/OpenSource/Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp(249) : void WebCore::AXIsolatedTree::applyPendingRootNodeLocked()
1   WebCore::AXIsolatedTree::applyPendingRootNodeLocked()
2   WebCore::AXIsolatedTree::applyPendingChangesLocked()
3   WebCore::AXIsolatedTree::applyPendingChanges()

The bug happens when the in-process FrameHost (AccessibilityScrollView)
is replaced with the out-of-process FrameHost, and then the FrameHost ends up in
both m_pendingSubtreeRemovals and m_pendingAppends in applyPendingChangesLocked.

We never hit this before because it doesn&apos;t happen with DOM nodes or
LayoutObjects - if the old object was removed from the AX cache, then
it won&apos;t end up in m_pendingAppends. In this case both the old Widget
and the new Widget exist, so the assertion is telling us we&apos;re leaking
the old isolated objects.

With this fix, and with also implementing the allowAXAuthenticationForTesting
check in the UI process to allow client accessibility APIs to succeed for any
process, http/tests/site-isolation/accessibility/client/simple-iframe.html can be enabled.

* LayoutTests/http/tests/site-isolation/accessibility/client/simple-iframe-expected.txt:
* LayoutTests/http/tests/site-isolation/accessibility/client/simple-iframe.html:
(WebCore::AXIsolatedTree::queueRemovals):
(WebCore::AXIsolatedTree::queueRemovalsLocked):
(WebCore::AXIsolatedTree::queueAppendsAndRemovals):
(WebCore::AXIsolatedTree::updateChildren):
(WebCore::AXIsolatedTree::removeNode):
(WebCore::AXIsolatedTree::removeStaleAppends):
(WebCore::AXIsolatedTree::applyPendingChangesLocked):
* Source/WebKit/UIProcess/API/C/WKContext.cpp:
(WKContextSetAllowAXAuthenticationForTesting):
* Source/WebKit/UIProcess/API/C/WKContextPrivate.h:
* Source/WebKit/UIProcess/Cocoa/WebProcessProxyCocoa.mm:
(WebKit::WebProcessProxy::isAXAuthenticated):
* Source/WebKit/UIProcess/WebProcessPool.h:
* Tools/WebKitTestRunner/mac/TestControllerMac.mm:
(WTR::TestController::initializeWebProcessAccessibility):

Canonical link: <a href="https://commits.webkit.org/310800@main">https://commits.webkit.org/310800@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/1cdbe10636467fd40f95cec637fca29b1df3b320

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/154945 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/28204 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/21364 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/163705 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/108416 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/22fdcd5d-61c3-4580-a722-c641e62c0268) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/156818 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/28343 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/28053 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/119879 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/84732 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/157904 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/22155 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/139148 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/100572 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/21240 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/19266 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/11531 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/130902 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/16992 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/166180 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/9702 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/18601 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/127979 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/27749 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/23305 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/128118 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/34774 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/27673 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/138785 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/84382 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/22997 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/15580 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/27365 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/91469 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/26943 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/27174 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/27016 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
+++ b/Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
@@ -415,7 +415,7 @@ void AXIsolatedTree::addUnconnectedNode(Ref<AccessibilityObject> axObject)
     m_unconnectedNodes.add(objectID);
 }
 
-void AXIsolatedTree::queueRemovals(Vector<AXID>&& subtreeRemovals)
+void AXIsolatedTree::queueRemovals(Vector<NodeAndParentID>&& subtreeRemovals)
 {
     AX_ASSERT(isMainThread());
 
@@ -423,12 +423,12 @@ void AXIsolatedTree::queueRemovals(Vector<AXID>&& subtreeRemovals)
     queueRemovalsLocked(WTF::move(subtreeRemovals));
 }
 
-void AXIsolatedTree::queueRemovalsLocked(Vector<AXID>&& subtreeRemovals)
+void AXIsolatedTree::queueRemovalsLocked(Vector<NodeAndParentID>&& subtreeRemovals)
 {
     AX_ASSERT(isMainThread());
     AX_ASSERT(m_changeLogLock.isLocked());
 
-    m_pendingSubtreeRemovals.addAll(WTF::move(subtreeRemovals));
+    m_pendingSubtreeRemovals.appendVector(WTF::move(subtreeRemovals));
     m_pendingProtectedFromDeletionIDs.addAll(std::exchange(m_protectedFromDeletionIDs, { }));
 }
 
@@ -478,7 +478,7 @@ Vector<AXIsolatedTree::NodeChange> AXIsolatedTree::resolveAppends()
     return resolvedAppends;
 }
 
-void AXIsolatedTree::queueAppendsAndRemovals(Vector<NodeChange>&& appends, Vector<AXID>&& subtreeRemovals)
+void AXIsolatedTree::queueAppendsAndRemovals(Vector<NodeChange>&& appends, Vector<NodeAndParentID>&& subtreeRemovals)
 {
     AX_ASSERT(isMainThread());
 
@@ -1103,7 +1103,9 @@ void AXIsolatedTree::updateChildren(AccessibilityObject& axObject, ResolveNodeCh
         updateDependentProperties(*axAncestor);
     }
 
-    m_subtreesToRemove.appendVector(WTF::move(oldChildrenIDs));
+    AXID parentID = axAncestor->objectID();
+    for (AXID childID : oldChildrenIDs)
+        m_subtreesToRemove.append({ childID, parentID });
     if (resolveNodeChanges == ResolveNodeChanges::Yes)
         queueRemovalsAndUnresolvedChanges();
 }
@@ -1323,7 +1325,7 @@ void AXIsolatedTree::removeNode(AXID axID, std::optional<AXID> parentID)
 
     m_unresolvedPendingAppends.remove(axID);
     removeSubtreeFromNodeMap(axID, parentID);
-    queueRemovals({ axID });
+    queueRemovals({ { axID, parentID } });
 }
 
 void AXIsolatedTree::removeSubtreeFromNodeMap(std::optional<AXID> objectID, std::optional<AXID> axParentID)
@@ -1429,6 +1431,103 @@ void AXIsolatedTree::clearTreeContentsLocked()
     // will be cleaned up automatically when the tree is destroyed.
 }
 
+// When a node is queued for both subtree removal and appending in the same
+// batch (e.g. when an in-process FrameHost is replaced with an out-of-process
+// one during site isolation), the old subtree snapshot and the new one can
+// both end up in m_pendingAppends. The removal root may not yet be in the
+// node map, so deleteSubtree won't encounter it.
+//
+// For each unique appended ID, this function walks up its parent chain to
+// check whether any ancestor was removed from the same parent it's being
+// appended under. Results are memoized so each node is visited at most once
+// across all walks, giving O(n) total work.
+//
+// Nodes that were removed from one parent but appended under a different
+// parent are reparenting operations and should be kept.
+void AXIsolatedTree::removeStaleAppends(const Vector<NodeAndParentID>& removals)
+{
+    // Build a map from removed AXID to the parent it was removed from, for
+    // O(1) lookups during the ancestor walk.
+    HashMap<AXID, Markable<AXID>> removedNodeToParentMap;
+    for (const auto& removal : removals)
+        removedNodeToParentMap.set(removal.nodeID, removal.parentID);
+
+    // Build a map from appended AXID to its final parentID (last entry wins,
+    // since later appends override earlier ones in the append loop).
+    HashMap<AXID, Markable<AXID>> appendedParentIDs;
+    for (const auto& item : m_pendingAppends)
+        appendedParentIDs.set(item.data.axID, item.data.parentID);
+
+    // For each unique appended ID, walk up the parent chain to determine
+    // whether it or any ancestor was removed from the same parent it's being
+    // appended under. Memoize results so shared ancestors are only resolved
+    // once.
+    HashMap<AXID, bool> ancestorWasRemoved;
+    HashSet<AXID> idsToRemove;
+
+    for (const auto& appendedID : appendedParentIDs.keys()) {
+        if (ancestorWasRemoved.contains(appendedID))
+            continue;
+
+        Vector<AXID> unresolvedAncestors;
+        AXID currentID = appendedID;
+        std::optional<bool> result;
+
+        while (!result) {
+            auto cachedIterator = ancestorWasRemoved.find(currentID);
+            if (cachedIterator != ancestorWasRemoved.end()) {
+                result = cachedIterator->value;
+                break;
+            }
+
+            unresolvedAncestors.append(currentID);
+
+            // Check if this node was removed. If so, compare parents: only
+            // keep it if both parents are known and differ (a reparent).
+            // If either parent is unknown, treat as stale to be safe.
+            auto removedIterator = removedNodeToParentMap.find(currentID);
+            if (removedIterator != removedNodeToParentMap.end()) {
+                auto currentAppendIterator = appendedParentIDs.find(currentID);
+                Markable<AXID> appendedParentID = (currentAppendIterator != appendedParentIDs.end()) ? currentAppendIterator->value : Markable<AXID> { };
+                Markable<AXID> removedParentID = removedIterator->value;
+                bool isReparent = appendedParentID && removedParentID && *appendedParentID != *removedParentID;
+                result = !isReparent;
+                break;
+            }
+
+            // Find this node's parent, first in pending appends, then in
+            // the existing tree.
+            auto appendedIterator = appendedParentIDs.find(currentID);
+            if (appendedIterator != appendedParentIDs.end() && appendedIterator->value) {
+                currentID = *appendedIterator->value;
+                continue;
+            }
+
+            if (auto* existingObject = objectForID(currentID)) {
+                if (auto existingParentID = existingObject->parent()) {
+                    currentID = *existingParentID;
+                    continue;
+                }
+            }
+
+            // Reached a root or dead end — no removed ancestor.
+            result = false;
+        }
+
+        for (const auto& axID : unresolvedAncestors) {
+            ancestorWasRemoved.set(axID, *result);
+            if (*result)
+                idsToRemove.add(axID);
+        }
+    }
+
+    if (!idsToRemove.isEmpty()) {
+        m_pendingAppends.removeAllMatching([&](const NodeChange& change) {
+            return idsToRemove.contains(change.data.axID);
+        });
+    }
+}
+
 void AXIsolatedTree::applyPendingChangesLocked()
 {
     AXTRACE("AXIsolatedTree::applyPendingChanges"_s);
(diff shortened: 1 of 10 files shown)


Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1446,6 +1446,12 @@ void AXIsolatedTree::applyPendingChangesLocked()
         m_focusedNodeID = m_pendingFocusedNodeID;
     }
 
+    // Collect the set of IDs pending removal before processing removals,
+    // so we can skip appending nodes that are also queued for removal.
+    HashSet<AXID> pendingRemovalIDs;
+    for (const auto& axID : m_pendingSubtreeRemovals)
+        pendingRemovalIDs.add(axID);
```
twilco: Couple of questions here:

1. `m_pendingSubtreeRemovals` is a `HashSet<AXID>` — would it not work to do this? `auto subtreeRemovalsCopy = m_pendingSubtreeRemovals`
2. m_pendingSubtreeRemovals deletes entire subtrees, but your addition to the `m_pendingAppends` only checks against subtree roots that were removed.
3. Should `deleteSubtree` just remove the element from `m_pendingAppends`?

minorninth: Your point #2 was spot on - we needed to check the whole subtrees. I was able to intermittently reproduce the need for that. I played around with it a while and came up with an algorithm that is efficient by walking up the ancestors and memoizing the results.

minorninth: Also: updated to specifically ensure that it only removes the subtree when it was the SAME parent, so it won't mess up reparenting. Existing tests all pass with isolated tree on.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -415,20 +415,21 @@ void AXIsolatedTree::addUnconnectedNode(Ref<AccessibilityObject> axObject)
     m_unconnectedNodes.add(objectID);
 }
 
-void AXIsolatedTree::queueRemovals(Vector<AXID>&& subtreeRemovals)
+void AXIsolatedTree::queueRemovals(Vector<std::pair<AXID, AXID>>&& subtreeRemovals)
```
twilco: IMO, this would be expressed more cleanly as a struct so the fields are named and thus less likely to be mixed up.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -415,20 +415,21 @@ void AXIsolatedTree::addUnconnectedNode(Ref<AccessibilityObject> axObject)
...
 
-void AXIsolatedTree::queueRemovalsLocked(Vector<AXID>&& subtreeRemovals)
+void AXIsolatedTree::queueRemovalsLocked(Vector<std::pair<AXID, AXID>>&& subtreeRemovals)
 {
     AX_ASSERT(isMainThread());
     AX_ASSERT(m_changeLogLock.isLocked());
 
-    m_pendingSubtreeRemovals.addAll(WTF::move(subtreeRemovals));
+    for (auto& [childID, parentID] : subtreeRemovals)
+        m_pendingSubtreeRemovals.set(childID, parentID);
```
twilco: I see this is a HashMap now. Is there any reason not to use a Vector? I guess it begs the question — is it ever possible for two parents to delete the same child? What should we do in that case?

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1315,7 +1318,8 @@ void AXIsolatedTree::removeNode(AXID axID, std::optional<AXID> parentID)
 
     m_unresolvedPendingAppends.remove(axID);
     removeSubtreeFromNodeMap(axID, parentID);
-    queueRemovals({ axID });
+    if (parentID)
+        queueRemovals({ { axID, *parentID } });
```
twilco: In theory the only thing for which something shouldn't have a parentID is when it's the root...but if we have any bugs where objects don't report a parent, even for a short time, this change will mean we never remove those.

Maybe we should `AX_ASSERT` this is the root?

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1421,6 +1425,96 @@ void AXIsolatedTree::clearTreeContentsLocked()
...
+            // Find this node's parent, first in pending appends, then in
+            // the existing tree.
+            auto appendedIterator = appendedParentIDs.find(currentID);
+            if (appendedIterator != appendedParentIDs.end() && appendedIterator->value) {
+                currentID = *appendedIterator->value;
+                continue;
+            }
+
+            auto existingIterator = m_readerThreadNodeMap.find(currentID);
+            if (existingIterator != m_readerThreadNodeMap.end()) {
```
twilco: Any reason not to just use `AXIsolatedTree::objectForID`?

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1489,6 +1590,9 @@ void AXIsolatedTree::applyPendingChangesLocked()
     }
     m_pendingProtectedFromDeletionIDs.clear();
 
+    if (!removedIDs.isEmpty())
+        removeStaleAppends(removedIDs);
```
twilco, line 1604: It's unfortunate to add a O(n) operation while locked. I'm sure it would be fine...but wanted to point it out in case there's any way we can think of to do this work outside the lock.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1421,6 +1425,96 @@ void AXIsolatedTree::clearTreeContentsLocked()
...
+    // once.
+    HashMap<AXID, bool> ancestorWasRemoved;
+    HashSet<AXID> idsToRemove;
+
+    for (const auto& appendedID : appendedParentIDs.keys()) {
+        if (ancestorWasRemoved.contains(appendedID))
+            continue;
+
+        Vector<AXID> unresolvedAncestors;
+        auto currentID = appendedID;
```
twilco: IMO when the type is AXID, we should just write AXID. It's the same number of characters as auto and makes the code more readable.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
minorninth: Changed to a vector. It seems possible for a single node to be reparented multiple times before the next isolated tree batch update, so we should be robust to that.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
minorninth: No good ideas here, sorry.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1322,8 +1316,11 @@ void AXIsolatedTree::removeNode(AXID axID, std::optional<AXID> parentID)
     AX_ASSERT(isMainThread());
 
     m_unresolvedPendingAppends.remove(axID);
+    // The only node without a parent should be the root.
+    ASSERT(parentID || !m_nodeMap.contains(axID) || !m_nodeMap.get(axID).parentID);
```
twilco: Can this be `AX_ASSERT` so we get it in release mode during local development too?

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1429,6 +1426,101 @@ void AXIsolatedTree::clearTreeContentsLocked()
...
+{
+    // Build a map from removed AXID to the parent it was removed from, for
+    // O(1) lookups during the ancestor walk.
+    HashMap<AXID, AXID> removedNodeToParentMap;
+    for (const auto& removal : removals)
+        removedNodeToParentMap.set(removal.nodeID, removal.parentID);
+
+    // Build a map from appended AXID to its final parentID (last entry wins,
+    // since later appends override earlier ones in the append loop).
+    HashMap<AXID, std::optional<AXID>> appendedParentIDs;
```
twilco: `Markable<AXID>` would be a bit more space efficient

twilco (approved):

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
minorninth: Unfortunately I hit this assert on the bots, but only intermittently. I wonder if it's during teardown. I modified it to explicitly represent a parentID not being known.

PR #61721, opened 2026-03-30
[JSC][WASM][Debugger] Extend WASM debugger trap interception to all IPInt exception sites

hyjorc1:
#### 6fa604090d4e14d47b354c9572006eeed49bcb7b
<pre>
[JSC][WASM][Debugger] Extend WASM debugger trap interception to all IPInt exception sites
<a href="https://bugs.webkit.org/show_bug.cgi?id=311138">https://bugs.webkit.org/show_bug.cgi?id=311138</a>
<a href="https://rdar.apple.com/173723442">rdar://173723442</a>

Reviewed by Mark Lam.

Previously the WASM debugger only intercepted the unreachable opcode.
This patch extends interception to all IPInt-dispatched traps (division
by zero, integer overflow, out-of-bounds memory, null reference, bad
signature, stack overflow, etc.) by routing all exception paths through
a single C++ handler, ipint_extern_handle_debugger_trap_if_needed.

Inline trap sites use the handleDebuggerTrapIfNeededAndThrowWasmTrap macro,
which remains 2 instructions (store + jmp) to avoid overflowing fixed-size
IPInt dispatch slots. On ARM64/ARM64E it jumps to the shared
_wasm_ipint_check_debugger_hook_and_throw_trap trampoline which saves IPInt
registers and calls the handler; on other platforms it jumps directly to
_wasm_throw_from_slow_path_trampoline since the WASM debugger is not supported there.

The operationCallMayThrowImpl exception path calls
ipint_extern_handle_debugger_trap_if_needed directly, using the IPInt
registers already on the stack from operationCallMayThrowImpl&apos;s own save sequence.

Stack-overflow and Termination traps go through
operationCallMayThrowPreservingVolatileRegisters, which takes the same
operationCallMayThrowImpl exception path with volatile registers preserved.

The fault-handler path (wasm_throw_from_fault_handler_trampoline_reg_instance)
copies instance-&gt;m_exception into the CFR ArgumentCountIncludingThis slot
before calling handleDebuggerTrapIfNeeded(), since IPInt registers remain
live when the Mach exception handler redirects the machine PC to the trampoline.

Four new tests cover one representative trap per execution path:
WasmDivByZeroTrapTestCase (inline asm trap)
WasmOutOfBoundsCallIndirectTrapTestCase (C++ slow-path trap)
WasmStackOverflowTrapTestCase (prologue/volatile-preserved trap)
WasmOobMemoryTrapTestCase (fault-handler path)

Canonical link: <a href="https://commits.webkit.org/310417@main">https://commits.webkit.org/310417@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/fc947a43f978cc6938c2ec47a0cc911d37b15841

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/153780 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/26564 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/20181 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/162531 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/107241 "Built successfully") | [loading 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/fc71b0f4-cddb-41b0-9ee3-ab49bc938184) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/27092 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/26886 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/118893 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/84072 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [loading 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/6f5acc01-03fe-4aa4-be38-1bf8f51dd777) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/156739 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/21159 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/138083 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/99603 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/a41d014e-fac6-49bb-9fc5-f0304c973576) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/20239 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/18197 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/10363 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/145793 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/129887 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/15942 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/165001 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/14604 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/8135 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/17536 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/126981 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/26361 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/22228 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/127148 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/34489 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/26363 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/137737 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/83041 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/22047 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/14520 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/185416 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/25980 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/90268 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/47559 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/25671 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/25831 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/25731 "Built successfully") | | | |

### DIFF
--- a/JSTests/wasm/debugger/resources/wasm/trap-div-by-zero.js
+++ b/JSTests/wasm/debugger/resources/wasm/trap-div-by-zero.js
@@ -1,0 +1,40 @@
+var wasm = new Uint8Array([
+    // [0x00] WASM header
+    0x00, 0x61, 0x73, 0x6d, // magic: \0asm
+    0x01, 0x00, 0x00, 0x00, // version: 1
+
+    // [0x08] Type section: 1 type, func () -> i32
+    0x01, 0x05, 0x01, 0x60, 0x00, 0x01, 0x7f,
+
+    // [0x0f] Function section: 1 function, type index 0
+    0x03, 0x02, 0x01, 0x00,
+
+    // [0x13] Export section: export function 0 as "triggerTrap"
+    0x07, 0x0f, 0x01, 0x0b, 0x74, 0x72, 0x69, 0x67, 0x67, 0x65, 0x72, 0x54, 0x72, 0x61, 0x70, 0x00, 0x00,
+
+    // [0x24] Code section
+    0x0a,       // section id = 10
+    0x09,       // section size = 9
+    0x01,       // 1 function body
+    0x07,       // body size = 7
+    0x00,       // 0 local declarations
+    0x41, 0x0a, // [0x29][0x2a] i32.const 10
+    0x41, 0x00, // [0x2b][0x2c] i32.const 0
+    0x6d,       // [0x2d] i32.div_s  <- DivisionByZero trap
+    0x0b        // [0x2e] end
+]);
+
+var module = new WebAssembly.Module(wasm);
+var instance = new WebAssembly.Instance(module);
+let triggerTrap = instance.exports.triggerTrap;
+
+let iteration = 0;
+for (;;) {
+    try {
+        triggerTrap();
+    } catch (e) {
+    }
+    iteration += 1;
+    if (iteration % 1e5 == 0)
+        print("iteration=", iteration);
+}
--- a/JSTests/wasm/debugger/resources/wasm/trap-oob-memory.js
+++ b/JSTests/wasm/debugger/resources/wasm/trap-oob-memory.js
@@ -1,0 +1,42 @@
+var wasm = new Uint8Array([
+    // [0x00] WASM header
+    0x00, 0x61, 0x73, 0x6d, // magic: \0asm
+    0x01, 0x00, 0x00, 0x00, // version: 1
+
+    // [0x08] Type section: 1 type, func () -> i32
+    0x01, 0x05, 0x01, 0x60, 0x00, 0x01, 0x7f,
+
+    // [0x0f] Function section: 1 function, type index 0
+    0x03, 0x02, 0x01, 0x00,
+
+    // [0x13] Memory section: 1 memory, min=1 page (64KB), no max
+    0x05, 0x03, 0x01, 0x00, 0x01,
+
+    // [0x18] Export section: export function 0 as "triggerTrap"
+    0x07, 0x0f, 0x01, 0x0b, 0x74, 0x72, 0x69, 0x67, 0x67, 0x65, 0x72, 0x54, 0x72, 0x61, 0x70, 0x00, 0x00,
+
+    // [0x29] Code section
+    0x0a,               // section id = 10
+    0x0b,               // section size = 11
+    0x01,               // 1 function body
+    0x09,               // body size = 9
+    0x00,               // 0 local declarations
+    0x41, 0x80, 0x80, 0x04, // [0x2e][0x2f][0x30][0x31] i32.const 65536 (1 page = 64KB, so OOB)
+    0x28, 0x02, 0x00,   // [0x32] i32.load align=2 offset=0  <- OutOfBoundsMemoryAccess trap
+    0x0b                // [0x35] end
+]);
+
+var module = new WebAssembly.Module(wasm);
+var instance = new WebAssembly.Instance(module);
+let triggerTrap = instance.exports.triggerTrap;
+
+let iteration = 0;
+for (;;) {
+    try {
+        triggerTrap();
+    } catch (e) {
+    }
+    iteration += 1;
+    if (iteration % 1e5 == 0)
+        print("iteration=", iteration);
+}
--- a/JSTests/wasm/debugger/resources/wasm/trap-stack-overflow.js
+++ b/JSTests/wasm/debugger/resources/wasm/trap-stack-overflow.js
@@ -1,0 +1,38 @@
+var wasm = new Uint8Array([
+    // [0x00] WASM header
+    0x00, 0x61, 0x73, 0x6d, // magic: \0asm
+    0x01, 0x00, 0x00, 0x00, // version: 1
+
+    // [0x08] Type section: 1 type, func () -> ()
+    0x01, 0x04, 0x01, 0x60, 0x00, 0x00,
+
+    // [0x0e] Function section: 1 function, type index 0
+    0x03, 0x02, 0x01, 0x00,
+
+    // [0x12] Export section: export function 0 as "triggerTrap"
+    0x07, 0x0f, 0x01, 0x0b, 0x74, 0x72, 0x69, 0x67, 0x67, 0x65, 0x72, 0x54, 0x72, 0x61, 0x70, 0x00, 0x00,
+
+    // [0x23] Code section
+    0x0a,       // section id = 10
+    0x06,       // section size = 6
+    0x01,       // 1 function body
+    0x04,       // body size = 4
+    0x00,       // 0 local declarations
+    0x10, 0x00, // [0x28][0x29] call func 0  <- infinite recursion, StackOverflow trap
+    0x0b,       // [0x2a] end
+]);
+
+var module = new WebAssembly.Module(wasm);
+var instance = new WebAssembly.Instance(module);
+let triggerTrap = instance.exports.triggerTrap;
+
+let iteration = 0;
+for (;;) {
+    try {
+        triggerTrap();
+    } catch (e) {
+    }
+    iteration += 1;
+    if (iteration % 1e5 == 0)
+        print("iteration=", iteration);
+}
--- a/Source/JavaScriptCore/llint/InPlaceInterpreter.asm
+++ b/Source/JavaScriptCore/llint/InPlaceInterpreter.asm
@@ -270,7 +270,7 @@ macro checkStackOverflow(callee, scratch)
 
 if not ADDRESS64
     bpbeq scratch, cfr, .checkTrapAwareSoftStackLimit
-    ipintException(StackOverflow)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(StackOverflow)
 .checkTrapAwareSoftStackLimit:
 end
     bpbeq JSWebAssemblyInstance::m_stackMirror + StackManager::Mirror::m_trapAwareSoftStackLimit[wasmInstance], scratch, .stackHeightOK
@@ -284,7 +284,8 @@ if X86_64
 else
         move callee, a2
 end
-        cCall3(_ipint_extern_check_stack_and_vm_traps)
+        move cfr, a3
+        cCall4(_ipint_extern_check_stack_and_vm_traps)
     end)
 
 .stackHeightOK:
@@ -391,6 +392,11 @@ macro operationCallMayThrowImpl(fn, sizeOfExtraRegistersPreserved)
     bpneq r1, (constexpr JSC::IPInt::SlowPathExceptionTag), .continuation
 
     storei r0, ArgumentCountIncludingThis + PayloadOffset[cfr]
+    if ARM64 or ARM64E
+        move cfr, a1
+        move sp, a2
+        operationCall(macro() cCall3(_ipint_extern_handle_debugger_trap_if_needed) end)
+    end
     addp sizeOfExtraRegistersPreserved + (4 * MachineRegisterSize), sp
     jmp _wasm_throw_from_slow_path_trampoline
 .continuation:
@@ -414,9 +420,16 @@ macro operationCallMayThrowPreservingVolatileRegisters(fn)
 end
 
 # Exception handling
-macro ipintException(exception)
+#
+# debugger-aware trap. 2 instructions; heavy logic in _wasm_ipint_check_debugger_hook_and_throw_trap due to fixed-size IPInt dispatch slots.
+macro handleDebuggerTrapIfNeededAndThrowWasmTrap(exception)
     storei constexpr Wasm::ExceptionType::%exception%, ArgumentCountIncludingThis + PayloadOffset[cfr]
+if ADDRESS64 and (ARM64 or ARM64E)
+   # Currently, only ARM64 and ARM64E with ADDRESS64 platforms support the WasmDebugger.
+    jmp _wasm_ipint_check_debugger_hook_and_throw_trap
+else
     jmp _wasm_throw_from_slow_path_trampoline
+end
 end
 
 # OSR
@@ -1166,6 +1179,27 @@ macro jumpToException()
     end
 end
 
+macro handleDebuggerTrapIfNeeded()
+    push PC, MC
+    push PL, ws0    # sp[0]=PL, sp[1]=ws0 (IPIntCallee*), sp[2]=PC, sp[3]=MC
+    move cfr, a1
+    move sp, a2     # a2 = pointer to saved [PL, ws0, PC, MC]
+    operationCall(macro() cCall3(_ipint_extern_handle_debugger_trap_if_needed) end)
+    addp 4 * MachineRegisterSize, sp
+end
+
+op(wasm_ipint_check_debugger_hook_and_throw_trap, macro ()
+    handleDebuggerTrapIfNeeded()
+    # r0 == 0 i.e. DebuggerTrapStatus::ResolvedByDebugger i.e. this was purely a debugger trap / breakpoint,
+    #              and has been handled.  We should continue executing because it's not a Wasm trap.
+    # r0 == 1 i.e. DebuggerTrapStatus::NotResolvedByDebugger i.e. this was a fatal Wasm trap.  We should
+    #              throw it to terminate Wasm execution.
+    btpz r0, .continue
+    jmp _wasm_throw_from_slow_path_trampoline
+.continue:
+    nextIPIntInstruction()
+end)
+
 op(wasm_throw_from_slow_path_trampoline, macro ()
     validateOpcodeConfig(t5)
     loadp JSWebAssemblyInstance::m_vm[wasmInstance], t5
@@ -1196,6 +1230,14 @@ op(wasm_unwind_from_slow_path_trampoline, macro()
 end)
 
 op(wasm_throw_from_fault_handler_trampoline_reg_instance, macro ()
+    # enableWasmDebugger disables BBQ/OMG, so this trampoline is only
+    # reached from IPInt when the debugger is active. The signal handler only patches
+    # the machine PC, so IPInt registers (PC, MC, PL, ws0, cfr) are still live.
+    # Exception type comes from instance->m_exception; copy to CFR slot for handle_debugger_trap_if_needed.
+    loadi JSWebAssemblyInstance::m_exception[wasmInstance], t0
+    storei t0, ArgumentCountIncludingThis + PayloadOffset[cfr]
+    handleDebuggerTrapIfNeeded()
+
     move wasmInstance, a2
     loadp JSWebAssemblyInstance::m_vm[a2], a0
     loadp VM::topEntryFrame[a0], a0
--- a/Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
+++ b/Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
@@ -235,26 +235,7 @@ end
     #############################
 
 ipintOp(_unreachable, macro()
-    # unreachable
-
-    # Push to stack for the handler
-    push PC, MC
-    push PL, ws0
-
-    move cfr, a1
-    move sp, a2
-    operationCall(macro() cCall3(_ipint_extern_unreachable_handler) end)
-
-    # Remove pushed values
-    addq 4 * SlotSize, sp
-
-    bqeq r0, 0, .exception
-
-.continue:
-    nextIPIntInstruction()
-
-.exception:
-    ipintException(Unreachable)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(Unreachable)
 end)
 
 ipintOp(_nop, macro()
@@ -396,7 +377,7 @@ ipintOp(_throw_ref, macro()
     jumpToException()
 
 .throw_null_ref:
-    throwException(NullExnrefReference)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(NullExnrefReference)
 end)
 
 macro uintDispatch()
@@ -927,7 +908,7 @@ macro ipintCheckMemoryBound(mem, scratch, size)
     # Memory indices are 32 bit
     leap size - 1[mem], scratch
     bpb scratch, boundsCheckingSize, .continuation
-    ipintException(OutOfBoundsMemoryAccess)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(OutOfBoundsMemoryAccess)
 .continuation:
 end
 
@@ -994,7 +975,7 @@ macro loadStoreAdvanceMCAndMakePointer(instrLenReg, wasmAddrReg, size, scratch, 
     jmp .done
 
 .outOfBounds:
-    ipintException(OutOfBoundsMemoryAccess)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(OutOfBoundsMemoryAccess)
 .done:
 end
 
@@ -1845,10 +1826,10 @@ ipintOp(_i32_div_s, macro()
     nextIPIntInstruction()
 
 .ipint_i32_div_s_throwDivisionByZero:
-    ipintException(DivisionByZero)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(DivisionByZero)
 
 .ipint_i32_div_s_throwIntegerOverflow:
-    ipintException(IntegerOverflow)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(IntegerOverflow)
 end)
 
 ipintOp(_i32_div_u, macro()
@@ -1870,7 +1851,7 @@ ipintOp(_i32_div_u, macro()
     nextIPIntInstruction()
 
 .ipint_i32_div_u_throwDivisionByZero:
-    ipintException(DivisionByZero)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(DivisionByZero)
 end)
 
 ipintOp(_i32_rem_s, macro()
@@ -1908,7 +1889,7 @@ ipintOp(_i32_rem_s, macro()
     nextIPIntInstruction()
 
 .ipint_i32_rem_s_throwDivisionByZero:
-    ipintException(DivisionByZero)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(DivisionByZero)
 end)
 
 ipintOp(_i32_rem_u, macro()
@@ -1934,7 +1915,7 @@ ipintOp(_i32_rem_u, macro()
     nextIPIntInstruction()
 
 .ipint_i32_rem_u_throwDivisionByZero:
-    ipintException(DivisionByZero)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(DivisionByZero)
 end)
 
 ipintOp(_i32_and, macro()
@@ -2117,10 +2098,10 @@ ipintOp(_i64_div_s, macro()
     nextIPIntInstruction()
 
 .ipint_i64_div_s_throwDivisionByZero:
-    ipintException(DivisionByZero)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(DivisionByZero)
 
 .ipint_i64_div_s_throwIntegerOverflow:
-    ipintException(IntegerOverflow)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(IntegerOverflow)
 end)
 
 ipintOp(_i64_div_u, macro()
@@ -2142,7 +2123,7 @@ ipintOp(_i64_div_u, macro()
     nextIPIntInstruction()
 
 .ipint_i64_div_u_throwDivisionByZero:
-    ipintException(DivisionByZero)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(DivisionByZero)
 end)
 
 ipintOp(_i64_rem_s, macro()
@@ -2180,7 +2161,7 @@ ipintOp(_i64_rem_s, macro()
     nextIPIntInstruction()
 
 .ipint_i64_rem_s_throwDivisionByZero:
-    ipintException(DivisionByZero)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(DivisionByZero)
 end)
 
 ipintOp(_i64_rem_u, macro()
@@ -2206,7 +2187,7 @@ ipintOp(_i64_rem_u, macro()
     nextIPIntInstruction()
 
 .ipint_i64_rem_u_throwDivisionByZero:
-    ipintException(DivisionByZero)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(DivisionByZero)
 end)
 
 ipintOp(_i64_and, macro()
@@ -2734,7 +2715,7 @@ ipintOp(_i32_trunc_f32_s, macro()
     nextIPIntInstruction()
 
 .ipint_trunc_i32_f32_s_outOfBoundsTrunc:
-    ipintException(OutOfBoundsTrunc)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(OutOfBoundsTrunc)
 end)
 
 ipintOp(_i32_trunc_f32_u, macro()
@@ -2753,7 +2734,7 @@ ipintOp(_i32_trunc_f32_u, macro()
     nextIPIntInstruction()
 
 .ipint_trunc_i32_f32_u_outOfBoundsTrunc:
-    ipintException(OutOfBoundsTrunc)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(OutOfBoundsTrunc)
 end)
 
 ipintOp(_i32_trunc_f64_s, macro()
@@ -2772,7 +2753,7 @@ ipintOp(_i32_trunc_f64_s, macro()
     nextIPIntInstruction()
 
 .ipint_trunc_i32_f64_s_outOfBoundsTrunc:
-    ipintException(OutOfBoundsTrunc)
+    handleDebuggerTrapIfNeededAndThrowWasmTrap(OutOfBoundsTrunc)
 end)
 
 ipintOp(_i32_trunc_f64_u, macro()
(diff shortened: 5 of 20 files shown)


JSTests/wasm/debugger/resources/wasm/trap-div-by-zero.js
```diff
@@ -0,0 +1,41 @@
...
+var module = new WebAssembly.Module(wasm);
+var instance = new WebAssembly.Instance(module);
+let crasher = instance.exports.crasher;
+
+let iteration = 0;
+for (;;) {
+    try {
+        crasher();
+    } catch (e) {
+
```
MenloDorian: nit: unneeded empty line.  Please remove.

JSTests/wasm/debugger/resources/wasm/trap-div-by-zero.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+    iteration += 1;
+    if (iteration % 1e5 == 0)
+        print("iteration=", iteration);
```
MenloDorian, line 39: Please put this behind a `verbose` flag that is false by default.  Don't make the JSC stress test dump lots of lines of text that obscures errors.

JSTests/wasm/debugger/resources/wasm/trap-oob-memory.js
```diff
@@ -0,0 +1,43 @@
...
+let iteration = 0;
+for (;;) {
+    try {
+        crasher();
+    } catch (e) {
+
+    }
+    iteration += 1;
+    if (iteration % 1e5 == 0)
+        print("iteration=", iteration);
```
MenloDorian, line 41: Ditto: put behind verbose flag defaulting to false.  And ditto for tests below too.

JSTests/wasm/debugger/resources/wasm/trap-oob-memory.js
```diff
@@ -0,0 +1,43 @@
...
+var module = new WebAssembly.Module(wasm);
+var instance = new WebAssembly.Instance(module);
+let crasher = instance.exports.crasher;
+
+let iteration = 0;
+for (;;) {
+    try {
+        crasher();
+    } catch (e) {
+
```
MenloDorian: nit: please remove.  And ditto for tests below too.

JSTests/wasm/debugger/resources/wasm/trap-div-by-zero.js
```diff
@@ -0,0 +1,41 @@
...
+    0x00,       // 0 local declarations
+    0x41, 0x0a, // [0x25][0x26] i32.const 10
+    0x41, 0x00, // [0x27][0x28] i32.const 0
+    0x6d,       // [0x29] i32.div_s  <- DivisionByZero trap
+    0x0b        // [0x2a] end
+]);
+
+var module = new WebAssembly.Module(wasm);
+var instance = new WebAssembly.Instance(module);
+let crasher = instance.exports.crasher;
```
MenloDorian: This is misleading.  Let's not call this `crasher` since it is not intended to crash.  Let's call it `throws` or `throwException` or `triggerTrap` or `triggerException` ... something like that.

Ditto for other tests below.

JSTests/wasm/debugger/resources/wasm/trap-div-by-zero.js
```diff
@@ -0,0 +1,41 @@
...
+
+let iteration = 0;
+for (;;) {
```
MenloDorian, line 32: Let's not use an infinite loop.  How long does it take for the debugger to observe this exception?  Can we make it deterministic and verify that the exception was observed?

My concern is that if the trigger function below fails to trigger the exception, we'd make these tests take a long time to time out, and slow down all test runs and EWS bots.   Ideally, we limit this to a small finite number of iterations, and set a success status.  Then, check success status after this loop, and if the test failed, print an error message and throw an uncaught exception.  This will uncaught exception will cause the test to fail more quickly instead of waiting for a time out in the test harness.

JSTests/wasm/debugger/resources/wasm/trap-stack-overflow.js
```diff
@@ -0,0 +1,39 @@
+var wasm = new Uint8Array([
```
MenloDorian, line 1: Since this is a StackOverflow test where the interesting bit is in the handling of the StackOverflow, I suggest requiring JSC option `maxPerThreadStackUsage` to set the stack to something smaller e.g 128K or 256K to shorten the test run time.  The current default limit is 5M, which I think is overkill for this test.

JSTests/wasm/debugger/resources/wasm/trap-div-by-zero.js
hyjorc1: `JSTests/wasm/debugger/test-wasm-debugger.py` is not yet integrated into EWS. The test launches two processes: one runs the jsc shell with the target test, and one runs LLDB. Since LLDB attachment can take some time, the test script needs to run long enough for LLDB to attach. Once attached, `JSTests/wasm/debugger/tests/tests.py` interacts with LLDB via pattern matching. After each test case passes, both processes are terminated.

JSTests/wasm/debugger/resources/wasm/trap-div-by-zero.js
hyjorc1: These are not tested under JSC stress test. See the reply at https://github.com/WebKit/WebKit/pull/61721#discussion_r3012637918.

JSTests/wasm/debugger/resources/wasm/trap-stack-overflow.js
hyjorc1: These are not tested under JSC stress test. See the reply at https://github.com/WebKit/WebKit/pull/61721#discussion_r3012637918. Will add a FIXME here for future WASM debugger test infrastructure update.

MenloDorian: Your script is invoking these jsc tests from your python script, correct?  Would it make sense to restrict their stack usage to 512K or lower across all these tests?  I presume that option will be easy to add, but I don't know if it'll break the tests.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -391,8 +392,9 @@ macro operationCallMayThrowImpl(fn, sizeOfExtraRegistersPreserved)
     bpneq r1, (constexpr JSC::IPInt::SlowPathExceptionTag), .continuation
 
     storei r0, ArgumentCountIncludingThis + PayloadOffset[cfr]
-    addp sizeOfExtraRegistersPreserved + (4 * MachineRegisterSize), sp
-    jmp _wasm_throw_from_slow_path_trampoline
+    # Pop any extra preserved registers so sp[0..3] = [PL, ws0, PC, MC].
+    addp sizeOfExtraRegistersPreserved, sp
+    jmp _wasm_trap_from_slow_path_with_debug_state
```
MenloDorian: This looks wrong.  `[PL, ws0, PC, MC]`  are the last 4 registers pushed at the top of this macro.  The `sizeOfExtraRegistersPreserved` parts were pushed by `preserveWasmVolatileRegisters()` back in `operationCallMayThrowPreservingVolatileRegisters` before calling `operationCallMayThrowImpl`.  So, you're getting rid of the `[PL, ws0, PC, MC]` values here and keeping some values from ``preserveWasmVolatileRegisters()`.

The correct thing to do would be to defer this `addp sizeOfExtraRegistersPreserved + (4 * MachineRegisterSize), sp` altogether till after the operation call in `wasm_trap_from_slow_path_with_debug_state`.

MenloDorian (changes requested):

JSTests/wasm/debugger/resources/wasm/trap-oob-memory.js
hyjorc1: These are not tested under JSC stress test. See the reply at https://github.com/WebKit/WebKit/pull/61721#discussion_r3012637918.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
hyjorc1: Thanks, fixed.

JSTests/wasm/debugger/resources/wasm/trap-stack-overflow.js
hyjorc1: Oh, `maxPerThreadStackUsage` already exists. Thanks, updated.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -414,9 +420,11 @@ macro operationCallMayThrowPreservingVolatileRegisters(fn)
 end
 
 # Exception handling
+#
+# debugger-aware trap. 2 instructions; heavy logic in _wasm_ipint_trap_handler_stub due to fixed-size IPInt dispatch slots.
 macro ipintException(exception)
```
MenloDorian: This `ipintException` name is not very descriptive.  Let's rename it to `handleDebuggerTrapIfNeededAndThrowWasmTrap` to clearly communicate what it does.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -414,9 +420,11 @@ macro operationCallMayThrowPreservingVolatileRegisters(fn)
...
     storei constexpr Wasm::ExceptionType::%exception%, ArgumentCountIncludingThis + PayloadOffset[cfr]
-    jmp _wasm_throw_from_slow_path_trampoline
+    jmp _wasm_ipint_trap_handler_stub
```
MenloDorian: Here, you should put the debugger supported checks:
```
if ADDRESS64 and (ARM64 or ARM64E)
   # Currently, only ARM64 and ARM64E with ADDRESS64 platforms support the WasmDebugger.
    jmp _wasm_ipint_handle_debugger_trap_if_needed_and_throw_wasm_trap
else
    jmp _wasm_throw_from_slow_path_trampoline
end
```

This way, you don't have to do the platform check in all the other places, and the code can read more coherently.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -270,7 +270,7 @@ macro checkStackOverflow(callee, scratch)
 
 if not ADDRESS64
     bpbeq scratch, cfr, .checkTrapAwareSoftStackLimit
-    ipintException(StackOverflow)
+    throwException(StackOverflow)
```
MenloDorian: Just have this call `handleDebuggerTrapIfNeededAndThrowWasmTrap` and let the macro decide whether to call the debugger hook or not.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -1166,6 +1174,25 @@ macro jumpToException()
...
+macro callIPIntTrapHandler()
+    push PC, MC
+    push PL, ws0    # sp[0]=PL, sp[1]=ws0 (IPIntCallee*), sp[2]=PC, sp[3]=MC
+    move cfr, a1
+    move sp, a2     # a2 = pointer to saved [PL, ws0, PC, MC]
+    operationCall(macro() cCall3(_ipint_extern_trap_handler) end)
+    addp 4 * MachineRegisterSize, sp
+end
+
+op(wasm_ipint_trap_handler_stub, macro ()
```
MenloDorian: Let's rename this to `wasm_ipint_check_debugger_hook_and_throw_trap` to describe what it actually does.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -1166,6 +1174,25 @@ macro jumpToException()
...
+    if ARM64 or ARM64E
```
MenloDorian: You can remove this condition now.  This stub should only ever be called if and only if the WasmDebugger is supported.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -1166,6 +1174,25 @@ macro jumpToException()
...
+        callIPIntTrapHandler()
+        btpz r0, .continue  # zero (false = breakpoint): continue execution; non-zero (true = throw)
+    end
```
MenloDorian: Ditto:  remove.

Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.cpp
```diff
@@ -1327,31 +1329,30 @@ static UNUSED_FUNCTION void displayWasmDebugState(JSWebAssemblyInstance* instanc
 }
 #endif
 
-WASM_IPINT_EXTERN_CPP_DECL(unreachable_handler, CallFrame* callFrame, Register* sp)
+WASM_IPINT_EXTERN_CPP_DECL(trap_handler, CallFrame* callFrame, Register* sp)
```
MenloDorian: Let's rename `trap_handler` to `handle_debugger_trap_if_needed` so as not to conflate it with other Wasm runtime traps.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -391,6 +392,11 @@ macro operationCallMayThrowImpl(fn, sizeOfExtraRegistersPreserved)
     bpneq r1, (constexpr JSC::IPInt::SlowPathExceptionTag), .continuation
 
     storei r0, ArgumentCountIncludingThis + PayloadOffset[cfr]
+    if ARM64 or ARM64E
+        move cfr, a1
+        move sp, a2
+        operationCall(macro() cCall3(_ipint_extern_trap_handler) end)
```
MenloDorian: Let's rename `_ipint_extern_trap_handler` to `_ipint_extern_handle_debugger_trap_if_needed`

Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.h
```diff
@@ -140,8 +140,8 @@ WASM_IPINT_EXTERN_CPP_HIDDEN_DECL(memory_atomic_wait32, uint64_t, uint32_t, uint
 WASM_IPINT_EXTERN_CPP_HIDDEN_DECL(memory_atomic_wait64, uint64_t, uint64_t, uint64_t);
 WASM_IPINT_EXTERN_CPP_HIDDEN_DECL(memory_atomic_notify, unsigned, unsigned, int32_t);
 
-WASM_IPINT_EXTERN_CPP_HIDDEN_DECL(check_stack_and_vm_traps, void* candidateNewStackPointer, Wasm::IPIntCallee*);
-WASM_IPINT_EXTERN_CPP_DECL(unreachable_handler, CallFrame*, Register*);
+WASM_IPINT_EXTERN_CPP_HIDDEN_DECL(check_stack_and_vm_traps, void* candidateNewStackPointer, Wasm::IPIntCallee*, CallFrame*);
+WASM_IPINT_EXTERN_CPP_DECL(trap_handler, CallFrame*, Register*);
```
MenloDorian: Let's rename `trap_handler` to `handle_debugger_trap_if_needed`.

Source/JavaScriptCore/wasm/debugger/WasmExecutionHandler.cpp
```diff
@@ -133,23 +134,31 @@ void ExecutionHandler::stopTheWorld(VM& debuggee, StopTheWorldEvent event)
     VMManager::singleton().notifyVMStop(debuggee, event);
 }
 
-bool ExecutionHandler::handleUnreachable(CallFrame* callFrame, JSWebAssemblyInstance* instance, IPIntCallee* callee, uint8_t* pc, uint8_t* mc, IPInt::IPIntLocal* locals, IPInt::IPIntStackEntry* stack)
+bool ExecutionHandler::handleBreakpointOrTrap(CallFrame* callFrame, JSWebAssemblyInstance* instance, IPIntCallee* callee, uint8_t* pc, uint8_t* mc, IPInt::IPIntLocal* locals, IPInt::IPIntStackEntry* stack, Wasm::ExceptionType exceptionType)
```
MenloDorian: Let's rename `handleBreakpointOrTrap` to `handleDebuggerTrapIfNeeded`.  Let's also have it return `DebuggerTrapStatus` instead (see below).

Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.cpp
```diff
@@ -1327,31 +1329,30 @@ static UNUSED_FUNCTION void displayWasmDebugState(JSWebAssemblyInstance* instanc
...
-            if (Options::verboseWasmDebugger())
-                displayWasmDebugState(instance, callee, stackPointer, pl);
-
-            breakpointHandled = debugServer.execution().handleUnreachable(callFrame, instance, callee, pc, mc, pl, stackPointer);
+            auto* callee = static_cast<Wasm::IPIntCallee*>(sp[1].pointer());
+            auto* stack = std::bit_cast<IPIntStackEntry*>(sp + 4);
+            auto exceptionType = static_cast<Wasm::ExceptionType>(callFrame->argumentCountIncludingThis());
+            if (Options::verboseWasmDebugger() && exceptionType == Wasm::ExceptionType::Unreachable)
+                displayWasmDebugState(instance, callee, stack, pl);
+            shouldThrow = debugServer.execution().handleBreakpointOrTrap(callFrame, instance, callee, pc, mc, pl, stack, exceptionType);
```
MenloDorian: Let's rename `handleBreakpointOrTrap` to `handleDebuggerTrapIsNeeded`.

Source/JavaScriptCore/wasm/debugger/WasmExecutionHandler.h
```diff
@@ -77,7 +77,7 @@ class ExecutionHandler {
     };
 
     ResumeMode stopCode(Locker<Lock>&, StopTheWorldEvent) WTF_REQUIRES_LOCK(m_lock);
-    bool handleUnreachable(CallFrame*, JSWebAssemblyInstance*, IPIntCallee*, uint8_t* pc, uint8_t* mc, IPInt::IPIntLocal* = nullptr, IPInt::IPIntStackEntry* = nullptr);
+    bool handleBreakpointOrTrap(CallFrame*, JSWebAssemblyInstance*, IPIntCallee*, uint8_t* pc, uint8_t* mc, IPInt::IPIntLocal*, IPInt::IPIntStackEntry*, Wasm::ExceptionType);
```
MenloDorian: Rename `handleBreakpointOrTrap` to `handleDebuggerTrapIfNeeded`.

Instead of returning a `bool` here, please add this:
```
enum class DebuggerTrapStatus : uint8_t {
    ResolvedByDebugger,
    NotResolvedByDebugger
};
```
... and have this function return  `DebuggerTrapStatus` instead.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -1166,6 +1174,25 @@ macro jumpToException()
     end
 end
 
+macro callIPIntTrapHandler()
```
MenloDorian: Please rename `callIPIntTrapHandler` to `handleDebuggerTrapIfNeeded`.

Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.cpp
```diff
@@ -1327,31 +1329,30 @@ static UNUSED_FUNCTION void displayWasmDebugState(JSWebAssemblyInstance* instanc
 }
 #endif
 
-WASM_IPINT_EXTERN_CPP_DECL(unreachable_handler, CallFrame* callFrame, Register* sp)
+WASM_IPINT_EXTERN_CPP_DECL(trap_handler, CallFrame* callFrame, Register* sp)
 {
-    bool breakpointHandled = false;
+    bool shouldThrow = true; // default: throw when debugger is not connected
```
MenloDorian: Instead of `bool shouldThrow`, let's use this instead:
```
// By default, we assume the trap is not a debugger only trap i.e. it's a Wasm trap, and the client needs to
// throw the Wasm trap.  If the debugger is connected, and determined that this trap its solely due to a
// debugger trap and has been handled by this function, it will change this status to DebuggerTrapStatus::ResolvedByDebugger.
DebuggerTrapStatus trapStatus = DebuggerTrapStatus::NotResolvedByDebugger;
```

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
```diff
@@ -396,7 +377,7 @@ ipintOp(_throw_ref, macro()
     jumpToException()
 
 .throw_null_ref:
-    throwException(NullExnrefReference)
+    ipintException(NullExnrefReference)
```
MenloDorian: Let's change all these `ipintException` to `handleDebuggerTrapIfNeededAndThrowWasmTrap` instead.  Same below.

Source/JavaScriptCore/llint/InPlaceInterpreter64.asm
```diff
@@ -235,25 +235,6 @@ end
...
-    # Remove pushed values
-    addq 4 * SlotSize, sp
-
-    bqeq r0, 0, .exception
-
-.continue:
-    nextIPIntInstruction()
-
-.exception:
     ipintException(Unreachable)
```
MenloDorian: Let's change this `ipintException` to `handleDebuggerTrapIfNeededAndThrowWasmTrap` instead.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -1196,6 +1223,16 @@ op(wasm_unwind_from_slow_path_trampoline, macro()
...
 
 op(wasm_throw_from_fault_handler_trampoline_reg_instance, macro ()
+    # enableWasmDebugger disables BBQ/OMG, so this trampoline is only
+    # reached from IPInt when the debugger is active. The signal handler only patches
+    # the machine PC, so IPInt registers (PC, MC, PL, ws0, cfr) are still live.
+    # Exception type comes from instance->m_exception; copy to CFR slot for trap_handler.
+    if ARM64 or ARM64E
+        loadi JSWebAssemblyInstance::m_exception[wasmInstance], t0
+        storei t0, ArgumentCountIncludingThis + PayloadOffset[cfr]
+        callIPIntTrapHandler()
```
MenloDorian: Let's change `callIPIntTrapHandler()` to `handleDebuggerTrapIfNeeded()`.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -1166,6 +1174,25 @@ macro jumpToException()
...
+    move cfr, a1
+    move sp, a2     # a2 = pointer to saved [PL, ws0, PC, MC]
+    operationCall(macro() cCall3(_ipint_extern_trap_handler) end)
+    addp 4 * MachineRegisterSize, sp
+end
+
+op(wasm_ipint_trap_handler_stub, macro ()
+    if ARM64 or ARM64E
+        callIPIntTrapHandler()
+        btpz r0, .continue  # zero (false = breakpoint): continue execution; non-zero (true = throw)
```
MenloDorian: nit: Can you replace this comment with a more elaborate one above instead e.g.
```
# r0 == 0 i.e. DebuggerTrapStatus::ResolvedByDebugger i.e. this was purely a debugger trap / breakpoint,
#              and has been handled.  We should continue executing because it's not a Wasm trap.
# r0 == 1 i.e. DebuggerTrapStatus::NotResolvedByDebugger i.e. this was a fatal Wasm trap.  We should
#              throw it to terminate Wasm execution.
```

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -1166,6 +1174,25 @@ macro jumpToException()
...
+    push PL, ws0    # sp[0]=PL, sp[1]=ws0 (IPIntCallee*), sp[2]=PC, sp[3]=MC
+    move cfr, a1
+    move sp, a2     # a2 = pointer to saved [PL, ws0, PC, MC]
+    operationCall(macro() cCall3(_ipint_extern_trap_handler) end)
+    addp 4 * MachineRegisterSize, sp
+end
+
+op(wasm_ipint_trap_handler_stub, macro ()
+    if ARM64 or ARM64E
+        callIPIntTrapHandler()
```
MenloDorian: Rename `callIPIntTrapHandler` to `handleDebuggerTrapIfNeeded`.

Source/JavaScriptCore/llint/InPlaceInterpreter.asm
```diff
@@ -284,7 +284,8 @@ if X86_64
 else
         move callee, a2
 end
-        cCall3(_ipint_extern_check_stack_and_vm_traps)
+        move cfr, a3
```
MenloDorian, line 287: I see that `cfr` has just been "pushed" at line 269 above with `subp cfr, scratch, scratch`.  AFAINT, the frame has not been initialized with any data yet before we get to this point.  Is it safe to pass it to `_ipint_extern_check_stack_and_vm_traps`?  What is `_`ipint_extern_check_stack_and_vm_traps`` doing with it?

It looks like you're passing it (in `setPrologueStopData`) to `StopData` to lldb.  If lldb reads from that CallFrame or tries to `bt` from it, you're going to have a bad time.  Should this be passing the caller CallFrame instead?

hyjorc1: The CFR is sufficiently initialized at this point. `preserveCallerPCAndCFR()` and `saveIPIntRegisters()` both run before `ipintEntry()` is called in `ipint_entry`, so `callerFrame()` and `cfr-8` (the outer call-site PC used by `collectCallStack`) are already valid. `collectCallStack ` never reads ArgumentCountIncludingThis, locals, PL, or the bytecode PC from this frame. The StopData's address comes from `callee->bytecode()` (the function's start address), not the PC. Passing the caller frame would drop the current function from the backtrace.

MenloDorian: Ah, indeed.  `op(ipint_entry...` does indeed initialize the CallFrame.  I only checked up to `macro ipintEntry()`.  Should have pealed back 1 more layer.

Source/JavaScriptCore/wasm/debugger/WasmDebugServerUtilities.h
```diff
@@ -192,26 +194,25 @@ struct DebugState {
...
-        stopData = makeUnique<StopData>(callee, instance, callFrame, pc, mc, locals, stack);
+        stopData = makeUnique<StopData>(callee, instance, callFrame, pc, mc, locals, stack, trapType);
     }
 
     bool atSystemCall() const { return !stopData; }
     bool atPrologue() const { return !!stopData && stopData->location == StopData::Location::Prologue; }
     bool atBreakpoint() const { return !!stopData && stopData->location == StopData::Location::Breakpoint; }
-    bool atTrap() const { return !!stopData && stopData->location == StopData::Location::Trap; }
-    bool atBreakpointOrTrap() const { return atBreakpoint() || atTrap(); }
+    bool isTrap() const { return !!stopData && stopData->trapType.has_value(); }
```
MenloDorian: This code deals with both debugger traps and Wasm traps.  In this case, I believe this one is for a Wasm trap.  Can you rename it to `isWasmTrap` to make it easier to reason about the code?

Source/JavaScriptCore/wasm/debugger/WasmDebugServerUtilities.h
```diff
@@ -192,26 +194,25 @@ struct DebugState {
...
+        stopData = makeUnique<StopData>(callee, instance, callFrame);
     }
 
     void setBreakpointStopData(Breakpoint::Type type, VirtualAddress address, uint8_t originalBytecode, uint8_t* pc, uint8_t* mc, IPInt::IPIntLocal* locals, IPInt::IPIntStackEntry* stack, IPIntCallee* callee, JSWebAssemblyInstance* instance, CallFrame* callFrame)
     {
         stopData = makeUnique<StopData>(type, address, originalBytecode, pc, mc, locals, stack, callee, instance, callFrame);
     }
 
-    void setTrapStopData(IPIntCallee* callee, JSWebAssemblyInstance* instance, CallFrame* callFrame, uint8_t* pc, uint8_t* mc, IPInt::IPIntLocal* locals, IPInt::IPIntStackEntry* stack)
+    void setTrapStopData(IPIntCallee* callee, JSWebAssemblyInstance* instance, CallFrame* callFrame, uint8_t* pc, uint8_t* mc, IPInt::IPIntLocal* locals, IPInt::IPIntStackEntry* stack, Wasm::ExceptionType trapType)
```
MenloDorian: Does the `Trap` in `setTrapStopData` refer to the `WasmTrap`?  If so, can you rename `setTrapStopData` accordingly to ``setWasmTrapStopData`?

nit: rename `trapType` to `wasmTrapType`.

Source/JavaScriptCore/wasm/debugger/WasmDebugServerUtilities.h
```diff
@@ -174,6 +175,7 @@ struct StopData {
     RefPtr<IPIntCallee> callee;
     JSWebAssemblyInstance* instance { nullptr };
     CallFrame* callFrame { nullptr };
+    std::optional<Wasm::ExceptionType> trapType;
```
MenloDorian: nit: rename to `wasmTrapType` mostly so that in places where it is used, it'll be clear that we're dealing with a Wasm trap.

Source/JavaScriptCore/wasm/debugger/WasmDebugServerUtilities.cpp
```diff
@@ -382,11 +383,11 @@ StopData::StopData(Breakpoint::Type type, VirtualAddress address, uint8_t origin
 {
 }
 
-StopData::StopData(IPIntCallee* callee, JSWebAssemblyInstance* instance, CallFrame* callFrame, uint8_t* pc, uint8_t* mc, IPInt::IPIntLocal* locals, IPInt::IPIntStackEntry* stack)
+StopData::StopData(IPIntCallee* callee, JSWebAssemblyInstance* instance, CallFrame* callFrame, uint8_t* pc, uint8_t* mc, IPInt::IPIntLocal* locals, IPInt::IPIntStackEntry* stack, Wasm::ExceptionType type)
     : StopData(Location::Trap, Code::Trap, VirtualAddress::toVirtual(instance, callee->functionIndex(), pc), 0, pc, mc, locals, stack, callee, instance, callFrame)
 {
+    trapType = type;
```
MenloDorian: Ditto.

Source/JavaScriptCore/wasm/debugger/WasmDebugServerUtilities.cpp
```diff
@@ -382,11 +383,11 @@ StopData::StopData(Breakpoint::Type type, VirtualAddress address, uint8_t origin
...
 }
-
```
MenloDorian, line 389: Let's undo this blank line removal.

MenloDorian (commented):
r=me with suggested improvements.

MenloDorian (approved):

PR #61750, opened 2026-03-31
[GTK4] TestWebKitAccessibility unit tests do not work

pgorszkowski-igalia:
#### 0a7326f424ae9058d353cd7dc7e89d131fc352ae
<pre>
[GTK4] TestWebKitAccessibility unit tests do not work
<a href="https://bugs.webkit.org/show_bug.cgi?id=311169">https://bugs.webkit.org/show_bug.cgi?id=311169</a>

Reviewed by NOBODY (OOPS!).

Fix TestWebKitAccessibility tests on GTK4 by replacing the AT-SPI registry
traversal approach with direct WebProcess root object resolution via GtkAtSpiSocket.

On GTK4, the headless (offscreen) GDK backend does not support AT-SPI
accessibility — GTK4 only registers with AT-SPI when a real Wayland or X11
surface is present. This means tests must run against a real compositor
(Weston) rather than a headless backend, and the AT-SPI tree is rooted
differently than on GTK3.

The new approach uses webkitWebViewBaseGetAccessibilitySocketForTesting() to
retrieve the GtkAtSpiSocket that was set up when the WebProcess registered its
AT-SPI plug. The socket&apos;s bus name and object path are passed directly to
_atspi_ref_accessible(), bypassing registry traversal entirely. No background
thread is needed because the D-Bus call targets the WebProcess, not the test
process itself.

A small number of tests (4) still time out or pass/fail. Their cause is unrelated
and will be investigated separately. These tests are currently marked as skipped
(to not slow down the API tests).

With this change, TestWebKitAccessibility API tests work on GTK4.

* Tools/Scripts/run-gtk-tests:
* Tools/TestWebKitAPI/Tests/WebKitGtk/TestWebKitAccessibility.cpp:
* Tools/TestWebKitAPI/glib/TestExpectations.json:
</pre>
https://github.com/WebKit/WebKit/commit/0a7326f424ae9058d353cd7dc7e89d131fc352ae

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/170166 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/44089 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/37505 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/179528 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/127878 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/172032 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/45333 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/43721 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/132652 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/93488 "2 flakes 12 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/173125 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/35838 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/153088 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/113154 "Passed tests") | 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/169487 "Passed tests") | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/34427 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/32788 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/28224 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/144910 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/32486 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/182125 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/34914 "Built successfully") | [❌ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/37131 "Found 1 new test failure: http/tests/site-isolation/https-load.html (failure)") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/141104 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/43746 "Built successfully") | | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/140929 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/44435 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/29470 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/108816 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/36200 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/116315 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/42945 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/7569 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/42337 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/42591 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/42480 "Built successfully") | | |

### DIFF
--- a/Source/WebKit/UIProcess/API/gtk/WebKitWebViewBase.cpp
+++ b/Source/WebKit/UIProcess/API/gtk/WebKitWebViewBase.cpp
@@ -3481,6 +3481,17 @@ void webkitWebViewBaseSetPlugID(WebKitWebViewBase* webViewBase, const String& pl
     UNUSED_PARAM(plugID);
 #endif // GTK_ACCESSIBILITY_ATSPI
 }
+
+GtkAtSpiSocket* webkitWebViewBaseGetAccessibilitySocketForTesting(WebKitWebViewBase* webViewBase)
+{
+#ifdef GTK_ACCESSIBILITY_ATSPI
+    auto* accessible = webViewBase->priv->socketAccessible.get();
+    return accessible ? GTK_AT_SPI_SOCKET(accessible) : nullptr;
+#else
+    UNUSED_PARAM(webViewBase);
+    return nullptr;
+#endif
+}
 #endif
 
 RendererBufferDescription webkitWebViewBaseGetRendererBufferDescription(WebKitWebViewBase* webViewBase)
--- a/Source/WebKit/UIProcess/API/gtk/WebKitWebViewBaseInternal.h
+++ b/Source/WebKit/UIProcess/API/gtk/WebKitWebViewBaseInternal.h
@@ -51,3 +51,7 @@ WK_EXPORT void webkitWebViewBaseSynthesizeWheelEvent(WebKitWebViewBase*, double 
 
 WK_EXPORT SkImage* webkitWebViewBaseSnapshotForTesting(WebKitWebViewBase*);
 
+#if USE(GTK4) && defined(GTK_ACCESSIBILITY_ATSPI)
+#include <gtk/a11y/gtkatspi.h>
+WK_EXPORT GtkAtSpiSocket* webkitWebViewBaseGetAccessibilitySocketForTesting(WebKitWebViewBase*);
+#endif
--- a/Tools/Scripts/run-gtk-tests
+++ b/Tools/Scripts/run-gtk-tests
@@ -62,13 +62,13 @@ class GtkTestRunner(TestRunner):
 
         return True
 
-    def _setup_testing_environment(self):
-        super(GtkTestRunner, self)._setup_testing_environment()
-
+    def _setup_testing_environment_for_driver(self, driver):
+        env = super(GtkTestRunner, self)._setup_testing_environment_for_driver(driver)
         # If we cannot start the accessibility daemons, we can just skip the accessibility tests.
         if not self._ensure_accessibility_service_is_running():
             print("Could not start accessibility bus, so disabling TestWebKitAccessibility")
             self._disabled_tests.append("WebKitGTK/TestWebKitAccessibility")
+        return env
 
     def _tear_down_testing_environment(self):
         super(GtkTestRunner, self)._tear_down_testing_environment()
--- a/Tools/Scripts/webkitpy/port/linux_container_sdk_utils.py
+++ b/Tools/Scripts/webkitpy/port/linux_container_sdk_utils.py
@@ -167,13 +167,18 @@ def _get_at_spi_bus_socket_or_dir_and_var(xdg):
         if m:
             path = _socket_or_none(m.group(1))
             if path:
-                return path, None
+                os.environ['AT_SPI_BUS_ADDRESS'] = 'unix:path={}'.format(path)
+                return path, 'AT_SPI_BUS_ADDRESS'
     except (FileNotFoundError, subprocess.TimeoutExpired, OSError):
         pass
 
     # 3 Fallback to usual directory
     xdg_dir = os.path.join(xdg, 'at-spi')
     if os.path.isdir(xdg_dir):
+        bus_socket = os.path.join(xdg_dir, 'bus')
+        if _socket_or_none(bus_socket):
+            os.environ['AT_SPI_BUS_ADDRESS'] = 'unix:path={}'.format(bus_socket)
+            return bus_socket, 'AT_SPI_BUS_ADDRESS'
         return xdg_dir, None
 
     return None, None
--- a/Tools/Scripts/webkitpy/port/linux_container_sdk_utils_unittest.py
+++ b/Tools/Scripts/webkitpy/port/linux_container_sdk_utils_unittest.py
@@ -126,7 +126,7 @@ class GetAtSpiBusDirTest(unittest.TestCase):
         with self._mock_gdbus("unix:path={},guid=abc".format(sock)):
             at_spi_path, at_spi_env_var = _get_at_spi_bus_socket_or_dir_and_var(self.tmpdir)
             self.assertEqual(at_spi_path, sock)
-            self.assertIsNone(at_spi_env_var)
+            self.assertEqual(at_spi_env_var, 'AT_SPI_BUS_ADDRESS')
 
     def test_gdbus_missing_socket_falls_through(self):
         with self._mock_gdbus("unix:path=/does/not/exist,guid=abc"):
--- a/Tools/TestWebKitAPI/Tests/WebKit/WKPage/gtk/TestWebKitAccessibility.cpp
+++ b/Tools/TestWebKitAPI/Tests/WebKit/WKPage/gtk/TestWebKitAccessibility.cpp
@@ -25,7 +25,11 @@
 // The libatspi headers don't use G_BEGIN_DECLS
 extern "C" {
 #include <atspi/atspi.h>
+// Private libatspi function to construct an AtspiAccessible from a known bus name and object path.
+AtspiAccessible* _atspi_ref_accessible(const gchar* app, const gchar* path);
 }
+
+#include <webkit/WebKitWebViewBaseInternal.h>
 
 struct AtspiEventDeleter {
     void operator()(AtspiEvent* event) const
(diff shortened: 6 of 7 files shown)


Tools/TestWebKitAPI/Tests/WebKitGtk/TestWebKitAccessibility.cpp
GeorgesStavracas: How about

```c
#if USE(GTK4) && defined(GTK_ACCESSIBILITY_ATSPI)
```

?

GeorgesStavracas, line 64: I understand what you've done here, and I think this will help with the test failures indeed, but I think there's still the risk of deadlocking since libatspi uses libdbus, and GTK/GLib uses GDBus. This risk is reduced by a big factor in this PR, but I don't think it's removed entirely.

GeorgesStavracas, line 3356: 🤔️

This is a little worrying, no? If it's generating warnings, something must be going wrong.

pgorszkowski-igalia: I agree, but we both know that the deadlock(libatspi uses libdbus, and GTK/GLib uses GDBus) is out of the scope of this. I mean we can't do anything with this here. The real fix should be in libdbus or GTK/GLib(its  accessibility mechanism implementation), right?

GeorgesStavracas: Yes. I'm mentioning this because I think we have to keep in mind that the tests might still deadlock, even with this patch.

pgorszkowski-igalia: ok, so my FATAL WARNINGS come from host applications which are registered in atspi. libatspi internally asks for the list of the application from desktop and that causes my FATAL WARNINGS. Probably it does not exist in case of CI. Example FATAL WARNINGS:

dbind-FATAL-WARNING: AT-SPI: Error in GetItems, sender=org.freedesktop.DBus, error=An AppArmor policy prevents this sender from sending this message to this recipient; type="method_call", sender=":1.687" (uid=1000 pid=1919973 comm="WebKitBuild/GTK/Release/bin/TestWebKitAPI/WebKitGT" label="crun (unconfined)") interface="org.a11y.atspi.Cache" member="GetItems" error name="(unset)" requested_reply="0" destination=":1.14" (uid=1000 pid=6923 comm="/snap/firefox/8054/usr/lib/firefox/firefox" label="snap.firefox.firefox (enforce)")

dbind-FATAL-WARNING: AT-SPI: Unable to open bus connection: Failed to connect to socket /run/user/1000/at-spi2-AOO8M3/socket: No such file or directory

The problem with the last one is that WKDEV does not bind(or create symlink from the /host/user...) at-spi sockets from the host.

Generally it is not a problem, I think, in case of CI, but if you run TestWebKitAccessibility locally in the container(wkdev) than all tests will fail.

csaavedra: > The problem with the last one is that WKDEV does not bind(or create symlink from the /host/user...) at-spi sockets from the host.

This could be fixed in the sdk, no?

pgorszkowski-igalia: Yes and no, when you start the wkdev it can create a sym links, but after that you can run some application outside the container, and that socket will not be symlinked so the problem will still exist.

pgorszkowski-igalia: in theory we can use "ATSPI_IN_TESTS" from https://gitlab.gnome.org/GNOME/at-spi2-core/-/commit/06997ae685055a0778a801651ee9e435cf62c799 to hide it. 

And actually I can try to upstream to https://gitlab.gnome.org/GNOME/at-spi2-core also using "ATSPI_IN_TESTS" in case of "dbind-FATAL-WARNING: AT-SPI: Error in GetItems"

mgorse: Fyi, I added a new function, atspi_set_g_main_loop_reentrancy, that can be used to make libatspi's DBus calls run a GMainLoop. This will not be the default behavior, but it helps for me with a couple of the deadlocks in the webKit tests, although it would requiring the up-coming GNOME 51 version of at-spi2-core.

mgorse: Libatspi already has ATSPI_NO_CACHE, intended to disable caching, and I just noticed that the code calls GetItems whether it is set or not, although there is no reason to do so if that variable is set. For the upcoming release, I just added a check so that it won't call GetItems anymore if ATSPI_NO_CACHE is set. I could apply the change to a stable release if it would help you (I'm not sure what version of GNOME the CI is loading), if you would like to disable caching for the tests. I'm not sure if I want to make ATSPI_IN_TESTS disable the GetItems warning--it looks specific to the WebKit environment--but I might be able to be convinced, or I could perhaps add a separate environment variable to control it. I'd like to get this resolved one way or the other so that we can get this merged (I have https://github.com/WebKit/WebKit/pull/62762, but my new tests are failing because we need this PR).

pgorszkowski-igalia: Hi @mgorse , thanks for your suggestion. Actually the problem with: `dbind-FATAL-WARNING: AT-SPI: Error in GetItems` I am observing only in my setup, this should not be visible on CI setup.

This case:
`dbind-FATAL-WARNING: AT-SPI: Unable to open bus connection: Failed to connect to socket /run/user/1000/at-spi2-AOO8M3/socket: No such file or directory`
is also no-CI setup, it happens only if I run these tests locally with WebKit SDK container.

@carlosgcampos : can you take a look and review it, thanks?

pgorszkowski-igalia: There are some regressions after rebasing to main. I need to check that

GeorgesStavracas (approved):
For the lack of better approaches, I think this is fine. We all understand the limitations of this patch so nothing should come as a surprise.

pgorszkowski-igalia:
> For the lack of better approaches, I think this is fine. We all understand the limitations of this patch so nothing should come as a surprise.

Thanks Georges for your review.

There is something wrong with the newer SDK and my change, that is why we see failings in api-gtk. I have to check that. It works with older SDK (dcc0e8790d9b0265df39fc11c6dace198f0539ed2c102bad95f999362cec2afd).

nikolaszimmermann:
Worth to CC @clopez - maybe related to the podman-in-pod changes?

@pgorszkowski-igalia pr does it even fail locally for you?

pgorszkowski-igalia:
> Worth to CC @clopez - maybe related to the podman-in-pod changes?
> 
> @pgorszkowski-igalia pr does it even fail locally for you?

@nikolaszimmermann on my local SDK, the tests with this PR work (the tests pass):

my local SDK:
```
WebKit SDK container version: 2.53-v6-d535e88

Git repository 'wkdev-sdk' ( main @ be706e12c64766031de783b11219026c2c2b93c6 )
 🠲  Last commit: Sat May 30 00:04:51 2026 +0200
```

The last logs from all tests:
```
Unexpected crashes (12)
    /WebKitGTK/TestResources
        /webkit/WebKitWebView/resources
        /webkit/WebKitWebView/history-cache
        /webkit/WebKitWebView/sync-request-on-max-conns
        /webkit/WebKitWebResource/loading
        /webkit/WebKitWebResource/response
        /webkit/WebKitWebResource/mime-type
        /webkit/WebKitWebResource/suggested-filename
        /webkit/WebKitWebResource/active-uri
        /webkit/WebKitWebResource/get-data
        /webkit/WebKitWebResource/get-data-error
        /webkit/WebKitWebResource/get-data-empty
        /webkit/WebKitWebPage/send-request

Unexpected passes (1)
    /WebKitGTK/TestWebKitWebView
        /webkit/WebKitWebView/is-playing-audio

Ran 2891 tests of 2897 with 2879 successful
```

nikolaszimmermann:
So I guess it's related to AT-SPI setup on the bots...

PR #61775, opened 2026-03-31
[CSS JIT] Compile :has() argument selectors

nullhook:
#### be1f27dcb8c0a43fe001fe2836e344c78bda3d06
<pre>
[CSS JIT] Compile :has() argument selectors
<a href="https://bugs.webkit.org/show_bug.cgi?id=311203">https://bugs.webkit.org/show_bug.cgi?id=311203</a>
<a href="https://rdar.apple.com/problem/173791568">rdar://problem/173791568</a>

Reviewed by Antti Koivisto.

When :has() is evaluated, the argument selector (e.g., .active in :has(&gt; .active))
is checked per-element via the SelectorChecker interpreter. This patch JIT-compiles
the argument selector so each per-element check runs as native code.

Argument selectors with adjacent/sibling combinators (+, ~) or functional
pseudo-classes (:not, :is) fall back to the interpreter because the CSS JIT
doesn&apos;t generate all the style relations needed for :has() invalidation with
sibling relationships. This can be addressed in a follow-up.

* Source/WebCore/css/SelectorChecker.cpp:
(WebCore::SelectorChecker::checkOne const):
(WebCore::compiledHasArgumentSelectorsMap):
(WebCore::SelectorChecker::clearCompiledHasArgumentSelectors):
(WebCore::canJITCompileHasArgument):
(WebCore::SelectorChecker::matchHasPseudoClass const):
(WebCore::SelectorChecker::matchHasArgumentSelector const):
* Source/WebCore/css/SelectorChecker.h:
(WebCore::SelectorCompiler::compileSelector):
(WebCore::SelectorCompiler::SelectorCodeGenerator::SelectorCodeGenerator):
(WebCore::SelectorCompiler::constructFragmentsInternal):
(WebCore::SelectorCompiler::constructFragments):
(WebCore::SelectorCompiler::SelectorCodeGenerator::generateElementMatching):
(WebCore::SelectorCompiler::SelectorCodeGenerator::generateElementMatchesHasScope):
* Source/WebCore/cssjit/SelectorCompiler.h:
* Source/WebCore/page/MemoryRelease.cpp:
(WebCore::releaseNoncriticalMemory):

Canonical link: <a href="https://commits.webkit.org/312288@main">https://commits.webkit.org/312288@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/b5f76b6fc5685bde0b36f4029f35ab98819dcd89

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/159303 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/32731 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/25836 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/168133 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/113681 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/161172 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/32799 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/32718 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/123426 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/86639 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/162260 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/25686 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/143108 "Passed tests") | [❌ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/104093 "Found 2 new API test failures: TestWebCore:IntSize.ConstrainedBetween, TestWebCore:AffineTransform.ValueConstruction (failure)") | | ⏳ 🛠 vision-apple 
| | [❌ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/24739 "Found 1 new test failure: imported/w3c/web-platform-tests/navigation-api/navigate-event/signal-abort-window-stop.html (failure)") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/23176 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/15906 "Built successfully") | | 
| [⏳ 🛠 🧪 jsc ](https://ews-build.webkit.org/#/builders/JSC-Tests-EWS "Waiting in queue, processing has not started yet") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/134426 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/20888 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/170627 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/16661 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/22514 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/131626 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/32420 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/27248 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/131739 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35662 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/32364 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/142681 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/90489 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/26421 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/19490 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/31875 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/98327 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/31395 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/31668 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/31550 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/css/SelectorChecker.cpp
+++ b/Source/WebCore/css/SelectorChecker.cpp
@@ -47,6 +47,8 @@
 #include "RenderElement.h"
 #include "RuleFeature.h"
 #include "SelectorCheckerTestFunctions.h"
+#include "SelectorCompiler.h"
+#include "Settings.h"
 #include "ShadowRoot.h"
 #include "StyleRule.h"
 #include "StyleScope.h"
@@ -54,6 +56,7 @@
 #include "TypedElementDescendantIteratorInlines.h"
 #include "ViewTransition.h"
 #include "ViewTransitionTypeSet.h"
+#include <wtf/NeverDestroyed.h>
 
 namespace WebCore {
 
@@ -984,13 +987,8 @@ bool SelectorChecker::checkOne(CheckingContext& checkingContext, LocalContext& c
                     matchType = localMatchType;
                 return hasMatchedAnything;
             }
-        case CSSSelector::PseudoClass::Has: {
-            for (auto& hasSelector : *selector.selectorList()) {
-                if (matchHasPseudoClass(checkingContext, element, hasSelector))
-                    return true;
-            }
-            return false;
-        }
+        case CSSSelector::PseudoClass::Has:
+            return matchHasPseudoClass(checkingContext, element, *selector.selectorList());
         case CSSSelector::PseudoClass::PlaceholderShown:
             if (auto* formControl = dynamicDowncast<HTMLTextFormControlElement>(element.get()))
                 return formControl->isPlaceholderVisible();
@@ -1432,7 +1430,51 @@ bool SelectorChecker::matchSelectorList(CheckingContext& checkingContext, const 
     return false;
 }
 
-bool SelectorChecker::matchHasPseudoClass(CheckingContext& checkingContext, const Element& element, const CSSSelector& hasSelector) const
+#if ENABLE(CSS_SELECTOR_JIT)
+static constexpr auto maximumCompiledHasArgumentSelectorsSize = 1024u;
+
+static HashMap<const CSSSelectorList*, FixedVector<CompiledSelector>>& NODELETE compiledHasArgumentSelectorsMap()
+{
+    static NeverDestroyed<HashMap<const CSSSelectorList*, FixedVector<CompiledSelector>>> map;
+    return map;
+}
+#endif
+
+void SelectorChecker::clearCompiledHasArgumentSelectors()
+{
+#if ENABLE(CSS_SELECTOR_JIT)
+    compiledHasArgumentSelectorsMap().clear();
+#endif
+}
+
+#if ENABLE(CSS_SELECTOR_JIT)
+// FIXME(https://bugs.webkit.org/show_bug.cgi?id=313164): The JIT doesn't generate the full set
+// of style relations needed for :has() invalidation with adjacent/sibling combinators. Skip JIT
+// for argument selectors containing these, or functional pseudo-classes that may nest them.
+static bool canJITCompileHasArgument(const CSSSelector& selector)
+{
+    for (auto* current = &selector; current; current = current->precedingInComplexSelector()) {
+        auto relation = current->relation();
+        if (relation == CSSSelector::Relation::DirectAdjacent || relation == CSSSelector::Relation::IndirectAdjacent)
+            return false;
+
+        if (current->match() == CSSSelector::Match::PseudoClass) {
+            switch (current->pseudoClass()) {
+            case CSSSelector::PseudoClass::Not:
+            case CSSSelector::PseudoClass::Is:
+            case CSSSelector::PseudoClass::Where:
+            case CSSSelector::PseudoClass::WebKitAny:
+                return false;
+            default:
+                break;
+            }
+        }
+    }
+    return true;
+}
+#endif
+
+bool SelectorChecker::matchHasPseudoClass(CheckingContext& checkingContext, const Element& element, const CSSSelectorList& selectorList) const
 {
     // :has() should never be nested with another :has()
     // This is generally discarded at parsing time, but
@@ -1440,6 +1482,37 @@ bool SelectorChecker::matchHasPseudoClass(CheckingContext& checkingContext, cons
     if (checkingContext.disallowHasPseudoClass)
         return false;
 
+#if ENABLE(CSS_SELECTOR_JIT)
+    FixedVector<CompiledSelector>* compiledSelectors = nullptr;
+    if (element.document().settings().cssSelectorJITCompilerEnabled()) {
+        auto& map = compiledHasArgumentSelectorsMap();
+        if (map.size() >= maximumCompiledHasArgumentSelectorsSize)
+            map.remove(map.random());
+        compiledSelectors = &map.ensure(&selectorList, [&] {
+            return FixedVector<CompiledSelector>(selectorList.size());
+        }).iterator->value;
+    }
+#endif
+
+#if ENABLE(CSS_SELECTOR_JIT)
+    unsigned argIndex = 0;
+#endif
+    for (auto& hasSelector : selectorList) {
+        CompiledSelector* compiled = nullptr;
+#if ENABLE(CSS_SELECTOR_JIT)
+        if (compiledSelectors)
+            compiled = &(*compiledSelectors)[argIndex];
+        ++argIndex;
+#endif
+        if (matchHasArgumentSelector(checkingContext, element, hasSelector, compiled))
+            return true;
+    }
+    return false;
+}
+
+bool SelectorChecker::matchHasArgumentSelector(CheckingContext& checkingContext, const Element& element, const CSSSelector& hasSelector, CompiledSelector* compiledSelector) const
+{
+    UNUSED_PARAM(compiledSelector);
     auto matchElement = Style::computeHasArgumentRelation(hasSelector);
 
     enum class HasTraversalType : uint8_t { Children, Descendants, Siblings, SiblingDescendants };
@@ -1539,6 +1612,25 @@ bool SelectorChecker::matchHasPseudoClass(CheckingContext& checkingContext, cons
     bool matchedInsideScope = false;
 
     auto checkRelative = [&](auto& elementToCheck) {
+#if ENABLE(CSS_SELECTOR_JIT)
+        if (compiledSelector) {
+            if (compiledSelector->status == SelectorCompilationStatus::NotCompiled) {
+                if (canJITCompileHasArgument(hasSelector))
+                    SelectorCompiler::compileSelector(*compiledSelector, hasSelector, SelectorCompiler::SelectorContext::RuleCollector, SelectorCompiler::SelectorPurpose::HasArgument);
+                else
+                    compiledSelector->status = SelectorCompilationStatus::CannotCompile;
+            }
+            if (compiledSelector->status == SelectorCompilationStatus::SelectorCheckerWithCheckingContext) {
+                const Element& elementRef = elementToCheck;
+                compiledSelector->wasUsed();
+                unsigned ignored = 0;
+                bool result = SelectorCompiler::ruleCollectorSelectorCheckerWithCheckingContext(*compiledSelector, &elementRef, &hasCheckingContext, &ignored);
+                if (hasCheckingContext.matchedInsideScope)
+                    matchedInsideScope = true;
+                return result;
+            }
+        }
+#endif
         LocalContext hasContext(hasSelector, elementToCheck, VisitedMatchType::Disabled, std::nullopt);
         hasContext.inFunctionalPseudoClass = true;
         hasContext.pseudoElementEffective = false;
--- a/Source/WebCore/cssjit/SelectorCompiler.cpp
+++ b/Source/WebCore/cssjit/SelectorCompiler.cpp
@@ -467,6 +467,8 @@ struct SelectorFragment {
     Vector<Vector<SelectorFragment>> anyFilters;
     const CSSSelector* pseudoElementSelector = nullptr;
 
+    bool matchesHasScope { false };
+
     // For quirks mode, follow this: http://quirks.spec.whatwg.org/#the-:active-and-:hover-quirk
     // In quirks mode, a compound selector 'selector' that matches the following conditions must not match elements that would not also match the ':any-link' selector.
     //
@@ -511,7 +513,7 @@ struct BacktrackingLevel {
 
 class SelectorCodeGenerator {
 public:
-    SelectorCodeGenerator(const CSSSelector&, SelectorContext);
+    SelectorCodeGenerator(const CSSSelector&, SelectorContext, SelectorPurpose = SelectorPurpose::Normal);
     SelectorCompilationStatus compile(JSC::MacroAssemblerCodeRef<JSC::CSSSelectorPtrTag>&);
 
 private:
@@ -580,6 +582,7 @@ private:
     void generateElementHasPseudoElement(Assembler::JumpList& failureCases, const SelectorFragment&);
     void generateElementIsRoot(Assembler::JumpList& failureCases);
     void generateElementIsScopeRoot(Assembler::JumpList& failureCases);
+    void generateElementMatchesHasScope(Assembler::JumpList& failureCases);
     void generateElementIsTarget(Assembler::JumpList& failureCases);
     void generateElementAndDocumentIsHTML(Assembler::JumpList& failureCases);
 
@@ -630,6 +633,7 @@ private:
     StackAllocator::StackReference m_lastVisitedElement;
     StackAllocator::StackReference m_startElement;
 
+    SelectorPurpose m_purpose;
     const CSSSelector& m_originalSelector;
 };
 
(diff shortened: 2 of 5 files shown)


Source/WebCore/css/CSSSelectorList.h
anttijk: Putting compiled selectors to CSSSelectorList is not really correct because in non-`:has()` cases argument selector lists are compiled into the top level selector so this just ends up wasting space.

There should be some other storage scheme for these. You could have a static `CSSSelectorList` -> `FixedVector<CompiledSelector>` map for example, CSSSelectorLists are hashable.

Source/WebCore/cssjit/SelectorCompiler.cpp
anttijk: Could HasArgument be another value in SelectorContext?

anttijk (changes requested):

Source/WebCore/cssjit/SelectorCompiler.cpp
nullhook: isn't `SelectorContext` about where the JIT is being invoked from? increasing the state space with `HasArgument` seems odd here

Source/WebCore/css/CSSSelectorList.h
nullhook: ah, good catch. i've made this storage singleton

Source/WebCore/css/SelectorChecker.cpp
```diff
@@ -985,9 +987,12 @@ bool SelectorChecker::checkOne(CheckingContext& checkingContext, LocalContext& c
                 return hasMatchedAnything;
             }
         case CSSSelector::PseudoClass::Has: {
-            for (auto& hasSelector : *selector.selectorList()) {
-                if (matchHasPseudoClass(checkingContext, element, hasSelector))
+            auto* selectorList = selector.selectorList();
+            unsigned argIndex = 0;
+            for (auto& hasSelector : *selectorList) {
+                if (matchHasPseudoClass(checkingContext, element, hasSelector, selectorList, argIndex))
                     return true;
```
anttijk: This seems like an awkward way to factor this. You could just pass selectorList and the index. Or even better move the entire loop in `matchHasPseudoClass`. This is the only caller so you can easily arrange things as you like.

Source/WebCore/css/SelectorChecker.cpp
```diff
@@ -1432,7 +1437,7 @@ bool SelectorChecker::matchSelectorList(CheckingContext& checkingContext, const
     return false;
 }
 
-bool SelectorChecker::matchHasPseudoClass(CheckingContext& checkingContext, const Element& element, const CSSSelector& hasSelector) const
+bool SelectorChecker::matchHasPseudoClass(CheckingContext& checkingContext, const Element& element, const CSSSelector& hasSelector, const CSSSelectorList* selectorList, unsigned argIndex) const
```
anttijk: selectorList is never nullptr so should be a reference

Source/WebCore/css/SelectorChecker.cpp
```diff
@@ -1511,6 +1516,25 @@ bool SelectorChecker::matchHasPseudoClass(CheckingContext& checkingContext, cons
     bool matchedInsideScope = false;
 
     auto checkRelative = [&](auto& elementToCheck) {
+        const Element& elementRef = elementToCheck;
+#if ENABLE(CSS_SELECTOR_JIT)
+        if (selectorList && element.document().settings().cssSelectorJITCompilerEnabled()) {
+            auto& compiled = SelectorCompiler::compiledHasArgumentSelectorForIndex(*selectorList, argIndex);
```
anttijk: because the looping is outside the function you'll make a hash lookup for every item on the list instead of just one per selectorList.

Source/WebCore/cssjit/SelectorCompiler.cpp
```diff
@@ -667,6 +673,41 @@ void compileSelector(CompiledSelector& compiledSelector, const CSSSelector& sele
...
+
+    SelectorCodeGenerator codeGenerator(selector, SelectorContext::RuleCollector, true /* isHasArgument */);
+    compiledSelector.status = codeGenerator.compile(compiledSelector.codeRef);
+
+#if defined(CSS_SELECTOR_JIT_PROFILING) && CSS_SELECTOR_JIT_PROFILING
+    compiledSelector.selector = &selector;
+#endif
+
+    ASSERT(compiledSelector.status != SelectorCompilationStatus::NotCompiled);
+}
```
anttijk: Why does this need a new compilation path? Doesn't `compileSelector` with an extra argument work?

Source/WebCore/cssjit/SelectorCompiler.cpp
```diff
@@ -667,6 +673,41 @@ void compileSelector(CompiledSelector& compiledSelector, const CSSSelector& sele
...
+
+CompiledSelector& compiledHasArgumentSelectorForIndex(const CSSSelectorList& selectorList, unsigned index)
+{
+    auto& map = compiledHasArgumentSelectorsMap();
+    auto& selectors = map.ensure(&selectorList, [&] {
+        return FixedVector<CompiledSelector>(selectorList.size());
+    }).iterator->value;
+    ASSERT(index < selectors.size());
+    return selectors[index];
+}
```
anttijk: This stuff does not belong to the `SelectorCompiler` as it is not about compilation. It should be in the client.

Source/WebCore/cssjit/SelectorCompiler.cpp
```diff
@@ -1478,22 +1519,24 @@ static inline FunctionType addPseudoClassType(const CSSSelector& selector, Selec
     return FunctionType::CannotCompile;
 }
 
-inline SelectorCodeGenerator::SelectorCodeGenerator(const CSSSelector& rootSelector, SelectorContext selectorContext)
+inline SelectorCodeGenerator::SelectorCodeGenerator(const CSSSelector& rootSelector, SelectorContext selectorContext, bool isHasArgument)
```
anttijk: The bool argument should be an enum class so it is clear from the call sites what is being passed.

anttijk: SelectorContext and has mode could probably be combined into a single EnumSet<> so no new arguments would be needed.

anttijk (changes requested):

Source/WebCore/cssjit/SelectorCompiler.cpp
nullhook: you're right, i've removed this.

Source/WebCore/cssjit/SelectorCompiler.cpp
```diff
@@ -644,20 +648,21 @@ enum class FragmentsLevel {
...
 
     if (!JSC::Options::useJIT()) {
         compiledSelector.status = SelectorCompilationStatus::CannotCompile;
         return;
     }
-    
-    SelectorCodeGenerator codeGenerator(selector, selectorContext);
+
+    bool isHasArgument = purpose == SelectorPurpose::HasArgument;
+    SelectorCodeGenerator codeGenerator(selector, selectorContext, isHasArgument);
```
anttijk: can you just pass SelectorPurpose itself instead of a bool?

Source/WebCore/css/SelectorChecker.cpp
```diff
@@ -1432,14 +1430,48 @@ bool SelectorChecker::matchSelectorList(CheckingContext& checkingContext, const
...
 }
 
-bool SelectorChecker::matchHasPseudoClass(CheckingContext& checkingContext, const Element& element, const CSSSelector& hasSelector) const
+#if ENABLE(CSS_SELECTOR_JIT)
+static HashMap<const CSSSelectorList*, FixedVector<CompiledSelector>>& NODELETE compiledHasArgumentSelectorsMap()
+{
+    static NeverDestroyed<HashMap<const CSSSelectorList*, FixedVector<CompiledSelector>>> map;
+    return map;
+}
+#endif
```
anttijk, line 1441: To prevent this from growing without bounds there should be some sort of eviction scheme. We usually use random eviction for simplicity. It should also be connected to the low memory handler. See for example `deduplicationMap()` in ImmutableStyleProperties.cpp how to do these things.

anttijk (changes requested):

Source/WebCore/css/SelectorChecker.cpp
nullhook: the lifetime for this map is global but should it be tied to page/document? alternatively, should we be clearing this map on page/document tear down?

Source/WebCore/style/StyleScope.cpp
```diff
@@ -195,6 +196,9 @@ void Scope::clearResolver()
     m_resolver = nullptr;
     customPropertyRegistry().clearRegisteredFromStylesheets();
     counterStyleRegistry().clearAuthorCounterStyles();
+
+    if (!m_shadowRoot)
+        SelectorChecker::clearCompiledHasArgumentSelectors();
```
nullhook: @anttijk presumably, this is the right place to clear the map. fixes all flakiness, locally at least.

nullhook:
i'm curious why `:has` with sibling combinator tests are failing on EWS but not locally. 

cc @anttijk any ideas?

edit: I ran failed sibling combinator tests in isolation on a EWS machine and it fails deterministically. so, it isn't test ordering. I'm looking into this more

nullhook:
> i'm curious why `:has` with sibling combinator tests are failing on EWS but not locally.
> 
> cc @anttijk any ideas?
> 
> edit: I ran failed sibling combinator tests in isolation on a EWS machine and it fails deterministically. so, it isn't test ordering. I'm looking into this more

This happens because JIT cannot generate full style relations with sibling relationships. I've fallback to interpreter for these cases now.

Source/WebCore/css/SelectorChecker.cpp
anttijk: Neither is necessary. Compiled selectors are globally shareable.

Source/WebCore/css/SelectorChecker.cpp
```diff
@@ -1432,14 +1430,92 @@ bool SelectorChecker::matchSelectorList(CheckingContext& checkingContext, const
...
+        auto& map = compiledHasArgumentSelectorsMap();
+        if (map.size() >= maximumCompiledHasArgumentSelectorsSize)
+            map.remove(map.random());
+        auto result = map.ensure(&selectorList, [&] {
+            return FixedVector<CompiledSelector>(selectorList.size());
+        });
+        auto& selectors = result.iterator->value;
+        // Validate the entry matches this selectorList. A stale entry from a
+        // deallocated CSSSelectorList at the same address may have a different size.
+        if (!result.isNewEntry && selectors.size() != selectorList.size())
```
anttijk: I don't understand this. How would there be a stale entry?  And would just size comparison be enough to detect it?

Source/WebCore/style/StyleScope.cpp
anttijk: You should clear it from `releaseNoncriticalMemory()` in MemoryRelease.cpp.

Source/WebCore/cssjit/SelectorCompiler.cpp
```diff
@@ -1639,6 +1647,11 @@ static FunctionType constructFragmentsInternal(const CSSSelector& rootSelector,
         case CSSSelector::Match::ForgivingUnknown:
         case CSSSelector::Match::ForgivingUnknownNestContaining:
         case CSSSelector::Match::HasScope:
+            if (isHasArgument) {
```
anttijk: This likely should not be affecting ForgivingUnknown/ForgivingUnknownNestContaining whatever those are.

Source/WebCore/cssjit/SelectorCompiler.cpp
```diff
@@ -1478,22 +1482,26 @@ static inline FunctionType addPseudoClassType(const CSSSelector& selector, Selec
...
     , m_visitedMode(VisitedMode::None)
     , m_descendantBacktrackingStartInUse(false)
+    , m_purpose(purpose)
     , m_originalSelector(rootSelector)
 {
     auto selectorTextUTF8 = m_originalSelector.selectorText().utf8();
     auto selectorTextSpan = selectorTextUTF8.span();
     dataLogFIf(shouldDumpCSSJITDisassembly(), "Compiling \"%.*s\"\n", static_cast<int>(selectorTextSpan.size()), selectorTextSpan.data());
 
+    bool isHasArgument = m_purpose == SelectorPurpose::HasArgument;
```
anttijk: why not just pass m_purpose?

Source/WebCore/css/SelectorChecker.cpp
```diff
@@ -1539,6 +1615,21 @@ bool SelectorChecker::matchHasPseudoClass(CheckingContext& checkingContext, cons
     bool matchedInsideScope = false;
 
     auto checkRelative = [&](auto& elementToCheck) {
+        const Element& elementRef = elementToCheck;
+#if ENABLE(CSS_SELECTOR_JIT)
```
anttijk: why this variable? also I think it would fail to compile if CSS_SELECTOR_JIT was disabled (because unused variable).

Source/WebCore/css/SelectorChecker.cpp
```diff
@@ -1432,14 +1430,92 @@ bool SelectorChecker::matchSelectorList(CheckingContext& checkingContext, const
...
+    }
+#endif
+
+    unsigned argIndex = 0;
+    for (auto& hasSelector : selectorList) {
+        CompiledSelector* compiled = nullptr;
+#if ENABLE(CSS_SELECTOR_JIT)
+        if (compiledSelectors && canJITCompileHasArgument(hasSelector))
+            compiled = &(*compiledSelectors)[argIndex];
+#endif
```
anttijk, line 1506: It is bit weird `canJITCompileHasArgument` and the actual compilation are in different places. Maybe youn you just compile here already and provide nullptr to matchHasArgumentSelector if it fails. Or move this check to where compilation is.

You could set SelectorCompilationStatus::CannotCompile in cache in this case too so you don't need to check repeatedly.

Source/WebCore/css/SelectorChecker.cpp
nullhook: this was added to potentially fix cross-test flakiness, but the real cause was different. i'll revert this change

Source/WebCore/cssjit/SelectorCompiler.cpp
nullhook: good catch

Source/WebCore/css/SelectorChecker.cpp
nullhook: moved

anttijk (approved):
Those watch and playstation build failures appear to be real. Please fix them before landing.

PR #61851, opened 2026-04-01
[Site Isolation] Console log is intentionally missing the full URL of a RemoteFrame for http/tests/security/block-top-level-navigations-by-sandboxed-iframe-with-propagated-user-gesture.html

atar13:
#### 80a2f83fa96b35052b2308ee781478d9dd9677bc
<pre>
[Site Isolation] Console log is intentionally missing the full URL of a RemoteFrame for http/tests/security/block-top-level-navigations-by-sandboxed-iframe-with-propagated-user-gesture.html
<a href="https://bugs.webkit.org/show_bug.cgi?id=311286">https://bugs.webkit.org/show_bug.cgi?id=311286</a>
<a href="https://rdar.apple.com/173883912">rdar://173883912</a>

Reviewed by Sihui Liu.

In http/tests/security/block-top-level-navigations-by-sandboxed-iframe-with-propagated-user-gesture.html,
with site isolation enabled the console message logging
&quot;Unsafe JavaScript attempt to initiate navigation&quot; will
not contain the full URL (including resource path) of RemoteFrames.

This is the diff:
```
@@ -1,4 +1,4 @@
-CONSOLE MESSAGE: Unsafe JavaScript attempt to initiate navigation for frame with URL &apos;<a href="http://127.0.0.1">http://127.0.0.1</a>:8000/security/block-top-level-navigations-by-sandboxed-iframe-with-propagated-user-gesture.html&apos; from frame with URL &apos;<a href="http://localhost">http://localhost</a>:8000/security/resources/navigate-top-level-frame-to-failure-page-via-message-handler.html&apos;. The frame attempting navigation of the top-level window is sandboxed, but the &apos;allow-top-navigation&apos; flag is not set.
+CONSOLE MESSAGE: Unsafe JavaScript attempt to initiate navigation for frame with URL &apos;<a href="http://127.0.0.1">http://127.0.0.1</a>:8000/&apos; from frame with URL &apos;<a href="http://localhost">http://localhost</a>:8000/security/resources/navigate-top-level-frame-to-failure-page-via-message-handler.html&apos;. The frame attempting navigation of the top-level window is sandboxed, but the &apos;allow-top-navigation&apos; flag is not set.

 CONSOLE MESSAGE: SecurityError: The operation is insecure.
 Test blocking of top-level navigations by an iframe with `sandbox=allow-top-navigation-by-user-activation` when the user gesture is propagated from another context.
```

This difference is thanks to <a href="https://commits.webkit.org/310093@main">https://commits.webkit.org/310093@main</a>
where we decided to prevent frames from fetching the full URL path of
remote frames in different processes.
http/tests/security/block-top-level-navigations-by-sandboxed-iframe-with-propagated-user-gesture.html
also has the same inconsistency in behavior with site isolation enabled vs disabled.

This patch moves the failing line in TestExpectations with the other intentional
failures from <a href="https://commits.webkit.org/310093@main">https://commits.webkit.org/310093@main</a>

* LayoutTests/platform/ios-site-isolation/TestExpectations:

Canonical link: <a href="https://commits.webkit.org/310523@main">https://commits.webkit.org/310523@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/4e757cb11dcdbfd963840c137f399252f81823f1

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/153749 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/26533 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/20150 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/162500 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/107209 "Built successfully") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/27061 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/26855 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/118874 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/84057 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/156708 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/21134 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/138058 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/99584 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/20213 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/18171 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/10332 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/129862 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/15917 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/164970 "Built successfully") | | 
| | | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/17511 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/126951 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/26330 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/22202 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/127118 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/34569 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/26332 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/137712 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/83010 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/22029 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/14494 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/25949 "Built successfully") | | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/25640 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/25800 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/25700 "Built successfully") | | | |

### DIFF
--- a/LayoutTests/platform/mac-site-isolation/TestExpectations
+++ b/LayoutTests/platform/mac-site-isolation/TestExpectations
@@ -249,7 +249,6 @@ http/tests/security/XFrameOptions/x-frame-options-ancestors-same-origin-deny.htm
 http/tests/security/XFrameOptions/x-frame-options-ignore-deny-meta-tag-parent-same-origin-deny.html [ Failure ]
 http/tests/security/XFrameOptions/x-frame-options-multiple-headers-sameorigin-deny.html [ Failure ]
 http/tests/security/XFrameOptions/x-frame-options-parent-same-origin-deny.html [ Failure ]
-http/tests/security/block-top-level-navigations-by-sandboxed-iframe-with-propagated-user-gesture.html [ Failure ]
 http/tests/security/block-top-level-navigations-by-third-party-sandboxed-iframe.html [ Failure ]
 http/tests/security/contentSecurityPolicy/embed-redirect-allowed.html [ Failure ]
 http/tests/security/contentSecurityPolicy/embed-redirect-allowed2.html [ Failure ]
@@ -520,12 +519,13 @@ imported/w3c/web-platform-tests/cookies/third-party-cookies/third-party-cookies.
 ####################################
 # Tests with intentional text fail #
 ####################################
-# The following two tests have a different console log with Site Isolation enabled
+# The following tests have a different console log with Site Isolation enabled
 # vs disabled. The difference is that with Site Isolation enabled, the error message
 # only includes the protocol and origin of a RemoteFrame when with Site Isolation
 # is disabled the other frame is a LocalFrame which we log the URL with the
 # full path.
 http/tests/security/block-top-level-navigation-to-different-scheme-by-third-party-iframes.html [ Failure ]
+http/tests/security/block-top-level-navigations-by-sandboxed-iframe-with-propagated-user-gesture.html [ Failure ]
 http/tests/security/block-top-level-navigations-by-third-party-iframes.html [ Failure ]
 
 # This test has different output with site isolation enabled. A copy with site isolation enabled is
--- a/LayoutTests/platform/ios-site-isolation/TestExpectations
+++ b/LayoutTests/platform/ios-site-isolation/TestExpectations
@@ -225,7 +225,6 @@ http/tests/security/XFrameOptions/x-frame-options-multiple-headers-sameorigin-de
 http/tests/security/XFrameOptions/x-frame-options-parent-same-origin-deny.html [ Failure ]
 http/tests/security/allow-top-level-navigations-by-third-party-iframes-with-previous-user-activation.html [ Failure ]
 http/tests/security/allow-top-level-navigations-by-third-party-iframes-with-user-activation.html [ Failure ]
-http/tests/security/block-top-level-navigations-by-sandboxed-iframe-with-propagated-user-gesture.html [ Failure ]
 http/tests/security/block-top-level-navigations-by-third-party-sandboxed-iframe.html [ Failure ]
 http/tests/security/contentSecurityPolicy/1.1/frame-ancestors/frame-ancestors-nested-cross-in-same-none-block.html [ Failure ]
 http/tests/security/contentSecurityPolicy/1.1/frame-ancestors/frame-ancestors-nested-cross-in-same-self-block.html [ Failure ]
@@ -442,10 +441,11 @@ fast/block/float/independent-align-positioning.html [ Missing ]
 ####################################
 # Tests with intentional text fail #
 ####################################
-# The following two tests have a different console log with Site Isolation enabled
+# The following tests have a different console log with Site Isolation enabled
 # vs disabled. The difference is that with Site Isolation enabled, the error message
 # only includes the protocol and origin of a RemoteFrame when with Site Isolation
 # is disabled the other frame is a LocalFrame which we log the URL with the
 # full path.
 http/tests/security/block-top-level-navigation-to-different-scheme-by-third-party-iframes.html [ Failure ]
+http/tests/security/block-top-level-navigations-by-sandboxed-iframe-with-propagated-user-gesture.html [ Failure ]
 http/tests/security/block-top-level-navigations-by-third-party-iframes.html [ Failure ]


LayoutTests/platform/mac-site-isolation/TestExpectations
```diff
@@ -526,10 +525,11 @@ imported/w3c/web-platform-tests/cookies/third-party-cookies/third-party-cookies.
 ####################################
 # Tests with intentional text fail #
 ####################################
-# The following two tests have a different console log with Site Isolation enabled
+# The following tests have a different console log with Site Isolation enabled
```
aproskuryakov, line 522: This seems like a wrong concept to have. If it's expected to be different, we should check the results in, not effectively disable the test.

But also, I don't know why Site Isolation having different result is ever anything other than a bug.

atar13: > If it's expected to be different, we should check the results in, not effectively disable the test.

I spoke with @szewai about it and this was the advice that she gave, but I'm open to other ways of distinguishing this expected difference. Having it marked as a failure could be confusing down the line until someone reads this comment in `*-site-isolation/TestExpectations` file.

> But also, I don't know why Site Isolation having different result is ever anything other than a bug.

Before [this commit](https://commits.webkit.org/310093@main), we had an IPC message that would allow a frame to fetch the full URL of a remote frame in a different process. This full URL would contain the full resource path as opposed to just the protocol and domain (the difference between `https://a.com` and `https://a.com/resource-here.txt`). The full URL contains valuable information which would be of interest to compromised frames of different WebContent processes. So, instead when logging these "Unsafe JavaScript attempt ..." messages, we now only allow the protocol and domain to be queried from a remote frame. See the commit message here for more details as well: https://commits.webkit.org/310093@main

aproskuryakov: > Having it marked as a failure could be confusing

The test being effectively disabled seems like the bigger problem to me, but yes, it's also confusing.

> Before [this commit](https://commits.webkit.org/310093@main), we had an IPC message... The full URL contains valuable information

It's not quite clear to me if it's Web exposed, or only test exposed (sorry for not having looked into it deeper myself).

- If Web exposed, and we think that this is an OK change to make in terms of web compatibility, perhaps we should consider making it unconditionally, whether site isolation is enabled or not. That would decrease confusion and testing matrix — although perhaps we'd also want to time such a change thoughtfully.

- If only test exposed, seems like we could have any behavior we want when running tests, and it won't affect security of the product.

atar13: > If Web exposed, and we think that this is an OK change to make in terms of web compatibility, perhaps we should consider making it unconditionally, whether site isolation is enabled or not.

Yes, it's Web exposed since a site can attempt to perform a navigation which will trigger the console log that reveals the mismatch with site isolation enabled.

We could potentially have the same log be printed with site isolation enabled vs disabled, but it seemed like folks were more inclined to leave the LocalFrame case as-is since there's no threat of information leakage compared to the RemoteFrame case.

aproskuryakov: One option is to have this behavior explicitly control with a feature flag that can be overridden in a test. Then most tests could opt in to old behavior until the new one is default, and one test could verify the default for Site Isolation.

szewai: We are actually actively tracking `[ Fail ]` ones (e.g. engineers actively look to fix `[ Fail ]`) so they are not "skipped"; adding comments like this helps avoid duplicate investigation for now.
We could think about whether we should have unified behavior or have a better long-term solution (like having the bot check the SI-specific result; the reason we haven't done that is the bot is running tests with SI on and off and comparing the results, not checking expected.txt).

szewai (approved):

PR #62019, opened 2026-04-04
[JSC] Eagerly build AST for likely IIFEs

ast-hugger:
#### 709e4e7e7ec7e5488be35a7fe48938e150fb7f22
<pre>
[JSC] Eagerly build AST for likely IIFEs
<a href="https://bugs.webkit.org/show_bug.cgi?id=311459">https://bugs.webkit.org/show_bug.cgi?id=311459</a>
<a href="https://rdar.apple.com/174062165">rdar://174062165</a>

Reviewed by Yusuke Suzuki.

This patch detects likely IIFEs (immediately invoked function expressions) and builds
their AST right away, bypassing the usual initial syntax check.

In general, we can&apos;t tell if a function expression is immediately invoked until after we
parse it, but the following heuristic catches most real world cases. A function expression
is expected to be an IIFE if the token preceding the `function` keyword is either `(` or
`!` (`!` is commonly used by minifiers to introduce a function expression using one fewer
character than wrapping it in parentheses). Another potential heuristic is if the
`function` keyword is preceded by `,` and the expression before the comma was an IIFE.
This heuristic is not used at this time because the cost/benefit tradeoff is not as clear
in this case.

Key changes:

- A FunctionNode created when eagerly parsing an IIFE is independent from the AST created
  for the containing code and has its own lifetime. To support this, we change ParserArena
  and ParserArenaRoot so that an arena can be shared by multiple roots. A root no longer
  takes ownership of the arena memory, instead it retains a Ref to the arena.

- Eager IIFE parsing is controlled by two classes, EagerIIFEParseScope and
  EagerIIFEParseState. The scope is created in Parser::parseFunctionInfo() when the IIFE
  prediction heuristic triggers. The scope contains an ASTBuilder to use for the IIFE and
  an EagerIIFEParseState. The state is registered with the parser as an indication of the
  special parsing mode it is currently in. It collects parsing artifacts such as function
  parameter and source elements until we are ready to create the final FunctionNode.
  When parseFunctionInfo exits, the scope is destroyed and its destructor returns the
  parser into the original state.

- At the end of parseFunctionInfo, if an IIFE AST was eagerly built, we create a
  FunctionNode stored in m_eagerIIFERegistry. Parser::parse() checks the cache when the
  function is invoked and takes the ready-to-use AST from the cache.

Testing:
    - Regression-checked by existing tests.
    - Eager parsing is not directly testable in stress tests.
    - Added `JSTests/stress/eager-iife-column-tracking.js` to test the source column
    adjustment required on the first line of IIFEs, as uncovered by one of the layout
    tests.

Canonical link: <a href="https://commits.webkit.org/312444@main">https://commits.webkit.org/312444@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/2e6a77f534b5ec92068e0cb34565734dda92dc6c

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/159772 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/33239 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/26346 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/168630 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/114153 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/7ff32a6f-4c53-4c46-93d9-b893738f07ba) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/33344 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/33243 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/123816 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/86865 "1 flakes 9 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/dfc16bcc-763f-4e56-8617-dcf8251d185a) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/162730 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/26065 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/143511 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/104448 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/7ba255d0-0c24-4ec6-9654-d16e00a1da44) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/25117 "Passed tests") | [❌ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/23591 "Found 2 new API test failures: TestWebKitAPI.WKWebExtension.LoadFromiOSAppExtensionBundle, TestWebKitAPI.WKBackForwardList.BackForwardNavigationSkipsItemsWithoutUserGestureFragment (failure)") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/16393 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/151828 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/134808 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/21283 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/171123 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/20609 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/17140 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/22920 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/132069 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/32918 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/27670 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/132119 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35786 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/32903 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/143077 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/91001 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/26736 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/19890 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/192056 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/32412 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/98809 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/49388 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/31909 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/32156 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/32060 "Built successfully") | | | |

### DIFF
--- a/Source/JavaScriptCore/parser/Parser.cpp
+++ b/Source/JavaScriptCore/parser/Parser.cpp
@@ -89,6 +89,36 @@ std::atomic<unsigned> globalParseCount { 0 };
 
 WTF_MAKE_TZONE_ALLOCATED_IMPL(ModuleScopeData);
 
+// When the parser encounters an immediately-invoked function expression (IIFE), it can
+// build the AST right away instead of the usual syntax-checking parse followed by a full
+// AST-building reparse.
+//
+// EagerIIFEParseScope and EagerIIFEParseState work together to enable this optimization.
+// EagerIIFEParseScope is created as a local in parseFunctionInfo when an IIFE candidate
+// is detected. It owns an ASTBuilder and constructs an EagerIIFEParseState that installs
+// itself into the parser. The ASTBuilder uses the parser's own arena, so the eagerly built
+// AST nodes live in the same arena as the rest of the parse. The FunctionNode created for
+// the IIFE holds a Ref to this shared arena, keeping it alive.
+
+template<typename LexerType>
+class EagerIIFEParseScope {
+public:
+    EagerIIFEParseScope(Parser<LexerType>* parser, unsigned startOffset)
+        : m_builder(const_cast<VM&>(parser->m_vm), parser->m_parserArena.copyRef(), const_cast<SourceCode*>(parser->m_source))
+        , m_parseState(*parser, &m_builder, startOffset)
+    { }
+
+private:
+    ASTBuilder m_builder;
+    Parser<LexerType>::EagerIIFEParseState m_parseState;
+};
+
+template <typename LexerType>
+CodeFeatures Parser<LexerType>::EagerIIFEParseState::features() const { return m_builder->features(); }
+
+template <typename LexerType>
+int Parser<LexerType>::EagerIIFEParseState::numConstants() const { return m_builder->numConstants(); }
+
 ALWAYS_INLINE static SourceParseMode getAsyncFunctionBodyParseMode(SourceParseMode parseMode)
 {
     if (isAsyncGeneratorWrapperParseMode(parseMode))
@@ -138,16 +168,18 @@ Parser<LexerType>::Parser(VM& vm, const SourceCode& source, ImplementationVisibi
     , m_implementationVisibility(implementationVisibility)
     , m_parsingBuiltin(builtinMode == JSParserBuiltinMode::Builtin)
     , m_isInsideOrdinaryFunction(isInsideOrdinaryFunction)
+    , m_parserArena(ParserArena::create())
     , m_hasStackOverflow(false)
     , m_debuggerParseData(debuggerParseData)
 {
     m_lexer = makeUnique<LexerType>(vm, builtinMode, scriptMode);
-    m_lexer->setCode(source, &m_parserArena);
+    m_lexer->setCode(source, m_parserArena.ptr());
     m_token.m_startPosition.line = source.firstLine().oneBasedInt();
     m_token.m_startPosition.offset = source.startOffset();
     m_token.m_startPosition.lineStartOffset = source.startOffset();
     m_token.m_endPosition.offset = source.startOffset();
     m_functionCache = vm.addSourceProviderCache(source.provider());
+    m_eagerIIFERegistry = vm.addEagerIIFERegistry(source.provider());
 
     Scope* scope = pushScope();
     scope->setLexicallyScopedFeatures(lexicallyScopedFeatures);
@@ -214,7 +246,7 @@ static ALWAYS_INLINE bool NODELETE isPrivateFieldName(UniquedStringImpl* uid)
 template <typename LexerType>
 Expected<typename Parser<LexerType>::ParseInnerResult, String> Parser<LexerType>::parseInner(const Identifier& calleeName, ParsingContext parsingContext, std::optional<int> functionConstructorParametersEndPosition, const FixedVector<UnlinkedFunctionExecutable::ClassElementDefinition>* classElementDefinitions, const PrivateNameEnvironment* parentScopePrivateNames)
 {
-    ASTBuilder context(const_cast<VM&>(m_vm), m_parserArena, const_cast<SourceCode*>(m_source));
+    ASTBuilder context(const_cast<VM&>(m_vm), m_parserArena.copyRef(), const_cast<SourceCode*>(m_source));
     SourceParseMode parseMode = sourceParseMode();
     Scope* scope = currentScope();
     scope->setIsLexicalScope();
@@ -320,21 +352,7 @@ Expected<typename Parser<LexerType>::ParseInnerResult, String> Parser<LexerType>
             context.propagateArgumentsUse();
     }
 
-    CodeFeatures features = context.features();
-    if (scope->shadowsArguments())
-        features |= ShadowsArgumentsFeature;
-    if (m_seenTaggedTemplateInNonReparsingFunctionMode)
-        features |= NoEvalCacheFeature;
-    if (scope->hasNonSimpleParameterList())
-        features |= NonSimpleParameterListFeature;
-    if (scope->usesImportMeta())
-        features |= ImportMetaFeature;
-    if (m_seenArgumentsDotLength && scope->hasDeclaredGlobalArguments())
-        features |= ArgumentsFeature;
-    if (scope->asyncFunctionBodyDoesNotUseAwait())
-        features |= AsyncFunctionWithoutAwaitFeature;
-    if (scope->usesAwait())
-        features |= AwaitFeature;
+    CodeFeatures features = collectCodeFeatures(context.features(), scope);
 
 #if ASSERT_ENABLED
     if (m_parsingBuiltin && isProgramParseMode(parseMode)) {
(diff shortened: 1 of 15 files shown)


Source/JavaScriptCore/parser/Parser.h
```diff
@@ -2073,7 +2152,12 @@ class CACHE_LINE_ALIGNED Parser {
         m_errorMessage = String();
     }
 
-    // Cache line 0 (hot)
+    ALWAYS_INLINE void setCurrentArena(ParserArena& arena) {
```
Constellation: Align to WebKit coding style ({ should be in the next line).

Source/JavaScriptCore/parser/Parser.cpp
```diff
@@ -89,6 +89,42 @@ std::atomic<unsigned> globalParseCount { 0 };
...
+class EagerIIFEParseScope {
+public:
+    EagerIIFEParseScope(Parser<LexerType>* parser, unsigned startOffset)
+        : m_arena()
+        , m_builder(const_cast<VM&>(parser->m_vm), m_arena, const_cast<SourceCode*>(parser->m_source))
+        , m_parseState(*parser, &m_builder, m_arena, startOffset)
+    { }
+
+private:
+    ParserArena m_arena;
```
Constellation: Can we have multiple ParserArena in one parsing implementation? Currently Identifier* is tied to ParserArena's IdentifierArena's allocated `Identifier`, which means that this pointer will be invalidated when arena gets destroyed.

Source/JavaScriptCore/parser/Parser.cpp
```diff
@@ -2764,14 +2804,43 @@ template <class TreeBuilder> bool Parser<LexerType>::parseFunctionInfo(TreeBuild
...
+        // start column. During eager parsing columns are absolute, but column counting
+        // logic expects them relative to the containing function's '('.
+        if (m_iifeParseState
+            && !m_iifeParseState->isAvailable()
+            && tokenLineStart() <= m_iifeParseState->startOffset()) [[unlikely]] {
+
+            int adjustment = m_iifeParseState->startOffset() - tokenLineStart();
+            startColumn -= adjustment;
+            functionInfo.parametersStartColumn = startColumn;
+        }
```
Constellation: Can you describe a bit detail about this with the example code?
I feel like needing a hack here sounds weird and error-prone.

Source/JavaScriptCore/parser/Parser.cpp
```diff
@@ -2908,6 +3023,43 @@ template <class TreeBuilder> bool Parser<LexerType>::parseFunctionInfo(TreeBuild
...
+        SourceCode functionSource = m_source->subExpression(
+            functionInfo.startOffset, functionInfo.endOffset,
+            functionInfo.startLine, functionInfo.parametersStartColumn);
+
+        JSTokenLocation endLocation;
+        endLocation.line = m_lastTokenLocation.line;
+        endLocation.startOffset = functionInfo.endOffset;
+        endLocation.lineStartOffset = m_lastTokenLocation.lineStartOffset;
+
+        FunctionNodeUniquePtr functionNode(new FunctionNode(
```
Constellation: Use `makeUnique`.

Source/JavaScriptCore/parser/Parser.h
ast-hugger: Interesting why the style checker didn't complain about this one.

Source/JavaScriptCore/parser/Parser.cpp
ast-hugger: Right, but with this change identifier creation is funneled into whatever is `m_currentArena->identifierArena()`. `setCurrentArena()` also reassigns the lexer's identifier arena to point at it. So an IIFE is parsed using a temporary arena, and then its `FunctionNode` takes ownership of that arena's contents and the parser goes back to its own `m_arena` as `m_currentArena`.

Source/JavaScriptCore/parser/Parser.cpp
ast-hugger: In normal lazy parsing when a function is nested, columns on its first line are tracked relative to its parameter list start because that's where the SourceCode seen by the parser begins. `UnlinkedFunctionExecutable` expects that column on the first line are relative. When eagerly parsing, the parser sees the broader uses the original SourceCode that contains the function expression, and the columns are absolute. 

This was uncovered by a layout test `js/dom/line-column-numbers-expected.html`. The test I added with this PR reproduces it. Case 1 is the simplest example. I agree this looks weird. Maybe we should revise column tracking so it's always absolute in the top-level source?

Source/JavaScriptCore/parser/Parser.cpp
ast-hugger: We can't use `makeUnique` here because FunctionNodeUniquePtr needs a custom deleter, which it needs because `SourceProviderCache.h` is a project header and it can't include `Nodes.h` directly, which it would need to have the full definition of `FunctionNode`. I don't like that custom deleter overall, but not sure which alternative to choose - keep IIFE cache somewhere else, or promote the header, or something else. Suggestions welcome.

ast-hugger: Fixed - IIFE registry is now a separate entity from `SourceProviderCache` with its separate header file, so it can include `Nodes.h` and have `FunctionNode` fully defined, and the use the standard `std::unique_ptr` and `makeUnique`.

Source/JavaScriptCore/parser/Parser.cpp
```diff
@@ -2890,9 +2955,55 @@ template <class TreeBuilder> bool Parser<LexerType>::parseFunctionInfo(TreeBuild
...
+        if (m_seenArgumentsDotLength && functionScope->hasDeclaredGlobalArguments())
+            iifeFeatures |= ArgumentsFeature;
+        if (functionScope->asyncFunctionBodyDoesNotUseAwait())
+            iifeFeatures |= AsyncFunctionWithoutAwaitFeature;
+
+        iifeVarDeclarations = functionScope->declaredVariables();
+        IdentifierSet capturedVariables;
+        functionScope->getCapturedVars(capturedVariables);
+        for (auto& entry : capturedVariables)
+            iifeVarDeclarations.markVariableAsCaptured(entry.get());
```
Constellation: Maybe these things can be unified with the normal Parser's final code? IIRC, we have similar code.

Constellation (approved):
r=me with comments.

Source/JavaScriptCore/parser/Parser.cpp
ast-hugger: Ok, I'll look into that.

PR #62142, opened 2026-04-06
CSS1: character styled with :first-letter is not selectable

alanbaradlay:
#### 1f01e406b70ecb01f1e605b490badbd9c9b3f9ac
<pre>
CSS1: character styled with :first-letter is not selectable
<a href="https://bugs.webkit.org/show_bug.cgi?id=6185">https://bugs.webkit.org/show_bug.cgi?id=6185</a>
&lt;<a href="https://rdar.apple.com/problem/5688237">rdar://problem/5688237</a>&gt;

Reviewed by Ryosuke Niwa.

Given

  &lt;style&gt;
        div::first-letter { color: red; }
  &lt;/style&gt;
  &lt;div&gt;Hello&lt;/div&gt;

the DOM (inside the block container) has a single text node &quot;Hello&quot;
but WebKit splits this into two renderers to be able to style the first letter.

  RenderBlock (div)
        RenderInline (::first-letter, anonymous)
          RenderText &quot;H&quot;            &lt;- offsets 0-1 (no DOM node)
        RenderTextFragment &quot;ello&quot;   &lt;- offsets 0-4 (owns the DOM text node)

The existing code assumes each DOM node maps to exactly one renderer.
Every call to node-&gt;renderer() or Position::renderer() returns the
remaining fragment - the first-letter renderer is never consulted.

This means DOM offset 0 (&quot;H&quot;) gets checked against the remaining
fragment&apos;s text boxes, which start at &quot;e&quot;. Selection, caret positioning,
and offset validation all fail for the first-letter range because the
wrong renderer is asked about offsets it doesn&apos;t own.

The fix is to never consult node-&gt;renderer() directly for text offset
operations on first-letter content. Instead, all call sites go through
Position::rendererAndOffset() which resolves the 1-DOM-node-to-2-renderers
split by picking the correct renderer and converting to its local offset.
This is the central invariant of the fix - every place that previously
called node-&gt;renderer() and used a DOM offset against it now uses this
helper (or one of its wrappers) to get the right renderer + offset pair.

Key helpers introduced:
- Position::rendererAndOffset(): the single source of truth for resolving
  a DOM offset to the correct renderer + fragment-local offset. Routes to
  the first-letter renderer for offsets below fragmentStart, otherwise
  returns the remaining fragment with adjusted offset.
- Position::resolvedTextRendererAndOffset(): convenience wrapper that also
  casts to RenderText.
- convertOffsetInTextFragmentToNodeOffset(): converts fragment-local offset back to
  DOM offset. Used when exiting renderer space (caretMaxOffset,
  createPositionWithAffinity, endPositionForLine, positionAtRightBoundary-
  OfBiDiRun, TextIterator::emitText).
- convertNodeOffsetToOffsetInTextFragment(): converts DOM offset to
  fragment-local. Used in Position::upstream/downstream where there is no
  Position object to call rendererAndOffset() on.
- crossesFirstLetterBoundary(): prevents upstream/downstream from
  canonicalizing positions across the first-letter split. Without this,
  VisiblePosition(offset 0) and VisiblePosition(offset 1) would collapse
  to the same canonical position, making the first letter unselectable.
  Must run before the isStreamer update to see the pre-update lastVisible.

Position::isCandidate, isRenderedCharacter, rendersInDifferentPosition:
  These check text offset validity against renderer text boxes. They now
  use resolvedTextRendererAndOffset() to ask the correct renderer
  (first-letter or remaining fragment) instead of always asking the
  remaining fragment with a raw DOM offset.

Position::inlineBoxAndOffset: resolves to the correct inline box for
  caret/selection positioning. Now delegates to rendererAndOffset()
  instead of duplicating the first-letter routing logic.

RenderedPosition: stores the original DOM node (m_node) from the Position
  during construction. Without this, positionAtLeftBoundaryOfBiDiRun()
  produces a null-node Position when the inline box belongs to the
  anonymous first-letter text, causing adjustEndpointsAtBidiBoundary()
  to collapse the selection when dragging left past the first letter.

VisiblePosition::rightVisuallyDistinctCandidate: when bidi traversal
  lands on the anonymous first-letter text box, resolves the DOM node
  via remainingTextFragmentForFirstLetter() before creating the Position.

RenderTextFragment::canBeSelectionLeaf: returns true for remaining
  fragments with first-letter, matching base RenderText behavior so
  cross-element selections (Cmd+A) include the remaining text.

* LayoutTests/editing/selection/first-letter-caret-single-char-expected.txt: Added.
* LayoutTests/editing/selection/first-letter-caret-single-char.html: Added.
* LayoutTests/editing/selection/first-letter-selection-cross-element-expected.txt: Added.
* LayoutTests/editing/selection/first-letter-selection-cross-element.html: Added.
* LayoutTests/editing/selection/first-letter-selection-expected.txt: Added.
* LayoutTests/editing/selection/first-letter-selection-painting-expected.html: Added.
* LayoutTests/editing/selection/first-letter-selection-painting.html: Added.
* LayoutTests/editing/selection/first-letter-selection.html: Added.
* LayoutTests/editing/text-iterator/first-letter-word-boundary-expected.txt:
  Progression. The backwards word boundary with white-space:pre now
  correctly finds offset 1 instead of 0.

* LayoutTests/fast/css/first-letter-pre-text-iterator-expected.txt: Added.
* LayoutTests/fast/css/first-letter-pre-text-iterator.html: Added.
* LayoutTests/fast/text/first-letter-partial-invalidation-crash-expected.txt:
  Progression. We match other engines now.

* LayoutTests/platform/glib/editing/selection/extend-by-word-002-expected.txt:
* LayoutTests/platform/ios/editing/selection/extend-by-word-002-expected.txt:
* LayoutTests/platform/mac-wk2/TestExpectations:
  Remove [ Crash ] for editing/text-iterator/backwards-text-iterator-basic.html.
  Progression — the null-check in handleFirstLetter prevents the crash.

* LayoutTests/platform/mac/editing/selection/extend-by-word-002-expected.txt:
  Progression. Content is selected now across &lt;li&gt;s with first-letter.

* Source/WebCore/dom/Position.cpp:
(WebCore::crossesFirstLetterBoundary):
(WebCore::Position::upstream const):
(WebCore::Position::downstream const):
(WebCore::Position::hasRenderedNonAnonymousDescendantsWithHeight):
  Was skipping ::first-letter pseudo renderers because they have no
  associated DOM node. This made positions after first-letter content
  fail isCandidate() -- the function thought the element had no visible
  content. The fix lets first-letter renderers count as rendered
  descendants with height.
(WebCore::Position::isCandidate const):
(WebCore::Position::isRenderedCharacter const):
(WebCore::Position::rendersInDifferentPosition const):
(WebCore::Position::rendererAndOffset const):
(WebCore::Position::resolvedTextRendererAndOffset const):
(WebCore::Position::inlineBoxAndOffset const):
* Source/WebCore/dom/Position.h:
* Source/WebCore/dom/PositionIterator.cpp:
(WebCore::PositionIterator::isCandidate const):
* Source/WebCore/editing/Editing.cpp:
(WebCore::caretMaxOffset):
(WebCore::convertOffsetInTextFragmentToNodeOffset):
(WebCore::convertNodeOffsetToOffsetInTextFragment):
* Source/WebCore/editing/Editing.h:
* Source/WebCore/editing/FrameSelection.cpp:
(WebCore::FrameSelection::updateAppearance):
* Source/WebCore/editing/RenderedPosition.cpp:
(WebCore::nodeFromPosition):
(WebCore::rendererFromPosition):
(WebCore::RenderedPosition::RenderedPosition):
(WebCore::RenderedPosition::positionAtLeftBoundaryOfBiDiRun const):
(WebCore::RenderedPosition::positionAtRightBoundaryOfBiDiRun const):
* Source/WebCore/editing/RenderedPosition.h:
(WebCore::TextIterator::handleTextNode):
(WebCore::TextIterator::handleTextRun):
(WebCore::TextIterator::emitText):
(WebCore::SimplifiedBackwardsTextIterator::handleFirstLetter):
  Null-check firstLetterRenderer before dereferencing. The first-letter
  renderer can be destroyed by DOM mutations (e.g. deleteFromDocument)
  while the backwards iterator is still referencing the text node.
* Source/WebCore/editing/VisiblePosition.cpp:
(WebCore::remainingTextFragmentForFirstLetter):
(WebCore::VisiblePosition::leftVisuallyDistinctCandidate const):
(WebCore::VisiblePosition::rightVisuallyDistinctCandidate const):
* Source/WebCore/editing/VisibleUnits.cpp:
(WebCore::endPositionForLine):
(WebCore::findEndOfParagraph):
* Source/WebCore/rendering/RenderObject.cpp:
(WebCore::RenderObject::createPositionWithAffinity const):
* Source/WebCore/rendering/RenderTextFragment.cpp:
(WebCore::RenderTextFragment::canBeSelectionLeaf const):

Canonical link: <a href="https://commits.webkit.org/311287@main">https://commits.webkit.org/311287@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/f03ebbf29a899974462d764b10b1aa4cf8c02655

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/156356 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/29691 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/22873 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/165177 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/110436 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/158227 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/29824 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/29694 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/121085 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85135 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/159314 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/23303 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/140412 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/101756 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/22371 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/20542 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/12949 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/132046 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/18243 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/167659 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/11772 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/19856 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/129210 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/29292 "Built successfully") | [❌ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/24609 "Found 1 new test failure: imported/w3c/web-platform-tests/shadow-dom/reference-target/tentative/form.html (failure)") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/129322 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35069 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/29214 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/140037 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/87010 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/24149 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/16836 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/28924 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/92880 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/28450 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/28678 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/28574 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/dom/Position.cpp
+++ b/Source/WebCore/dom/Position.cpp
@@ -31,6 +31,7 @@
 #include "CSSComputedStyleDeclaration.h"
 #include "ContainerNodeInlines.h"
 #include "EditingInlines.h"
+#include "Editing.h"
 #include "ElementInlines.h"
 #include "HTMLBRElement.h"
 #include "HTMLBodyElement.h"
@@ -54,6 +55,7 @@
 #include "RenderLineBreak.h"
 #include "RenderObjectInlines.h"
 #include "RenderText.h"
+#include "RenderTextFragment.h"
 #include "SVGElementTypeHelpers.h"
 #include "SVGTextElement.h"
 #include "Text.h"
@@ -642,6 +644,26 @@ static Node* enclosingVisualBoundary(SUPPRESS_UNCHECKED_LOCAL Node* node)
     return node;
 }
 
+// The first-letter and remaining text are separate renderers but share one DOM
+// text node. upstream/downstream must not cross this boundary, just like they
+// must not cross visually distinct node boundaries.
+static bool crossesFirstLetterBoundary(const RenderObject& renderer, const PositionIterator& current, const PositionIterator& previousLastVisible)
+{
+    auto* textFragment = dynamicDowncast<RenderTextFragment>(renderer);
+    if (!textFragment || !textFragment->firstLetter())
+        return false;
+    if (current.node() != previousLastVisible.node())
+        return false;
+    auto fragmentStart = static_cast<int>(textFragment->start());
+    auto currentOffset = current.offsetInLeafNode();
+    auto lastOffset = previousLastVisible.offsetInLeafNode();
+
+    bool currentOffsetIsInFirstLetter = currentOffset < fragmentStart;
+    bool lastOffsetIsInFirstLetter = lastOffset < fragmentStart;
+    return currentOffsetIsInFirstLetter != lastOffsetIsInFirstLetter;
+}
+
+
 // upstream() and downstream() want to return positions that are either in a
 // text node or at just before a non-text node.  This method checks for that.
 static bool isStreamer(const PositionIterator& pos)
@@ -711,7 +733,10 @@ Position Position::upstream(EditingBoundaryCrossingRule rule) const
             lastVisible = currentPosition;
             break;
         }
-        
+
+        if (crossesFirstLetterBoundary(*renderer, currentPosition, lastVisible))
+            return lastVisible;
+
         // track last visible streamer position
         if (isStreamer(currentPosition))
             lastVisible = currentPosition;
@@ -740,10 +765,10 @@ Position Position::upstream(EditingBoundaryCrossingRule rule) const
                 // render tree which can have a different length due to case transformation.
                 // Until we resolve that, disable this so we can run the layout tests!
                 //ASSERT(currentOffset >= renderer->caretMaxOffset());
-                return makeDeprecatedLegacyPosition(currentNode.ptr(), renderer->caretMaxOffset());
+                return makeDeprecatedLegacyPosition(currentNode.ptr(), caretMaxOffset(currentNode));
             }
 
-            unsigned textOffset = currentPosition.offsetInLeafNode();
+            auto textOffset = convertNodeOffsetToOffsetInTextFragment(*textRenderer, currentPosition.offsetInLeafNode());
             for (auto box = firstTextBox; box;) {
                 if (textOffset > box->start() && textOffset <= box->end())
                     return currentPosition;
@@ -827,6 +852,9 @@ Position Position::downstream(EditingBoundaryCrossingRule rule) const
             break;
         }
 
+        if (crossesFirstLetterBoundary(*renderer, currentPosition, lastVisible))
+            return lastVisible;
+
         // track last visible streamer position
         if (isStreamer(currentPosition))
             lastVisible = currentPosition;
@@ -849,7 +877,7 @@ Position Position::downstream(EditingBoundaryCrossingRule rule) const
                 return makeContainerOffsetPosition(currentNode.ptr(), textRenderer->caretMinOffset());
             }
 
-            unsigned textOffset = currentPosition.offsetInLeafNode();
+            auto textOffset = convertNodeOffsetToOffsetInTextFragment(*textRenderer, currentPosition.offsetInLeafNode());
             for (auto box = firstTextBox; box;) {
                 if (!box->length() && textOffset == box->start())
                     return currentPosition;
@@ -898,7 +926,13 @@ bool Position::hasRenderedNonAnonymousDescendantsWithHeight(const RenderElement&
     auto isHorizontal = renderer.isHorizontalWritingMode();
     CheckedPtr stop = renderer.nextInPreOrderAfterChildren();
     for (CheckedPtr descendant = renderer.firstChild(); descendant && descendant != stop; descendant = descendant->nextInPreOrder()) {
-        if (!descendant->nonPseudoNode())
+        auto shouldSkip = [&] {
+            if (descendant->nonPseudoNode())
+                return false;
+            CheckedPtr renderElement = dynamicDowncast<RenderElement>(*descendant);
+            return !renderElement || !renderElement->isFirstLetter();
+        };
+        if (shouldSkip())
             continue;
 
         auto boundingBoxLogicalHeight = [&](auto rect) {
@@ -916,7 +950,7 @@ bool Position::hasRenderedNonAnonymousDescendantsWithHeight(const RenderElement&
             continue;
         }
         if (CheckedPtr renderInline = dynamicDowncast<RenderInline>(*descendant)) {
-            if (isEmptyInline(*renderInline) && boundingBoxLogicalHeight(renderInline->linesBoundingBox()))
+            if ((renderInline->isFirstLetter() || isEmptyInline(*renderInline)) && boundingBoxLogicalHeight(renderInline->linesBoundingBox()))
                 return true;
             continue;
         }
@@ -985,8 +1019,10 @@ bool Position::isCandidate() const
         return !m_offset && m_anchorType != PositionIsAfterAnchor && !nodeIsUserSelectNone(node->parentNode());
     }
 
-    if (CheckedPtr renderText = dynamicDowncast<RenderText>(*renderer))
-        return !nodeIsUserSelectNone(node.get()) && renderText->containsCaretOffset(m_offset);
+    if (is<RenderText>(*renderer)) {
+        auto [resolvedText, resolvedOffset] = resolvedTextRendererAndOffset();
+        return !nodeIsUserSelectNone(node.get()) && resolvedText && resolvedText->containsCaretOffset(resolvedOffset);
+    }
 
     if (positionBeforeOrAfterNodeIsCandidate(*node)) {
         return ((atFirstEditingPositionForNode() && m_anchorType == PositionIsBeforeAnchor)
@@ -1013,9 +1049,10 @@ bool Position::isCandidate() const
 
 bool Position::isRenderedCharacter() const
 {
-    RefPtr text = dynamicDowncast<Text>(deprecatedNode());
-    CheckedPtr renderer = text ? text->renderer() : nullptr;
-    return renderer && renderer->containsRenderedCharacterOffset(m_offset);
+    if (!dynamicDowncast<Text>(deprecatedNode()))
+        return false;
+    auto [renderText, offset] = resolvedTextRendererAndOffset();
+    return renderText && renderText->containsRenderedCharacterOffset(offset);
 }
 
 static bool inSameEnclosingBlockFlowElement(Node* a, Node* b)
@@ -1061,16 +1098,16 @@ bool Position::rendersInDifferentPosition(const Position& position) const
     if (!inSameEnclosingBlockFlowElement(node.get(), positionNode.get()))
         return true;
 
-    CheckedPtr textRenderer = dynamicDowncast<RenderText>(*renderer);
-    if (textRenderer && !textRenderer->containsCaretOffset(m_offset))
+    auto [textRenderer, thisOffset] = resolvedTextRendererAndOffset();
+    if (textRenderer && !textRenderer->containsCaretOffset(thisOffset))
         return false;
 
-    CheckedPtr textPositionRenderer = dynamicDowncast<RenderText>(*positionRenderer);
-    if (textPositionRenderer && !textPositionRenderer->containsCaretOffset(position.m_offset))
+    auto [textPositionRenderer, positionOffset] = position.resolvedTextRendererAndOffset();
+    if (textPositionRenderer && !textPositionRenderer->containsCaretOffset(positionOffset))
         return false;
 
-    unsigned thisRenderedOffset = textRenderer ? textRenderer->countRenderedCharacterOffsetsUntil(m_offset) : m_offset;
-    unsigned positionRenderedOffset = textPositionRenderer ? textPositionRenderer->countRenderedCharacterOffsetsUntil(position.m_offset) : position.m_offset;
+    unsigned thisRenderedOffset = textRenderer ? textRenderer->countRenderedCharacterOffsetsUntil(thisOffset) : m_offset;
+    unsigned positionRenderedOffset = textPositionRenderer ? textPositionRenderer->countRenderedCharacterOffsetsUntil(positionOffset) : position.m_offset;
 
     if (renderer == positionRenderer && thisRenderedOffset == positionRenderedOffset)
         return false;
@@ -1195,16 +1232,52 @@ static Position upstreamIgnoringEditingBoundaries(Position position)
     return position;
 }
 
+std::pair<RenderObject*, unsigned> Position::rendererAndOffset() const
+{
+    auto* node = anchorNode();
+    if (!node)
+        return { nullptr, 0 };
+    auto* renderer = node->renderer();
+    if (!renderer)
+        return { nullptr, 0 };
+    unsigned offset = deprecatedEditingOffset();
+    CheckedPtr textFragment = dynamicDowncast<RenderTextFragment>(*renderer);
+    if (!textFragment || !textFragment->firstLetter())
+        return { renderer, offset };
+
+    if (offset >= textFragment->start() && textFragment->text().length())
+        return { renderer, offset - textFragment->start() };
+
+    // It looks like we are on the first letter renderer.
+    CheckedPtr firstLetterRenderer = dynamicDowncast<RenderText>(textFragment->firstLetter()->firstChild());
+    if (!firstLetterRenderer) {
+        ASSERT_NOT_REACHED();
+        return { nullptr, 0 };
+    }
+
+    // The first-letter text may not start at DOM offset 0 (e.g. collapsed whitespace precedes the first letter).
+    auto firstLetterStart = textFragment->start() - firstLetterRenderer->text().length();
+    if (offset < firstLetterStart)
+        return { nullptr, 0 };
+    return { firstLetterRenderer, offset - firstLetterStart };
+}
+
+std::pair<RenderText*, unsigned> Position::resolvedTextRendererAndOffset() const
+{
+    auto [renderer, offset] = rendererAndOffset();
+    if (auto* renderText = dynamicDowncast<RenderText>(renderer))
+        return { renderText, offset };
+    return { nullptr, 0 };
+}
+
 InlineBoxAndOffset Position::inlineBoxAndOffset(Affinity affinity, TextDirection primaryDirection) const
 {
     auto caretOffset = static_cast<unsigned>(deprecatedEditingOffset());
 
-    RefPtr node = deprecatedNode();
-    if (!node)
-        return { { }, caretOffset };
-    CheckedPtr renderer = node->renderer();
+    auto [renderer, resolvedOffset] = rendererAndOffset();
     if (!renderer)
         return { { }, caretOffset };
+    caretOffset = resolvedOffset;
 
     InlineIterator::LeafBoxIterator box;
 
--- a/Source/WebCore/dom/PositionIterator.cpp
+++ b/Source/WebCore/dom/PositionIterator.cpp
@@ -175,8 +175,12 @@ bool PositionIterator::isCandidate() const
     if (renderer->isBR())
         return Position(*this).isCandidate();
 
-    if (CheckedPtr renderText = dynamicDowncast<RenderText>(*renderer))
-        return !Position::nodeIsUserSelectNone(anchorNode.get()) && renderText->containsCaretOffset(m_offsetInAnchor);
+    if (is<RenderText>(*renderer)) {
+        if (Position::nodeIsUserSelectNone(anchorNode.get()))
+            return false;
+        auto [resolvedText, resolvedOffset] = Position(*this).resolvedTextRendererAndOffset();
+        return resolvedText && resolvedText->containsCaretOffset(resolvedOffset);
+    }
 
     if (positionBeforeOrAfterNodeIsCandidate(*anchorNode))
         return (atStartOfNode() || atEndOfNode()) && !Position::nodeIsUserSelectNone(anchorNode->parentNode());
--- a/Source/WebCore/editing/Editing.cpp
+++ b/Source/WebCore/editing/Editing.cpp
@@ -66,6 +66,7 @@
 #include "RenderStyle+GettersInlines.h"
 #include "RenderTableCell.h"
 #include "RenderTextControlSingleLine.h"
+#include "RenderTextFragment.h"
 #include "RenderedPosition.h"
 #include "ShadowRoot.h"
 #include "Text.h"
@@ -904,9 +905,23 @@ int caretMaxOffset(const Node& node)
     // For rendered text nodes, return the last position that a caret could occupy.
     if (auto* text = dynamicDowncast<Text>(node)) {
         if (CheckedPtr renderer = text->renderer())
-            return renderer->caretMaxOffset();
+            return convertOffsetInTextFragmentToNodeOffset(*renderer, renderer->caretMaxOffset());
     }
     return lastOffsetForEditing(node);
+}
+
+unsigned convertOffsetInTextFragmentToNodeOffset(const RenderObject& renderer, unsigned offset)
+{
+    if (auto* textFragment = dynamicDowncast<RenderTextFragment>(renderer); textFragment && textFragment->firstLetter())
+        return offset + textFragment->start();
+    return offset;
+}
+
+unsigned convertNodeOffsetToOffsetInTextFragment(const RenderObject& renderer, unsigned offset)
+{
+    if (auto* textFragment = dynamicDowncast<RenderTextFragment>(renderer); textFragment && textFragment->firstLetter() && offset >= textFragment->start())
+        return offset - textFragment->start();
+    return offset;
 }
 
 bool lineBreakExistsAtVisiblePosition(const VisiblePosition& position)
--- a/Source/WebCore/editing/RenderedPosition.cpp
+++ b/Source/WebCore/editing/RenderedPosition.cpp
@@ -32,6 +32,7 @@
 #include "RenderedPosition.h"
 
 #include "CaretRectComputation.h"
+#include "Editing.h"
 #include "InlineRunAndOffset.h"
 #include "NodeInlines.h"
 #include "RenderObjectInlines.h"
@@ -39,15 +40,15 @@
 
 namespace WebCore {
 
-static inline const RenderObject* NODELETE rendererFromPosition(const Position& position)
+static inline Node* NODELETE nodeFromPosition(const Position& position)
 {
     ASSERT(position.isNotNull());
-    Node* rendererNode = nullptr;
+    Node* node = nullptr;
     switch (position.anchorType()) {
     case Position::PositionIsOffsetInAnchor:
-        rendererNode = position.computeNodeAfterPosition();
-        if (!rendererNode || !rendererNode->renderer())
-            rendererNode = position.anchorNode()->lastChild();
+        node = position.computeNodeAfterPosition();
+        if (!node || !node->renderer())
+            node = position.anchorNode()->lastChild();
         break;
 
     case Position::PositionIsBeforeAnchor:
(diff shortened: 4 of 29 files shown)


Source/WebCore/dom/Position.cpp
```diff
@@ -712,6 +740,10 @@ Position Position::upstream(EditingBoundaryCrossingRule rule) const
             break;
         }
         
+        // Don't move past the first-letter boundary within a text node.
```
rniwa: I don't think this comment is necessary. The function name conveys the same information.

Source/WebCore/dom/Position.cpp
```diff
@@ -642,6 +643,33 @@ static Node* enclosingVisualBoundary(SUPPRESS_UNCHECKED_LOCAL Node* node)
...
+        return false;
+    // The boundary check only applies when both positions are in the same text node.
+    if (current.node() != previousLastVisible.node())
+        return false;
+    auto fragmentStart = static_cast<int>(textFragment->start());
+    auto currentOffset = current.offsetInLeafNode();
+    auto lastOffset = previousLastVisible.offsetInLeafNode();
+    // One offset is in the first-letter range and the other is not.
+    // Offsets at fragmentStart belong to the remaining fragment, not the first-letter.
+    return (currentOffset < fragmentStart) != (lastOffset < fragmentStart);
```
rniwa: Maybe better to define local variables with descriptive names like this:
```cpp
bool currentOffsetIsInFirstLetter = currentOffset < fragmentStart;
bool lastOffsetIsInFirstLetter = lastOffset < fragmentStart;
return currentOffsetIsInFirstLetter == lastOffsetIsInFirstLetter;
```

Source/WebCore/dom/Position.cpp
```diff
@@ -642,6 +643,33 @@ static Node* enclosingVisualBoundary(SUPPRESS_UNCHECKED_LOCAL Node* node)
...
 
+// The first-letter and remaining text are separate renderers but share one DOM
+// text node. upstream/downstream must not cross this boundary, just like they
+// must not cross visually distinct node boundaries.
+static bool crossesFirstLetterBoundary(const RenderObject& renderer, const PositionIterator& current, const PositionIterator& previousLastVisible)
+{
+    auto* textFragment = dynamicDowncast<RenderTextFragment>(renderer);
+    if (!textFragment || !textFragment->firstLetter())
+        return false;
+    // The boundary check only applies when both positions are in the same text node.
```
rniwa: I don't think this comment is necessary. It's pretty obvious from the function name & context.

Source/WebCore/dom/Position.cpp
```diff
@@ -642,6 +643,33 @@ static Node* enclosingVisualBoundary(SUPPRESS_UNCHECKED_LOCAL Node* node)
...
+    auto fragmentStart = static_cast<int>(textFragment->start());
+    auto currentOffset = current.offsetInLeafNode();
+    auto lastOffset = previousLastVisible.offsetInLeafNode();
+    // One offset is in the first-letter range and the other is not.
+    // Offsets at fragmentStart belong to the remaining fragment, not the first-letter.
+    return (currentOffset < fragmentStart) != (lastOffset < fragmentStart);
+}
+
+// Convert a DOM offset to fragment-local for remaining text fragments with first-letter.
+static unsigned firstLetterAdjustedFragmentLocalOffset(const RenderObject& renderer, unsigned offset)
```
rniwa: Instead of adding a comment like this, we should just give this a more descriptive name:
e.g. `convertNodeOffsetToOffsetInTextFragment`.

Source/WebCore/dom/Position.cpp
```diff
@@ -827,6 +859,10 @@ Position Position::downstream(EditingBoundaryCrossingRule rule) const
             break;
         }
 
+        // Don't move past the first-letter boundary within a text node.
```
rniwa: Ditto, no need for this comment.

Source/WebCore/dom/Position.cpp
```diff
@@ -898,7 +934,13 @@ bool Position::hasRenderedNonAnonymousDescendantsWithHeight(const RenderElement&
     auto isHorizontal = renderer.isHorizontalWritingMode();
     CheckedPtr stop = renderer.nextInPreOrderAfterChildren();
     for (CheckedPtr descendant = renderer.firstChild(); descendant && descendant != stop; descendant = descendant->nextInPreOrder()) {
-        if (!descendant->nonPseudoNode())
+        auto shouldSkip = [&] {
+            if (descendant->nonPseudoNode())
+                return false;
+            CheckedPtr renderElement = dynamicDowncast<RenderElement>(*descendant);
+            return !renderElement || !renderElement->isFirstLetter();
```
rniwa, line 933: What is this change about? Could you explain in the commit message?

Source/WebCore/dom/Position.cpp
```diff
@@ -1195,16 +1240,52 @@ static Position upstreamIgnoringEditingBoundaries(Position position)
...
+        return { { }, { } };
+    auto* renderer = node->renderer();
+    if (!renderer)
+        return { { }, { } };
+    unsigned offset = deprecatedEditingOffset();
+    CheckedPtr textFragment = dynamicDowncast<RenderTextFragment>(*renderer);
+    if (!textFragment || !textFragment->firstLetter())
+        return { renderer, offset };
+
+    // When the remaining fragment is empty (entire text is the first letter),
```
rniwa: This comment doesn't seem to match the code below??

Source/WebCore/dom/Position.cpp
```diff
@@ -985,8 +1027,10 @@ bool Position::isCandidate() const
         return !m_offset && m_anchorType != PositionIsAfterAnchor && !nodeIsUserSelectNone(node->parentNode());
     }
 
-    if (CheckedPtr renderText = dynamicDowncast<RenderText>(*renderer))
-        return !nodeIsUserSelectNone(node.get()) && renderText->containsCaretOffset(m_offset);
+    if (CheckedPtr renderText = dynamicDowncast<RenderText>(*renderer)) {
```
rniwa: It seems like we should be using `is<RenderText>` instead?

Source/WebCore/dom/Position.cpp
```diff
@@ -1195,16 +1240,52 @@ static Position upstreamIgnoringEditingBoundaries(Position position)
     return position;
 }
 
+std::pair<RenderObject*, unsigned> Position::rendererAndOffset() const
+{
+    auto* node = deprecatedNode();
+    if (!node)
+        return { { }, { } };
```
rniwa: Hm... maybe better to use `nullptr` and `0 respectively to be more explicit?

Source/WebCore/dom/Position.cpp
```diff
@@ -1195,16 +1240,52 @@ static Position upstreamIgnoringEditingBoundaries(Position position)
...
+    auto* renderer = node->renderer();
+    if (!renderer)
+        return { { }, { } };
```
rniwa: Ditto.

Source/WebCore/dom/Position.cpp
```diff
@@ -1195,16 +1240,52 @@ static Position upstreamIgnoringEditingBoundaries(Position position)
...
+    // When the remaining fragment is empty (entire text is the first letter),
+    // route to the first-letter renderer's end position so "after the last character" is a valid caret position.
+    if (offset >= textFragment->start() && textFragment->text().length())
+        return { renderer, offset - textFragment->start() };
+
+    // It looks like we are on the first letter renderer.
+    CheckedPtr firstLetterRenderer = dynamicDowncast<RenderText>(textFragment->firstLetter()->firstChild());
+    if (!firstLetterRenderer) {
+        ASSERT_NOT_REACHED();
+        return { { }, { } };
```
rniwa: Ditto.

Source/WebCore/dom/Position.cpp
```diff
@@ -1195,16 +1240,52 @@ static Position upstreamIgnoringEditingBoundaries(Position position)
...
+    }
+
+    // The first-letter text may not start at DOM offset 0 (e.g. collapsed whitespace precedes the first letter).
+    auto firstLetterStart = textFragment->start() - firstLetterRenderer->text().length();
+    return offset >= firstLetterStart ? std::make_pair(firstLetterRenderer, offset - firstLetterStart) : std::make_pair(nullptr, 0);
```
rniwa: I think it's more clear to do:
```cpp
if (offset < firstLetterStart)
    return { nullptr, 0 };
return { firstLetterRenderer, offset - firstLetterStart };
```

Source/WebCore/dom/PositionIterator.cpp
```diff
@@ -175,8 +175,12 @@ bool PositionIterator::isCandidate() const
     if (renderer->isBR())
         return Position(*this).isCandidate();
 
-    if (CheckedPtr renderText = dynamicDowncast<RenderText>(*renderer))
-        return !Position::nodeIsUserSelectNone(anchorNode.get()) && renderText->containsCaretOffset(m_offsetInAnchor);
+    if (CheckedPtr renderText = dynamicDowncast<RenderText>(*renderer)) {
```
rniwa: It seems like we can use `is<RenderText>` instead?

Source/WebCore/editing/Editing.cpp
```diff
@@ -904,11 +905,18 @@ int caretMaxOffset(const Node& node)
     // For rendered text nodes, return the last position that a caret could occupy.
     if (auto* text = dynamicDowncast<Text>(node)) {
         if (CheckedPtr renderer = text->renderer())
-            return renderer->caretMaxOffset();
+            return firstLetterAdjustedDOMOffset(*renderer, renderer->caretMaxOffset());
     }
     return lastOffsetForEditing(node);
 }
 
+unsigned firstLetterAdjustedDOMOffset(const RenderObject& renderer, unsigned offset)
```
rniwa: Maybe say `NodeOffset` instead of `DOMOffset`?

Source/WebCore/editing/RenderedPosition.cpp
```diff
@@ -85,6 +86,7 @@ RenderedPosition::RenderedPosition(const Position& position, Affinity affinity)
     if (position.isNull())
         return;
 
+    m_node = position.deprecatedNode();
```
rniwa: Oh, please don't use `deprecatedNode`.
We should extra the code to find the appropriate node given a `Position` from `rendererFromPosition`.

Source/WebCore/editing/TextIterator.cpp
```diff
@@ -607,6 +607,8 @@ bool TextIterator::handleTextNode()
     CheckedRef renderer = *textNode->renderer();
     m_lastTextNode = textNode.ptr();
     auto rendererText = rendererTextForBehavior(renderer.get());
+    // textFragment is the renderer that holds the trailing content and has a reference to the first letter renderer.
+    CheckedPtr textFragment = dynamicDowncast<RenderTextFragment>(renderer.get());
```
rniwa: Instead of adding a comment like this, we should give it a more descriptive name like `textFragmentWithRemainingTextAfterFirstLetter` or something.

Source/WebCore/editing/TextIterator.cpp
```diff
@@ -627,9 +629,15 @@ bool TextIterator::handleTextNode()
         }
         if (renderer->style().visibility() != Visibility::Visible && !m_behaviors.contains(TextIteratorBehavior::IgnoresStyleVisibility))
             return false;
-        int rendererTextLength = rendererText.length();
-        int end = (textNode.ptr() == m_endContainer) ? m_endOffset : INT_MAX;
-        int runEnd = std::min(rendererTextLength, end);
+        auto rendererTextLength = static_cast<int>(rendererText.length());
+        auto end = (textNode.ptr() == m_endContainer) ? m_endOffset : INT_MAX;
+        if (textFragment && textFragment->firstLetter()) {
+            // The remaining text fragment's text is fragment-local. Convert DOM offsets to fragment-local so we can compare against the renderer's text length.
```
rniwa: Can we reuse `convertNodeOffsetToOffsetInTextFragment` here?

Source/WebCore/editing/TextIterator.cpp
```diff
@@ -665,12 +673,21 @@ void TextIterator::handleTextRun()
...
-        rangeEnd = m_endOffset;
+    // For remaining text fragments after a first-letter split, text box offsets are fragment-local but m_offset/m_endOffset are DOM offsets.
+    unsigned remainingFragmentStart = 0;
+    if (auto* renderText = dynamicDowncast<RenderTextFragment>(renderer.get()); renderText && renderText->firstLetter())
+        remainingFragmentStart = renderText->start();
+
+    auto toFragmentLocal = [&](unsigned domOffset) {
+        return domOffset > remainingFragmentStart ? domOffset - remainingFragmentStart : 0;
+    };
+    auto toDOMOffset = [&](unsigned localOffset) {
```
rniwa: I'd call this `toNodeOffset` instead of `toDOMOffset`.

Source/WebCore/editing/Editing.cpp
rniwa: I'd call this something like `convertOffsetInTextFragmentToNodeOffset` or something.

Source/WebCore/editing/TextIterator.cpp
```diff
@@ -1201,8 +1218,11 @@ void TextIterator::emitText(Text& textNode, RenderText& renderer, int textStartO
 
     m_positionNode = textNode;
     m_positionOffsetBaseNode = nullptr;
-    m_positionStartOffset = textStartOffset;
-    m_positionEndOffset = textEndOffset;
+    // For remaining text fragments after a first-letter split, the text offsets are
+    // fragment-local but position offsets need to be DOM-relative for range() to
+    // return correct boundary points (used by word/sentence boundary detection).
+    m_positionStartOffset = firstLetterAdjustedDOMOffset(renderer, textStartOffset);
```
rniwa: This is converting a offset in TextFragment to offset in DOM node.
I'd give it a more descriptive name after that.

Source/WebCore/rendering/RenderTextFragment.cpp
```diff
@@ -68,7 +68,14 @@ RenderTextFragment::~RenderTextFragment()
 
 bool RenderTextFragment::canBeSelectionLeaf() const
 {
-    return textNode() && textNode()->hasEditableStyle();
+    if (RefPtr textNode = this->textNode()) {
+        // Remaining (trailing) text fragments with first-letter are always selectable,
+        // matching the base RenderText::canBeSelectionLeaf() behavior.
+        return firstLetter() ? true : textNode->hasEditableStyle();
```
rniwa: Maybe you can do: `firstLetter() || textNode->hasEditableStyle()`?

Source/WebCore/editing/VisiblePosition.cpp
```diff
@@ -58,6 +60,16 @@ namespace WebCore {
 
 using namespace HTMLNames;
 
+// The anonymous text renderer inside ::first-letter has no DOM node.
+static RenderTextFragment* remainingTextFragmentForFirstLetter(const RenderObject& renderer)
+{
+    if (renderer.node())
+        return { };
+    if (auto* parent = dynamicDowncast<RenderBoxModelObject>(renderer.parent()))
```
rniwa, line 68: Is it always true that the parent will have `firstLetterRemainingText`? Wouldn't it be possible for an anonymous renderer to exist to account for inline-block mismatch?

Source/WebCore/dom/Position.cpp
alanbaradlay: yeah, some leftover, sorry about that.

Source/WebCore/editing/VisiblePosition.cpp
alanbaradlay: The relation between the RenderText and its first letter fragment (through the anon RenderInline) is pretty tight. I can't think of any case where an anon renderer may be inserted in-between the parent and the fragment. I looked at the tree builder code and tested a few cases but nothing seemed to indicate that a renderer may get inserted (e.g. display: content, or float, or initial-letter, or inline-block). Let me know if you have a specific test case in mind.

Source/WebCore/editing/TextIterator.cpp
```diff
@@ -627,9 +628,13 @@ bool TextIterator::handleTextNode()
         }
         if (renderer->style().visibility() != Visibility::Visible && !m_behaviors.contains(TextIteratorBehavior::IgnoresStyleVisibility))
             return false;
-        int rendererTextLength = rendererText.length();
-        int end = (textNode.ptr() == m_endContainer) ? m_endOffset : INT_MAX;
-        int runEnd = std::min(rendererTextLength, end);
+        auto rendererTextLength = static_cast<int>(rendererText.length());
+        auto end = (textNode.ptr() == m_endContainer) ? m_endOffset : INT_MAX;
```
rniwa: Seems like parenthesis here is unnecessary?

Source/WebCore/editing/TextIterator.cpp
```diff
@@ -665,12 +670,21 @@ void TextIterator::handleTextRun()
...
-    unsigned rangeStart = m_offset;
-    auto rangeEnd = std::optional<unsigned> { };
-    if (textNode.ptr() == m_endContainer)
-        rangeEnd = m_endOffset;
+    // For remaining text fragments after a first-letter split, text box offsets are fragment-local but m_offset/m_endOffset are DOM offsets.
+    unsigned remainingFragmentStart = 0;
+    if (auto* renderText = dynamicDowncast<RenderTextFragment>(renderer.get()); renderText && renderText->firstLetter())
+        remainingFragmentStart = renderText->start();
+
+    auto toFragmentLocal = [&](unsigned domOffset) {
```
rniwa: `domOffset` -> `nodeOffset` for consistency.

Source/WebCore/editing/TextIterator.cpp
```diff
@@ -1361,70 +1378,52 @@ bool SimplifiedBackwardsTextIterator::handleTextNode()
 {
     m_lastTextNode = downcast<Text>(m_node.get());
 
-    int startOffset;
-    int offsetInNode;
-    CheckedPtr renderer = handleFirstLetter(startOffset, offsetInNode);
-    if (!renderer)
-        return true;
+    CheckedRef renderer = downcast<RenderText>(*m_node->renderer());
+    auto startOffset = (m_node == m_startContainer) ? m_startOffset : 0;
```
rniwa: Parenthesis here seems unnecessary.

Source/WebCore/editing/TextIterator.cpp
```diff
@@ -1361,70 +1378,52 @@ bool SimplifiedBackwardsTextIterator::handleTextNode()
...
+    unsigned fragmentStart = 0;
+    if (auto* renderText = dynamicDowncast<RenderTextFragment>(renderer.get()); renderText && renderText->firstLetter() && startOffset >= static_cast<int>(renderText->start()))
+        fragmentStart = renderText->start();
 
-CheckedPtr<RenderText> SimplifiedBackwardsTextIterator::handleFirstLetter(int& startOffset, int& offsetInNode)
-{
-    CheckedRef renderer = downcast<RenderText>(*m_node->renderer());
-    startOffset = (m_node == m_startContainer) ? m_startOffset : 0;
+    // After editing mutations (e.g. deleting the last character),
+    // m_positionEndOffset can exceed the text node's current length in DOM space.
```
rniwa: I don't think this makes a lot of sense. We should be updating these offsets before making DOM mutations. What if characters were removed from the middle? Clamping the value to the end of `Text` node won't be correct in that case.

rniwa (approved):

Source/WebCore/editing/TextIterator.cpp
alanbaradlay: ok let me check what fails If we don't do this here and figure out what's going on.

alanbaradlay: You were right, this was just papering over an issue (without that clamp we assert in 3 tests).
The fix is to put the original `SimplifiedBackwardsTextIterator::handleFirstLetter` back (current PR has a lot less red) and now we even progress a couple of more tests. I don't remember why I removed `m_shouldHandleFirstLetter`, I  think it was along the lines of 'we don't have to have a dedicated first letter handling anymore here', but apparently we do.
Anyway `handleFirstLetter` is back and 1, makes the assert go away 2, more progression 3, no mysterious clamping.

Source/WebCore/editing/TextIterator.cpp
```diff
@@ -627,9 +628,13 @@ bool TextIterator::handleTextNode()
...
-        int rendererTextLength = rendererText.length();
-        int end = (textNode.ptr() == m_endContainer) ? m_endOffset : INT_MAX;
-        int runEnd = std::min(rendererTextLength, end);
+        auto rendererTextLength = static_cast<int>(rendererText.length());
+        auto end = textNode.ptr() == m_endContainer ? m_endOffset : INT_MAX;
+        if (textFragmentWithRemainingTextAfterFirstLetter && textFragmentWithRemainingTextAfterFirstLetter->firstLetter()) {
+            runStart = convertNodeOffsetToOffsetInTextFragment(*textFragmentWithRemainingTextAfterFirstLetter, std::max(0, runStart));
+            end = end == INT_MAX ? INT_MAX : static_cast<int>(convertNodeOffsetToOffsetInTextFragment(*textFragmentWithRemainingTextAfterFirstLetter, end));
+        }
+        auto runEnd = std::min(rendererTextLength, end);
```
rniwa: Do we still need this?

rniwa (approved):

PR #62271, opened 2026-04-08
[EME][GStreamer][OCDM] Persistent licenses on CDMThunder

eocanha:
#### e3c739f5d15c05a8c7635b28d7c2bd0ac46d188f
<pre>
[EME][GStreamer][OCDM] Persistent licenses on CDMThunder
<a href="https://bugs.webkit.org/show_bug.cgi?id=311719">https://bugs.webkit.org/show_bug.cgi?id=311719</a>

The mechanism to store/retrieve persistent EME licenses has been defined
on opencdm_construct_session() and opencdm_session_load() in a backwards
compatible way (no API changes, just using the current API in a specific
way). A way to use that feature should be added to CDMThunder.

The way in which this has been defined is that
opencdm_construct_session() would get the previously stored sessionID at
construction time in its CDMData parameter and specifying
PersistentLicense as licenseType. If that session exists, a valid
session will be returned. The actual data load will happen in
opencdm_session_load().

See: <a href="https://github.com/WebPlatformForEmbedded/WPEWebKit/issues/1546">https://github.com/WebPlatformForEmbedded/WPEWebKit/issues/1546</a>

This change adds a mechanism to MediaKeySession to check that no other
active session with the same id for the same security origin exists (a
FIXME pending in the preexisting code). The spec[1] mandates that this
check must be done synchronously at load time, but that&apos;s not enough in
practice. Sessions are created by MediaKeys.createSession() in an
uninitialized state and with an empty session id. In the case of
persistent sessions, that id is populated by load() in a queued task, so
there&apos;s some time after the synchronous check is done and after the
session id is set in MediaKeySession when the set id is still empty.
That can create a race condition when two sessions are trying to load
with the same id. That&apos;s why the check is done again when the queued
task is processed, to give another session loading in parallel time to
set its id, so we can check that no other session is using the same
value we&apos;re trying to set.

The proper code to load the session using OpenCDM API has been added to
CDMInstanceSessionThunder.

[1] <a href="https://www.w3.org/TR/encrypted-media/#dom-mediakeysession-load">https://www.w3.org/TR/encrypted-media/#dom-mediakeysession-load</a> (step 8.3)
&quot;do not create a session if a non-closed session, regardless of type,
already exists for this sanitized session ID in this browsing context&quot;

* Source/WebCore/Modules/encryptedmedia/MediaKeySession.cpp:
(WebCore::MediaKeySession::hasSecurityOrigin const): Added API to check if a session has a specific security origin, without revealing what the session security origin is.
(WebCore::MediaKeySession::load): Immediately reject loading if there&apos;s already another open session with the same id for the same security origin. Capture &quot;origin&quot; (security origin) and pass it to the lambda of the enqueued task that processes the promise and repeat the check later, right before resolving the promise, to avoid the race condition explained before.
* Source/WebCore/Modules/encryptedmedia/MediaKeySession.h: Added hasSecurityOrigin().
* Source/WebCore/Modules/encryptedmedia/MediaKeys.cpp:
(WebCore::MediaKeys::hasOpenSessionWithIdForOrigin const): Check if there&apos;s another open session with the supplied id. It&apos;s a way to check that without revealing the session id to the caller.
* Source/WebCore/Modules/encryptedmedia/MediaKeys.h: Added hasOpenSessionWithIdForOrigin().
* Source/WebCore/platform/graphics/gstreamer/eme/CDMThunder.cpp:
(WebCore::CDMInstanceSessionThunder::CDMInstanceSessionThunder): Removed the assert for zero challengeLength, since persistent license loading can generate a challenge with length 0.
(WebCore::CDMInstanceSessionThunder::challengeGeneratedCallback): Empty challenges (generated by license loading) aren&apos;t processed, since there&apos;s no data to process. Exit early.
(WebCore::CDMInstanceSessionThunder::loadSession): Implement loading by constructing the session passing the session id as CDMData (and its length). Check for error conditions and finally set the session id if loading succeeds. Finally, load the data. Some safety assumptions (initialized and open MediaKeySession, sanitized session id, license type being one of the persistent ones) are guaranteed by the upper WebCore layer and aren&apos;t checked again here. The changed callbacks are kept, since they might be useful if the session changes (renewal is required, etc.). I can&apos;t check this in practice until the real integration is done. I&apos;ve been working with a mockup to check the code until now.

Canonical link: <a href="https://commits.webkit.org/310914@main">https://commits.webkit.org/310914@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/3a22d9c5d4478c84ed55d926c87a5989fab93c9d

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/155298 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/28558 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/21717 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/164059 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/109095 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/157171 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/28698 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/28408 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/120176 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/109095 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/158257 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/22405 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/139469 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/100871 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/21491 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/19567 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/11885 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/131159 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/17302 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/166537 "Built successfully") | 
| | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/10698 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/18912 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/128289 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/28102 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/23601 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/128420 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/34850 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/28026 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/139095 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/85532 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/23276 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/15892 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/27720 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/91823 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/27297 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/27527 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/27370 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/Modules/encryptedmedia/MediaKeySession.cpp
+++ b/Source/WebCore/Modules/encryptedmedia/MediaKeySession.cpp
@@ -132,6 +132,14 @@ const String& MediaKeySession::sessionId() const
     return m_sessionId;
 }
 
+bool MediaKeySession::hasSecurityOrigin(const String& origin) const
+{
+    String sessionOrigin;
+    if (RefPtr document = downcast<Document>(scriptExecutionContext()))
+        sessionOrigin = protect(document->securityOrigin())->toString();
+    return origin == sessionOrigin;
+}
+
 double MediaKeySession::expiration() const
 {
     return m_expiration;
@@ -323,19 +331,27 @@ void MediaKeySession::load(const String& sessionId, Ref<DeferredPromise>&& promi
             return;
         }
 
-        // 8.3. If there is a MediaKeySession object that is not closed in this object's Document whose sessionId attribute is sanitized session ID, reject promise with a QuotaExceededError.
-        // FIXME: This needs a global MediaKeySession tracker.
-
         String origin;
         if (RefPtr document = downcast<Document>(session.scriptExecutionContext()))
-            origin = document->securityOrigin().toString();
+            origin = protect(document->securityOrigin())->toString();
+
+        // 8.3. If there is a MediaKeySession object that is not closed in this object's Document whose sessionId attribute is sanitized session ID, reject promise with a QuotaExceededError.
+        if (session.m_keys && session.m_keys->hasOpenSessionWithIdForOrigin(sanitizedSessionId.value(), WTF::move(origin))) {
+            ERROR_LOG_WITH_THIS(&session, identifier, "Rejected: Another open MediaKeySession with the same session ID already exists for this security origin");
+            promise->reject(ExceptionCode::QuotaExceededError);
+            return;
+        }
 
         // 8.4. Let expiration time be NaN.
         // 8.5. Let message be null.
         // 8.6. Let message type be null.
         // 8.7. Let cdm be the CDM instance represented by this object's cdm instance value.
         // 8.8. Use the cdm to execute the following steps:
-        session.m_instanceSession->loadSession(session.m_sessionType, *sanitizedSessionId, origin, [weakThis = WeakPtr { session }, promise = WTF::move(promise), sanitizedSessionId = *sanitizedSessionId, identifier = WTF::move(identifier)] (std::optional<CDMInstanceSession::KeyStatusVector>&& knownKeys, std::optional<double>&& expiration, std::optional<CDMInstanceSession::Message>&& message, CDMInstanceSession::SuccessValue succeeded, CDMInstanceSession::SessionLoadFailure failure) mutable {
+        session.m_instanceSession->loadSession(session.m_sessionType, *sanitizedSessionId, origin,
+            [weakThis = WeakPtr { session }, promise = WTF::move(promise), sanitizedSessionId = *sanitizedSessionId, identifier = WTF::move(identifier), origin]
+            (std::optional<CDMInstanceSession::KeyStatusVector>&& knownKeys, std::optional<double>&& expiration,
+            std::optional<CDMInstanceSession::Message>&& message, CDMInstanceSession::SuccessValue succeeded,
+            CDMInstanceSession::SessionLoadFailure failure) mutable {
             RefPtr protectedThis = weakThis.get();
             if (!protectedThis) {
                 promise->reject(ExceptionCode::InvalidStateError);
@@ -375,7 +391,17 @@ void MediaKeySession::load(const String& sessionId, Ref<DeferredPromise>&& promi
             }
 
             // 8.9. Queue a task to run the following steps:
-            queueTaskKeepingObjectAlive(*protectedThis, TaskSource::Networking, [knownKeys = WTF::move(knownKeys), expiration = WTF::move(expiration), message = WTF::move(message), sanitizedSessionId, succeeded, promise = WTF::move(promise), identifier = WTF::move(identifier)](auto& session) mutable {
+            queueTaskKeepingObjectAlive(*protectedThis, TaskSource::Networking, [knownKeys = WTF::move(knownKeys), expiration = WTF::move(expiration), message = WTF::move(message), sanitizedSessionId, succeeded, promise = WTF::move(promise), identifier = WTF::move(identifier), origin = WTF::move(origin)](auto& session) mutable {
+                // We must reevaluate condition 8.3 to avoid a race condition in case the id of the sessions
+                // compared before were empty at the moment of comparison because they were in the middle of a
+                // session loading that completed asynchronously. Their ids should have settled now, so let's
+                // check again.
+                if (session.m_keys && session.m_keys->hasOpenSessionWithIdForOrigin(sanitizedSessionId, origin)) {
+                    ERROR_LOG_WITH_THIS(&session, identifier, "Rejected (later): Another open MediaKeySession with the same session ID already exists for this security origin");
+                    promise->reject(ExceptionCode::QuotaExceededError);
+                    return;
+                }
+
                 // 8.9.1. If any of the preceding steps failed, reject promise with a the appropriate error name.
                 if (succeeded == CDMInstanceSession::SuccessValue::Failed) {
                     ERROR_LOG_WITH_THIS(&session, identifier, "::task() Rejected: Other failure");
--- a/Source/WebCore/Modules/encryptedmedia/MediaKeys.cpp
+++ b/Source/WebCore/Modules/encryptedmedia/MediaKeys.cpp
@@ -182,6 +182,15 @@ bool MediaKeys::hasOpenSessions() const
         });
 }
 
+bool MediaKeys::hasOpenSessionWithIdForOrigin(const String& sessionId, const String& origin) const
+{
+    return std::any_of(m_sessions.begin(), m_sessions.end(),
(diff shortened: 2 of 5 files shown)


Source/WebCore/platform/graphics/gstreamer/eme/CDMThunder.cpp
calvaris: Cache in stack the CString resulting of calling sessionID.utf8(). And being at it it would be better to use the length of the CString instead of the String one because String::length gives you the size in UTF16 characters if it were not ASCII so we would be screwing up.

calvaris, line 661: This is an interesting question. Did you check if this runs at all? If not I would remove it. But I suspect it is being called, otherwise we would not be notifying the new available key IDs appearing as result of the load.

calvaris: this should be sessionID, please check the debug build.

calvaris, line 642: Even if you're falling to the default it would be good to add the break here, or at least there used to be some FALLBACK_something macro to avoid compiler issues with this.

Source/WebCore/Modules/encryptedmedia/MediaKeys.cpp
calvaris: Let's capture sessionId and origin by reference.

calvaris (commented):
Commit message typos: "peviously" → "previously"; "trying to a load" → "trying to load".

And let's wait for the result of the callback check I mentioned below.

This is looking good.

Source/WebCore/platform/graphics/gstreamer/eme/CDMThunder.cpp
eocanha: No, I couldn't check because the underlying OpenCDM code that would make the whole feature work hasn't been added yet by the team in charge. Until now, I've only been able to test persistent license storage by adding mockup code downstream (that's how I detected the duplicated session id race condition in [MediaKeySession::load()](https://github.com/WebKit/WebKit/pull/62271/changes/BASE..781d2eaf09261568a04d1ae2ebb0a00a39ed7404#diff-8e5bea46fda04bd2b03aca3f4a50889b8b76b93b26fadee7155e5b9e7913bb24R394)).
This patch is needed in order to integrate everything downstream. Once that happens, I can reevaluate if the response callback is ever needed and remove it in a future commit in case it's not.

Source/WebCore/Modules/encryptedmedia/MediaKeySession.cpp
calvaris: Extra line

calvaris, line 351: You could also move origin here. Sorry that I missed this before.

Source/WebCore/platform/graphics/gstreamer/eme/CDMThunder.cpp
calvaris: "Ignoring empty challenge (maybe persistent license load)"

calvaris: I would print session ID as well here, and maybe (session %p).

calvaris, line 626: Please document the convention here for persistent licenses

calvaris (approved):

Source/WebCore/platform/graphics/gstreamer/eme/CDMThunder.cpp
eocanha: I'm going to use "... empty challenge response ...", since what is being processed in this code is a response.

PR #54, opened 2021-12-08
Bring up another iOS GPU Processes tester

ryanhaddad:
#### ce589ba815cff18d6f6dcebb05337236b021ee7c
<pre>
Bring up another iOS GPU Processes tester
<a href="https://bugs.webkit.org/show_bug.cgi?id=234016">https://bugs.webkit.org/show_bug.cgi?id=234016</a>

Reviewed by NOBODY (OOPS!).

* CISupport/build-webkit-org/config.json: Repurpose the Catalina GPU bot for this queue.
* CISupport/build-webkit-org/factories_unittest.py:
(TestExpectedBuildSteps): Update unit tests.
</pre>

PR #62328, opened 2026-04-09
[Build Speed] Reduce the cost of StructureInlines.h

jernoble:
#### 5ff0c08af8bfe29647b8ae2c25532b18bad58268
<pre>
[Build Speed] Reduce the cost of StructureInlines.h
<a href="https://rdar.apple.com/174363603">rdar://174363603</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=311778">https://bugs.webkit.org/show_bug.cgi?id=311778</a>

Reviewed by Geoffrey Garen.

StructureInlines.h was the 4th most expensive header in the JSC build
(included 180 times, avg 866ms) and 2nd most expensive in the WebCore build
(included 225 times, avg 1096ms). JSCJSValueInlines.h was the 7th most
expensive in WebCore (included 221 times, avg 871ms). The high cost comes
from heavy transitive includes and circular dependencies between core JSC
headers.

This patch reduces these costs through several strategies:

1. Split StructureInlines.h into lighter headers:

   - StructureCreateInlines.h: Contains Structure::create(),
     createStructure(), subspaceFor(), finishCreation(), and
     JSObject::didBecomePrototype(). For code that only defines
     createStructure() methods.

   - StructureArrayStorageInlines.h: Contains
     mayInterceptIndexedAccesses() and holesMustForwardToPrototype(),
     breaking the circular dependency between StructureInlines.h,
     JSObjectInlines.h, and JSArrayInlines.h.

2. Create JSCJSValueCell.h to break the JSCJSValue/JSCell circular
   dependency. This is a regular header (not Inlines) that any header can
   include. It provides JSValue type-checking functions (isString, isObject,
   isBigInt, toPrimitive, toBoolean, etc.) that previously required
   JSCJSValueCellInlines.h. Structure.h includes it so all downstream
   headers get these functions.

3. Rename JSCJSValueCellInlines.h to JSCJSValueStructure.h since it is no
   longer an Inlines header — it just breaks the circular dependency on
   Structure.h, providing JSValue::inherits, classInfoOrNull, and
   structureOrNull.

4. Decompose JSCJSValueInlines.h by moving functions to headers that
   already have the needed types:

   - JSCJSValue.h: getUInt32, equal
   - JSBigInt.h: asHeapBigInt, isZeroBigInt, isNegativeBigInt
   - JSObject.h: put, putInline, putByIndex, getPrototype
   - ExceptionHelpers.h: toIndex, requireObjectCoercible
   - JSGlobalProxy.h: isThisValueAltered
   - New JSCJSValuePropertyInlines.h: all get/getPropertySlot/
     getOwnPropertySlot/getAs overloads (lightweight alternative to
     JSCJSValueInlines.h for property access)

   This removes 7 includes from JSCJSValueInlines.h: Error.h,
   ExceptionHelpers.h, InternalFunction.h, JSGlobalProxy.h, MathCommon.h,
   MakeString.h, JSCJSValueCellInlines.h.

5. Split Operations.h into Operations.h (declarations only) and
   OperationsInlines.h (definitions). Operations.h is reduced from 6 heavy
   includes to just JSExportMacros.h and Forward.h with forward
   declarations. The JSBigInt ComparisonMode and ComparisonResult enums are
   moved to namespace scope (as JSBigIntComparisonMode and
   JSBigIntComparisonResult) to avoid requiring JSBigInt.h.

6. Move JSCell::classInfo() and JSCell::inherits() definitions to
   Structure.h (where Structure is complete). Extract the sweeping
   validation ASSERT to an out-of-line validateIsNotSweeping() in
   JSCell.cpp to avoid pulling vm() into the header.

7. Mark HeapCell::heap() and HeapCell::vm() declarations as inline to catch
   missing HeapCellInlines.h at compile time rather than link time. Remove
   vm() calls from headers by passing VM&amp; parameters
   (FunctionRareData::clearInternalFunctionAllocationProfile,
   JSBoundFunction::name/nameString) or using getVM(globalObject)
   (JSString.h verifyCanGC). Change reportZappedCellAndCrash signature to
   take just const JSCell* to avoid heap() call in JSCast.h.

8. Move Structure::checkOffsetConsistency and checkConsistency to
   Structure.cpp since they are only used in debug builds and only called
   from within Structure.cpp itself.

ClonedArguments::length(), DirectArguments::length(), and
ScopedArguments::length() are moved out-of-line to their respective .cpp
files since their inline definitions called JSObject::get() which was
previously visible through the transitive include chain being broken.

Long (&gt;4 line) inline function definitions in Structure.h are moved to the
bottom of the file to improve API readability, with declarations marked
inline and annotated with &quot;// Defined below&quot;.

After this change, StructureInlines.h drops from the top 5 most expensive
headers to #54 in WebCore, with average cost per include reduced from
~1096ms to ~294ms. In JSC it drops off the top expensive headers list
entirely.

Canonical link: <a href="https://commits.webkit.org/311458@main">https://commits.webkit.org/311458@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/a1ac330b117a7c7b720dcefef731c22ccec2b9be

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/156890 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/30226 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/23417 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/165713 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/110972 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/158761 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/30362 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/30229 "Built successfully") | [❌ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/121515 "Found 1 new test failure: accessibility/aria-owns-deep-chain-no-timeout.html (failure)") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85328 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/159848 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/23749 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/140888 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/102183 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/22803 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/21018 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/13485 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/148940 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/132484 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/18716 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/168198 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/17725 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/12357 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/20336 "Passed tests") | [❌ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/129630 "Found 1 new test failure: accessibility/aria-owns-deep-chain-no-timeout.html (failure)") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/29828 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/25096 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/129737 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35171 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/29751 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/140510 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/87555 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/24566 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/17314 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/188852 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/29461 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/93476 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/48497 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/28984 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/29214 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/29110 "Built successfully") | | | |


Source/JavaScriptCore/runtime/Structure.h
```diff
@@ -630,13 +634,40 @@ class Structure : public JSCell {
...
+            return false;
+        if (hasAnyKindOfGetterSetterProperties())
+            return false;
+        if (isUncacheableDictionary())
+            return false;
+        // Cannot perform fast [[Put]] to |target| if the property names of the |source| contain "__proto__".
+        if (hasUnderscoreProtoPropertyExcludingOriginalProto())
+            return false;
+        return true;
+    }
```
kmiller68: A lot of these definitions are pretty big. Can we move them out of line? It makes the API harder to read when there are so many large inline definitions.

jernoble: I was intentionally not moving things out-of-line if I could avoid it, and the call was plausibly in a fast path somewhere. I assume this was made inline because it needed to be inline; that may have been an invalid assumption, but it's where I was starting at.

jernoble: That said, I could make this inline, but at the bottom of the file.

jernoble: Resolved!

Source/JavaScriptCore/bytecode/ComplexGetStatus.cpp
```diff
@@ -26,7 +26,6 @@
 #include "config.h"
 #include "ComplexGetStatus.h"
 
-#include "StructureInlines.h"
```
darinadler:
```suggestion
```

Source/JavaScriptCore/runtime/ExceptionHelpers.h
```diff
@@ -95,4 +97,46 @@ JS_EXPORT_PRIVATE Exception* throwStackOverflowError(JSGlobalObject*, ThrowScope
         RELEASE_ASSERT_WITH_MESSAGE(assertion, __VA_ARGS__); \
     } while (false)
 
+// Defined here rather than in JSCJSValue.h because they need throwException/createRangeError/createNotAnObjectError.
```
darinadler, line 100: How did you decide on putting these here rather than a separate `JSValueInlinesWithExceptions.h`, so it can be included even less? Seems likely many includers of `ExceptionHelpers.h` aren’t calling these `JSValue` functions.

Source/JavaScriptCore/runtime/JSCJSValue.h
```diff
@@ -1410,4 +1410,28 @@ inline int32_t JSValue::bigInt32AsInt32() const
 }
 #endif // USE(BIGINT32)
 
+ALWAYS_INLINE bool JSValue::getUInt32(uint32_t& v) const
```
darinadler, line 1412: Moving this here doesn’t seem beneficial. I read the change log and while the other moves make sense, this one doesn’t.

Source/JavaScriptCore/runtime/JSCJSValue.h
```diff
@@ -1410,4 +1410,28 @@ inline int32_t JSValue::bigInt32AsInt32() const
...
+    if (isDouble()) {
+        double d = asDouble();
+        v = static_cast<uint32_t>(d);
+        return v == d;
+    }
+    return false;
+}
+
+// ECMA 11.9.3
+inline bool JSValue::equal(JSGlobalObject* globalObject, JSValue v1, JSValue v2)
```
darinadler, line 1428: Moving this here doesn’t seem beneficial. I read the change log and while the other moves make sense, this one doesn’t.

Source/JavaScriptCore/runtime/JSCJSValueInlines.h
```diff
@@ -29,21 +29,15 @@
 
 WTF_ALLOW_UNSAFE_BUFFER_USAGE_BEGIN
 
-#include <JavaScriptCore/Error.h>
-#include <JavaScriptCore/ExceptionHelpers.h>
 #include <JavaScriptCore/Identifier.h>
-#include <JavaScriptCore/InternalFunction.h>
 #include <JavaScriptCore/JSBigInt.h>
```
darinadler, line 33: I don’t understand why this is still here. We moved some inline functions from here to `JSBigInt.h`, and that’s not valuable if this includes `JSBigInt.h`!

darinadler (commented):
A few initial comments. I’ll review again later.

geoffreygaren (approved):
This looks sensible to me and EWS passes, and @kmiller68 already took a review pass, so I'll say r+.

...oh, but also please address Darin's comments, which appeared concurrently.

Source/JavaScriptCore/runtime/ExceptionHelpers.h
jernoble: I only made new headers when necessary to break an include cycle. It's certainly a valid strategy to make a new `JSCJSValueExceptionHelpers.h` file that only includes both `JSCJSValue.h` and `ExceptionHelpers.h` and move these inlines there, but I biased towards using existing headers if doing so didn't require additional includes.

darinadler: I do (and did) understand that you made that choice, but not why. Since our goal is faster compilation, it might be worthwhile to minimize any increase in size of existing headers, depending on how widely included they are.

Source/JavaScriptCore/runtime/JSCJSValue.h
jernoble: These were moved back into `JSCJSValue.h` from `JSCJSValueInlines.h` because they have no dependencies on any header other than `JSCJSValue.h`. Requiring callers to import `JSCJSValueInlines.h` means this function can't be called in other non-Inline.h inlined functions. (It is called from `CommonSlowPaths.h`.)

darinadler: Seems like this will slow down any translation unit that includes `JSCJSValue.h` and doesn’t include `CommonSlowPaths.h`. Maybe that’s an obviously good tradeoff for the benefits you describe above. Hard for me to do the math in my head.

Source/JavaScriptCore/runtime/JSCJSValue.h
jernoble: This was moved back in for the same reason as `getUInt32()`, though there are no existing header-file callers. The benefit of not having to pull in `JSCJSValueInlines.h` is limited to only a few .cpp files.

Source/JavaScriptCore/runtime/ExceptionHelpers.h
jernoble: My goal was not an absolute reduction in size of every header, but to reduce the overall time spent during parsing. Moving a declaration from a more widely included header to a less widely included one met that goal. That said, I fully intend to iterate, and we'll definitely have another chance to take a bite of this particular apple.

Source/JavaScriptCore/runtime/JSCJSValue.h
jernoble: IMO, reducing the cost of the header itself is a different strategy than reducing the transitive cost of all that header's headers. This patch is about the latter, not the former. The former is still a valid strategy! But my intuition says that this function definition won't meaningfully increase the parse time of `JSCJSValue.h`, and ClangBuildAnalyzer output tends to agree.

Source/JavaScriptCore/runtime/JSCJSValueInlines.h
jernoble: It's still valuable if `JSCJSValueInlines.h` is included in fewer places, which it is, and specifically of callers of those methods which were moved don't need to include `JSCJSValueInlines.h` to do so.

If memory serves, removing `JSBigInt.h` caused unrelated build errors for files which were transitively getting that header.  Claude reminds me that `equalSlowCaseInline`, `strictEqualForCells`, `strictEqual`, and `pureStrictEqual` still require `JSBigInt.h`. But they themselves can't be moved into `JSBigInt.h` because they have other, non-JSBigInt requirements.

darinadler: I don’t follow.

Here’s my feedback: We moved `asHeapBigInt` to `JSBigInt.h`, making that header bigger. Moving it did not allow us to remove an include from `JSCJSValueInlines.h`. That change will not make the build faster. I recommend not making that change.

darinadler: Are you saying that `asHeapBigInt` is used in files that no longer include `JSCJSValueInlines.h`? Do we have an example of one?

jernoble: Yes. ParserArena.cpp is one. It includes `JSBigInt.h`, it calls `asHeapBigInt()`, and it used to get `JSCJSValueInlines.h` transitively, but doesn't need it.

jernoble: > Moving it did not allow us to remove an include from `JSCJSValueInlines.h`. That change will not make the build faster.

It makes the build faster by having fewer files need to include `JSCJSValueInlines.h`, specifically files that already included `JSBigInt.h`, because they needed to perform operations on the return value of `asHeapBigInt()`.

This patch wasn't really about reducing the number of places that included `JSCJSValueInlines.h`; all the changes in it were compiler and linker error fallout from the changes to `StructureInlines.h`. And the biggest change to `StructureInlines.h` was the removal of an include of `JSCJSValueInline.h`. 

Before this patch, `StructureInlines.h` was the 4th most expensive header in JSC, and `JSCJSValueInlines.h` was 2nd.  Afterward, `StructureInlines.h` isn't even in the top 1000, and `JSCJSValueInlines.h` is ... still 2nd. That just means we still have work to do, not that this is pointless or counterproductive.

jernoble: ....All that said, making a new file called `JSCJSValueBigInt.h` to put these definitions in would be fine too, and would appear to address your feedback, so I'll just do that.

jernoble: The new patch adds `JSCJSValueBigInt.h`, but in so doing, I discovered that 2 of the 3 functions to be moved to that header were only used in a single .cpp file. So I refactored those as static methods and moved them into the one file that needed them. `JSCJSValueBigInt.h` only has a single method in it, `asHeapInt32()` which is largely duplicative with the `JSBigInt.h` method `asHeapInt32(JSValue)`. One way to achieve the net-zero increase in `JSBigInt.h` would have been to remove that method.

jernoble: And this new patch cuts down frontend parser time by 317s, or about 5m, mostly through reductions in both the average time spent parsing `StructureInlines.h` (6.2m -> 1.7m) and the number of times that file was parsed (326 -> 260). And because `StructureInlines.h` was included via `JSCJSValueInlines.h`, nearly identical decreases were seen in the number and average time spent parsing that header as well.

All in all, `JSCJSValueInlines.h` dropped from 9th most expensive header to 44th, `StructureInlines.h` dropped from 10th to 74th.

PR #62423, opened 2026-04-10
[BFCache] Add multi-process suspension for BFCache with Site Isolation

basuke:
#### 41894db0ac68ab9f13087416e2579726077cee8f
<pre>
[BFCache] Add multi-process suspension for BFCache with Site Isolation
<a href="https://bugs.webkit.org/show_bug.cgi?id=311224">https://bugs.webkit.org/show_bug.cgi?id=311224</a>
<a href="https://rdar.apple.com/173743585">rdar://173743585</a>

Reviewed by Sihui Liu.

When site isolation is active and multi-process BFCache is enabled, suspend
subframe processes alongside the main frame when a page enters the
back/forward cache. Each subframe process receives a SetSubframesSuspended
IPC — along with the main frame&apos;s BackForwardFrameItemIdentifier — that
caches its page in the WebCore BackForwardCache. A CallbackAggregator tracks
completion across all processes; if any fails, the BFCache entry is removed
and teardown cleans up the subframe processes.

The UIProcess passes fromItem-&gt;mainFrameItem().identifier() to each iframe
process. This identifier is unique per navigation entry, stable across iframe
cross-site navigations, and the same for all iframe processes — no
per-process frame tree traversal needed. Each iframe process stores its
CachedPage under this key in its own per-process BackForwardCache singleton.

Also refactor BackForwardCache::addIfCacheable to share logic between the
existing HistoryItem-based overload and the new identifier-based overload
used by iframe processes.

Tests: Tools/TestWebKitAPI/Tests/WebKit/WKWebView/SiteIsolation.mm

* Source/WebCore/history/BackForwardCache.cpp:
(WebCore::BackForwardCache::addIfCacheable):
* Source/WebCore/history/BackForwardCache.h:
(WebKit::BrowsingContextGroup::removePage):
(WebKit::BrowsingContextGroup::closeRemotePagesForPage):
(WebKit::BrowsingContextGroup::hasMultiplePages const):
* Source/WebKit/UIProcess/BrowsingContextGroup.h:
* Source/WebKit/UIProcess/SuspendedPageProxy.cpp:
(WebKit::SuspendedPageProxy::findReusableSuspendedPageProcess):
(WebKit::SuspendedPageProxy::SuspendedPageProxy):
(WebKit::SuspendedPageProxy::startSuspension):
(WebKit::SuspendedPageProxy::~SuspendedPageProxy):
(WebKit::SuspendedPageProxy::teardown):
(WebKit::SuspendedPageProxy::waitUntilReadyToUnsuspend):
(WebKit::SuspendedPageProxy::didProcessRequestToSuspend):
(WebKit::SuspendedPageProxy::suspensionTimedOut):
(WebKit::SuspendedPageProxy::suspendSubframeProcesses):
(WebKit::SuspendedPageProxy::hasSubframeInProcess const):
(WebKit::SuspendedPageProxy::maybeCompleteSuspension):
* Source/WebKit/UIProcess/SuspendedPageProxy.h:
* Source/WebKit/UIProcess/WebBackForwardCache.cpp:
(WebKit::WebBackForwardCache::removeEntriesForProcess):
(WebKit::WebPageProxy::suspendCurrentPageIfPossible):
(WebKit::WebPage::setSubframesSuspended):
* Tools/TestWebKitAPI/Helpers/cocoa/TestNavigationDelegate.h:
* Tools/TestWebKitAPI/Helpers/cocoa/TestNavigationDelegate.mm:
(-[TestNavigationDelegate waitForDidFinishNavigationAndLoadInSubframe]):
(TestWebKitAPI::(SiteIsolation, MultiProcessBFCacheIframeProcessSurvival)):
(TestWebKitAPI::(SiteIsolation, MultiProcessBFCacheOpenerSkipsBFCache)):

Canonical link: <a href="https://commits.webkit.org/311727@main">https://commits.webkit.org/311727@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/df2d66250fc70f3c2755308b471878a59b9f0dcf

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/157824 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/31161 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/24354 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/166654 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/111906 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/159695 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/31296 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/31163 "Built successfully") | [  ~~🧪 wpe-wk2~~](https://ews-build.webkit.org/#/builders/34/builds/122218 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/111906 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/160782 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/24497 "Passed tests") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/141725 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-wpe~~](https://ews-build.webkit.org/#/builders/41/builds/102884 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | ⏳ 🛠 vision-apple 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/23553 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/21840 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/14419 "Built successfully") | | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/133234 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/19535 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/169148 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/13898 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/21157 "Passed tests") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/1/builds/130390 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [  ~~🛠 vision~~](https://ews-build.webkit.org/#/builders/153/builds/30907 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/25909 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-gtk~~](https://ews-build.webkit.org/#/builders/21/builds/130503 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35336 "Built successfully and passed tests") | [  ~~🛠 vision-sim~~](https://ews-build.webkit.org/#/builders/160/builds/30845 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/141329 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/88695 "Built successfully") | | 
| | [  ~~🧪 vision-wk2~~](https://ews-build.webkit.org/#/builders/160/builds/30845 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/18135 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/30397 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/95230 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/29918 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/30148 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/30045 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/history/BackForwardCache.cpp
+++ b/Source/WebCore/history/BackForwardCache.cpp
@@ -530,38 +530,40 @@ std::unique_ptr<CachedPage> BackForwardCache::trySuspendPage(Page& page, ForceSu
     return makeUnique<CachedPage>(page);
 }
 
-bool BackForwardCache::addIfCacheable(HistoryItem& item, Page* page)
+bool BackForwardCache::addIfCacheable(BackForwardFrameItemIdentifier identifier, Page& page)
 {
-    if (item.isInBackForwardCache())
+    if (isInBackForwardCache(identifier))
         return false;
 
-    if (!page)
-        return false;
-
-    // FIXME (rdar://173799983): Remove this workaround once the UIProcess has a
-    // mechanism to detach RemotePageProxy objects from the BrowsingContextGroup
-    // during in-process (same-origin) navigations. Cross-process suspension
-    // (suspendPage with ForceSuspension::Yes) already handles this via
-    // BrowsingContextGroup exchange in suspendCurrentPageIfPossible.
-    if (page->settings().siteIsolationEnabled())
-        return false;
-
-    auto cachedPage = trySuspendPage(*page, ForceSuspension::No);
+    auto cachedPage = trySuspendPage(page, ForceSuspension::No);
     if (!cachedPage)
         return false;
 
     {
-        // Make sure we don't fire any JS events in this scope.
         ScriptDisallowedScope::InMainThread scriptDisallowedScope;
-
-        m_cachedPageMap.set(item.frameItemID(), makeUniqueRefFromNonNullUniquePtr(WTF::move(cachedPage)));
-        m_items.add(item.frameItemID());
-        item.notifyChanged();
+        m_cachedPageMap.set(identifier, makeUniqueRefFromNonNullUniquePtr(WTF::move(cachedPage)));
+        m_items.add(identifier);
     }
     prune(PruningReason::ReachedMaxSize);
+    RELEASE_LOG(BackForwardCache, "BackForwardCache::addIfCacheable frameItemID: %s, size: %u / %u", identifier.toString().utf8().data(), pageCount(), maxSize());
+    return true;
+}
 
-    RELEASE_LOG(BackForwardCache, "BackForwardCache::addIfCacheable item: %s, size: %u / %u", item.itemID().toString().utf8().data(), pageCount(), maxSize());
+bool BackForwardCache::addIfCacheable(HistoryItem& item, Page* page)
+{
+    if (!page)
+        return false;
 
+    // Same-site BFCache is not yet supported under Site Isolation — same-origin
+    // navigations can leave stale RemotePageProxy objects in the BCG.
+    // Cross-site (multi-process) BFCache uses the other addIfCacheable overload
+    // which is only called when both SI and multiProcessBackForwardCache are enabled.
+    if (page->settings().siteIsolationEnabled())
+        return false;
+
+    if (!addIfCacheable(item.frameItemID(), *page))
+        return false;
+    item.notifyChanged();
     return true;
 }
 
--- a/Source/WebKit/UIProcess/SuspendedPageProxy.cpp
+++ b/Source/WebKit/UIProcess/SuspendedPageProxy.cpp
@@ -33,6 +33,7 @@
 #include "HandleMessage.h"
 #include "Logging.h"
 #include "MessageSenderInlines.h"
+#include "RemotePageProxy.h"
 #include "WebBackForwardCache.h"
 #include "WebBackForwardList.h"
 #include "WebBackForwardListMessages.h"
@@ -42,6 +43,7 @@
 #include "WebPageProxyMessages.h"
 #include "WebProcessMessages.h"
 #include "WebProcessPool.h"
+#include <wtf/CallbackAggregator.h>
 #include <wtf/DebugUtilities.h>
 #include <wtf/HexNumber.h>
 #include <wtf/TZoneMallocInlines.h>
@@ -64,6 +66,8 @@ WTF_MAKE_TZONE_ALLOCATED_IMPL(SuspendedPageProxy);
 RefPtr<WebProcessProxy> SuspendedPageProxy::findReusableSuspendedPageProcess(WebProcessPool& processPool, const RegistrableDomain& registrableDomain, WebsiteDataStore& dataStore, WebProcessProxy::LockdownMode lockdownMode, EnhancedSecurity enhancedSecurity, const API::PageConfiguration& pageConfiguration)
 {
     for (Ref suspendedPage : allSuspendedPages()) {
+        if (!suspendedPage->hasSuspensionStarted())
+            continue;
         Ref process = suspendedPage->process();
         if (&process->processPool() == &processPool
             && process->site() && process->site()->domain() == registrableDomain
@@ -135,15 +139,6 @@ SuspendedPageProxy::SuspendedPageProxy(WebPageProxy& page, Ref<WebProcessProxy>&
 #endif
 {
     allSuspendedPages().add(*this);
-    m_process->addSuspendedPageProxy(*this);
-    m_messageReceiverRegistration.startReceivingMessages(m_process, m_webPageID, *this, *this);
-    m_suspensionTimeoutTimer.startOneShot(suspensionTimeout);
-    sendWithAsyncReply(Messages::WebPage::SetIsSuspended(true), [weakThis = WeakPtr { *this }](std::optional<bool> didSuspend) {
-        RefPtr protectedThis = weakThis.get();
-        if (!protectedThis || !didSuspend)
-            return;
-        protectedThis->didProcessRequestToSuspend(*didSuspend ? SuspensionState::Suspended : SuspensionState::FailedToSuspend);
-    });
 }
 
 template<typename M>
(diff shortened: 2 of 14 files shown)


basuke:
**Note:** This PR covers **suspension only**. Restoration (goBack) will be in a follow-up PR that adds:

- `unsuspend()` flow for iframe processes: reinitialize WebPage, restore CachedFrame tree, reconnect BrowsingContextGroup
- Bypass `BackForwardCache::addIfCacheable()` SI guard for iframe processes (rdar://173799983)
- `MultiProcessBFCacheIframeRestoration` test — verifies frame tree + process PID preservation after goBack

---

**Update (latest push):** Fixed `MultiProcessBFCacheIframeProcessSurvival` iOS EWS failure — the iframe's back/forward list frame item wasn't populated yet when navigating to c.com. Added `waitForDidFinishNavigationAndLoadInSubframe` to `TestNavigationDelegate` to atomically wait for both main frame navigation and subframe load completion.

Source/WebKit/UIProcess/SuspendedPageProxy.cpp
szewai: I think the original code falls through to case SuspensionState::Suspended: so the change is not needed

Source/WebKit/WebProcess/WebPage/WebPage.cpp
szewai: `addIfCacheable` calls `trySuspendPage(*page, ForceSuspension::No)` which suspends the entire Page and deconstructs the frame tree on the first call, making subsequent iterations incorrect.

Why do we need to cache each frame?  `CacheFrame` will cache its children on construction, so caching the main frame should suffice?

basuke: Good catch. You're right on both counts — the loop is broken and unnecessary.

`addIfCacheable` → `trySuspendPage` suspends the entire Page and deconstructs the frame tree on the first call, so subsequent `traverseNext()` iterations would be walking a dismantled tree. And as you noted, `CachedFrame` recursively caches children, so a single call suffices.

Looking deeper, I think the root issue is that `SetSubframesSuspended` IPC doesn't pass any identifier — the iframe process has to find a HistoryItem on its own, which led to this loop. But `BackForwardCache` internally already keys on `BackForwardFrameItemIdentifier`, and `ClearCachedPage` already passes the identifier directly.

I'll fix this by having the UIProcess pass the `BackForwardFrameItemIdentifier` in the IPC, then call `trySuspendPage` once and store the CachedPage under that identifier — no loop, no HistoryItem lookup needed.

Source/WebKit/UIProcess/SuspendedPageProxy.cpp
basuke: Right. After the refactoring, this became the same code.

Source/WebCore/history/BackForwardCache.cpp
szewai: Why is it ok to add `!page.settings().multiProcessBackForwardCacheEnabled()` for `bool BackForwardCache::addIfCacheable(HistoryItem& item, Page* page) `case?
It seems `addIfCacheable(HistoryItem..)` currently is also invoked by `FrameLoader::commitProvisionalLoad()`, which means same-process navigation will invoke it; but we should avoid cache under Site Isolation unless UI process agrees? (i.e. we should add the check when the function is invoked by IPC message from UI process, not in-process).

Source/WebKit/UIProcess/SuspendedPageProxy.cpp
szewai: `startSuspension` has to be invoked after `SuspendedPageProxy` is created, so I think we should keep it in constructor and add parameter `WebBackForwardListItem*` to constructor instead -- then we can avoid risk of callers forget invoking the additional function.

szewai: protect is not needed because `m_browsingContextGroup` is const Ref

Source/WebKit/UIProcess/WebBackForwardCache.cpp
szewai: Nit: keep capturing `processIdentifier..` instead of `&` unless we need other variables.

Source/WebKit/UIProcess/SuspendedPageProxy.cpp
basuke: I can't agree with that. Constructers should basically not do anything other than initialization. 

Since there is a state that manages the suspension status, we should use it properly to handle the management. Looking at it now, however, the status prior to the suspension start is currently not being managed. We should add that.

Source/WebCore/history/BackForwardCache.cpp
basuke: For same-site navigations (no process swap), `FrameLoader::commitProvisionalLoad()` → `addIfCacheable()` is the only caching opportunity — `suspendCurrentPageIfPossible()` is not called in that path. If we block caching here under SI, same-site navigations would lose BFCache entirely.

BFCache caching decisions have historically been made by the WebProcess; the UIProcess controls only the cross-process suspension path via `suspendCurrentPageIfPossible()`. Making the UIProcess gate all caching decisions (including same-process) would be a larger architectural change beyond the scope of this patch.

szewai: My point is, b/f cache under SI (both same-site and cross-site navigation) is not supported before this PR, so the original condition in `addIfCacheable` is:
```
if (page->settings().siteIsolationEnabled())
        return false;
```

This PR fixes the cross-site navigation b/f cache under SI with `multiProcessBackForwardCacheEnabled`, so it changes the condition to:
`if (page.settings().siteIsolationEnabled() && !page.settings().multiProcessBackForwardCacheEnabled())`

However, the PR does not fix same-site navigation under SI (this will be a bigger change as you mentioned), so it should not update the condition for same-site navigation: i.e. `bool BackForwardCache::addIfCacheable(HistoryItem& item, Page* page)` should still have check:
```
if (page->settings().siteIsolationEnabled())
        return false;
```
and not fall back to use the updated check `bool BackForwardCache::addIfCacheable(BackForwardFrameItemIdentifier identifier, Page& page)`

Or are you saying the same-site navigation b/f cache will just work with this PR?

basuke: Ah, I got the point. Same site navigation case should be denied as-is. Right.

Source/WebKit/UIProcess/SuspendedPageProxy.cpp
szewai: I could understand you want to make constructor lightweight by moving the initialization out, but the risk is we will have a partially initialized `SuspendedPageProxy` if caller forgets to invoke `startSuspension` after creating it. I see you added a new state `BeforeStart`, but you don't add the necessary checks for this new state in all functions, specifically `SuspendedPageProxy::findReusableSuspendedPageProcess` and `SuspendedPageProxy::teardown`.

For `findReusableSuspendedPageProcess`, in `BeforeStart` state, the object is in `allSuspendedPages` but not registered with its process (because you move `m_process->addSuspendedPageProxy(*this);` to `startSuspension`).

For `teardown`, in `BeforeStart` state, `removeSuspendedPageProxy` would try to remove something that was never added.

So I would still recommend keeping the original implementation for simplicity; or update all the other functions to ensure there are check for the new states (i.e. to either skip the unnecessary steps if `startSuspension` is not called or to release assert `startSuspension` must be called)

basuke: Got it. Still I don't agree to keep it as-is so I'll do the right way to complete the state management in all place.

szewai: Yeah I think it's also acceptable to add state checks in corresponding functions

szewai (approved):

PR #62481, opened 2026-04-10
[Linux] Use QOS::UserInitiated for NetworkProcess' Cache Storage thread

spenap:
#### c446a0f81300c097ee5be9b75c292e0175e6fa87
<pre>
[Linux] Use QOS::UserInitiated for NetworkProcess&apos; Cache Storage thread
<a href="https://bugs.webkit.org/show_bug.cgi?id=311975">https://bugs.webkit.org/show_bug.cgi?id=311975</a>

This conditionally uses QOS::UserInitiated on Linux for the NetworkProcess&apos;
Cache Storage thread, which maps into SCHED_OTHER. Keeping UserInteractive,
which maps to SCHED_RR, would consistently hit RT limits under moderate load.

Tested on a debug build, confirming that the NetworkProcess thread is no
longer promoted (nor subsequently demoted, as before)

* Source/WebKit/NetworkProcess/cache/NetworkCacheStorage.cpp:
(WebKit::NetworkCache::Storage::Storage):

Canonical link: <a href="https://commits.webkit.org/311064@main">https://commits.webkit.org/311064@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/0de8a1ebcd8099c3f9f449c7c826af3fcf20690e

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/155814 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/29072 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/22231 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/164576 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/109628 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/29219 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/28922 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/120609 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/84972 "1 flakes 3 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/158771 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/22791 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/139922 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/101298 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/21877 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/20022 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/12406 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/131546 "Passed tests") | | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/167056 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/11230 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/19365 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/128729 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/28616 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/24056 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/128861 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/34937 "Built successfully and passed tests") | [loading 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/28540 "Build is in progress. Recent messages:OS: Tahoe (26.2), Xcode: 26.2; Running apply-patch; Checked out pull request; Compiled WebKit (warnings)") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/139547 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/86388 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/23684 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/16345 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/28234 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/92337 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/27811 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/28041 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/27884 "Built successfully") | | |

### DIFF
--- a/Source/WebKit/NetworkProcess/cache/NetworkCacheStorage.cpp
+++ b/Source/WebKit/NetworkProcess/cache/NetworkCacheStorage.cpp
@@ -406,7 +406,13 @@ Storage::Storage(const String& baseDirectoryPath, Mode mode, Salt salt, size_t c
     , m_capacity(capacity)
     , m_readOperationTimeoutTimer(*this, &Storage::cancelAllReadOperations)
     , m_writeOperationDispatchTimer(*this, &Storage::dispatchPendingWriteOperations)
-    , m_ioQueue(ConcurrentWorkQueue::create("com.apple.WebKit.Cache.Storage"_s, WorkQueue::QOS::UserInteractive))
+    , m_ioQueue(ConcurrentWorkQueue::create("com.apple.WebKit.Cache.Storage"_s,
+#if OS(LINUX)
+                WorkQueue::QOS::UserInitiated
+#else
+                WorkQueue::QOS::UserInteractive
+#endif
+                ))
     , m_backgroundIOQueue(ConcurrentWorkQueue::create("com.apple.WebKit.Cache.Storage.background"_s, WorkQueue::QOS::Utility))
     , m_serialBackgroundIOQueue(WorkQueue::create("com.apple.WebKit.Cache.Storage.serialBackground"_s, WorkQueue::QOS::Utility))
     , m_blobStorage(makeBlobDirectoryPath(baseDirectoryPath), m_salt, mainResourceBlobMemoryCacheFileLimit)


spenap:
See https://commits.webkit.org/292576@main when the Cache Storage thread's QOS was updated

mcatanzaro:
I think it should use QOS::UserInitiated on all platforms. Semantically, if a thread is UserInteractive, then it absolutely must not perform long-running blocking operations. There's nothing platform-specific about that expectation. So the Linux semantics of QOS::UserInteractive align with reasonable developer expectations, and we shouldn't need to make Linux-specific decisions when setting the QOS.

Maybe we should add a warning about the kill behavior to the documentation of the QOS enum, though, because Apple developers probably won't expect that it can cause the entire process to be killed if it blocks for too long.

mcatanzaro:
I assume CacheStorage thread is intentionally performing blocking I/O. Blocking implies it must not be UserInteractive.

(If the thread is _not_ expected to block, then perhaps we need to investigate why that is happening. But I assume it's probably expected.)

Source/WebKit/NetworkProcess/cache/NetworkCacheStorage.cpp
```diff
@@ -406,7 +406,13 @@ Storage::Storage(const String& baseDirectoryPath, Mode mode, Salt salt, size_t c
     , m_capacity(capacity)
     , m_readOperationTimeoutTimer(*this, &Storage::cancelAllReadOperations)
     , m_writeOperationDispatchTimer(*this, &Storage::dispatchPendingWriteOperations)
-    , m_ioQueue(ConcurrentWorkQueue::create("com.apple.WebKit.Cache.Storage"_s, WorkQueue::QOS::UserInteractive))
+    , m_ioQueue(ConcurrentWorkQueue::create("com.apple.WebKit.Cache.Storage"_s,
+#if OS(LINUX)
+                WorkQueue::QOS::UserInitiated
```
carlosgcampos, line 411: Why not WorkQueue::QOS::Utility? it's the same in linux, but it's consistent with the other queues

carlosgcampos:
> I think it should use QOS::UserInitiated on all platforms. Semantically, if a thread is UserInteractive, then it absolutely must not perform long-running blocking operations. There's nothing platform-specific about that expectation. So the Linux semantics of QOS::UserInteractive align with reasonable developer expectations, and we shouldn't need to make Linux-specific decisions when setting the QOS.
> 
> Maybe we should add a warning about the kill behavior to the documentation of the QOS enum, though, because Apple developers probably won't expect that it can cause the entire process to be killed if it blocks for too long.

We can't compare QOS values used in apple platform with linux, because in linux we can only make the thread real time or not, and it has many limitations. So, we can't make all threads marked a user interactive as real time in the end.

Source/WebKit/NetworkProcess/cache/NetworkCacheStorage.cpp
spenap: Because `QOS_CLASS_USER_INITIATED` is the second-highest priority in Darwin, so this could live as an #ifdef temporarily, and then they might be keen to use the same QoS as us. If that happened, our mapping would be non-Real Time.

I guess the choice depends on whether we assume we will diverge here in the long term -- in which case we can use Utility -- or whether there's a chance they accept `UserInitiated` -- in which case we could drop the ifdef

carlosgcampos (approved):

PR #62624, opened 2026-04-13
Improve <img>'s naturalWidth & naturalHeight with SVG images

annevk:
#### 7a59ef69ef562614ed25ae6bd54081a674e44d23
<pre>
Improve &lt;img&gt;&apos;s naturalWidth &amp; naturalHeight with SVG images
<a href="https://bugs.webkit.org/show_bug.cgi?id=284341">https://bugs.webkit.org/show_bug.cgi?id=284341</a>
<a href="https://rdar.apple.com/141196049">rdar://141196049</a>

Reviewed by Said Abou-Hallawa (OOPS\!).

For SVG images that lack some intrinsic dimensions, we need to use the
300 x 150 CSS defaults for naturalWidth &amp; naturalHeight. However,
srcset&apos;s density should only scale intrinsic dimensions, not these
defaults. This is standardized in HTML and more background is
available here:

    <a href="https://github.com/whatwg/html/issues/11287">https://github.com/whatwg/html/issues/11287</a>

To fix this we add SVGImage::resolvedIntrinsicSize(density) which
resolves missing dimensions using the defaults and the aspect ratio, if
any, applying density only to intrinsic dimensions. We add
SVGImage::hasIntrinsicWidth/Height() (and the corresponding Image
virtuals) so callers can query which dimensions exist.

We thread a density parameter through CachedImage&apos;s sizing methods
(separate from the existing multiplier) so that
StyleCachedImage::imageSize() can pass it from the renderer&apos;s
imageDevicePixelRatio to internalImageSizeForRenderer().

Canonical link: <a href="https://commits.webkit.org/312552@main">https://commits.webkit.org/312552@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/437841de86d4cb3b93473484e2824beb71353276

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/160309 "4 style errors") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/33782 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/26883 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/169211 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/115375 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/d2b4a120-eff8-4817-827c-507f72ef2034) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/162178 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/33883 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/33782 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/124289 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/87816 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/e4b8e30e-1522-4d1d-9e7d-1a3d7c4d3e3f) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/163267 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/26527 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/143994 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/104888 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/afb60bcf-2f87-4fd0-b19d-176906c29592) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/25579 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/24080 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/16931 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/135272 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/21757 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/171689 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/18047 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/23397 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/132539 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/33455 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/28174 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/132563 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35839 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/33440 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/143552 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/95222 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/27177 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/20366 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/32949 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/99710 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/32450 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/32694 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/32598 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/html/HTMLImageElement.cpp
+++ b/Source/WebCore/html/HTMLImageElement.cpp
@@ -656,7 +656,7 @@ unsigned HTMLImageElement::width()
 
         // if the image is available, use its width
         if (RefPtr image = m_imageLoader->image())
-            return image->imageSizeForRenderer(nullptr, 1.0f).width().toUnsigned();
+            return image->imageSizeForRenderer(nullptr, 1.0f, CachedImage::IntrinsicSize).width().toUnsigned();
     }
 
     CheckedPtr box = renderBox();
@@ -679,7 +679,7 @@ unsigned HTMLImageElement::height()
 
         // if the image is available, use its height
         if (RefPtr image = m_imageLoader->image())
-            return image->imageSizeForRenderer(nullptr, 1.0f).height().toUnsigned();
+            return image->imageSizeForRenderer(nullptr, 1.0f, CachedImage::IntrinsicSize).height().toUnsigned();
     }
 
     CheckedPtr box = renderBox();
@@ -689,29 +689,20 @@ unsigned HTMLImageElement::height()
     return Style::adjustLayoutUnitForAbsoluteZoom(contentRect.height(), *box).round();
 }
 
-float HTMLImageElement::effectiveImageDevicePixelRatio() const
-{
-    RefPtr cachedImage = m_imageLoader->image();
-    if (!cachedImage)
-        return 1.0f;
-
-    RefPtr image = cachedImage->image();
-    if (image && image->drawsSVGImage())
-        return 1.0f;
-
-    return m_imageDevicePixelRatio;
-}
-
 unsigned HTMLImageElement::naturalWidth() const
 {
     RefPtr image = m_imageLoader->image();
-    return image ? image->unclampedImageSizeForRenderer(protect(renderer()).get(), effectiveImageDevicePixelRatio()).width().toUnsigned() : 0;
+    if (!image)
+        return 0;
+    return image->unclampedImageSizeForRenderer(protect(renderer()).get(), 1.0f, CachedImage::IntrinsicSize, m_imageDevicePixelRatio).width().toUnsigned();
 }
 
 unsigned HTMLImageElement::naturalHeight() const
 {
     RefPtr image = m_imageLoader->image();
-    return image ? image->unclampedImageSizeForRenderer(protect(renderer()).get(), effectiveImageDevicePixelRatio()).height().toUnsigned() : 0;
+    if (!image)
+        return 0;
+    return image->unclampedImageSizeForRenderer(protect(renderer()).get(), 1.0f, CachedImage::IntrinsicSize, m_imageDevicePixelRatio).height().toUnsigned();
 }
 
 bool HTMLImageElement::isURLAttribute(const Attribute& attribute) const
--- a/Source/WebCore/loader/cache/CachedImage.cpp
+++ b/Source/WebCore/loader/cache/CachedImage.cpp
@@ -308,39 +308,54 @@ void CachedImage::setContainerContextForClient(const CachedImageClient& client, 
     m_svgImageCache->setContainerContextForClient(client, containerSize, containerZoom, imageURL);
 }
 
-FloatSize CachedImage::imageSizeForRenderer(const RenderElement* renderer, SizeType sizeType) const
+FloatSize CachedImage::internalImageSizeForRenderer(const RenderElement* renderer, float multiplier, SizeType sizeType, float density) const
 {
     RefPtr image = m_image;
     if (!image)
         return { };
 
+    if (RefPtr svgImage = dynamicDowncast<SVGImage>(*image)) {
+        FloatSize size;
+        if (sizeType == UsedSize) {
+            // The SVG cache size is already in CSS pixels (set by the layout
+            // system via setContainerContextForClient), so density does not apply.
+            size = m_svgImageCache->imageSizeForRenderer(renderer);
+        } else
+            size = svgImage->resolvedIntrinsicSize(density);
+        if (multiplier != 1.0f)
+            size.scale(multiplier);
+        return size;
+    }
+
+    FloatSize imageSize;
 #if ENABLE(MULTI_REPRESENTATION_HEIC)
     if (CheckedPtr renderImage = dynamicDowncast<RenderImage>(renderer); renderImage && renderImage->isMultiRepresentationHEIC())
-        return renderImage->style().fontCascade().primaryFont().metricsForMultiRepresentationHEIC().size();
+        imageSize = renderImage->style().fontCascade().primaryFont().metricsForMultiRepresentationHEIC().size();
+    else
 #endif
+        imageSize = image->size(renderer ? renderer->imageOrientation() : ImageOrientation(ImageOrientation::Orientation::FromImage));
 
-    if (image->drawsSVGImage() && sizeType == UsedSize)
-        return m_svgImageCache->imageSizeForRenderer(renderer);
-
-    return image->size(renderer ? renderer->imageOrientation() : ImageOrientation(ImageOrientation::Orientation::FromImage));
+    float scaleFactor = multiplier * density;
+    float widthScale = image->hasRelativeWidth() ? 1.0f : scaleFactor;
+    float heightScale = image->hasRelativeHeight() ? 1.0f : scaleFactor;
+    if (widthScale != 1.0f || heightScale != 1.0f)
+        imageSize.scale(widthScale, heightScale);
+    return imageSize;
 }
 
-
-LayoutSize CachedImage::unclampedImageSizeForRenderer(const RenderElement* renderer, float multiplier, SizeType sizeType) const
+FloatSize CachedImage::imageSizeForRenderer(const RenderElement* renderer) const
 {
-    LayoutSize imageSize = LayoutSize(imageSizeForRenderer(renderer, sizeType));
-    if (imageSize.isEmpty() || multiplier == 1.0f)
-        return imageSize;
-
-    float widthScale = m_image->hasRelativeWidth() ? 1.0f : multiplier;
-    float heightScale = m_image->hasRelativeHeight() ? 1.0f : multiplier;
-    imageSize.scale(widthScale, heightScale);
-    return imageSize;    
+    return internalImageSizeForRenderer(renderer, 1.0f, UsedSize, 1.0f);
 }
 
-LayoutSize CachedImage::imageSizeForRenderer(const RenderElement* renderer, float multiplier, SizeType sizeType) const
+LayoutSize CachedImage::unclampedImageSizeForRenderer(const RenderElement* renderer, float multiplier, SizeType sizeType, float density) const
 {
-    auto imageSize = unclampedImageSizeForRenderer(renderer, multiplier, sizeType);
+    return LayoutSize(internalImageSizeForRenderer(renderer, multiplier, sizeType, density));
+}
+
+LayoutSize CachedImage::imageSizeForRenderer(const RenderElement* renderer, float multiplier, SizeType sizeType, float density) const
+{
+    auto imageSize = unclampedImageSizeForRenderer(renderer, multiplier, sizeType, density);
     if (imageSize.isEmpty() || multiplier == 1.0f)
         return imageSize;
 
--- a/Source/WebCore/rendering/RenderImageResource.cpp
+++ b/Source/WebCore/rendering/RenderImageResource.cpp
@@ -29,10 +29,8 @@
 #include "RenderImageResource.h"
 
 #include "CachedImage.h"
-#include "Image.h"
 #include "NullGraphicsContext.h"
 #include "RenderElement.h"
-#include "RenderImage.h"
 #include "RenderStyle+GettersInlines.h"
 #include "StyleCachedImage.h"
 #include "StyleInvalidImage.h"
@@ -144,10 +142,7 @@ LayoutSize RenderImageResource::imageSize(float multiplier, CachedImage::SizeTyp
 {
     if (!m_styleImage)
         return { };
-    auto size = LayoutSize(m_styleImage->imageSize(m_renderer.get(), multiplier, type));
-    if (auto* renderImage = dynamicDowncast<RenderImage>(m_renderer.get()))
-        size.scale(renderImage->imageDevicePixelRatio());
-    return size;
+    return LayoutSize(m_styleImage->imageSize(m_renderer.get(), multiplier, type));
 }
 
 } // namespace WebCore
--- a/Source/WebCore/style/values/images/kinds/StyleCachedImage.cpp
+++ b/Source/WebCore/style/values/images/kinds/StyleCachedImage.cpp
@@ -30,6 +30,7 @@
 #include "ContainerNodeInlines.h"
 #include "ReferencedSVGResources.h"
 #include "RenderElement.h"
+#include "RenderImage.h"
 #include "RenderObjectInlines.h"
 #include "RenderSVGResourceMasker.h"
 #include "RenderView.h"
@@ -236,7 +237,10 @@ FloatSize CachedImage::imageSize(const RenderElement* renderer, float multiplier
         return m_containerSize;
     if (!m_cachedImage)
         return { };
-    return m_cachedImage->imageSizeForRenderer(renderer, multiplier, sizeType) / m_scaleFactor;
+    float density = 1.0f;
+    if (CheckedPtr renderImage = dynamicDowncast<RenderImage>(renderer))
+        density = renderImage->imageDevicePixelRatio();
+    return m_cachedImage->imageSizeForRenderer(renderer, multiplier, sizeType, density) / m_scaleFactor;
 }
 
 bool CachedImage::imageHasRelativeWidth() const
--- a/Source/WebCore/svg/graphics/SVGImage.cpp
+++ b/Source/WebCore/svg/graphics/SVGImage.cpp
@@ -100,6 +100,45 @@ RefPtr<SVGSVGElement> SVGImage::rootElement() const
     return DocumentSVG::rootElement(*localMainFrame->document());
 }
 
+FloatSize SVGImage::resolvedIntrinsicSize(float density) const
+{
+    constexpr float defaultWidth = 300;
+    constexpr float defaultHeight = 150;
+
+    RefPtr rootElement = this->rootElement();
+    if (!rootElement)
+        return { defaultWidth, defaultHeight };
+
+    std::optional<float> aspectRatio;
+    auto viewBox = rootElement->viewBox();
+    if (!viewBox.isEmpty())
+        aspectRatio = viewBox.width() / viewBox.height();
+
+    constexpr float defaultRatio = defaultWidth / defaultHeight;
+    float width = defaultWidth;
+    float height = defaultHeight;
+
+    // Four cases for { hasIntrinsicWidth, hasIntrinsicHeight }.
+    if (rootElement->hasIntrinsicWidth()) {
+        width = rootElement->intrinsicWidth() * density;
+        if (rootElement->hasIntrinsicHeight())
+            height = rootElement->intrinsicHeight() * density;  // Case 1: { true, true }
+        else if (aspectRatio)                                   // Case 2: { true, false }
+            height = width / *aspectRatio;
+    } else if (rootElement->hasIntrinsicHeight()) {
+        height = rootElement->intrinsicHeight() * density;
+        if (aspectRatio)                                        // Case 3: { false, true }
+            width = height * *aspectRatio;
+    } else if (aspectRatio) {
+        if (*aspectRatio >= defaultRatio)                       // Case 4: { false, false }
+            height = width / *aspectRatio;
+        else
+            width = height * *aspectRatio;
+    }
+
+    return { width, height };
+}
+
 bool SVGImage::renderingTaintsOrigin() const
 {
     RefPtr rootElement = this->rootElement();
@@ -373,15 +412,27 @@ LocalFrameView* SVGImage::frameView() const
     return localMainFrame->view();
 }
 
+bool SVGImage::hasIntrinsicWidth() const
+{
+    RefPtr rootElement = this->rootElement();
+    return rootElement && rootElement->hasIntrinsicWidth();
+}
+
+bool SVGImage::hasIntrinsicHeight() const
+{
+    RefPtr rootElement = this->rootElement();
+    return rootElement && rootElement->hasIntrinsicHeight();
+}
+
 bool SVGImage::hasRelativeWidth() const
 {
-    // FIXME: This seems wrong.
+    // FIXME: Delete this function and replace all the calls to it with !hasIntrinsicWidth().
     return false;
 }
 
 bool SVGImage::hasRelativeHeight() const
 {
-    // FIXME: This seems wrong.
+    // FIXME: Delete this function and replace all the calls to it with !hasIntrinsicHeight().
     return false;
 }
 
--- a/LayoutTests/imported/w3c/web-platform-tests/html/semantics/embedded-content/the-img-element/naturalWidth-naturalHeight-width-height-expected.txt
+++ b/LayoutTests/imported/w3c/web-platform-tests/html/semantics/embedded-content/the-img-element/naturalWidth-naturalHeight-width-height-expected.txt
@@ -10,46 +10,46 @@ PASS non existent image with width/height attributes, no natural dimensions
 PASS non existent image with width/height attributes, no natural dimensions (when not rendered)
 PASS SVG image, no natural dimensions
 PASS SVG image, no natural dimensions (when not rendered)
-FAIL SVG image with width/height attrs, no natural dimensions assert_equals: naturalWidth expected 300 but got 40
+PASS SVG image with width/height attrs, no natural dimensions
 PASS SVG image with width/height attrs, no natural dimensions (when not rendered)
-FAIL SVG image with width attr, no natural dimensions assert_equals: naturalWidth expected 300 but got 40
+PASS SVG image with width attr, no natural dimensions
 PASS SVG image with width attr, no natural dimensions (when not rendered)
-FAIL SVG image with height attr, no natural dimensions assert_equals: naturalHeight expected 150 but got 10
+PASS SVG image with height attr, no natural dimensions
 PASS SVG image with height attr, no natural dimensions (when not rendered)
 PASS SVG image, percengage natural dimensions
 PASS SVG image, percengage natural dimensions (when not rendered)
 PASS SVG image, negative percengage natural dimensions
 PASS SVG image, negative percengage natural dimensions (when not rendered)
 PASS SVG image, with natural width
-FAIL SVG image, with natural width (when not rendered) assert_equals: naturalWidth when not rendered expected 60 but got 300
+PASS SVG image, with natural width (when not rendered)
 PASS SVG image, with natural height
-FAIL SVG image, with natural height (when not rendered) assert_equals: naturalHeight when not rendered expected 60 but got 150
-FAIL SVG image, with natural width of 0 assert_equals: naturalWidth expected 0 but got 300
-FAIL SVG image, with natural width of 0 (when not rendered) assert_equals: naturalWidth when not rendered expected 0 but got 300
-FAIL SVG image, with natural height of 0 assert_equals: naturalHeight expected 0 but got 150
-FAIL SVG image, with natural height of 0 (when not rendered) assert_equals: naturalHeight when not rendered expected 0 but got 150
+PASS SVG image, with natural height (when not rendered)
+PASS SVG image, with natural width of 0
+PASS SVG image, with natural width of 0 (when not rendered)
+PASS SVG image, with natural height of 0
+PASS SVG image, with natural height of 0 (when not rendered)
 PASS SVG image, with natural width being negative
 PASS SVG image, with natural width being negative (when not rendered)
 PASS SVG image, with natural height being negative
 PASS SVG image, with natural height being negative (when not rendered)
-FAIL SVG image, no natural dimensions, and aspect ratio from viewBox assert_equals: naturalWidth expected 300 but got 720
-FAIL SVG image, no natural dimensions, and aspect ratio from viewBox (when not rendered) assert_equals: naturalWidth when not rendered expected 300 but got 600
-FAIL SVG image, percengage natural dimensions, and aspect ratio from viewBox assert_equals: naturalWidth expected 300 but got 720
-FAIL SVG image, percengage natural dimensions, and aspect ratio from viewBox (when not rendered) assert_equals: naturalWidth when not rendered expected 300 but got 600
-FAIL SVG image, negative percengage natural dimensions, and aspect ratio from viewBox assert_equals: naturalWidth expected 300 but got 720
-FAIL SVG image, negative percengage natural dimensions, and aspect ratio from viewBox (when not rendered) assert_equals: naturalWidth when not rendered expected 300 but got 600
+PASS SVG image, no natural dimensions, and aspect ratio from viewBox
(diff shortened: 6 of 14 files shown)


Ahmad-S792:
Oh super nice!!

annevk:
> the resulting height should not depend

Oops, it should!

Source/WebCore/rendering/RenderImageResource.cpp
```diff
@@ -145,8 +146,19 @@ LayoutSize RenderImageResource::imageSize(float multiplier, CachedImage::SizeTyp
...
-        size.scale(renderImage->imageDevicePixelRatio());
+    if (CheckedPtr renderImage = dynamicDowncast<RenderImage>(m_renderer.get())) {
+        auto dpr = renderImage->imageDevicePixelRatio();
+        if (type == CachedImage::IntrinsicSize) {
+            if (RefPtr cachedImage = renderImage->cachedImage()) {
+                auto intrinsic = cachedImage->intrinsicImageSize();
+                if (intrinsic.width)
+                    size.setWidth(size.width() * dpr);
+                if (intrinsic.height)
+                    size.setHeight(size.height() * dpr);
```
shallawa: I do not understand what is expected here if one of the dimensions is missing or the two are missing. We are keeping the dimension from the size we get from `imageSize()` but we are not scaling it with `imageDevicePixelRatio` as before. Is this intentional?

shallawa: I would suggest the scaling to happen at the end in all cases:
```
if (type == CachedImage::IntrinsicSize) {
    ...
    size.setWidth(size.width());
    ...
}
size.scale(dpr);
```

Source/WebCore/svg/graphics/SVGImage.cpp
```diff
@@ -100,6 +101,28 @@ RefPtr<SVGSVGElement> SVGImage::rootElement() const
...
+    std::optional<float> width;
+    if (rootElement->hasIntrinsicWidth())
+        width = rootElement->intrinsicWidth();
+
+    std::optional<float> height;
+    if (rootElement->hasIntrinsicHeight())
+        height = rootElement->intrinsicHeight();
+
+    std::optional<float> aspectRatio;
+    FloatRect vb = rootElement->viewBox();
```
shallawa: vb -> viewBox

I think we do not like this kind of abbreviations

Source/WebCore/svg/graphics/SVGImage.cpp
```diff
@@ -100,6 +101,28 @@ RefPtr<SVGSVGElement> SVGImage::rootElement() const
...
+    if (vb.width() > 0 && vb.height() > 0)
```
shallawa: I think you can use `vb.isEmpty()`.

Source/WebCore/svg/SVGSVGElement.cpp
```diff
@@ -666,12 +666,18 @@ FloatSize SVGSVGElement::currentViewportSizeExcludingZoom() const
 
 bool SVGSVGElement::hasIntrinsicWidth() const
 {
-    return width().lengthType() != SVGLengthType::Percentage;
+    if (width().lengthType() == SVGLengthType::Percentage)
+        return false;
+    SVGLengthContext lengthContext(this);
+    return width().value(lengthContext) >= 0;
```
shallawa: I do not understand why this condition is added here. What is the difference between 0 or -1 as Intrinsic width?

Source/WebCore/rendering/RenderImageResource.cpp
```diff
@@ -145,8 +146,19 @@ LayoutSize RenderImageResource::imageSize(float multiplier, CachedImage::SizeTyp
     if (!m_styleImage)
         return { };
     auto size = LayoutSize(m_styleImage->imageSize(m_renderer.get(), multiplier, type));
```
shallawa: I do not like getting the `m_styleImage->imageSize()` here and then below we get the `intrinsicImageSize()`
```
        if (type == CachedImage::IntrinsicSize) {
            if (RefPtr cachedImage = renderImage->cachedImage()) {
                auto intrinsic = cachedImage->intrinsicImageSize();
```

In fact `m_styleImage->imageSize()` will end up calling `SVGImage::size()` which surprisingly returns this
```
FloatSize size(ImageOrientation = ImageOrientation::Orientation::FromImage) const final { return m_intrinsicSize; }
```

shallawa: I think CachedImage::imageSizeForRenderer() should propagate the `SizeType` to `SVGImageCache::imageSizeForRenderer()` which propagates the `SizeType` to `SVGImage::size()`.

shallawa (commented):
Names became messy. `SVGImage::size()` returns its `m_intrinsicSize`. And you are adding `SVGImage::intrinsicDimensions()` also to the same class.

In `RenderImageResource::imageSize()` you are passing `CachedImage::SizeType` to `Style::Image::imageSize()`. But it seems to be ignored when its value is `IntrinsicSize`. Then you are adding new code to  deal with `CachedImage::SizeType::IntrinsicSize` by getting `CachedImage::intrinsicImageSize()` which get it from `SVGImage::intrinsicDimensions()`

And you are making `ImageIntrinsicSize` to work for all types of images even it makes sense only with `SVGImage`.

My suggestion is move all the calculations to `SVGImage::size()` with passing the necessary flags so we do not touch many different areas. And more importantly we need to have a clear story about what is returned by `size()`, `imageSize()`, `intrinsicImageSize()`   and `intrinsicDimensions()`

Source/WebCore/rendering/RenderImageResource.cpp
annevk: Yes this is intentional to preserve the 300 x 150 default and not scale that.

Source/WebCore/rendering/RenderImageResource.cpp
```diff
@@ -145,8 +145,19 @@ LayoutSize RenderImageResource::imageSize(float multiplier, CachedImage::SizeTyp
...
-    if (auto* renderImage = dynamicDowncast<RenderImage>(m_renderer.get()))
-        size.scale(renderImage->imageDevicePixelRatio());
+    if (CheckedPtr renderImage = dynamicDowncast<RenderImage>(m_renderer.get())) {
+        auto dpr = renderImage->imageDevicePixelRatio();
+        if (type == CachedImage::IntrinsicSize) {
```
shallawa: Please make an early return here:

```
auto dpr = renderImage->imageDevicePixelRatio();
if (WTF::areEssentiallyEqual(dpr, 1.0f)) [[likely]]
    return size;
if (type != CachedImage::IntrinsicSize)
    return size.scaled(dpr);
```

Source/WebCore/rendering/RenderImageResource.cpp
```diff
@@ -145,8 +145,19 @@ LayoutSize RenderImageResource::imageSize(float multiplier, CachedImage::SizeTyp
...
+            if (RefPtr cachedImage = renderImage->cachedImage()) {
+                RefPtr image = cachedImage->image();
```
shallawa: `image` can be nullptr. So should we add `imageHasIntrinsicWidth()` and `imageHasIntrinsicHeight()` to `CachedImage`? It has already these:
```
    bool imageHasRelativeWidth() const { return m_image && m_image->hasRelativeWidth(); }
    bool imageHasRelativeHeight() const { return m_image && m_image->hasRelativeHeight(); }
```

Source/WebCore/svg/graphics/SVGImage.cpp
```diff
@@ -100,6 +100,49 @@ RefPtr<SVGSVGElement> SVGImage::rootElement() const
...
+            return { *height * *aspectRatio, *height };
+        return { defaultWidth, *height };
+    }
+    if (aspectRatio) {
+        constexpr float defaultRatio = defaultWidth / defaultHeight;
+        if (*aspectRatio >= defaultRatio)
+            return { defaultWidth, defaultWidth / *aspectRatio };
+        return { defaultHeight * *aspectRatio, defaultHeight };
+    }
+    return { defaultWidth, defaultHeight };
```
shallawa:
```
    float width = defaultWidth;
    float height = defaultHeight;
    
    // Four cases for { hasIntrinsicWidth,  hasIntrinsicHeight }
    if (rootElement->hasIntrinsicWidth()) {
        width = rootElement->intrinsicWidth() * density;
        if (rootElement->hasIntrinsicHeight())
            height = rootElement->intrinsicHeight() * density;  // Case 1: { true, true }
        else if (aspectRatio)                                   // Case 2: { true, false }
            height = width / *aspectRatio;
    } else if (rootElement->hasIntrinsicHeight()) {
        height = rootElement->intrinsicHeight() * density;
        if (aspectRatio)                                        // Case 3: { fasle, true }
            width = height * *aspectRatio;
    } else if (aspectRatio) {
        if (*aspectRatio >= defaultRatio)                       // Case 4: { fasle, false }
            height = width / *aspectRatio;
        else
            width = height * *aspectRatio;
    }

    return { width, height };
```

Source/WebCore/svg/graphics/SVGImage.cpp
```diff
@@ -373,6 +416,18 @@ LocalFrameView* SVGImage::frameView() const
...
+
+bool SVGImage::hasIntrinsicHeight() const
+{
+    RefPtr rootElement = this->rootElement();
+    return rootElement && rootElement->hasIntrinsicHeight();
+}
+
 bool SVGImage::hasRelativeWidth() const
 {
     // FIXME: This seems wrong.
```
shallawa: Should we change this comment to be something like this:
```
// FIXME: Delete this function and replace all the calls to it with !hasIntrinsicWidth()
```

Source/WebCore/html/HTMLImageElement.cpp
```diff
@@ -695,24 +695,23 @@ float HTMLImageElement::effectiveImageDevicePixelRatio() const
...
 }
 
 unsigned HTMLImageElement::naturalWidth() const
 {
-    RefPtr image = m_imageLoader->image();
-    return image ? image->unclampedImageSizeForRenderer(protect(renderer()).get(), effectiveImageDevicePixelRatio()).width().toUnsigned() : 0;
+    RefPtr cachedImage = m_imageLoader->image();
+    if (!cachedImage)
+        return 0;
+    return cachedImage->unclampedImageSizeForRenderer(nullptr, 1.0f, CachedImage::IntrinsicSize, effectiveImageDevicePixelRatio()).width().toUnsigned();
```
shallawa: I do not understand this change. This function used to take `effectiveImageDevicePixelRatio()` as the argument `multiplier`:
```
LayoutSize unclampedImageSizeForRenderer(const RenderElement* renderer, float multiplier, SizeType = UsedSize) const;
```
Now we replacing this with `multiplier` = 1.0f and `density` = effectiveImageDevicePixelRatio().

This function scales the imageSize with `multiplier`. With this change it not going to do that.

```
    float widthScale = m_image->hasRelativeWidth() ? 1.0f : multiplier;
    float heightScale = m_image->hasRelativeHeight() ? 1.0f : multiplier;
    imageSize.scale(widthScale, heightScale);
    return imageSize; 
```

Is not scaling the `imageSize` with `effectiveImageDevicePixelRatio().` in `unclampedImageSizeForRenderer()` intentional?

annevk: Oh yeah, this is weird. We can just use `imageSizeForRenderer` here.

LayoutTests/imported/w3c/web-platform-tests/html/semantics/embedded-content/the-img-element/naturalWidth-naturalHeight-width-height.tentative-expected.txt
```diff
@@ -160,72 +160,72 @@ PASS non existent image, no natural dimensions (with srcset/2x)
...
-FAIL SVG image, negative percengage natural dimensions, and aspect ratio from viewBox (with srcset/2x) (when not rendered) assert_equals: naturalWidth when not rendered expected 150 but got 600
+FAIL SVG image, no natural dimensions, and aspect ratio from viewBox (with srcset/2x) assert_equals: naturalWidth expected 150 but got 300
+FAIL SVG image, no natural dimensions, and aspect ratio from viewBox (with srcset/2x) (when not rendered) assert_equals: naturalWidth when not rendered expected 150 but got 300
+FAIL SVG image, percengage natural dimensions, and aspect ratio from viewBox (with srcset/2x) assert_equals: naturalWidth expected 150 but got 300
+FAIL SVG image, percengage natural dimensions, and aspect ratio from viewBox (with srcset/2x) (when not rendered) assert_equals: naturalWidth when not rendered expected 150 but got 300
+FAIL SVG image, negative percengage natural dimensions, and aspect ratio from viewBox (with srcset/2x) assert_equals: naturalWidth expected 150 but got 300
+FAIL SVG image, negative percengage natural dimensions, and aspect ratio from viewBox (with srcset/2x) (when not rendered) assert_equals: naturalWidth when not rendered expected 150 but got 300
```
shallawa: Why are these cases still failing? Should not they be fixed by thus PR? no natural dimensions and density = 2 should return width 150. right?

Source/WebCore/loader/cache/CachedImage.cpp
```diff
@@ -319,36 +319,48 @@ FloatSize CachedImage::imageSizeForRenderer(const RenderElement* renderer, SizeT
         return renderImage->style().fontCascade().primaryFont().metricsForMultiRepresentationHEIC().size();
 #endif
```
shallawa: The returned size used to be scaled by `multiplier` in `unclampedImageSizeForRenderer()`. Now it is not scaled here or in `unclampedImageSizeForRenderer()`.

Source/WebCore/loader/cache/CachedImage.cpp
```diff
@@ -319,36 +319,48 @@ FloatSize CachedImage::imageSizeForRenderer(const RenderElement* renderer, SizeT
...
 
-    if (image->drawsSVGImage() && sizeType == UsedSize)
-        return m_svgImageCache->imageSizeForRenderer(renderer);
+    if (RefPtr svgImage = dynamicDowncast<SVGImage>(*image)) {
+        if (sizeType == UsedSize) {
+            ASSERT(density == 1.0f);
```
shallawa: Why is this ASSERT necessary? Why do not we pass `density` to this function even `if (sizeType == CachedImage::IntrinsicSize)`

Source/WebCore/loader/cache/CachedImage.cpp
```diff
@@ -319,36 +319,48 @@ FloatSize CachedImage::imageSizeForRenderer(const RenderElement* renderer, SizeT
...
+        if (zoom != 1.0f)
+            size.scale(zoom);
+        return size;
+    }
 
-    return image->size(renderer ? renderer->imageOrientation() : ImageOrientation(ImageOrientation::Orientation::FromImage));
+    auto size = image->size(renderer ? renderer->imageOrientation() : ImageOrientation(ImageOrientation::Orientation::FromImage));
+    float scale = zoom * density;
+    if (scale != 1.0f)
+        size.scale(scale);
```
shallawa: I think this scaling should check the `image->hasRelativeWidth()` and `image->hasRelativeWidth()` as we used to do in `unclampedImageSizeForRenderer()`
```
    float widthScale = m_image->hasRelativeWidth() ? 1.0f : multiplier;
    float heightScale = m_image->hasRelativeHeight() ? 1.0f : multiplier;
    imageSize.scale(widthScale, heightScale);
```

Source/WebCore/loader/cache/CachedImage.cpp
```diff
@@ -319,36 +319,48 @@ FloatSize CachedImage::imageSizeForRenderer(const RenderElement* renderer, SizeT
         return renderImage->style().fontCascade().primaryFont().metricsForMultiRepresentationHEIC().size();
 #endif
 
-    if (image->drawsSVGImage() && sizeType == UsedSize)
-        return m_svgImageCache->imageSizeForRenderer(renderer);
+    if (RefPtr svgImage = dynamicDowncast<SVGImage>(*image)) {
```
shallawa, line 317: `drawsSVGImage()` means the image is either `SVGImage` or `SVGImageForContainer`. This change will exclude `SVGImageForContainer`. Is this intentional?

Source/WebCore/rendering/RenderImageResource.cpp
```diff
@@ -145,8 +145,22 @@ LayoutSize RenderImageResource::imageSize(float multiplier, CachedImage::SizeTyp
...
+    if (type != CachedImage::IntrinsicSize) {
+        size.scale(dpr);
+        return size;
+    }
+    if (RefPtr cachedImage = renderImage->cachedImage()) {
+        if (cachedImage->imageHasIntrinsicWidth())
+            size.setWidth(size.width() * dpr);
+        if (cachedImage->imageHasIntrinsicHeight())
+            size.setHeight(size.height() * dpr);
+    }
```
shallawa: Why do not we pass `renderImage->imageDevicePixelRatio()` as the `density` to `StyleImage::imageSize()` so we do not need to do the scaling in two places: here and in `CachedImage::internalImageSizeForRenderer()`?

1. Add a new argument `density` to `Style::Image::imageSize()` with default value = 1.0f
2. Make Style::CachedImage::imageSize() pass the `density` to CachedImage::imageSizeForRenderer()
3. Make `CachedImage::internalImageSizeForRenderer()` scales the size with density `if (type == CachedImage::IntrinsicSize)`

Source/WebCore/loader/cache/CachedImage.cpp
annevk: `m_image` is never an `SVGImageForContainer` and since we needed the `downcast` for the call below this seemed like a nice simplification.

Source/WebCore/loader/cache/CachedImage.cpp
annevk: I added a comment to explain.

Source/WebCore/loader/cache/CachedImage.cpp
```diff
@@ -308,47 +308,65 @@ void CachedImage::setContainerContextForClient(const CachedImageClient& client,
     m_svgImageCache->setContainerContextForClient(client, containerSize, containerZoom, imageURL);
 }
 
-FloatSize CachedImage::imageSizeForRenderer(const RenderElement* renderer, SizeType sizeType) const
+static FloatSize applyScaling(FloatSize size, const Image& image, float zoom, float density)
```
shallawa: This could be just a lambda inside `internalImageSizeForRenderer()` since it is the only caller.

Source/WebCore/loader/cache/CachedImage.cpp
```diff
@@ -308,47 +308,65 @@ void CachedImage::setContainerContextForClient(const CachedImageClient& client,
...
+{
+    float widthScale = image.hasRelativeWidth() ? 1.0f : zoom * density;
+    float heightScale = image.hasRelativeHeight() ? 1.0f : zoom * density;
```
shallawa: Instead of repeating the multiplication, this function can take only argument for scaling: `multiplier`, `scaleFactor` or something similar. The callers will send `zoom * density` for this argument.

Source/WebCore/loader/cache/CachedImage.cpp
```diff
@@ -308,47 +308,65 @@ void CachedImage::setContainerContextForClient(const CachedImageClient& client,
...
+FloatSize CachedImage::internalImageSizeForRenderer(const RenderElement* renderer, float zoom, SizeType sizeType, float density) const
 {
     RefPtr image = m_image;
     if (!image)
         return { };
 
 #if ENABLE(MULTI_REPRESENTATION_HEIC)
     if (CheckedPtr renderImage = dynamicDowncast<RenderImage>(renderer); renderImage && renderImage->isMultiRepresentationHEIC())
-        return renderImage->style().fontCascade().primaryFont().metricsForMultiRepresentationHEIC().size();
+        return applyScaling(renderImage->style().fontCascade().primaryFont().metricsForMultiRepresentationHEIC().size(), *image, zoom, density);
```
shallawa: This line is now hard to read. I think we can rearrange the code in this function to make it more readable:

```
// Handling the SVGImage comes first.
...
FloatSize imageSize;
#if ENABLE(MULTI_REPRESENTATION_HEIC)
    if (CheckedPtr renderImage = ...)
        imageSize = renderImage->style()...
    else
#endif
        imageSize = image->size(renderer ...)

return applyScaling(...);
```

Source/WebCore/loader/cache/CachedImage.cpp
```diff
@@ -308,47 +308,65 @@ void CachedImage::setContainerContextForClient(const CachedImageClient& client,
...
+            // system via setContainerContextForClient), so density does not apply.
+            auto size = m_svgImageCache->imageSizeForRenderer(renderer);
+            if (zoom != 1.0f)
+                size.scale(zoom);
+            return size;
+        }
+        auto size = svgImage->resolvedIntrinsicSize(density);
+        if (zoom != 1.0f)
+            size.scale(zoom);
+        return size;
```
shallawa, line 327:
```
    FloatSize imageSize;
    if (sizeType == UsedSize)
        imageSize = m_svgImageCache->...
    else
        imageSize = svgImage->...

    if (zoom != 1.0f)
        size.scale(zoom);
    return size;
```

Source/WebCore/loader/cache/CachedImage.cpp
```diff
@@ -308,47 +308,65 @@ void CachedImage::setContainerContextForClient(const CachedImageClient& client,
...
+static FloatSize applyScaling(FloatSize size, const Image& image, float zoom, float density)
+{
+    float widthScale = image.hasRelativeWidth() ? 1.0f : zoom * density;
+    float heightScale = image.hasRelativeHeight() ? 1.0f : zoom * density;
+    if (widthScale != 1.0f || heightScale != 1.0f)
+        size.scale(widthScale, heightScale);
+    return size;
+}
+
+FloatSize CachedImage::internalImageSizeForRenderer(const RenderElement* renderer, float zoom, SizeType sizeType, float density) const
```
shallawa: I think the name `zoom` is a little bit confusing. Some callers pass the `zoomFactor` and others pass `imageDevicePixelRatio` which in fact the `1 / ImageCandidate.density`. So I think keeping the name `multiplier` might be a better choice.

Source/WebCore/rendering/RenderImageResource.cpp
```diff
@@ -144,10 +143,10 @@ LayoutSize RenderImageResource::imageSize(float multiplier, CachedImage::SizeTyp
...
     if (!m_styleImage)
         return { };
-    auto size = LayoutSize(m_styleImage->imageSize(m_renderer.get(), multiplier, type));
-    if (auto* renderImage = dynamicDowncast<RenderImage>(m_renderer.get()))
-        size.scale(renderImage->imageDevicePixelRatio());
-    return size;
+    float density = 1.0f;
+    if (CheckedPtr renderImage = dynamicDowncast<RenderImage>(m_renderer.get()))
+        density = renderImage->imageDevicePixelRatio();
+    return LayoutSize(m_styleImage->imageSize(m_renderer.get(), multiplier, type, density));
```
shallawa: Here we use `multiplier` not `zoom`.

I think `multiplier` is a better name given the callers use it to scale the size for different purposes.

Source/WebCore/rendering/RenderImageResource.cpp
```diff
@@ -144,10 +143,10 @@ LayoutSize RenderImageResource::imageSize(float multiplier, CachedImage::SizeTyp
 {
     if (!m_styleImage)
         return { };
-    auto size = LayoutSize(m_styleImage->imageSize(m_renderer.get(), multiplier, type));
-    if (auto* renderImage = dynamicDowncast<RenderImage>(m_renderer.get()))
-        size.scale(renderImage->imageDevicePixelRatio());
-    return size;
+    float density = 1.0f;
+    if (CheckedPtr renderImage = dynamicDowncast<RenderImage>(m_renderer.get()))
+        density = renderImage->imageDevicePixelRatio();
```
shallawa: Since `Style::CachedImage::imageSize()` is the only function that uses `density`. I think you can move getting `density` to it and not doing here.

Source/WebCore/style/values/images/kinds/StyleCachedImage.cpp
```diff
@@ -230,13 +230,13 @@ bool CachedImage::errorOccurred() const
     return m_cachedImage->errorOccurred();
 }
 
-FloatSize CachedImage::imageSize(const RenderElement* renderer, float multiplier, WebCore::CachedImage::SizeType sizeType) const
+FloatSize CachedImage::imageSize(const RenderElement* renderer, float zoom, WebCore::CachedImage::SizeType sizeType, float density) const
```
shallawa: I think you can avoid sending `density` to this function since you can get it from the `renderer`.

Source/WebCore/style/values/images/kinds/StyleImage.h
```diff
@@ -75,7 +75,7 @@ class Image : public RefCountedAndCanMakeWeakPtr<Image> {
     virtual bool hasClient(RenderElement&) const = 0;
 
     // Size / scale.
-    virtual FloatSize imageSize(const RenderElement*, float multiplier, WebCore::CachedImage::SizeType = WebCore::CachedImage::UsedSize) const = 0;
+    virtual FloatSize imageSize(const RenderElement*, float zoom, WebCore::CachedImage::SizeType = WebCore::CachedImage::UsedSize, float density = 1.0f) const = 0;
```
shallawa: This prototype can stay the same.

Source/WebCore/html/HTMLImageElement.cpp
shallawa: Why do we pass nullptr for the renderer here. This will break `isMultiRepresentationHEIC()` condition in `CachedImage::internalImageSizeForRenderer()` since the renderer is always nullptr. It will also ignore the CSS `image-orientation` and uses the `FromImage`

Source/WebCore/html/HTMLImageElement.cpp
```diff
@@ -695,24 +695,23 @@ float HTMLImageElement::effectiveImageDevicePixelRatio() const
...
 }
 
 unsigned HTMLImageElement::naturalHeight() const
 {
-    RefPtr image = m_imageLoader->image();
-    return image ? image->unclampedImageSizeForRenderer(protect(renderer()).get(), effectiveImageDevicePixelRatio()).height().toUnsigned() : 0;
+    RefPtr cachedImage = m_imageLoader->image();
+    if (!cachedImage)
+        return 0;
+    return cachedImage->unclampedImageSizeForRenderer(nullptr, 1.0f, CachedImage::IntrinsicSize, effectiveImageDevicePixelRatio()).height().toUnsigned();
```
shallawa: Same here. nullptr for renderer.

shallawa (commented):
Please update the commit message to describe the fix. I think it is currently describes the problem only.

Source/WebCore/loader/cache/CachedImage.cpp
```diff
@@ -308,39 +308,58 @@ void CachedImage::setContainerContextForClient(const CachedImageClient& client,
...
+        imageSize = renderImage->style().fontCascade().primaryFont().metricsForMultiRepresentationHEIC().size();
+    else
 #endif
+        imageSize = image->size(renderer ? renderer->imageOrientation() : ImageOrientation(ImageOrientation::Orientation::FromImage));
 
-    if (image->drawsSVGImage() && sizeType == UsedSize)
-        return m_svgImageCache->imageSizeForRenderer(renderer);
-
-    return image->size(renderer ? renderer->imageOrientation() : ImageOrientation(ImageOrientation::Orientation::FromImage));
+    return applyScaling(imageSize);
```
shallawa: Since this lambda is called only once, maybe there is no need for it anymore. But it is fine as is.

shallawa (approved):

Source/WebCore/loader/cache/CachedImage.cpp
annevk: I was wondering the same thing. I've made that change now.

karlcow:
this is great.

PR #62772, opened 2026-04-14
[CMake] bmalloc build should match Xcode on Apple platforms

geoffreygaren:
#### 726d68cf77612026c4de5bb4a3730a736cd342c8
<pre>
[CMake] bmalloc build should match Xcode on Apple platforms
<a href="https://bugs.webkit.org/show_bug.cgi?id=312320">https://bugs.webkit.org/show_bug.cgi?id=312320</a>
&lt;<a href="https://rdar.apple.com/problem/174781392">rdar://problem/174781392</a>&gt;

Reviewed by NOBODY (OOPS!).

Set DPAS_BMALLOC_HIDDEN because we don&apos;t export bmalloc as API.

Added missing warnings.

Now we have this heads-up comparison with Xcodebuild:

Build Type   Xcodebuild   Ninja      Speedup
────────────────────────────────────────────
Clean             23.2s    3.9s     6x (19s)
No-op             13.3s    0.1s   133x (13s)

* Source/bmalloc/CMakeLists.txt:
* Source/cmake/WebKitCompilerFlags.cmake:
</pre>
https://github.com/WebKit/WebKit/commit/726d68cf77612026c4de5bb4a3730a736cd342c8

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/156425 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/29760 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/22942 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/165246 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/110505 "Built successfully") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/29893 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/29764 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/121148 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85160 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/159383 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/23373 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/140476 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/101817 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/22437 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/20606 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/13018 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/148475 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/132116 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/18307 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/167728 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/17260 "Built successfully and passed tests") | | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/19920 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/129269 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/29361 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/24674 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/129380 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/29283 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/140101 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/87079 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/24207 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/16900 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/188308 "Built successfully") | | 
| | [loading 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/28993 "Build is in progress. Recent messages:OS: Tahoe (26.2), Xcode: 26.2; Running apply-patch; Checked out pull request; Running compile-webkit") | | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/48407 "Passed tests") | | 
| | [loading 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/28519 "Build is in progress. Recent messages:OS: Tahoe (26.2), Xcode: 26.2; Running apply-patch; Checked out pull request; Running compile-webkit") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/28747 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/28643 "Built successfully") | | | |

### DIFF
--- a/Source/cmake/OptionsMac.cmake
+++ b/Source/cmake/OptionsMac.cmake
@@ -128,6 +128,7 @@ WEBKIT_OPTION_END()
 
 set(SWIFT_REQUIRED ON)
 
+
 # FIXME: AV1 decoding requires dav1d which uses meson and isn't built by CMake yet.
 SET_AND_EXPOSE_TO_BUILD(ENABLE_AV1 OFF)
 
--- a/Source/cmake/WebKitCompilerFlags.cmake
+++ b/Source/cmake/WebKitCompilerFlags.cmake
@@ -281,6 +281,39 @@ if (COMPILER_IS_GCC_OR_CLANG AND NOT MSVC)
     WEBKIT_PREPEND_GLOBAL_COMPILER_FLAGS(-Wall -Wextra)
 endif ()
 
+# Match Apple Xcode xcconfig warnings.
+if (APPLE)
+    # Set -Werror directly to skip the flag-support check, which fails because
+    # -Werror turns existing warnings in the test compilation into errors.
+    set(CMAKE_C_FLAGS "-Werror ${CMAKE_C_FLAGS}")
+    set(CMAKE_CXX_FLAGS "-Werror ${CMAKE_CXX_FLAGS}")
+
+    WEBKIT_PREPEND_GLOBAL_COMPILER_FLAGS(
+        -Wc99-designator
+        -Wconditional-uninitialized
+        -Wdeprecated-enum-enum-conversion
+        -Wdeprecated-enum-float-conversion
+        -Wenum-float-conversion
+        -Wfinal-dtor-non-final-class
+        -Wformat=2
+        -Wpointer-to-int-cast
+        -Wreorder-init-list
+        -Wunused-but-set-variable
+        -Wvla
+        -Werror=unguarded-availability
+        -Wcast-qual
+        -Wexit-time-destructors
+        -Wextra-tokens
+        -Wglobal-constructors
+        -Wmissing-noreturn
+        -Wpacked
+        -Wredundant-decls
+        -Wtautological-compare
+        -Wthread-safety
+        -Wwrite-strings
+    )
+endif ()
+
 # Ninja tricks compilers into turning off color support.
 if (CMAKE_GENERATOR STREQUAL "Ninja")
     WEBKIT_PREPEND_GLOBAL_COMPILER_FLAGS(-fcolor-diagnostics
--- a/Source/bmalloc/CMakeLists.txt
+++ b/Source/bmalloc/CMakeLists.txt
@@ -692,6 +692,7 @@ endif ()
 
 add_definitions(-D_GNU_SOURCE)
 add_definitions(-DPAS_BMALLOC=1)
+add_definitions(-DPAS_BMALLOC_HIDDEN=1)
 
 if (USE_MIMALLOC)
     add_definitions(-DUSE_MIMALLOC=1)


justinmichaud:
Is there any reason why we shouldn't do this for non-Apple platforms too?

justinmichaud (approved):

geoffreygaren:
> Is there any reason why we shouldn't do this for non-Apple platforms too?

It depends on whether any platform offers libpas as public API. Probably they don't?

geoffreygaren:
Darn, I'm going to have to rework this to accommodate WTF being a framework in the CMake build.

Source/cmake/WebKitCompilerFlags.cmake
```diff
@@ -281,6 +281,39 @@ if (COMPILER_IS_GCC_OR_CLANG AND NOT MSVC)
     WEBKIT_PREPEND_GLOBAL_COMPILER_FLAGS(-Wall -Wextra)
 endif ()
 
+# Match Apple Xcode xcconfig warnings.
+if (APPLE)
+    # Set -Werror directly to skip the flag-support check, which fails because
+    # -Werror turns existing warnings in the test compilation into errors.
+    set(CMAKE_C_FLAGS "-Werror ${CMAKE_C_FLAGS}")
```
stwrt, line 288: In CMake this propagates to every build target including third party ones.
This leads to build errors like the below in third party libwebrtc code.

```
libwebm/webm_parser/src/byte_parser.h:65:5: error: 'WEBM_FUZZER_BYTE_ELEMENT_SIZE_LIMIT'
      is not defined, evaluates to 0 [-Werror,-Wundef]                                                                     
                                                                                                                           
opus/src/silk/SigProc_FIX.h:413:5: error: 'EMBEDDED_ARM'                                                                 
      is not defined, evaluates to 0 [-Werror,-Wundef]          
```

Xcode appears to not have these checks for third party code based on some investigation with Claude.

```
In the Xcode build, this isn't a problem because GCC_TREAT_WARNINGS_AS_ERRORS = YES is set in Base.xcconfig, and         
  third-party targets like libwebm use their own minimal xcconfig chain (libwebm.xcconfig → BaseTarget-libwebm.xcconfig)   
  that deliberately does not inherit Base.xcconfig.
```

Now, we can move this error into WEBKIT_FRAMEWORK, WEBKIT_LIBRARY, and WEBKIT_EXECUTABLE, which are WebKit owned targets to give the same effect as how Xcode does it.


```
diff --git a/Source/cmake/WebKitMacros.cmake b/Source/cmake/WebKitMacros.cmake
index 42e2f3cd9f03..88219889d0ec 100644
--- a/Source/cmake/WebKitMacros.cmake
+++ b/Source/cmake/WebKitMacros.cmake
@@ -361,6 +361,12 @@ macro(WEBKIT_FRAMEWORK _target)
     _WEBKIT_TARGET(${_target})
     _WEBKIT_TARGET_ANALYZE(${_target})
 
+    # Match Xcode's per-target GCC_TREAT_WARNINGS_AS_ERRORS = YES, scoped to
+    # WebKit's own targets so third-party code is not affected.
+    if (APPLE)
+        target_compile_options(${_target} PRIVATE -Werror)
+    endif ()
+
     # Apply PGO compile flags only to library targets (not executables) to avoid duplicate symbol errors
     # Link flags are applied globally via CMAKE_SHARED_LINKER_FLAGS for LTO compatibility
     if (PGO_COMPILE_OPTIONS)
@@ -393,6 +399,10 @@ macro(WEBKIT_LIBRARY _target)
     _WEBKIT_TARGET(${_target})
     _WEBKIT_TARGET_ANALYZE(${_target})
 
+    if (APPLE)
+        target_compile_options(${_target} PRIVATE -Werror)
+    endif ()
+
     # Apply PGO compile flags only to library targets (not executables) to avoid duplicate symbol errors
     # Link flags are applied globally via CMAKE_SHARED_LINKER_FLAGS for LTO compatibility
     if (PGO_COMPILE_OPTIONS)
@@ -411,6 +421,10 @@ macro(WEBKIT_EXECUTABLE _target)
     _WEBKIT_TARGET(${_target})
     _WEBKIT_TARGET_ANALYZE(${_target})
 
+    if (APPLE)
+        target_compile_options(${_target} PRIVATE -Werror)
+    endif ()
+
     if (${_target}_OUTPUT_NAME)
         set_target_properties(${_target} PROPERTIES OUTPUT_NAME ${${_target}_OUTPUT_NAME})
     endif ()
```

stwrt (changes requested):

Source/cmake/OptionsMac.cmake
```diff
@@ -128,6 +128,7 @@ WEBKIT_OPTION_END()
 
 set(SWIFT_REQUIRED ON)
 
+
```
mcatanzaro, line 131:
```suggestion
```

Source/cmake/WebKitCompilerFlags.cmake
mcatanzaro: It's not necessary at all. CMake build uses -Werror if `DEVELOPER_MODE` is enabled and `DEVELOPER_MODE_FATAL_WARNINGS` is not enabled. Otherwise, it doesn't. Better not try to second-guess this.

Source/cmake/WebKitCompilerFlags.cmake
```diff
@@ -281,6 +281,39 @@ if (COMPILER_IS_GCC_OR_CLANG AND NOT MSVC)
...
+        -Wexit-time-destructors
+        -Wextra-tokens
+        -Wglobal-constructors
+        -Wmissing-noreturn
+        -Wpacked
+        -Wredundant-decls
+        -Wtautological-compare
+        -Wthread-safety
+        -Wwrite-strings
+    )
```
mcatanzaro, line 314: I think we should be able to enable most of these on every platform, not just Apple.

We might need to leave a couple of them Apple-specific initially, though, like -Wexit-time-destructors. But Linux developers could follow up in the future to try to avoid those.

mcatanzaro (changes requested):

Source/cmake/WebKitCompilerFlags.cmake
mcatanzaro: Oh, and build-webkit itself enables `DEVELOPER_MODE`. The idea is: WebKit developers fix warnings, and end users who just want to build WebKit don't have to wade through spurious build failures when future compiler versions add new warnings.

geoffreygaren: @mcatanzaro Is the expectation that non-developers build `Release`, and developers build `Debug` or `RelWithDebInfo`? -- If so, do `Debug` and `RelWithDebInfo` include `-Werror` by default?

I'm trying to make sure that all developers building the Mac port with CMake get the right warnings (including `-Werror`) by default, without having to set extra environment variables or pass extra arguments. Even if they don't use the `build-webkit` script.

mcatanzaro: > Is the expectation that non-developers build Release, and developers build Debug or RelWithDebInfo?

No. Non-developers will generally use `RelWithDebInfo`, or possibly `Release` with `-g` manually added to CFLAGS. Developers probably do so too.

It's been a very long time since I last attempted a Debug build. They are very slow, and they crash a lot because we have most assertions disabled in release builds. I'm not sure how many developers actually use Debug builds, but I guess not many. (My $0.02: we should enable all non-slow assertions in release builds, and fix whatever ones actually get hit.)

> If so, do Debug and RelWithDebInfo include -Werror by default?

No. -Werror is triggered by `DEVELOPER_MODE`, not by the build type. If you use CMake directly rather than using `build-webkit`, then you would need to set `-DDEVELOPER_MODE=ON`.

I think the rule to not enable -Werror by default is very well-accepted by most open source projects, and I don't recommend ignoring it. But if you want to do things differently for the Apple ports, I think you have two good options. (1) Enable `DEVELOPER_MODE` by default (using `WEBKIT_OPTION_DEFAULT_PORT_VALUE` in OptionsMac.cmake). Or (2) rename `DEVELOPER_MODE_FATAL_WARNINGS` to just `FATAL_WARNINGS`, change it to not depend on `DEVELOPER_MODE`, set its default value based on `DEVELOPER_MODE` to avoid changing behavior for most ports, and override the default value to `ON` using `WEBKIT_OPTION_DEFAULT_PORT_VALUE` in OptionsMac.cmake. Basically, look up about 100 lines in WebKitCompilerFlags.cmake and try to reuse that instead of separately adding -Werror.

geoffreygaren:
This PR is supplanted by https://commits.webkit.org/312155@main

PR #62849, opened 2026-04-15
Fix backtraces on linux DEVELOPER_MODE builds caused by CRASH_DUMP_INFO

justinmichaud:
#### 1f83254ec13ed6387025d9f797d2536729c187aa
<pre>
Fix backtraces on linux DEVELOPER_MODE builds caused by CRASH_DUMP_INFO
<a href="https://bugs.webkit.org/show_bug.cgi?id=312413">https://bugs.webkit.org/show_bug.cgi?id=312413</a>

Reviewed by NOBODY (OOPS!).

See <a href="https://bugs.webkit.org/show_bug.cgi?id=311796">https://bugs.webkit.org/show_bug.cgi?id=311796</a> for context. CRASH_DUMP_INFO
is important for production builds, but we disable it for DEVELOPER_MODE
builds so that we can use gdb/lldb properly.

The problem is that a NO_RETURN function has no need to save it&apos;s fp. We
revert this back to the old behaviour, and also re-add some compiler flags
that were rolled out a while back to further aid in debugging
(see <a href="https://bugs.webkit.org/show_bug.cgi?id=293765).">https://bugs.webkit.org/show_bug.cgi?id=293765).</a>
</pre>
https://github.com/WebKit/WebKit/commit/a5110b3f1d9619923eba1e29f9ee9739067410eb

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/157104 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/30440 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/23631 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/165927 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/111186 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/30576 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/30443 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/121675 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85435 "1 flakes 2 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/160062 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/23923 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/141079 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/102343 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/22978 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/21210 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/13699 "Built successfully") | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/149154 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/132658 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/18907 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/168412 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/17939 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/12571 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/20527 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/129804 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/30042 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/25288 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/129912 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/29965 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/140701 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/87786 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/24739 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/17505 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/189067 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/29676 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/93690 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/48527 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/29198 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/29428 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/29325 "Built successfully") | | |


Source/cmake/OptionsWPE.cmake
```diff
@@ -123,7 +123,7 @@ WEBKIT_OPTION_DEFINE(USE_SYSTEM_SYSPROF_CAPTURE "Whether to use a system-provide
 WEBKIT_OPTION_DEFINE(USE_SYSTEM_UNIFDEF "Whether to use a system-provided unifdef" PRIVATE ON)
 WEBKIT_OPTION_DEFINE(ENABLE_COG "Toggle Cog browser compilation." PRIVATE OFF)
 WEBKIT_OPTION_DEFINE(ENABLE_JSC_RESTRICTED_OPTIONS_BY_DEFAULT "Whether to enable dangerous development options in JSC by default." PRIVATE OFF)
-WEBKIT_OPTION_DEFINE(ENABLE_CRASH_DUMP_INFO "Enables extra crash investigation tools that might increase binary size." PRIVATE ${ENABLE_DEVELOPER_MODE})
+WEBKIT_OPTION_DEFINE(ENABLE_CRASH_DUMP_INFO "Enables extra crash investigation tools that might increase binary size." PRIVATE OFF)
```
mcatanzaro: Is this intentional? We enable crash dump info for GTK and JSCOnly to aid debuggability, but disable it for WPE to save... 4 MB ([310454@main](https://commits.webkit.org/310454@main))?

If the space is truly that precious, then OK.

Source/cmake/OptionsGTK.cmake
```diff
@@ -60,6 +60,12 @@ WEBKIT_OPTION_DEFINE(USE_LIBSECRET "Whether to enable the persistent credential
 WEBKIT_OPTION_DEFINE(USE_SKIA_OPENTYPE_SVG "Whether to use the Skia built-in support for OpenType SVG fonts." PUBLIC ON)
 WEBKIT_OPTION_DEFINE(USE_SYSTEM_SYSPROF_CAPTURE "Whether to use a system-provided libsysprof-capture" PUBLIC ON)
 WEBKIT_OPTION_DEFINE(ENABLE_JSC_RESTRICTED_OPTIONS_BY_DEFAULT "Whether to enable dangerous development options in JSC by default." PRIVATE OFF)
+if (DEVELOPER_MODE)
+    set(ENABLE_CRASH_DUMP_INFO_DEFAULT OFF)
+else ()
+    set(ENABLE_CRASH_DUMP_INFO_DEFAULT ON)
+endif ()
+WEBKIT_OPTION_DEFINE(ENABLE_CRASH_DUMP_INFO "Enables extra crash investigation tools that might increase binary size." PRIVATE ${ENABLE_CRASH_DUMP_INFO_DEFAULT})
```
mcatanzaro: Can it go into OptionsCommon.cmake, so it doesn't have to be handled differently by each port? Since it only depends on `DEVELOPER_MODE`, and since `DEVELOPER_MODE` exists for every CMake port, I would expect it doesn't need to be platform-specific.

Source/WTF/wtf/Assertions.h
```diff
@@ -348,19 +347,10 @@ WTF_EXPORT_PRIVATE bool WTFIsDebuggerAttached(void);
...
 } while (0)
-#elif !ENABLE(DEVELOPER_MODE) && !OS(DARWIN)
-#ifdef __cplusplus
-#define CRASH() std::abort()
-#define CRASH_UNDER_CONSTEXPR_CONTEXT() WTFBreakpointTrapUnderConstexprContext()
-#else
-#define CRASH() abort()
-#define CRASH_UNDER_CONSTEXPR_CONTEXT() WTFBreakpointTrapUnderConstexprContext()
-#endif // __cplusplus
-#else
```
mcatanzaro, line 359: I wondered why this is so complicated, so I did a blame. Naturally, it's my fault, from [206647@main](https://commits.webkit.org/206647@main). That commit message referenced a mailing list thread, so I did some archaeology and found [this discussion](https://lists.webkit.org/archives/list/webkit-dev@lists.webkit.org/thread/5IPJDDG64WIAZEM5XRAXVL5JIJJFKHN7/).

I can't think of any reason this complexity needs to be _here_, though. Perhaps `WTFCrash` should just prefer to crash with `std::abort()`?

It's really seriously confusing that half the logic is in the header file, half is in the source file, it all references itself, and there are preprocessor guards all over the place. Every platform is different, every build configuration is different. The simpler we can make it, the better.

Source/WTF/wtf/Assertions.h
```diff
@@ -348,19 +347,10 @@ WTF_EXPORT_PRIVATE bool WTFIsDebuggerAttached(void);
...
+#else // defined(NDEBUG) && ENABLE(CRASH_DUMP_INFO)
 #define CRASH() WTFCrash()
```
mcatanzaro, line 351: One more thing. We now will hit this case in practice. But how is WTFCrash defined?

```
#if !defined(NDEBUG) || !(OS(DARWIN) || PLATFORM(PLAYSTATION) || OS(LINUX)) || !ENABLE(CRASH_DUMP_INFO)
// ... (more confusing code here) ...
#else
// We need to keep WTFCrash() around (even on non-debug OS(DARWIN) builds) as a workaround
// for presently shipping (circa early 2016) SafariForWebKitDevelopment binaries which still
// expects to link to it.
void WTFCrash()
{
    CRASH();
}
#endif // !defined(NDEBUG) || !(OS(DARWIN) || PLATFORM(PLAYSTATION))
```

So that's potentially circular if we mess up the preprocessor conditions, which are difficult to reason about. It's confusing and not great. Have you thought about simplifying this? I bet this complexity is not actually needed.

Source/WTF/wtf/Assertions.cpp
```diff
@@ -734,6 +734,14 @@ void WTFCrashWithInfoImpl(int, const char*, const char*, UCPURegister, UCPURegis
 void WTFCrashWithInfoImpl(int, const char*, const char*, UCPURegister, UCPURegister) { CRASH(); }
 void WTFCrashWithInfoImpl(int, const char*, const char*, UCPURegister) { CRASH(); }
 
+void WTFCrashWithInfo(int, const char*, const char*)
+#if COMPILER(CLANG)
+    __attribute__((optnone))
```
fujii: This is no longer needed. `optnone` was added by <https://commits.webkit.org/203263@main> to make it an inline function.

Source/WTF/wtf/Assertions.h
```diff
@@ -335,8 +335,7 @@ WTF_EXPORT_PRIVATE bool WTFIsDebuggerAttached(void);
 #define WTFBreakpointTrapUnderConstexprContext() __builtin_trap()
 
 #ifndef CRASH
-
-#if defined(NDEBUG)
+#if defined(NDEBUG) && ENABLE(CRASH_DUMP_INFO)
```
fujii, line 338: We shouldn't change PlayStation port. They don't use `WTFCrash()` function in release build.

Bug 197932 – [PlayStation] WTFCrash should preserve register state.
https://bugs.webkit.org/show_bug.cgi?id=197932

Source/WTF/wtf/Assertions.h
fujii: See https://www.mail-archive.com/webkit-dev@lists.webkit.org/msg28213.html if https://lists.webkit.org/ has a trouble for you.

Source/cmake/OptionsWPE.cmake
justinmichaud: Yeah, they really care. I'm a bit surprised TBH

Source/WTF/wtf/Assertions.h
justinmichaud: Good call, I removed all the checks and always define it now

Source/WTF/wtf/Assertions.h
justinmichaud: Should we ENABLE(CRASH_DUMP_INFO) for them by default then?

Source/cmake/OptionsGTK.cmake
```diff
@@ -60,6 +60,7 @@ WEBKIT_OPTION_DEFINE(USE_LIBSECRET "Whether to enable the persistent credential
 WEBKIT_OPTION_DEFINE(USE_SKIA_OPENTYPE_SVG "Whether to use the Skia built-in support for OpenType SVG fonts." PUBLIC ON)
 WEBKIT_OPTION_DEFINE(USE_SYSTEM_SYSPROF_CAPTURE "Whether to use a system-provided libsysprof-capture" PUBLIC ON)
 WEBKIT_OPTION_DEFINE(ENABLE_JSC_RESTRICTED_OPTIONS_BY_DEFAULT "Whether to enable dangerous development options in JSC by default." PRIVATE OFF)
+WEBKIT_OPTION_DEFINE(ENABLE_CRASH_DUMP_INFO "Enables extra crash investigation tools that might increase binary size." PRIVATE ${ENABLE_CRASH_DUMP_INFO_DEFAULT})
```
mcatanzaro: Please define it in OptionsCommon.cmake, right after defining `ENABLE_CRASH_DUMP_INFO_DEFAULT`. Then you don't need to add it to OptionsGTK.cmake, OptionsJSCOnly.cmake, or OptionsPlayStation.cmake. You can use `WEBKIT_OPTION_DEFAULT_PORT_VALUE` in OptionsWPE.cmake to override the default value to OFF.

mcatanzaro (approved):
I'm slightly disappointed that it will no longer crash with SIGABRT, but that's OK.

If you want to take this even further, we might be able to completely delete `WTFCrash()` by merging its implementation into the `CRASH()` macro. (Unless Safari really does still link to it, but I presume likely not.)

Source/cmake/OptionsGTK.cmake
justinmichaud: I saw no other definition there, that's why I was hesitant, but sounds good to me

mcatanzaro: OptionsCommon.cmake does define several other options, but more towards the top of the file rather than the bottom.

mcatanzaro: (So although it doesn't matter, it would be better feng shui to put this code not at the very bottom of the file.)

Source/cmake/OptionsJSCOnly.cmake
```diff
@@ -24,6 +24,7 @@ if (NOT WIN32)
 endif ()
 WEBKIT_OPTION_DEFINE(ENABLE_JSC_GLIB_API "Whether to enable the JavaScriptCore GLib API." PUBLIC OFF)
 WEBKIT_OPTION_DEFINE(USE_SYSTEM_UNIFDEF "Whether to use a system-provided unifdef" PRIVATE OFF)
+WEBKIT_OPTION_DEFINE(ENABLE_CRASH_DUMP_INFO "Enables extra crash investigation tools that might increase binary size." PRIVATE ${ENABLE_CRASH_DUMP_INFO_DEFAULT})
```
mcatanzaro:
```suggestion
```

Source/WTF/wtf/Assertions.h
fujii: I think `ENABLE_CRASH_DUMP_INFO` should be 1 by default to preserve the original.

fujii: On the other hand, this condition makes me feel weird.
I think `ENABLE_CRASH_DUMP_INFO` shouldn't affect `CRASH()`.
For example, GTK and WPE disable `ENABLE_CRASH_DUMP_INFO` for some condition, why does it change `CRASH()` behavior?

justinmichaud: It should hopefully change CRASH so that CRASH can preserve a unique crash address for each distinct CRASH macro. For development builds or to reduce binary size, we can reduce the amount of crash info we get, but I agree that this doesn't feel like "crash info"

justinmichaud:
I believe I have responded to everyone and there has been enough time for comments. I'm not opposed to making sigabrt work on linux, although for now I like that the Apple/linux implementations match. I'll leave this here for now to make sure that folks get unbroken, and to see if there is any damage to the Apple internal builds.

@gavin-apple (I sure hope you are the right Gavin) and @MenloDorian, I'd love a review from the Apple side before landing. This shouldn't change Apple crashes, but I want to give you a heads up just in case I missed something.

fujii:
IIUC, with the combination of x86_64 Linux, gdb and production builds, gdb can't unwind the callstack of `CRASH()`.
We have the same problem as Michael mentioned in <https://www.mail-archive.com/webkit-dev@lists.webkit.org/msg28213.html>
Right? Any other problems?

fujii:
We shouldn't change PlayStation port. They don't use WTFCrash() function in release build.

Bug 197932 – [PlayStation] WTFCrash should preserve register state.
https://bugs.webkit.org/show_bug.cgi?id=197932

mcatanzaro (changes requested):
Fujii is right. This breaks stack traces when `DEVELOPER_MODE` is disabled. Effectively, that means no more crash reports from users. I tested this by adding some sabotage:

```
diff --git a/Source/WebKit/WebProcess/gtk/WebProcessMainGtk.cpp b/Source/WebKit/WebProcess/gtk/WebProcessMainGtk.cpp
index abac72c8ce61..60599086aa3c 100644
--- a/Source/WebKit/WebProcess/gtk/WebProcessMainGtk.cpp
+++ b/Source/WebKit/WebProcess/gtk/WebProcessMainGtk.cpp
@@ -81,7 +81,7 @@ public:
 
         bindtextdomain(GETTEXT_PACKAGE, LOCALEDIR);
         bind_textdomain_codeset(GETTEXT_PACKAGE, "UTF-8");
-
+CRASH();
         return true;
     }
 ```

Here is the stack trace with `DEVELOPER_MODE` enabled:

```
Core was generated by `/home/mcatanzaro/Projects/GNOME/install/libexec/webkitgtk-6.0/WebKitWebProcess 4 56'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  WTFCrash () at /home/mcatanzaro/Projects/WebKit/Source/WTF/wtf/Assertions.cpp:378
378	    *(int *)(uintptr_t)0xbbadbeef = 0;
(gdb) bt
#0  WTFCrash () at /home/mcatanzaro/Projects/WebKit/Source/WTF/wtf/Assertions.cpp:378
#1  0x00007f66cce22578 in WebKit::WebProcessMainGtk::platformInitialize (this=<optimized out>)
    at /home/mcatanzaro/Projects/WebKit/Source/WebKit/WebProcess/gtk/WebProcessMainGtk.cpp:84
#2  0x00007f66cce1f8fc in WebKit::AuxiliaryProcessMainBase<WebKit::WebProcess, true>::run (this=0x7ffcf41b87c0, argc=<optimized out>, 
    argv=<optimized out>) at /home/mcatanzaro/Projects/WebKit/Source/WebKit/Shared/AuxiliaryProcessMain.h:68
#3  WebKit::AuxiliaryProcessMain<WebKit::WebProcessMainGtk> (argc=3, argv=0x7ffcf41b8988)
    at /home/mcatanzaro/Projects/WebKit/Source/WebKit/Shared/AuxiliaryProcessMain.h:103
#4  0x00007f66cce1f899 in WebKit::WebProcessMain (argc=-1012958944, argv=0x382d46)
    at /home/mcatanzaro/Projects/WebKit/Source/WebKit/WebProcess/gtk/WebProcessMainGtk.cpp:98
#5  0x0000000000201899 in main (argc=-1012958944, argv=0x382d46)
    at /home/mcatanzaro/Projects/WebKit/Source/WebKit/WebProcess/EntryPoint/unix/WebProcessMain.cpp:31
```

And with `DEVELOPER_MODE` disabled:

```
Core was generated by `/home/mcatanzaro/Projects/GNOME/install/libexec/webkitgtk-6.0/WebKitWebProcess 4 69'.
Program terminated with signal SIGTRAP, Trace/breakpoint trap.
#0  0x00007f72a815abbf in ?? () from /home/mcatanzaro/Projects/GNOME/install/lib/libwebkitgtk-6.0.so.4
(gdb) bt
#0  0x00007f72a815abbf in ?? () from /home/mcatanzaro/Projects/GNOME/install/lib/libwebkitgtk-6.0.so.4
#1  0x0000000000000000 in ?? ()
```

I think we need to not use `__builtin_trap()`. Whatever that does seems to break gdb.

**Comment by justinmichaud** [comment]
> I think we need to not use __builtin_trap(). Whatever that does seems to break gdb.

I don't think we expect users to attach gdb, right? The important part here is that automated crash reporting works, and that we are left with interesting things in the registers after the crash.

On release builds in general, we can't expect the backtrace to be too interesting because the compiler is free to do a bunch of optimizations (including combining multiple asserts into one crash). Based on your example above, if you added a second crash and some condition that isn't taken, I'm not sure it will tell you which one was the true cause of the crash. 

ex:
```
if (getenv("TEST1"))
    CRASH();
else
    CRASH();
```

 I think having distinct crash locations is more important that having gdb backtraces work for user reporting, but I don't know how linux projects collect this info for crash reports.
 
 In general though, I don't know why the compiler is so insistent on not saving frame pointers. I think we could test the performance in both cases to be sure, but my guess is that it is faster this way.
 
 With all of that said, I don't have the time to do all of this testing, and I can include this patch in downstream only. Without proof that at least one other project has automated crash reporting that could benefit from this, it is best to revert the original patch.

mcatanzaro:
> I don't think we expect users to attach gdb, right?

Attaching while the process is running: no, not unless as an absolute last resort, because attaching is complicated, especially if the goal is to attach to a child process. Trying to teach users how to do this in an issue report is not fun.

But users are certainly going to want to use gdb to get a stack trace from a core dump. Otherwise, how do you expect to ever receive any crash reports?

> The important part here is that automated crash reporting works, and that we are left with interesting things in the registers after the crash.

How can automated crash reporting possibly work if gdb does not work?

I only know of three automated crash report systems for Linux: KDE's Dr. Konqi, which only processes KDE crashes and so is irrelevant for WebKitGTK, Fedora's ABRT, which is _nearly_ irrelevant because it is completely abandoned and likely to be retired soon, and Ubuntu's apport, which is sort of worth considering, except Ubuntu never forwards WebKitGTK apport reports to us, so even then not really. All three of these systems rely on gdb, though, so they'll be totally useless if you make this change, right? The whole point is to collect stack traces, but your change will prevent that?

[Here are my instructions for crash reporting](https://handbook.gnome.org/issues/stack-traces.html), to give you an idea of what I expect from bug reports. (The instructions are written for GNOME, but I use them for WebKit and other Linux projects too, since there's nothing GNOME-specific there.)

> On release builds in general, we can't expect the backtrace to be too interesting because the compiler is free to do a bunch of optimizations (including combining multiple asserts into one crash). Based on your example above, if you added a second crash and some condition that isn't taken, I'm not sure it will tell you which one was the true cause of the crash.

I'm pretty confused. We almost _exclusively_ look at stack traces from release builds? When was the last time you saw a bug report with a stack trace from a non-release build? Guessing what's wrong when optimizations interfere with a stack trace is certainly not fun and has tricked me many times, but there's not really anything we can do about this. It's not like users are going to start running debug builds of WebKit?

> In general though, I don't know why the compiler is so insistent on not saving frame pointers. I think we could test the performance in both cases to be sure, but my guess is that it is faster this way.

You can build with -fno-omit-frame-pointer and -fno-omit-leaf-frame-pointer if you want to keep frame pointers. Distros are split on whether to do this. Fedora and Ubuntu both recently started doing so, because it allows full-desktop performance analysis. But it also makes your program slower.

P.S. I think most of this patch is good cleanup that's worth landing. It just shouldn't use `__builtin_trap()`.

PR #62893, opened 2026-04-16
Web Inspector: Elements: Editing an attribute value that contains double quotes results in a malformed value

rcaliman-apple:
#### c7b3ebaa936a6c3cb28974d1f8b2f64c2c99d5ad
<pre>
Web Inspector: Elements: Editing an attribute value that contains double quotes results in a malformed value
<a href="https://bugs.webkit.org/show_bug.cgi?id=312465">https://bugs.webkit.org/show_bug.cgi?id=312465</a>
<a href="https://rdar.apple.com/149523483">rdar://149523483</a>

Reviewed by Devin Rousso.

Attributes in the Elements DOM tree outline are rendered by wrapping the value in double quotes
(`WI.DOMTreeElement.prototype._buildAttributeDOM()`).

Entering editing mode turns the `&lt;span`&gt; element containing the represented attribute
into a `contenteditable=plaintext` element. (`WI.startEditing()`).

When committing the changed value, the `textContent` of this editable element
is sent to the backend as a text string (`target.DOMAgent.setAttributesAsText()`)
which interpolates the string into a dummy element to parse the result and update the node.
(`InspectorDOMAgent::setAttributesAsText()`). This is done to handle cases where
the committed value contains additional attributes, not just a single changed attribute.

If the changed attribute value contains double quotes, these conflict with the double quotes used to wrap the value in the DOM tree outline.
Web Inspector does not use the same type of quotes as in the original HTML markup.

For example, markup like `style=&apos;filter: url(&quot;#glow&quot;)&apos;`, is rendered and offered for editing
as `style=&quot;filter: url(&quot;#glow&quot;)&quot;` and committed as `style=\&quot;filter: url(\&quot;#glow\&quot;)\&quot;`.

The interpolation on the backend results in a malformed string
that causes the HTML parser to break the value at double quote limits.

This patch changes the type of quotes used in Web Inspector DOM tree outline to represent
the attribute if the attribute value already contains double quotes.

* Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js:
(WI.DOMTreeElement.prototype._buildAttributeDOM):

Canonical link: <a href="https://commits.webkit.org/311717@main">https://commits.webkit.org/311717@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/09f3009052211841ca06ebddee1453e25824addf

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/157046 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/30382 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/23573 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/165869 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/111128 "Built successfully") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/30518 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/30385 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/121622 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85398 "1 failures") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/160004 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/23865 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/141026 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/102290 "Passed tests") | | [❌ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/2b6e9c71-9eb5-4f51-be8a-c789032755b7) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/22919 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/21150 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/13641 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/132600 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/18854 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/168354 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/12513 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/20474 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/129749 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/29984 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/25228 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/129857 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35332 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/29907 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/140648 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/87724 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/24681 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/17452 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/29618 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/93632 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/29140 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/29370 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/29267 "Built successfully") | | | |

### DIFF
--- a/Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
+++ b/Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
@@ -1453,8 +1453,10 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
         let attrNameElement = attrSpanElement.createChild("span", "html-attribute-name");
         attrNameElement.textContent = name;
         let attrValueElement = null;
+
+        let quote = value.includes('"') ? "'" : "\"";
         if (hasText)
-            attrSpanElement.append("=\u200B\"");
+            attrSpanElement.append("=\u200B", quote);
 
         if (name === "src" || /\bhref\b/.test(name)) {
             let baseURL = node.frame ? node.frame.url : null;
@@ -1511,7 +1513,7 @@ WI.DOMTreeElement = class DOMTreeElement extends WI.TreeElement
         }
 
         if (hasText)
-            attrSpanElement.append("\"");
+            attrSpanElement.append(quote);
 
         this._createModifiedAnimation(name, value, hasText ? attrValueElement : attrNameElement);
 


rcaliman-apple:
Test case: [inline-filter.html](https://github.com/user-attachments/files/26784747/inline-filter.html)

## Screen recording

|Before|After|
|---|---|
|https://github.com/user-attachments/assets/5170f848-e451-4aa6-a697-f746a7cac0bd|https://github.com/user-attachments/assets/91c9d882-9393-4196-b84b-511ee2b6e3ff|

Source/WebInspectorUI/UserInterface/Views/DOMTreeElement.js
dcrousso, line 1457: doesn't the same logic also apply when using single quotes (i.e. what if the value also has a single quote in it)?

i think in other similar situations we've just replaced all `"` in the value with `\"`

the-chenergy:
```suggestion
            attrSpanElement.append("=\u200B" + quote);
```

dcrousso: if we're gonna use `append()` then that takes multiple arguments so we don't even need the `+`

rcaliman-apple: Backslash `\` isn't an escape character for the HTML parser. It's treated literally (i.e. goes into _Anything else_):

https://html.spec.whatwg.org/multipage/parsing.html#attribute-value-(double-quoted)-state

Some examples:

|HTML|Element.getAttribute('style')|
|---|---|
|`<span style='filter: url(\'#glow-1\')'>`|`"filter: url(\\"`|
|`<span style='filter: url(\\'#glow-1\\')'`>|`"filter: url(\\\\"`|
|`<span style='filter: url(\\\'#glow-1\\\')'`>|`"filter: url(\\\\\\"`|
|`<span style="filter: url(&quot;#glow-1&quot;)">`|`"filter: url(\"#glow-1\")"`|
|`<span style="filter: url(&apos;#glow-1&apos;)">`|`"filter: url('#glow-1')"`|

(`Element.getAttribute('style')` is what `Element.textContent` on the DOM tree element returns when entering edit more, showing what Webkit sees)

We can replace the inner double quotes with `&quot;` for the correct behavior before sending to the backedn for interpolating. But the regex to handle that in all cases turns narly. 

It has to handle both a simple attribute text like:
`style=\"filter: url(\"#glow-1\")\" `

and committed text that adds more attributes:

`style=\"filter: url(\"#glow-1\")\" test=\"a\" test`

(Quotes escaped by WebKit when getting `Element.textContent`. This is the text that `DOMTreeElement._attributeEditingCommitted()` receives, regardless of using `\"` or `&quot;` when creating the representation of the attribute in the DOM tree outline)

It gets error prone to figure out _exactly_ which double quotes are within an attribute value and which are wrapping the value itself. 

Ultimately, this fixes an edge case. The fix to use single quotes to wrap attributes when their values contain double quotes is the most straightforward one for the way Web Inspector handles editing attribute values by interpolating the text into a dummy element on the backend to parse it. 

I stand by this proposal.

rcaliman-apple: Paradoxically, using escaped single quotes (`&apos;`), there is no issue with interpolation within a double-quoted attribute value (i.e.: without this patch applied):

https://github.com/user-attachments/assets/867c684b-4e93-4015-bb43-57753986f894

But if using escaped double quotes (`&quot;`), it fails in a different way  🤷‍♂️:

https://github.com/user-attachments/assets/6d949ae2-70cd-4bc6-bbb0-1c275e567f85

This patch supports both cases.

dcrousso (approved):

PR #62925, opened 2026-04-16
[JSC][JSPI] Scan PinballCompletion conservative roots via a GC marking constraint

ast-hugger:
#### 4c00d7bcee9551888f474a7f1b1841312a71cc38
<pre>
[JSC][JSPI] Scan PinballCompletion conservative roots via a GC marking constraint
<a href="https://bugs.webkit.org/show_bug.cgi?id=307564">https://bugs.webkit.org/show_bug.cgi?id=307564</a>
<a href="https://rdar.apple.com/170156450">rdar://170156450</a>

Reviewed by Keith Miller.

In the existing JSPI implementation, evacuated Wasm stack slices and callee saves are
registered with the VM and scanned as conservative roots. This creates a potential memory
leak. Here is the reference structure:

    Promise --&gt; PinballCompletion --&gt; EvacuatedStacks, Callee Saves --&gt; ...
                                                ^
                                                |
                                               VM

The Promise is the promise that caused JSPI suspension, and it owns (transitively via
resolution handlers) the PinballCompletion with the suspended computation. Normally, when
the promise is fulfilled or rejected, PinballCompletion unregisters the evacuated stacks
and callee saves it owns, and all these objects become garbage-collectable. Also, if the
promise is forgotten without being resolved and becomes unmarked together with
PinballCompletion, PinballCompletion destructor unregisters the evacuated stacks and
callee saves, so everything is collected. However, if there happens to be a reference
chain starting in evacuated stacks or callee saves and leading back to the promise, this
transitive strong reference from the VM conservative roots will keep the promise and all
the other objects involved alive forever.

This patch prevents such a scenario. It stops treating evacuated stacks and callee saves
as VM conservative roots. Instead, a new GC constraint is introduced for PinballCompletion
objects, to scan the evacuated stacks and callee saves owned by a completion for
conservative roots only when the completion itself is marked.

This change requires some care in the implantation of evacuated stack slices. In the
original implementation the top slice is removed from a PinballCompletion and stored in
PinballHandlerContext by pinballHandlerInitContext(). Control would then return to
assembly code which would prepare the stack space for the slice and then call
pinballHandlerImplantSlice() to copy the slice data onto the stack. This creates a time
window when a stack slice has been taken out of its owner PinballCompletion, but its
contents have not yet been copied onto the stack. If GC were to run in this window, the
slice would not be scanned for conservative roots. This is not an immediate problem
because currently we do nothing in that time window that could start a GC, but it is a
fragility in case the code is changed in the future.

To avoid this fragility, we change the logic so that instead of storing the slice to
implant in PinballHandlerContext, we store the pinball object itself, and the slice is
only taken out of a PinballCompletion immediately before being implanted. We also refactor
the initialization code to make it an initializer of the PinballHandlerContext struct
instead of a static helper function.

Unified sources fixes
---------------------

The change in unified sources order with this patch, combined with the revision of JSC
header includes by <a href="https://bugs.webkit.org/show_bug.cgi?id=312585">https://bugs.webkit.org/show_bug.cgi?id=312585</a> caused build breakage.
The patch includes these additional changes unrelated to its core logic to fix the
breakage:

- TemporalPlainYearMonth.h / .cpp: Moved the addDurationToYearMonth template definition
  from the header into the .cpp file and added explicit template instantiations for
  AddOrSubtract::Add and AddOrSubtract::Subtract. The header only had a forward
  declaration of JSGlobalObject, so the globalObject-&gt;vm() call on line 91 failed with an
  incomplete type error.

- WeakMapImplInlines.h: Added #include &quot;Symbol.h&quot;. The canBeHeldWeakly function calls
  asSymbol() which is declared in Symbol.h, but the header only had a forward declaration
  of Symbol via the include chain.

- PinballHandlerContext.cpp: Added #include &quot;StackAlignment.h&quot;. The constructor asserts
  !(sliceByteSize % stackAlignmentBytes()), but stackAlignmentBytes() is declared in
  StackAlignment.h which wasn&apos;t included directly.

Testing
-------

- Added `jspi-gc-abandoned-suspensions.js` to reproduce the memory leak scenario described
  above.

- Refactoring changes are covered by existing tests.

Canonical link: <a href="https://commits.webkit.org/312253@main">https://commits.webkit.org/312253@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/4b3a3cada0950955e548056369e04cb9839a5d8a

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/159295 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/32723 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/25828 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/168125 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/113673 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [❌ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/6608f394-b52c-4c4a-b580-3a8bf86477d6) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/32791 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/32710 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/123427 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/113673 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [loading 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b1e0d2ff-d336-402e-86d1-9294dd067b39) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/162252 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/25683 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/143094 "Passed tests") | [❌ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/104094 "Found 2 new API test failures: TestWebCore:Color.RGBToHSL_Blue, TestWebCore:CookieJar.ShouldNotIncludeSecureCookiesForNonLocalHostnames (failure)") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/3e150883-80b5-426b-9256-e001e6b1216f) 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/24736 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/23170 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/15897 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/151345 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/134423 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/20874 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/170619 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/20128 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/16653 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/22500 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/131638 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/32412 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/27242 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/131750 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35640 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/32356 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/142667 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/90481 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/26416 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/19476 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/191578 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/31867 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/98312 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/49230 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/31387 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/31660 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/31542 "Built successfully") | | | |

### DIFF
--- a/Source/JavaScriptCore/heap/Heap.cpp
+++ b/Source/JavaScriptCore/heap/Heap.cpp
@@ -906,10 +906,7 @@ void Heap::gatherVMRoots(ConservativeRoots& roots)
         vm.scanSideState(roots);
     }
 #endif
-#if ENABLE(WEBASSEMBLY)
-    vm.gatherEvacuatedStackRoots(roots);
-#endif
-#if !(ENABLE(DFG_JIT) || ENABLE(WEBASSEMBLY))
+#if !ENABLE(DFG_JIT)
     UNUSED_PARAM(roots);
     UNUSED_VARIABLE(vm);
 #endif
@@ -3139,7 +3136,36 @@ void Heap::addCoreConstraints()
         })),
         ConstraintVolatility::GreyedByMarking,
         ConstraintParallelism::Parallel);
-    
+
+#if ENABLE(WEBASSEMBLY)
+    m_constraintSet->add(
+        "Pbc", "Pinball Completions",
+        MAKE_MARKING_CONSTRAINT_EXECUTOR_PAIR(([this] (auto& visitor) {
+            // FIXME: Unlike the "Cs" constraint which is skipped during verification
+            // because conservative roots are not stable, this skip is only here because
+            // ConservativeRoots::genericAddPointer asserts isMarking(), which doesn't
+            // hold during verification. This constraint could run always otherwise, but
+            // that would require rethinking the assumptions in ConservativeRoots.
+            if (m_isMarkingForGCVerifier)
+                return;
+            IsoSubspace* subspace = m_pinballCompletionSpace.get();
+            if (!subspace)
+                return;
+            ASSERT(worldIsStopped());
+            // FIXME: Add a second CellState for PinballCompletion so we can skip
+            // pinballs whose conservative roots have already been gathered this cycle.
+            ConservativeRoots conservativeRoots(*this);
+            subspace->forEachMarkedCell([&](HeapCell* cell, HeapCell::Kind) {
+                auto* pinball = uncheckedDowncast<PinballCompletion>(static_cast<JSCell*>(cell));
+                pinball->gatherConservativeRoots(conservativeRoots);
+            });
+            SetRootMarkReasonScope rootScope(visitor, RootMarkReason::PinballCompletionConservativeRoots);
+            visitor.append(conservativeRoots);
+        })),
+        ConstraintVolatility::GreyedByMarking,
+        ConstraintConcurrency::Sequential);
+#endif
+
 #if ENABLE(JIT)
     if (Options::useJIT()) {
         m_constraintSet->add(
--- a/JSTests/wasm/stress/jspi-gc-abandoned-suspensions.js
+++ b/JSTests/wasm/stress/jspi-gc-abandoned-suspensions.js
@@ -1,0 +1,55 @@
+//@ requireOptions("--useJSPI=1")
+
+import { instantiate } from "../wabt-wrapper.js"
+
+// When a suspending promise is abandoned without being resolved, the associated objects
+// should be garbage-collected even if there are reference cycles from conservative roots
+// on the evacuated stack back to the suspending promise.
+
+let wat = `
+(module
+  (import "env" "suspend" (func $suspend (param externref) (result i32)))
+  (func $entry (export "entry") (param $obj externref) (result i32)
+    local.get $obj
+    call $suspend
+  )
+)`;
+
+const count = 200;
+let weakRefs = [];
+
+{
+    const instance = await instantiate(wat, {
+        env: {
+            suspend: new WebAssembly.Suspending(function (obj) {
+                let promise = new Promise(() => {});
+                obj.promise = promise;
+                return promise;
+            })
+        }
+    });
+    const entry = WebAssembly.promising(instance.exports.entry);
+
+    for (let i = 0; i < count; i++) {
+        let obj = { index: i };
+        weakRefs.push(new WeakRef(obj));
+        entry(obj);
+        // Don't store the returned promise — abandon it immediately.
+    }
+}
+// instance, entry, obj, and all promises are now unreferenced.
+
+// Use setTimeout to cross a job boundary (required by the WeakRef spec
+// for targets to become clearable), then trigger a full GC.
+setTimeout(() => {
+    fullGC();
+
+    let collected = 0;
+    for (let weakRef of weakRefs) {
+        if (weakRef.deref() === undefined)
+            collected++;
+    }
+
+    if (collected == 0)
+        throw new Error("Expected at least some abandoned JSPI objects to be collected, but none were");
+}, 0);
--- a/Source/JavaScriptCore/JavaScriptCore.xcodeproj/project.pbxproj
+++ b/Source/JavaScriptCore/JavaScriptCore.xcodeproj/project.pbxproj
@@ -4051,6 +4051,7 @@
 		23A3568F2E9790F40039C82A /* PinballCompletion.cpp */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.cpp; path = PinballCompletion.cpp; sourceTree = "<group>"; };
 		23A969692E95A903005A36F5 /* EvacuatedStack.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = EvacuatedStack.h; sourceTree = "<group>"; };
 		23A9696A2E95A903005A36F5 /* EvacuatedStack.cpp */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.cpp; path = EvacuatedStack.cpp; sourceTree = "<group>"; };
+		23CD2E2C2F908FC900119878 /* PinballHandlerContext.cpp */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.cpp; path = PinballHandlerContext.cpp; sourceTree = "<group>"; };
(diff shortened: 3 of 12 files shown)


Source/JavaScriptCore/heap/Heap.cpp
```diff
@@ -3134,7 +3128,29 @@ void Heap::addCoreConstraints()
...
         ConstraintVolatility::GreyedByMarking,
         ConstraintParallelism::Parallel);
-    
+
+#if ENABLE(WEBASSEMBLY)
+    m_constraintSet->add(
+        "Pbc", "Pinball Completions",
+        MAKE_MARKING_CONSTRAINT_EXECUTOR_PAIR(([this] (auto& visitor) {
+            if (m_isMarkingForGCVerifier)
+                return;
```
Constellation, line 3150: Why do we need this check?

ast-hugger: My understanding is that ConservativeRoots can only be used in a real marking phase, as `genericAddPointer()` calls `isLiveCell()` saying `isMarking` is true, and asserts that the heap is marking. "Cs" has a similar guard, though in that case there's also an issue that roots on the stack may have changed, which shouldn't be a concern in our case.

Source/JavaScriptCore/heap/Heap.cpp
```diff
@@ -3137,7 +3134,29 @@ void Heap::addCoreConstraints()
...
+            IsoSubspace* subspace = m_pinballCompletionSpace.get();
+            if (!subspace)
+                return;
+            ASSERT(worldIsStopped());
+            ConservativeRoots conservativeRoots(*this);
+            subspace->forEachMarkedCell([&](HeapCell* cell, HeapCell::Kind) {
+                auto* pinball = uncheckedDowncast<PinballCompletion>(static_cast<JSCell*>(cell));
```
kmiller68, line 3159: Probably worth a FIXME to add a second CellState for this constraint on the PinballCompletion object so we don't have to rescan every pinball each time the constraint runs.

Source/JavaScriptCore/heap/Heap.cpp
kmiller68: We should probably add a FIXME that this might not be needed but it would require deep thoughts and changing some ASSERTs.

Source/JavaScriptCore/heap/Heap.cpp
```diff
@@ -3137,7 +3134,29 @@ void Heap::addCoreConstraints()
...
+                pinball->gatherConservativeRoots(conservativeRoots);
+            });
+            SetRootMarkReasonScope rootScope(visitor, RootMarkReason::ConservativeScan);
```
kmiller68: Probably worth a unique RootMarkReason.

kmiller68 (approved):
r=me if you add the new root mark reasons and FIXMEs

PR #62274, opened 2026-04-08
Parse 'systemLanguage' as a comma separated list

karlcow:
#### 94ce6308f5028935574e1f569f55df46de8a0473
<pre>
Parse &apos;systemLanguage&apos; as a comma separated list
<a href="https://bugs.webkit.org/show_bug.cgi?id=262146">https://bugs.webkit.org/show_bug.cgi?id=262146</a>
<a href="https://rdar.apple.com/116427520">rdar://116427520</a>

Reviewed by Darin Adler and Nikolas Zimmermann.

The SVG2 specification defines systemLanguage as
&quot;a set of comma-separated tokens&quot; (per the HTML definition),
where each token is a BCP 47 language tag. WebKit was parsing
it as space-separated, which meant &quot;en,fr,de&quot; was treated as
a single token.

SVGStringList now exposes two public methods:
* setFromCommaSeparatedTokens()
* setFromSpaceSeparatedTokens()
and as two private methods:
* parseCommaSeparatedTokens()
* parseSpaceSeparatedTokens()
systemLanguage uses the comma path;
requiredExtensions continues to use the space path.

Note: This PR is about parsing only. The validation logic
(matching user preferences against the parsed tags) is tracked
separately in <a href="https://webkit.org/b/262140.">https://webkit.org/b/262140.</a>

Note: This also invites to create a more general parseCommaSeparatedTokens
and parseSpaceSeparatedTokens, because there are other parts of WebCore
using similar needs. See <a href="https://webkit.org/b/312153.">https://webkit.org/b/312153.</a>

Spec: <a href="https://w3c.github.io/svgwg/svg2-draft/struct.html#ConditionalProcessingSystemLanguageAttribute">https://w3c.github.io/svgwg/svg2-draft/struct.html#ConditionalProcessingSystemLanguageAttribute</a>
An issue has been opened on the spec to clarify it too.
<a href="https://github.com/w3c/svgwg/issues/1089">https://github.com/w3c/svgwg/issues/1089</a>

Test: imported/w3c/web-platform-tests/svg/struct/systemLanguage-parsing.html

* LayoutTests/imported/w3c/web-platform-tests/svg/struct/systemLanguage-parsing-expected.txt: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/struct/systemLanguage-parsing.html: Added.
* LayoutTests/svg/dom/svg-list-properties-parser-leading-trailing-spaces-expected.txt:
[...]

PR #62975, opened 2026-04-17
SVGTests::isValid() does not correctly implement systemLanguage matching per SVG2

karlcow:
#### bcfac781bec780c78e7ef90f71293d8792a1d517
<pre>
SVGTests::isValid() does not correctly implement systemLanguage matching per SVG2
<a href="https://bugs.webkit.org/show_bug.cgi?id=262140">https://bugs.webkit.org/show_bug.cgi?id=262140</a>
<a href="https://rdar.apple.com/116424326">rdar://116424326</a>

Reviewed by Nikolas Zimmermann and Anne van Kesteren.

SVGTests::isValid() did not correctly implement the systemLanguage
matching algorithm defined in SVG2 5.6.5 and RFC 4647 3.3.1 (Basic
Filtering). It had four bugs:

1. Required ALL listed tags to match (should be: true if ANY
user preference matches ANY listed tag).
2. Truncated the user language to 2 characters, losing script
and region subtags (e.g. &quot;zh-Hans&quot; became &quot;zh&quot;).
3. Used case-sensitive comparison.
4. Did not implement BCP 47 prefix matching (user &quot;en&quot; should
match listed &quot;en-US&quot;).

The fix replaces the broken loop with two static helpers
matchesLanguageTag() implementing RFC 4647 Basic Filtering,
and userLanguageMatchesAnyListedTag() synthesizing parent language tags
by iterative subtag stripping and iterates over userPreferredLanguages()
instead of truncating defaultLanguage().

BUT in a way that preserve the privacy of the end users. This is the way
WebKit is architectured. The code deviates from the specification on
purpose.

Tests: imported/w3c/web-platform-tests/svg/struct/systemLanguage-matching.html
       svg/dom/systemLanguage-matching.html

* LayoutTests/imported/w3c/web-platform-tests/svg/struct/systemLanguage-matching-expected.txt: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/struct/systemLanguage-matching.html: Added.
* LayoutTests/svg/dom/systemLanguage-matching-expected.txt: Added.
* LayoutTests/svg/dom/systemLanguage-matching.html: Added.
* Source/WebCore/svg/SVGTests.cpp:
  (WebCore::matchesLanguageTag):
  (WebCore::userLanguageMatchesAnyListedTag):
  (WebCore::SVGTests::isValid const):

Canonical link: <a href="https://commits.webkit.org/315952@main">https://commits.webkit.org/315952@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/0ac466d7370d40a04676c43fc08078a2908ce8aa

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/170181 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/44104 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/37520 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/179543 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/127893 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/f215556f-9b4b-4d8b-8e8f-d6ee090a401d) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/172047 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/45348 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/43736 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/132666 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/93498 "13 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/896d0e39-3fb1-4ee6-b697-998b4236eb27) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/173140 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/35851 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/153096 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/113167 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/7554e49d-c7cd-4adc-9119-795b6a32ea0b) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/34440 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/32799 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/28239 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/144923 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/32494 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/182140 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/34929 "Built successfully") | [❌ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/36967 "Found 1 new test failure: http/tests/site-isolation/frame-index.html (failure)") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/141117 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/43761 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/152566 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/140941 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/38045 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/44450 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/29478 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/108836 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/36209 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/116330 "Built successfully") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/42960 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/7577 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/42352 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/42606 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/42495 "Built successfully") | | | |

### DIFF
--- a/LayoutTests/svg/dom/systemLanguage-matching-expected.txt
+++ b/LayoutTests/svg/dom/systemLanguage-matching-expected.txt
@@ -1,0 +1,29 @@
+systemLanguage matches against the same primary user language that navigator.language exposes. Parent language tags are synthesized from the primary language per the SVG2 5.6.5 implementation note.
+
+On success, you will see a series of "PASS" messages, followed by "TEST COMPLETE".
+
+
+--- User language en-US: parent-tag synthesis matches en, en-US, en-Latn-US ---
+PASS isRendered(createRect("en")) is true
+PASS isRendered(createRect("en-US")) is true
+PASS isRendered(createRect("en-Latn-US")) is true
+PASS isRendered(createRect("eng")) is false
+PASS isRendered(createRect("fr")) is false
+
+--- User language zh-Hans-CN: parent-tag synthesis matches zh, zh-Hans, zh-Hans-CN ---
+PASS isRendered(createRect("zh")) is true
+PASS isRendered(createRect("zh-Hans")) is true
+PASS isRendered(createRect("zh-Hans-CN")) is true
+PASS isRendered(createRect("zh-Hant")) is true
+
+--- Privacy: only the primary user language is matched, not secondary preferences ---
+PASS isRendered(createRect("fr")) is true
+PASS isRendered(createRect("fr-FR")) is true
+PASS isRendered(createRect("en")) is false
+PASS isRendered(createRect("en-US")) is false
+PASS isRendered(createRect("ja")) is false
+PASS isRendered(createRect("ja-JP")) is false
+PASS successfullyParsed is true
+
+TEST COMPLETE
+
--- a/Source/WebCore/svg/SVGTests.cpp
+++ b/Source/WebCore/svg/SVGTests.cpp
@@ -29,7 +29,6 @@
 #include "SVGPropertyOwnerRegistry.h"
 #include "SVGStringList.h"
 #include <wtf/Language.h>
-#include <wtf/SortedArrayMap.h>
 #include <wtf/TZoneMallocInlines.h>
 
 #if ENABLE(MATHML)
@@ -68,18 +67,50 @@ bool SVGTests::hasExtension(const String& extension)
     return extension == HTMLNames::xhtmlNamespaceURI;
 }
 
+// FIXME: RFC 4647 Basic Filtering with parent language tags can cause cross-script
+// matches (e.g., "zh" matching "zh-Hant" when user prefers "zh-Hans"). Consider
+// Extended Filtering (RFC 4647 3.3.2) for script-aware matching.
+static bool matchesLanguageTag(StringView range, StringView tag)
+{
+    if (range.isEmpty())
+        return false;
+    if (range.length() == tag.length())
+        return equalIgnoringASCIICase(range, tag);
+    return range.length() < tag.length()
+        && tag[range.length()] == '-'
+        && equalIgnoringASCIICase(range, tag.left(range.length()));
+}
+
+// SVG2 5.6.5 implementation note: try the full user language tag, then iteratively
+// strip subtags to synthesize parent language tags (e.g. "zh-Hans-CN" → "zh-Hans" → "zh").
+static bool userLanguageMatchesAnyListedTag(StringView userLanguage, const Vector<String>& listedTags)
+{
+    for (StringView range = userLanguage; !range.isEmpty(); ) {
+        for (auto& listedTag : listedTags) {
+            if (matchesLanguageTag(range, listedTag))
+                return true;
+        }
+        auto hyphenIndex = range.reverseFind('-');
+        if (hyphenIndex == notFound)
+            return false;
+        range = range.left(hyphenIndex);
+    }
+    return false;
+}
+
 bool SVGTests::isValid() const
 {
     auto attributes = conditionalProcessingAttributesIfExists();
     if (!attributes)
         return true;
 
-    String defaultLanguage = WTF::defaultLanguage();
-    auto genericDefaultLanguage = StringView(defaultLanguage).left(2);
-    for (auto& language : attributes->systemLanguage().items()) {
-        if (language != genericDefaultLanguage)
+    auto& systemLanguageList = attributes->systemLanguage().items();
+    if (!systemLanguageList.isEmpty()) {
+        // Match against the same value navigator.language / navigator.languages exposes.
+        if (!userLanguageMatchesAnyListedTag(defaultLanguage(), systemLanguageList))
             return false;
     }
+
     for (auto& extension : attributes->requiredExtensions().items()) {
         if (!hasExtension(extension))
             return false;
--- a/LayoutTests/imported/w3c/web-platform-tests/svg/struct/systemLanguage-matching-expected.txt
+++ b/LayoutTests/imported/w3c/web-platform-tests/svg/struct/systemLanguage-matching-expected.txt
@@ -1,0 +1,16 @@
+
+PASS Exact match with navigator.language renders the element
+PASS Case-insensitive exact match renders the element
+PASS Non-matching language tag does not render
+PASS Empty systemLanguage evaluates to false
+PASS Absent systemLanguage evaluates to true
+PASS Comma-separated list with matching tag first renders
+PASS Comma-separated list with matching tag second renders
+PASS Comma-separated list with matching tag in the middle renders
+PASS Comma-separated list with no matching tag does not render
+PASS BCP 47 prefix matching: user language prefix matches listed tag with extra subtag
+PASS switch element: first matching child renders, others do not
+PASS switch element: skips non-matching, renders first match
+PASS switch element: fallback child without systemLanguage renders when no match
+PASS switch element: nothing renders when no child matches and no fallback
+
--- a/LayoutTests/imported/w3c/web-platform-tests/svg/struct/systemLanguage-matching.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/svg/struct/systemLanguage-matching.html
@@ -1,0 +1,121 @@
+<!DOCTYPE html>
+<title>systemLanguage attribute matching per BCP 47 basic filtering</title>
+<link rel="help" href="https://w3c.github.io/svgwg/svg2-draft/struct.html#ConditionalProcessingSystemLanguageAttribute">
+<link rel="help" href="https://www.rfc-editor.org/rfc/rfc4647#section-3.3.1">
+<meta name="assert" content="The systemLanguage attribute evaluates to true if any user preferred language matches any listed tag per BCP 47 basic filtering.">
+<script src="/resources/testharness.js"></script>
+<script src="/resources/testharnessreport.js"></script>
+<svg id="svg" width="400" height="400" style="position: absolute; top: -9999px;">
+</svg>
+<script>
+const svgNS = "http://www.w3.org/2000/svg";
+const svg = document.getElementById("svg");
+const userLang = navigator.language;
+
+function isRendered(el) {
+    const bbox = el.getBoundingClientRect();
+    return bbox.width > 0 && bbox.height > 0;
+}
+
+function createRect(systemLanguage) {
+    const rect = document.createElementNS(svgNS, "rect");
+    rect.setAttribute("width", "10");
+    rect.setAttribute("height", "10");
+    if (systemLanguage !== null)
+        rect.setAttribute("systemLanguage", systemLanguage);
+    svg.appendChild(rect);
+    return rect;
+}
+
+function createSwitchWithRects(langList) {
+    const sw = document.createElementNS(svgNS, "switch");
+    const rects = [];
+    for (const lang of langList) {
+        const rect = document.createElementNS(svgNS, "rect");
+        rect.setAttribute("width", "10");
+        rect.setAttribute("height", "10");
+        if (lang !== null)
+            rect.setAttribute("systemLanguage", lang);
+        sw.appendChild(rect);
+        rects.push(rect);
+    }
+    svg.appendChild(sw);
+    return rects;
+}
+
+test(() => {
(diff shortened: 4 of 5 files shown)


karlcow:
This is the followup fix after https://github.com/WebKit/WebKit/pull/62274
to fix https://bugs.webkit.org/show_bug.cgi?id=262140

## Before the patch
Screenshot 2026-04-17 at 23 36 31

## After the patch
Screenshot 2026-04-17 at 23 36 13

Source/WebCore/svg/SVGTests.cpp
```diff
@@ -68,18 +67,49 @@ bool SVGTests::hasExtension(const String& extension)
...
         return true;
 
-    String defaultLanguage = WTF::defaultLanguage();
-    auto genericDefaultLanguage = StringView(defaultLanguage).left(2);
-    for (auto& language : attributes->systemLanguage().items()) {
-        if (language != genericDefaultLanguage)
+    auto& systemLanguageList = attributes->systemLanguage().items();
+    if (!systemLanguageList.isEmpty()) {
+        auto userLanguages = WTF::userPreferredLanguages();
+        bool matched = userLanguages.containsIf([&](const String& userLanguage) {
```
nikolaszimmermann: This looks really convoluted, how about extracting a static helper method, just like you did for matchesLanguageTag?

```
  // SVG2 5.6.5 implementation note: try the full user language tag, then iteratively
  // strip subtags to synthesize parent language tags (e.g. "zh-Hans-CN" → "zh-Hans" → "zh").
  static bool userLanguageMatchesAnyListedTag(StringView userLanguage, const Vector<String>& listedTags)
  {
      for (StringView range = userLanguage; !range.isEmpty(); ) {
          for (auto& listedTag : listedTags) {
              if (matchesLanguageTag(range, listedTag))
                  return true;
          }
          auto hyphenIndex = range.reverseFind('-');
          if (hyphenIndex == notFound)
              return false;
          range = range.left(hyphenIndex);
      }
      return false;
  }
```

This is more readable then:

```
      auto& systemLanguageList = attributes->systemLanguage().items();
      if (!systemLanguageList.isEmpty()) {
          bool anyUserLanguageMatches = WTF::userPreferredLanguages().containsIf([&](auto& userLanguage) {
              return userLanguageMatchesAnyListedTag(userLanguage, systemLanguageList);
          });
          if (!anyUserLanguageMatches)
              return false;
      }
```

karlcow: @nikolaszimmermann I created a new version of the code.

annevk:
@karlcow don't we have this logic elsewhere already for `:lang` and such? I think ideally we'd make all of that share the same underlying primitives.

karlcow:
@annevk that would be a great followup and specifically to check if the rules in svg2 spec CR are aligned with the rules in things like html and other parts of the code doing language selections. There's a couple of another thing which needs to be done after this PR, like https://bugs.webkit.org/show_bug.cgi?id=317386 to align the duck with the spec and other implementations.

karlcow:
1. `WTF::indexOfBestMatchingLanguageInList()`.  Exported, wrong semantics for SVG2, not the same rules, not RFC 4647 basic filtering.  https://searchfox.org/wubkat/rev/d221d71a6a560cb45bb3e9b23e526b6bcfa9e4b3/Source/WTF/wtf/Language.cpp#199-241
2. `WebCore::matchesLangPseudoClass()` + `containslanguageSubtagMatchingRange()`.  semantically correct, not reusable as-is, but maybe that could be changed in the future? But that's a different task, and maybe an SVG2 spec change? (to double check).

We could in a future PR extract `containslanguageSubtagMatchingRange` (and its caller's per-range loop) out of `SelectorCheckerTestFunctions.h` into `WTF::Language` as a generic  `languageRangeMatchesTag(StringView range, StringView tag)`, then having both `:lang()` and `SVGTests::isValid()` call it. But that's a bigger refactoring.

karlcow:
@annevk I opened https://bugs.webkit.org/show_bug.cgi?id=317462

nikolaszimmermann (approved):
Looks good to me, the refactoring with :lang can be addressed in a follow-up. Ok with you @annevk ?

Source/WebCore/svg/SVGTests.cpp
```diff
@@ -68,18 +67,52 @@ bool SVGTests::hasExtension(const String& extension)
...
     if (!attributes)
         return true;
 
-    String defaultLanguage = WTF::defaultLanguage();
-    auto genericDefaultLanguage = StringView(defaultLanguage).left(2);
-    for (auto& language : attributes->systemLanguage().items()) {
-        if (language != genericDefaultLanguage)
+    auto& systemLanguageList = attributes->systemLanguage().items();
+    if (!systemLanguageList.isEmpty()) {
+        bool anyUserLanguageMatches = WTF::userPreferredLanguages().containsIf([&](auto& userLanguage) {
```
annevk: I actually have a different concern. This would expose information that I don't think is otherwise exposed to this granularity. If we want to support this, it needs to return the exact same value as `navigator.language` (possibly `navigator.languages` if we implement that). And given the privacy risk I do think we want a shared path for that.

annevk (changes requested):

Source/WebCore/svg/SVGTests.cpp
karlcow: @annevk could you provide a testcase for your concern, so I can understand better.

karlcow: And if you need the current spec prose. 
https://w3c.github.io/svgwg/svg2-draft/single-page.html#struct-ConditionalProcessingSystemLanguageAttribute

annevk: What exactly is unclear? This ends up exposing information from the end user. We don't expose this raw data elsewhere (except for captions which are otherwise sandboxed). So we should use the same kind of protections existing public accessors have for this kind of data.

karlcow: The PR is implementing the current prose in the the spec which says "ordered list of preferred languages."

> [To evaluate the systemLanguage attribute with value V](https://w3c.github.io/svgwg/svg2-draft/single-page.html#struct-ConditionalProcessingSystemLanguageAttribute)
> …
> 4. Let user languages be the user's ordered list of preferred languages.
>

This privacy concern looks like a spec issue. The SVG2 spec could be changed to address this. Probably an issue should be opened to discuss it this week. 

I don't know who is **we** in the sentence:
> We don't expose this raw data

Nothing is unclear. I'm asking, could you come up with a test that will exhibit/expose the privacy issue? So It is easier to test across browsers. I'm also thinking about all the WebKit consumers, which are not Safari. I perfectly understand for the Safari case. I'm just trying to be careful.

TingPing: The test case is a user has a list of preferred languages (`en-US`, `en-Latn-US`). WebKit purposefully does not expose the full list, `navigator.languages` returns `["en-US"]`. This change allows iterating over every language to construct that full list.

annevk: End user languages is not something that is interoperable across browsers. We (as in WebKit) just shouldn't deviate from how we expose them today.

karlcow:
@annevk new version + a test to record the deviation from the spec.

LayoutTests/svg/dom/systemLanguage-matching-expected.txt
```diff
@@ -0,0 +1,29 @@
+WebKit deviates from SVG2 5.6.5 by matching systemLanguage against the primary user language only (the same value navigator.language exposes), not the full user preference list. Parent language tags are synthesized from the primary language per the SVG2 5.6.5 implementation note.
```
annevk: I'm not sure this is a deviation. Specifications can at best make SHOULD-level requirements about user interface.

Source/WebCore/svg/SVGTests.cpp
```diff
@@ -68,18 +67,56 @@ bool SVGTests::hasExtension(const String& extension)
...
-        if (language != genericDefaultLanguage)
+    auto& systemLanguageList = attributes->systemLanguage().items();
+    if (!systemLanguageList.isEmpty()) {
+        // SVG2 5.6.5 specifies matching against the user's "ordered list of preferred
+        // languages", but exposing that full list here would let a page probe the user's
+        // language preferences with finer granularity than navigator.language /
+        // navigator.languages (which intentionally surface only the primary language for
+        // privacy — see NavigatorBase::languages()). Match against the primary language
+        // only; parent-tag synthesis still applies, but only re-derives subtags already
+        // contained in the primary language.
```
annevk: I don't think this comment really helps. Perhaps we should point out that this needs to match navigator.languages, but that's about it.

Source/WebCore/svg/SVGTests.cpp
```diff
@@ -68,18 +67,56 @@ bool SVGTests::hasExtension(const String& extension)
...
+        if (!userLanguageMatchesAnyListedTag(WTF::defaultLanguage(), systemLanguageList))
```
annevk: We don't need the `WTF::` prefix here.

annevk (approved):
It seems okay to do the code-sharing as a follow-up.

Source/WebCore/svg/SVGTests.cpp
karlcow: modified.

PR #63018, opened 2026-04-18
Font invalidation: replace full style rebuild with targeted walk and lazy font cascade resolution

vitorroriz:
#### 7354211e17af8c7a95bd02e163a78237a2bbfa6e
<pre>
Font invalidation: replace full style rebuild with targeted walk and lazy font cascade resolution
<a href="https://bugs.webkit.org/show_bug.cgi?id=311368">https://bugs.webkit.org/show_bug.cgi?id=311368</a>
<a href="https://rdar.apple.com/174044985">rdar://174044985</a>

When a web font loads, WebKit calls scheduleFullStyleRebuild(), forcing
every element to be restyled and relaid out. On large DOMs this causes
significant main-thread blocking.

A font load is not a style change: the same declarations match the same
elements, and what changed is font data. So instead of invalidating
style, we mark the affected FontCascadeFonts as invalid and set
Document::m_needsFontInvalidation. On the next rendering update a single
walk decides, per element, whether it needs a restyle or only a
relayout, and font resolution happens lazily when something actually
asks for font data.

Only elements whose computed values were baked from font metrics need a
restyle: font-metric-relative units, font-size-adjust and
font-size:math. Everything else recovers through layout, which
re-measures text against the refreshed fonts.

Renderers that cache font-derived state are the exception, because they
relied on styleDidChange() firing whenever fonts changed and it no
longer does. RenderObject::fontsDidChange() gives each of them a place
to drop or re-derive that state next to the styleDidChange() code that
owns it.

Two performance tests are added, both invalidating a 15,000 element tree
when a font becomes available. The first uses 8,000 CSS rules that
collide on their rightmost selector, so style resolution dominates the
way it did on the reported page, and goes from 350ms to 37ms. The second
uses a font no element references, which is the case where the ideal cost
is zero, and goes from 44.5ms to 20ms.

Future optimization: per-font-family tracking would let us invalidate
only the FCFs that reference the font that loaded. The unused-font test
measures 21ms of invalidation for a font no element references, where the
ideal cost is zero, so that is the size of the prize. It would also make
it affordable to restyle affected elements rather than refresh fonts in
place, letting the existing styleDidChange machinery handle every
font-derived cache.
</pre>
https://github.com/WebKit/WebKit/commit/7354211e17af8c7a95bd02e163a78237a2bbfa6e

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/180429 "Failed to checkout and rebase branch from PR 63018") | [❌ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/54374 "Failed to checkout and rebase branch from PR 63018") | [❌ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/48172 "Failed to checkout and rebase branch from PR 63018") | [❌ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/190640 "Failed to checkout and rebase branch from PR 63018") | [❌ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/144750 "Failed to checkout and rebase branch from PR 63018") | ⏳ 🛠 ios-apple 
| [❌ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/182291 "Failed to checkout and rebase branch from PR 63018") | [❌ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/55672 "Failed to checkout and rebase branch from PR 63018") | [❌ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/54090 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/5/builds/190640 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/59/builds/144750 "Failed to checkout and rebase branch from PR 63018") | ⏳ 🛠 mac-apple 
| [❌ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/183384 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/155/builds/55672 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 api-mac](https://ews-build.webkit.org/#/builders/168/builds/48172 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/5/builds/190640 "Failed to checkout and rebase branch from PR 63018") | | ⏳ 🛠 vision-apple 
| | [❌ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/155/builds/55672 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/156/builds/54090 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 jsc-wpe](https://ews-build.webkit.org/#/builders/5/builds/190640 "Failed to checkout and rebase branch from PR 63018") | | 
| [❌ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/168/builds/48172 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 api-ios](https://ews-build.webkit.org/#/builders/155/builds/55672 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/168/builds/48172 "Failed to checkout and rebase branch from PR 63018") | [❌ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/1799 "Failed to checkout and rebase branch from PR 63018") | | 
| | [❌ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/46973 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/156/builds/54090 "Failed to checkout and rebase branch from PR 63018") | [❌ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/192575 "Failed to checkout and rebase branch from PR 63018") | | 
| | [❌ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/54040 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/168/builds/48172 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/2/builds/192575 "Failed to checkout and rebase branch from PR 63018") | | 
| | [❌ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/54728 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/168/builds/48172 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/2/builds/192575 "Failed to checkout and rebase branch from PR 63018") | | 
| | [❌ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/160/builds/54728 "Failed to checkout and rebase branch from PR 63018") | [❌ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/126991 "Failed to checkout and rebase branch from PR 63018") | [❌ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/122153 "Failed to checkout and rebase branch from PR 63018") | | 
| | [❌ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/53245 "Failed to checkout and rebase branch from PR 63018") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/168/builds/48172 "Failed to checkout and rebase branch from PR 63018") | | | 
| | [❌ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/52627 "Failed to checkout and rebase branch from PR 63018") | | | | 
| | [❌ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/52864 "Failed to checkout and rebase branch from PR 63018") | | | | 
| | [❌ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/52692 "Failed to checkout and rebase branch from PR 63018") | | | |


Source/WebCore/css/CSSFontSelector.cpp
```diff
@@ -315,13 +317,39 @@ void CSSFontSelector::opportunisticallyStartFontDataURLLoading(const FontCascade
...
+        break;
+    case CSSFontFaceSet::FontModificationKind::Removal:
+        invalidateAssociatedFontCascadeFonts(FontInvalidationReason::FontDeleted);
+        break;
+    case CSSFontFaceSet::FontModificationKind::PropertyChange:
+        invalidateAssociatedFontCascadeFonts(FontInvalidationReason::Generic);
+        break;
+    }
+
+    for (auto& client : copyToVector(m_clients)) {
```
fantasai: Why are we making a copy?

Source/WebCore/css/CSSSegmentedFontFace.cpp
```diff
@@ -91,6 +91,8 @@ class CSSFontAccessor final : public FontAccessor {
         return m_result && m_result.value() && m_result.value()->isInterstitial();
     }
 
+    bool isCSSFontAccessor() const final { return true; }
```
fantasai: Should this be marked as `override`?

Source/WebCore/dom/Document.cpp
```diff
@@ -2883,6 +2888,9 @@ void Document::resolveStyle(ResolveStyleType type)
         Style::PostResolutionCallbackDisabler disabler(*this);
         WidgetHierarchyUpdatesSuspensionScope suspendWidgetHierarchyUpdates;
 
+        if (std::exchange(m_needsFontInvalidation, false))
+            performFontInvalidation();
```
fantasai: Would it make sense to just check the boolean here and set m_needsFontInvalidation to false inside performFontInvalidation()?

Source/WebCore/dom/Document.cpp
```diff
@@ -3481,9 +3499,90 @@ void Document::invalidateMatchedPropertiesCacheAndForceStyleRecalc()
...
+{
+    if (!renderView())
+        return;
+
+    bool needsStyleRecalc = false;
+    if (RefPtr root = documentElement()) {
+        for (Ref element : descendantsOfType<Element>(*root)) {
+            auto dependsOnFontMetrics = [](const Style::ComputedStyle* style) {
+                return style && (style->dependsOnFontMetrics()
+                    || !style->fontDescription().fontSizeAdjust().isNone());
```
fantasai: Should this check be folded into `dependsOnFontMetrics()`?

fantasai (commented):
Got partway through before my brain started melting. 🫠 Will have to take another pass when I'm less exhausted. :) Here's what I've got so far.

Source/WebCore/dom/Document.cpp
vitorroriz: yes!

Source/WebCore/dom/Document.cpp
```diff
@@ -3435,9 +3450,72 @@ void Document::invalidateMatchedPropertiesCacheAndForceStyleRecalc()
...
+        }
+        return false;
+    };
+
+    // Phase 1: restyle elements whose computed value was baked from font metrics.
+    // Selector matching and the cascade never change, only these specific values do.
+    bool needsStyleRecalc = false;
+    for (Ref element : descendantsOfType<Element>(*this)) {
+        if (!anyStyleDependsOnFontMetrics(element))
+            continue;
```
anttijk, line 3494: This is misses all shadow trees. You could use `composedTreeDescendants` or recursive traversal (as order does not matter). There should be a test for this. 

Your perf test uses shadow trees so numbers may be bogus for that reason too.

Generally there should be correctness tests triggering the optimized case specifically as I don't think it happens that much during existing tests.

Source/WebCore/dom/Document.cpp
```diff
@@ -3435,9 +3450,72 @@ void Document::invalidateMatchedPropertiesCacheAndForceStyleRecalc()
...
+
+        element->invalidateStyle();
+        needsStyleRecalc = true;
+    }
+
+    // Phase 2: relayout renderers whose font actually went invalid, regardless of
+    // whether their style has any font-metric-dependent value.
+    Vector<CheckedRef<RenderObject>> renderersWithInvalidatedFonts;
+    for (CheckedPtr descendant = renderView()->firstChild(); descendant; descendant = descendant->nextInPreOrder(renderView())) {
+        RefPtr fontCascadeFonts = descendant->style().fontCascade().existingFonts();
```
anttijk, line 3504: Phase 1 looks into pseudo-element styles but this doesn't. It will still get most cases right but not the ones that don't generate boxes like `::first-line/::selected/etc`. Also testable.

Source/WebCore/dom/Document.cpp
```diff
@@ -3435,9 +3450,72 @@ void Document::invalidateMatchedPropertiesCacheAndForceStyleRecalc()
...
+            continue;
+
+        element->invalidateStyle();
+        needsStyleRecalc = true;
+    }
+
+    // Phase 2: relayout renderers whose font actually went invalid, regardless of
+    // whether their style has any font-metric-dependent value.
+    Vector<CheckedRef<RenderObject>> renderersWithInvalidatedFonts;
+    for (CheckedPtr descendant = renderView()->firstChild(); descendant; descendant = descendant->nextInPreOrder(renderView())) {
```
anttijk, line 3503: This could use `descendantsOfType<RenderElement>`. `RenderText`s have the same style as their parent element.

PerformanceTests/CSS/FontInvalidationComplexStyle.html
```diff
@@ -0,0 +1,132 @@
...
+    frameDocument.fonts.delete(probeFont);
+    frameDocument.body.offsetHeight;
+    const widthWithoutProbeFont = probeElement.offsetWidth;
+    frameDocument.fonts.add(probeFont);
+    frameDocument.body.offsetHeight;
+    if (probeElement.offsetWidth === widthWithoutProbeFont)
+        throw new Error("Adding the font did not change layout within the measured region.");
+
+    const test = {
+        description: "Measures invalidating a large, style-heavy DOM when a web font the page uses becomes available. The 8,000 rules all end in the same class so they land in one RuleSet bucket, which makes selector matching, and therefore style resolution, dominate rather than layout. That is the ratio reported on the page this work came from: style recalc 97% of the freeze, layout 3%. Parameterised via the query string so the size can be calibrated.",
```
anttijk, line 113: Claude tells me these perf test do not test what it claims to test. 

```
  The perf numbers measure the wrong path

  Both perf tests drive invalidation with frameDocument.fonts.add() / .delete(). That goes through
  CSSFontFaceSet's FontModifiedObserver → CSSFontSelector::fontModified() →
  dispatchInvalidationCallbacks() with the default Generic reason — which invalidates every
  FontCascadeFonts on the selector. Only CSSFontSelector::fontLoaded() passes FontLoaded, i.e. the
  isLoadingCustomFonts()-filtered path this patch is actually about.

  So FontInvalidationUnusedFont doesn't measure "a font no element references" at all: under
  Generic, every FCF is invalid, Phase 2 dirties layout on all 15,000 elements, and the 20ms is a
  full relayout plus two full-tree walks. The commit message's "the unused-font test measures 21ms
  of invalidation for a font no element references... so that is the size of the prize" is a
  conclusion about per-font-family tracking drawn from a path where the per-font filter is bypassed
  entirely. FontInvalidationComplexStyle's 350ms→37ms is a real win but also isn't the "web font the
  page uses becomes available" scenario it describes.

  The async case in your layout test already gets the ordering right — add(face) then face.load().
  The perf tests do load() then add().

  Relatedly: a document-level "some style depends on font metrics" bit, mirroring
  hasStyleWithViewportUnits(), would let performFontInvalidation skip Phase 1 outright. A good share
  of the residual cost is the two unconditional tree walks, not the missing per-family tracking.
```

Source/WebCore/layout/integration/inline/LayoutIntegrationLineLayout.cpp
```diff
@@ -1500,6 +1513,23 @@ void LineLayout::releaseCachesAndResetDamage()
...
+void LineLayout::invalidateCachesIfFontChanged()
+{
+    if (auto* fontSelector = m_document ? m_document->fontSelectorIfExists() : nullptr) {
+        auto currentVersion = fontSelector->version();
+        if (m_fontSelectorVersion != currentVersion) {
+            m_fontSelectorVersion = currentVersion;
+            fontsNeedUpdate();
+        }
+    }
+}
```
anttijk, line 1531: This seems strictly worse than the more fine-grained tree-walking invalidation path. We end up dropping these very perf sensitive caches unnecessarily. Can we just drop the caches during tree walking invalidation if needed?

Source/WebCore/dom/Document.cpp
```diff
@@ -3435,9 +3450,72 @@ void Document::invalidateMatchedPropertiesCacheAndForceStyleRecalc()
 
     if (backForwardCacheState() != NotInBackForwardCache || !renderView())
         return;
+
     scheduleFullStyleRebuild();
 }
 
+void Document::performFontInvalidation()
```
anttijk, line 3457: `Style::DocumentScope` might be a better place for this invalidation code as `Document` is bloated already.

anttijk (changes requested):

Source/WebCore/dom/Document.cpp
```diff
@@ -3435,9 +3450,72 @@ void Document::invalidateMatchedPropertiesCacheAndForceStyleRecalc()
...
+
+    // Phase 2: relayout renderers whose font actually went invalid, regardless of
+    // whether their style has any font-metric-dependent value.
+    Vector<CheckedRef<RenderObject>> renderersWithInvalidatedFonts;
+    for (CheckedPtr descendant = renderView()->firstChild(); descendant; descendant = descendant->nextInPreOrder(renderView())) {
+        RefPtr fontCascadeFonts = descendant->style().fontCascade().existingFonts();
+        if (!fontCascadeFonts || fontCascadeFonts->isValid())
+            continue;
+
+        renderersWithInvalidatedFonts.append(*descendant);
```
anttijk, line 3508: Can the render tree invalidation just be handled by `RenderTreeUpdater`? That's what it is for.

vitorroriz: The problem is that style won't necessarily change when a font loads and I believe RenderTreeUpdater relies on that (?)

vitorroriz: I can try to expand RenderTreeUpdater to handle it but we would still need to do our own walk I believe

PR #63082, opened 2026-04-19
Add site-specific quirk to make anchor elements mouse-focusable on thesaurus.com and dictionary.com

swpatters:
#### 4509531a0302a134f49aae23194372ed4ba9d0d8
<pre>
Add site-specific quirk to make anchor elements mouse-focusable on thesaurus.com and dictionary.com
<a href="https://bugs.webkit.org/show_bug.cgi?id=312692">https://bugs.webkit.org/show_bug.cgi?id=312692</a>
<a href="https://rdar.apple.com/174959285">rdar://174959285</a>

Reviewed by Abrar Protyasha.

Clicking links on thesaurus.com and dictionary.com causes blur events with a
null relatedTarget, breaking dropdown menus that rely on relatedTarget to stay
open.

Rather than changing the default behavior for all sites, add a site-specific
quirk following the existing needsFormControlToBeMouseFocusable pattern. The
broader behavior discussion is tracked in bug 22261.

Widen platform guards from PLATFORM(MAC) to PLATFORM(COCOA) so the quirk is
also effective on iPad desktop-class browsing and visionOS.

* LayoutTests/fast/events/anchor-mouse-focusable-quirk-expected.txt: Added.
* LayoutTests/fast/events/anchor-mouse-focusable-quirk.html: Added.
* Source/WebCore/html/HTMLAnchorElement.cpp:
(WebCore::HTMLAnchorElement::isMouseFocusable):
(WebCore::Quirks::needsAnchorToBeMouseFocusable):
(WebCore::handleThesaurusQuirks):
(WebCore::handleDictionaryQuirks):
(WebCore::Quirks::determineRelevantQuirks):
* Source/WebCore/page/Quirks.h:

Canonical link: <a href="https://commits.webkit.org/311907@main">https://commits.webkit.org/311907@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/653e5173fb45d33e228f6b87bba3886e6370be24

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/158187 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/31524 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/24718 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/167016 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/112270 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/160058 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/31661 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/31542 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/122503 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85992 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/161145 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/24775 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/142075 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/103172 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/23831 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/22189 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/14788 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/133515 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/19867 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/169505 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/14859 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/21490 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/130687 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/31270 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/26253 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/130802 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35453 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/31208 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/141661 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/89104 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/25509 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/18467 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/30760 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/96293 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/30281 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/30511 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/30408 "Built successfully") | | |

### DIFF
--- a/LayoutTests/fast/events/anchor-mouse-focusable-quirk.html
+++ b/LayoutTests/fast/events/anchor-mouse-focusable-quirk.html
@@ -1,0 +1,42 @@
+<!DOCTYPE html>
+<!-- Test for the anchor mouse-focusability quirk on thesaurus.com and dictionary.com.
+     Anchor elements with href but no explicit tabindex are normally not mouse-focusable.
+     The quirk makes them focusable on click for these specific sites. -->
+<html>
+<head>
+<script src="../../resources/js-test.js"></script>
+</head>
+<body>
+<a id="link" href="#">Click me</a>
+<script>
+description("Tests that the anchor mouse-focusability quirk works on thesaurus.com and dictionary.com, but not on unrelated sites.");
+
+var link = document.getElementById("link");
+
+function clickLink() {
+    let rect = link.getBoundingClientRect();
+    eventSender.mouseMoveTo(rect.left + rect.width / 2, rect.top + rect.height / 2);
+    eventSender.mouseDown();
+    eventSender.mouseUp();
+}
+
+// Test 1: thesaurus.com — anchor should become focused on click.
+document.activeElement?.blur();
+window.internals.setTopDocumentURLForQuirks("https://www.thesaurus.com");
+clickLink();
+shouldBe("document.activeElement", "link");
+
+// Test 2: dictionary.com — anchor should become focused on click.
+document.activeElement?.blur();
+window.internals.setTopDocumentURLForQuirks("https://www.dictionary.com");
+clickLink();
+shouldBe("document.activeElement", "link");
+
+// Test 3: unrelated site — anchor should NOT become focused on click.
+document.activeElement?.blur();
+window.internals.setTopDocumentURLForQuirks("https://www.example.com");
+clickLink();
+shouldNotBe("document.activeElement", "link");
+</script>
+</body>
+</html>
--- a/Source/WebCore/html/HTMLAnchorElement.cpp
+++ b/Source/WebCore/html/HTMLAnchorElement.cpp
@@ -30,6 +30,7 @@
 #include "ContainerNodeInlines.h"
 #include "DOMTokenList.h"
 #include "DocumentPage.h"
+#include "DocumentQuirks.h"
 #include "DocumentSecurityOrigin.h"
 #include "ElementAncestorIteratorInlines.h"
 #include "EventHandler.h"
@@ -110,7 +111,7 @@ bool HTMLAnchorElement::isMouseFocusable() const
 {
 #if !(PLATFORM(GTK) || PLATFORM(WPE))
     // Only allow links with tabIndex or contentEditable to be mouse focusable.
-    if (isLink())
+    if (isLink() && !protect(document())->quirks().needsAnchorToBeMouseFocusable())
         return HTMLElement::supportsFocus();
 #endif
 
--- a/Source/WebCore/page/Quirks.cpp
+++ b/Source/WebCore/page/Quirks.cpp
@@ -247,6 +247,18 @@ bool Quirks::isEmbedDomain(const String& domainString) const
     return RegistrableDomain(document->url()).string() == domainString;
 }
 
+// thesaurus.com, dictionary.com https://bugs.webkit.org/show_bug.cgi?id=312692 rdar://174959285
+bool Quirks::needsAnchorToBeMouseFocusable() const
+{
+#if PLATFORM(COCOA)
+    QUIRKS_EARLY_RETURN_IF_DISABLED_WITH_VALUE(false);
+
+    return m_quirksData.quirkIsEnabled(QuirksData::SiteSpecificQuirk::NeedsAnchorToBeMouseFocusableQuirk);
+#else
+    return false;
+#endif // PLATFORM(COCOA)
+}
+
 // ceac.state.gov https://bugs.webkit.org/show_bug.cgi?id=193478
 // weather.com rdar://139689157
 // madisoncity.k12.al.us https://bugs.webkit.org/show_bug.cgi?id=296989
@@ -2766,26 +2778,16 @@ static void handleSlackQuirks(QuirksData& quirksData, const URL&, const String& 
 #endif
 }
 
+#if ENABLE(DESKTOP_CONTENT_MODE_QUIRKS)
 static void handleScriptToEvaluateBeforeRunningScriptFromURLQuirk(QuirksData& quirksData, const URL& /* quirksURL */, const String& topDomain, const URL& /* documentURL */)
 {
-    if (topDomain == "dictionary.com"_s) {
-        quirksData.isDictionary = true;
-        quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk);
-    }
-
-    if (topDomain == "thesaurus.com"_s) {
-        quirksData.isThesaurus = true;
-        quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk);
-    }
-
-#if ENABLE(DESKTOP_CONTENT_MODE_QUIRKS)
     if (topDomain == "webex.com"_s) {
         quirksData.isWebEx = true;
         quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk);
     }
-#endif
 }
 #endif
+#endif // PLATFORM(IOS_FAMILY)
 
 #if ENABLE(TWO_PHASE_CLICKS)
 static void handleWalmartQuirks(QuirksData& quirksData, const URL& /* quirksURL */, const String& quirksDomainString, const URL&  /* documentURL */)
@@ -2892,6 +2894,32 @@ static void handleWPDevelopmentQuirks(QuirksData& quirksData, const URL& /* quir
     quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsFormControlToBeMouseFocusableQuirk);
 }
 #endif
+
+static void handleThesaurusQuirks(QuirksData& quirksData, const URL& /* quirksURL */, const String& quirksDomainString, const URL&  /* documentURL */)
+{
+    QUIRKS_EARLY_RETURN_IF_NOT_DOMAIN("thesaurus.com"_s);
+
+    quirksData.isThesaurus = true;
+#if PLATFORM(IOS_FAMILY)
+    quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk);
+#endif
+#if PLATFORM(COCOA)
+    quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsAnchorToBeMouseFocusableQuirk);
+#endif
+}
+
+static void handleDictionaryQuirks(QuirksData& quirksData, const URL& /* quirksURL */, const String& quirksDomainString, const URL&  /* documentURL */)
+{
+    QUIRKS_EARLY_RETURN_IF_NOT_DOMAIN("dictionary.com"_s);
+
+    quirksData.isDictionary = true;
+#if PLATFORM(IOS_FAMILY)
+    quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk);
+#endif
+#if PLATFORM(COCOA)
+    quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsAnchorToBeMouseFocusableQuirk);
+#endif
+}
 
 static void handleTikTokQuirks(QuirksData& quirksData, const URL& /* quirksURL */, const String& quirksDomainString, const URL&  /* documentURL */)
 {
@@ -3802,8 +3830,8 @@ void Quirks::determineRelevantQuirks()
         { "crunchyroll"_s, &handleCrunchyRollQuirks },
         { "t-mobile"_s, &handleTMobileQuirks },
         { "descript"_s, &handleDescriptQuirks },
+        { "dictionary"_s, &handleDictionaryQuirks },
 #if PLATFORM(IOS_FAMILY)
-        { "dictionary"_s, &handleScriptToEvaluateBeforeRunningScriptFromURLQuirk },
         { "disneyplus"_s, &handleDisneyPlusQuirks },
 #endif
         { "ea"_s, &handleEAQuirks },
@@ -3881,8 +3909,8 @@ void Quirks::determineRelevantQuirks()
         { "state"_s, &handleCEACStateGovQuirks },
 #if PLATFORM(IOS_FAMILY)
         { "theguardian"_s, &handleGuardianQuirks },
-        { "thesaurus"_s, &handleScriptToEvaluateBeforeRunningScriptFromURLQuirk },
 #endif
+        { "thesaurus"_s, &handleThesaurusQuirks },
         { "tiktok"_s, &handleTikTokQuirks },
 #if PLATFORM(MAC)
         { "trix-editor"_s, &handleTrixEditorQuirks },
--- a/LayoutTests/fast/events/anchor-mouse-focusable-quirk-expected.txt
+++ b/LayoutTests/fast/events/anchor-mouse-focusable-quirk-expected.txt
@@ -1,0 +1,12 @@
+Tests that the anchor mouse-focusability quirk works on thesaurus.com and dictionary.com, but not on unrelated sites.
+
+On success, you will see a series of "PASS" messages, followed by "TEST COMPLETE".
+
+
+PASS document.activeElement is link
+PASS document.activeElement is link
+PASS document.activeElement is not link
+PASS successfullyParsed is true
+
+TEST COMPLETE
+Click me
--- a/LayoutTests/platform/glib/TestExpectations
+++ b/LayoutTests/platform/glib/TestExpectations
@@ -5484,6 +5484,9 @@ webkit.org/b/313058 media/media-vp8-webm.html [ Pass Timeout ImageOnlyFailure ]
 
 webkit.org/b/312985 svg/transforms/svg-transform-scale-with-and-without-layer.html [ ImageOnlyFailure ]
 
+# Anchor mouse-focusability quirk is Cocoa-only.
+webkit.org/b/312692 fast/events/anchor-mouse-focusable-quirk.html [ Skip ]
+
 # End: Common failures between GTK and WPE.
 
 #////////////////////////////////////////////////////////////////////////////////////////
--- a/LayoutTests/platform/ios/TestExpectations
+++ b/LayoutTests/platform/ios/TestExpectations
@@ -5672,6 +5672,7 @@ fast/dom/Geolocation/startUpdatingOnlyWhenPageVisible.html [ Skip ]
 fast/dom/Geolocation/stopUpdatingForHiddenPage.html [ Skip ]
 
 # eventSender.mouseDown is not implemented
+fast/events/anchor-mouse-focusable-quirk.html [ Skip ]
 fast/loader/location-hash-user-gesture.html [ Skip ]
 imported/blink/editing/selection/selectstart-event-crash.html [ Skip ]
 imported/blink/fast/forms/label/label-contains-other-interactive-content.html [ Skip ]
--- a/LayoutTests/platform/win/TestExpectations
+++ b/LayoutTests/platform/win/TestExpectations
@@ -2432,6 +2432,7 @@ fast/canvas/font-family-system-ui-canvas.html [ Skip ] # ImageOnlyFailure
 fast/css-grid-layout/grid-align-baseline-vertical.html [ Failure ]
 fast/dom/call-a-constructor-as-a-function.html [ Failure ]
 fast/dynamic/float-in-trailing-whitespace-after-last-line-break-2.html [ Failure ]
+fast/events/anchor-mouse-focusable-quirk.html [ Skip ] # Cocoa-only quirk
 fast/events/attempt-scroll-with-no-scrollbars.html [ Failure ]
 fast/events/autoscroll-when-input-is-offscreen.html [ Skip ] # UIScript
 fast/events/autoscroll-with-software-keyboard.html [ Skip ] # UIScript
--- a/Source/WebCore/page/Quirks.h
+++ b/Source/WebCore/page/Quirks.h
@@ -71,6 +71,7 @@ public:
     bool shouldDeferIntersectionObserversDuringResize() const;
     bool NODELETE shouldSilenceMediaQueryListChangeEvents() const;
     bool shouldIgnoreInvalidSignal() const;
+    bool needsAnchorToBeMouseFocusable() const;
     bool needsFormControlToBeMouseFocusable() const;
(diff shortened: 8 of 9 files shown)


aprotyas (changes requested):
The technical approach is sound, and it is a real web compat problem that you're trying to fix, but this touches a deliberately-maintained platform behavior with extensive history (https://bugs.webkit.org/show_bug.cgi?id=22261). We'd need broader discussion before landing.

Also, worth considering: HTMLFormControlElement::isMouseFocusable() has a site-specific quirk to work around this as a stop-gap. Maybe we can employ something like that for anchors, which would allow us to fix thesaurus.com without committing to a blanket behavior change right now.

karlcow (commented):
for form elements. some quirks have been put in place to make some websites working. 

https://github.com/WebKit/WebKit/blob/367f77ea6640540501134dc9aaf9ca73e7da0886/Source/WebCore/page/Quirks.cpp#L253-L262

https://github.com/WebKit/WebKit/blob/367f77ea6640540501134dc9aaf9ca73e7da0886/Source/WebCore/html/HTMLFormControlElement.cpp#L230-L241

Just for information.

swpatters:
> The technical approach is sound, and it is a real web compat problem that you're trying to fix, but this touches a deliberately-maintained platform behavior with extensive history (https://bugs.webkit.org/show_bug.cgi?id=22261). We'd need broader discussion before landing.
> 
> Also, worth considering: HTMLFormControlElement::isMouseFocusable() has a site-specific quirk to work around this as a stop-gap. Maybe we can employ something like that for anchors, which would allow us to fix thesaurus.com without committing to a blanket behavior change right now.

That's fair. You're right. The original blocker — no way to distinguish mouse vs keyboard focus — is now solved.  Support for :focus-visible shipped and is implemented properly. The original reason for the Mac restriction no longer applies.  But, yes, there should be more discussion. Maybe a quirk is the route to take.

karlcow:
@swpatters Do you want me to create the quirk or would you like to give a stab at it?

swpatters:
> @swpatters Do you want me to create the quirk or would you like to give a stab at it?

Whatever works. I'm inclined to wait for more conversation about it overall. @karlcow If you think a quirk is better suited here, I can change it to a quirk for thesaurus.com.

waterplea:
Why are you making quirks targeting specific websites? We all want a working browser, what's so special about thesaurus?

othermaciej:
> That's fair. You're right. The original blocker — no way to distinguish mouse vs keyboard focus — is now solved. Support for :focus-visible shipped and is implemented properly. The original reason for the Mac restriction no longer applies. But, yes, there should be more discussion. Maybe a quirk is the route to take.

Speaking as someone who was around back when it was implemented: visible focus ring wasn't the original reason for the Mac mouse and keyboard focus behavior. The original intent was to be consistent with the Mac platform and HI guidelines. On Mac, in native apps, only controls that the user can type in such as text fields can receive keyboard and mouse focus by default. The system setting "Keyboard Navigation" overrides this and makes all controls focusable. Note that even if no focus ring is drawn, focusing a link or button changes where the focus will go next when the user hits TAB.

We realize that this longstanding behavior is at odds with web developer expectations, since other browsers (even on Mac) implement Windows-like behavior by default, where any control can be focused by clicking. We have been discussing whether we need to entirely give up on the historical desire for consistency with Mac native apps, or if we can come up with some partway measure that preserves both Mac-like behavior and web compatibility. Possibilities include focusing but hiding the focus ring (imperfect), setting focus temporarily but then unsetting (unclear if this will be compatible enough) or hiding the focus ring and reseting focus to top level when the user performs keyboard focus navigation (probably compatible but kind of complicated). An additional complication is accessibility. VoiceOver users are notified of focus changes. If a user uses VoiceOver without Keyboard Navigation, telling them about a focus change that will be reset as soon as they hit TAB could be confusing.

Sorry for the longwinded explanation but I wanted to make sure folks proceed based on correct information about the intent of this behavior and the state of discussions about changing it.

waterplea:
> setting focus temporarily but then unsetting (unclear if this will be compatible enough) or hiding the focus ring and reseting focus to top level when the user performs keyboard focus navigation 

All of these will end up being Lovecraftian level horrors for the web devs trying to create a simple dropdown menu that should close when you tab away to the next element. 

What do you mean focusing but not showing the focus ring being imperfect? This is literally expected behaviour that has been in place everywhere but in Safari for 20+ years

othermaciej:
> What do you mean focusing but not showing the focus ring being imperfect? This is literally expected behaviour that has been in place everywhere but in Safari for 20+ years

@waterplea if by "everywhere" you mean "other web browsers" then you are correct. If by "everywhere" you mean in other Mac (or iPad w/ keyboard) applications, you are not correct. The Mac has never had this behavior in native UI. WebKit has had a goal to maximally fit in with native Mac UI since its inception. It is even used to emulate otherwise native-looking UI in some apps, where the user doesn't even know they are looking at web content. We have to balance that against web compatibility and expectations set by other browsers of course. I just want to make folks understand, if the situation seems simple and obvious to you, it is more complicated than you think.

Rest assured also that we will do thorough testing before unleashing any Lovecraftian horrors[*]. And we will do something simple if we can't make a complicated solution work. Examples of sites that are dependent on links or buttons getting click-focus would help!

[*] A lot of browser internals are Lovecraftian horrors by necessity, so just the fact that a solution is internally complicated and horrible doesn't mean it's wrong (see JS semicolon insertion, HTML parsing algorithm, document.all...). But we do want to avoid unleashing needless horror on users or web developers.

Source/WebCore/page/Quirks.cpp
```diff
@@ -2753,6 +2766,10 @@ static void handleScriptToEvaluateBeforeRunningScriptFromURLQuirk(QuirksData& qu
     if (topDomain == "thesaurus.com"_s) {
         quirksData.isThesaurus = true;
         quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk);
+#if PLATFORM(MAC)
+        // thesaurus.com https://bugs.webkit.org/show_bug.cgi?id=312692 rdar://174959285
+        quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsAnchorToBeMouseFocusableQuirk);
```
pxlcoder: Does not seem right to be enabling this quirk under `handleScriptToEvaluateBeforeRunningScriptFromURLQuirk`.

Source/WebCore/page/Quirks.cpp
```diff
@@ -3851,8 +3867,8 @@ void Quirks::determineRelevantQuirks()
         { "state"_s, &handleCEACStateGovQuirks },
 #if PLATFORM(IOS_FAMILY)
         { "theguardian"_s, &handleGuardianQuirks },
-        { "thesaurus"_s, &handleScriptToEvaluateBeforeRunningScriptFromURLQuirk },
 #endif
+        { "thesaurus"_s, &handleScriptToEvaluateBeforeRunningScriptFromURLQuirk },
```
pxlcoder: I think you want to add a new method `handleThesaurusQuirks`? We don't want the script quirk for Mac.

Source/WebCore/page/Quirks.cpp
```diff
@@ -2753,6 +2766,10 @@ static void handleScriptToEvaluateBeforeRunningScriptFromURLQuirk(QuirksData& qu
     if (topDomain == "thesaurus.com"_s) {
         quirksData.isThesaurus = true;
         quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk);
+#if PLATFORM(MAC)
+        // thesaurus.com https://bugs.webkit.org/show_bug.cgi?id=312692 rdar://174959285
+        quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsAnchorToBeMouseFocusableQuirk);
+#endif
```
karlcow: You need to put it outside of `handleScriptToEvaluateBeforeRunningScriptFromURLQuirk`, as there is no script to run in this case. 

The `handleScriptToEvaluateBeforeRunningScriptFromURLQuirk` is used only when doing a script injection. Here the code needs to enable the Focusability only when the domain matches `thesaurus.com`

Source/WebCore/page/Quirks.cpp
swpatters: Good call. Thank you.

Source/WebCore/page/Quirks.cpp
swpatters: Thank you for pointing that out. Fixing.

Source/WebCore/page/Quirks.cpp
```diff
@@ -3852,6 +3884,9 @@ void Quirks::determineRelevantQuirks()
 #if PLATFORM(IOS_FAMILY)
         { "theguardian"_s, &handleGuardianQuirks },
         { "thesaurus"_s, &handleScriptToEvaluateBeforeRunningScriptFromURLQuirk },
```
pxlcoder: We can now remove the `handleScriptToEvaluateBeforeRunningScriptFromURLQuirk` entry for both thesaurus.com and dictionary.com, remove them `handleScriptToEvaluateBeforeRunningScriptFromURLQuirk` itself and update the newly added methods to include `quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk);`

swpatters: Ah, yes. Thank you again.

waterplea:
@othermaciej by "everywhere" I mean every OS and every browser in the world, except Safari. If you adopt this behavior it wouldn't even change anything visually for the person, I don't understand what is the big deal here, aside from maybe accessibility tools. But then again, if the whole world works this way + Safari can also work this way if you manually add tabindex="0" to everything – then I'm pretty sure a11y will also be alright.

> Examples of sites that are dependent on links or buttons getting click-focus would help!

Here's a very common case where Safari makes web platform jump through notorious hoops:

image


When we navigate using keyboard and we hit tab after "Button" – we get "Dropdown" focused. Then we hit enter or arrow down which opens the dropdown. Then if we hit tab again, focus moves to "Another button" and dropdown must be closed, because focus is no longer ON its host and neither it is INSIDE dropdown content. So to know when to close the dropdown we must watch focus. Now imagine after we opened the dropdown we click on a radio button inside the dropdown. In Safari focus goes nowhere and we must figure out ourselves with overcomplicated machinery whether this focus loss warrants dropdown closure or if it's just Safari acting up.

Source/WebCore/html/HTMLAnchorElement.cpp
```diff
@@ -110,7 +111,7 @@ bool HTMLAnchorElement::isMouseFocusable() const
 {
 #if !(PLATFORM(GTK) || PLATFORM(WPE))
     // Only allow links with tabIndex or contentEditable to be mouse focusable.
-    if (isLink())
+    if (isLink() && !protect(document())->quirks().needsAnchorToBeMouseFocusable())
```
brentfulgham, line 114: It's a shame we have to ref document every time `isMouseFocusable()` gets called. It would be nice if this could be cached so that it only fired the first time (since it is true for the life of the WCP). But probably only worth doing if it shows up in a perf trace.

brentfulgham (approved):
Looks good. Thank you for resolving this! r=me

Source/WebCore/page/Quirks.cpp
```diff
@@ -2870,6 +2873,30 @@ static void handleWPDevelopmentQuirks(QuirksData& quirksData, const URL& /* quir
 }
 #endif
 
+static void handleThesaurusQuirks(QuirksData& quirksData, const URL& /* quirksURL */, const String& quirksDomainString, const URL&  /* documentURL */)
+{
+    QUIRKS_EARLY_RETURN_IF_NOT_DOMAIN("thesaurus.com"_s);
+
+    quirksData.isThesaurus = true;
+    quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk);
+#if PLATFORM(MAC)
```
aprotyas: I think the platform guards are too restrictive. You can totally hit this same case on iPad desktop-class browsing

aprotyas: We probably want to `PLATFORM(COCOA)` instead.

swpatters: Good call.

aprotyas (commented):
The quirk makes sense to me modulo having it be effective on other Cocoa platforms too.

Can you add test coverage? We have an established pattern for testing quirk behavior using `window.internals.setTopDocumentURLForQuirks()`. See LayoutTests/fast/events/mac/mouse-down-can-start-selection-quirk.html (Spotify) or LayoutTests/fast/css/expedia-group-animation-quirk.html (Expedia) for examples.

I'm imagining something like this:

```js
  // <a id="link" href="#">Click me</a>

  window.internals.setTopDocumentURLForQuirks("https://www.thesaurus.com");
  eventSender.mouseMoveTo({ anchor coordinates });
  eventSender.mouseDown();
  eventSender.mouseUp();
  shouldBe("document.activeElement", "link");
```

Source/WebCore/page/Quirks.cpp
brentfulgham: Yes, let's please make this PLATFORM(COCOA).

Source/WebCore/html/HTMLAnchorElement.cpp
swpatters: 100 anchors x 3 iterations                                                                                                          
With quirk (thesaurus.com):  861, 834, 1258 ms   avg=984 ms
Without quirk (example.com): 832, 1173, 857 ms   avg=954 ms                                                                         
																																  
No meaningful difference — ~984ms vs ~954ms, well within the noise (note the 1258ms and 1173ms outliers swapping between runs). The protect(document()) overhead is not measurable here.

LayoutTests/fast/events/anchor-mouse-focusable-quirk.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
+<!-- Test for the anchor mouse-focusability quirk on thesaurus.com and dictionary.com.
+     https://bugs.webkit.org/show_bug.cgi?id=312692 rdar://174959285
```
aprotyas: I would either link to both tracking bugs or neither. (I'm leaning towards the latter)

LayoutTests/fast/events/anchor-mouse-focusable-quirk.html
```diff
@@ -0,0 +1,43 @@
...
+<html>
+<head>
+<script src="../../resources/js-test.js"></script>
+</head>
+<body>
+<a id="link" href="#">Click me</a>
+<script>
+description("Tests that the anchor mouse-focusability quirk works on thesaurus.com and dictionary.com, but not on unrelated sites.");
+
+var link = document.getElementById("link");
```
aprotyas, line 14:
```suggestion
let link = document.getElementById("link");
```

LayoutTests/fast/events/anchor-mouse-focusable-quirk.html
```diff
@@ -0,0 +1,43 @@
...
+
+function clickLink() {
+    var rect = link.getBoundingClientRect();
```
aprotyas:
```suggestion
    let rect = link.getBoundingClientRect();
```

Source/WebCore/html/HTMLAnchorElement.cpp
swpatters: But I'll keep an eye out. Thank you.

aprotyas: @swpatters what is this measurement?

aprotyas: I think Brent was referring to more systematic tracing with our perf infra.

Source/WebCore/page/Quirks.cpp
```diff
@@ -247,6 +247,19 @@ bool Quirks::isEmbedDomain(const String& domainString) const
     return RegistrableDomain(document->url()).string() == domainString;
 }
 
+// thesaurus.com https://bugs.webkit.org/show_bug.cgi?id=312692 rdar://174959285
+// dictionary.com https://bugs.webkit.org/show_bug.cgi?id=312692 rdar://174959285
```
aprotyas: These both point to the thesaurus.com bug, I would adjust the bug trackers for dictionary.com

Source/WebCore/page/Quirks.cpp
```diff
@@ -2870,6 +2873,30 @@ static void handleWPDevelopmentQuirks(QuirksData& quirksData, const URL& /* quir
 }
 #endif
 
+static void handleThesaurusQuirks(QuirksData& quirksData, const URL& /* quirksURL */, const String& quirksDomainString, const URL&  /* documentURL */)
+{
+    QUIRKS_EARLY_RETURN_IF_NOT_DOMAIN("thesaurus.com"_s);
+
+    quirksData.isThesaurus = true;
+    quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk);
```
aprotyas, line 2904: Shouldn't these quirks only be enabled for `#if ENABLE(DESKTOP_CONTENT_MODE_QUIRKS)` configs?

Source/WebCore/page/Quirks.cpp
```diff
@@ -2870,6 +2873,30 @@ static void handleWPDevelopmentQuirks(QuirksData& quirksData, const URL& /* quir
...
+    quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsAnchorToBeMouseFocusableQuirk);
+#endif
+}
+
+static void handleDictionaryQuirks(QuirksData& quirksData, const URL& /* quirksURL */, const String& quirksDomainString, const URL&  /* documentURL */)
+{
+    QUIRKS_EARLY_RETURN_IF_NOT_DOMAIN("dictionary.com"_s);
+
+    quirksData.isDictionary = true;
+    quirksData.enableQuirk(QuirksData::SiteSpecificQuirk::NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk);
```
aprotyas, line 2917: Same comment here about `DESKTOP_CONTENT_MODE_QUIRKS`

aprotyas (approved):
Thanks for iterating. Two action items for using the correct bug numbers and considering the scope of the `NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk` quirk.

Source/WebCore/page/Quirks.cpp
aprotyas: (previously, these were only enabled in the `#if PLATFORM(IOS_FAMILY) && ENABLE(DESKTOP_CONTENT_MODE_QUIRKS)` path.)

LayoutTests/fast/events/anchor-mouse-focusable-quirk.html
swpatters: I see. Removed.

LayoutTests/fast/events/anchor-mouse-focusable-quirk.html
swpatters: var link kept as var — needed because shouldBe() uses eval() which can't access let variables from a different script scope.

Source/WebCore/page/Quirks.cpp
swpatters: I wrapped it with #if PLATFORM(IOS_FAMILY) to match the original behavior — and that the original code was only IOS_FAMILY-guarded (not DESKTOP_CONTENT_MODE_QUIRKS), see lines 3775-3776 and 3853-3854.

Source/WebCore/page/Quirks.cpp
swpatters: Similar issue as above.

Source/WebCore/html/HTMLAnchorElement.cpp
swpatters: I ran some local perf tests to see if could see anything ahead of infrastructure. But yes, Brent's concern is still valid and we'll wait and see.

Source/WebCore/page/Quirks.cpp
aprotyas: Sounds good, I don't feel strongly about this in either direction. There's a latent risk that somewhere else in newer code we may end up reading `SiteSpecificQuirk::NeedsScriptToEvaluateBeforeRunningScriptFromURLQuirk` as enabled on macOS, but not a big deal.

PR #63096, opened 2026-04-20
Speed up IDL bindings generation build step

cdumez:
#### 4ed26b82fdb27bf273ff90a7026a46f958ce193a
<pre>
Speed up IDL bindings generation build step
<a href="https://bugs.webkit.org/show_bug.cgi?id=312747">https://bugs.webkit.org/show_bug.cgi?id=312747</a>

Reviewed by Sam Weinig.

This PR applies the following optimization to the IDL bindings generation
build step:

applyPreprocessor() unconditionally forks clang to preprocess every
file, even though most files don&apos;t contain C preprocessor directives.
For those files, reading and stripping comments in Perl avoids the
expensive fork+exec of clang/gcc (~25ms per invocation).

For the files that do use directives, they all follow simple
(including nesting), so evaluate these in Perl as well, eliminating all
clang invocations during the build step. This covers both IDL files
(25 out of ~1,823 use directives) and user agent CSS stylesheets.

Note: // line comments are only stripped for .idl files, not for CSS
files where // appears in URLs and is not a comment delimiter.

This reduces the generate-bindings-all.pl build step from 1 minute 36 seconds
to 21 seconds, on a M4 Max machine.

Additionally, remove the --preprocessor command-line option from all scripts
and the CODE_GENERATOR_PREPROCESSOR CMake variable. The CMake build was passing
an explicit preprocessor path (e.g. clang -E -P -x c++) that did exactly the
same thing the Xcode build was doing implicitly via a fallback default in
applyPreprocessor(). Since the Perl-based preprocessor now handles all cases,
there is no longer any need for an external preprocessor codepath.

* Source/WebCore/bindings/scripts/preprocessor.pm:
(applyPreprocessor): Handle preprocessing entirely in Perl: read the file,
strip comments (/* */ for all files, // only for .idl files), and if the
file contains preprocessor directives, evaluate them using
_evaluatePreprocessorDirectives instead of forking clang. Remove the
$preprocessor parameter and the _applyExternalPreprocessor function.
(_evaluatePreprocessorDirectives): Added. #if / #else / #endif evaluator
using a stack to track nesting.
(_evaluateCondition): Added. Evaluates preprocessor condition expressions
supporting conjunctions (&amp;&amp;), defined() checks, negation (!(...)), and
parenthesized sub-expressions.
* Source/WebCore/bindings/scripts/CodeGenerator.pm:
(new): Remove $preprocessor parameter.
* Source/WebCore/bindings/scripts/IDLParser.pm:
(Parse): Remove $preprocessor parameter.
* Source/WebCore/bindings/scripts/generate-bindings.pl: Remove --preprocessor option.
* Source/WebCore/bindings/scripts/generate-bindings-all.pl: Remove --preprocessor option.
* Source/WebCore/bindings/scripts/preprocess-idls.pl: Remove --preprocessor option.
* Source/WebCore/css/make-css-file-arrays.pl: Remove --preprocessor option.
* Source/WebCore/preprocess-localizable-strings.pl: Remove --preprocessor option.
* Source/WebCore/CMakeLists.txt: Remove --preprocessor from make-css-file-arrays.pl invocation.
* Source/WebCore/WebCoreMacros.cmake: Remove --preprocessor from bindings generation args.
* Source/WebKit/PlatformGTK.cmake: Remove --preprocessor from make-css-file-arrays.pl invocation.
* Source/WebKit/PlatformWPE.cmake: Remove --preprocessor from make-css-file-arrays.pl invocation.
* Source/cmake/WebKitCompilerFlags.cmake: Remove CODE_GENERATOR_PREPROCESSOR variable.

Canonical link: <a href="https://commits.webkit.org/311901@main">https://commits.webkit.org/311901@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/1389aaae27a4928ccd2f6506746f7bea8a428621

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/158159 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/31496 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/24690 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/166988 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/112242 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/160030 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/31633 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/31514 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/122485 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85983 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/161117 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/24756 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/142051 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/103154 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/23812 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/22163 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/14760 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/150210 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/133496 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/19843 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/169477 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/18994 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/14831 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/21466 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/130666 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/31242 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/26227 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/130781 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35450 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/31180 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/141637 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/89076 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/25491 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/18443 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/190288 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/30732 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/96265 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/48840 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/30253 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/30483 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/30380 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/bindings/scripts/preprocessor.pm
+++ b/Source/WebCore/bindings/scripts/preprocessor.pm
@@ -21,11 +21,6 @@
 use strict;
 use warnings;
 
-use Config;
-use IPC::Open2;
-use IPC::Open3;
-use Text::ParseWords;
-
 BEGIN {
    use Exporter   ();
    our ($VERSION, @ISA, @EXPORT, @EXPORT_OK, %EXPORT_TAGS);
@@ -36,84 +31,106 @@ BEGIN {
    @EXPORT_OK   = ();
 }
 
+# Reads a file, strips comments, and evaluates simple #if / #else / #endif
+# preprocessor directives without forking an external process.
 # Returns an array of lines.
 sub applyPreprocessor
 {
     my $fileName = shift;
     my $defines = shift;
-    my $preprocessor = shift;
     my $keepComments = shift;
+    my $stripLineComments = shift;
 
-    my @args = ();
-    if (!$preprocessor) {
-        if ($Config::Config{"osname"} eq "MSWin32") {
-            $preprocessor = $ENV{CC} || "cl";
-            push(@args, qw(/nologo /EP /TP));
-        } else {
-            $preprocessor = $ENV{CC} || (-x "/usr/bin/clang" ? "/usr/bin/clang" : "/usr/bin/gcc");
-            push(@args, qw(-E -P -x c++));
-            if ($keepComments) {
-                push(@args, qw(-C));
-            }
-        }
+    open(my $fh, '<', $fileName) or die "Failed to open $fileName";
+    local $/;
+    my $content = <$fh>;
+    close($fh);
+
+    unless ($keepComments) {
+        # Strip block comments.
+        $content =~ s|/\*.*?\*/||gs;
+        # Strip // line comments only when requested. CSS files use // in URLs
+        # so callers processing CSS must not pass $stripLineComments.
+        $content =~ s|//[^\n]*||g if $stripLineComments;
     }
 
-    if ($Config::Config{"osname"} eq "darwin") {
-        if ($ENV{ARCHS}) {
-            (my $arch, @_) = split(" ", $ENV{ARCHS});
-            my $vendor = $ENV{LLVM_TARGET_TRIPLE_VENDOR};
-            my $os = $ENV{LLVM_TARGET_TRIPLE_OS_VERSION} . ($ENV{LLVM_TARGET_TRIPLE_SUFFIX} // "");
-            push(@args, "-target", "$arch-$vendor-$os");
-        }
-        # When ARCHS is unset (CMake builds), clang defaults to the host
-        # triple, which is correct for preprocessing.
-        push(@args, "-I" . $ENV{BUILT_PRODUCTS_DIR} . "/usr/local/include") if $ENV{BUILT_PRODUCTS_DIR};
-        push(@args, "-isysroot", $ENV{SDKROOT}) if $ENV{SDKROOT};
+    if ($content !~ /^\s*#/m) {
+        return split(/^/m, $content);
     }
 
-    my @macros;
+    # The file has preprocessor directives. Evaluate simple
+    # #if defined(X) && X / #if !(defined(X) && X) / #else / #endif
+    # patterns in Perl instead of forking clang.
+    my %macros;
     if ($defines) {
-        # Remove double quotations from $defines and extract macros.
-        # For example, if $defines is ' "A=1" "B=1" C=1 ""    D  ',
-        # then it is converted into four macros -DA=1, -DB=1, -DC=1 and -DD.
-        $defines =~ s/\"//g;
-        @macros = grep { $_ } split(/\s+/, $defines); # grep skips empty macros.
-        @macros = map { "-D$_" } @macros;
+        $macros{$_} = 1 for grep { $_ } split(/\s+/, $defines =~ s/"//gr);
+    }
+    return split(/^/m, _evaluatePreprocessorDirectives($content, \%macros));
+}
+
+# Simple #if / #else / #endif evaluator. Supports:
+#   #if defined(X) && X
+#   #if defined(X)
+#   #if defined(X) && X && defined(Y) && Y  (conjunctions)
+#   #if !(defined(X) && X)                  (negation)
+#   #else
+#   #endif
+#   #endif // comment
+#   Nesting
+sub _evaluatePreprocessorDirectives
+{
+    my ($content, $macros) = @_;
+    my $output = "";
+    # Stack of [emittingBeforeThisIf, conditionWasTrue].
+    my @ifStack = ();
+    my $emitting = 1;
+
+    for my $line (split(/^/m, $content)) {
+        if ($line =~ /^\s*#\s*if\s+(.*)/) {
+            my $cond = _evaluateCondition($1, $macros);
+            push @ifStack, [$emitting, $cond];
+            $emitting = $emitting && $cond;
+        } elsif ($line =~ /^\s*#\s*else\b/) {
+            my $frame = $ifStack[-1];
+            $emitting = $frame->[0] && !$frame->[1];
+        } elsif ($line =~ /^\s*#\s*endif\b/) {
+            my $frame = pop @ifStack;
+            $emitting = $frame->[0];
+        } else {
+            $output .= $line if $emitting;
+        }
+    }
+    return $output;
+}
+
+sub _evaluateCondition
+{
+    my ($expr, $macros) = @_;
+
+    # Handle negation: "!(defined(X) && X)"
+    if ($expr =~ /^\s*!\s*\((.+)\)\s*$/) {
+        return _evaluateCondition($1, $macros) ? 0 : 1;
     }
 
-    my $pid = 0;
-    if ($Config{osname} eq "cygwin") {
-        $ENV{PATH} = "$ENV{PATH}:/cygdrive/c/cygwin/bin";
-        my @preprocessorAndFlags = shellwords($preprocessor);
-        if ($preprocessorAndFlags[0] =~ "cl.exe") {
-            $fileName = Cygwin::posix_to_win_path($fileName);
-        }
-        $preprocessorAndFlags[0] = Cygwin::win_to_posix_path($preprocessorAndFlags[0]);
-        # This call can fail if Windows rebases cygwin, so retry a few times until it succeeds.
-        for (my $tries = 0; !$pid && ($tries < 20); $tries++) {
-            eval {
-                # Suppress STDERR so that if we're using cl.exe, the output
-                # name isn't needlessly echoed.
-                use Symbol 'gensym'; my $err = gensym;
-                $pid = open3(\*PP_IN, \*PP_OUT, $err, @preprocessorAndFlags, @args, @macros, $fileName);
-                1;
-            } or do {
-                sleep 1;
-            }
-        };
-    } elsif ($Config::Config{"osname"} eq "MSWin32") {
-        # Suppress STDERR so that if we're using cl.exe, the output
-        # name isn't needlessly echoed.
-        use Symbol 'gensym'; my $err = gensym;
-        $pid = open3(\*PP_IN, \*PP_OUT, $err, $preprocessor, @args, @macros, $fileName);
-    } else {
-        $pid = open2(\*PP_OUT, \*PP_IN, shellwords($preprocessor), @args, @macros, $fileName);
+    # Handle parenthesized expression: "(defined(X) && X)"
+    if ($expr =~ /^\s*\((.+)\)\s*$/) {
+        return _evaluateCondition($1, $macros);
     }
-    close PP_IN;
-    my @documentContent = <PP_OUT>;
-    close PP_OUT;
-    waitpid($pid, 0);
-    return @documentContent;
+
+    # Handle conjunctions: "defined(X) && X && defined(Y) && Y"
+    my @parts = split(/\s*&&\s*/, $expr);
+    for my $part (@parts) {
+        $part =~ s/^\s+|\s+$//g;
+        if ($part =~ /^\(?defined\((\w+)\)\)?$/) {
+            return 0 unless $macros->{$1};
+        } elsif ($part =~ /^(\w+)$/) {
+            return 0 unless $macros->{$1};
+        } else {
+            # Unrecognized expression — conservatively treat as true.
+            next;
+        }
+    }
+    return 1;
 }
 
 1;
--- a/Source/WebCore/CMakeLists.txt
+++ b/Source/WebCore/CMakeLists.txt
@@ -2864,7 +2864,7 @@ add_custom_command(
     OUTPUT ${WebCore_DERIVED_SOURCES_DIR}/UserAgentStyleSheetsData.cpp ${WebCore_DERIVED_SOURCES_DIR}/UserAgentStyleSheets.h
     MAIN_DEPENDENCY ${WEBCORE_DIR}/css/make-css-file-arrays.pl
     DEPENDS ${WebCore_USER_AGENT_STYLE_SHEETS} ${WEBCORE_DIR}/bindings/scripts/preprocessor.pm
-    COMMAND ${PERL_EXECUTABLE} ${WEBCORE_DIR}/css/make-css-file-arrays.pl --defines "${FEATURE_DEFINES_WITH_SPACE_SEPARATOR}" --preprocessor "${CODE_GENERATOR_PREPROCESSOR}" ${WebCore_DERIVED_SOURCES_DIR}/UserAgentStyleSheets.h ${WebCore_DERIVED_SOURCES_DIR}/UserAgentStyleSheetsData.cpp ${WebCore_USER_AGENT_STYLE_SHEETS}
+    COMMAND ${PERL_EXECUTABLE} ${WEBCORE_DIR}/css/make-css-file-arrays.pl --defines "${FEATURE_DEFINES_WITH_SPACE_SEPARATOR}" ${WebCore_DERIVED_SOURCES_DIR}/UserAgentStyleSheets.h ${WebCore_DERIVED_SOURCES_DIR}/UserAgentStyleSheetsData.cpp ${WebCore_USER_AGENT_STYLE_SHEETS}
     VERBATIM)
 list(APPEND WebCore_SOURCES ${WebCore_DERIVED_SOURCES_DIR}/UserAgentStyleSheetsData.cpp)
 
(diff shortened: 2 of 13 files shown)


weinig:
> Together, these reduce the preprocess-idls.pl + generate-bindings-all.pl
build steps from 1 minute 36 seconds to 6 seconds.

Wow! 

(For completeness, what machine / configuration is that on.)

weinig:
(Going to make my forthcoming python re-write less impressive, but fortunately the goal for the re-write is not performance and is just getting things into a language more contributors are likely to know)

Source/WebCore/bindings/scripts/preprocessor.pm
```diff
@@ -44,6 +44,35 @@ sub applyPreprocessor
...
+        open(my $fh, '<', $fileName) or die "Failed to open $fileName";
+        local $/;
+        my $content = <$fh>;
+        close($fh);
+
+        # Strip block comments like the preprocessor would.
+        $content =~ s|/\*.*?\*/||gs;
+        # Strip // line comments only for IDL files. Do NOT strip them for
+        # CSS files where // appears in URLs and is not a comment delimiter.
+        $content =~ s|//[^\n]*||g if $fileName =~ /\.idl$/;
```
weinig: Hard coding .idl here seems weird. This seems like something that should be a parameter to the function.

(Also, if this is used for non-bindings things, we should probably move this perl module to somewhere other than WebCore/bindings/scripts, perhaps a new WebCore/scripts directory for shared functionality).

Source/WebCore/bindings/scripts/preprocessor.pm
```diff
@@ -44,6 +44,35 @@ sub applyPreprocessor
...
+        # Perl instead of forking clang.
+        my %macros;
+        if ($defines) {
+            $macros{$_} = 1 for grep { $_ } split(/\s+/, $defines =~ s/"//gr);
+        }
+        return split(/^/m, _evaluatePreprocessorDirectives($content, \%macros));
+    }
+
     my @args = ();
     if (!$preprocessor) {
```
weinig: Will this ever be true now given the above `if (!$preprocessor) {` always returns?

Source/WebCore/bindings/scripts/preprocessor.pm
```diff
@@ -44,6 +44,35 @@ sub applyPreprocessor
     my $preprocessor = shift;
     my $keepComments = shift;
 
+    # Fast path: if no explicit preprocessor was requested, check whether the
```
weinig: Seems like we should split this into two functions, one that wants to support the full preprocessor set of functionality and one that just wants the simple version. That also solves the problem below where we have the a second dead `if (!$preprocessor) {` check.

Source/WebCore/DerivedSources.make
```diff
@@ -2688,10 +2688,12 @@ $(IDL_INTERMEDIATE_PATTERNS) : $(PREPROCESS_IDLS_SCRIPTS) $(IDL_ATTRIBUTES_FILE)
 vpath %.idl $(ADDITIONAL_BINDING_IDLS_PATHS) $(WebCore)/bindings/scripts
 
 # -------------------------------------------------
-JS_DOM_HEADERS_PATTERNS = $(subst .h,%h,$(JS_DOM_HEADERS))
-JS_DOM_IMPLEMENTATIONS_PATTERNS = $(subst .cpp,%cpp,$(JS_DOM_IMPLEMENTATIONS))
+# Use a stamp file so that make checks dependencies only once for all JS
+# bindings, instead of evaluating ~3,646 pattern targets individually against
+# ~1,823 IDL prerequisites (which causes O(n*m) stat calls and ~15s overhead).
+JS_BINDINGS_STAMP = .js-bindings-stamp
```
weinig, line 2694: In the interest of bisection (and determining individual perf results), can we split this timestamp part of the change into its own PR?

weinig (changes requested):
If we never pass a preprocessor to generate-bindings.pl or generate-bindings-all.pl (which it would be great to add to WebCore.xcodeproj so we can see it in Xcode), we should remove the `preprocessor` option from the GetOptions() calls.

hortont424:
@emw-apple made me revert the stamp file change from my patch to adopt the combined bindings generator script, but I also didn't have numbers like this. Anyway, check with them in case there was a reason.

weinig:
I think someone more versed in makefiles would be better suited to review the makefile part of this. It's been long enough since I dove into that part that I don't have a good "big picture" understanding of the build process for the bindings anymore. From a quick peek at the code to refresh me, my understanding is that works something like this:

- First, the `preprocess-idls.pl` (which is a completely different usage of "preprocess" that has nothing to do with macros or the c++ preprocessor so should probably be named differently) script is called, with all IDL files passed to it. It generates some additional IDL files (like the FooConstructors.idl files) and generates the makefile dependency data in the form of SupplementalDependencies.dep. By default, it doesn't do a real parse of the IDL files, but rather uses regular expressions to try and extract the information it needs (I added the ability to validate it against the real IDL parser, and this hack annoys me to this day).

- Second, the `generate-bindings-all.pl` script is called, also with all the IDL files passed to it, but also with the supplemental dependency information generated by `preprocess-idls.pl` in the first step, and it decides which JSFoo.cpp/JSFoo.h files need to be generated, using the `generate-bindings.pl` script to do that. It also looks like it can call `preprocess-idls.pl`, so I am not sure why DerivedSources.make also needs/wants to do that.

(Other uses of the IDL infrastructure, like DumpRenderTree, WebKitTestRunner and WebKit's WebExtensions, use `generate-bindings.pl` directly, instead of using `generate-bindings-all.pl`. It seems like converging on one or the other would be a good simplification).

Since you are touching a bunch of different parts of this, it would be great if you could jot down some documentation on this, as its pretty tricky and getting things wrong can cause subtle but very annoying problems like failing to build locally because some file is not regenerating.

Source/WebCore/bindings/scripts/preprocessor.pm
emw-apple: Might be easier to solve this by anchoring the search to the start of each line? I think that's what `/m` is for.


```suggestion
        $content =~ s|^\s*//[^\n]*||mg if $fileName =~ /\.idl$/;
```

Source/WebCore/DerivedSources.make
```diff
@@ -2688,10 +2688,12 @@ $(IDL_INTERMEDIATE_PATTERNS) : $(PREPROCESS_IDLS_SCRIPTS) $(IDL_ATTRIBUTES_FILE)
...
 
-$(JS_DOM_HEADERS_PATTERNS) $(JS_DOM_IMPLEMENTATIONS_PATTERNS): $(JS_BINDING_IDLS) $(JS_BINDINGS_SCRIPTS) \
+$(JS_BINDINGS_STAMP): $(JS_BINDING_IDLS) $(JS_BINDINGS_SCRIPTS) \
```
emw-apple: Changing from using pattern rules to a single target means that we can't express when we only need to generate a subset of the overall bindings (i.e. when one IDL changes and we rebuild). But that seems fine here, since iirc the perl code performs its own stat checks, and we aren't using any variables like `$?`.

cdumez:
> > Together, these reduce the preprocess-idls.pl + generate-bindings-all.pl
> > build steps from 1 minute 36 seconds to 6 seconds.
> 
> Wow!
> 
> (For completeness, what machine / configuration is that on.)

I am on a M4 Max machine.

Source/WebCore/DerivedSources.make
cdumez: Yes, I can split them. The Makefile change is a bit riskier and saves less than 15 seconds.
The preprocessor fix saves 1 minute and 15 seconds.

cdumez: I created https://github.com/WebKit/WebKit/pull/63181 for the DerivedSources.make changes.

Source/WebCore/bindings/scripts/preprocessor.pm
cdumez: @emw-apple No, I don't think that would work. IDL files have `//` comments at the end of code lines, not just on their own lines. For example:
```
  FetchRequestCredentials credentials = "same-origin"; // credentials is only used if type is "module"                                                           
```

The anchored regex `^\s*//[^\n]*` would only strip full-line comments and miss these end-of-line comments. The IDL parser expects all `//` comments to be stripped (as clang would do previously), so leaving them in would break parsing.

Source/WebCore/bindings/scripts/preprocessor.pm
cdumez: No, this is dead code. I'll remove.

cdumez:
> (Going to make my forthcoming python re-write less impressive, but fortunately the goal for the re-write is not performance and is just getting things into a language more contributors are likely to know)

Well, Claude knows Perl just as well as Python so I'm not sure it makes a lot of difference anymore :P

cdumez:
> If we never pass a preprocessor to generate-bindings.pl or generate-bindings-all.pl (which it would be great to add to WebCore.xcodeproj so we can see it in Xcode), we should remove the `preprocessor` option from the GetOptions() calls.

The CMake build DOES pass --preprocessor, so the explicit preprocessor path is still reachable on CMake builds.

weinig:
> > If we never pass a preprocessor to generate-bindings.pl or generate-bindings-all.pl (which it would be great to add to WebCore.xcodeproj so we can see it in Xcode), we should remove the `preprocessor` option from the GetOptions() calls.
> 
> The CMake build DOES pass --preprocessor, so the explicit preprocessor path is still reachable on CMake builds.

But do any IDLs rely on it? If not, it should be removed. If so, we shouldn’t make this change at all.

cdumez:
> > > If we never pass a preprocessor to generate-bindings.pl or generate-bindings-all.pl (which it would be great to add to WebCore.xcodeproj so we can see it in Xcode), we should remove the `preprocessor` option from the GetOptions() calls.
> > 
> > 
> > The CMake build DOES pass --preprocessor, so the explicit preprocessor path is still reachable on CMake builds.
> 
> But do any IDLs rely on it? If not, it should be removed. If so, we shouldn’t make this change at all.

I don't understand your feedback.

IDL files do rely on preprocessor directives. There are 26 IDL files (out of ~1,823) that use `#if defined(ENABLE_*) && ENABLE_* / #else / #endif` guards to conditionally include/exclude API members. For example, HTMLMediaElement.idl has:                                                                             
```
  #if defined(ENABLE_MEDIA_STREAM) && ENABLE_MEDIA_STREAM                                                                                                        
      attribute MediaStream srcObject;                                                                                                                           
  #endif                                                                                                                                                         
```                                                                                                                                  
So the preprocessing step cannot be removed, it's needed on all build systems.                                                                                
                                                                                                                                                                 
What this patch does is avoid forking clang for that preprocessing, not getting rid of preprocessing.

The Xcode build never passed `--preprocessor`. When no `--preprocessor` was given, `applyPreprocessor()` would default to `/usr/bin/clang` (or $CC) and fork it with `-E -P -x c++` for *every single IDL file*. This patch replaces that implicit clang fork with a fast in-Perl implementation that reads each file directly, strips comments, and evaluates the simple `#if / #else / #endif` patterns itself, cutting the build step from 1m36s to 21s. The explicit `--preprocessor` codepath is preserved for CMake builds that pass it.

cdumez:
> > > > If we never pass a preprocessor to generate-bindings.pl or generate-bindings-all.pl (which it would be great to add to WebCore.xcodeproj so we can see it in Xcode), we should remove the `preprocessor` option from the GetOptions() calls.
> > > 
> > > 
> > > The CMake build DOES pass --preprocessor, so the explicit preprocessor path is still reachable on CMake builds.
> > 
> > 
> > But do any IDLs rely on it? If not, it should be removed. If so, we shouldn’t make this change at all.
> 
> I don't understand your feedback.
> 
> IDL files do rely on preprocessor directives. There are 26 IDL files (out of ~1,823) that use `#if defined(ENABLE_*) && ENABLE_* / #else / #endif` guards to conditionally include/exclude API members. For example, HTMLMediaElement.idl has:
> 
> ```
>   #if defined(ENABLE_MEDIA_STREAM) && ENABLE_MEDIA_STREAM                                                                                                        
>       attribute MediaStream srcObject;                                                                                                                           
>   #endif                                                                                                                                                         
> ```
> 
> So the preprocessing step cannot be removed, it's needed on all build systems.
> 
> What this patch does is avoid forking clang for that preprocessing, not getting rid of preprocessing.
> 
> The Xcode build never passed `--preprocessor`. When no `--preprocessor` was given, `applyPreprocessor()` would default to `/usr/bin/clang` (or $CC) and fork it with `-E -P -x c++` for _every single IDL file_. This patch replaces that implicit clang fork with a fast in-Perl implementation that reads each file directly, strips comments, and evaluates the simple `#if / #else / #endif` patterns itself, cutting the build step from 1m36s to 21s. The explicit `--preprocessor` codepath is preserved for CMake builds that pass it.

@weinig Does this clarify things?

weinig:
I’m not sure it does. What is the reason the CMake build needs to pass the flag when the Xcode build does not?

cdumez:
> I’m not sure it does. What is the reason the CMake build needs to pass the flag when the Xcode build does not?

Ok, I'll update CMake to not pass the flag.

cdumez:
> > I’m not sure it does. What is the reason the CMake build needs to pass the flag when the Xcode build does not?
> 
> Ok, I'll update CMake to not pass the flag.

Ok, I have updated the PR. The flag is gone and now all ports use the fast path, not just Cocoa ones.

weinig (approved):

PR #17228, opened 2023-08-30
Optimize StyledElement::rebuildPresentationalHintStyle()

cdumez:
#### 8249ec2a7b199f145256e194ad2fc97c61dbe404
<pre>
Optimize StyledElement::rebuildPresentationalHintStyle()
<a href="https://bugs.webkit.org/show_bug.cgi?id=258316">https://bugs.webkit.org/show_bug.cgi?id=258316</a>
rdar://111406451

Reviewed by Darin Adler.

Optimize StyledElement::rebuildPresentationalHintStyle() by leveraging
the NodeName enumeration to avoid a large HashMap lookup.

* Source/WebCore/dom/StyledElement.cpp:
(WebCore::StyledElement::rebuildPresentationalHintStyle):
* Source/WebCore/dom/StyledElement.h:
(WebCore::StyledElement::additionalPresentationalHintStyle const):
(WebCore::StyledElement::collectExtraStyleForPresentationalHints): Deleted.
* Source/WebCore/html/HTMLImageElement.h:
* Source/WebCore/svg/SVGElement.cpp:
(WebCore::SVGElement::cssPropertyIdForSVGAttributeName):
(WebCore::createAttributeNameToCSSPropertyIDMap): Deleted.

Canonical link: <a href="https://commits.webkit.org/267482@main">https://commits.webkit.org/267482@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/58a49db053dfd69fff4cc040d1fd009a10244875

| Misc | iOS, tvOS & watchOS  | macOS  | Linux |  Windows |
| ----- | ---------------------- | ------- |  ----- |  --------- |
| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/16614 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/26/builds/16937 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/14/builds/17383 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/18396 "Built successfully") | [✅ 🛠 wincairo](https://ews-build.webkit.org/#/builders/32/builds/15582 "Built successfully") 
[...]

PR #63207, opened 2026-04-21
CSS transform none value doesn't override transform attribute on SVG element

karlcow:
#### 79ef02060e796b0ce7145d2f3dadedb3f92d93be
<pre>
CSS transform none value doesn&apos;t override transform attribute on SVG element
<a href="https://bugs.webkit.org/show_bug.cgi?id=217286">https://bugs.webkit.org/show_bug.cgi?id=217286</a>
<a href="https://rdar.apple.com/69934881">rdar://69934881</a>

Reviewed by NOBODY (OOPS!).

Map the SVG transform attribute into the CSS cascade as a presentation
hint so that CSS transform declarations (including none, initial, inherit)
correctly override it.

This is JUST A DRAFT PR. Please ignore for now.

Tests: imported/w3c/web-platform-tests/svg/styling/transform-none-overrides-attribute-ref.html
       imported/w3c/web-platform-tests/svg/styling/transform-none-overrides-attribute.html
       imported/w3c/web-platform-tests/svg/styling/transform-attribute-applies-without-css-ref.html
       imported/w3c/web-platform-tests/svg/styling/transform-attribute-applies-without-css.html
       imported/w3c/web-platform-tests/svg/styling/transform-attribute-computed-style.html
       imported/w3c/web-platform-tests/svg/styling/transform-css-value-overrides-attribute-ref.html
       imported/w3c/web-platform-tests/svg/styling/transform-css-value-overrides-attribute.html
       imported/w3c/web-platform-tests/svg/styling/transform-inherit-overrides-attribute-ref.html
       imported/w3c/web-platform-tests/svg/styling/transform-inherit-overrides-attribute.html
       imported/w3c/web-platform-tests/svg/styling/transform-initial-overrides-attribute-ref.html
       imported/w3c/web-platform-tests/svg/styling/transform-initial-overrides-attribute.html

* LayoutTests/imported/w3c/web-platform-tests/svg/styling/presentation-attributes-relevant-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/presentation-attributes-special-cases-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/symbol-g-graphics-element-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-attribute-applies-without-css-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-attribute-applies-without-css-ref.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-attribute-applies-without-css.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-attribute-computed-style-expected.txt: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-attribute-computed-style.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-css-value-overrides-attribute-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-css-value-overrides-attribute-ref.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-css-value-overrides-attribute.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-inherit-overrides-attribute-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-inherit-overrides-attribute-ref.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-inherit-overrides-attribute.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-initial-overrides-attribute-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-initial-overrides-attribute-ref.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-initial-overrides-attribute.html: Added.
* LayoutTests/platform/mac/imported/w3c/web-platform-tests/svg/styling/css-var-on-length-attributes-02-expected.png: Added.
* LayoutTests/platform/mac/imported/w3c/web-platform-tests/svg/styling/css-var-on-length-attributes-04-expected.png: Added.
* LayoutTests/platform/mac/imported/w3c/web-platform-tests/svg/styling/css-var-on-length-attributes-05-expected.png: Added.
* LayoutTests/svg/repaint/foreign-object-repainting-after-modifying-transform-repaint-rects-expected.txt:
* LayoutTests/svg/repaint/text-repainting-after-modifying-transform-repaint-rects-expected.txt:
* Source/WebCore/rendering/RenderLayerModelObject.cpp:
(WebCore::RenderLayerModelObject::applySVGTransform const):
* Source/WebCore/style/values/transforms/StyleTransformFunction.cpp:
(WebCore::Style::createMatrixTransformFunction):
* Source/WebCore/svg/SVGGraphicsElement.cpp:
(WebCore::SVGGraphicsElement::hasPresentationalHintsForAttribute const):
(WebCore::SVGGraphicsElement::addTransformPresentationalHint):
(WebCore::SVGGraphicsElement::collectPresentationalHintsForAttribute):
(WebCore::SVGGraphicsElement::collectExtraStyleForPresentationalHints):
(WebCore::SVGGraphicsElement::animatedLocalTransform const):
(WebCore::SVGGraphicsElement::svgAttributeChanged):
* Source/WebCore/svg/SVGGraphicsElement.h:
* Source/WebCore/svg/properties/SVGAnimatedPropertyAnimatorImpl.h:
</pre>
https://github.com/WebKit/WebKit/commit/79ef02060e796b0ce7145d2f3dadedb3f92d93be

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/158923 "check-webkit-style (failure)") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/32350 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/25455 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/167752 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/113007 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/fcdf4f27-67dc-41ae-ad6e-5dd34dcae8fa) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/160792 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/32417 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/32337 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/123134 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/86459 "Passed tests") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/64b69921-a167-4f6e-aecf-43f14e2dd6cb) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/161880 "Passed tests") | [❌ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/25403 "Found 1 new test failure: svg/repaint/text-repainting-after-modifying-transform-repaint-rects.html (failure)") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/142770 "Passed tests") | [❌ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/103802 "Found 1 new API test failure: WPE/TestUIClient:/webkit/WebKitWebView/usermedia-enumeratedevices-permission-check (failure)") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/862f55d3-3045-4c34-ba07-70e8712e2fb9) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/24459 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/22861 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/15524 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/134145 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/20550 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/170245 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/15987 "Built successfully") | [❌ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/22176 "1 flakes") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/131322 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/32039 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/26928 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/131435 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/31984 "Built successfully") | [❌ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/142343 "1 flakes") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/90029 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/26139 "Passed tests") | [❌ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/19152 "1 flakes") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/31495 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/97509 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/31015 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/31288 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/31169 "Built successfully") | | | |


karlcow:
The results before and after
Screenshot 2026-04-21 at 17 18 17

using https://codepen.io/webcompat/pen/myrgZvE

From left to right:

* MiniBrowser + patch
* STP 241
* Firefox Nightly 
* Chrome Canary

LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-none-overrides-attribute.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS transform: none overrides SVG transform attribute</title>
+<link rel="help" href="https://www.w3.org/TR/css-transforms-1/#transform-property">
+<link rel="help" href="https://svgwg.org/svg2-draft/styling.html#PresentationAttributes">
```
karlcow:
```suggestion
<link rel="help" href="https://w3c.github.io/svgwg/svg2-draft/styling.html#PresentationAttributes">
```

LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-none-overrides-attribute-expected.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="utf-8">
+<title>CSS transform: none overrides SVG transform attribute (reference)</title>
+</head>
+<body>
+<p>Test passes if there is a small green circle (not scaled).</p>
```
nikolaszimmermann: Hmm, the difference between reftest and expectation is not as drastic as it could be IMHO.

"Test passes if there is a green circle covering the whole viewport."
<circle cx="100" cy="100" r="100" fill="green"/> for the expected result
<circle cx="100" cy="100" r="100" transform="scale(0.01)" fill="green"/> + transform: none for the actual test

Much easier to detect visually than a scale(2) difference, and the vague statement of a "small circle" (not scaled).

Source/WebCore/css/CSSProperties.json
```diff
@@ -11969,6 +11969,8 @@
                 ],
                 "animation-wrapper-acceleration": "always",
                 "style-extractor-custom": true,
+                "computed-style-has-explicitly-set-policy": "value-only",
```
nikolaszimmermann: Looks fine - but I'm not in expert in the CSS property descriptors after all the changes in the past years - would appreciate an ack from @weinig here.

nikolaszimmermann (approved):
Looks good, please double-check with @weinig regarding the CSSProperties.json changes though.

LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-none-overrides-attribute-expected.html
karlcow: Yes good point. Let me make a better test.

LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-none-overrides-attribute.html
karlcow: done

LayoutTests/imported/w3c/web-platform-tests/svg/styling/transform-none-overrides-attribute-expected.html
karlcow: done

Source/WebCore/css/CSSProperties.json
weinig: The "has-explicitly-set" stuff is not well defined, and the stuff I added in CSSProperties.json to support it was just modeling what existed already. There is no concept in the specs that matches it, and the implementation is not exactly consistent. For instance, a property that gets reverted or unset will still the explicitly set bit set. 

My understanding is that so far its mainly been use for heuristic driven purposes where it was alright for it be wrong some of the time.

That said, what is the rationale behind using "value-only"?

Here are some questions to ask:

Should `transform: initial` override?
Should `transform: inherit` override?

If so, you may want to use `"computed-style-has-explicitly-set-policy": "all-author-origin"` or add a new option, perhaps `computed-style-has-explicitly-set-policy": "all"`, that unconditionally sets it for each builder application.

weinig (commented):
Where is this specified?

Source/WebCore/css/CSSProperties.json
karlcow: @weinig

the transform attribute is defined in https://w3c.github.io/svgwg/svg2-draft/coords.html#TransformProperty which basically says:

> User agents must support the [transform](https://w3c.github.io/svgwg/svg2-draft/coords.html#TransformProperty) property and presentation attribute as defined in [[css-transforms-1](https://w3c.github.io/svgwg/svg2-draft/refs.html#ref-css-transforms-1)].

```
<svg>
  <style>circle { transform: none; }</style>
  <circle r="50" transform="scale(2)"/>
</svg>
```

currently Both Firefox and Chrome cancel the `scale(2)` set by the transform attribute when the CSS specifies `transform: none`. Safari doesn't. 

The CSS spec defines:
https://drafts.csswg.org/css-transforms/#transform-property

Value: none | [`<transform-list>`](https://drafts.csswg.org/css-transforms/#typedef-transform-list)
Initial: none
inherited: no

> Any computed value other than none for the transform affects containing block and stacking context, as described in [§ 2 The Transform Rendering Model](https://drafts.csswg.org/css-transforms/#transform-rendering).
> 
> `<transform-list>` = [`<transform-function>`](https://drafts.csswg.org/css-transforms/#typedef-transform-function)[+](https://drafts.csswg.org/css-values-4/#mult-one-plus)

**BUT** the spec has also a section about the SVG transform attribute. 
https://drafts.csswg.org/css-transforms/#svg-transform

On the other hand, there is a circularity issue, because the CSS spec is sending back SVG 1.1, not SVG2. And issue to open on the CSS WG.

To note that if we do `none`, for example `stroke` it cancels.

https://codepen.io/editor/webcompat/pen/019dbd30-6a56-7275-a823-63c0c8de1590
in this Safari returns `none` for both even when the value is set for scale.

**without specified CSS**
Safari:
circle computed stroke value: rgb(255, 0, 0)
square computed transform value: none

Firefox/Chrome:
circle computed stroke value: rgb(255, 0, 0)
square computed transform value: matrix(0.01, 0, 0, 0.01, 0, 0)

**Now let's apply a CSS**
https://codepen.io/editor/webcompat/pen/019dbd4a-603c-7880-89f3-0936d3a55895

```
   <style>
      circle {stroke: none}
      rect {transform: none}
    </style>
```

The stroke is being canceled in Safari, but the transform is not. 

Safari/Firefox/Chrome reports:
circle computed stroke value: none
square computed transform value: none

In Firefox/Chrome the square is visible as it should. It is not in Safari.

We can see that there is no get computed style being applied.

For `initial` and `inherit`:
Firefox/Chrome treats it as none. Safari doesn't do anything with it.

There is no synchronisation in between the CSS and the presentation attributes for these values.
On the other end if I do 
      `rect {transform: scale(2)}`
Then here we get: square computed transform value: matrix(2, 0, 0, 2, 0, 0)

so this is happening just for `none`, `initial`, `inherit`.

Does it help? 
What do you suggest for the values in CSSProperties.json?

karlcow: Oh I have forgotten also about 
https://w3c.github.io/svgwg/svg2-draft/styling.html#PresentationAttributes

Some styling properties can be specified not only in style sheets and ‘[style](https://w3c.github.io/svgwg/svg2-draft/styling.html#StyleAttribute)’ attributes, but also in presentation attributes. These are attributes whose name matches (or is similar to) a given CSS property and whose value is parsed as a value of that property. Presentation attributes contribute to the [author level](https://drafts.csswg.org/css-cascade-3/#preshint) of the cascade, followed by all other author-level style sheets, and have specificity 0.

Since presentation attributes are parsed as CSS values, not declarations, an [!important declaration](https://www.w3.org/TR/2011/REC-CSS2-20110607/cascade.html#important-rules) within a presentation attribute will cause it to have an [invalid value](https://w3c.github.io/svgwg/svg2-draft/types.html#TermInvalidValue). See [Attribute syntax](https://w3c.github.io/svgwg/svg2-draft/types.html#presentation-attribute-css-value) for details on how presentation attributes are parsed.

Not all style properties that can affect SVG rendering have a corresponding presentation attribute. Other attributes (which happen to share the name of a style property) must not be parsed as a presentation attribute and must not affect CSS cascading and inheritance. Also, only elements in the SVG namespace support presentation attributes. Most SVG presentation attributes may be specified on any element in the SVG namespace where there is not a name clash with an existing attribute. However, the [geometry properties](https://w3c.github.io/svgwg/svg2-draft/geometry.html#geometry-properties) only have equivalent presentation attributes on designated elements. Attributes of the same name on other elements must not affect CSS cascading and inheritance.

Except as noted in the table for the [transform](https://w3c.github.io/svgwg/svg2-draft/coords.html#TransformProperty) presentation attributes, the presentation attribute name is the same as the property name, in lower-case letters.

karlcow: Also https://w3c.github.io/svgwg/svg2-draft/types.html#presentation-attribute-css-value

The Initial value column gives the [initial value](https://w3c.github.io/svgwg/svg2-draft/types.html#TermInitialValue) for the attribute. When an attribute fails to parse according to the specified CSS Value Definition Syntax, ABNF or EBNF grammar, or if parsing according to the URL Standard or by the prose describing how to parse the attribute indicates failure, the attribute is assumed to have been specified as the given [initial value](https://w3c.github.io/svgwg/svg2-draft/types.html#TermInitialValue).

The initial value of a [presentation attribute](https://w3c.github.io/svgwg/svg2-draft/styling.html#TermPresentationAttribute) is its corresponding property's initial value. Since the use of an [invalid value](https://w3c.github.io/svgwg/svg2-draft/types.html#TermInvalidValue) in a [presentation attribute](https://w3c.github.io/svgwg/svg2-draft/styling.html#TermPresentationAttribute) will be treated as if the initial value was specified, this value can override values that come from lower priority style sheet rules, such as those from the user agent style sheet.

karlcow: stroke is defined with 
https://searchfox.org/wubkat/rev/15e55d72a144964cda9fc7d0635437c4a2bb3e2d/Source/WebCore/css/CSSProperties.json#7999-8031

```
        "stroke": {
            "animation-type": "by computed value",
            "inherited": true,
            "initial": "none",
            "values": [
                "none",
                {
                    "value": "context-fill",
                    "status": "unimplemented"
                },
                {
                    "value": "context-stroke",
                    "status": "unimplemented"
                }
            ],
            "codegen-properties": {
                "computed-style-storage-path": ["m_svgData", "strokeData"],
                "computed-style-type": "Style::SVGPaint",
                "computed-style-visited-link-storage-path": ["m_svgData", "strokeData"],
                "color-property": true,
                "color-property-traits-color-custom": true,
                "color-property-traits-visited-link-color-custom": true,
                "visited-link-color-support": true,
                "parser-grammar": "none | <color> | [ <url> [ none | <color> ]? ]",
                "parser-grammar-comment": "Current spec grammar is 'none | <color> | [ <url> [ none | <color> ]? ] | context-fill | context-stroke'",
                "comment": "Also specified (with different grammar) in CSS Fill and Stroke Module Level 3"
            },
            "specification": {
                "category": "svg",
                "url": "https://svgwg.org/svg2-draft/painting.html#SpecifyingStrokePaint"
            }
        },   
```

karlcow: But basically, any author-origin CSS declaration for transform (`none`, `initial`, `inherit`, function) should override the presentation attribute.

Plus SVG2 discourage to create future presentation attributes and to rely on CSS only. So that seems logical too.

karlcow: so yes you are right `all-author-origin` seems a better choice.

karlcow:
OK PR has been sent again. 
locally the test is passing. Let's see with the bots.

karlcow:
@weinig do you have a primer explaining all the vocabulary in CSSProperties.json, since you modified it. :) Just to make sure to use the right thing next time.

weinig (changes requested):
I had trouble following what you were trying to say above, but from the definition of ["presentation attribute"](https://w3c.github.io/svgwg/svg2-draft/styling.html#PresentationAttributes) in SVG2, it seems like the real issue here is that we are not mapping the "transform" attribute into the cascade. 

The correct fix for this likely to make the `transform` attribute (or `patternTransform` for `patten` elements, and `gradientTransform` for `linearGradient`/`radialGradient` elements) use the normal "presentation hint" infrastructure. (see SVGElement::cssPropertyIdForSVGAttributeName).

An example test case where the current approach would be incorrect is something like:

```
<!DOCTYPE html>
<html>
<body>
  <svg>
    <circle id="test" r="50" transform="scale(2)"/>
  </svg>

  <pre id="output"></pre>

  <script>
    const test = document.getElementById('test');
    const testComputedStyle = window.getComputedStyle(test);

    let testTransformValue = testComputedStyle.getPropertyValue('transform');

    document.getElementById('output').textContent += "test:       " + testTransformValue + "\n";
  </script>
</body>
</html>
```

In WebKit, this will always output "test:       none", but Firefox correctly outputs "test:       matrix(2, 0, 0, 2, 0, 0)".

I think adding an "hasExplictlySet" here is the wrong direction and is just papering over the real issue.

karlcow:
Thanks for the guidance @weinig I will look at it. This is useful.

karlcow:
I need to double check certain things, because this change seems to have happened with
https://github.com/WebKit/WebKit/pull/17228/changes

but transform seems to be already missing there. I need to double check if any presentation attributes  from the table other than transform family are missing.

And there is this [note](https://w3c.github.io/svgwg/svg2-draft/types.html#DOMInterfacesForReflectingSVGAttributes) : 

> In the future, any new properties that apply to SVG content will not gain presentation attributes. Therefore, authors are suggested to use styling properties, either through inline ‘[style](https://w3c.github.io/svgwg/svg2-draft/styling.html#StyleAttribute)’ properties or style sheets, rather than presentation attributes, for styling SVG content.

I guess this is for covering the future.

karlcow:
Note to myself, also for `transform` make sure we have invalidation tests for this. Check the table for all the requirements. 

> [transform](https://w3c.github.io/svgwg/svg2-draft/coords.html#TransformProperty)
> Any element in the SVG namespace with the exception of the ‘[pattern](https://w3c.github.io/svgwg/svg2-draft/pservers.html#PatternElement)’, ‘[linearGradient](https://w3c.github.io/svgwg/svg2-draft/pservers.html#LinearGradientElement)’ and ‘[radialGradient](https://w3c.github.io/svgwg/svg2-draft/pservers.html#RadialGradientElement)’ elements.

`transform` should not have any effect on these elements.

karlcow:
> In WebKit, this will always output "test: none", but Firefox correctly outputs "test: matrix(2, 0, 0, 2, 0, 0)".

Yes these are my test cases above. The codepen ones. It's what I was showing.

nullhook:
This will require using the presentational machinery. You'd be dirtying the presentational hints when the transform attribute changes, which will trigger style-resolution. In addition to having good test coverage, I suggest running perf tests to ensure there are no regressions. Also, take a look at how the `d` attribute is mapped and flows through the cascade. It also has a significant performance impact.

FWIW, this PR looks more involved.

karlcow:
I also filed https://bugs.webkit.org/show_bug.cgi?id=313374

nikolaszimmermann:
It was left out on purpose so far, because it always involved the performance regression compared to the hand-optimized SVG transform parser. Last time I tried this was two years ago in earlier LBSE work, and even after adding transform specific fast paths to the CSS Parser I couldn't get it to the same performance as the current implementation - might be worse to retry the experiment.

karlcow:
I moved to draft. And I will try to see if it breaks anything.

NO NEED TO REVIEW FOR NOW :p

PR #63228, opened 2026-04-21
[CMake] Default the build to Release instead of RelWithDebInfo when CMAKE_BUILD_TYPE is not specified.

clopez:
#### d45cb4091f3848d3190216a0bc401bedcfb415eb
<pre>
[CMake] Default the build to Release instead of RelWithDebInfo when CMAKE_BUILD_TYPE is not specified.
<a href="https://bugs.webkit.org/show_bug.cgi?id=312890">https://bugs.webkit.org/show_bug.cgi?id=312890</a>

Reviewed by NOBODY (OOPS!).

At <a href="https://commits.webkit.org/194203@main">194203@main</a> some code was added to warn the user when the
CMAKE_BUILD_TYPE is not specified and then to default the build
in that case to RelWithDebInfo.

But I think now it would be better to default that instead to
Release, because Release enables -O3 but RelWithDebInfo only -O2

Note: this does not affect builds done via build-webkit script as
build-webkit correctly passes the right -DCMAKE_BUILD_TYPE value.

* Source/cmake/WebKitCommon.cmake:
</pre>
https://github.com/WebKit/WebKit/commit/d45cb4091f3848d3190216a0bc401bedcfb415eb

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/157724 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/31061 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/24254 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/166548 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/111806 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/31196 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/31063 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/122128 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/85767 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/160682 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/24411 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/141630 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/102797 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/23467 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/21754 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/14319 "Built successfully") | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/149775 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/133147 "Passed tests") | | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/169037 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/18559 "Built successfully and passed tests") | | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/21066 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/130297 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/30807 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/25823 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/130414 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/30745 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/141238 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/88591 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/25188 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/18043 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/189852 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/30297 "Built successfully") | | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/48759 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/29818 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/30048 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/29945 "Built successfully") | | |


mcatanzaro:
Shrug. RelWithDebInfo seems like a good default to me because it matches what all Linux distros do in practice, which is -O2 -g. (This change doesn't affect Fedora because we do arbitrarily specify a build type, `Release`. But we also replace -O3 with -O2.)

I can tell you that Red Hat's toolchain team is pretty adamant that -O2 is the best general default for performance. They recommend -O3 only if you do performance analysis to check whether it is actually better than -O2. In particular, -O3 increases code size and therefore harms cache locality, so it tends to make code slower rather than faster. I wonder whether we have investigated -O2 vs. -O3 for WebKit.

If you want distros to build with -O3, then consider making an announcement. We can switch if you have investigated and say it is better.

mcatanzaro:
Also, very notably, if you don't specify a build type and do not manually enable -g, then after this change you will of course no longer be able to debug crashes, which is not great.

aperezdc:
I am with @mcatanzaro in this one, this seems like change for the sake of change, without any real advantage. If people want to reap performance benefits provided by the toolchain, instead of using `-O3` they would be better off using LTO + PGO (we landed some support in CMake for the latter recently), plus the set of linker flags that many distro packages make use of (something like `-Wl,-O1,--as-needed,--gc-sections,--sort-common,--no-undefined,-z,now,-z,relro,-z,noexecstack`).

Also, I think it's not the job of the build system to hardcode compiler/linker flags to enable, because that will annoy packagers to no end. If we want to recommend a set of compiler/linker flags, we are better off either documenting them in a “how to build” document (which would be good to have anyway), or by having an opt-in CMake option (something like `-DENABLE_RECOMMENDED_TOOLCHAIN_FLAGS=ON`).

PR #63146, opened 2026-04-20
[TextureMapper] Fix drop-shadow filter to align with the layer's transform

zhani:
#### 099eb0a77aff9b12a423ef10b80cfd9159b6ca9c
<pre>
[TextureMapper] Fix drop-shadow filter to align with the layer&apos;s transform
<a href="https://bugs.webkit.org/show_bug.cgi?id=218734">https://bugs.webkit.org/show_bug.cgi?id=218734</a>

Reviewed by NOBODY (OOPS!).

When a composited layer has a CSS filter, the layer&apos;s transform was
baked into the intermediate surface before the filter ran, so
drop-shadow(20px 20px) on a rotated layer produced an axis-aligned
shadow in surface space instead of one that rotates with the content.

Paint the intermediate surface in the layer&apos;s local (untransformed)
coordinate space, apply the filter there, and use the layer&apos;s
combined transform as the modelView when drawing the surface back.

Test: compositing/filters/drop-shadow-transform.html

* LayoutTests/compositing/filters/drop-shadow-transform-expected.html: Added.
* LayoutTests/compositing/filters/drop-shadow-transform.html: Added.
* Source/WebCore/platform/graphics/texmap/TextureMapperLayer.cpp:
(WebCore::TextureMapperLayer::computeConsolidatedOverlapRegionRects):
(WebCore::TextureMapperLayer::paintSelfChildrenFilterAndMask):
(WebCore::TextureMapperLayer::paintSelfAndChildrenWithIntermediateSurface):
* Source/WebCore/platform/graphics/texmap/TextureMapperLayer.h:
</pre>
https://github.com/WebKit/WebKit/commit/4fba0c022d41d1d8cd90a2b8a633b7e11782edfd

| Misc | iOS, visionOS, tvOS & watchOS  | macOS  | Linux |  Windows |
| ----- | ---------------------- | ------- |  ----- |  --------- |
| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/158246 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/31583 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/24778 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/167075 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/112330 "Built successfully") 
[...]

PR #63233, opened 2026-04-21
[GTK][WPE] Fix SkiaCompositingLayer drop-shadow transform rendering

zhani:
#### 2bb731602a5685bb9c537c4e985e67bc9e64dc55
<pre>
[GTK][WPE] Fix SkiaCompositingLayer drop-shadow transform rendering
<a href="https://bugs.webkit.org/show_bug.cgi?id=312898">https://bugs.webkit.org/show_bug.cgi?id=312898</a>

Reviewed by NOBODY (OOPS!).

When a composited layer has both a CSS filter and a transform, the
drop-shadow keeps falling in a fixed screen direction instead of rotating
with the content, and a blur is not scaled by the layer transform or the
device scale factor either.

The filtered subtree is painted into a device-space intermediate surface,
with the layer transform already baked into its pixels, and the
SkImageFilter is applied when that surface is drawn back with an identity
CTM. Skia maps filter parameters through the CTM, so the drop-shadow offset
follows the device axes and the blur sigma stays unscaled.

Give the filter the linear part of the layer transform as a local matrix,
which remaps its parameters without resampling the surface. Skia's blur
only accepts a scale-translate matrix, so for a rotation factor the scale
out and apply the remaining rotation around the filter with MatrixTransform
and Compose. Axis-aligned rotations are exact; other angles resample the
content twice.

Windows still composites with TextureMapper, which has the same bug, so the
test is marked as an image failure there.

Test: compositing/filters/drop-shadow-transform.html

* LayoutTests/compositing/filters/drop-shadow-transform-expected.html: Added.
* LayoutTests/compositing/filters/drop-shadow-transform.html: Added.
* LayoutTests/platform/win/TestExpectations:
(WebCore::filterForLayerTransform):
(WebCore::SkiaCompositingLayer::paintWithFilterAndMask):
</pre>
https://github.com/WebKit/WebKit/commit/701c931c7ad9b74627694bbd2bdd6ebaf621dd8b

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/185416 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/59328 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/53145 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/195740 "Built successfully") | [loading 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/150192 "Build is in progress. Recent messages:Printed configuration; Checked out pull request; Compiled WebKit") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/187278 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/60693 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/59055 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/144898 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/100533 "Exiting early after 60 failures. 60 tests run. 61 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/188371 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/48578 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/165481 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/125190 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/47306 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/45362 "Passed tests") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/157673 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/45053 "Passed tests") | [✅ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/6791 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/51984 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/49750 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/197688 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/58992 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/164989 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/154411 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/59701 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/41658 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/154063 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/48546 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/132042 "Built successfully") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/128888 "The change is no longer eligible for processing.") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/58179 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/20075 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/57551 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/57794 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/57618 "Built successfully") | | |


zhani:
EWS fails because TextureMapper does not pass the test introduced by this patch. There is a separate TextureMapper fix in #63146 . That TextureMapper change can be treated as a hotfix, while this patch addresses the review comments on #63146 and applies the approach suggested by @fujii to SkiaCompositingLayer

zhani:
@carlosgcampos I’ve rebased and updated this PR so it should be ready for review again. It’s been open for a while, sorry about the long gap on my side.

LayoutTests/compositing/filters/drop-shadow-transform.html
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <style>
+      .x div {
+	      will-change: transform;
```
nikolaszimmermann: Seems to have a tab in here, rest is tab-free.

LayoutTests/compositing/filters/drop-shadow-transform.html
```diff
@@ -0,0 +1,38 @@
...
+      }
+      div div {
```
nikolaszimmermann: In general I'd prefer class matching instead of "div div" so that it's clear what you apply this to without having to scroll down :-)

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
nikolaszimmermann, line 1210: Don't we need to use the new `filterForLayerTransform` also here for the backdrop-filters? :-)

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -1269,6 +1269,27 @@ void SkiaCompositingLayer::paintWithMaskAndBackdrop(SkCanvas& canvas, PaintConte
     paintWithBlendMode(canvas, context);
 }
 
+static sk_sp<SkImageFilter> filterForLayerTransform(const sk_sp<SkImageFilter>& filter, const TransformationMatrix& transform)
+{
+    SkMatrix matrix = SkMatrix::MakeAll(transform.a(), transform.c(), 0, transform.b(), transform.d(), 0, 0, 0, 1);
+    if (auto localFilter = filter->makeWithLocalMatrix(matrix))
```
nikolaszimmermann, line 1275: TIL makeWithLocalMatrix - neat way to optimize having to construct another new filter...

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -1269,6 +1269,27 @@ void SkiaCompositingLayer::paintWithMaskAndBackdrop(SkCanvas& canvas, PaintConte
     paintWithBlendMode(canvas, context);
 }
 
+static sk_sp<SkImageFilter> filterForLayerTransform(const sk_sp<SkImageFilter>& filter, const TransformationMatrix& transform)
+{
+    SkMatrix matrix = SkMatrix::MakeAll(transform.a(), transform.c(), 0, transform.b(), transform.d(), 0, 0, 0, 1);
```
nikolaszimmermann, line 1274: Hmmm, while this is 100% correct for the affine case, I think it's not correct in general, if perspective transformations are involved.... Need to think about this...

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -1269,6 +1269,27 @@ void SkiaCompositingLayer::paintWithMaskAndBackdrop(SkCanvas& canvas, PaintConte
...
+    if (!matrix.decomposeScale(&scale, &remaining))
+        return filter;
+    auto inverseRemaining = remaining.invert();
+    if (!inverseRemaining)
+        return filter;
+    auto localFilter = filter->makeWithLocalMatrix(SkMatrix::Scale(scale.width(), scale.height()));
+    if (!localFilter)
+        return filter;
+
+    SkSamplingOptions sampling(SkFilterMode::kLinear);
```
nikolaszimmermann, line 1289: Always unconditional bi-linear filtering? We recently changed the code to always check if we can't better use Nearest filtering instead

fujii:
Looks much better than my suggested approach. I'm sorry for disrupting.
I added a test case. PTAL at https://bugs.webkit.org/show_bug.cgi?id=312898#c3

LayoutTests/compositing/filters/drop-shadow-transform.html
zhani: fixed

LayoutTests/compositing/filters/drop-shadow-transform.html
zhani: fixed

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
zhani: Yes, it has the same problem but I'd like to keep that out of this patch and do it as a separate commit with its own test

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
zhani: Agreed, it is exact only for the affine case. Before this patch a perspective layer ignored the transform entirely, so this is strictly an improvement for that case. I was planning to do follow up patch which deals with perspective transformations

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
zhani: Not sure which code you mean, I assume SkiaUtilities::samplingOptionsForMatrix()? This fallback runs only for matrices that are not scale-translate so call it it here would still yield linear.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
nikolaszimmermann: I thought about this, and I think it needs a non-trivial amount of code to fix it for the perspective case... and code we really don't need if we alter the way how we handle transforms with filters.

Your PR inspired mine: https://github.com/WebKit/WebKit/pull/73501
Please take a look @zhani - I think this is a generic solution which utilitizes Skias ability to handle the transform decompositon for affine and perspective for us - also fixes backdrop-filter  etc..

zhani: nice, I will drop this then

nikolaszimmermann: @zhani but please check, if you see any flaws - another pair of eyes can't hurt.

zhani:
Dropping this in favor of #73501

PR #63323, opened 2026-04-22
[WPE][GTK] Initialize Vulkan in a way that will allow sharing allocated buffers with OpenGL

aperezdc:
#### 5bc4ddb9f31fb9002336127c75a643786f1d7b56
<pre>
[WPE][GTK] Initialize Vulkan in a way that will allow sharing allocated buffers with OpenGL
<a href="https://bugs.webkit.org/show_bug.cgi?id=312992">https://bugs.webkit.org/show_bug.cgi?id=312992</a>

This imports the needed CMake bits to detect and use Volk, which is a
Vulkan loader that works in a way similar to Epoxy. This way it is not
required at runtime to have Vulkan installed, and the built WebKit
library does not depend on a fixed version of libvulkan.

The rest of the patch is adding scaffolding for dealing with Vulkan
types in a more idiomatic and type-safe way, reusing C++ types from
WTF, and includes initialization code for the Vulkan context in the
WebProcess.

Follow-up patches will wire device selection, and buffer allocation.

* Source/WebCore/PlatformGTK.cmake:
* Source/WebCore/PlatformWPE.cmake:
* Source/WebCore/SourcesGTK.txt:
* Source/WebCore/SourcesWPE.txt:
* Source/WebCore/platform/Logging.h:
* Source/WebCore/platform/graphics/vulkan/VulkanTypes.cpp: Added.
(WebCore::Vulkan::ApplicationInfo::ApplicationInfo):
(WebCore::Vulkan::InstanceCreateInfo::InstanceCreateInfo):
(WebCore::Vulkan::PhysicalDevice::fillProperties const):
(WebCore::Vulkan::Instance::availableLayers):
(WebCore::Vulkan::Instance::hasLayers):
(WebCore::Vulkan::Instance::hasLayer):
(WebCore::Vulkan::Instance::availableExtensions):
(WebCore::Vulkan::Instance::hasExtensions):
(WebCore::Vulkan::Instance::hasExtension):
(WebCore::Vulkan::Instance::setSharedInstance):
(WebCore::Vulkan::Instance::sharedInstanceIfExists):
(WebCore::Vulkan::Instance::sharedInstance):
(WebCore::Vulkan::Instance::create):
(WebCore::Vulkan::Instance::~Instance):
(WebCore::Vulkan::Instance::availableDevices const):
(WebCore::Vulkan::debugUtilsMessengerHandleMessage):
(WebCore::Vulkan::Instance::installDebugMessenger):
(WebCore::Vulkan::Instance::installDebugMessenger const):
* Source/WebCore/platform/graphics/vulkan/VulkanTypes.h: Added.
(WebCore::Vulkan::BaseStruct::BaseStruct):
(WebCore::Vulkan::BaseStruct::operator=):
(WebCore::Vulkan::BaseStruct::ptr const):
(WebCore::Vulkan::BaseStruct::ptr):
(WebCore::Vulkan::Structure::Structure):
(WebCore::Vulkan::Structure::next):
(WebCore::Vulkan::Structure::operator-&gt; const):
(WebCore::Vulkan::Structure::operator-&gt;):
(WebCore::Vulkan::ApplicationInfo::ApplicationInfo):
(WebCore::Vulkan::InstanceCreateInfo::InstanceCreateInfo):
(WebCore::Vulkan::PhysicalDeviceProperties::operator-&gt; const):
(WebCore::Vulkan::PhysicalDeviceProperties::operator-&gt;):
(WebCore::Vulkan::Instance::Instance):
(WebCore::Vulkan::Instance::operator=):
* Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp: Added.
(WebCore::Vulkan::resultString):
(WebCore::Vulkan::initializeIfNeeded):
* Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h: Added.
(WebCore::Vulkan::resultString):
* Source/WebKit/Platform/Logging.h:
(WebKit::WebChromeClient::createGraphicsContextGL const):
* Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/DrawingAreaCoordinatedGraphicsGLib.cpp:
(WebKit::DrawingAreaCoordinatedGraphics::updatePreferences):
* Source/WebKit/WebProcess/WebProcess.h:
* Source/WebKit/WebProcess/glib/WebProcessGLib.cpp:
(WebKit::WebProcess::initializeVulkanIfNeeded):
(WebKit::WebProcess::platformInitializeWebProcess):
* Source/cmake/OptionsGTK.cmake:
* Source/cmake/OptionsWPE.cmake:

Canonical link: <a href="https://commits.webkit.org/312614@main">https://commits.webkit.org/312614@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/b10d0588bcc42ce0a425342b241c60e40f922724

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/160550 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/34023 "Built successfully") | [  ~~🛠 mac~~](https://ews-build.webkit.org/#/builders/168/builds/27124 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/169453 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/115616 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/162419 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/34124 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/34023 "Built successfully") | [  ~~🧪 wpe-wk2~~](https://ews-build.webkit.org/#/builders/34/builds/124504 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/115616 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/163509 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/26744 "Passed tests") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/144216 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/105104 "Passed tests") | 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/25796 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/24299 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/17172 "Built successfully") | 
| [  ~~🛠 🧪 jsc~~](https://ews-build.webkit.org/#/builders/20/builds/152606 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/135490 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [  ~~🧪 mac-wk1~~](https://ews-build.webkit.org/#/builders/167/builds/21979 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/171932 "Built successfully") | 
| [  ~~🛠 🧪 jsc-debug-arm64~~](https://ews-build.webkit.org/#/builders/171/builds/21387 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/18714 "Built successfully") | [  ~~🧪 mac-wk2~~](https://ews-build.webkit.org/#/builders/169/builds/23619 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/1/builds/132768 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/33697 "Built successfully") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/28392 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/132783 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35877 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/33681 "Built successfully") | [  ~~🧪 mac-wk2-stress~~](https://ews-build.webkit.org/#/builders/8/builds/143774 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/95477 "Built successfully") | 
| | [  ~~🧪 vision-wk2~~](https://ews-build.webkit.org/#/builders/164/builds/27378 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/20588 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/192949 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/33192 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/100317 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/49687 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/32691 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/32935 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/32839 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/platform/graphics/vulkan/VulkanTypes.h
+++ b/Source/WebCore/platform/graphics/vulkan/VulkanTypes.h
@@ -1,0 +1,207 @@
+/*
+ * Copyright (C) 2026 Igalia S.L.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL APPLE INC. OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ * OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#if USE(VULKAN)
+#include <expected>
+#include <volk.h>
+#include <wtf/Noncopyable.h>
+
+namespace WebCore {
+namespace Vulkan {
+
+template <typename Type>
+using Result = std::expected<Type, VkResult>;
+
+template <typename VulkanType>
+struct BaseStruct {
+    WTF_MAKE_NONCOPYABLE(BaseStruct)
+
+    using Base = BaseStruct<VulkanType>;
+
+    BaseStruct()
+    {
+        zeroBytes(m_inner);
+    }
+
+    BaseStruct(VulkanType&& inner)
+        : m_inner(WTF::move(inner))
+    {
+    }
+
+    ~BaseStruct() = default;
+
+    BaseStruct(BaseStruct&& other)
+        : BaseStruct()
+    {
+        std::swap(m_inner, other.m_inner);
+    }
+
+    BaseStruct& operator=(BaseStruct&& other)
+    {
+        if (this != &other) [[likely]]
+            std::swap(m_inner, other.m_inner);
+        return *this;
+    }
+
+    const VulkanType* LIFETIME_BOUND ptr() const { return &m_inner; }
+    VulkanType* LIFETIME_BOUND ptr() { return &m_inner; }
+
+protected:
+    VulkanType m_inner;
+};
+
+template <typename VulkanType, VkStructureType vulkanStructureType>
+struct Structure : BaseStruct<VulkanType> {
+    using Type = VulkanType;
+    using Base = BaseStruct<VulkanType>;
+
+    static constexpr VkStructureType typeCode = vulkanStructureType;
+
+    Structure()
+    {
+        this->m_inner.sType = typeCode;
+    }
+
+    Structure(VulkanType&& inner)
+        : Base(WTF::move(inner))
+    {
+        ASSERT(this->m_inner.sType == typeCode);
+    }
+
+    Structure(Structure&& other)
+        : Structure(WTF::move(other.m_inner))
+    {
+    }
+
+    // Create another structure, make it the "next" to this, and return it.
+    // This enables the following handy idiom to create chained Structure
+    // subtype instances:
+    //
+    //   struct A : Structure<...> { };
+    //   struct B : Structure<...> { };
+    //   struct C : Structure<...> { };
+    //
+    //   A first;
+    //   auto second = first.next<B>();
+    //   auto third = second.next<C>();
+    //
+    // Then the resulting chain of structures is: first -> second -> third.
+    //
+    template <typename OtherType, typename... Args>
+    requires std::derived_from<OtherType, Structure<typename OtherType::Type, OtherType::typeCode>>
+    [[nodiscard]] OtherType next(Args&&... params)
+    {
+        OtherType nextStructure { std::forward<Args>(params)... };
+        this->m_inner.pNext = nextStructure.ptr();
+        return nextStructure;
+    }
+
+    const VulkanType* LIFETIME_BOUND operator->() const { return this->ptr(); }
+    VulkanType* LIFETIME_BOUND operator->() { return this->ptr(); }
+};
+
+struct ApplicationInfo : Structure<VkApplicationInfo, VK_STRUCTURE_TYPE_APPLICATION_INFO> {
+    ApplicationInfo(const String& applicationName, uint32_t apiVersion = VK_API_VERSION_1_3)
+        : ApplicationInfo(applicationName.utf8().data(), apiVersion)
+    {
+    }
+
+private:
+    ApplicationInfo(const char* applicationName, uint32_t apiVersion);
+};
+
+struct InstanceCreateInfo : Structure<VkInstanceCreateInfo, VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO> {
+    InstanceCreateInfo(const ApplicationInfo& applicationInfo LIFETIME_BOUND, std::span<const char* const> enabledLayers LIFETIME_BOUND = { }, std::span<const char* const> enabledExtensions LIFETIME_BOUND = { });
+};
+
+struct PhysicalDeviceProperties : Structure<VkPhysicalDeviceProperties2, VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2> {
+    const VkPhysicalDeviceProperties* LIFETIME_BOUND operator->() const { return &m_inner.properties; }
+    VkPhysicalDeviceProperties* LIFETIME_BOUND operator->() { return &m_inner.properties; }
+};
+
+struct PhysicalDeviceDRMProperties : Structure<VkPhysicalDeviceDrmPropertiesEXT, VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DRM_PROPERTIES_EXT> {
+};
+
+struct PhysicalDevice : BaseStruct<VkPhysicalDevice> {
+    void fillProperties(PhysicalDeviceProperties&) const;
+};
+
+struct Instance : BaseStruct<VkInstance> {
+    [[nodiscard]] static const Vector<VkLayerProperties>& availableLayers();
+    [[nodiscard]] static bool hasLayers(std::span<const char* const> layerNames);
+    [[nodiscard]] static bool hasLayer(const char* const layerName);
+
+    [[nodiscard]] static Result<Vector<VkExtensionProperties>> availableExtensions(const char* layerName = nullptr);
+    [[nodiscard]] static bool hasExtensions(const Vector<VkExtensionProperties>&, std::span<const char* const> extensionNames);
+    [[nodiscard]] static bool hasExtension(const Vector<VkExtensionProperties>&, const char* const extensionName);
+    [[nodiscard]] static bool hasExtensions(std::span<const char* const> extensionNames, const char* layerName = nullptr);
+    [[nodiscard]] static bool hasExtension(const char* const extensionName, const char* layerName = nullptr);
+
+    static void setSharedInstance(Instance&&);
+    [[nodiscard]] static Instance* sharedInstanceIfExists();
+    [[nodiscard]] static Instance& sharedInstance();
+
+    [[nodiscard]] static Result<Instance> create(const InstanceCreateInfo&);
+    ~Instance();
+
+    Instance(Instance&& other)
+        : Base()
+    {
+        *this = WTF::move(other);
+    }
+
+    Instance& operator=(Instance&& other)
+    {
+        if (this != &other) {
+            std::swap(m_inner, other.m_inner);
+#ifdef VK_EXT_debug_utils
+            std::swap(m_debugMessenger, other.m_debugMessenger);
+#endif
+        }
+        return *this;
+    }
+
+    [[nodiscard]] Result<Vector<PhysicalDevice>> availableDevices() const;
+    [[nodiscard]] VkResult installDebugMessenger();
+
+private:
+    Instance(VkInstance inner)
+        : Base(WTF::move(inner))
+    {
+    }
+
+    static Instance s_sharedInstance;
+
+#ifdef VK_EXT_debug_utils
+    VkDebugUtilsMessengerEXT m_debugMessenger { nullptr };
+#endif
+};
+
+} // namespace Vulkan
+} // namespace WebCore
+
+#endif // USE(VULKAN)
--- a/Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
+++ b/Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
@@ -1,0 +1,278 @@
+/*
+ * Copyright (C) 2026 Igalia S.L.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL APPLE INC. OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ * OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "VulkanUtilities.h"
+
+#if USE(VULKAN)
+#include "Logging.h"
+#include <wtf/RuntimeApplicationChecks.h>
+#include <wtf/StdLibExtras.h>
+#include <wtf/Vector.h>
+#include <wtf/text/ASCIILiteral.h>
+#include <wtf/text/MakeString.h>
+#include <wtf/text/StringBuilder.h>
+
+#if USE(GLIB)
+#include "ApplicationGLib.h"
+#endif
+
+#ifndef VK_KHR_EXTERNAL_MEMORY_CAPABILITIES_EXTENSION_NAME
+#define VK_KHR_EXTERNAL_MEMORY_CAPABILITIES_EXTENSION_NAME "VK_KHR_external_memory_capabilities"
+#endif
+
+namespace WebCore {
+namespace Vulkan {
+
+const char* resultString(VkResult result)
+{
+    switch (result) {
+    case VK_SUCCESS:
+        return "SUCCESS";
+    case VK_NOT_READY:
+        return "NOT_READY";
+    case VK_TIMEOUT:
+        return "TIMEOUT";
+    case VK_EVENT_SET:
+        return "EVENT_SET";
+    case VK_EVENT_RESET:
+        return "EVENT_RESET";
+    case VK_INCOMPLETE:
+        return "INCOMPLETE";
+    case VK_ERROR_OUT_OF_HOST_MEMORY:
+        return "ERROR_OUT_OF_HOST_MEMORY";
+    case VK_ERROR_OUT_OF_DEVICE_MEMORY:
+        return "ERROR_OUT_OF_DEVICE_MEMORY";
+    case VK_ERROR_INITIALIZATION_FAILED:
+        return "ERROR_INITIALIZATION_FAILED";
+    case VK_ERROR_DEVICE_LOST:
+        return "ERROR_DEVICE_LOST";
+    case VK_ERROR_MEMORY_MAP_FAILED:
+        return "ERROR_MEMORY_MAP_FAILED";
+    case VK_ERROR_LAYER_NOT_PRESENT:
+        return "ERROR_LAYER_NOT_PRESENT";
+    case VK_ERROR_EXTENSION_NOT_PRESENT:
+        return "ERROR_EXTENSION_NOT_PRESENT";
+    case VK_ERROR_FEATURE_NOT_PRESENT:
+        return "ERROR_FEATURE_NOT_PRESENT";
+    case VK_ERROR_INCOMPATIBLE_DRIVER:
+        return "ERROR_INCOMPATIBLE_DRIVER";
+    case VK_ERROR_TOO_MANY_OBJECTS:
+        return "ERROR_TOO_MANY_OBJECTS";
+    case VK_ERROR_FORMAT_NOT_SUPPORTED:
+        return "ERROR_FORMAT_NOT_SUPPORTED";
+    case VK_ERROR_FRAGMENTED_POOL:
+        return "ERROR_FRAGMENTED_POOL";
+    case VK_ERROR_UNKNOWN:
+        return "ERROR_UNKNOWN";
+    case VK_ERROR_OUT_OF_POOL_MEMORY:
+        return "ERROR_OUT_OF_POOL_MEMORY";
+    case VK_ERROR_INVALID_EXTERNAL_HANDLE:
+        return "ERROR_INVALID_EXTERNAL_HANDLE";
+    case VK_ERROR_INVALID_OPAQUE_CAPTURE_ADDRESS:
+        return "ERROR_INVALID_OPAQUE_CAPTURE_ADDRESS";
+    case VK_ERROR_FRAGMENTATION:
+        return "ERROR_FRAGMENTATION";
+    case VK_PIPELINE_COMPILE_REQUIRED:
+        return "PIPELINE_COMPILE_REQUIRED";
+#ifdef VK_EXT_global_priority
+    case VK_ERROR_NOT_PERMITTED_EXT:
+        return "ERROR_NOT_PERMITTED_EXT";
+#endif
+#ifdef VK_KHR_surface
+    case VK_ERROR_SURFACE_LOST_KHR:
+        return "ERROR_SURFACE_LOST_KHR";
+    case VK_ERROR_NATIVE_WINDOW_IN_USE_KHR:
+        return "ERROR_NATIVE_WINDOW_IN_USE_KHR";
+#endif
+#ifdef VK_KHR_swapchain
+    case VK_SUBOPTIMAL_KHR:
+        return "SUBOPTIMAL_KHR";
+    case VK_ERROR_OUT_OF_DATE_KHR:
+        return "ERROR_OUT_OF_DATE_KHR";
+#endif
+#ifdef VK_KHR_display_swapchain
+    case VK_ERROR_INCOMPATIBLE_DISPLAY_KHR:
+        return "ERROR_INCOMPATIBLE_DISPLAY_KHR";
+#endif
+#ifdef VK_NV_glsl_shader
+    case VK_ERROR_INVALID_SHADER_NV:
+        return "ERROR_INVALID_SHADER_NV";
+#endif
+#ifdef VK_KHR_video_queue
+    case VK_ERROR_IMAGE_USAGE_NOT_SUPPORTED_KHR:
+        return "ERROR_IMAGE_USAGE_NOT_SUPPORTED_KHR";
+    case VK_ERROR_VIDEO_PICTURE_LAYOUT_NOT_SUPPORTED_KHR:
+        return "ERROR_VIDEO_PICTURE_LAYOUT_NOT_SUPPORTED_KHR";
+    case VK_ERROR_VIDEO_PROFILE_OPERATION_NOT_SUPPORTED_KHR:
+        return "ERROR_VIDEO_PROFILE_OPERATION_NOT_SUPPORTED_KHR";
+    case VK_ERROR_VIDEO_PROFILE_FORMAT_NOT_SUPPORTED_KHR:
+        return "ERROR_VIDEO_PROFILE_FORMAT_NOT_SUPPORTED_KHR";
+    case VK_ERROR_VIDEO_PROFILE_CODEC_NOT_SUPPORTED_KHR:
+        return "ERROR_VIDEO_PROFILE_CODEC_NOT_SUPPORTED_KHR";
+    case VK_ERROR_VIDEO_STD_VERSION_NOT_SUPPORTED_KHR:
+        return "ERROR_VIDEO_STD_VERSION_NOT_SUPPORTED_KHR";
+#endif
+#ifdef VK_EXT_image_drm_format_modifier
+    case VK_ERROR_INVALID_DRM_FORMAT_MODIFIER_PLANE_LAYOUT_EXT:
+        return "ERROR_INVALID_DRM_FORMAT_MODIFIER_PLANE_LAYOUT_EXT";
+#endif
+#ifdef VK_EXT_full_screen_exclusive
+    case VK_ERROR_FULL_SCREEN_EXCLUSIVE_MODE_LOST_EXT:
+        return "ERROR_FULL_SCREEN_EXCLUSIVE_MODE_LOST_EXT";
+#endif
+#ifdef VK_KHR_deferred_host_operations
+    case VK_THREAD_IDLE_KHR:
+        return "THREAD_IDLE_KHR";
+    case VK_THREAD_DONE_KHR:
+        return "THREAD_DONE_KHR";
+    case VK_OPERATION_DEFERRED_KHR:
+        return "OPERATION_DEFERRED_KHR";
+    case VK_OPERATION_NOT_DEFERRED_KHR:
+        return "OPERATION_NOT_DEFERRED_KHR";
+#endif
+#ifdef VK_KHR_video_encode_queue
+    case VK_ERROR_INVALID_VIDEO_STD_PARAMETERS_KHR:
+        return "ERROR_INVALID_VIDEO_STD_PARAMETERS_KHR";
+#endif
+#ifdef VK_EXT_image_compression_control
+    case VK_ERROR_COMPRESSION_EXHAUSTED_EXT:
+        return "ERROR_COMPRESSION_EXHAUSTED_EXT";
+#endif
+#ifdef VK_EXT_shader_object
+    case VK_ERROR_INCOMPATIBLE_SHADER_BINARY_EXT:
+        return "ERROR_INCOMPATIBLE_SHADER_BINARY_EXT";
+#endif
+#ifdef VK_EXT_debug_report
+    case VK_ERROR_VALIDATION_FAILED_EXT:
+        return "ERROR_VALIDATION_FAILED_EXT";
+#endif
+    default:
+        return "UNKNOWN";
+    }
+}
+
+void initializeIfNeeded()
+{
+    if (Instance::sharedInstanceIfExists())
+        return;
+
+    if (auto result = volkInitialize(); result != VK_SUCCESS) {
+        RELEASE_LOG(Vulkan, "Vulkan initialization failed: %s (runtime libraries missing?)", resultString(result));
+        return;
+    }
+
+    if (!Instance::hasExtension(VK_KHR_EXTERNAL_MEMORY_CAPABILITIES_EXTENSION_NAME)) {
+        RELEASE_LOG_ERROR(Vulkan, "Required extension %s not present", VK_KHR_EXTERNAL_MEMORY_CAPABILITIES_EXTENSION_NAME);
+        return;
+    }
+
+    Vector<const char*> layerNames;
+    Vector<const char*> extensionNames {
+        VK_KHR_EXTERNAL_MEMORY_CAPABILITIES_EXTENSION_NAME,
+    };
+
+#ifdef VK_EXT_debug_utils
+    bool shouldInstallDebugMessenger = false;
+#endif // VK_EXT_debug_utils
+
+    if (LOG_CHANNEL(Vulkan).state != WTFLogChannelState::Off) [[unlikely]] {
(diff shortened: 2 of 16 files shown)


Source/WebKit/WebProcess/glib/WebProcessGLib.cpp
```diff
@@ -228,6 +331,10 @@ void WebProcess::platformInitializeWebProcess(WebProcessCreationParameters& para
     if (!parameters.applicationName.isEmpty())
         WebCore::setApplicationName(parameters.applicationName);
 
+#if USE(VULKAN)
+    initializeVulkan();
```
nikolaszimmermann: I would do it righter after the initializePlatformDisplayIfNeeded code.

Source/WebKit/WebProcess/glib/WebProcessGLib.cpp
```diff
@@ -180,6 +185,104 @@ void WebProcess::initializePlatformDisplayIfNeeded() const
     CRASH();
 }
 
+#if USE(VULKAN)
+void WebProcess::initializeVulkan()
```
nikolaszimmermann: I'm surprised this lives here, shouldn't we try to keep WebProcessGLib.cpp "minimal" and move that into another class doing the actual initializtion? Maybe VulkanInitialization.cpp in platform/graphics/vulkan/?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+#include <wtf/Forward.h>
+#include <wtf/Noncopyable.h>
+
+namespace WebCore {
+namespace Vulkan {
+
+template <typename Type>
+using Result = std::expected<Type, VkResult>;
+
+[[nodiscard]] const char* resultString(VkResult);
```
nikolaszimmermann, line 35: Hmm, I was about to complain that this is named in a very generic way, but it's in Vulkan sub-namespace so maybe ok...

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+template <typename Type>
+[[nodiscard]] const char* resultString(const Result<Type>& result)
+{
+    return resultString(result.error_or(VK_SUCCESS));
+}
+
+[[nodiscard]] std::span<const char* const> validationLayerNames();
+
+template <typename VulkanType>
+struct BaseStruct {
```
nikolaszimmermann: IMHO this would fit nicely into a dedicated VulkanTypes.h

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+    const Type* LIFETIME_BOUND ptr() const { return &m_inner; }
+    Type* LIFETIME_BOUND ptr() { return &m_inner; }
+
+protected:
+    Type m_inner;
+};
+
+template <typename VulkanType, VkStructureType vulkanStructureType>
+struct Structure : BaseStruct<VulkanType> {
+    using Type = BaseStruct<VulkanType>::Type;
```
nikolaszimmermann: you could also have a `using Base = BaseStruct<VulkanType>` here, to avoid repeating BaseStruct<VulkanType> in the constructor, and save the `Type` using statement instead.. no big deal.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+    static constexpr VkStructureType typeCode = vulkanStructureType;
+
+    Structure()
+    {
+        this->m_inner.sType = typeCode;
```
nikolaszimmermann: Why the explicit this-> usage here and elsewhere?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+        ASSERT(this->m_inner.sType == typeCode);
+    }
+
+    Structure(Structure&& other)
+        : Structure(WTF::move(other.m_inner))
+    {
+    }
+
+    template <typename WrapperType>
+    void chain(WrapperType& next)
```
nikolaszimmermann: I have no idea how this is intended to be used - I would appreciate more inline comments, just reading this without a Vulkan background, I have no idea why you need BaseStruct/Structure/chain() method, etc.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+
+private:
+    Instance(VkInstance inner)
+        : BaseStruct(WTF::move(inner))
+    {
+    }
+
+    static Instance s_sharedInstance;
+
+#ifdef VK_EXT_debug_utils
```
nikolaszimmermann: I guess that definition comes straight from the Vulkan headers?

Source/cmake/OptionsWPE.cmake
```diff
@@ -261,6 +262,13 @@ if (FREETYPE_WOFF2_SUPPORT_IS_AVAILABLE)
     endif ()
 endif ()
 
+if (USE_VULKAN)
+    find_package(volk CONFIG)
```
nikolaszimmermann: Where can we get volk from? Do you plan to supply it with the SDK? How is distro availabilty? yocto/buildroot?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+    return resultString(result.error_or(VK_SUCCESS));
+}
+
+[[nodiscard]] std::span<const char* const> validationLayerNames();
+
+template <typename VulkanType>
+struct BaseStruct {
+    WTF_MAKE_NONCOPYABLE(BaseStruct)
+
+    using Type = VulkanType;
```
nikolaszimmermann: I think that doesn't help much, I'd omit this definition.

nikolaszimmermann: Structure does this `using Type = BaseStruct<VulkanType>::Type;` -- so that's just VulkanType, no?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+struct ApplicationInfo : Structure<VkApplicationInfo, VK_STRUCTURE_TYPE_APPLICATION_INFO> {
+    ApplicationInfo(const ASCIILiteral& applicationName, uint32_t apiVersion = VK_API_VERSION_1_3);
+};
+
+struct InstanceCreateInfo : Structure<VkInstanceCreateInfo, VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO> {
+    InstanceCreateInfo(const ApplicationInfo& applicationInfo LIFETIME_BOUND, std::span<const char* const> enabledLayers LIFETIME_BOUND = { }, std::span<const char* const> enabledExtensions LIFETIME_BOUND = { });
+};
+
+struct PhysicalDeviceProperties : Structure<VkPhysicalDeviceProperties2, VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2> {
+    const VkPhysicalDeviceProperties* LIFETIME_BOUND operator->() const { return &m_inner.properties; }
```
nikolaszimmermann: Aren't these missing a "2" suffix for the type?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
nikolaszimmermann: Hmmm, also I can chain here every single type that has a ptr() method -- that seems weird.

Maybe a use-case for concepts?  Untested: `requires { { next.ptr() } -> std::convertible_to<const void*>; })` plus a check that next is actually a `Structure<xyz>` would catch accidental chaining at compile time?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+    VkPhysicalDeviceProperties* LIFETIME_BOUND operator->() { return &m_inner.properties; }
+};
+
+struct PhysicalDeviceDRMProperties : Structure<VkPhysicalDeviceDrmPropertiesEXT, VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DRM_PROPERTIES_EXT> {
+};
+
+struct PhysicalDevice : BaseStruct<VkPhysicalDevice> {
+    void getProperties(PhysicalDeviceProperties&) const;
```
nikolaszimmermann: PhysicalDeviceProperties cannot be moved? otherwise I'd assume to have a "PhysicalDeviceProperties properties() const" accessor here?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+};
+
+struct Instance : BaseStruct<VkInstance> {
+    [[nodiscard]] static const Vector<VkLayerProperties>& availableLayers();
+    [[nodiscard]] static bool hasLayers(std::span<const char* const> layerNames);
```
nikolaszimmermann: I guess this is a Vulkan concept :-) Would be really helpful to have some comments for these group of methods linking to Vulkan docs regarding layers...

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+
+    [[nodiscard]] static Result<Vector<VkExtensionProperties>> availableExtensions(const char* layerName = nullptr);
+    [[nodiscard]] static bool hasExtensions(const Vector<VkExtensionProperties>&, std::span<const char* const> extensionNames);
```
nikolaszimmermann: Why the span vs const char overloads?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+    [[nodiscard]] static bool hasExtension(const Vector<VkExtensionProperties>&, const char* const extensionName);
+    [[nodiscard]] static bool hasExtensions(std::span<const char* const> extensionNames, const char* layerName = nullptr);
+    [[nodiscard]] static bool hasExtension(const char* const extensionName, const char *layerName = nullptr);
```
nikolaszimmermann: Misplaced star.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+
+    static void setSharedInstance(Instance&&);
+    [[nodiscard]] static Instance* sharedInstanceIfExists();
+    [[nodiscard]] static Instance& sharedInstance();
```
nikolaszimmermann: ensureShardInstance?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+
+    [[nodiscard]] static Result<Instance> create(const InstanceCreateInfo&);
```
nikolaszimmermann: Should Instance creation always get routed through this method? Then I would try to make Instance constructor private.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+    Instance(Instance&&) = default;
+    ~Instance();
+
+    Instance& operator=(Instance&&) = default;
```
nikolaszimmermann: What about deleting operator=(const Instance&)?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
```diff
@@ -0,0 +1,181 @@
...
+
+    [[nodiscard]] Result<Vector<PhysicalDevice>> availableDevices() const;
+    [[nodiscard]] VkResult installDebugMessenger();
+
+private:
+    Instance(VkInstance inner)
+        : BaseStruct(WTF::move(inner))
```
nikolaszimmermann: The move here is noise, no? VkInstance is a pointer-shape type, right?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
```diff
@@ -0,0 +1,360 @@
...
+    switch (result) {
+#define RESULT_STRING_CASE(r) case VK_ ## r : return #r;
+        FOR_EACH_VULKAN_RESULT(RESULT_STRING_CASE)
+#undef RESULT_STRING_CASE
+    default:
+        return "UNKNOWN";
+    }
+}
+
+std::span<const char* const> validationLayerNames()
```
nikolaszimmermann: Do you plan to extend this with more layers? Looks unnecesary complicated no.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
```diff
@@ -0,0 +1,360 @@
...
+{
+    static constexpr char const* layerNames[] = {
+        "VK_LAYER_KHRONOS_validation",
+    };
+    return std::span(layerNames);
+}
+
+ApplicationInfo::ApplicationInfo(const ASCIILiteral& applicationName, uint32_t apiVersion)
+{
+    m_inner.pApplicationName = applicationName.characters();
```
nikolaszimmermann: I wonder if m_inner is a good name, it's a detail that its wrapped and thus "inner". How about "m_vulkan" / m_vkType / m_type / ... ?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
```diff
@@ -0,0 +1,360 @@
...
+}
+
+void PhysicalDevice::getProperties(PhysicalDeviceProperties& properties) const
+{
+    vkGetPhysicalDeviceProperties2(m_inner, properties.ptr());
+}
+
+const Vector<VkLayerProperties>& Instance::availableLayers()
+{
+    static const auto layerProperties = ([]() -> Vector<VkLayerProperties> {
```
nikolaszimmermann: Why the lambda here? Doesn't buy you much.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
```diff
@@ -0,0 +1,360 @@
...
+        return result;
+    })();
+    return layerProperties;
+}
+
+bool Instance::hasLayers(std::span<const char* const> layerNames)
+{
+    return std::ranges::all_of(layerNames, [](auto* layerName) -> bool {
+        return availableLayers().containsIf([layerName](const auto& layer) -> bool {
+            return CStringView::unsafeFromUTF8(layerName) == CStringView::unsafeFromUTF8(layer.layerName);
```
nikolaszimmermann: No better way to compare those names then these conversions? :-)

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
```diff
@@ -0,0 +1,360 @@
...
+    volkLoadInstanceOnly(instance);
+    return Instance(instance);
+}
+
+Instance::~Instance()
+{
+#ifdef VK_EXT_debug_utils
+    if (m_debugMessenger) {
+        vkDestroyDebugUtilsMessengerEXT(m_inner, m_debugMessenger, nullptr);
+        m_debugMessenger = nullptr;
```
nikolaszimmermann: No need to nullify things?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
```diff
@@ -0,0 +1,360 @@
...
+    m(RESULT_MAX_ENUM)
+
+namespace WebCore {
+namespace Vulkan {
+
+const char* resultString(VkResult result)
+{
+    switch (result) {
+#define RESULT_STRING_CASE(r) case VK_ ## r : return #r;
+        FOR_EACH_VULKAN_RESULT(RESULT_STRING_CASE)
```
nikolaszimmermann: Ewww macros, how about:

```
  const char* resultString(VkResult result)                                                                                                                                                   
  {                                                                                                                                                                                           
      struct Entry {                                                                                                                                                                          
          VkResult value;                                                                                                                                                                     
          ASCIILiteral name;                                                               
      };                              
      static constexpr std::array table {
          Entry { VK_SUCCESS,                              "SUCCESS"_s },
...
          Entry { VK_ERROR_OUT_OF_DATE_KHR,                "ERROR_OUT_OF_DATE_KHR"_s },                                                                                                       
      };                                                                                                                                                                                      
      for (const auto& entry : table) {                                                    
          if (entry.value == result)                                                                                                                                                          
              return entry.name;                                                           
      }                                                                                                                                                                                       
      return "UNKNOWN";                                                                    
  }                                   
```

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
```diff
@@ -0,0 +1,360 @@
...
+
+#ifdef VK_EXT_debug_utils
+struct DebugUtilsMessengerCreateInfo : Structure<VkDebugUtilsMessengerCreateInfoEXT, VK_STRUCTURE_TYPE_DEBUG_UTILS_MESSENGER_CREATE_INFO_EXT> {
+};
+
+static VkBool32 debugUtilsMessengerHandleMessage(VkDebugUtilsMessageSeverityFlagBitsEXT messageSeverity, VkDebugUtilsMessageTypeFlagsEXT messageTypes, const VkDebugUtilsMessengerCallbackDataEXT* data, void*)
+{
+    ASSERT(data->pMessage);
+
+    WTFLogLevel logLevel;
```
nikolaszimmermann:
```
  auto logLevel = [&] {                                                                                                                                                                       
      switch (messageSeverity) {
      case VK_DEBUG_UTILS_MESSAGE_SEVERITY_ERROR_BIT_EXT:
          return WTFLogLevel::Error;                                                                                                        
...
      }                                                                                                                                                                                       
      RELEASE_ASSERT_NOT_REACHED();                                                                                                                                                           
  }();                                                                                                                                                                                        
```
?

nikolaszimmermann (commented):
Sorry for the long stream of comments :-)

Source/WebKit/WebProcess/glib/WebProcessGLib.cpp
aperezdc: Ah, I remember thinking that it would go well here after the call to `WebCore::setApplicationName()` in order to use the saved value for the `Vulkan::ApplicationInfo` passed when instantiating the `Vulkan::Context`.

I think it makes sense to move it into next to `initializePlatformDisplayIfNeeded()` anyway, so let's do that.

Source/WebKit/WebProcess/glib/WebProcessGLib.cpp
aperezdc: Good point, and having the initialization routine elsewhere we can reuse it later down the road when/if we want to use Vulkan also in the UIProcess or the GPUProcess.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: I like the idea of a separate `VulanTypes.h` header where to have the base templates and the simpler struct/class definitions. It would make it easier if we wanted to move the bigger ones into their own source files in the future.

Source/WebKit/WebProcess/glib/WebProcessGLib.cpp
aperezdc: Actually, I think with the wrapper types living in `VulkanTypes.{h,cpp}`, the initialization routine could be in `VulkanUtilities.{h,cpp}` directly. I would only split it to its own `VulkanInitialization.cpp` file if `VulkanUtilities.cpp` becomes big (many thousands of LOC).

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: The whole idea behind adding a `WebCore::Vulkan::` namespace is being able to have short meaningful names that do not conflict with other existing things 😉

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: I tried your idea and indeed makes things look a bit nicer, thanks!

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: Accessing members from a template base class needs an explicit `this`, as explained e.g. [in this article](https://www.geeksforgeeks.org/cpp/why-should-i-access-template-base-class-members-through-this-pointer/).

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: I agree this needs to be a bit more concrete, only wrapper types that derive from `Structure<T>` should be allowed here.

Using the `pNext` pointer to chain structures is how additional parameters can be passed to/from functions which can have their functionality extended (either by higher Vulkan versions that convert extensions into base spec functionality, or by explicitly enabled extensions). I can write a comment next to this to explain a bit.

Also I am not super happy that one needs to manually do things like:

```cpp
A a;
B b;
C c;

a.chain(b);
b.chain(c);
```

Maybe using template varargs and `std::forward()` it would be possible to do instead something like:

```cpp
A a { /* designated initializer values for A */ };
B b = a.chained<B>({ /* designated initializer values for B */ });
C c = b.chained<C>({ /* designated initializer values for C */ });
```

Or even:

```cpp
A a { /* ... */ };
B b { /* ... */ };
C c { /* ... */ };
chain(a, b, c);  // Connects a.pNext = &b; b.pNext = &c;
```

I will give it a bit of thought, but I may leave making this prettier for a follow-up patch.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: Yes, the headers define a macro that is exactly the extension name for each extension that has their declarations included in the headers (this is the same as in OpenGL).

Source/cmake/OptionsWPE.cmake
aperezdc: I checked a bit before deciding to use it. Some distros have a copy in their official repositories:

* Ubuntu [has it](https://packages.ubuntu.com/noble/libvolk-dev) since 24.04 LTS.
* Debian [has it](https://packages.debian.org/trixie/libvolk-dev) since 13.x Trixie (current stable).
* Arch Linux [has it](https://archlinux.org/packages/extra/x86_64/volk/).
* Yocto [has it](https://git.openembedded.org/openembedded-core/tree/meta/recipes-graphics/vulkan/vulkan-volk_1.4.341.0.bb).

Others do not, e.g. Fedora or Buildroot.

Amusingly, the bundled ANGLE already has a copy of Volk, under `Source/ThirdParty/ANGLE/src/third_party/volk/volk.c` &mdash; we could reuse that, or include a different copy under `Source/ThirdParty/volk/` (but I think we may as well reuse ANGLE's).

A very nice thing about Volk is that it is only needed at build time, because it is always a static library, or only two files, `volk.{h,c}` compiled along the rest of WebKit.

aperezdc: FWIW, I can package it for Buildroot myself and submit a patch, no big deal there either.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: Agreed, let's remove those.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: This is correct withbout the suffix.

The original `VkPhysicalDeviceProperties` from Vulkan 1.0 did not have a `pNext` pointer. Later `VkPhysicalDeviceProperties2` was added, and it is basically a container for a `pNext` pointer + a member for the old struct.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: Calling `getProperties()` fills the passed `PhysicalDeviceProperties` and any other structures chained through their `pNext` pointers. Something like `fillProperties()` is probably a better name. I will also add a comment here.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: The ones with spans check that all the layers/extensions listed are available, the overload checks for a single layer/extension name without having to awkwardly build a span with a single item.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: Yes, [layers](https://docs.vulkan.org/guide/latest/layers.html) are a Vulkan concept. It basically formalizes the idea of dynamically loading optional components at runtime.

I think educating people about how Vulkan works does not belong in a source header, though. Literally all the reference documentation is at [docs.vulkan.org](https://docs.vulkan.org), including browseable lists of layers and extensions, with their reference documentation. Would a general comment at the top of `VulkanTypes.h` linking to that be enough?

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: This is following the same pattern as `PlatformDisplay`, which seems to be also the most common one.

There was some discussion about naming in `webkit-dev`, the last time [the thread](https://lists.webkit.org/archives/list/webkit-dev@lists.webkit.org/thread/JS2MR7WS24FIUV6YFQPWVVV2NCIWVHYA/) stopped getting messages on January 2025 without a clear conclusion.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: It is already private. The one that is public is the move-constructor.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: This type is movable, but not copyable. Defining the operator is needed for the assignment inside `::setSharedInstance()`.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: The move is still needed because the `Base()` constructor expects a rvalue-reference.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
aperezdc: I would avoid anything with “type” in the name. How about `m_value`? Although personally I see no problem in a wrapper having knowledge about itself being... well, a wrapper 😉

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
aperezdc: It gets us threadsafe static initialization without needing `std::once` or `LazyNeverDestroyed` 😉

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
aperezdc: Sadly, they are `const char*` and `char[VK_MAX_EXTENSION_NAME_SIZE]`, so we can either do this or use `strcmp()` adding the unsafe buffer usage suppresion macros. Neither option is great.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
aperezdc: I used a “generator macro” thinking that the error codes might be need to be enumerated for something else; but the need has not appeared so far.

Instead of something fancier, I think it's perfectly fine to write the `switch` in full by hand, so that's what the next iteration of the patch will have.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
aperezdc: Why? This would move the same `switch` inside the lambda so I don't see the advantage. The compiler will warn if the `logLevel` would be left uninitialized in case none of the `case` labels matched.

Source/WebKit/WebProcess/glib/WebProcessGLib.cpp
aperezdc: Done.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: Done.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: Done.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
aperezdc: Actually no. I first thought there would be more than one for this. I will replace the extra function with direct use of the layer name.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
aperezdc: Fixed.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
aperezdc: Done, and also while manually writing the `switch` I added guards for definitions added by extensions.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
aperezdc: I ended up with this:

```cpp
template <typename OtherType, typename... Args>
requires std::derived_from<OtherType, Structure<typename OtherType::Type, OtherType::typeCode>>
[[nodiscard]] OtherType next(Args&&... params)
{
    OtherType nextStructure { std::forward<Args>(params)... };
    this->m_inner.pNext = nextStructure.ptr();
    return nextStructure;
}
```

which allows doing:

```cpp
struct A : Structure<...> { };
struct B : Structure<...> { };
struct C : Structure<...> { };

A first;                     // first → ø
B second = first.next<B>();  // first → second → ø
C third = second.next<C>();  // first → second → third → ø
```

(Where the `ø` symbol represents a null `pNext` pointer.)

aperezdc:
This should be ready to take another look at it. Also I have added the missing bits for the GTK port in the CMake build system.

Source/cmake/OptionsWPE.cmake
nikolaszimmermann: Sounds great.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
nikolaszimmermann: m_wrapped? :-) Choose what you like this is not a deal-breaker.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
nikolaszimmermann: Shorter, no breaks?

Source/WebCore/platform/graphics/vulkan/VulkanTypes.h
```diff
@@ -0,0 +1,193 @@
...
+    //   struct C : Structure<...> { };
+    //
+    //   A first;
+    //   auto second = first.next<B>();
+    //   auto third = second.next<C>();
+    //
+    // Then the resulting chain of structures is: first -> second -> third.
+    //
+    template <typename OtherType, typename... Args>
+    requires std::derived_from<OtherType, Structure<typename OtherType::Type, OtherType::typeCode>>
```
nikolaszimmermann, line 115: Nice.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
nikolaszimmermann: Ah right, thanks.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
nikolaszimmermann: Ok.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
nikolaszimmermann: Yes that's enough.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
nikolaszimmermann: Ok, no strong opinion either.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.h
nikolaszimmermann: Ah right.

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
nikolaszimmermann: Uh-oh, interesting pattern.

nikolaszimmermann (approved):
Looks great!

Source/WebCore/platform/graphics/vulkan/VulkanUtilities.cpp
aperezdc: I left it as `m_inner` in the end.

aperezdc:
Also I have left the feature disabled by default and added a runtime check on `WEBKIT_VULKAN_ENABLED=1` (from the environment) to opt-in into the new stuff.

PR #63355, opened 2026-04-22
[Site Isolation] Add multi-process BFCache restoration with Site Isolation

basuke:
#### fceeb850c1e2dc758581520cdce13e04c4f0ff25
<pre>
[Site Isolation] Add multi-process BFCache restoration with Site Isolation
<a href="https://bugs.webkit.org/show_bug.cgi?id=313027">https://bugs.webkit.org/show_bug.cgi?id=313027</a>
<a href="https://rdar.apple.com/173743733">rdar://173743733</a>

Reviewed by Sihui Liu.

Complete multi-process BFCache restoration: when navigating back to a BFCache&apos;d page under Site
Isolation, restore cached iframe processes alongside the main frame.

Key changes:
- Split WebPage::setIsSuspendedWithFrameItem into WebPage::suspendWithFrameItem and
  WebPage::restoreWithFrameItem, backed by two distinct IPC messages
  (SuspendWithFrameItem / RestoreWithFrameItem). The new functions update
  m_isSuspended only after the underlying BackForwardCache operation succeeds,
  eliminating the transient-state window of the previous bool-flag design.
- In SuspendedPageProxy::unsuspend(mainFrameItemID), send RestoreWithFrameItem to
  each subframe process with aggregated failure handling that triggers a reload fallback.
- In WebProcessPool::processForNavigation, give a SuspendedPage explicit priority
  over site-based routing under Site Isolation by calling prepareProcessForNavigation
  directly when the target back/forward item has a suspended page.
- In receivedNavigationActionPolicyDecision, use the suspended page&apos;s BrowsingContextGroup for
  the new navigation so preserved subframe RemotePageProxies are reachable during restoration,
  and guard removeEntriesForPageAndProcess with frame-&gt;isMainFrame() so that cross-site iframe
  process swaps do not evict main-frame BFCache entries from the destination process.
- Make BrowsingContextGroup::addPage() idempotent when the page is already a member, guarded
  by ASSERT(!hasMultiplePages()) - the idempotent branch is only valid during suspended-BCG
  restoration, where the BCG holds exactly this one page. At suspension the BCG stays inert
  while held by SuspendedPageProxy, so m_pages can keep the entry through the suspend/restore
  cycle without a separate detach step.
- In WebPage::suspendForProcessSwap, guard against a null corePage and switch to the
  frame-item-keyed addIfCacheable signature.
- Strengthen MultiProcessBFCacheGoForward and MultiProcessBFCacheSameSiteNavAfterRestore with
  window-scoped markers that only survive when the page is restored from BFCache, not when it
  is reloaded from the network.

Test: Tools/TestWebKitAPI/Tests/WebKit/WKWebView/SiteIsolation.mm

(WebKit::BrowsingContextGroup::addPage):
(WebKit::ProvisionalPageProxy::ProvisionalPageProxy):
* Source/WebKit/UIProcess/SuspendedPageProxy.cpp:
(WebKit::SuspendedPageProxy::startSuspension):
(WebKit::SuspendedPageProxy::unsuspend):
(WebKit::SuspendedPageProxy::suspendSubframeProcesses):
* Source/WebKit/UIProcess/SuspendedPageProxy.h:
* Source/WebKit/UIProcess/WebProcessPool.cpp:
(WebKit::WebProcessPool::processForNavigation):
(WebKit::WebPage::suspendForProcessSwap):
(WebKit::WebPage::suspendWithFrameItem):
(WebKit::WebPage::restoreWithFrameItem):
(TestWebKitAPI::(SiteIsolation, MultiProcessBFCacheGoForwardSimple)):
(TestWebKitAPI::(SiteIsolation, MultiProcessBFCacheGoForward)):
(TestWebKitAPI::(SiteIsolation, MultiProcessBFCacheSameSiteNavAfterRestore)):

Canonical link: <a href="https://commits.webkit.org/313059@main">https://commits.webkit.org/313059@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/f9a0b9c6ae6626afa8ce2995205419fda0bc6367

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/161975 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/35450 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/28555 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/170883 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/118346 "Built successfully") | [loading 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/336099df-2464-40ab-88a5-89312544ee2d) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/163844 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/35552 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/35451 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/125692 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/88567 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/67a351b7-2fae-4351-95bd-36e424a76733) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/164934 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/28004 "Passed tests") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/145487 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/106274 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/27053 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/25564 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/18603 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/136746 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/23241 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/173371 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/20462 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/24882 "Passed tests") | [❌ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/133984 "Found 1 new test failure: accessibility/aria-owns-deep-chain-no-timeout.html (failure)") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/35123 "Built successfully") | [❌ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/29649 "Found 60 new test failures: accessibility/ARIA-reflection.html accessibility/combobox/aria-combobox-control-owns-elements.html accessibility/combobox/aria-combobox-hierarchy.html accessibility/combobox/aria-combobox-no-owns.html accessibility/combobox/aria-combobox.html accessibility/combobox/mac/aria-combobox-activedescendant.html accessibility/combobox/mac/combobox-activedescendant-notifications.html accessibility/combobox/mac/combobox-inherited-role.html accessibility/combobox/mac/combobox-value.html accessibility/custom-elements/autocomplete.html ... (failure)") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/133990 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36176 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/35107 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/145036 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/97220 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/28516 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/21861 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/34617 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/102102 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/34118 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/34360 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/34266 "Built successfully") | | | |

### DIFF
--- a/Source/WebKit/UIProcess/BrowsingContextGroup.cpp
+++ b/Source/WebKit/UIProcess/BrowsingContextGroup.cpp
@@ -207,7 +207,11 @@ void BrowsingContextGroup::removeFrameProcess(FrameProcess& process)
 
 void BrowsingContextGroup::addPage(WebPageProxy& page)
 {
-    ASSERT(!m_pages.contains(page));
+    if (m_pages.contains(page)) {
+        // This only happens when restoring a page from a suspended BCG, which holds exactly this one page.
+        ASSERT(!hasMultiplePages());
+        return;
+    }
     m_pages.add(page);
     auto& set = m_remotePages.ensure(page, [] {
         return HashSet<Ref<RemotePageProxy>> { };
--- a/Source/WebKit/UIProcess/SuspendedPageProxy.cpp
+++ b/Source/WebKit/UIProcess/SuspendedPageProxy.cpp
@@ -26,6 +26,7 @@
 #include "config.h"
 #include "SuspendedPageProxy.h"
 
+#include "APINavigation.h"
 #include "APIPageConfiguration.h"
 #include "BrowsingContextGroup.h"
 #include "DrawingAreaProxy.h"
@@ -36,6 +37,7 @@
 #include "RemotePageProxy.h"
 #include "WebBackForwardCache.h"
 #include "WebBackForwardList.h"
+#include "WebBackForwardListFrameItem.h"
 #include "WebBackForwardListMessages.h"
 #include "WebFrameProxy.h"
 #include "WebPageMessages.h"
@@ -168,6 +170,17 @@ void SuspendedPageProxy::startSuspension(std::optional<BackForwardFrameItemIdent
 
     m_messageReceiverRegistration.startReceivingMessages(m_process, m_webPageID, *this, *this);
     m_suspensionTimeoutTimer.startOneShot(suspensionTimeout);
+
+    if (mainFrameItemID) {
+        sendWithAsyncReply(Messages::WebPage::SuspendWithFrameItem(*mainFrameItemID), [weakThis = WeakPtr { *this }](bool didSuspend) {
+            RefPtr protectedThis = weakThis.get();
+            if (!protectedThis)
+                return;
+            protectedThis->didProcessRequestToSuspend(didSuspend ? SuspensionState::Suspended : SuspensionState::FailedToSuspend);
+        });
+        return;
+    }
+
     sendWithAsyncReply(Messages::WebPage::SetIsSuspended(true), [weakThis = WeakPtr { *this }](std::optional<bool> didSuspend) {
         RefPtr protectedThis = weakThis.get();
         if (!protectedThis || !didSuspend)
@@ -239,13 +252,37 @@ void SuspendedPageProxy::waitUntilReadyToUnsuspend(CompletionHandler<void(Suspen
     }
 }
 
-void SuspendedPageProxy::unsuspend()
+void SuspendedPageProxy::unsuspend(WebCore::BackForwardFrameItemIdentifier mainFrameItemID)
 {
     ASSERT(m_suspensionState == SuspensionState::Suspended);
 
     m_suspensionState = SuspensionState::Resumed;
     sendWithAsyncReply(Messages::WebPage::SetIsSuspended(false), [](std::optional<bool> didSuspend) {
         ASSERT(!didSuspend.has_value());
+    });
+
+    RefPtr page = m_page.get();
+    if (!page)
+        return;
+
+    // Notify each subframe process to restore from BFCache.
+    // Each process finds its own local frames and restores their cached pages.
+    auto aggregator = MainRunLoopSuccessCallbackAggregator::create([weakPage = m_page](bool success) {
+        if (success)
+            return;
+        RefPtr page = weakPage.get();
+        if (!page)
+            return;
+        RELEASE_LOG_ERROR(ProcessSwapping, "SuspendedPageProxy::unsuspend: subframe restoration failed, reloading page");
+        page->reload(WebCore::ReloadOption::ExpiredOnly);
+    });
+
+    m_browsingContextGroup->forEachRemotePage(*page, [suspendedPage = Ref { *this }, &aggregator, mainFrameItemID](auto& remotePage) {
+        Ref process = remotePage.siteIsolatedProcess();
+        if (!suspendedPage->hasSubframeInProcess(process->coreProcessIdentifier()))
+            return;
+        RELEASE_LOG(ProcessSwapping, "%p - SuspendedPageProxy::unsuspend: Sending RestoreWithFrameItem to pid %i", &suspendedPage, process->processID());
+        process->sendWithAsyncReply(Messages::WebPage::RestoreWithFrameItem(mainFrameItemID), aggregator->chain(), remotePage.identifierInSiteIsolatedProcess());
     });
 }
 
@@ -360,11 +397,13 @@ void SuspendedPageProxy::suspendSubframeProcesses(BackForwardFrameItemIdentifier
 
     m_browsingContextGroup->forEachRemotePage(*page, [suspendedPage = Ref { *this }, &aggregator, mainFrameItemID](auto& remotePage) {
         Ref process = remotePage.siteIsolatedProcess();
+        if (!suspendedPage->hasSubframeInProcess(process->coreProcessIdentifier()))
+            return;
         process->addSuspendedPageProxy(suspendedPage);
 
-        RELEASE_LOG(ProcessSwapping, "%p - SuspendedPageProxy::suspendSubframeProcesses: Sending SetIsSuspendedWithFrameItem to pid %i", &suspendedPage, process->processID());
+        RELEASE_LOG(ProcessSwapping, "%p - SuspendedPageProxy::suspendSubframeProcesses: Sending SuspendWithFrameItem to pid %i", &suspendedPage, process->processID());
 
-        process->sendWithAsyncReply(Messages::WebPage::SetIsSuspendedWithFrameItem(true, mainFrameItemID), aggregator->chain(), remotePage.identifierInSiteIsolatedProcess());
+        process->sendWithAsyncReply(Messages::WebPage::SuspendWithFrameItem(mainFrameItemID), aggregator->chain(), remotePage.identifierInSiteIsolatedProcess());
     });
 }
 
--- a/Source/WebKit/UIProcess/SuspendedPageProxy.h
+++ b/Source/WebKit/UIProcess/SuspendedPageProxy.h
@@ -84,7 +84,7 @@ public:
 
     void startSuspension(std::optional<WebCore::BackForwardFrameItemIdentifier>);
     void waitUntilReadyToUnsuspend(CompletionHandler<void(SuspendedPageProxy*)>&&);
-    void unsuspend();
+    void unsuspend(WebCore::BackForwardFrameItemIdentifier);
 
     bool hasSubframeInProcess(WebCore::ProcessIdentifier) const;
 
--- a/Source/WebKit/UIProcess/WebProcessPool.cpp
+++ b/Source/WebKit/UIProcess/WebProcessPool.cpp
@@ -2159,6 +2159,20 @@ void WebProcessPool::processForNavigation(WebPageProxy& page, WebFrameProxy& fra
     }
 
     if (siteIsolationEnabled && !site.isEmpty()) {
+        if (RefPtr targetItem = navigation.targetItem(); targetItem && frame.isMainFrame()) {
+            if (RefPtr suspendedPage = targetItem->suspendedPage()) {
+                Ref process = suspendedPage->process();
+                if (process->state() != AuxiliaryProcessProxy::State::Terminated) {
+                    ASSERT(isolatedProcessType != WebProcessProxy::IsolatedProcessType::Shared);
+                    prepareProcessForNavigation(WTF::move(process), page, suspendedPage.get(),
+                        "Using target back/forward item's process and suspended page"_s, isolatedProcessType,
+                        site, mainFrameSite, navigation, lockdownMode, enhancedSecurity, loadedWebArchive,
+                        WTF::move(dataStore), WTF::move(completionHandler));
+                    return;
+                }
+            }
+        }
+
         ASSERT(frameInfo.isMainFrame ? site == mainFrameSite : Site(URL(protect(page.pageLoadState())->activeURL())) == mainFrameSite);
         if (!frame.isMainFrame() && site == mainFrameSite) {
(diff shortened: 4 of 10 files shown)


Tools/TestWebKitAPI/Tests/WebKit/WKWebView/SiteIsolation.mm
```diff
@@ -8743,4 +8743,136 @@ HTTPServer server({
 }
 #endif
 
+
+TEST(SiteIsolation, MultiProcessBFCacheGoForwardSimple)
```
rr-codes, line 8838: For future tests, optionally consider using Swift Testing, which has a ton of advantages over the classic way of testing, and is much more ergonomic and versatile!

basuke: Are there any onboarding document or something? Or the prompt I should ask Claude if everything is already in the repo.

rr-codes: there is no document, that's a great idea though! I'll work on one today!

Source/WebKit/UIProcess/BrowsingContextGroup.cpp
```diff
@@ -222,6 +223,7 @@ void BrowsingContextGroup::addPage(WebPageProxy& page)
 
         if (process->process().coreProcessIdentifier() == page.legacyMainFrameProcess().coreProcessIdentifier())
             return false;
+
```
basuke: Delete this line.

Source/WebKit/UIProcess/SuspendedPageProxy.cpp
```diff
@@ -26,6 +26,7 @@
 #include "config.h"
 #include "SuspendedPageProxy.h"
 
+#include "APINavigation.h"
```
basuke, line 29: Do we need this?

Source/WebKit/UIProcess/SuspendedPageProxy.h
```diff
@@ -47,6 +47,7 @@ namespace WebKit {
 class BrowsingContextGroup;
 class RemotePageProxy;
 class WebBackForwardCache;
+class WebBackForwardListItem;
```
basuke: Do we need this?

Source/WebKit/UIProcess/WebProcessPool.cpp
```diff
@@ -2162,7 +2162,7 @@ void WebProcessPool::processForNavigation(WebPageProxy& page, WebFrameProxy& fra
         return completionHandler(sourceProcess.copyRef(), nullptr, "Navigation is treated as same-site (archive load)"_s);
     }
 
-    if (siteIsolationEnabled && !site.isEmpty()) {
+    if (siteIsolationEnabled && !site.isEmpty() && (!navigation.targetItem() || !navigation.targetItem()->suspendedPage())) {
```
szewai: The current approach introduces a priority change in process selection: before, under Site Isolation, `processForSite` always took precedence and `processForNavigationInternal` was never reached. 
Now, when the target item has a suspended page, we skip `processForSite` and fall through to `processForNavigationInternal`, which means all the earlier conditions there (opener checks, `hasOpenedPage`, etc.) now participate in the decision — even though the intent is only to let the suspended page take priority over site-based routing. 
We might extract the suspended-page lookup into a direct check (like `processFromTargetItem()`) in `processForNavigation` before the site-based routing, so we get the intended priority ("suspended page > processForSite") without inadvertently exposing the flow to unrelated conditions in `processForNavigationInternal`.

Source/WebKit/UIProcess/BrowsingContextGroup.cpp
```diff
@@ -207,7 +207,8 @@ void BrowsingContextGroup::removeFrameProcess(FrameProcess& process)
 
 void BrowsingContextGroup::addPage(WebPageProxy& page)
 {
-    ASSERT(!m_pages.contains(page));
+    if (m_pages.contains(page))
```
szewai: Can we add an assertion here so we still have some sanity check?
```
// This only happens when restoring a page from a suspended BCG, which holds exactly this one page.
ASSERT(!hasMultiplePages());
```

Source/WebKit/WebProcess/WebPage/WebPage.cpp
```diff
@@ -8634,21 +8638,28 @@ void WebPage::setIsSuspendedWithFrameItem(bool suspended, BackForwardFrameItemId
...
+        WEBPAGE_RELEASE_LOG_ERROR(ProcessSwapping, "setIsSuspendedWithFrameItem: No corePage");
+        return completionHandler(false);
+    }
+
     if (!suspended) {
-        // FIXME: Restore path (follow-up patch).
+        auto cachedPage = BackForwardCache::singleton().take(identifier, page);
+        if (!cachedPage) {
+            WEBPAGE_RELEASE_LOG_ERROR(ProcessSwapping, "setIsSuspendedWithFrameItem: take failed, cache entry missing or expired");
+            m_isSuspended = true;
```
szewai: I think we should split `setIsSuspendedWithFrameItem` into two separate functions (e.g. `suspendWithFrameItem` / `restoreWithFrameItem`) given it's no longer a simple flag setter now; and the bool parameter with bool return value is a bit confusing -- does `true` mean the flag is successfully set or the flag's current value is true.
This would also fix the transient incorrect state issue: on the restore path, `m_isSuspended` is set to `false` at the top of the function before the `take()` call, and only reset to `true` on failure. Between those two points, `m_isSuspended` is briefly `false` even though the page hasn't actually been restored yet.

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/SiteIsolation.mm
```diff
@@ -8743,4 +8743,136 @@ HTTPServer server({
...
+        EXPECT_TRUE([[webView objectByEvaluatingJavaScript:@"window.__bfcacheMarker ? true : false"] boolValue]);
+
+        [webView goForward];
+        [navigationDelegate waitForDidFinishNavigation];
+        EXPECT_WK_STREQ(@"https://b.com/b", [webView URL].absoluteString);
+        EXPECT_TRUE([[webView objectByEvaluatingJavaScript:@"window.__bfcacheMarker ? true : false"] boolValue]);
+    }
+}
+
+TEST(SiteIsolation, MultiProcessBFCacheGoForward)
```
szewai, line 8874: should we add `window.__bfcacheMarker` check too?

Source/WebKit/UIProcess/WebProcessPool.cpp
basuke: I totally agree. I've been uncomfortable with processForNavigationInternal returning {WebProcessProxy, SuspendedPageProxy*, reason} when only one path ever populates the SuspendedPageProxy*. With your proposed refactor, we can drop it from the return type entirely and clean up all the nullptr values in the remaining return statements. I think we should do this refactoring in advance to this PR.

basuke: I totally misread your comment based on my demand. That's a wrong direction. I'll do a minimum change.

Source/WebKit/UIProcess/WebProcessPool.cpp
szewai: The condition here is relaxed to include SI with `multiProcessBackForwardCacheEnabled`. This seems problematic because this block contains two paths: (1) the suspended page path, and (2) the `processForIdentifier(lastProcessIdentifier)` fallback. 

Under Site Isolation, process allocation should be based on the process's current associated site, not whether the process was previously used for the given site — so path (2) should not be enabled under SI. For path (1), since `processForNavigation` now handles the suspended page case directly before site-based routing, this code is no longer reachable under SI. 
So it seems this condition change should simply be reverted (keep the original `!siteIsolationEnabled`), unless I misunderstood the intention.

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/SiteIsolation.mm
szewai, line 8917: Consider verifying frame tree structure and iframe states to ensure iframe is also restored, like:
`checkFrameTreesInProcesses(webView.get(), { { "https://a.com"_s, { { RemoteFrame } } }, { RemoteFrame, { { "https://b.com"_s } } }});`
Can be done as followup.

szewai (approved):

PR #63278, opened 2026-04-21
Xcode Build system cleanups

brentfulgham:
#### 209240dbaa963266a6f0ff548bb5e0ffd0ccc820
<pre>
Xcode Build system cleanups
<a href="https://bugs.webkit.org/show_bug.cgi?id=312936">https://bugs.webkit.org/show_bug.cgi?id=312936</a>
<a href="https://rdar.apple.com/175291313">rdar://175291313</a>

Reviewed by Richard Robinson.

This patch resolves some script phases with missing outputs. This should help reduce
the time required to run no-change builds.

Several Xcode script phases produce warnings because they don&apos;t declare outputs, causing
them to run unconditionally on every build.

  * Add a stamp file output to libwebrtc&apos;s &quot;Check For Inappropriate Objective-C Class
    Names&quot; phase. Since it already declares the built binary as an input, the stamp file
    lets Xcode skip the check when the binary hasn&apos;t changed.

  * Add an input directory and stamp file output to WebCore&apos;s &quot;Copy modern media controls
    code and assets&quot; phase. This avoids running rsync on every build when the source
    images haven&apos;t changed.

  * Add an input directory and stamp file output to WebInspectorUI&apos;s &quot;Copy User Interface
    Resources&quot; phase. This avoids re-running the Perl copy/combine script on every build
    when the UserInterface sources haven&apos;t changed.

  * Set alwaysOutOfDate on the &quot;Unlock Keychain&quot; phases in JavaScriptCore and WebKit, as
    well as the GPU, Networking, and Model XPC service targets in WebKit. These are cheap
    scripts that don&apos;t produce files, and are only run one time per build.

  * Set alwaysOutOfDate on JavaScriptCore&apos;s &quot;Create Versions/Current/Helpers symlink&quot;
    phase. This is a cheap symlink operation that doesn&apos;t produce trackable output.

  * Set ALWAYS_SEARCH_USER_PATHS=NO in the Base.xcconfig for MiniBrowser, TestWebKitAPI
    WebKitTestRunner, DumpRenderTree, and SwiftBrowser to silence traditional headermap
    deprecation warnings across ~20 targets.

  * Add stamp file outputs to the five &quot;Update Info.plist for RunningBoard management&quot;
    phases in WebKit (WebContent, WebContent.EnhancedSecurity, WebContent.CaptivePortal,
    Networking, GPU). These already declare the Info.plist as input, but lacked outputs.

[...]

PR #63361, opened 2026-04-22
Further Xcode build improvements

brentfulgham:
#### 3d1372667d1584ab687c9f7234a439d08dbb877a
<pre>
Further Xcode build improvements
<a href="https://bugs.webkit.org/show_bug.cgi?id=313025">https://bugs.webkit.org/show_bug.cgi?id=313025</a>
<a href="https://rdar.apple.com/175367709">rdar://175367709</a>

Reviewed by Elliott Williams.

This patch is a follow-up to PR #63278 that improves Xcode build phase dependency
declarations and fixes an ImageDiff build warning.

Elliott pointed out that directory paths used as inputPaths do not introduce dependencies
on their contents, so changes to files inside those directories would not re-run the
script phases.

This patch also changes the `copy-user-interface-resources.pl` script to read from the
new UserInterface-input.xcfileslist when performing its copy/build operation. This
forces us to keep the xcfilelist up-to-date.

These steps allow the build system to recognize a no-change build for WebInspectorUI
and skip the process entirely, rather than always performing the copy/build operations.

* Source/JavaScriptCore/JavaScriptCore.xcodeproj/project.pbxproj:
    Remove alwaysOutOfDate from &quot;Create Versions/Current/Helpers symlink&quot; phase and
    declare the symlink path as an output instead.

    Change the &quot;Copy modern media controls code and assets&quot; into a PBXCopyFilesBuildPhase
    gated by platformFilters, rather than using rsync. This should allow this step to
    be skipped when files have not changed.

* Source/WebInspectorUI/Scripts/copy-user-interface-resources.pl:
(readInputFileList):
(copyFilesFromList):

* Source/WebInspectorUI/UserInterface-input.xcfilelist: Added.
    Add a new script to generate the xcfilelist representing the files that need to be
    processed to generate the UI. This will make the dependency checking step of the
    build capable of recognizing when this work can be skipped.

* Source/WebInspectorUI/WebInspectorUI.xcodeproj/project.pbxproj:
    Added a new &quot;Check .xcfilelists&quot; PBXShellScriptBuildPhase that runs when building
    against the internal SDK. This will regenerate the xcfilelist if it detects additions
    or removals from the directory.

Canonical link: <a href="https://commits.webkit.org/312305@main">https://commits.webkit.org/312305@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/38422adc48d465aceab429cc229908015c55ea8e

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/159505 "Passed style check") | [  ~~🛠 ios~~](https://ews-build.webkit.org/#/builders/159/builds/32972 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 mac~~](https://ews-build.webkit.org/#/builders/168/builds/26050 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 wpe~~](https://ews-build.webkit.org/#/builders/5/builds/168343 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/113886 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [❌ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/eb283ce4-4242-48bc-8124-66c1e84530ed) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/161374 "Passed tests") | [  ~~🛠 ios-sim~~](https://ews-build.webkit.org/#/builders/155/builds/33006 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/32925 "Built successfully") | [  ~~🧪 wpe-wk2~~](https://ews-build.webkit.org/#/builders/5/builds/168343 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/113886 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [❌ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/27fc6185-33e8-4783-8688-6f7ae05af1e9) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/162462 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/155/builds/33006 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/168/builds/26050 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-wpe~~](https://ews-build.webkit.org/#/builders/5/builds/168343 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | [❌ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/dd19202b-06cf-45c4-9ea9-e76c7a76a039) 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/155/builds/33006 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/23354 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 gtk3-libwebrtc~~](https://ews-build.webkit.org/#/builders/173/builds/16113 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | | 
| [  ~~🛠 🧪 jsc~~](https://ews-build.webkit.org/#/builders/20/builds/151561 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/155/builds/33006 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-wk1~~](https://ews-build.webkit.org/#/builders/168/builds/26050 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 gtk~~](https://ews-build.webkit.org/#/builders/2/builds/170836 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/20342 "Built successfully and passed tests") | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/16873 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-wk2~~](https://ews-build.webkit.org/#/builders/168/builds/26050 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/2/builds/170836 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [  ~~🛠 vision~~](https://ews-build.webkit.org/#/builders/153/builds/32627 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/27430 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-gtk~~](https://ews-build.webkit.org/#/builders/2/builds/170836 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35672 "Built successfully and passed tests") | [  ~~🛠 vision-sim~~](https://ews-build.webkit.org/#/builders/160/builds/32593 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-wk2-stress~~](https://ews-build.webkit.org/#/builders/168/builds/26050 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/90718 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [  ~~🧪 vision-wk2~~](https://ews-build.webkit.org/#/builders/160/builds/32593 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/168/builds/26050 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/191789 "Built successfully") | | 
| | [  ~~🛠 tv~~](https://ews-build.webkit.org/#/builders/158/builds/32145 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 mac-safer-cpp~~](https://ews-build.webkit.org/#/builders/120/builds/98541 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/49278 "Passed tests") | | 
| | [  ~~🛠 tv-sim~~](https://ews-build.webkit.org/#/builders/157/builds/31606 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [  ~~🛠 watch~~](https://ews-build.webkit.org/#/builders/163/builds/31876 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | | 
| | [  ~~🛠 watch-sim~~](https://ews-build.webkit.org/#/builders/152/builds/31793 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | |


Source/WebCore/ModernMediaControls-images-input.xcfilelist
emw-apple, line 1: This isn't an ideal fix; IMO the better thing to do would be to import the whole directory to the Xcode project and add a Copy Files phase. But it will work. Could you add a comment like:
```
// FIXME: Replace the rsync script phase that uses this file list with a native Copy Files phase.
```

Also, I think it would be worth decluttering by moving this xcfilelist to `Source/WebCore/Modules/modern-media-controls`.

emw-apple (changes requested):
re: the large xcfilelists. I don't think they are an ideal way to solve the problem, because they aren't _actually_ used to decide what to copy; they are only used as metadata to the build system. So they will go out of date very easily.

Like I said in the last PR: It would be better to import these directories to their respective Xcode projects (using a "directory reference" to avoid having to track each file individually) and add a native Copy Files build phase that copies them to the correct location. That should give the build system all the info it needs to copy incrementally as needed.

brentfulgham:
> re: the large xcfilelists. I don't think they are an ideal way to solve the problem, because they aren't _actually_ used to decide what to copy; they are only used as metadata to the build system. So they will go out of date very easily.
> 
> Like I said in the last PR: It would be better to import these directories to their respective Xcode projects (using a "directory reference" to avoid having to track each file individually) and add a native Copy Files build phase that copies them to the correct location. That should give the build system all the info it needs to copy incrementally as needed.

Oh! I misunderstood. I'll give that a try.

Source/WebCore/ModernMediaControls-images-input.xcfilelist
brentfulgham: Can a native Copy Files phase gracefully handle the different files needed on different platforms? I assume it must?

brentfulgham: This worked! Done.

Source/WebInspectorUI/WebInspectorUI.xcodeproj/project.pbxproj
```diff
@@ -202,6 +207,24 @@
...
+			);
+			name = "Check .xcfilelists";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "Scripts/check-xcfilelists.sh\n";
+		};
```
emw-apple: This script phase always runs. Is it really faster/better to run this instead of running `copy-user-interface-resources.pl` unconditionally?

The other "Check .xcfilelists" script phases in the build list a `DerivedSources.make` file as their input, since they keep a project's Makefile in sync with its filelists. But there is no analog here since your code is comparing the build system against the filesystem, so I'm not sure what to recommend.

emw-apple (changes requested):
The WebInspectorUI change is curious — I can see how it's modeled after what we do in projects with a `DerivedSources.make` file, but it's conceptually quite different, since it is syncing the build with the filesystem, instead of syncing the build to a Makefile. I'm not sure that is a good approach; at the very least it'll require us to run the checking script once per build (see below).

---

As an alternative, what if you changed `copy-user-interface-resources.pl` to _read_ the input xcfilelist, instead of doing full directory copies? That would make it impossible to go out of sync, and you could remove all this checking logic.

The xcfilelist, with path variables expanded, is available during the script's execution at `$SCRIPT_OUTPUT_FILE_LIST_0`, iirc. So you could change the perl script to read paths from that file in its `ditto()` subroutine, instead of calling ditto or `File::Copy::Recursive::dircopy`.

Source/WebInspectorUI/WebInspectorUI.xcodeproj/project.pbxproj
brentfulgham: Your idea about using the xfilelist as input to the script seems good. My hope was that Xcode could use the xcfilelist to be aware of changes to relevant files, and skip the step if no files were out of date. But that's clearly not the case since the check script has to run.

I do wonder if just pulling the ~1500 files into the project file might be the best way to ensure dependency analysis recognizes a no-action case to avoid a build. But that was the approach a few iterations ago which you also pushed back on.

I'll try the approach of using xcfilelist as input.

emw-apple: > My hope was that Xcode could use the xcfilelist to be aware of changes to relevant files, and skip the step if no files were out of date. 

It can and will do that! But as written here, the check-xcfilelists.sh script doesn't list the xcfilelist as an input (and it shouldn't—since it needs to run to detect stale xcfilelist content).

So yeah, I think finding a way to make the xcfilelist actually control what gets copied is the key.

Source/WebInspectorUI/Scripts/copy-user-interface-resources.pl
```diff
@@ -73,6 +73,37 @@ ($$)
...
+    my @files;
+    while (my $line = <$fh>) {
+        chomp $line;
+        next if $line =~ /^\s*\/\//;
+        next if $line =~ /^\s*$/;
+        $line =~ s/\$\(SRCROOT\)/$srcroot/g;
+        push @files, $line;
+    }
+    close($fh);
+    return @files;
```
emw-apple: This works, but Xcode will give you a dynamically produced xcfilelist at `$SCRIPT_INPUT_FILE_LIST_0`, with e.g. comments stripped and $(SRCROOT) expanded. So you could simplify to:


```suggestion
    open(my $fh, '<', $ENV{'SCRIPT_INPUT_FILE_LIST_0'}) or die "Unable to open SCRIPT_INPUT_FILE_LIST_0: $!";
    my @files;
    while (my $line = <$fh>) {
        chomp $line;
        push @files, $line;
    }
    close($fh);
    return @files;
```

Source/WebInspectorUI/Scripts/copy-user-interface-resources.pl
```diff
@@ -187,12 +218,13 @@ ($)
 my $shouldCombineTest = defined $ENV{'COMBINE_TEST_RESOURCES'} && ($ENV{'COMBINE_TEST_RESOURCES'} eq 'YES');
 my $shouldIncludeBrowserInspectorFrontendHost = defined $ENV{'INCLUDE_BROWSER_INSPECTOR_FRONTEND_HOST'} && ($ENV{'INCLUDE_BROWSER_INSPECTOR_FRONTEND_HOST'} eq 'YES');
 my $combineResourcesCmd = File::Spec->catfile($scriptsRoot, 'combine-resources.pl');
+my @inputFiles = readInputFileList();
```
emw-apple: This script gets run on non-Xcode builds, so maybe we should do:

```suggestion
my @inputFiles = readInputFileList() if exists $ENV{'SCRIPT_INPUT_FILE_LIST_0}';
```

Then keep the `ditto` subroutine, and have it use the xcfilelist only when it is present. Otherwise we lock the CMake build into depending on an xcfilelist, which is super weird.

(Btw, the CMake build has the same problem: the script copies from the file system which may pick up files that are not declared as input dependencies. Something to think about fixing down the road.)

Source/WebInspectorUI/Scripts/generate-xcfilelists.sh
emw-apple, line 1: I don't think we need this anymore. It's easy enough to write the xcfilelist by hand.

emw-apple (approved):
Looks great! One more idea for improvement, but I don't think I need to re-review at this point.

PR #63372, opened 2026-04-22
Build WebKit with ENABLE_USER_SCRIPT_SANDBOXING=YES

brentfulgham:
#### 8f1f5a27db3d8bfef9db2aae60b6b841cb9dd8e1
<pre>
Build WebKit with ENABLE_USER_SCRIPT_SANDBOXING=YES
<a href="https://bugs.webkit.org/show_bug.cgi?id=313032">https://bugs.webkit.org/show_bug.cgi?id=313032</a>
&lt;<a href="https://rdar.apple.com/problem/90190559">rdar://problem/90190559</a>&gt;

Reviewed by Elliott Williams.

Enable Xcode&apos;s user script sandboxing across WTF, JavaScriptCore, WebCore, WebKit,
and WebKitLegacy. When sandboxing is enabled, Xcode can run Copy Headers phases in
parallel with script phases instead of serializing them, improving build parallelism.

The sandbox restricts script phases to reading declared inputs and writing declared
outputs. This required three categories of fixes:

  1. Script phases that need broad file system access (code generation, validation
     scripts, entitlement processing) are excluded from sandboxing by name using the
     EXCLUDED_USER_SCRIPT_SANDBOXING_PHASE_NAMES build setting.

  2. Build rules that invoke `postprocess-header-rule` now declare their WebKitAdditions
     script dependencies (when present) as inputFiles, allowing the sandbox to grant
     read access.

JavaScriptCore&apos;s LLInt-related targets (libJavaScriptCore, JSCLLIntSettingsExtractor,
JSCLLIntOffsetsExtractor) require target-level sandbox overrides because their offline
assembler build rules dynamically include generated .asm files from DerivedSources
that cannot be enumerated as inputFiles.

Additionally, undeclared script dependencies were added across all projects:
  * Tools/Scripts check scripts declared as inputPaths.
  * Scripts/check-xcfilelists.sh declared as inputPaths.
  * Cryptex symlink phases declare output paths.
  * WebCore&apos;s SourcesCocoaInternalSDK.txt declared as input to &apos;Generate Unified Sources&apos;.
  * JSC&apos;s `postprocess-header-rule` declares its script in inputFiles.

This change adopts $(WK_WEBKITADDITIONS_HEADERS_FOLDER_PATH), instead of the existing
$(BUILT_PRODUCTS_DIR)$(WK_LIBRARY_HEADERS_FOLDER_PATH)/WebKitAdditions, since the former
cleanly handles our Internal and Open Source build environments so that we can safely
declare project dependencies.

Finally, WebKitLegacy&apos;s header handling is modified to match the WebKit frameworks
header processing. This allows us to opt it into the script sandbox, and to allow Xcode
to handle dependency checks to decide which files need copying.

* Configurations/CommonBase.xcconfig:
* Configurations/WebKitProjectPaths.xcconfig:
* Source/JavaScriptCore/Configurations/Base.xcconfig:
* Source/JavaScriptCore/Configurations/LLIntExtractor.xcconfig: Added.
* Source/JavaScriptCore/Configurations/libJavaScriptCore.xcconfig:
* Source/JavaScriptCore/DerivedSources-input.xcfilelist:
* Source/JavaScriptCore/DerivedSources.make:
* Source/JavaScriptCore/JavaScriptCore.xcodeproj/project.pbxproj:
* Source/WTF/Configurations/Base.xcconfig:
* Source/WebCore/Configurations/Base.xcconfig:
* Source/WebGPU/WebGPU.xcodeproj/project.pbxproj:
* Source/WebInspectorUI/Configurations/Base.xcconfig:
* Source/WebKit/Configurations/Base.xcconfig:
* Source/WebKitLegacy/WebKitLegacy.xcodeproj/project.pbxproj:
* Source/WebKitLegacy/mac/Configurations/Base.xcconfig:
* Source/WebKitLegacy/scripts/postprocess-header-rule:

Canonical link: <a href="https://commits.webkit.org/313328@main">https://commits.webkit.org/313328@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/e5a82674d8131012490e64a511cec8207717ff7c

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/162772 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/36248 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/29401 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/171710 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/119218 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/164641 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/36391 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/36252 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/126344 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/88983 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/165731 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/162/builds/28725 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/146258 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/106921 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/27771 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/26269 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/19410 "Built successfully") | | 
| [⏳ 🛠 🧪 jsc ](https://ews-build.webkit.org/#/builders/JSC-Tests-EWS "Waiting in queue, processing has not started yet") | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/137480 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/23987 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/174214 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/23611 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/21706 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/25632 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/134655 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/35922 "Built successfully") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/30371 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/134650 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36330 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/35906 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/145783 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/98320 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/29186 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/22619 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/195173 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/35416 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/103367 "Built successfully") | [  ~~🧪 jsc-armv7-tests~~](https://ews-build.webkit.org/#/builders/25/builds/50563 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/34917 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/35162 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/35065 "Built successfully") | | | |


Source/JavaScriptCore/JavaScriptCore.xcodeproj/project.pbxproj
```diff
@@ -14383,27 +14394,31 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = BC021BF2136900C300FC5467 /* ToolExecutable.xcconfig */;
 			buildSettings = {
+				ENABLE_USER_SCRIPT_SANDBOXING = NO;
```
emw-apple: Please move this (and others) to the relevant xcconfig files; we want to keep build settings from being overridden in the pbxproj because it's hard to look for.

Source/JavaScriptCore/JavaScriptCore.xcodeproj/project.pbxproj
```diff
@@ -13338,11 +13345,13 @@
 			inputFileListPaths = (
 			);
 			inputPaths = (
+				"$(SRCROOT)/Scripts/create-symlink-to-cryptex.sh",
 			);
 			name = "Create Symlink to Cryptex Path";
 			outputFileListPaths = (
 			);
 			outputPaths = (
+				"${OUTPUT_CRYPTEX_SYMLINK_PATH}",
```
emw-apple: nit: These should be Make-like variables using parentheses. I think it works either way, but please fix for consistency.

Source/WebKitLegacy/scripts/postprocess-header-rule
```diff
@@ -29,10 +29,10 @@ if [[ -z "${SCRIPT_HEADER_VISIBILITY}" ]]; then
 fi
 
 SOURCE_FILE="${SCRIPT_INPUT_FILE}"
-WORK_FILE="${SCRIPT_OUTPUT_FILE_0}.tmp"
+WORK_FILE="${DERIVED_FILE_DIR}/$(basename "${SCRIPT_OUTPUT_FILE_0}").tmp"
 DEST_FILE="${SCRIPT_OUTPUT_FILE_0}"
 
-ditto "${SOURCE_FILE}" "${WORK_FILE}"
+cp "${SOURCE_FILE}" "${WORK_FILE}"
```
emw-apple: why?

Source/WebKitLegacy/WebKitLegacy.xcodeproj/project.pbxproj
```diff
@@ -671,10 +671,14 @@
...
 			inputFiles = (
 				"$(SRCROOT)/scripts/postprocess-header-rule",
+				"$(BUILT_PRODUCTS_DIR)$(WK_LIBRARY_HEADERS_FOLDER_PATH)/WebKitAdditions/Scripts/postprocess-framework-headers-definitions",
+				"$(BUILT_PRODUCTS_DIR)$(WK_LIBRARY_HEADERS_FOLDER_PATH)/WebKitAdditions/Scripts/branch_config.json",
 			);
 			isEditable = 1;
 			outputFiles = (
 				"$(HEADER_OUTPUT_DIR)/$(INPUT_FILE_NAME)",
+				"$(DERIVED_FILE_DIR)/$(INPUT_FILE_NAME).tmp",
+				"$(DERIVED_FILE_DIR)/$(INPUT_FILE_NAME).tmp.unifdef",
```
emw-apple: It's kind of weird to reveal the .tmp files like this, because they aren't supposed to exist once the script phase has completed. So they need to be written to, but they aren't really "outputs" in the traditional sense.

Maybe you could instead try using `mktemp` / reading $TMPDIR to find a location for these files? It's annoying because we have to consider cross-filesystem moves, though.

emw-apple (commented):
How did you come up with the list for EXCLUDED_USER_SCRIPT_SANDBOXING_PHASE_NAMES?

emw-apple:
Also, this is going to have implications for the Apple Internal build, so please make sure you pay attention to Internal EWS or test it on your own.

brentfulgham:
> How did you come up with the list for EXCLUDED_USER_SCRIPT_SANDBOXING_PHASE_NAMES?

This was trial-and-error by running clean builds until I got all the way through.

Source/JavaScriptCore/JavaScriptCore.xcodeproj/project.pbxproj
brentfulgham: Got it. Will do!

aproskuryakov:
cc @geoffreygaren, who attempted sandboxing a few days ago.

Source/WebKitLegacy/scripts/postprocess-header-rule
brentfulgham: ditto creates hidden temporary files (.BC.T_* prefix with random suffix) in the destination directory as part of its atomic write strategy. The sandbox only allows writing to exact declared output paths, so these weird temp files get blocked.

This was frustrating to figure out.

cp writes directly to the target path without intermediaries. so avoided the sandbox problem. Maybe ditto has a feature flag to avoid the hidden temporary file, but I couldn't easily see it with a `ditto --help`.

Source/WebKitLegacy/WebKitLegacy.xcodeproj/project.pbxproj
brentfulgham: Yeah. I'm not sure how to make this better. If I don't include them, I get build failures.

The sandbox only allows writing to declared output paths. DERIVED_FILE_DIR isn't freely writable like it is for script phases. So the extra output declarations are needed to grants write permission to those temp file paths.

It seems like WebKit doesn't do the temporary file thing, so maybe the postprocess-header-rule in WKL could be changed to match WK?

Source/WebKitLegacy/scripts/postprocess-header-rule
aproskuryakov: According to manpage, it's supposed to do it when using the `--keepBinaries` option. I wonder if it's a bug if this happens by default.

Source/WebKitLegacy/WebKitLegacy.xcodeproj/project.pbxproj
emw-apple: > The sandbox only allows writing to declared output paths. DERIVED_FILE_DIR isn't freely writable like it is for script phases. So the extra output declarations are needed to grants write permission to those temp file paths.

My impression is that it only blocks access to paths in the build and source roots, so $TMPDIR would be free to write to, for instance.

> It seems like WebKit doesn't do the temporary file thing, so maybe the postprocess-header-rule in WKL could be changed to match WK?

Would be scary to change IMO. The point of the temporary file is to make the output file atomic, so that if the build is interrupted while it is being written, it doesn't become stuck and break the incremental build. (Keep in mind that the build system uses modification times to decide what needs to be rewritten, so a malformed output is not distinguishable from a well-formed output).

Source/JavaScriptCore/JavaScriptCore.xcodeproj/project.pbxproj
brentfulgham: Will fix.

Source/WebKitLegacy/scripts/postprocess-header-rule
brentfulgham: Interesting. The help output implies `--keepBinaries` is not present by default. I could try adding `--nonAtomicCopies` to the command.

emw-apple: > cp writes directly to the target path without intermediaries. so avoided the sandbox problem. Maybe ditto has a feature flag to avoid the hidden temporary file, but I couldn't easily see it with a `ditto --help`.

This is a feature and we probably depend on it! Suppose a build gets interrupted (because a user types Ctrl-C or a task in another lane fails). `cp` will get terminated, and if it's in the middle of a copy, that output file is now corrupted. But to the build system, it _looks_ like it was successfully build (because its modification time is newer than its source file's), so now the incremental build is broken until someone manually deleted the file.

I'm not sure what to advise. Maybe we should just get rid of atomic writes and hope that this^ issue isn't that bad in practice? Maybe we should cp to $TMPDIR (which would outside of the sandbox) and manually `mv` it ourselves?

brentfulgham: I'll back this out, as I don't believe this particular step worked well for sandboxing yet.

Configurations/CommonBase.xcconfig
```diff
@@ -28,6 +28,17 @@
...
+__ALLOW_EXCLUDED_USER_SCRIPT_SANDBOXING_PHASE_NAMES = YES;
+
+WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT = $(WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT_$(USE_INTERNAL_SDK)_$(ACTION));
+WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT_YES_ = $(WK_WEBKITADDITIONS_HEADERS_FOLDER_PATH)/Scripts/postprocess-framework-headers-definitions;
+WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT_YES_build = $(WK_WEBKITADDITIONS_HEADERS_FOLDER_PATH)/Scripts/postprocess-framework-headers-definitions;
+WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT_YES_install = $(WK_WEBKITADDITIONS_HEADERS_FOLDER_PATH)/Scripts/postprocess-framework-headers-definitions;
+WK_WEBKITADDITIONS_BRANCH_CONFIG_INPUT = $(WK_WEBKITADDITIONS_BRANCH_CONFIG_INPUT_$(USE_INTERNAL_SDK)_$(ACTION));
+WK_WEBKITADDITIONS_BRANCH_CONFIG_INPUT_YES_ = $(WK_WEBKITADDITIONS_HEADERS_FOLDER_PATH)/Scripts/branch_config.json;
+WK_WEBKITADDITIONS_BRANCH_CONFIG_INPUT_YES_build = $(WK_WEBKITADDITIONS_HEADERS_FOLDER_PATH)/Scripts/branch_config.json;
+WK_WEBKITADDITIONS_BRANCH_CONFIG_INPUT_YES_install = $(WK_WEBKITADDITIONS_HEADERS_FOLDER_PATH)/Scripts/branch_config.json;
```
emw-apple: nit: Please move these to `WebKitProjectPaths.xcconfig`; that's a better place for paths that are shared among different projects.

Configurations/CommonBase.xcconfig
```diff
@@ -28,6 +28,17 @@
 #include "WebKitProjectPaths.xcconfig"
 #include "SDKAdditions.xcconfig"
 
+__ALLOW_EXCLUDED_USER_SCRIPT_SANDBOXING_PHASE_NAMES = YES;
+
+WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT = $(WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT_$(USE_INTERNAL_SDK)_$(ACTION));
+WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT_YES_ = $(WK_WEBKITADDITIONS_HEADERS_FOLDER_PATH)/Scripts/postprocess-framework-headers-definitions;
+WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT_YES_build = $(WK_WEBKITADDITIONS_HEADERS_FOLDER_PATH)/Scripts/postprocess-framework-headers-definitions;
+WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT_YES_install = $(WK_WEBKITADDITIONS_HEADERS_FOLDER_PATH)/Scripts/postprocess-framework-headers-definitions;
```
emw-apple: re: checking $(ACTION), is the goal here to avoid the dependency edge in `installhdrs` builds?

It might be easier to use $(WK_IS_INSTALLHDRS). Then you could write:


```suggestion
WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT = $(WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT_$(WK_AND_$(USE_INTERNAL_SDK)_$(WK_IS_INSTALLHDRS)));
WK_WEBKITADDITIONS_POSTPROCESS_SCRIPTS_INPUT_YES = $(WK_WEBKITADDITIONS_HEADERS_FOLDER_PATH)/Scripts/postprocess-framework-headers-definitions;
```

Source/JavaScriptCore/DerivedSources-input.xcfilelist
emw-apple, line 1: Were these added by hand? Normally this file should be updated by adding dependencies to `DerivedSources.make` and then letting the xcfilelist generator re-run.

emw-apple (approved):

Configurations/CommonBase.xcconfig
brentfulgham: Will do!

Configurations/CommonBase.xcconfig
brentfulgham: Will do!

Source/JavaScriptCore/DerivedSources-input.xcfilelist
brentfulgham: Yes -- I will move them to `DerivedSources.make` and check in the xcfilelist after it updates.

PR #63380, opened 2026-04-23
Remove `unsafe` from use of Data in WebKit model source

mwyrzykowski:
#### 3d1cee5c55dcac580b3979588de6baddd9d088f4
<pre>
Remove `unsafe` from use of Data in WebKit model source
<a href="https://bugs.webkit.org/show_bug.cgi?id=313009">https://bugs.webkit.org/show_bug.cgi?id=313009</a>
<a href="https://rdar.apple.com/173226084">rdar://173226084</a>

Reviewed by Adrian Taylor.

Based on Adrian&apos;s patches:  <a href="https://github.com/WebKit/WebKit/compare/main...adetaylor">https://github.com/WebKit/WebKit/compare/main...adetaylor</a>:WebKit:zap-data-stuff
remove / coalesce unsafe usages.

* Source/WebKit/GPUProcess/graphics/Model/ModelBridge.swift:
(WKBridgeSkinningData.jointTransforms):
(WKBridgeSkinningData.inverseBindPoses):
(WKBridgeSkinningData.influenceJointIndices):
(WKBridgeSkinningData.influenceWeights):
* Source/WebKit/GPUProcess/graphics/Model/ModelUtils.swift:
(copyDataIntoBuffer(_:from:)):
(_Proto_LowLevelMeshResource_v1.replaceVertexData(_:)):
(_Proto_LowLevelMeshResource_v1.replaceIndexData(_:)):
* Source/WebKit/GPUProcess/graphics/Model/USDModel.swift:
(makeConstantSGNode(from:)):
(fromWKBridgeConstant(_:)):

Canonical link: <a href="https://commits.webkit.org/311970@main">https://commits.webkit.org/311970@main</a>
</pre>

### DIFF
--- a/Source/WebKit/GPUProcess/graphics/Model/ModelBridge.swift
+++ b/Source/WebKit/GPUProcess/graphics/Model/ModelBridge.swift
@@ -302,22 +302,12 @@ extension WKBridgeUpdateMesh {
 }
 
 #if ENABLE_GPU_PROCESS_MODEL && canImport(RealityCoreTextureProcessing, _version: 19)
-func decodeValues<T>(from data: Data) -> [T] {
+func decodeValues<T>(from data: Data) -> [T] where T: BitwiseCopyable {
     let stride = MemoryLayout<T>.stride
-
-    guard data.count > 0, data.count % stride == 0 else {
-        return []
-    }
-
-    return unsafe data.withUnsafeBytes { rawBufferPointer in
-        guard let baseAddress = rawBufferPointer.baseAddress else {
-            return []
-        }
-
-        let count = rawBufferPointer.count / stride
-        let pointer = unsafe baseAddress.assumingMemoryBound(to: T.self)
-        return (0..<count).map { unsafe pointer[$0] }
-    }
+    guard !data.isEmpty, data.count % stride == 0 else { return [] }
+    // rdar://164559261 - this is needed because there is no way to represnt an NSArray of
+    // primitive types in Objective-C
+    return unsafe data.withUnsafeBytes { unsafe Array(unsafe $0.bindMemory(to: T.self)) }
 }
 
 extension WKBridgeBlendShapeData {
@@ -632,10 +622,9 @@ extension WKBridgeMaterialGraph {
 #if ENABLE_GPU_PROCESS_MODEL && canImport(RealityCoreTextureProcessing, _version: 19)
 
 func toData<T>(_ input: [T]) -> Data {
-    // FIXME: (rdar://164559261) understand/document/remove unsafety
-    unsafe input.withUnsafeBytes { bufferPointer in
-        unsafe Data(bufferPointer)
-    }
+    // rdar://164559261 - this is needed because there is no way to represnt an NSArray of
+    // primitive types in Objective-C
+    unsafe input.withUnsafeBytes { unsafe Data($0) }
 }
 
 private func toDataArray<T>(_ input: [[T]]) -> [Data] {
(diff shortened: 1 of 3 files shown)


Source/WebKit/GPUProcess/graphics/Model/ModelBridge.swift
```diff
@@ -304,20 +304,8 @@ extension WKBridgeUpdateMesh {
...
-        guard let baseAddress = rawBufferPointer.baseAddress else {
-            return []
-        }
-
-        let count = rawBufferPointer.count / stride
-        let pointer = unsafe baseAddress.assumingMemoryBound(to: T.self)
-        return (0..<count).map { unsafe pointer[$0] }
-    }
+    guard !data.isEmpty, data.count % stride == 0 else { return [] }
+    return unsafe data.withUnsafeBytes { unsafe Array(unsafe $0.bindMemory(to: T.self)) }
```
adetaylor, line 310: Please can you keep/add a comment for each `unsafe` explaining why it's unavoidable and more importantly, why it's safe

Source/WebKit/GPUProcess/graphics/Model/ModelBridge.swift
```diff
@@ -632,10 +620,7 @@ extension WKBridgeMaterialGraph {
 #if ENABLE_GPU_PROCESS_MODEL && canImport(RealityCoreTextureProcessing, _version: 19)
 
 func toData<T>(_ input: [T]) -> Data {
-    // FIXME: (rdar://164559261) understand/document/remove unsafety
-    unsafe input.withUnsafeBytes { bufferPointer in
-        unsafe Data(bufferPointer)
-    }
+    unsafe input.withUnsafeBytes { unsafe Data($0) }
```
adetaylor, line 627: Ditto

Source/WebKit/GPUProcess/graphics/Model/ModelUtils.swift
```diff
@@ -108,31 +108,21 @@ extension _Proto_LowLevelMeshResource_v1.Descriptor {
     }
 }
 
+@_lifetime(buffer: copy buffer)
+private func copyDataIntoBuffer(_ buffer: inout MutableRawSpan, from data: Data) {
+    unsafe buffer.withUnsafeMutableBytes { unsafe $0.copyBytes(from: data) }
```
adetaylor, line 114: Ditto

Source/WebKit/GPUProcess/graphics/Model/ModelBridge.swift
adetaylor: Also, irrespective of the comment, I wonder if this actually _is_ memory safe for all `T`.

Since this function is not, itself, marked `@unsafe` it must be impossible for future engineers to call it with a `T` that results in a memory safety bug in here. Are there any types of `T` where this would go wrong? I think there might be some kinds of `T` which couldn't be safely reconstituted like this?

If that's so either this function needs to be marked `@unsafe` (which will prompt future engineers calling this function to think about this) or you should constrain `T`, e.g. maybe `BitwiseCopyable` is sufficient?

adetaylor (approved):
I've added comments about a couple of things that need addressing but otherwise this seems like an improvement

Source/WebKit/GPUProcess/graphics/Model/ModelBridge.swift
mwyrzykowski: `where T: BitwiseCopyable` should be sufficient, thank you

mwyrzykowski:
This causes a regression on some usd files, need to investigate

PR #63073, opened 2026-04-19
Import WPT test for sticky user activation across same-origin navigations

kate-k-lee:
#### 713dd53f52a7c6c6d881db3b458ffe09ccdbb33b
<pre>
Import WPT test for sticky user activation across same-origin navigations
<a href="https://bugs.webkit.org/show_bug.cgi?id=312241">https://bugs.webkit.org/show_bug.cgi?id=312241</a>

Reviewed by Anne van Kesteren.

Import navigate-to-sameorigin.html from upstream WPT.
The test is marked as [ Skip ] in TestExpectations because
WebKitTestRunner&apos;s eventSender cannot dispatch events to popup
windows (Bug 312719).

Upstream commit: <a href="https://github.com/web-platform-tests/wpt/commit/60af73525190d80aa07d8ca67e9e4ad2fd28982c">https://github.com/web-platform-tests/wpt/commit/60af73525190d80aa07d8ca67e9e4ad2fd28982c</a>

* LayoutTests/TestExpectations:
* LayoutTests/imported/w3c/web-platform-tests/html/user-activation/navigate-to-sameorigin.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/html/user-activation/resources/opened-window.html: Added.

Canonical link: <a href="https://commits.webkit.org/311953@main">https://commits.webkit.org/311953@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/cbcccdff4c527fca0dcc4d7dfd81f2a4dee3d508

| Misc | iOS, visionOS, tvOS & watchOS  | macOS  | Linux |  Windows |
| ----- | ---------------------- | ------- |  ----- |  --------- |
| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/158516 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/31942 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/25049 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/167346 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/112601 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
[...]

PR #63430, opened 2026-04-23
Support eventSender dispatching click events to popup windows

kate-k-lee:
#### 75e2fd9ca577e7e2d8ed354418c8817d3beeec1d
<pre>
Support eventSender dispatching click events to popup windows
<a href="https://bugs.webkit.org/show_bug.cgi?id=312719">https://bugs.webkit.org/show_bug.cgi?id=312719</a>

eventSender always dispatched click events to mainWebView(), making it
impossible to click elements in popup windows opened via window.open().
This caused WPT tests using cross-window test_driver.click() to timeout.

The fix introduces a targetView that is set based on the source page of
the script message, so click events are routed to the correct view.

* LayoutTests/imported/w3c/web-platform-tests/css/css-view-transitions/navigation/pageswap-push-from-click-expected.txt:
* LayoutTests/resources/testdriver-vendor.js:
(window.test_driver_internal.click):
* Tools/WebKitTestRunner/TestController.cpp:
(WTR::TestController::viewForPage):
(WTR::TestController::didReceiveScriptMessage):
(WTR::TestController::doAfterProcessingAllPendingMouseEvents):
* Tools/WebKitTestRunner/TestController.h:
(WTR::TestController::targetView):
(WTR::TestController::setTargetView):
* Tools/WebKitTestRunner/gtk/EventSenderProxyGtk.cpp:
(WTR::EventSenderProxy::mouseDown):
(WTR::EventSenderProxy::mouseUp):
(WTR::EventSenderProxy::mouseMoveTo):
* Tools/WebKitTestRunner/libwpe/EventSenderProxyClientLibWPE.cpp:
(WTR::viewBackend):
* Tools/WebKitTestRunner/mac/EventSenderProxy.mm:
(WTR::EventSenderProxy::mouseDown):
(WTR::EventSenderProxy::mouseUp):
(WTR::EventSenderProxy::mouseMoveTo):
* Tools/WebKitTestRunner/win/EventSenderProxyWin.cpp:
(WTR::EventSenderProxy::dispatchMessage):
* Tools/WebKitTestRunner/wpe/EventSenderProxyClientWPE.cpp:
(WTR::EventSenderProxyClientWPE::mouseDown):
(WTR::EventSenderProxyClientWPE::mouseUp):
(WTR::EventSenderProxyClientWPE::mouseMoveTo):

Canonical link: <a href="https://commits.webkit.org/312424@main">https://commits.webkit.org/312424@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/2f8cd1a4a72b431a0574d798cf770403f5270180

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/159850 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/33318 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/26424 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/168713 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/114231 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/161719 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/33422 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/33321 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/123872 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/86895 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/162808 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/26131 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/143576 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/104502 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/25183 "Passed tests") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/23656 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/16471 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/134875 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/21347 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/171203 "Built successfully") | 
| | | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/22985 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/132135 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/32996 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/27735 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/132175 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35770 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/32981 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/143141 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/91079 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/26790 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/19955 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/32490 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/31987 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/32234 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/32138 "Built successfully") | | |

### DIFF
--- a/Tools/WebKitTestRunner/TestController.cpp
+++ b/Tools/WebKitTestRunner/TestController.cpp
@@ -589,6 +589,24 @@ void TestController::closeOtherPage(WKPageRef page, PlatformWebView* view)
         m_auxiliaryWebViews.removeAt(index);
 }
 
+PlatformWebView* TestController::viewForPage(WKPageRef page)
+{
+    if (mainWebView() && mainWebView()->page() == page)
+        return mainWebView();
+    for (auto& auxiliaryWebView : m_auxiliaryWebViews) {
+        if (auxiliaryWebView->page() == page)
+            return auxiliaryWebView.ptr();
+    }
+    return mainWebView();
+}
+
+void TestController::setTargetViewFromMessage(WKScriptMessageRef message)
+{
+    auto* frameInfo = WKScriptMessageGetFrameInfo(message);
+    auto* sourcePage = frameInfo ? WKFrameInfoGetPage(frameInfo) : nullptr;
+    setTargetView(sourcePage ? viewForPage(sourcePage) : mainWebView());
+}
+
 WKPageRef TestController::createOtherPage(WKPageRef, WKPageConfigurationRef configuration, WKNavigationActionRef navigationAction, WKWindowFeaturesRef windowFeatures, const void *clientInfo)
 {
     PlatformWebView* parentView = static_cast<PlatformWebView*>(const_cast<void*>(clientInfo));
@@ -2755,13 +2773,19 @@ void TestController::didReceiveScriptMessage(WKScriptMessageRef message, Complet
         const auto y = doubleValue(argument2);
         const auto pointerType = stringValue(argument3);
 
+        // Route event to the view that sent this message (e.g. a popup window).
+        setTargetViewFromMessage(message);
+
         auto array = adoptWK(WKMutableArrayCreate());
         WKArrayAppendItem(array.get(), argument);
         WKArrayAppendItem(array.get(), argument2);
-        WKPagePostMessageToInjectedBundle(mainWebView()->page(), toWK("SetMousePosition").get(), array.get());
+        WKPagePostMessageToInjectedBundle(targetView()->page(), toWK("SetMousePosition").get(), array.get());
 
         m_eventSenderProxy->mouseMoveTo(x, y, pointerType,
-            [completionHandler = WTF::move(completionHandler)] mutable { completionHandler(nullptr); });
+            [this, completionHandler = WTF::move(completionHandler)] mutable {
+                setTargetView(nullptr);
+                completionHandler(nullptr);
+            });
         return;
     }
 
@@ -2769,8 +2793,14 @@ void TestController::didReceiveScriptMessage(WKScriptMessageRef message, Complet
         const auto button = static_cast<uint64_t>(doubleValue(argument));
         const auto array = arrayValue(argument2);
         const auto pointerType = stringValue(argument3);
+
+        setTargetViewFromMessage(message);
+
         m_eventSenderProxy->mouseDown(button, parseModifierArray(array), pointerType,
-            [completionHandler = WTF::move(completionHandler)] mutable { completionHandler(nullptr); });
+            [this, completionHandler = WTF::move(completionHandler)] mutable {
+                setTargetView(nullptr);
+                completionHandler(nullptr);
+            });
         return;
     }
 
@@ -2778,8 +2808,14 @@ void TestController::didReceiveScriptMessage(WKScriptMessageRef message, Complet
         const auto button = static_cast<uint64_t>(doubleValue(argument));
         const auto array = arrayValue(argument2);
         const auto pointerType = stringValue(argument3);
+
+        setTargetViewFromMessage(message);
+
         m_eventSenderProxy->mouseUp(button, parseModifierArray(array), pointerType,
-            [completionHandler = WTF::move(completionHandler)] mutable { completionHandler(nullptr); });
+            [this, completionHandler = WTF::move(completionHandler)] mutable {
+                setTargetView(nullptr);
+                completionHandler(nullptr);
+            });
         return;
     }
 
@@ -5773,7 +5809,7 @@ void TestController::handleAXPerformAction(WKDictionaryRef messageBody)
 void TestController::doAfterProcessingAllPendingMouseEvents(CompletionHandler<void()>&& handler)
 {
     auto* completionHandler = new CompletionHandler<void()>(WTF::move(handler));
-    WKPageDoAfterProcessingAllPendingMouseEvents(mainWebView()->page(), completionHandler, [](void* userData) {
+    WKPageDoAfterProcessingAllPendingMouseEvents(targetView()->page(), completionHandler, [](void* userData) {
         auto* completionHandler = static_cast<CompletionHandler<void()>*>(userData);
         (*completionHandler)();
         delete completionHandler;
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-view-transitions/navigation/pageswap-push-from-click-expected.txt
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-view-transitions/navigation/pageswap-push-from-click-expected.txt
@@ -1,5 +1,3 @@
 
-Harness Error (TIMEOUT), message = null
-
-TIMEOUT pageswap on navigation from user click Test timed out
+PASS pageswap on navigation from user click
 
--- a/LayoutTests/imported/w3c/web-platform-tests/html/cross-origin-opener-policy/navigate-top-to-aboutblank.https-expected.txt
+++ b/LayoutTests/imported/w3c/web-platform-tests/html/cross-origin-opener-policy/navigate-top-to-aboutblank.https-expected.txt
@@ -6,7 +6,7 @@ Opener opens a new window, to a network document with the same origin and COOP v
 Opener's iframe navigates its parent frame (opener) to about:blank.
 Verify the openee still has access to its opener.
 
-FAIL Navigate top to about:blank from iframe with opener COOP: |header(Cross-Origin-Opener-Policy,same-origin), iframe origin: https://web-platform.test:9443 assert_equals: expected "about:blank loaded" but got "about:blank NOT loaded"
-FAIL Navigate top to about:blank from iframe with opener COOP: |header(Cross-Origin-Opener-Policy,same-origin), iframe origin: https://www1.web-platform.test:9443 assert_equals: expected "about:blank loaded" but got "about:blank NOT loaded"
-FAIL Navigate top to about:blank from iframe with opener COOP: |header(Cross-Origin-Opener-Policy,same-origin-allow-popups), iframe origin: https://www1.web-platform.test:9443 assert_equals: expected "about:blank loaded" but got "about:blank NOT loaded"
+PASS Navigate top to about:blank from iframe with opener COOP: |header(Cross-Origin-Opener-Policy,same-origin), iframe origin: https://web-platform.test:9443
+PASS Navigate top to about:blank from iframe with opener COOP: |header(Cross-Origin-Opener-Policy,same-origin), iframe origin: https://www1.web-platform.test:9443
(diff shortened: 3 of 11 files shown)


aproskuryakov:
Are there security concerns here, such as perhaps with targeting settings windows in the browser which should be user only?

This PR touches a lot of shared code, could you please clarify why it's best to think of it as "[WPE][GTK]"?

kate-k-lee:
@aproskuryakov
On security: eventSender is only available in WebKitTestRunner, not in production browsers. The viewForPage() lookup is limited to m_auxiliaryWebViews, which only contains pages created via window.open() during a test. It cannot target browser-internal windows.                                                                                                               
                                                    
On the [WPE][GTK] tag: you're right, the shared code changes (TestController, testdriver-vendor.js) are platform-independent. I used [WPE][GTK] because the platform-specific EventSenderProxy changes are only for WPE and GTK.  I didn't modify the macOS or Windows EventSenderProxy since I couldn't verify them. I'm happy to drop the prefix or change it if you prefer.

Thank you.

kate-k-lee:
To clarify the platform impact: the shared code (TestController, testdriver-vendor.js) sets up the target view routing for all platforms, but the actual event dispatching change is only in the WPE and GTK EventSenderProxy clients. On macOS and Windows, the EventSenderProxy still uses mainWebView(), so WebKitTestRunner still cannot dispatch click events to popup windows on those platforms — no regression, but no improvement either. Those platforms can be updated in follow-up patches once verified.

clopez:
you need to rebaseline the -expected.txt for the test `imported/w3c/web-platform-tests/css/css-view-transitions/navigation/pageswap-push-from-click.html` in this patch for platforms gtk and wpe so it matches the new generated result (which is a pass now)
https://ews-build.s3-us-west-2.amazonaws.com/WPE-WK2-Tests-EWS/744e578d-122572/imported/w3c/web-platform-tests/css/css-view-transitions/navigation/pageswap-push-from-click-diff.txt

LayoutTests/platform/glib/imported/w3c/web-platform-tests/css/css-view-transitions/navigation/pageswap-push-from-click-expected.txt
```diff
@@ -0,0 +1,3 @@
+
+PASS pageswap on navigation from user click
+
```
clopez: Since this file is the same than the wpe one you don't need to duplicate it.
You can put it under the LayoutTests/platform/glib/ directory which is shared for gtk and wpe.

Tools/WebKitTestRunner/TestController.cpp
```diff
@@ -852,6 +863,7 @@ PlatformWebView* TestController::createOtherPlatformWebView(PlatformWebView* par
     WKPageSetPageInjectedBundleClient(newPage, &injectedBundleClient.base);
 
     view->didInitializeClients();
+    view->addToWindow();
```
clopez: What does this and why is it needed here?

kate-k-lee: Sorry, this was left over from debugging. The HeadlessViewBackend already sets the necessary activity states on creation, so this call is unnecessary. Removed it.

Tools/WebKitTestRunner/TestController.cpp
```diff
@@ -2762,31 +2772,53 @@ void TestController::didReceiveScriptMessage(WKScriptMessageRef message, Complet
...
         return;
     }
 
     if (WKStringIsEqualToUTF8CString(command, "AsyncMouseDown")) {
         const auto button = static_cast<uint64_t>(doubleValue(argument));
         const auto array = arrayValue(argument2);
         const auto pointerType = stringValue(argument3);
+
+        auto* sourcePage = WKFrameInfoGetPage(WKScriptMessageGetFrameInfo(message));
+        setTargetView(viewForPage(sourcePage));
```
clopez: In the case of `AsyncMouseMoveTo`, the return value of `WKScriptMessageGetFrameInfo(message)` is null-checked but not in this case, why is that?

Tools/WebKitTestRunner/TestController.cpp
```diff
@@ -2762,31 +2772,53 @@ void TestController::didReceiveScriptMessage(WKScriptMessageRef message, Complet
...
         return;
     }
 
     if (WKStringIsEqualToUTF8CString(command, "AsyncMouseUp")) {
         const auto button = static_cast<uint64_t>(doubleValue(argument));
         const auto array = arrayValue(argument2);
         const auto pointerType = stringValue(argument3);
+
+        auto* sourcePage = WKFrameInfoGetPage(WKScriptMessageGetFrameInfo(message));
+        setTargetView(viewForPage(sourcePage));
```
clopez: also here, the return value of `WKScriptMessageGetFrameInfo(message)` is not null checked.

Tools/WebKitTestRunner/TestController.cpp
kate-k-lee: thanks you! fixed.

Tools/WebKitTestRunner/TestController.cpp
kate-k-lee: Updated the patch to cover all platforms (macOS, Windows, WPE, and GTK). Also fixed the null-check inconsistency. This PR only addresses click events (mouseMoveTo/mouseDown/mouseUp). Support for other event types (key, touch, scroll) in popup windows will be handled in a follow-up.

clopez:
One question, in the bugzilla bug report about this you mention test `imported/w3c/web-platform-tests/html/user-activation/navigate-to-sameorigin.html` which is currently marked as Skip in `LayoutTests/TestExpectations` pointing to the same bug this PR fixes.

Does that test passes after this fix? In that case you should remove that Skip line from `LayoutTests/TestExpectations`

Otherwise, if is still broken after this fix, then please report a different bug about it and change the bug number.

Tools/WebKitTestRunner/TestController.cpp
clopez: Since this 3 lines are repeated here and on the 2 other cases below, maybe is worth to create a `setTargetViewFromMessage(message)` helper function to eliminate duplicated code and to avoid one drifting from the other in the future?

Edit: I have no idea why github doesn't render this comment on the diff, but I'm referring to the 3 lines after the comment `// Route event to the view that sent this message (e.g. a popup window)` on `Tools/WebKitTestRunner/TestController.cpp`

kate-k-lee: That test requires two fixes: 
(1) eventSender support for popup window clicks (this PR), and (2) the actual sticky activation carry-over implementation, which is tracked separately in Bug 313716. The spec (whatwg/html#11454) is not finalized yet, so only a draft PR exists for the implementation.                                                                                                                                                                          

With just this eventSender fix, the test no longer times out but still fails because the activation behavior isn't implemented yet. I'll update the Skip line to point to Bug 313716 in the test import PR (#63073), since that's where the TestExpectations entry was added.

Thanks,

kate-k-lee:
The WKPageGetScaleFactorNotZero timeout appears to be unrelated to this patch. it's a page scale factor API test, not related to eventSender or popup windows. Is this a known flaky test?

csaavedra:
> The WKPageGetScaleFactorNotZero timeout appears to be unrelated to this patch. it's a page scale factor API test, not related to eventSender or popup windows. Is this a known flaky test?

Yes, looks like it's flaky.

Tools/WebKitTestRunner/TestController.cpp
clopez: We don't usually Skip failing tests, but mark those as expected to fail or as expected to timeout or expected to crash. That is because otherwise becomes impossible tracking when a test changes behavior due to a webkit patch. Skipped tests are sent to /dev/null.

And in the specific case of the imported tests (from WPT) we don't mark those as Failure, but we commit the generated expectation with FAIL lines. The idea behind that is the same: better track when a WebKit patch makes some of the sub-test to PASS even when the whole test is still not passing.
As you can see here, your patch is updating two tests expectations that were previously passing (from the Webkit test runner PoV, the -expected.txt always contained what the test was generating), is just that the -expected.txt now contains PASS lines instead of FAIL.

So I suggest to remove that Skip line and instead commit in the WebKit repository the generated expected file, even when that one has FAIL lines.

Then the commit that would fix it would simply update the expectation to contain PASS lines like this one.

clopez (approved):

clopez:
do you want me to set merge-queue label so it lands?

kate-k-lee:
@clopez 
Thanks for the explanation about expected files — that makes a lot of sense for tracking test behavior changes. Yes, please go ahead and add the merge-queue label. I don't have commit access yet. Thanks!

PR #63476, opened 2026-04-23
[LBSE] Introduce DOM-order children cache infrastructure for SVG layers

nikolaszimmermann:
#### 315fb715d136d0057225087b703abe6bffac7a82
<pre>
[LBSE] Introduce DOM-order children cache infrastructure for SVG layers
<a href="https://bugs.webkit.org/show_bug.cgi?id=313177">https://bugs.webkit.org/show_bug.cgi?id=313177</a>

Reviewed by Simon Fraser.

Introduce the storage, invalidation, and collection of a cached, DOM-order
list of children for SVG layers. This infrastructure prepares for upcoming
SVG-specific painting and hit-testing that needs to walk children in DOM
order while interleaving layered and non-layered subtrees, and while
honoring z-index.

A new RenderLayerSVGAdditions.h header defines SVGPaintOrderLayerItem,
which classifies each entry via three factories:

 1) makeLayered: the child has its own RenderLayer; the consumer will
    delegate to the layer&apos;s own paint/hit-test machinery.
 2) makeAtomic: a transformed non-layer SVG renderer, or any leaf - painted
    in a single recursive pass at the precomputed ancestor offset.
 3) makeOutlineOnly: a non-layer container whose subtree contains
    independently painted descendants. The container is kept in the list
    so its own self-outline still paints in DOM order, while its
    descendants appear as separate entries; the container&apos;s own paint
    must therefore skip child recursion to avoid double-painting.

RenderLayer::SVGData gains the cached vector plus a dirty flag.
childrenInDOMOrderForSVG() is the lazy accessor and rebuilds via
collectChildrenInDOMOrderForSVG() when dirty. Collection walks the render
tree recursively: layered descendants and transformed non-layer descendants
are appended directly, while non-layer containers are recursed
speculatively — kept as a split makeOutlineOnly entry if the subtree contains
any independently painted descendants, otherwise collapsed into a single
makeAtomic entry. The final list is stable-sorted by z-index only when at
least one child uses a non-zero z-index, so the common case stays in DOM
order without sort overhead. Storage uses Vector::shrink(0) on
invalidation and rebuild to retain capacity across churn.

The cache is dirtied via dirtyChildrenInDOMOrderForSVG(), wired into every
site that can change the list&apos;s membership, classification, or sort order:

  - RenderLayer::dirtyPaintOrderListsOnChildChange, dirtyZOrderLists, and
    dirtyNormalFlowList — layer-children changes within an SVG ancestor.
    Note that the existing dirtyStackingContextZOrderLists() only dirties
    the stacking context ancestor, not necessarily the immediate parent,
    so an explicit dirty is required.
  - The z-index branch of RenderLayer::styleChanged — z-index drives sort
    order, so the parent layer&apos;s cache must be dirtied when it changes.
  - RenderElement::insertedIntoTree and willBeRemovedFromTree, via the
    new dirtyEnclosingLayerSVGChildrenIfNeeded() helper — non-layer SVG
    renderers do not go through RenderLayer::addChild/removeChild, so
    their insertion and removal must be intercepted at the RenderElement
    level.
  - RenderLayerModelObject::updateHasSVGTransformFlags and
    repaintOrRelayoutAfterSVGTransformChange — when isTransformed() flips
    on a non-layer renderer, its classification flips between Atomic and
    the container path, requiring a rebuild of the enclosing layer&apos;s
    cache.
  - RenderTreeBuilder::detachFromRenderElement on the early-teardown
    path, where resetRendererStateOnDetach() does not run and would
    otherwise leave stale entries pointing into the detached subtree.

The cache has no consumers yet; they follow in subsequent patches.
Covered by existing tests once conditional-layer support is enabled.

(WebCore::RenderElement::dirtyEnclosingLayerSVGChildrenIfNeeded):
(WebCore::RenderElement::insertedIntoTree):
(WebCore::RenderElement::willBeRemovedFromTree):
* Source/WebCore/rendering/RenderElement.h:
* Source/WebCore/rendering/RenderLayer.cpp:
(WebCore::RenderLayer::dirtyPaintOrderListsOnChildChange):
(WebCore::RenderLayer::dirtyZOrderLists):
(WebCore::RenderLayer::dirtyNormalFlowList):
(WebCore::RenderLayer::calculateClipRects const):
* Source/WebCore/rendering/RenderLayer.h:
* Source/WebCore/rendering/RenderLayerModelObject.cpp:
(WebCore::RenderLayerModelObject::updateHasSVGTransformFlags):
(WebCore::RenderLayerModelObject::repaintOrRelayoutAfterSVGTransformChange):
* Source/WebCore/rendering/RenderLayerSVGAdditions.cpp:
(WebCore::RenderLayer::dirtyChildrenInDOMOrderForSVG):
(WebCore::RenderLayer::collectChildrenInDOMOrderForSVG):
(WebCore::RenderLayer::appendChildrenInDOMOrderForSVG):
(WebCore::RenderLayer::childrenInDOMOrderForSVG):
* Source/WebCore/rendering/RenderLayerSVGAdditions.h: Added.
(WebCore::SVGPaintOrderLayerItem::makeLayered):
(WebCore::SVGPaintOrderLayerItem::makeAtomic):
(WebCore::SVGPaintOrderLayerItem::makeOutlineOnly):
(WebCore::SVGPaintOrderLayerItem::SVGPaintOrderLayerItem):
* Source/WebCore/rendering/updating/RenderTreeBuilder.cpp:
(WebCore::RenderTreeBuilder::detachFromRenderElement):

Canonical link: <a href="https://commits.webkit.org/312337@main">https://commits.webkit.org/312337@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/e1fabfb8b82ccb0d328e4bc4962e42ea5348ade2

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/159588 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/33055 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/26162 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/168435 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/113969 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/161457 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/33160 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/33059 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/123653 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/113969 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/162545 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/25914 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/143342 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/104305 "Passed tests") | 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/24965 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/23421 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/16200 "Built successfully") | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/134657 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/21114 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/170922 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/16956 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/22748 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/131880 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/32734 "Built successfully") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/27494 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/131966 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35715 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/32719 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/142907 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/90801 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/26609 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/19721 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/32228 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/98624 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/31725 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/31972 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/31876 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/rendering/RenderElement.cpp
+++ b/Source/WebCore/rendering/RenderElement.cpp
@@ -1195,6 +1195,15 @@ void RenderElement::styleDidChange(Style::Difference diff, const RenderStyle* ol
         LayoutIntegration::LineLayout::updateStyle(*this);
 }
 
+void RenderElement::dirtyEnclosingLayerSVGChildrenIfNeeded()
+{
+    ASSERT(isSVGLayerAwareRenderer());
+    if (!hasLayer() && document().settings().layerBasedSVGEngineEnabled()) {
+        if (CheckedPtr layer = enclosingLayer())
+            layer->dirtyChildrenInDOMOrderForSVG();
+    }
+}
+
 void RenderElement::insertedIntoTree()
 {
     // Keep our layer hierarchy updated. Optimize for the common case where we don't have any children
@@ -1210,6 +1219,11 @@ void RenderElement::insertedIntoTree()
         if (CheckedPtr parentLayer = layerParent())
             parentLayer->dirtyVisibleContentStatus();
     }
+
+    // Dirty the enclosing layer's SVG children list when a non-layer SVG renderer is inserted.
+    // Layer children are handled by RenderLayer::addChild -> dirtyPaintOrderListsOnChildChange.
+    if (isSVGLayerAwareRenderer())
+        dirtyEnclosingLayerSVGChildrenIfNeeded();
 
     RenderObject::insertedIntoTree();
 }
@@ -1227,6 +1241,12 @@ void RenderElement::willBeRemovedFromTree()
         if (CheckedPtr enclosingLayer = parent()->enclosingLayer())
             enclosingLayer->dirtyVisibleContentStatus();
     }
+
+    // Dirty the enclosing layer's SVG children list when a non-layer SVG renderer is removed.
+    // Layer children are handled by RenderLayer::removeChild -> dirtyPaintOrderListsOnChildChange.
+    if (isSVGLayerAwareRenderer())
+        dirtyEnclosingLayerSVGChildrenIfNeeded();
+
     // Keep our layer hierarchy updated.
     if (firstChild() || hasLayer())
         removeLayers();
--- a/Source/WebCore/rendering/RenderLayerModelObject.cpp
+++ b/Source/WebCore/rendering/RenderLayerModelObject.cpp
@@ -455,9 +455,20 @@ void RenderLayerModelObject::updateHasSVGTransformFlags()
 {
     ASSERT(document().settings().layerBasedSVGEngineEnabled());
 
+    bool wasTransformed = isTransformed();
     bool hasSVGTransform = needsHasSVGTransformFlags();
     setHasTransformRelatedProperty(hasSVGTransform || style().hasTransformRelatedProperty());
     setHasSVGTransform(hasSVGTransform);
+
+    // When the isTransformed() state changes, the enclosing layer's SVG children order cache
+    // must be rebuilt. A transformed non-layer renderer is collected as a single atomic entry
+    // without recursing into its subtree, so any layered descendants are not registered in
+    // the DOM-order list. A stale classification either drops those descendants (when toggled
+    // to transformed) or leaves them un-wrapped by the new transform (when toggled away).
+    if (isTransformed() != wasTransformed) {
+        if (CheckedPtr layer = enclosingLayer())
+            layer->dirtyChildrenInDOMOrderForSVG();
+    }
 }
 
 RenderSVGResourceClipper* RenderLayerModelObject::svgClipperResourceFromStyle() const
--- a/Source/WebCore/rendering/RenderLayerSVGAdditions.cpp
+++ b/Source/WebCore/rendering/RenderLayerSVGAdditions.cpp
@@ -22,13 +22,17 @@
 #include "CSSFilterRenderer.h"
 #include "ReferencedSVGResources.h"
 #include "RenderAncestorIterator.h"
+#include "RenderDescendantIterator.h"
+#include "RenderElementInlines.h"
 #include "RenderLayerFilters.h"
 #include "RenderLayerInlines.h"
 #include "RenderLayerModelObject.h"
 #include "RenderLayerSVGAdditionsInlines.h"
+#include "RenderObjectInlines.h"
 #include "RenderSVGContainer.h"
 #include "RenderSVGHiddenContainer.h"
 #include "RenderSVGModelObject.h"
+#include "RenderSVGModelObjectInlines.h"
 #include "RenderSVGResourceClipper.h"
 #include "RenderSVGResourceContainer.h"
 #include "RenderSVGRoot.h"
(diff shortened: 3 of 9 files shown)


Source/WebCore/rendering/RenderLayerSVGAdditions.cpp
```diff
@@ -176,4 +180,124 @@ void RenderLayer::updateAncestorDependentStateForSVG()
...
+            // Without the shrink, A's normal recursive paint would visit children that
+            // are still present in the list, causing the same double-paint problem.
+            //
+            // Case B — layered descendants exist: keep their entries (so they sort into
+            // the overall z-order) and append a "split" entry for the container itself.
+            // The consumer paints only the container's own non-content contribution
+            // (outlines) and skips child recursion, since each layered descendant is
+            // painted independently from the sorted list.
+            size_t startIndex = allChildren.size();
+            bool hasLayeredDescendants = self(self, child.get(), childOffset);
```
rwlbuis: The self part I find a bit confusing, for one the name changes from collectChildren to self. I wonder if a static of RenderLayer method is not better here.

Source/WebCore/rendering/RenderLayerSVGAdditions.h
```diff
@@ -0,0 +1,52 @@
...
+namespace WebCore {
+
+class RenderElement;
+class RenderLayer;
+
+struct SVGPaintOrderAwareChild {
+    CheckedPtr<RenderElement> renderer;
+    CheckedPtr<RenderLayer> layer; // null for non-layer children
+    int zIndex { 0 };
+    bool selfOutlineOnly { false };
```
rwlbuis: This does not seem used, may be better to not add it until it is used somewhere.

Source/WebCore/rendering/RenderLayerSVGAdditions.h
```diff
@@ -0,0 +1,52 @@
...
+    LayoutSize accumulatedAncestorOffset; // Precomputed offset from non-layered ancestors between child and layer's renderer.
```
rwlbuis: Ditto.

rwlbuis (commented):
some comments.

Source/WebCore/rendering/RenderLayerSVGAdditions.cpp
nikolaszimmermann: Indeed, it's not a pattern we usually follow in WebKit - I'll reorganize this.

Source/WebCore/rendering/RenderLayerSVGAdditions.h
nikolaszimmermann: The selfOutlineOnly is the "split marker" - so it's really part of the core logic, since only for outline handling we need the splitting part (otherwise the container that's split doesn't have to be part of the childrenInDOMOrder structure).

Source/WebCore/rendering/RenderLayerSVGAdditions.h
nikolaszimmermann: We're computing the ancestorOffset in the collection method, and store it in that structure?

Source/WebCore/rendering/RenderLayerSVGAdditions.h
nikolaszimmermann: I'll rename the creation method 'makeSplit' and rename to 'makeOutlineOnly'

Source/WebCore/rendering/RenderLayerSVGAdditions.h
rwlbuis: Oh, my string search did not seem to show up anything, but indeed it is stored, nm.

Source/WebCore/rendering/RenderLayerSVGAdditions.h
rwlbuis: Fair enough :) I did a simple search and did not realise it is at least being stored.

Source/WebCore/rendering/RenderLayerSVGAdditions.h
```diff
@@ -0,0 +1,65 @@
...
+#include <WebCore/PaintPhase.h>
+#include <wtf/CheckedPtr.h>
+#include <wtf/OptionSet.h>
+
+namespace WebCore {
+
+class RenderElement;
+class RenderLayer;
+
+class SVGPaintOrderAwareChild {
```
smfr: I think the name could improve; child of what? Maybe `SVGPaintOrderLayerItem` or something?

Source/WebCore/rendering/RenderElement.cpp
```diff
@@ -1211,6 +1219,10 @@ void RenderElement::insertedIntoTree()
             parentLayer->dirtyVisibleContentStatus();
     }
 
+    // Dirty the enclosing layer's SVG children list when a non-layer SVG renderer is inserted.
+    // Layer children are handled by RenderLayer::addChild -> dirtyPaintOrderListsOnChildChange.
+    dirtyEnclosingLayerSVGChildrenIfNeeded();
```
smfr: Maybe do the `if (isSVGLayerAwareRenderer())` check here?

Source/WebCore/rendering/RenderElement.cpp
```diff
@@ -1227,6 +1239,11 @@ void RenderElement::willBeRemovedFromTree()
         if (CheckedPtr enclosingLayer = parent()->enclosingLayer())
             enclosingLayer->dirtyVisibleContentStatus();
     }
+
+    // Dirty the enclosing layer's SVG children list when a non-layer SVG renderer is removed.
+    // Layer children are handled by RenderLayer::removeChild -> dirtyPaintOrderListsOnChildChange.
+    dirtyEnclosingLayerSVGChildrenIfNeeded();
```
smfr: Ditto

Source/WebCore/rendering/RenderLayerModelObject.cpp
```diff
@@ -676,8 +688,20 @@ void RenderLayerModelObject::repaintOrRelayoutAfterSVGTransformChange()
...
+    bool wasTransformed = isTransformed();
     updateHasSVGTransformFlags();
 
+    // If the isTransformed() state changed, the enclosing layer's SVG child paint order
+    // must be rebuilt. A transformed non-layer renderer is collected as a single atomic
+    // entry and its subtree is not recursed into, so layered descendants are never
+    // registered in the DOM-order list. When the transform flag toggles, the classification
+    // (atomic vs. outline-only-with-recursion) changes, and a stale list either drops
+    // layered descendants entirely or fails to wrap them in a now-required transform.
+    if (isTransformed() != wasTransformed) {
```
smfr: Share with the code above?

smfr (approved):

Source/WebCore/rendering/RenderLayerSVGAdditions.h
nikolaszimmermann: That's indeed a better name, will adapt.

Source/WebCore/rendering/RenderLayerModelObject.cpp
nikolaszimmermann: Indeed, `updateHasSVGTransformFlags` can execute the logic for both layer/non-layered case, then we don't need it here at all.

PR #63584, opened 2026-04-25
[CSS Calc] Trig functions should convert degrees to radians during evaluateDouble()

Ahmad-S792:
#### 529c5725dd7d740cc2d2d7801ede2a0cde8a709b
<pre>
[CSS Calc] Trig functions should convert degrees to radians during evaluateDouble()
<a href="https://bugs.webkit.org/show_bug.cgi?id=313300">https://bugs.webkit.org/show_bug.cgi?id=313300</a>
<a href="https://rdar.apple.com/175575617">rdar://175575617</a>

Reviewed by Sam Weinig.

The simplification path for sin()/cos()/tan() correctly calls deg2rad()
when the argument is a CanonicalDimension (angle in degrees). However,
when the argument is a complex expression that cannot be simplified at
parse time (e.g. it contains sibling-index() or other runtime-dependent
values), the trig node is left as-is in the tree. Later, evaluateDouble()
evaluates it by passing the raw double — which is in degrees — directly
to std::cos() et al., which expect radians.

This causes incorrect results for any property that resolves calc values
through the evaluateDouble() path (e.g. rotate, skew) rather than the
copyAndSimplify() path (e.g. translateX with percentage). It explains
why using @property to register custom properties as &lt;number&gt; works
around the bug: the runtime-dependent values resolve to plain numbers
before substitution, so cos() can be fully simplified at parse time
with the correct deg2rad conversion.

The fix introduces an explicit Deg2Rad conversion node that the parser
inserts inside sin(), cos() and tan() whenever the argument is an
&lt;angle&gt;. Evaluation, simplification, and serialization handle Deg2Rad
directly — Deg2Rad(CanonicalDimension(angle)) simplifies to
Number(deg2rad(value)) at parse time, Deg2Rad is transparent during
serialization, and its evaluation multiplies by the radians-per-degree
constant. This makes the degrees-to-radians conversion explicit in
the tree structure, so evaluateTrig() no longer has to inspect its
child&apos;s Type to decide whether to convert (previously the only place
outside the parser and simplification that called getType()).

For conversion to Style::Calculation::Tree (which has no Deg2Rad
node), Deg2Rad is materialized as a Product by radiansPerDegreeDouble,
preserving semantics for the rare case where a Deg2Rad survives
simplification all the way to conversion.

Test: imported/w3c/web-platform-tests/css/css-values/tree-counting/trig-functions-with-runtime-angle-arguments.html

* LayoutTests/imported/w3c/web-platform-tests/css/css-values/tree-counting/trig-functions-with-runtime-angle-arguments-expected.txt: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-values/tree-counting/trig-functions-with-runtime-angle-arguments.html: Added.
* Source/WebCore/css/calc/CSSCalcTree+Evaluation.cpp:
(WebCore::CSSCalc::evaluate):
(WebCore::CSSCalc::evaluateTrig):
* Source/WebCore/css/calc/CSSCalcTree+Parser.cpp:
(WebCore::CSSCalc::consumeExactlyOneArgument):
* Source/WebCore/css/calc/CSSCalcTree+Serialization.cpp:
(WebCore::CSSCalc::serializeCalculationTree):
(WebCore::CSSCalc::serializeMathFunction):
* Source/WebCore/css/calc/CSSCalcTree+Simplification.cpp:
(WebCore::CSSCalc::simplifyForTrig):
(WebCore::CSSCalc::simplify):
* Source/WebCore/css/calc/CSSCalcTree+Simplification.h:
* Source/WebCore/css/calc/CSSCalcTree.cpp:
(WebCore::CSSCalc::toType):
* Source/WebCore/css/calc/CSSCalcTree.h:
(WebCore::CSSCalc::get):
* Source/WebCore/style/calc/StyleCalculationTree+Conversion.cpp:
(WebCore::Style::Calculation::toStyle):

&gt; Progressions:
* LayoutTests/imported/w3c/web-platform-tests/css/css-values/acos-asin-atan-atan2-computed-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/css/css-values/sin-cos-tan-computed-expected.txt:

Canonical link: <a href="https://commits.webkit.org/312860@main">https://commits.webkit.org/312860@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/2ad99801be014cb4bbfd8f48f4c385f9f0956591

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/161167 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/34640 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/27741 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/170070 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/117538 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/10fed16b-afa5-4ca5-bf13-d6d2d97bf8b8) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/163036 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/34741 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/34641 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/125016 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/117538 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/81669e12-999b-43c7-bff4-f0fc3b8ad49c) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/164126 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/27295 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/144767 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/105613 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/bbbe7213-e421-4268-9f36-10cf9f737238) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/26343 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/24843 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/17790 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/136037 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/22528 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/172553 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/19574 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/24169 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/133294 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/34314 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/28936 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/133304 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36028 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/34298 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/144323 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/96156 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/27851 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/21141 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/33809 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/101234 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/33308 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/33552 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/33457 "Built successfully") | | | |


Source/WebCore/css/calc/CSSCalcTree+Parser.cpp
```diff
@@ -1339,6 +1339,7 @@ std::optional<TypedChild> parseCalcFunction(CSSParserTokenRange& tokens, CSSValu
             return { };
         if (state.propertyParserState.currentProperty == CSSPropertyInvalid)
             return { };
+        state.requiresConversionData = true;
```
weinig, line 1360: Can these parser changes be done in a separate PR. They seems unrelated.

Source/WebCore/css/calc/CSSCalcTree+Evaluation.cpp
```diff
@@ -193,6 +197,33 @@ std::optional<double> evaluate(const IndirectNode<Hypot>& root, const Evaluation
     return executeVariadicMathOperationAfterUnwrapping(root, options);
 }
 
+template<typename Op> static std::optional<double> evaluateTrig(const IndirectNode<Op>& root, const EvaluationOptions& options)
+{
+    auto a = evaluate(root->a, options);
+    if (!a)
+        return std::nullopt;
+
+    auto type = getType(root->a);
```
weinig: This is pretty interesting. This is, I think, the only place outside of the parser and simplification that getType() is used. It's certainly a point against the "evaluation" optimization, where the alternative would be to re-invoke simplification here. 

Perhaps a good idea would be to add an assertion in the root evaluateDouble() that asserts that the result of the evaluation matches what would happen if simplification had been used (I filed https://bugs.webkit.org/show_bug.cgi?id=313309 to track this, no need to in this PR).

I think this is ok, but I also think a solution where we make this explicit in the calc structure via either:

a) splitting the trig functions into SinRad/SinDeg, CosRad/CosDeg, TanRad/TanDeg nodes
b) adding an explicit Deg2Rad node, that gets inserted at parse time

might be better.

(I'm trying to think through if there could be a similar issue when converting to Style::CalculationTree, where instead of a sibling-*() function causing simplification of a trig function to bail, an unresolved percentage does, but I can't off the top of my head think of a way for a percentage to be a valid child of a trig() node since they only allow `<angle>` or `<number>`).

weinig (approved):

Source/WebCore/css/calc/CSSCalcTree+Parser.cpp
Ahmad-S792: Done - https://github.com/WebKit/WebKit/pull/63597 (now let me modify this PR to undo it).

Source/WebCore/css/calc/CSSCalcTree+Evaluation.cpp
Ahmad-S792: Thanks for the guidance, really appreciate it as always! I agree that using getType() in evaluation is not ideal since it's the only place outside the parser and simplification that does this.

From the two suggested approaches, I think option (b) — adding an explicit Deg2Rad node at parse time seems like the better long-term solution. But I just wanted to check, should I explore it in this PR or would you prefer to tackle it separately as a follow-up?

Ahmad-S792: Meanwhile I pushed PR to undo other suggested changes.

weinig: >  But I just wanted to check, should I explore it in this PR or would you prefer to tackle it separately as a follow-up?

No, you should go ahead with this.

Ahmad-S792: @weinig - I significantly changed based on options discussed, can you have a look once more?

weinig (approved):

PR #63690, opened 2026-04-27
SVG2 geometry presentation attributes accepted on wrong elements such as g

karlcow:
#### 9db42a8aaf4a56d40c3642fbad90599174952696
<pre>
SVG2 geometry presentation attributes accepted on wrong elements such as g
<a href="https://bugs.webkit.org/show_bug.cgi?id=313380">https://bugs.webkit.org/show_bug.cgi?id=313380</a>
<a href="https://rdar.apple.com/175672111">rdar://175672111</a>

WebKit treated the SVG2 geometry attributes (cx, cy, r, rx, ry,
x, y, width, height, d) as CSS presentation attributes on every
SVG element. Per SVG2 each is a presentation attribute only on
its owning elements; elsewhere getComputedStyle must return the
initial value.

Fold the element-restriction check inline into
SVGElement::cssPropertyIdForSVGAttributeName: each geometry case
returns its CSS property only when *this is an owning element
type, otherwise CSSPropertyInvalid. The mapping stays non-virtual.

* cx, cy: circle, ellipse
* r: circle
* rx, ry: ellipse, rect
* x, y: foreignObject, image, rect, svg, symbol, use
* width, height: foreignObject, image, rect, outermost svg only
* d: path (gated by Settings::cssDPropertyEnabled)

Drop now-unnecessary setPresentationalHintStyleIsDirty() from
SVGFilterElement, SVGMaskElement, SVGPatternElement. Refactor
parse-length.html / parse-calc-length.html to use spec-correct
elements (circle for cx/cy/r).

Spec: <a href="https://w3c.github.io/svgwg/svg2-draft/geometry.html">https://w3c.github.io/svgwg/svg2-draft/geometry.html</a>

* LayoutTests/imported/w3c/web-platform-tests/svg/geometry/parsing/height-computed-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/svg/geometry/parsing/width-computed-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/presentation-attributes-special-cases-expected.txt:
* LayoutTests/imported/w3c/web-platform-tests/svg/styling/presentation-attributes-special-cases.html:
* LayoutTests/platform/mac-tahoe-wk2-lbse-text/imported/w3c/web-platform-tests/svg/geometry/parsing/height-computed-expected.txt:
* LayoutTests/platform/mac-tahoe-wk2-lbse-text/imported/w3c/web-platform-tests/svg/geometry/parsing/width-computed-expected.txt:
* LayoutTests/svg/css/parse-calc-length-expected.txt:
* LayoutTests/svg/css/parse-calc-length.html:
* LayoutTests/svg/css/parse-length-expected.txt:
* LayoutTests/svg/css/parse-length.html:
(WebCore::SVGElement::cssPropertyIdForSVGAttributeName const):
(WebCore::SVGElement::hasPresentationalHintsForAttribute const):
(WebCore::SVGElement::collectPresentationalHintsForAttribute):
(WebCore::SVGElement::svgAttributeChanged):
(WebCore::SVGElement::cssPropertyIdForSVGAttributeName): Deleted.
* Source/WebCore/svg/SVGElement.h:
* Source/WebCore/svg/SVGFilterElement.cpp:
(WebCore::SVGFilterElement::svgAttributeChanged):
* Source/WebCore/svg/SVGFontFaceElement.cpp:
(WebCore::SVGFontFaceElement::attributeChanged):
* Source/WebCore/svg/SVGMaskElement.cpp:
(WebCore::SVGMaskElement::svgAttributeChanged):
(WebCore::SVGPathElement::collectDPresentationalHint):
* Source/WebCore/svg/SVGPatternElement.cpp:
(WebCore::SVGPatternElement::svgAttributeChanged):

Canonical link: <a href="https://commits.webkit.org/315946@main">https://commits.webkit.org/315946@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/0dddf141cf0f8bf49ffed986d0592798651ed5cc

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/170421 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/44344 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/37761 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/179796 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/128133 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b63be64f-ab00-4951-ab5e-7ce2608156f9) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/172287 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/45589 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/43976 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/132892 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/93689 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/716daa83-504a-4a36-a62e-d20d4e3bbeed) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/173380 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/36070 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/153322 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/113392 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/110c396e-383b-4f34-bbfe-92f33b2685bd) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/34694 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/33029 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/28479 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/145154 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/32720 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/182381 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/35184 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/37197 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/141342 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/44001 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/152792 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/141160 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/38039 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/44690 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/29708 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/109161 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/36428 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/116585 "Built successfully") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/43200 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/7808 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/42606 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/42846 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/42735 "Built successfully") | | | |


Source/WebCore/svg/SVGElement.cpp
```diff
@@ -1014,10 +1014,56 @@ CSSPropertyID SVGElement::cssPropertyIdForSVGAttributeName(const QualifiedName&
...
+        case SVG::circle:
+        case SVG::ellipse:
+            return true;
+        default:
+            return false;
+        }
+    case CSSPropertyR:
+        return element.elementName() == SVG::circle;
+    case CSSPropertyRx:
+    case CSSPropertyRy:
```
rwlbuis: These can be combined with cx,cy.

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -1014,10 +1014,56 @@ CSSPropertyID SVGElement::cssPropertyIdForSVGAttributeName(const QualifiedName&
...
+    switch (propertyID) {
+    case CSSPropertyCx:
+    case CSSPropertyCy:
+        switch (element.elementName()) {
+        case SVG::circle:
+        case SVG::ellipse:
+            return true;
+        default:
+            return false;
+        }
```
rwlbuis: Why not do return is\<SVGCircleElement\>(element) || is\<SVGEllipseElement\>(element); style?

rwlbuis (commented):
Some comments.

Source/WebCore/svg/SVGElement.cpp
karlcow: hmm… but
* cx, cy -> circle + ellipse
* rx, ry ->  rect + ellipse

So they can't really be merged.

rwlbuis: Ah, nm, I misread.

Source/WebCore/svg/SVGElement.cpp
karlcow: I was thinking about perf. With `is<>`

The code would look like this. 

```cpp
static bool isApplicableGeometryPresentationAttribute(CSSPropertyID propertyID, const SVGElement& element)
{
    switch (propertyID) {
    case CSSPropertyCx:
    case CSSPropertyCy:
        return is<SVGCircleElement>(element) || is<SVGEllipseElement>(element);
    case CSSPropertyR:
        return is<SVGCircleElement>(element);
    case CSSPropertyRx:
    case CSSPropertyRy:
        return is<SVGEllipseElement>(element) || is<SVGRectElement>(element);
    case CSSPropertyX:
    case CSSPropertyY:
    case CSSPropertyWidth:
    case CSSPropertyHeight:
        return is<SVGForeignObjectElement>(element) || is<SVGImageElement>(element)
            || is<SVGRectElement>(element) || is<SVGSVGElement>(element)
            || is<SVGSymbolElement>(element) || is<SVGUseElement>(element);
    default:
        return true;
    }
}
```

For the x/y/width/height case, that's 6 calls evaluated left to right until one matches. With switch it is only one jump. But maybe it doesn't matter.

nullhook:
This change only filters the collection pass of the presentational hint style infrastructure. However, there are elements (e.g `<mask>`) that still dirty the presentational hint style via `setPresentationalHintStyleIsDirty()`, which will trigger a full style-resolution pass even though `collectPresentationalHintsForAttribute` will now reject their CSS side of geometry attributes.

Can we get @weinig's opinion?

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -1014,10 +1014,56 @@ CSSPropertyID SVGElement::cssPropertyIdForSVGAttributeName(const QualifiedName&
     return CSSPropertyInvalid;
 }
 
-bool SVGElement::hasPresentationalHintsForAttribute(const QualifiedName& name) const
+static bool isApplicableGeometryPresentationAttribute(CSSPropertyID propertyID, const SVGElement& element)
```
nullhook: the function name looks misleading. it returns `true` for all non-geometry properties via `default: return true`

karlcow: hmmm… :/ indeed not good. Let me refactor this. I also noticed some missing cases.

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -1014,10 +1014,68 @@ CSSPropertyID SVGElement::cssPropertyIdForSVGAttributeName(const QualifiedName&
...
+        }
+    case CSSPropertyWidth:
+    case CSSPropertyHeight:
+        switch (element.elementName()) {
+        case SVG::foreignObject:
+        case SVG::image:
+        case SVG::rect:
+            return true;
+        case SVG::svg:
+            return element.isOutermostSVGSVGElement();
```
karlcow: This is one of the constraints of the SVG spec.

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -1014,10 +1014,68 @@ CSSPropertyID SVGElement::cssPropertyIdForSVGAttributeName(const QualifiedName&
     return CSSPropertyInvalid;
 }
 
-bool SVGElement::hasPresentationalHintsForAttribute(const QualifiedName& name) const
+static bool isApplicableSVGPresentationAttribute(CSSPropertyID propertyID, const SVGElement& element)
```
shallawa: I think the name of this function should be `isApplicableCSSProperty()` or something similar. It does not take an attribute to have `SVGPresentationAttribute` in its name.

shallawa: Also in SVGElement we sometimes enumerate the types of the SVGElements instead of adding a new virtual method, like this one you are adding here. I think it makes more sense to add this as a virtual method and let each type decide what CSS attribute they can support:
```
bool SVGCircleElement::isApplicableCSSProperty(CSSPropertyID propertyID)
{
    return propertyID == CSSPropertyCx || propertyID == CSSPropertyCy || propertyID == CSSPropertyR;
}
```

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -1014,10 +1014,68 @@ CSSPropertyID SVGElement::cssPropertyIdForSVGAttributeName(const QualifiedName&
...
+    default:
         return true;
+    }
+}
+
+bool SVGElement::hasPresentationalHintsForAttribute(const QualifiedName& name) const
+{
+    CSSPropertyID propertyID = cssPropertyIdForSVGAttributeName(name, document().settings());
+    if (propertyID > 0)
+        return isApplicableSVGPresentationAttribute(propertyID, *this);
```
shallawa: I think the check of `isApplicableSVGPresentationAttribute()` should be included inside `cssPropertyIdForSVGAttributeName()`. `cssPropertyIdForSVGAttributeName()` should return `CSSPropertyInvalid` if the element does not support the CSS attribute.

If this is done this function will not be changed.

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -1026,9 +1084,10 @@ bool SVGElement::hasPresentationalHintsForAttribute(const QualifiedName& name) c
 void SVGElement::collectPresentationalHintsForAttribute(const QualifiedName& name, const AtomString& value, MutableStyleProperties& style)
 {
     CSSPropertyID propertyID = cssPropertyIdForSVGAttributeName(name, document().settings());
-    if (propertyID > 0)
-        addPropertyToPresentationalHintStyle(style, propertyID, value);
-    else if (name.matches(XMLNames::langAttr) || (name.matches(HTMLNames::langAttr) && !hasAttributeWithoutSynchronization(XMLNames::langAttr)))
+    if (propertyID > 0) {
+        if (isApplicableSVGPresentationAttribute(propertyID, *this))
+            addPropertyToPresentationalHintStyle(style, propertyID, value);
+    } else if (name.matches(XMLNames::langAttr) || (name.matches(HTMLNames::langAttr) && !hasAttributeWithoutSynchronization(XMLNames::langAttr)))
```
shallawa: Ditto.

Source/WebCore/svg/SVGElement.cpp
karlcow: @shallawa 
* `cssPropertyIdForSVGAttributeName` is currently static.  mapping the SVG attribute name with the CSS Property ID.
* `isApplicableSVGPresentationAttribute` (the patch) tries to answer the question is this a valid presentation attribute for this element.  

so if I understand your latest comment you are proposing to move the logic to each individual element.

We would need to take into account all the properties which apply to *Any element in the SVG namespace*:
see https://w3c.github.io/svgwg/svg2-draft/styling.html#TermPresentationAttribute

[alignment-baseline](https://w3c.github.io/svgwg/svg2-draft/text.html#AlignmentBaselineProperty), [baseline-shift](https://w3c.github.io/svgwg/svg2-draft/text.html#BaselineShiftProperty), [clip-path](https://drafts.fxtf.org/css-masking-1/#the-clip-path), [clip-rule](https://drafts.fxtf.org/css-masking-1/#the-clip-rule), [color](https://w3c.github.io/svgwg/svg2-draft/painting.html#ColorProperty), [color-interpolation](https://w3c.github.io/svgwg/svg2-draft/painting.html#ColorInterpolationProperty), [color-interpolation-filters](https://drafts.fxtf.org/filter-effects/#ColorInterpolationFiltersProperty), [cursor](https://www.w3.org/TR/css-ui-3/#cursor), [direction](https://w3c.github.io/svgwg/svg2-draft/text.html#DirectionProperty), [display](https://w3c.github.io/svgwg/svg2-draft/render.html#VisibilityControl), [dominant-baseline](https://w3c.github.io/svgwg/svg2-draft/text.html#DominantBaselineProperty), [fill-opacity](https://w3c.github.io/svgwg/svg2-draft/painting.html#FillOpacityProperty), [fill-rule](https://w3c.github.io/svgwg/svg2-draft/painting.html#FillRuleProperty), [filter](https://drafts.fxtf.org/filter-effects/#FilterProperty), [flood-color](https://drafts.fxtf.org/filter-effects/#FloodColorProperty), [flood-opacity](https://drafts.fxtf.org/filter-effects/#FloodOpacityProperty), [font-family](https://www.w3.org/TR/css-fonts-3/#font-family-prop), [font-size](https://www.w3.org/TR/css-fonts-3/#font-size-prop), [font-size-adjust](https://www.w3.org/TR/css-fonts-3/#font-size-adjust-prop), [font-stretch](https://www.w3.org/TR/css-fonts-3/#font-stretch-prop), [font-style](https://www.w3.org/TR/css-fonts-3/#font-style-prop), [font-variant](https://www.w3.org/TR/css-fonts-3/#propdef-font-variant), [font-weight](https://www.w3.org/TR/css-fonts-3/#font-weight-prop), [glyph-orientation-vertical](https://w3c.github.io/svgwg/svg2-draft/text.html#GlyphOrientationVerticalProperty), [image-rendering](https://w3c.github.io/svgwg/svg2-draft/painting.html#ImageRenderingProperty), [letter-spacing](https://www.w3.org/TR/css-text-3/#letter-spacing-property), [lighting-color](https://drafts.fxtf.org/filter-effects/#LightingColorProperty), [marker-end](https://w3c.github.io/svgwg/svg2-draft/painting.html#MarkerEndProperty), [marker-mid](https://w3c.github.io/svgwg/svg2-draft/painting.html#MarkerMidProperty), [marker-start](https://w3c.github.io/svgwg/svg2-draft/painting.html#MarkerStartProperty), [mask](https://drafts.fxtf.org/css-masking-1/#propdef-mask), [mask-type](https://drafts.fxtf.org/css-masking-1/#propdef-mask-type), [opacity](https://w3c.github.io/svgwg/svg2-draft/render.html#ObjectAndGroupOpacityProperties), [overflow](https://w3c.github.io/svgwg/svg2-draft/render.html#OverflowAndClipProperties), [paint-order](https://w3c.github.io/svgwg/svg2-draft/painting.html#PaintOrderProperty), [pointer-events](https://w3c.github.io/svgwg/svg2-draft/interact.html#PointerEventsProperty), [shape-rendering](https://w3c.github.io/svgwg/svg2-draft/painting.html#ShapeRenderingProperty), [stop-color](https://w3c.github.io/svgwg/svg2-draft/pservers.html#StopColorProperty), [stop-opacity](https://w3c.github.io/svgwg/svg2-draft/pservers.html#StopOpacityProperty), [stroke](https://w3c.github.io/svgwg/svg2-draft/painting.html#StrokeProperty), [stroke-dasharray](https://w3c.github.io/svgwg/svg2-draft/painting.html#StrokeDasharrayProperty), [stroke-dashoffset](https://w3c.github.io/svgwg/svg2-draft/painting.html#StrokeDashoffsetProperty), [stroke-linecap](https://w3c.github.io/svgwg/svg2-draft/painting.html#StrokeLinecapProperty), [stroke-linejoin](https://w3c.github.io/svgwg/svg2-draft/painting.html#StrokeLinejoinProperty), [stroke-miterlimit](https://w3c.github.io/svgwg/svg2-draft/painting.html#StrokeMiterlimitProperty), [stroke-opacity](https://w3c.github.io/svgwg/svg2-draft/painting.html#StrokeOpacityProperty), [stroke-width](https://w3c.github.io/svgwg/svg2-draft/painting.html#StrokeWidthProperty), [text-anchor](https://w3c.github.io/svgwg/svg2-draft/text.html#TextAnchorProperty), [text-decoration](https://w3c.github.io/svgwg/svg2-draft/text.html#TextDecorationProperties), [text-overflow](https://w3c.github.io/svgwg/svg2-draft/text.html#TextOverflowProperty), [text-rendering](https://w3c.github.io/svgwg/svg2-draft/painting.html#TextRenderingProperty), [transform-origin](https://drafts.csswg.org/css-transforms/#transform-origin-property), [unicode-bidi](https://www.w3.org/TR/css-writing-modes-3/#unicode-bidi), [vector-effect](https://w3c.github.io/svgwg/svg2-draft/coords.html#VectorEffectProperty), [visibility](https://w3c.github.io/svgwg/svg2-draft/render.html#VisibilityControl), [white-space](https://www.w3.org/TR/css-text-3/#white-space-property), [word-spacing](https://www.w3.org/TR/css-text-3/#word-spacing-property), [writing-mode](https://w3c.github.io/svgwg/svg2-draft/text.html#WritingModeProperty)

That's the `default: return true` in the current code. So if we go the subclass route, the code would look like something like:

```cpp
  bool SVGCircleElement::isApplicableCSSProperty(CSSPropertyID id) {                                                                                                                     
      switch (id) {                                                              
      case CSSPropertyCx:                                                                                                                                                                
      case CSSPropertyCy:
      case CSSPropertyR:                                                                                                                                                                 
          return true;                                                           
      default:                                                                                                                                                                           
          return SVGElement::isApplicableCSSProperty(id);                        
      }
  }  
```

@shallawa What do you think?

karlcow: It also means that we would have to reject all geometry properties in `SVGElement.cpp`?

```cpp
  bool SVGElement::isApplicableCSSProperty(CSSPropertyID propertyID) const                                                                                                               
  {                                                                                                                                                                                      
      switch (propertyID) {                                                                                                                                                              
      case CSSPropertyCx:                                                        
      case CSSPropertyCy:                                                                                                                                                                
      case CSSPropertyR:                                                                                                                                                                 
      case CSSPropertyRx:                                                                                                                                                                
      case CSSPropertyRy:                                                                                                                                                                
      case CSSPropertyX:                                                                                                                                                                 
      case CSSPropertyY:                                                         
      case CSSPropertyWidth:                                                                                                                                                             
      case CSSPropertyHeight:                                                    
      case CSSPropertyD:                                                                                                                                                                 
          return false;                                                                                                                                                                  
      default:                                                                                                                                                                           
          return true;                                                                                                                                                                   
      }                                                                          
  } 
```

karlcow: If we fold everything into `cssPropertyIdForSVGAttributeName`, the first alternate proposal. So no distribution, we would change the scope of this function by not doing only mapping but also a validation phase for the applicability.

🤔

karlcow: searching for parallels. `patternTransform` and `gradientTransform` are not handled as presentation attributes at all. They are handled by their own elements.  So that makes a precedent. We could require each element to handle its own geometry attributes.

karlcow: Still thinking… 
we could remove the geometry attributes from `cssPropertyIdForSVGAttributeName` and just deal with them at Element level.

shallawa: This will work. But you have to list all the other CSSPropertyIDs in `SVGElement::isApplicableCSSProperty()`

What I meant was to change the cases which are not common to all SVGElements like `cx`, `cy`, `r,` etc in `cssPropertyIdForSVGAttributeName()` like this:

```
bool SVGCircleElement::isApplicableCSSProperty(CSSPropertyID propertyID)
{
    return propertyID == CSSPropertyCx || propertyID == CSSPropertyCy || propertyID == CSSPropertyR;
}

CSSPropertyID SVGElement::cssPropertyIdForSVGAttributeName(const QualifiedName& attrName, const Settings& settings)
{
    if (!attrName.namespaceURI().isNull())
        return CSSPropertyInvalid;

    switch (attrName.nodeName()) {
    ...
    case AttributeNames::colorAttr:
        return CSSPropertyColor;
    ...
    case AttributeNames::cxAttr:
        return isApplicableCSSProperty(CSSPropertyCx) ? CSSPropertyCx : CSSPropertyInvalid;
    case AttributeNames::cyAttr:
        return isApplicableCSSProperty(CSSPropertyCy) ? CSSPropertyCy : CSSPropertyInvalid;
    ...
    case AttributeNames::rAttr:
        return isApplicableCSSProperty(CSSPropertyR) ? CSSPropertyR : CSSPropertyInvalid;
    ...
    default:
        break;
    }

    return CSSPropertyInvalid;
}
```

So `cssPropertyIdForSVGAttributeName()` will return `CSSPropertyInvalid` for `cxAttr`, `cyAttr`, `rAttr` unless the element is `SVGCircleElement`

karlcow: We could do for example

```cpp
  bool SVGCircleElement::hasPresentationalHintsForAttribute(const QualifiedName& name) const
  {                                                                                                                                                                                      
      if (name == SVGNames::cxAttr || name == SVGNames::cyAttr || name == SVGNames::rAttr)                                                                                               
          return true;                                                                                                                                                                   
      return SVGGeometryElement::hasPresentationalHintsForAttribute(name);                                                                                                               
  }                                                                                                                                                                                      
                                                                                                                                                                                         
  void SVGCircleElement::collectPresentationalHintsForAttribute(const QualifiedName& name, const AtomString& value, MutableStyleProperties& style)                                       
  {                                                                                                                                                                                      
      if (name == SVGNames::cxAttr)                                              
          addPropertyToPresentationalHintStyle(style, CSSPropertyCx, value);                                                                                                             
      else if (name == SVGNames::cyAttr)                                                                                                                                                 
          addPropertyToPresentationalHintStyle(style, CSSPropertyCy, value);                                                                                                             
      else if (name == SVGNames::rAttr)                                                                                                                                                  
          addPropertyToPresentationalHintStyle(style, CSSPropertyR, value);                                                                                                              
      else                                                                                                                                                                               
          SVGGeometryElement::collectPresentationalHintsForAttribute(name, value, style);                                                                                                
  }                                                                                                                                                                                      
```

Source/WebCore/svg/SVGElement.h
```diff
@@ -205,7 +205,8 @@ class SVGElement : public StyledElement, public SVGPropertyOwner {
     SVGElementRareData& ensureSVGRareData();
 
     void reportAttributeParsingError(SVGParsingError, const QualifiedName&, const AtomString&);
-    static CSSPropertyID NODELETE cssPropertyIdForSVGAttributeName(const QualifiedName&, const Settings&);
+    CSSPropertyID cssPropertyIdForSVGAttributeName(const QualifiedName&) const;
+    virtual bool isApplicableCSSProperty(CSSPropertyID) const { return false; }
```
weinig: What is your thinking behind adding a virtual `isApplicableCSSProperty` and not just making `cssPropertyIdForSVGAttributeName` virtual?

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -883,12 +883,12 @@ CSSPropertyID SVGElement::cssPropertyIdForSVGAttributeName(const QualifiedName&
...
     case AttributeNames::cxAttr:
-        return CSSPropertyCx;
+        return isApplicableCSSProperty(CSSPropertyCx) ? CSSPropertyCx : CSSPropertyInvalid;
     case AttributeNames::cyAttr:
-        return CSSPropertyCy;
+        return isApplicableCSSProperty(CSSPropertyCy) ? CSSPropertyCy : CSSPropertyInvalid;
     case AttributeNames::dAttr:
-        if (settings.cssDPropertyEnabled())
-            return CSSPropertyD;
+        if (document().settings().cssDPropertyEnabled())
```
shallawa: I think this condition can be moved to `SVGPathElement`.

Source/WebCore/svg/SVGElement.h
karlcow: Part of it was the maintenance burden. 
If you need to add a new presentation attribute, you then to find and modify all the elements it applies to. 

**But** maybe that doesn't really matter given that it's a rare event to add a new presentation attribute, and it is mostly discouraged. I could try a solution with making `cssPropertyIdForSVGAttributeName` virtual 🤔

karlcow: @weinig ok I followed your approach.

Source/WebCore/svg/SVGElement.cpp
karlcow: @shallawa It has been moved.

nikolaszimmermann (commented):
It's a correct fix, and looks tempting, however making this virtual could be a perf-hit - you need some A/B testing and checks if that hurts perf or not. Otherwise we need to find other means to fix this.

karlcow:
haha!  Now I feel the torture 😆 

1. @weinig says go `virtual` 
2. @nikolaszimmermann says avoid `virtual` because maybe perf issues.

The not virtual was the previous proposal, which I just undid.

karlcow:
> haha! Now I feel the torture 😆
> 
> 1. @weinig says go `virtual`
> 2. @nikolaszimmermann says avoid `virtual` because maybe perf issues.
> 
> The not virtual was the previous proposal, which I just undid.

ok sam on another channel is saying, go with Nikolas. Let me workout a new solution. :) Thanks!

karlcow:
ok let's try again.

karlcow:
@nikolaszimmermann last comment about this? Is it in good shapes for you? Or are there a couple of things I should change?

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -884,13 +895,17 @@ CSSPropertyID SVGElement::cssPropertyIdForSVGAttributeName(const QualifiedName&
     case AttributeNames::cursorAttr:
         return CSSPropertyCursor;
     case AttributeNames::cxAttr:
-        return CSSPropertyCx;
+        if (is<SVGCircleElement>(*this) || is<SVGEllipseElement>(*this))
```
nikolaszimmermann: While all these if statements look a bit ugly, it's likely more performant in this way.

Source/WebCore/svg/SVGElement.h
```diff
@@ -209,7 +209,7 @@ class SVGElement : public StyledElement, public SVGPropertyOwner {
     SVGElementRareData& ensureSVGRareData();
 
     void reportAttributeParsingError(SVGParsingError, const QualifiedName&, const AtomString&);
-    static CSSPropertyID NODELETE cssPropertyIdForSVGAttributeName(const QualifiedName&, const Settings&);
```
nikolaszimmermann, line 212: I'm not fluent in NODELETE - but is there a reason we no longer need it?

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -857,8 +862,14 @@ bool SVGElement::rendererIsNeeded(const Style::ComputedStyle& style)
     return false;
 }
 
-CSSPropertyID SVGElement::cssPropertyIdForSVGAttributeName(const QualifiedName& attrName, const Settings& settings)
+CSSPropertyID SVGElement::cssPropertyIdForSVGAttributeName(const QualifiedName& attrName) const
```
nikolaszimmermann, line 865:
```
CSSPropertyID SVGElement::cssPropertyIdForSVGAttributeName(const QualifiedName& attrName) const
  {
      if (!attrName.namespaceURI().isNull())
          return CSSPropertyInvalid;

      // The SVG2 geometry presentation attributes apply only on their owning
      // element types. https://w3c.github.io/svgwg/svg2-draft/geometry.html
      auto supportsXY = [&] {
          return is<SVGForeignObjectElement>(*this) || is<SVGImageElement>(*this) || is<SVGRectElement>(*this)
              || is<SVGSVGElement>(*this) || is<SVGSymbolElement>(*this) || is<SVGUseElement>(*this);
      };
      auto supportsWidthHeight = [&] {
          if (is<SVGForeignObjectElement>(*this) || is<SVGImageElement>(*this) || is<SVGRectElement>(*this))
              return true;
          auto* svg = dynamicDowncast<SVGSVGElement>(*this);
          return svg && svg->isOutermostSVGSVGElement();
      };

      switch (attrName.nodeName()) {
      case AttributeNames::alignment_baselineAttr:
          return CSSPropertyAlignmentBaseline;
      case AttributeNames::baseline_shiftAttr:
          return CSSPropertyBaselineShift;
      case AttributeNames::buffered_renderingAttr:
          return CSSPropertyBufferedRendering;
      case AttributeNames::clipAttr:
          return CSSPropertyClip;
      case AttributeNames::clip_pathAttr:
          return CSSPropertyClipPath;
      case AttributeNames::clip_ruleAttr:
          return CSSPropertyClipRule;
      case AttributeNames::colorAttr:
          return CSSPropertyColor;
      case AttributeNames::color_interpolationAttr:
          return CSSPropertyColorInterpolation;
      case AttributeNames::color_interpolation_filtersAttr:
          return CSSPropertyColorInterpolationFilters;
      case AttributeNames::cursorAttr:
          return CSSPropertyCursor;
      case AttributeNames::cxAttr:
          return is<SVGCircleElement>(*this) || is<SVGEllipseElement>(*this) ? CSSPropertyCx : CSSPropertyInvalid;
      case AttributeNames::cyAttr:
          return is<SVGCircleElement>(*this) || is<SVGEllipseElement>(*this) ? CSSPropertyCy : CSSPropertyInvalid;
      case AttributeNames::dAttr:
          return is<SVGPathElement>(*this) && document().settings().cssDPropertyEnabled() ? CSSPropertyD : CSSPropertyInvalid;
      case AttributeNames::directionAttr:
          return CSSPropertyDirection;
      case AttributeNames::displayAttr:
          return CSSPropertyDisplay;
...
    case AttributeNames::xAttr:
        return supportsXY() ? CSSPropertyX : CSSPropertyInvalid;
    case AttributeNames::yAttr:
        return supportsXY() ? CSSPropertyY : CSSPropertyInvalid;
    default:
        break;
    }

    return CSSPropertyInvalid;
}
```

I think it's easier to read this way.

nikolaszimmermann (approved):
@karlcow looks good, I'd just like to make the big-switch-function a bit more readable, left a comment for that.

Source/WebCore/svg/SVGElement.h
karlcow: I didn't remember at first.
hehe. Now I remember… it's because of safer CPP. :)

Source/WebCore/svg/SVGElement.cpp
karlcow: fixed it.

Source/WebCore/svg/SVGElement.cpp
karlcow: I changed it.

PR #63704, opened 2026-04-27
[TSan] bmalloc: fall through to system allocator under ThreadSanitizer

sheeparegreat:
#### e05f64d91daff6fa89cd3e630a65dad791da84f2
<pre>
[TSan] bmalloc: fall through to system allocator under ThreadSanitizer
<a href="https://bugs.webkit.org/show_bug.cgi?id=313440">https://bugs.webkit.org/show_bug.cgi?id=313440</a>

Reviewed by Marcus Plutowski.

libpas manages memory directly via mmap without going through
malloc/free, so TSan never sees the happens-before edges between
an object&apos;s last free and its next allocation, and reports every
allocator-recycled slot as a use-after-free race. Under TSan,
force BUSE_SYSTEM_MALLOC so the libpas allocator is bypassed
entirely and TSan can track allocation lifetimes through the
system allocator.

Add BTSAN_ENABLED macro to BCompiler.h (matching the existing
BASAN_ENABLED pattern), then gate BUSE_SYSTEM_MALLOC on it in
BPlatform.h. This cascades through the existing preprocessor
logic: BUSE_SYSTEM_MALLOC=1 disables BENABLE_LIBPAS, which in
turn disables BUSE_LIBPAS, BUSE_TZONE, and BENABLE_MTE.

No behaviour change in non-TSan builds.

* Source/bmalloc/bmalloc/BCompiler.h:
* Source/bmalloc/bmalloc/BPlatform.h:

Canonical link: <a href="https://commits.webkit.org/312658@main">https://commits.webkit.org/312658@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/4b69ec94b0423b0714f07ab3376cfcad302ecf37

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/160438 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/33911 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/27012 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/169341 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/115504 "Built successfully") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/34012 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/33911 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/124398 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/87867 "1 flakes 2 failures") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/163396 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/26644 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/144118 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/104997 "Passed tests") | | [❌ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/0cbbd73b-8ee1-479b-bada-22102cb23016) 
| | [❌ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/25696 "Found 1 new test failure: imported/w3c/web-platform-tests/navigation-api/navigate-event/signal-abort-window-stop.html (failure)") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/24199 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/17060 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/152494 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/135390 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/21881 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/171819 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/21275 "Built successfully and passed tests") | | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/23521 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/132656 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/33585 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/28293 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/132677 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/33569 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/143676 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/95351 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/24447 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/27280 "Passed tests") | [❌ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/20490 "Found 1 new test failure: scrollingcoordinator/mac/latching/horizontal-overflow-back-swipe.html (failure)") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/192837 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/33079 "Built successfully") | | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/49657 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/32579 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/32823 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/32727 "Built successfully") | | | |


Source/bmalloc/bmalloc/bmalloc.h
```diff
@@ -46,6 +46,15 @@ BALLOW_UNSAFE_BUFFER_USAGE_BEGIN
 #include "mimalloc.h"
 #endif
 
+// Under TSan, bypass libpas and use the system allocator so TSan can track
+// memory reuse. libpas's thread-local caches reuse memory without going
+// through malloc/free, which makes TSan unable to see happens-before edges.
+#if __has_feature(thread_sanitizer)
```
ddkilzer: This needs a `BTSAN_ENABLED` macro defined in `Source/bmalloc/bmalloc/BCompiler.h` next to `BASAN_ENABLED`.

ddkilzer (changes requested):
Need to use `BTSAN_ENABLED` macro.

Source/bmalloc/bmalloc/bmalloc.h
```diff
@@ -62,7 +71,7 @@ inline pas_primitive_heap_ref& heapForKind(Gigacage::Kind kind)
 // Returns null on failure.
 BINLINE void* tryMalloc(size_t size, CompactAllocationMode mode, HeapKind kind = HeapKind::Primary)
 {
-#if BUSE(LIBPAS)
+#if BUSE(LIBPAS) && !BMALLOC_USE_SYSTEM_MALLOC_FOR_TSAN
```
Achierius: Why not gate BUSE(LIBPAS) on BMALLOC_USE_SYSTEM_MALLOC_FOR_TSAN instead? We don't want to rely on them being coupled in case people add uses of the former in the future.

Source/bmalloc/bmalloc/bmalloc.h
```diff
@@ -46,6 +46,15 @@ BALLOW_UNSAFE_BUFFER_USAGE_BEGIN
 #include "mimalloc.h"
 #endif
 
+// Under TSan, bypass libpas and use the system allocator so TSan can track
+// memory reuse. libpas's thread-local caches reuse memory without going
+// through malloc/free, which makes TSan unable to see happens-before edges.
+#if BTSAN_ENABLED
```
Achierius: If this is added as part of BUSE_LIBPAS then it doesn't need a separate macro def, it can just use BTSAN_ENABLED

Achierius (changes requested):

Source/bmalloc/bmalloc/bmalloc.h
```diff
@@ -46,6 +46,15 @@ BALLOW_UNSAFE_BUFFER_USAGE_BEGIN
 #include "mimalloc.h"
 #endif
 
+// Under TSan, bypass libpas and use the system allocator so TSan can track
+// memory reuse. libpas's thread-local caches reuse memory without going
+// through malloc/free, which makes TSan unable to see happens-before edges.
```
Achierius: I don't fully understand what you mean by "thread-local caches reuse memory without going through malloc/free" -- all allocators reuse memory internally, and since they implement malloc/free they can't rely on it. In principle you could have an allocator which has every allocation/free call bottom out in mmap, but that would be far too slow to be practical.

I wouldn't be surprised if the TLCs _did_ interact poorly with TSan, but it wouldn't be for this reason. My guess would be that the problem comes from the scavenger + relevant safepointing mechanics are the problem, as TSan wouldn't understand why it's safe for the scavenger to go mucking around in other thread's TLCs. I ask in part because, if it's *not* too big of a lift, then it would be preferable to have the TSan pass use libpas, to better match the execution paths we see on the real system.

Am I correct in assuming that TSan works when mimalloc is enabled? Or does it need to fall through all the way to xzmalloc?

ddkilzer: Maybe @DanBlackwell can answer this question.

DanBlackwell: I looked at this a while ago, so it's not too fresh in my memory. I agree that in the long run, we should find a way to have TSan using the same paths as the release builds. 

> Am I correct in assuming that TSan works when mimalloc is enabled? Or does it need to fall through all the way to xzmalloc?

TSan intercepts and replaces malloc/free calls with calls into its own allocator. Therefore, I think it's fine to 'use' mimalloc as it has the regular allocator interface - just be aware that behind the scenes it's actually using TSan's own allocator.

There is one benefit to using TSan's allocator, which is getting allocation site traces - see the end here:
```
==================
WARNING: ThreadSanitizer: data race (pid=72756)
  Write of size 4 at 0x00010bb00710 by thread T2:
    #0 worker tsan_race.c:9 (tsan_race:arm64+0x1000007b8)

  Previous write of size 4 at 0x00010bb00710 by thread T1:
    #0 worker tsan_race.c:9 (tsan_race:arm64+0x1000007b8)

  Location is heap block of size 4 at 0x00010bb00710 allocated by main thread:
    #0 malloc sanitizer_malloc_mac.inc:135 (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x66044)
    #1 _malloc_type_malloc_outlined <null> (libsystem_malloc.dylib:arm64e+0x1e9b8)
    #2 main tsan_race.c:20 (tsan_race:arm64+0x100000674)
```

> I wouldn't be surprised if the TLCs did interact poorly with TSan, but it wouldn't be for this reason. My guess would be that the problem comes from the scavenger + relevant safepointing mechanics are the problem, as TSan wouldn't understand why it's safe for the scavenger to go mucking around in other thread's TLCs.

Is there an edge (store-release to load-acquire) that TSan should be able to see between the threads? Perhaps there's some `TSAN_ANNOTATE_HAPPENS_BEFORE`-like annotations we can use to make TSan happy.


One thing that makes TSan support in WebKit tricky is that the TSan model of happens-before doesn't work with memory fences, I see some of these here: https://github.com/WebKit/WebKit/blob/a5a8dcdb7013507fb5334c990ab8b516ef027b2a/Source/WTF/wtf/Atomics.h#L251. It may be the case that annotating these with the appropriate `TSAN_ANNOTATE_HAPPENS_BEFORE` et al. or converting them to std::atomic load-acq store-rel operations will make things better.

DanBlackwell: >// Under TSan, bypass libpas and use the system allocator so TSan can track
// memory reuse. libpas's thread-local caches reuse memory without going
// through malloc/free, which makes TSan unable to see happens-before edges.

I think that the issue here is not the reusing TLCs, but specifically reusing TLCs across threads combined with TSan not being able to see the happens-before relationship between the 'free on Thread 1' and the 'alloc on Thread 2'. Using the 'system allocator' (really TSan's own allocator) is the easy way to eliminate false positives caused by this. 

(side-note is TLC really the right term if it can be reused across threads?)

Achierius: Re.: the source of the problem:

> side-note is TLC really the right term if it can be reused across threads?

The TLC itself is thread-local, but the backing memory (of various kinds, not just object payloads) can be reclaimed by the allocator. Otherwise, you end up with the situation that mimalloc used to have where you can have e.g. a sleeping thread hang on to a bunch of unallocated memory in its cache, causing unnecessary memory pressure. Libpas would safepoint the thread, remove the memory from the TLC, and decommit it.

Separately, the memory allocated by a TLC is *not* actually thread-local. Libpas is OK with allocating memory from one thread and deallocating on another -- at `free`-time the freeing thread just pushes the allocation onto its own deallocation-log (not that belonging to the thread which allocated it), which is then reconciled with the originating TLC (if applicable) when the scavenger next processes that thread's log.

I expect that this is probably the issue -- to slightly amend what you said earlier

> reusing TLCs across threads combined with TSan not being able to see the happens-before relationship between the 'free on Thread 1' and the 'alloc on Thread 2'.

Rather than reusing the TLCs, either memory migrating between TLCs or objects being pulled from one thread's decommit queue to another thread's allocation-bits via the scavenger.

> Is there an edge (store-release to load-acquire) that TSan should be able to see between the threads?

Unfortunately I don't think so -- at least not one I'd be willing to vouch for with any confidence. A lot of synchronization (esp. for performance-sensitive areas like the deallocation log) relies on implicit or explicit barriers and avoids the use of atomics for performance.

Achierius: > TSan intercepts and replaces malloc/free calls with calls into its own allocator. Therefore, I think it's fine to 'use' mimalloc as it has the regular allocator interface - just be aware that behind the scenes it's actually using TSan's own allocator.

Either mimalloc or libmalloc works fine then. I'd softly prefer bottoming out at libmalloc (for consistency / clarity -- generally when we expect to be interposed we call libmalloc's apis) but Windows in particular has mimalloc as the 'allocator of last resort' since the system malloc has an incompatible API.

DanBlackwell: > > Is there an edge (store-release to load-acquire) that TSan should be able to see between the threads?

> Unfortunately I don't think so -- at least not one I'd be willing to vouch for with any confidence. A lot of synchronization (esp. for performance-sensitive areas like the deallocation log) relies on implicit or explicit barriers and avoids the use of atomics for performance.

Makes sense

Source/bmalloc/bmalloc/bmalloc.h
ddkilzer: Fixed

Source/bmalloc/bmalloc/bmalloc.h
ddkilzer: Fixed

Achierius (approved):
r=me

PR #25851, opened 2024-03-13
RegisteredContentScripts.db should only be created if needed.

xeenon:
#### 5745aee1167a4b0f77891efe606302d5160fdbde
<pre>
RegisteredContentScripts.db should only be created if needed.
<a href="https://webkit.org/b/270931">https://webkit.org/b/270931</a>
<a href="https://rdar.apple.com/problem/124554561">rdar://problem/124554561</a>

Reviewed by Sihui Liu and Jeff Miller.

Change the APIs that use SQL store to use the new _openDatabaseIfNecessaryReturningErrorMessage:createIfNecessary:
method and pass NO when the access is deleting or getting data. This makes sure we don&apos;t create an empty database
just to read or delete no data.

* Source/WebKit/Shared/Extensions/_WKWebExtensionRegisteredScriptsSQLiteStore.mm:
(-[_WKWebExtensionRegisteredScriptsSQLiteStore deleteScriptsWithIDs:completionHandler:]): Pass createIfNecessary:NO.
(-[_WKWebExtensionRegisteredScriptsSQLiteStore _getScriptsWithOutErrorMessage:]): Ditto.
* Source/WebKit/Shared/Extensions/_WKWebExtensionSQLiteDatabase.mm:
(-[_WKWebExtensionSQLiteDatabase _openWithFlags:vfs:error:]): Return no error if thsi is a can&apos;t open error and
we were not asekd to create the file.
* Source/WebKit/Shared/Extensions/_WKWebExtensionSQLiteStore.h:
* Source/WebKit/Shared/Extensions/_WKWebExtensionSQLiteStore.mm:
(-[_WKWebExtensionSQLiteStore _openDatabaseIfNecessaryReturningErrorMessage:]): Call the new method with YES.
(-[_WKWebExtensionSQLiteStore _openDatabaseIfNecessaryReturningErrorMessage:createIfNecessary:]): Added.
(-[_WKWebExtensionSQLiteStore _openDatabase:withAccessType:deleteDatabaseFileOnError:]): Added accessType param.
(-[_WKWebExtensionSQLiteStore _deleteDatabaseFileAtURL:reopenDatabase:]): Pass SQLiteDatabaseAccessTypeReadWriteCreate.
(-[_WKWebExtensionSQLiteStore _openDatabase:deleteDatabaseFileOnError:]): Deleted.
* Source/WebKit/UIProcess/Extensions/Cocoa/_WKWebExtensionDeclarativeNetRequestSQLiteStore.mm:
[...]

PR #63735, opened 2026-04-27
Bounding box does not account for skinning transforms

mwyrzykowski:
#### f9fdeabc55e9440ff8cb05e1210631b266696b4f
<pre>
Bounding box does not account for skinning transforms
<a href="https://bugs.webkit.org/show_bug.cgi?id=313476">https://bugs.webkit.org/show_bug.cgi?id=313476</a>
<a href="https://rdar.apple.com/174193414">rdar://174193414</a>

Reviewed by Etienne Segonzac.

In computing the bounding box of a mesh, we considered the instance
transform but failed to consider the skinning transform. Correct this
by taking that transform into account.

Also fix constantly recomputing bounding boxes of skinned meshes
which is not compliant with the visionOS implementation.

* Source/WebKit/GPUProcess/graphics/Model/ModelBridge.swift:
(WKBridgeSkinningData.rootJointIndices):
* Source/WebKit/GPUProcess/graphics/Model/ModelTypes.h:
* Source/WebKit/GPUProcess/graphics/Model/USDModel.swift:
(rootSkinMatrices(for:)):
(rootJointIndices(forMeshAt:in:)):
(rootJointIndices(from:)):
(updateMesh(_:)):
(USDModelLoader.processMeshUpdates(_:)):
* Source/WebKit/GPUProcess/graphics/Model/WebKitMesh.mm:
(WebModel::convert):
* Source/WebKit/Shared/Model.serialization.in:
* Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp:
(WebKit::computeMinAndMaxCorners):
(WebKit::RemoteMeshProxy::update):
* Source/WebKit/WebProcess/Model/ModelInlineConverters.h:
(WebKit::convert):

Canonical link: <a href="https://commits.webkit.org/313775@main">https://commits.webkit.org/313775@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/311fbb0110a1d5b8d8b43a0ea7f16eb7e204e2ee

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/163961 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/37445 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/30664 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/172990 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/121212 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [loading 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b45cdcce-445b-4631-be10-0fc935fd0d50) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/37777 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/37445 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/127373 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/121212 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [loading 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/4fc3157b-3807-46a3-8319-308c592bffcd) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/166920 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/29659 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/147402 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/107921 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/92f0b172-203a-454a-922e-05ba5fd74d85) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/28705 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/27327 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/20754 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/138606 "Passed tests") | | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/175515 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/28337 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/26787 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/135683 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/37115 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/31439 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/135655 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36584 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/37900 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/146914 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/100063 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/30955 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/23770 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/36709 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/109756 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/36108 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/1810 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/36358 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/36255 "Built successfully") | | | |

### DIFF
--- a/Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
+++ b/Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
@@ -354,6 +354,100 @@ internal func logInfo(_ info: String) {
     Logger.modelGPU.info("\(info)")
 }
 
+// Mirrors RemoteMeshProxy.cpp::computeMinAndMaxCorners: for each root joint, computes
+// the skin matrix that places the mesh in world space for AABB purposes, returning
+// one matrix per root joint. Using only root joints avoids wildly over-inflating the
+// bounds — child joints move sub-regions that stay within the local AABB extent.
+private func rootSkinMatrices(_ skinningData: WKBridgeSkinningData) -> [simd_float4x4] {
+    let jointTransforms = skinningData.jointTransforms
+    guard !jointTransforms.isEmpty else { return [] }
+    let inverseBindPoses = skinningData.inverseBindPoses
+    let geomBind = skinningData.geometryBindTransform
+    // Fall back to index 0 when rootJointIndices is empty (single-root or unknown).
+    let indices = skinningData.rootJointIndices.isEmpty ? [UInt32(0)] : skinningData.rootJointIndices
+    return indices.compactMap { idx -> simd_float4x4? in
+        let i = Int(idx)
+        guard i < jointTransforms.count else { return nil }
+        let invBind = i < inverseBindPoses.count ? inverseBindPoses[i] : matrix_identity_float4x4
+        return simd_mul(simd_mul(jointTransforms[i], invBind), geomBind)
+    }
+}
+
+// Computes root joint indices for the mesh at meshPath by reading the skeleton's
+// joint token paths from the UsdStage. A root joint is one whose parent token path
+// is not present in the skeleton's joint list.
+private func rootJointIndices(forMeshAt meshPath: String, in stage: UsdStage) -> [UInt32] {
+    let meshPrim = stage.prim(at: SdfPath(meshPath))
+    guard meshPrim.isValid else { return [] }
+
+    // Walk up the prim hierarchy to find a skel:skeleton relationship.
+    var skelPrimPath: SdfPath? = nil
+    var current: UsdPrim? = meshPrim
+    while let prim = current {
+        if let rel = prim.relationship(named: "skel:skeleton"),
+            let targets = rel.forwardedTargets, let target = targets.first
+        {
+            skelPrimPath = target
+            break
+        }
+        current = prim.parent
+    }
+
+    guard let skelPath = skelPrimPath else { return [] }
+    let skelPrim = stage.prim(at: skelPath)
+    guard skelPrim.isValid,
+        let jointsAttr = skelPrim.attribute("joints"),
+        let jointTokens: TokenArray = jointsAttr.get()
+    else { return [] }
+
+    return rootJointIndices(from: jointTokens.map(\.string))
+}
+
+// Pure string computation: returns the indices of joints whose parent path does
+// not appear in the joints array. USD requires parents before children, so root
+// joints always appear before their descendants but there can be multiple roots
+// (e.g. ["A", "A/B", "C", "C/D/E"] has roots at indices 0 and 2).
+private func rootJointIndices(from joints: [String]) -> [UInt32] {
+    let jointSet = Set(joints)
+    return joints.enumerated()
+        .compactMap { index, joint -> UInt32? in
+            var path = joint
+            while let slash = path.lastIndex(of: "/") {
+                path = String(path[..<slash])
+                if jointSet.contains(path) { return nil }
+            }
+            return UInt32(index)
+        }
+}
+
+// Transforms a local-space AABB by one or more skin matrices and returns the
+// union of the resulting world-space AABBs. Passing multiple matrices handles
+// multi-root skeletons correctly.
+private func computeSkinningAABB(
+    _ skinMatrices: [simd_float4x4],
+    _ localMin: SIMD3<Float>,
+    _ localMax: SIMD3<Float>
+) -> (SIMD3<Float>, SIMD3<Float>) {
+    let matrices = skinMatrices.isEmpty ? [matrix_identity_float4x4] : skinMatrices
+    var effectiveMin = SIMD3<Float>(repeating: .greatestFiniteMagnitude)
+    var effectiveMax = SIMD3<Float>(repeating: -.greatestFiniteMagnitude)
+    for skinMatrix in matrices {
+        for i in 0..<8 {
+            let corner = SIMD4<Float>(
+                (i & 1) != 0 ? localMax.x : localMin.x,
+                (i & 2) != 0 ? localMax.y : localMin.y,
+                (i & 4) != 0 ? localMax.z : localMin.z,
+                1
+            )
+            let transformed = simd_mul(skinMatrix, corner)
+            let p = SIMD3<Float>(transformed.x, transformed.y, transformed.z)
+            effectiveMin = simd_min(effectiveMin, p)
+            effectiveMax = simd_max(effectiveMax, p)
+        }
+    }
+    return (effectiveMin, effectiveMax)
+}
+
 @objc
 @implementation
 extension WKBridgeUSDConfiguration {
@@ -791,6 +885,7 @@ extension WKBridgeReceiver {
 
                         var deferredMeshUpdate = DeferredMeshUpdate(identifier: identifier, type: .newMesh, updatedInstances: [])
 
+                        let skinMatrices = meshData.deformationData?.skinningData.map { rootSkinMatrices($0) } ?? []
                         for (partIndex, part) in meshData.parts.enumerated() {
                             if part.materialIndex >= meshData.assignedMaterials.count {
                                 fatalError(
@@ -811,15 +906,21 @@ extension WKBridgeReceiver {
                                 )
                             )
 
+                            let (effectiveMin, effectiveMax) = computeSkinningAABB(
+                                skinMatrices,
+                                part.boundsMin,
+                                part.boundsMax
+                            )
+
                             let meshPart = try renderContext.makeMeshPart(
                                 resource: meshResource,
-                                indexOffset: meshData.parts[partIndex].indexOffset,
-                                indexCount: meshData.parts[partIndex].indexCount,
-                                primitive: meshData.parts[partIndex].topology,
+                                indexOffset: part.indexOffset,
+                                indexCount: part.indexCount,
+                                primitive: part.topology,
                                 windingOrder: .counterClockwise,
                                 bounds: .init(
-                                    boxMin: meshData.parts[partIndex].boundsMin,
-                                    boxMax: meshData.parts[partIndex].boundsMax
+                                    boxMin: effectiveMin,
+                                    boxMax: effectiveMax
                                 )
                             )
 
@@ -1028,7 +1129,8 @@ private func webUpdateTextureRequestFromUpdateTextureRequest(_ request: _Proto_T
 }
 
 private func webUpdateMeshRequestFromUpdateMeshRequest(
-    _ request: _Proto_MeshDataUpdate_v1
+    _ request: _Proto_MeshDataUpdate_v1,
+    rootJointIndices: [UInt32]
 ) -> WKBridgeUpdateMesh {
     var descriptor: WKBridgeMeshDescriptor?
     if let requestDescriptor = request.descriptor {
@@ -1045,7 +1147,7 @@ private func webUpdateMeshRequestFromUpdateMeshRequest(
         instanceTransforms: toData(request.instanceTransformsCompat()),
         instanceTransformsCount: request.instanceTransformsCompat().count,
         assignedMaterials: convert(request.assignedMaterials),
-        deformationData: .init(request.deformationData)
+        deformationData: .init(request.deformationData, rootJointIndices: rootJointIndices)
     )
 }
 
@@ -1832,6 +1934,7 @@ final class USDModelLoader {
     fileprivate var stage: UsdStage?
     fileprivate var data: Data?
     private let objcLoader: WKBridgeModelLoader
+    private var rootJointIndicesCache: [String: [UInt32]] = [:]
 
     @nonobjc
     fileprivate var time: TimeInterval = 0
@@ -1851,6 +1954,7 @@ final class USDModelLoader {
     }
 
     func loadModel(from url: Foundation.URL) {
+        rootJointIndicesCache.removeAll()
         do {
             let stage = try UsdStage.open(url)
             self.setupTimes(from: stage)
@@ -1861,6 +1965,7 @@ final class USDModelLoader {
     }
 
     func loadModel(data: Foundation.Data) -> Bool {
+        rootJointIndicesCache.removeAll()
         do {
             self.stage = try UsdStage.open(buffer: data)
             guard let stage = self.stage else {
@@ -1990,7 +2095,25 @@ final class USDModelLoader {
     }
 
     private func processMeshUpdates(_ updates: [_Proto_MeshDataUpdate_v1]) {
-        self.objcLoader.updateMesh(webRequest: updates.map { webUpdateMeshRequestFromUpdateMeshRequest($0) })
+        let stage = self.stage
+        self.objcLoader.updateMesh(
+            webRequest: updates.map { update in
+                let rootIndices: [UInt32]
+                if let stage, update.deformationData?.skinningData != nil {
+                    let path = update.id.path
+                    if let cached = rootJointIndicesCache[path] {
+                        rootIndices = cached
+                    } else {
+                        let computed = rootJointIndices(forMeshAt: path, in: stage)
+                        rootJointIndicesCache[path] = computed
+                        rootIndices = computed
+                    }
+                } else {
+                    rootIndices = []
+                }
+                return webUpdateMeshRequestFromUpdateMeshRequest(update, rootJointIndices: rootIndices)
+            }
+        )
     }
 }
 
--- a/Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
+++ b/Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
@@ -70,7 +70,7 @@ static WebModel::Float4x4 makeTransformMatrix(
     return result;
 }
 
-static std::pair<simd_float4, simd_float4> computeMinAndMaxCorners(const Vector<WebModel::MeshPart>& parts, const Vector<WebModel::Float4x4>& instanceTransforms)
+static std::pair<simd_float4, simd_float4> computeMinAndMaxCorners(const Vector<WebModel::MeshPart>& parts, const Vector<WebModel::Float4x4>& instanceTransforms, const std::optional<WebModel::DeformationData>& deformationData)
 {
     simd_float4 minCorner4 = simd_make_float4(FLT_MAX, FLT_MAX, FLT_MAX, 1.f);
     simd_float4 maxCorner4 = simd_make_float4(-FLT_MAX, -FLT_MAX, -FLT_MAX, 1.f);
@@ -87,13 +87,35 @@ static std::pair<simd_float4, simd_float4> computeMinAndMaxCorners(const Vector<
         corners[6] = simd_make_float3(part.boundsMin.x, part.boundsMax.y, part.boundsMax.z);
         corners[7] = simd_make_float3(part.boundsMax.x, part.boundsMax.y, part.boundsMax.z);
 
-        for (auto& transform : instanceTransforms) {
-            for (int j = 0; j < 8; ++j) {
-                simd_float4 corner4 = simd_make_float4(corners[j].x, corners[j].y, corners[j].z, 1.f);
-                simd_float4 transformedCorner = simd_mul(transform, corner4);
+        Vector<simd_float4x4> rootSkinMatrices;
+        if (deformationData && deformationData->skinningData && !deformationData->skinningData->jointTransforms.isEmpty()) {
+            const auto& skinning = *deformationData->skinningData;
+            const simd_float4x4 geomBind = skinning.geometryBindTransform;
+            const auto& rootIndices = skinning.rootJointIndices;
+            const bool hasRootIndices = !rootIndices.isEmpty();
+            const size_t iterCount = hasRootIndices ? rootIndices.size() : 1;
+            rootSkinMatrices.reserveInitialCapacity(iterCount);
+            for (size_t r = 0; r < iterCount; ++r) {
+                uint32_t rootIdx = hasRootIndices ? rootIndices[r] : 0;
+                if (rootIdx >= skinning.jointTransforms.size())
+                    continue;
+                const simd_float4x4 invBind = (rootIdx < skinning.inverseBindPoses.size())
+                    ? static_cast<simd_float4x4>(skinning.inverseBindPoses[rootIdx])
+                    : matrix_identity_float4x4;
+                rootSkinMatrices.append(simd_mul(simd_mul(skinning.jointTransforms[rootIdx], invBind), geomBind));
+            }
+        } else
+            rootSkinMatrices.append(matrix_identity_float4x4);
 
-                minCorner4 = simd_min(transformedCorner, minCorner4);
-                maxCorner4 = simd_max(transformedCorner, maxCorner4);
+        for (auto& transform : instanceTransforms) {
+            for (const auto& skinMatrix : rootSkinMatrices) {
+                const simd_float4x4 worldTransform = simd_mul(transform, skinMatrix);
+                for (int j = 0; j < 8; ++j) {
+                    simd_float4 corner4 = simd_make_float4(corners[j].x, corners[j].y, corners[j].z, 1.f);
+                    simd_float4 transformedCorner = simd_mul(worldTransform, corner4);
+                    minCorner4 = simd_min(transformedCorner, minCorner4);
+                    maxCorner4 = simd_max(transformedCorner, maxCorner4);
+                }
             }
         }
     }
@@ -124,24 +146,21 @@ RemoteMeshProxy::~RemoteMeshProxy()
 void RemoteMeshProxy::update(Vector<WebModel::UpdateMeshDescriptor>&& descriptorArray)
 {
 #if ENABLE(GPU_PROCESS_MODEL)
-    bool anyBoundingBoxChanged = false;
-    for (auto& descriptor : descriptorArray) {
-        auto [minCorner, maxCorner] = computeMinAndMaxCorners(descriptor.parts, descriptor.instanceTransforms);
-        auto boundingBoxChanged = minCorner.x <= maxCorner.x && minCorner.y <= maxCorner.y && minCorner.z <= maxCorner.z;
-        boundingBoxChanged = boundingBoxChanged && (!simd_equal(m_minCorner, minCorner) || !simd_equal(m_maxCorner, maxCorner));
-        if (boundingBoxChanged) {
+    bool needBoundingBoxUpdate = m_minCorner.x > m_maxCorner.x;
+    if (needBoundingBoxUpdate) {
+        for (auto& descriptor : descriptorArray) {
+            auto [minCorner, maxCorner] = computeMinAndMaxCorners(descriptor.parts, descriptor.instanceTransforms, descriptor.deformationData);
             m_minCorner = simd_min(m_minCorner, minCorner);
             m_maxCorner = simd_max(m_maxCorner, maxCorner);
         }
-        anyBoundingBoxChanged = anyBoundingBoxChanged || boundingBoxChanged;
     }
 
     auto sendResult = sendWithAsyncReply(Messages::RemoteMesh::Update(WTF::move(descriptorArray)), [](auto) mutable {
     });
     UNUSED_VARIABLE(sendResult);
-    if (anyBoundingBoxChanged)
-        computeTransform();
 
+    if (needBoundingBoxUpdate)
+        computeTransform();
 #else
     UNUSED_PARAM(descriptorArray);
 #endif
--- a/Source/WebKit/GPUProcess/graphics/Model/ModelBridge.swift
+++ b/Source/WebKit/GPUProcess/graphics/Model/ModelBridge.swift
@@ -134,6 +134,7 @@ extension WKBridgeSkinningData {
     let influenceJointIndicesData: Data?
     let influenceWeightsData: Data?
     let geometryBindTransform: simd_float4x4
+    let rootJointIndicesData: Data?
 
     init(
         influencePerVertexCount: UInt8,
@@ -141,7 +142,8 @@ extension WKBridgeSkinningData {
         inverseBindPoses: Data?,
         influenceJointIndices: Data?,
         influenceWeights: Data?,
-        geometryBindTransform: simd_float4x4
+        geometryBindTransform: simd_float4x4,
+        rootJointIndices: Data?
     ) {
         self.influencePerVertexCount = influencePerVertexCount
         self.jointTransformsData = jointTransforms
@@ -149,6 +151,7 @@ extension WKBridgeSkinningData {
         self.influenceJointIndicesData = influenceJointIndices
         self.influenceWeightsData = influenceWeights
         self.geometryBindTransform = geometryBindTransform
+        self.rootJointIndicesData = rootJointIndices
     }
 }
 
@@ -684,9 +687,10 @@ extension WKBridgeSkinningData {
     var inverseBindPoses: [simd_float4x4] { inverseBindPosesData.map { decodeValues(from: $0) } ?? [] }
     var influenceJointIndices: [UInt32] { influenceJointIndicesData.map { decodeValues(from: $0) } ?? [] }
     var influenceWeights: [Float] { influenceWeightsData.map { decodeValues(from: $0) } ?? [] }
+    var rootJointIndices: [UInt32] { rootJointIndicesData.map { decodeValues(from: $0) } ?? [] }
 
     @nonobjc
-    convenience init?(_ request: _Proto_DeformationData_v1.SkinningData?) {
+    convenience init?(_ request: _Proto_DeformationData_v1.SkinningData?, rootJointIndices: [UInt32] = []) {
         guard let request else {
             return nil
         }
@@ -697,7 +701,8 @@ extension WKBridgeSkinningData {
             inverseBindPoses: toData(request.inverseBindPosesCompat()),
             influenceJointIndices: toData(request.influenceJointIndices),
             influenceWeights: toData(request.influenceWeights),
-            geometryBindTransform: request.geometryBindTransformCompat()
+            geometryBindTransform: request.geometryBindTransformCompat(),
+            rootJointIndices: rootJointIndices.isEmpty ? nil : toData(rootJointIndices)
         )
     }
 }
@@ -731,13 +736,13 @@ extension WKBridgeRenormalizationData {
 }
 extension WKBridgeDeformationData {
     @nonobjc
-    convenience init?(_ request: _Proto_DeformationData_v1?) {
+    convenience init?(_ request: _Proto_DeformationData_v1?, rootJointIndices: [UInt32] = []) {
         guard let request else {
             return nil
         }
 
         self.init(
-            skinningData: .init(request.skinningData),
+            skinningData: .init(request.skinningData, rootJointIndices: rootJointIndices),
             blendShapeData: .init(request.blendShapeData),
             renormalizationData: .init(request.renormalizationData)
         )
--- a/Source/WebKit/GPUProcess/graphics/Model/ModelTypes.h
+++ b/Source/WebKit/GPUProcess/graphics/Model/ModelTypes.h
@@ -137,9 +137,10 @@ typedef NS_ENUM(uint8_t, WKBridgeVertexSemantic) {
 @property (nonatomic, readonly, nullable) NSData *influenceJointIndicesData; // [UInt32]
 @property (nonatomic, readonly, nullable) NSData *influenceWeightsData; // [Float]
 @property (nonatomic, readonly) simd_float4x4 geometryBindTransform;
+@property (nonatomic, readonly, nullable) NSData *rootJointIndicesData; // [UInt32] indices of joints with no parent in the skeleton
 
 - (instancetype)init NS_UNAVAILABLE;
-- (instancetype)initWithInfluencePerVertexCount:(uint8_t)influencePerVertexCount jointTransforms:(nullable NSData *)jointTransforms inverseBindPoses:(nullable NSData *)inverseBindPoses influenceJointIndices:(nullable NSData *)influenceJointIndices influenceWeights:(nullable NSData *)influenceWeights geometryBindTransform:(simd_float4x4)geometryBindTransform NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithInfluencePerVertexCount:(uint8_t)influencePerVertexCount jointTransforms:(nullable NSData *)jointTransforms inverseBindPoses:(nullable NSData *)inverseBindPoses influenceJointIndices:(nullable NSData *)influenceJointIndices influenceWeights:(nullable NSData *)influenceWeights geometryBindTransform:(simd_float4x4)geometryBindTransform rootJointIndices:(nullable NSData *)rootJointIndices NS_DESIGNATED_INITIALIZER;
 
 @end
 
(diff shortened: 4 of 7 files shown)


Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -1966,6 +1966,39 @@ final class USDModelLoader {
         usdStageSession = _Proto_UsdStageSession_v1.noMetalSessionWithSynchronizedUpdate(gpuFamily: gpuFamily)
     }
 
+    func rootExtents(time: UsdTimeCode = .default) -> (min: SIMD3<Float>, max: SIMD3<Float>) {
+        let invalidBoundingBox = (SIMD3<Float>.one * Float.greatestFiniteMagnitude, SIMD3<Float>.one * -Float.greatestFiniteMagnitude)
+        guard let stage = stage else { return invalidBoundingBox }
```
rr-codes:
```suggestion
        guard let stage else { return invalidBoundingBox }
```

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -1966,6 +1966,39 @@ final class USDModelLoader {
...
+        let combinedRange: GfRange3d
+        guard hasStaticExtent else {
+            return invalidBoundingBox
+        }
+        let cache = UsdGeomBBoxCache(
+            time: .default,
+            includedPurposes: [TfToken("default"), TfToken("render")],
+            useExtentHints: true
+        )
+        combinedRange = cache.computeWorldBound(for: rootPrim).computeAlignedRange()
```
rr-codes:
```suggestion
        let combinedRange = cache.computeWorldBound(for: rootPrim).computeAlignedRange()
```

and then can remove the `        let combinedRange: GfRange3d`

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -1966,6 +1966,39 @@ final class USDModelLoader {
...
+        var max = SIMD3<Float>(combinedRange.max.simd * metersPerUnit)
+
+        if stage.upAxis == TfToken("Z") {
+            (min, max) = (
+                SIMD3<Float>(min.x, min.z, -max.y),
+                SIMD3<Float>(max.x, max.z, -min.y)
+            )
+        }
+
+        return (min: min, max: max)
```
rr-codes:
```suggestion
        return (min, max)
```

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -2025,6 +2058,14 @@ final class USDModelLoader {
     func loadModel(from data: Data) {
     }
 
+    @objc(computeRootExtents)
```
rr-codes: shouldn't be needed

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -1966,6 +1966,39 @@ final class USDModelLoader {
         usdStageSession = _Proto_UsdStageSession_v1.noMetalSessionWithSynchronizedUpdate(gpuFamily: gpuFamily)
     }
 
+    func rootExtents(time: UsdTimeCode = .default) -> (min: SIMD3<Float>, max: SIMD3<Float>) {
+        let invalidBoundingBox = (SIMD3<Float>.one * Float.greatestFiniteMagnitude, SIMD3<Float>.one * -Float.greatestFiniteMagnitude)
+        guard let stage = stage else { return invalidBoundingBox }
+        let rootPrim = stage.defaultPrim ?? stage.pseudoRoot
+
+        let extentAttr = rootPrim.attribute(TfToken("extent"))
+        let hasStaticExtent = (extentAttr?.numTimeSamples == 0) && (extentAttr?.hasValue == true)
```
rr-codes:
```suggestion
        guard let extentAttr = rootPrim.attribute(TfToken("extent")) else {
        return invalidBoundingBox
        }
        
        
        let hasStaticExtent = extentAttr.numTimeSamples == 0 && extentAttr.hasValue
```

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
mwyrzykowski: I added it due to a compiler error where swift couldn't find the corresponding objective-C declaration, but I can't reproduce it, so it's removed now :)

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -2025,6 +2056,13 @@ final class USDModelLoader {
     func loadModel(from data: Data) {
     }
 
+    func computeRootExtents() -> simd_float2x3 {
+        // Use startTimeCode so we get valid extents for animated files, which only
+        // author extent as time samples and have no value at UsdTimeCode.default.
```
etiennesegonzac: not sure I follow, it seems like we always pass .default in rootExtents anyway

etiennesegonzac: I mean if the comment is true then the implementation is buggy :)

mwyrzykowski:
Using the user authored extents leads to differences from visionOS.

There is a much better way to solve this which is to look at the positions post-skinning upon load

mwyrzykowski:
I root caused the actual problem here, we were failing to account for the joint transforms when computing the AABB, so putting this back up for review

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
mwyrzykowski: Indeed this was the wrong thing to do, I've updated the PR and removed this change

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
```diff
@@ -124,24 +135,19 @@ RemoteMeshProxy::~RemoteMeshProxy()
...
-        auto [minCorner, maxCorner] = computeMinAndMaxCorners(descriptor.parts, descriptor.instanceTransforms);
-        auto boundingBoxChanged = minCorner.x <= maxCorner.x && minCorner.y <= maxCorner.y && minCorner.z <= maxCorner.z;
-        boundingBoxChanged = boundingBoxChanged && (!simd_equal(m_minCorner, minCorner) || !simd_equal(m_maxCorner, maxCorner));
-        if (boundingBoxChanged) {
+    auto sendResult = sendWithAsyncReply(Messages::RemoteMesh::Update(WTF::move(descriptorArray)), [](auto) mutable {
+    });
+    UNUSED_VARIABLE(sendResult);
+
+    if (m_minCorner.x > m_maxCorner.x) {
+        for (auto& descriptor : descriptorArray) {
```
ruthvikkonda, line 151: I think this is a use-after-move. We `WTF::move(descriptorArray)` right above. Should we change the order to compute bounds + transform first and then send the message after?

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
```diff
@@ -87,11 +87,22 @@ static std::pair<simd_float4, simd_float4> computeMinAndMaxCorners(const Vector<
...
+        if (deformationData && deformationData->skinningData && !deformationData->skinningData->jointTransforms.isEmpty()) {
+            // For skeleton-animated meshes, use the root joint (index 0) to establish world placement.
+            // Child joints only move sub-regions of the mesh and stay within the physical extent of the
+            // local AABB, so using all joints would wildly over-inflate the bounds.
+            const auto& skinning = *deformationData->skinningData;
+            const simd_float4x4 geomBind = skinning.geometryBindTransform;
+            simd_float4x4 invBind = !skinning.inverseBindPoses.isEmpty() ? (simd_float4x4)skinning.inverseBindPoses[0] : matrix_identity_float4x4;
+            skinMatrix = simd_mul(simd_mul((simd_float4x4)skinning.jointTransforms[0], invBind), geomBind);
+        }
+
```
ruthvikkonda: Nit: Since `skinMatrix` only depends on `deformationData`, can we pull this logic outside the loop?

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -806,14 +842,21 @@ extension WKBridgeReceiver {
                                 )
                             )
 
+                            let skinMatrix = computeRootSkinMatrix(meshData.deformationData?.skinningData)
```
ruthvikkonda: Nit: Since `skinMatrix` only depends on `deformationData`, can we pull this logic outside the loop?

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
ruthvikkonda: The Swift side of this (USDModel.swift:358) factors this out into a helper: `computeRootSkinMatrix`. What do you think about doing the same here for the C++ side? I think it'd help maintain consistency and make the two implementations easier to compare side by side when verifying they agree.

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
```diff
@@ -87,11 +87,22 @@ static std::pair<simd_float4, simd_float4> computeMinAndMaxCorners(const Vector<
...
         for (auto& transform : instanceTransforms) {
+            simd_float4x4 worldTransform = simd_mul((simd_float4x4)transform, skinMatrix);
             for (int j = 0; j < 8; ++j) {
                 simd_float4 corner4 = simd_make_float4(corners[j].x, corners[j].y, corners[j].z, 1.f);
-                simd_float4 transformedCorner = simd_mul(transform, corner4);
-
+                simd_float4 transformedCorner = simd_mul(worldTransform, corner4);
                 minCorner4 = simd_min(transformedCorner, minCorner4);
                 maxCorner4 = simd_max(transformedCorner, maxCorner4);
             }
```
ruthvikkonda: The Swift side of this (USDModel.swift:368) factors this out into a helper: `computeSkinningAABB`. What do you think about doing the same here for the C++ side? I think it'd help maintain consistency and make the two implementations easier to compare side by side when verifying they agree.

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -355,6 +355,42 @@ internal func logInfo(_ info: String) {
     Logger.modelGPU.info("\(info)")
 }
 
+// Mirrors RemoteMeshProxy.cpp::computeMinAndMaxExtents: computes the skin matrix that places the
```
ruthvikkonda: Nit: Typo here at `computeMinandMaxExtents`. Looks like it should be `computeMinAndMaxCorners`. If we end up going the helper function route discussed in the other comment, this should instead reference the new helper function.

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
```diff
@@ -87,11 +87,22 @@ static std::pair<simd_float4, simd_float4> computeMinAndMaxCorners(const Vector<
         corners[6] = simd_make_float3(part.boundsMin.x, part.boundsMax.y, part.boundsMax.z);
         corners[7] = simd_make_float3(part.boundsMax.x, part.boundsMax.y, part.boundsMax.z);
 
+        simd_float4x4 skinMatrix = matrix_identity_float4x4;
+        if (deformationData && deformationData->skinningData && !deformationData->skinningData->jointTransforms.isEmpty()) {
+            // For skeleton-animated meshes, use the root joint (index 0) to establish world placement.
```
ruthvikkonda: Nit: Since index 0 being the root is more of a convention rather than an enforced part of the schema, is it worth adding a link to a spec/reference here citing the convention?

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -806,14 +842,21 @@ extension WKBridgeReceiver {
...
+
                             let meshPart = try renderContext.makeMeshPart(
                                 resource: meshResource,
                                 indexOffset: meshData.parts[partIndex].indexOffset,
                                 indexCount: meshData.parts[partIndex].indexCount,
                                 primitive: meshData.parts[partIndex].topology,
                                 windingOrder: .counterClockwise,
-                                boundsMin: meshData.parts[partIndex].boundsMin,
-                                boundsMax: meshData.parts[partIndex].boundsMax
+                                boundsMin: effectiveMin,
```
ruthvikkonda: Not about this line specifically, but about the surrounding 
`if meshToMeshInstances[identifier] == nil` branch this lives in. `makeMeshPart` 
is only called on first creation, so the bounds set here are frozen at bind pose 
values. Subsequent updates take the else branch which only refreshes the instance 
transforms.

I believe RCR uses these bounds for culling. So as the skeleton animates and the rendered 
geometry extends beyond the bind pose AABB, parts could get incorrectly culled. The "fix constantly recomputing bounding boxes of skinned meshes which is not compliant with the visionOS implementation" and freezing on the web process side applies to the framing transform only. This per part AABB issue on the other hand might be a real pre-existing bug?

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -355,6 +355,42 @@ internal func logInfo(_ info: String) {
     Logger.modelGPU.info("\(info)")
 }
 
+// Mirrors RemoteMeshProxy.cpp::computeMinAndMaxExtents: computes the skin matrix that places the
+// mesh in world space for bounds purposes (using root joint 0 only to avoid over-inflation).
```
ruthvikkonda: Nit: Since index 0 being the root is more of a convention rather than an enforced part of the schema, is it worth adding a link to a spec/reference here citing the convention?

ruthvikkonda (commented):
(not an official webkit reviewer)

Approach looks good! Left a few comments

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -806,14 +842,21 @@ extension WKBridgeReceiver {
                                 )
                             )
 
+                            let skinMatrix = computeRootSkinMatrix(meshData.deformationData?.skinningData)
+                            let (effectiveMin, effectiveMax) = computeSkinningAABB(
+                                skinMatrix,
+                                meshData.parts[partIndex].boundsMin,
```
etiennesegonzac: nit: don't we have `meshData.parts[partIndex]` already bound to `part`?

etiennesegonzac (commented):
great review @ruthvikkonda!

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
mwyrzykowski: I probably should remove the comment to avoid it getting out of date

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
mwyrzykowski: oh yes, `(partIndex, part) in meshData.parts.enumerated()` would do that 🙂

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
mwyrzykowski: This is a good point, perhaps we should detect root joints instead as noted in:
https://openusd.org/release/api/_usd_skel__schemas.html#UsdSkel_JointHierarchy

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
mwyrzykowski: Ideally they are even the same function, maybe I can do that too

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
mwyrzykowski: I think I should try and fix this, thank you for noticing

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
mwyrzykowski: do you think we should pass:
```
boundsMin: -.one * Float.greatestFiniteMagnitude,
boundsMin: +.one * Float.greatestFiniteMagnitude,
```
instead? I think that should be fine

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
ruthvikkonda: Sounds good to me. Oh there can be multiple root joints! Glad we found this

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
ruthvikkonda: Thinking about this again. I wonder if we even need to do anything. We compute per parts bounds once at creation time and after this, even with later animation, I don't think we move our camera frustum at all. So the initial bind pose bounds always stay within the frustum no matter what animation takes place, so culling never occurs anyway. The only situation still worth considering is where the stale bounds lead to unnecessary draw calls.

Imagine a scenario where animation causes big translation where the stale bounds are still within the frustum but now the animated model is outside it. In that case, we wouldn't cull and would still issue draw calls incorrectly. This would also occur with the infinite bounds approach. 

The ideal way to solve this would be to do per vertex skinning each frame and update a tight AABB for each mesh part but that seems too expensive. 

Some approximation ideas in decreasing accuracy:

1. Union of per-joint transformed AABBs (tracks all joint motion; requires load-time per-joint AABB computation)
2. Dominant joint per mesh part (tracks part's main joint motion; approximate for multi-joint parts)
3. Root joint matrix per part each frame in the else branch (tracks only root translation; no-op for in-place animations)

Maybe Option 3 is worth doing? We could do that in the else branch and update each frame. It would help the specific case of characters translated via root-joint skinning. But it's a no-op for typical in-place character animation, so the value is narrow. If we want to just keep things simple, we could keep bind pose AABB as is or do infinite bounds approach (both are practically the same here I think), but technically those would be semantically incomplete.

mwyrzykowski: should it be outside the scope of this PR? The behavior is pre-existing, it would be good to have a repro case where we can definitively say there is an issue and now it is fixed. Similar to the chair USD file fixed by this issue

ruthvikkonda: Agreed, sounds good!

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
mwyrzykowski: I removed the Swift side computation so factoring this out probably is not needed, since its only used in a single place now

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
mwyrzykowski: Now it depends on `rootIdx` so I don't think we can pull it outside of the loop so easily

mwyrzykowski:
fyi @etiennesegonzac @ruthvikkonda  safer-cpp bot failure can be ignored, it's in WasmConstExprGenerator.cpp

 https://ews-build.s3-us-west-2.amazonaws.com/macOS-Safer-CPP-Checks-EWS/484206a9-103084/scan-build-output/StaticAnalyzerRegressions/WebKitLegacy/StaticAnalyzerReports/report-WasmConstExprGenerator.cpp-unknown-0-d9f7bb.html#EndPath

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -355,6 +355,99 @@ internal func logInfo(_ info: String) {
...
+        let jointTokens: TokenArray = jointsAttr.get()
+    else { return [] }
+
+    return rootJointIndices(from: jointTokens.map(\.description))
+}
+
+// Pure string computation: returns the indices of joints whose parent path does
+// not appear in the joints array. USD requires parents before children, so root
+// joints always appear before their descendants but there can be multiple roots
+// (e.g. ["A", "A/B", "C", "C/D/E"] has roots at indices 0 and 2).
```
ruthvikkonda, line 409: I think this algorithm is wrong. The comment says ["A", "A/B", "C", "C/D/E"] should yield [0, 2], but running the function on that input returns [0, 2, 3]. I ran some local testing to confirm. "C/D/E" gets classified as a root because its immediate parent "C/D" isn't in the set — but "C" is, and per the [USD joint hierarchy spec](https://openusd.org/release/api/_usd_skel__schemas.html#UsdSkel_JointHierarchy), the parent is the longest array entry that's an ancestor, not just the immediate parent.

The below solution should fix it. Basically, walk ancestors until we find one or run out:

```
private func rootJointIndices(from joints: [String]) -> [UInt32] {
    let jointSet = Set(joints)
    return joints.enumerated().compactMap { index, joint -> UInt32? in
        var path = joint
        while let slash = path.lastIndex(of: "/") {
            path = String(path[..<slash])
            if jointSet.contains(path) { return nil }
        }
        return UInt32(index)
    }
}
```

mwyrzykowski: Thank you! I will try it out

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
```diff
@@ -87,11 +87,39 @@ static std::pair<simd_float4, simd_float4> computeMinAndMaxCorners(const Vector<
...
+            bool hasRootIndices = !rootIndices.isEmpty();
+            size_t iterCount = hasRootIndices ? rootIndices.size() : 1;
+
+            for (auto& transform : instanceTransforms) {
+                for (size_t r = 0; r < iterCount; ++r) {
+                    uint32_t rootIdx = hasRootIndices ? rootIndices[r] : 0;
+                    if (rootIdx >= skinning.jointTransforms.size())
+                        continue;
+                    simd_float4x4 invBind = (rootIdx < skinning.inverseBindPoses.size()) ? (simd_float4x4)skinning.inverseBindPoses[rootIdx] : matrix_identity_float4x4;
+                    skinMatrix = simd_mul(simd_mul((simd_float4x4)skinning.jointTransforms[rootIdx], invBind), geomBind);
```
ruthvikkonda: Nit:

> Now it depends on rootIdx so I don't think we can pull it outside of the loop so easily

It depends on rootIdx but not on the instance transform! So we can at least pull it out of the inner instanceTransforms loop. Otherwise, we'd be recomputing skinMatrix O(|instanceTransforms| × |roots|) times.

Something like below could work:

```
Vector<simd_float4x4> rootSkinMatrices;
rootSkinMatrices.reserveInitialCapacity(iterCount);
for (size_t r = 0; r < iterCount; ++r) {
    uint32_t rootIdx = hasRootIndices ? rootIndices[r] : 0;
    if (rootIdx >= skinning.jointTransforms.size())
        continue;
    const simd_float4x4 invBind = (rootIdx < skinning.inverseBindPoses.size())
        ? (simd_float4x4)skinning.inverseBindPoses[rootIdx]
        : matrix_identity_float4x4;
    rootSkinMatrices.append(simd_mul(simd_mul((simd_float4x4)skinning.jointTransforms[rootIdx], invBind), geomBind));
}

for (auto& transform : instanceTransforms) {
    for (const auto& skinMatrix : rootSkinMatrices) {
        const simd_float4x4 worldTransform = simd_mul((simd_float4x4)transform, skinMatrix);
        // transform 8 corners, update min/max
    }
}
```

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
```diff
@@ -87,11 +87,39 @@ static std::pair<simd_float4, simd_float4> computeMinAndMaxCorners(const Vector<
         corners[6] = simd_make_float3(part.boundsMin.x, part.boundsMax.y, part.boundsMax.z);
         corners[7] = simd_make_float3(part.boundsMax.x, part.boundsMax.y, part.boundsMax.z);
 
+        simd_float4x4 skinMatrix = matrix_identity_float4x4;
+        if (deformationData && deformationData->skinningData && !deformationData->skinningData->jointTransforms.isEmpty()) {
+            const auto& skinning = *deformationData->skinningData;
+            const simd_float4x4 geomBind = skinning.geometryBindTransform;
+            // Fall back to index 0 when rootJointIndices is empty (single-root assumption).
+            auto& rootIndices = skinning.rootJointIndices;
```
ruthvikkonda: Nit: Should we mark `invBind`, `hasRootIndices`, `iterCount`, `rootIndices` as `const`?

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
```diff
@@ -87,11 +87,39 @@ static std::pair<simd_float4, simd_float4> computeMinAndMaxCorners(const Vector<
...
+            bool hasRootIndices = !rootIndices.isEmpty();
+            size_t iterCount = hasRootIndices ? rootIndices.size() : 1;
+
+            for (auto& transform : instanceTransforms) {
+                for (size_t r = 0; r < iterCount; ++r) {
+                    uint32_t rootIdx = hasRootIndices ? rootIndices[r] : 0;
+                    if (rootIdx >= skinning.jointTransforms.size())
+                        continue;
+                    simd_float4x4 invBind = (rootIdx < skinning.inverseBindPoses.size()) ? (simd_float4x4)skinning.inverseBindPoses[rootIdx] : matrix_identity_float4x4;
```
ruthvikkonda: Nit: the 4 C-style (simd_float4x4) casts here and below are not needed. WebModel::Float4x4 has implicit conversion operators to simd_float4x4.

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -355,6 +355,99 @@ internal func logInfo(_ info: String) {
...
+            let parentPath = String(joint[joint.startIndex..<lastSlash])
+            return jointSet.contains(parentPath) ? nil : UInt32(index)
+        }
+}
+
+// Transforms a local-space AABB by one or more skin matrices and returns the
+// union of the resulting world-space AABBs. Passing multiple matrices handles
+// multi-root skeletons correctly.
+private func computeSkinningAABB(
+    _ skinMatrices: [simd_float4x4],
```
ruthvikkonda, line 427: Nit: I don't know Swift style super well, but I'm wondering if these parameters should have labels too since the other functions above like `rootSkinMatrices` do?

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
mwyrzykowski: we can just switch the order of the outer and inner loops too, right?

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
mwyrzykowski: `rootSkinMatrices` should probably drop the label as it is obvious, but not completely sure

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
mwyrzykowski: I think at least this one is needed since its inside a ternary:
```
/Volumes/Data/OpenSource/Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp:102:92: error: conditional expression is ambiguous; 'const WebModel::Float4x4' can be converted to 'const simd_float4x4' and vice versa
  102 |                 const simd_float4x4 invBind = (rootIdx < skinning.inverseBindPoses.size()) ? skinning.inverseBindPoses[rootIdx] : matrix_identity_float4x4;
```
but I will switch to static_cast

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -355,6 +355,99 @@ internal func logInfo(_ info: String) {
...
+
+    guard let skelPath = skelPrimPath else { return [] }
+    let skelPrim = stage.prim(at: skelPath)
+    guard skelPrim.isValid,
+        let jointsAttr = skelPrim.attribute("joints"),
+        let jointTokens: TokenArray = jointsAttr.get()
+    else { return [] }
+
+    return rootJointIndices(from: jointTokens.map(\.description))
+}
```
ruthvikkonda, line 404: Nit: 

For skinned meshes whose joint transforms change each frame, `webUpdateMeshRequestFromUpdateMeshRequest` fires per mesh per frame, which means this function does a prim-walk + relationship lookup + attribute read + Set build every frame.

But the data being read here is the skeleton schema. `prim.relationship(named: "skel:skeleton")` and `skelPrim.attribute("joints")` are both fixed. Joint transforms change per frame, but the joint list/hierarchy is fixed once the stage is open.

So can we cache the result? Something like a [String: [UInt32]] keyed by the mesh prim path inside USDModelLoader would let subsequent calls return the cached [UInt32] directly. And we can invalidate on stage reload.

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
mwyrzykowski: Works great :)

Source/WebKit/WebProcess/GPU/graphics/Model/RemoteMeshProxy.cpp
```diff
@@ -87,11 +87,38 @@ static std::pair<simd_float4, simd_float4> computeMinAndMaxCorners(const Vector<
...
+                        simd_float4 corner4 = simd_make_float4(corners[j].x, corners[j].y, corners[j].z, 1.f);
+                        simd_float4 transformedCorner = simd_mul(worldTransform, corner4);
+                        minCorner4 = simd_min(transformedCorner, minCorner4);
+                        maxCorner4 = simd_max(transformedCorner, maxCorner4);
+                    }
+                }
+            }
+            continue; // instanceTransforms loop already handled above
+        }
+
```
ruthvikkonda, line 109: Nit: Loop swap works for the perf concern. But the inner corner transformation body is duplicated between the skinning branch and the non-skinning fallback. Both do "for each instance transform, transform 8 corners, update min/max." The skinning branch ends with continue; to skip the fallback for that part.

If we want to get rid of this duplication too, building a single Vector<simd_float4x4> of skin matrices upfront —non-skinning case fills it with one identity matrix — gets us to one unified inner loop:

```
Vector<simd_float4x4> rootSkinMatrices;
if (deformationData && deformationData->skinningData && !deformationData->skinningData->jointTransforms.isEmpty()) {
    const auto& skinning = *deformationData->skinningData;
    const simd_float4x4 geomBind = skinning.geometryBindTransform;
    const auto& rootIndices = skinning.rootJointIndices;
    const bool hasRootIndices = !rootIndices.isEmpty();
    const size_t iterCount = hasRootIndices ? rootIndices.size() : 1;
    rootSkinMatrices.reserveInitialCapacity(iterCount);
    for (size_t r = 0; r < iterCount; ++r) {
        uint32_t rootIdx = hasRootIndices ? rootIndices[r] : 0;
        if (rootIdx >= skinning.jointTransforms.size())
            continue;
        const simd_float4x4 invBind = (rootIdx < skinning.inverseBindPoses.size())
            ? static_cast<simd_float4x4>(skinning.inverseBindPoses[rootIdx])
            : matrix_identity_float4x4;
        rootSkinMatrices.append(simd_mul(simd_mul(skinning.jointTransforms[rootIdx], invBind), geomBind));
    }
} else {
    rootSkinMatrices.append(matrix_identity_float4x4);
}

for (auto& transform : instanceTransforms) {
    for (const auto& skinMatrix : rootSkinMatrices) {
        const simd_float4x4 worldTransform = simd_mul(transform, skinMatrix);
        for (int j = 0; j < 8; ++j) {
            simd_float4 corner4 = simd_make_float4(corners[j].x, corners[j].y, corners[j].z, 1.f);
            simd_float4 transformedCorner = simd_mul(worldTransform, corner4);
            minCorner4 = simd_min(transformedCorner, minCorner4);
            maxCorner4 = simd_max(transformedCorner, maxCorner4);
        }
    }
}
```

ruthvikkonda (commented):
(not an official webkit reviewer)
r = me

Thank you for addressing earlier comments! Looks a lot better! Left a few nits.

Also, one overall design consideration I was just thinking about (feel free to discard):

I wonder if we can actually just apply skinning to per part bounds on webcontent side and then send the bounds over that way instead of all the downstream logic. I think that would make things a lot cleaner. We have all the skinning data on webcontent side anyway since USDKit runs there. 

When we build the WKBridgeUpdateMesh, the per-part post-skin AABB could be computed there once and stamped into WKBridgeMeshPart.boundsMin/Max. This would change the field's contract from "authored static AABB" to "post-skin AABB" .

Then in GPU process code, we could drop the logic for skinning and feed bounds directly through. And even on WebContent side, in RemoteMeshProxy::computeMinAndMaxCorners, we could then use bounds directly and remove skinning logic.

One trade-off would be that the framing AABB derived in RemoteMeshProxy::computeMinAndMaxCorners would become a superset of today's bound when the instance transform has rotation/shear, because we'd be re-bounding an intermediate axis-aligned post-skin box rather than transforming original mesh corners through the combined (instanceTransform × skinMatrix).

mwyrzykowski:
> ### iOS Safer C++ Build [#25851](https://ews-build.webkit.org/#/builders/174/builds/25851) ([ad31009](https://github.com/WebKit/WebKit/commit/ad31009db350d28891d1489aee495610c8f02c9d))
> ❌ Found [1 failing file with 1 issue](https://ews-build.s3-us-west-2.amazonaws.com/Apple-iOS-26-Safer-CPP-Checks-EWS/ad31009d-25851/scan-build-output/new-results.html). Please address these issues before landing. See [WebKit Guidelines for Safer C++ Programming](https://github.com/WebKit/WebKit/wiki/Safer-CPP-Guidelines). (cc @rniwa)

Looks like the `WasmConstExprGenerator.cpp` failure again, unrelated to this PR

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
mwyrzykowski: I resolved this via a cache

mwyrzykowski: Thank you for the suggestion :) `rootJointIndicesCache`

Source/WebKit/GPUProcess/graphics/Model/USDModel.swift
```diff
@@ -811,15 +905,22 @@ extension WKBridgeReceiver {
                                 )
                             )
 
+                            let skinMatrices = meshData.deformationData?.skinningData.map { rootSkinMatrices($0) } ?? []
```
etiennesegonzac: can we move this outside of the loop?

etiennesegonzac (approved):

ruthvikkonda:
r=me

Looks great, thank you!

heycam:
For expediency, I will put this in the merge queue, and @mwyrzykowski you can address Etienne's review comment in a followup.

heycam:
(Except it has merge conflicts...)

mwyrzykowski:
I will try to resolve the conflicts today

mwyrzykowski:
> (not an official webkit reviewer) r = me
> 
> Thank you for addressing earlier comments! Looks a lot better! Left a few nits.
> 
> Also, one overall design consideration I was just thinking about (feel free to discard):
> 
> I wonder if we can actually just apply skinning to per part bounds on webcontent side and then send the bounds over that way instead of all the downstream logic. I think that would make things a lot cleaner. We have all the skinning data on webcontent side anyway since USDKit runs there.
> 
> When we build the WKBridgeUpdateMesh, the per-part post-skin AABB could be computed there once and stamped into WKBridgeMeshPart.boundsMin/Max. This would change the field's contract from "authored static AABB" to "post-skin AABB" .
> 
> Then in GPU process code, we could drop the logic for skinning and feed bounds directly through. And even on WebContent side, in RemoteMeshProxy::computeMinAndMaxCorners, we could then use bounds directly and remove skinning logic.
> 
> One trade-off would be that the framing AABB derived in RemoteMeshProxy::computeMinAndMaxCorners would become a superset of today's bound when the instance transform has rotation/shear, because we'd be re-bounding an intermediate axis-aligned post-skin box rather than transforming original mesh corners through the combined (instanceTransform × skinMatrix).

It's a good idea - I filed https://bugs.webkit.org/show_bug.cgi?id=315409

Please feel free to take the re-factor

PR #63885, opened 2026-04-29
[WebXR Layers][OpenXR] WebKit crashes in immersiveweb's quad-ab-test.html

svillar:
#### e5dac2f2642fedd5c1a289f5491875972e666ebd
<pre>
[WebXR Layers][OpenXR] WebKit crashes in immersiveweb&apos;s quad-ab-test.html
<a href="https://bugs.webkit.org/show_bug.cgi?id=313625">https://bugs.webkit.org/show_bug.cgi?id=313625</a>

Reviewed by Dan Glastonbury.

The immersiveweb&apos;s quad AB example is a test that demonstrates rendering
the same picture using a quad layer or embedded in the projection layer.
The UIProcess was crashing soon after starting the WebXR session (it
requires the not landed yet cylinder layer support from PR#63785).

The sequence was more or less the following:
1. Layer A is created, and updateRenderState() is
   called so the list of active layers becomes [ A ]
2. Another layer B is created, but updateRenderState is not called yet.
3. populateFrameData() called startFrame() over all *created* layers,
   i.e. [A, B]
4. submitFrame() was later called with only [ A ] as updateRenderState()
   was not called yet
5. On the render thread endFrame() processes layer A (calling
   releaseImage()). Layer B was never submitted, so B-&gt;endFrame() is
   never called.
6. In the next frame populateFrameData() was calling startFrame over [A,
   B] again, but since B&apos;s texture was not released, the call to
   acquireImage() failed.

As it can be seen the problem was that startFrame() was always called
over all created layers, but endFrame() was created only over the active
layers in WebXR. OpenXR should only operate on the layers selected by
WebXR.

We actually tried to solve this previously adding an acquiredTexture
check in endFrame() but that was wrong because it was indeed fixing the
crash (the sympthon) but not the proper flow of calls (ensuring that
startFrame/endFrame -acquireImage/releaseImage- happen always in pairs).
This patch replaced that with ASSERTs.

Now we keep a list of active textures on the PlatformXR side which is
updated everytime the user calls updateRenderState() in WebXR. Actually
we need to keep two lists in order to use the same list during the whole
lifetime of a frame while allowing the code to update that list for the
next one.

Last but not least, this allowed us to unveil a bug when setting
passthroughFullyObscured. We were resetting it to false if no
requestData was provided, That&apos;s wrong, requestData is only received
when the render state changes, so if it&apos;s null then it means that
nothing should change and that it should keep the previous value.

No new tests as this can only be reproduced with an actual OpenXR
runtime runing.

* Source/WebCore/Modules/webxr/WebXRLayer.h:
* Source/WebCore/Modules/webxr/WebXRSession.cpp:
(WebCore::WebXRSession::applyPendingRenderState):
* Source/WebCore/Modules/webxr/WebXRWebGLLayer.cpp:
(WebCore::createOpaqueFramebuffer):
(WebCore::WebXRWebGLLayer::create):
(WebCore::WebXRWebGLLayer::layerHandle const):
* Source/WebCore/Modules/webxr/WebXRWebGLLayer.h:
* Source/WebCore/Modules/webxr/XRCompositionLayer.cpp:
(WebCore::XRCompositionLayer::layerHandle const):
* Source/WebCore/Modules/webxr/XRCompositionLayer.h:
* Source/WebCore/platform/xr/PlatformXR.h:
* Source/WebKit/Shared/XR/PlatformXR.serialization.in:
* Source/WebKit/UIProcess/XR/openxr/OpenXRLayer.cpp:
* Source/WebKit/UIProcess/XR/openxr/PlatformXROpenXR.cpp:
(WebKit::OpenXRCoordinator::scheduleAnimationFrame):
(WebKit::OpenXRCoordinator::populateFrameData):
(WebKit::OpenXRCoordinator::beginFrame):
(WebKit::OpenXRCoordinator::endFrame):

Canonical link: <a href="https://commits.webkit.org/312422@main">https://commits.webkit.org/312422@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/72f97d19b53a8acdb36ca872ee3a0bab0bd1b67d

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/159452 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/32881 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/25988 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/168284 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/113832 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/161321 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/32948 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/32869 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/123547 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/86721 "1 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/162409 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/25791 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/143223 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/104210 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/24851 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/23308 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/16055 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/134542 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/21003 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/170776 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/16810 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/22629 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/131753 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/32569 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/27375 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/131866 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35768 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/32513 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/142796 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/90650 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/26526 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/19605 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/32024 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/98476 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/31544 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/31817 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/31699 "Built successfully") | | |

### DIFF
--- a/Source/WebKit/UIProcess/XR/openxr/PlatformXROpenXR.cpp
+++ b/Source/WebKit/UIProcess/XR/openxr/PlatformXROpenXR.cpp
@@ -65,6 +65,8 @@ struct OpenXRCoordinator::RenderState {
     PlatformXR::Device::RequestFrameCallback onFrameUpdate;
     XrFrameState frameState;
     bool passthroughFullyObscured { false };
+    Vector<PlatformXR::LayerHandle> activeLayerHandles;
+    Vector<PlatformXR::LayerHandle> currentFrameLayerHandles;
 #if ENABLE(WEBXR_HIT_TEST)
     HashMap<PlatformXR::HitTestSource, UniqueRef<PlatformXR::HitTestOptions>> hitTestSources;
     HashMap<PlatformXR::TransientInputHitTestSource, UniqueRef<PlatformXR::TransientInputHitTestOptions>> transientInputHitTestSources;
@@ -429,7 +431,10 @@ void OpenXRCoordinator::scheduleAnimationFrame(WebPageProxy& page, std::optional
             }
 
             active.renderQueue->dispatch([this, renderState = active.renderState, requestData = WTF::move(requestData), onFrameUpdateCallback = WTF::move(onFrameUpdateCallback)]() mutable {
-                renderState->passthroughFullyObscured = requestData ? requestData->isPassthroughFullyObscured : false;
+                if (requestData) {
+                    renderState->passthroughFullyObscured = requestData->isPassthroughFullyObscured;
+                    renderState->activeLayerHandles = requestData->activeLayerHandles;
+                }
                 renderState->onFrameUpdate = WTF::move(onFrameUpdateCallback);
                 renderLoop(renderState);
             });
@@ -1104,6 +1109,8 @@ PlatformXR::FrameData OpenXRCoordinator::populateFrameData(Box<RenderState> rend
         frameData.floorTransform = XrIdentityPose();
 
     for (auto& layer : m_layers) {
+        if (!renderState->currentFrameLayerHandles.contains(layer.key))
+            continue;
         auto layerData = layer.value->startFrame();
         if (layerData) {
             auto layerDataRef = makeUniqueRef<PlatformXR::FrameData::LayerData>(WTF::move(*layerData));
@@ -1224,6 +1231,7 @@ void OpenXRCoordinator::beginFrame(Box<RenderState> renderState)
 
     // We should not directly use renderState->frameState in the xrWaitFrame() in order not to override the previous (ongoing) value.
     renderState->frameState = frameState;
+    renderState->currentFrameLayerHandles = renderState->activeLayerHandles;
 
     createReferenceSpacesIfNeeded(renderState);
     PlatformXR::FrameData frameData = populateFrameData(renderState);
@@ -1247,6 +1255,8 @@ void OpenXRCoordinator::endFrame(Box<RenderState> renderState, Vector<PlatformXR
 
     Vector<XrCompositionLayerBaseHeader*> frameEndLayers;
     for (auto& layer : layers) {
+        ASSERT(renderState->currentFrameLayerHandles.contains(layer.handle));
+
         auto it = m_layers.find(layer.handle);
         if (it == m_layers.end()) {
             LOG(XR, "Didn't find a OpenXRLayer with %d handle", layer.handle);
--- a/Source/WebCore/Modules/webxr/WebXRLayer.h
+++ b/Source/WebCore/Modules/webxr/WebXRLayer.h
@@ -49,6 +49,7 @@ public:
 
     virtual void startFrame(PlatformXR::FrameData&) = 0;
     virtual PlatformXR::DeviceLayer endFrame() = 0;
+    virtual PlatformXR::LayerHandle layerHandle() const = 0;
 
     virtual bool isWebXRWebGLLayer() const { return false; }
     virtual bool isXRCompositionLayer() const { return false; }
--- a/Source/WebCore/Modules/webxr/WebXRSession.cpp
+++ b/Source/WebCore/Modules/webxr/WebXRSession.cpp
@@ -585,9 +585,19 @@ void WebXRSession::applyPendingRenderState()
         m_activeRenderState->setOutputCanvas(nullptr);
     }
 
-    m_requestData = { {
+    m_requestData = {
         .isPassthroughFullyObscured = m_activeRenderState->passthroughFullyObscured().value_or(false),
-        .depthRange = PlatformXR::DepthRange { static_cast<float>(m_activeRenderState->depthNear()), static_cast<float>(m_activeRenderState->depthFar()) } }}; // NOLINT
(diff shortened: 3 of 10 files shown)


Source/WebKit/UIProcess/XR/openxr/PlatformXROpenXR.cpp
```diff
@@ -1104,6 +1109,8 @@ PlatformXR::FrameData OpenXRCoordinator::populateFrameData(Box<RenderState> rend
         frameData.floorTransform = XrIdentityPose();
 
     for (auto& layer : m_layers) {
+        if (!renderState->currentFrameLayerHandles.contains(layer.key))
+            continue;
```
djg, line 1113: Is there a guarantee that `updateRenderState()` is always called before the first frame that has layers? If `requestData` is null on the first `scheduleAnimationFrame` call, `activeLayerHandles`, and `currentFrameLayerHandles` will be empty, and `populateFrameData()` will skip all layers. That appears correct for "no active layers = nothing to render", but I wanted to confirm there's no path where layers are active but `activeLayerHandles` hasn't been populated yet.

djg (approved):
:shipit:

Source/WebKit/UIProcess/XR/openxr/PlatformXROpenXR.cpp
svillar: We only provide the list of active layers to OpenXR whenever `requestFrame()` is called with a non null `requestData`. And that only happens after the user has called `updateRenderState()`. It's indeed possible to create layers and not use them at all, but that's fine from the spec POV. 

The `activeRenderState` is only modified in `applyPendingRenderState()` which is where we build the list of active layers as well. And the key is that we build that list before calling `requestFrame()` so WebXR will always have the most updated list of layers, even before the platform code. It seems to me that is not possible to get a list of handles from the platform code that do not correspond to the active layers in the active render state.

But I am not sure I completely understood your doubts so feel free to ask as many questions as you have. I can hold the horses for a while :)

PR #63914, opened 2026-04-29
Web Inspector: Record first input far too small within Graphics tab

rcaliman-apple:
#### b5ebc12da8948321a58cf8aef37923e61692c072
<pre>
Web Inspector: Record first input far too small within Graphics tab
<a href="https://bugs.webkit.org/show_bug.cgi?id=297082">https://bugs.webkit.org/show_bug.cgi?id=297082</a>
<a href="https://rdar.apple.com/157787230">rdar://157787230</a>

Reviewed by Qianlang Chen.

The custom autosizing of the input field based on its content is replaced
with the standard CSS `field-sizing: content` property.

Also, increase the `min-width` to account for the number input increment/decrement
buttons in contemporary macOS. This is only applicable when the placeholder is set.
When a value is set, the `field-sizing: content` appropriately sizes the element.

The UX of the number input field was also broken making it impossible to type more
than one character at a time because on every keypress the DOM structure for the label,
which included the input field itself, would get rebuilt to represent the plural form
for the number of frames in the label text. This caused input focus to be lost on every keypress.

Previously, this was worked around using programmatic selection. That does not work anymore after
<a href="https://commits.webkit.org/247274@main">https://commits.webkit.org/247274@main</a> which made WebKit spec compliant and removed
support for `selectionStart` / `selectionEnd` from number input fields.

This patch changes the string used for the label so that it is no longer necessary
to account for pluralization and no need to update the DOM structure at runtime.

* Source/WebInspectorUI/UserInterface/Views/CanvasOverviewContentView.css:
(.navigation-bar &gt; .item.canvas-recording-auto-capture &gt; label &gt; input):
* Source/WebInspectorUI/UserInterface/Views/CanvasOverviewContentView.js:
(WI.CanvasOverviewContentView):
(WI.CanvasOverviewContentView.prototype._handleRecordingAutoCaptureInput):
(WI.CanvasOverviewContentView.prototype._handleCanvasRecordingAutoCaptureFrameCountChanged):
(WI.CanvasOverviewContentView.prototype._updateRecordingAutoCaptureCheckboxLabel): Deleted.
(WI.CanvasOverviewContentView.prototype._updateRecordingAutoCaptureInputElementSize): Deleted.

Canonical link: <a href="https://commits.webkit.org/312433@main">https://commits.webkit.org/312433@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/98493e67062147b74a650e55a7dbd516936f3b20

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/159726 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/33193 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/26300 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/168582 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/114107 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/c7ba4b44-6372-432b-86ca-132a45d4290a) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/33298 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/33197 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/123761 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/86840 "9 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/cdcbc992-5ab4-4d22-b7b8-aa61134f7abe) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/162684 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/26024 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/143462 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/104404 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/01f3c8fc-d8d9-4654-a75d-237b0d496045) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/25075 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/23546 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/16347 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/134767 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/21234 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/171075 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/17094 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/22870 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/132015 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/32872 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/27623 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/132074 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/35779 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/32857 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/143028 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/90940 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/26696 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/19841 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/32366 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/98762 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/31863 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/32110 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/32014 "Built successfully") | | | |

### DIFF
--- a/Source/WebInspectorUI/UserInterface/Views/CanvasOverviewContentView.js
+++ b/Source/WebInspectorUI/UserInterface/Views/CanvasOverviewContentView.js
@@ -56,9 +56,14 @@ WI.CanvasOverviewContentView = class CanvasOverviewContentView extends WI.Collec
         this._recordingAutoCaptureNavigationItem.visibilityPriority = WI.NavigationItem.VisibilityPriority.Low;
         this._recordingAutoCaptureNavigationItem.addEventListener(WI.CheckboxNavigationItem.Event.CheckedDidChange, this._handleRecordingAutoCaptureCheckedDidChange, this);
 
-        let frameCount = this._updateRecordingAutoCaptureInputElementSize();
-        this._setRecordingAutoCaptureFrameCount(frameCount);
-        this._updateRecordingAutoCaptureCheckboxLabel(frameCount);
+        let fragment = document.createDocumentFragment();
+        String.format(WI.UIString("Recording frame count: %s", "Label for input for number of frames to record @ Canvas section of Graphics tab"), [this._recordingAutoCaptureFrameCountInputElement], String.standardFormatters, fragment, (a, b) => {
+            a.append(b);
+            return a;
+        });
+        this._recordingAutoCaptureNavigationItem.label = fragment;
+
+        this._setRecordingAutoCaptureFrameCount(this._recordingAutoCaptureFrameCountInputElementValue);
 
         importNavigationItem.addEventListener(WI.ButtonNavigationItem.Event.Clicked, this._handleImportButtonNavigationItemClicked, this);
 
@@ -176,28 +181,6 @@ WI.CanvasOverviewContentView = class CanvasOverviewContentView extends WI.Collec
         WI.canvasManager.setRecordingAutoCaptureFrameCount(enabled, frameCount);
     }
 
-    _updateRecordingAutoCaptureCheckboxLabel(frameCount)
-    {
-        let active = document.activeElement === this._recordingAutoCaptureFrameCountInputElement;
-        let selectionStart = this._recordingAutoCaptureFrameCountInputElement.selectionStart;
-        let selectionEnd = this._recordingAutoCaptureFrameCountInputElement.selectionEnd;
-        let direction = this._recordingAutoCaptureFrameCountInputElement.direction;
-
-        let label = frameCount === 1 ? WI.UIString("Record first %s frame") : WI.UIString("Record first %s frames");
-        let fragment = document.createDocumentFragment();
-        String.format(label, [this._recordingAutoCaptureFrameCountInputElement], String.standardFormatters, fragment, (a, b) => {
-            a.append(b);
-            return a;
-        });
-        this._recordingAutoCaptureNavigationItem.label = fragment;
-
-        if (active) {
-            this._recordingAutoCaptureFrameCountInputElement.selectionStart = selectionStart;
-            this._recordingAutoCaptureFrameCountInputElement.selectionEnd = selectionEnd;
-            this._recordingAutoCaptureFrameCountInputElement.direction = direction;
-        }
-    }
-
     get _recordingAutoCaptureFrameCountInputElementValue()
     {
         return parseInt(this._recordingAutoCaptureFrameCountInputElement.value);
@@ -209,19 +192,6 @@ WI.CanvasOverviewContentView = class CanvasOverviewContentView extends WI.Collec
             this._recordingAutoCaptureFrameCountInputElement.value = frameCount;
 
         this._recordingAutoCaptureFrameCountInputElement.placeholder = frameCount;
-    }
-
-    _updateRecordingAutoCaptureInputElementSize()
-    {
-        let frameCount = this._recordingAutoCaptureFrameCountInputElementValue;
-        if (isNaN(frameCount) || frameCount < 0) {
-            frameCount = 0;
-            this._recordingAutoCaptureFrameCountInputElementValue = frameCount;
-        }
-
-        this._recordingAutoCaptureFrameCountInputElement.autosize();
-
-        return frameCount;
     }
 
     _addSavedRecording(recording)
@@ -253,7 +223,12 @@ WI.CanvasOverviewContentView = class CanvasOverviewContentView extends WI.Collec
 
     _handleRecordingAutoCaptureInput(event)
     {
-        let frameCount = this._updateRecordingAutoCaptureInputElementSize();
+        let frameCount = this._recordingAutoCaptureFrameCountInputElementValue;
+        if (isNaN(frameCount) || frameCount < 0) {
+            frameCount = 0;
+            this._recordingAutoCaptureFrameCountInputElementValue = frameCount;
+        }
+
         this._recordingAutoCaptureNavigationItem.checked = !!frameCount;
 
         this._setRecordingAutoCaptureFrameCount(frameCount);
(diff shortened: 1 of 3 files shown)


rcaliman-apple:
Screen recordings

|Before|After|
|---|---|
|https://github.com/user-attachments/assets/37b49c67-3764-4949-8a74-7f1d96b2a19d|https://github.com/user-attachments/assets/9174d452-3dae-426e-bfa3-64313e76bef9|

Source/WebInspectorUI/UserInterface/Views/CanvasOverviewContentView.js
rcaliman-apple, line 61: No need to sanitize input value in the constructor right after building the input field with the value from `WI.settings.canvasRecordingAutoCaptureFrameCount`. We can just use the value from the setting because it has been sanitized before being set (or is the default).

the-chenergy:
Commit message nit: "ofr" -> "for" near the end of the paragraph about programmatic selection

the-chenergy (approved):

Source/WebInspectorUI/UserInterface/Views/CanvasOverviewContentView.js
dcrousso, line 66: we only ever pass `_recordingAutoCaptureFrameCountInputElementValue` so i think we should just remove the argument altogether

dcrousso: ditto

dcrousso: id much rather us use `String.format` as that will handle any other specifiers (e.g. `%d`), as well as if the localized string is missing an expected `%s` (or any other potential things we may add to it in the future that the developer would then also have to remember to change here too)

as such, please revert this back to using `String.format`

dcrousso: if the issue is that it removes the `<input>` from the DOM (and therefore messes with the current cursor position) then id suggest either rephrasing the localized string so that it doesn't require using `String.format` in the first place or to modify `WI.CheckboxNavigationItem` such that there's a way to directly manipulate the children without removal

dcrousso, line 249: should we drop this `if` too?

rcaliman-apple: That's true for most of the cases, aside for the sanitization in `_handleRecordingAutoCaptureInput`.

rcaliman-apple: Yes, the issue is replacing the `WI.CheckboxNavigationItem` label, which contains the `<input>`, that causes it to lose focus and makes it impossible to type more than one character at a time. 

`%s` will never go away in this format because it represents the `<input>`. It might move around the string, but it will always exist.

> (or any other potential things we may add to it in the future that the developer would then also have to remember to change here too)

I think it's unlikely we'd change this label

rcaliman-apple: I investigated, but there's a little UX tweak attached to it.

The input value can be `NaN` when typing non-numeric values which will disable the checkbox. The input will show `0` as the placeholder. 

Enabling the checkbox in this state will attempt to set the value, and use `1` as a fallback:
https://github.com/WebKit/WebKit/blob/95c755e8936a2d310aa64716ea89d226cc2702ee/Source/WebInspectorUI/UserInterface/Views/CanvasOverviewContentView.js#L171-L172

This is persisted in the setting in `WI.CanvasManager.prototype.setRecordingAutoCaptureFrameCount()` and the change is reflected back to the input field value via this event handler.

(Setting the value to `0` disables recording):
https://github.com/WebKit/WebKit/blob/92693c2591b2db3161960f6b03fbe74ad9cfe55e/Source/WebCore/inspector/agents/InspectorCanvasAgent.cpp#L227-L231
I'll keep it.

rcaliman-apple: >id suggest either rephrasing the localized string so that it doesn't require using String.format in the first place

The cleanest solution was this. Updated the patch and removed the dynamic label update.

Image

`String.format` will always move the `<input>` via the `Node.prototype.append()` used in the callback.

PR #63930, opened 2026-04-29
[Site Isolation] CSP upgrade-insecure-requests not applied when cross-process iframe navigates top frame

roberto-apple:
#### 888c80c640da3d5a5224f727e3e45fbd6fbdf5b8
<pre>
[Site Isolation] CSP upgrade-insecure-requests not applied when cross-process iframe navigates top frame
<a href="https://bugs.webkit.org/show_bug.cgi?id=313690">https://bugs.webkit.org/show_bug.cgi?id=313690</a>
<a href="https://rdar.apple.com/175889340">rdar://175889340</a>

Reviewed by Alex Christensen.

When a cross-origin iframe navigates window.top.location.href to an http:// URL under site isolation,
the parent&apos;s upgrade-insecure-requests CSP is never applied. The upgrade normally happens in
FrameLoader::changeLocation() using the target frame&apos;s document CSP, but the site isolation path
bypasses that code entirely and lands in a provisional page where no document exists to check.

Sync each frame&apos;s CSP navigation upgrade origins to its WebFrameProxy in the UIProcess. The set is
sent in DidCommitLoadForFrame when the directive arrives via response header, or via a new
DidChangeCSPOriginsThatUpgradeInsecureNavigations IPC to WebFrameProxy when added dynamically via
meta tag. Then in receivedNavigationActionPolicyDecision, consult the set and upgrade the URL before
process routing.

* Source/WebCore/loader/LocalFrameLoaderClient.h:
* Source/WebCore/page/csp/ContentSecurityPolicy.cpp:
(WebCore::ContentSecurityPolicy::setUpgradeInsecureRequests):
(WebCore::ContentSecurityPolicy::inheritInsecureNavigationRequestsToUpgradeFromOpener):
(WebCore::ContentSecurityPolicy::setInsecureNavigationRequestsToUpgrade):
(WebCore::ContentSecurityPolicy::notifyInsecureNavigationRequestsToUpgradeChanged const):
* Source/WebCore/page/csp/ContentSecurityPolicy.h:
* Source/WebCore/platform/network/ResourceRequestBase.h:
* Source/WebKit/UIProcess/API/APINavigation.cpp:
(API::Navigation::upgradeCurrentInsecureRequest):
* Source/WebKit/UIProcess/API/APINavigation.h:
(WebKit::ProvisionalPageProxy::didCommitLoadForFrame):
* Source/WebKit/UIProcess/ProvisionalPageProxy.h:
(WebKit::WebFrameProxy::didCommitLoad):
(WebKit::WebFrameProxy::commitProvisionalFrame):
(WebKit::WebFrameProxy::didChangeCSPOriginsThatUpgradeInsecureNavigations):
(WebKit::WebFrameProxy::cspOriginsThatUpgradeInsecureNavigations const):
(WebKit::WebFrameProxy::setCSPOriginsThatUpgradeInsecureNavigations):
* Source/WebKit/UIProcess/WebFrameProxy.messages.in:
(WebKit::WebPageProxy::didCommitLoadForFrame):
* Source/WebKit/WebProcess/WebCoreSupport/WebLocalFrameLoaderClient.cpp:
(WebKit::WebLocalFrameLoaderClient::dispatchDidChangeCSPOriginsThatUpgradeInsecureNavigations):
(WebKit::WebLocalFrameLoaderClient::dispatchDidCommitLoad):
* Source/WebKit/WebProcess/WebCoreSupport/WebLocalFrameLoaderClient.h:

Canonical link: <a href="https://commits.webkit.org/314221@main">https://commits.webkit.org/314221@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/585d874165182671fe5b2203f13d54718243daca

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/165447 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/38937 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/32518 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/174489 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/122702 "Built successfully") | ⏳ 🛠 ios-apple 
| [  ~~🧪 bindings~~](https://ews-build.webkit.org/#/builders/9/builds/167313 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/39273 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/38941 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/128571 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/90495 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/168406 "Passed tests") | [❌ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/30918 "Found 2 new test failures: fast/events/touch/ios/touch-event-regions-layer-tree/mousemove-mouseup.html fast/forms/ios/drag-range-track.html (failure)") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/148641 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/109096 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/29925 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/28556 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/22564 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/139820 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/26339 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/177013 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/29921 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/28077 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/136895 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/39006 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/32729 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/136831 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36888 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/39694 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/148135 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/102069 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/32103 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/25002 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/38204 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/111278 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/37601 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/3082 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/37847 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/37745 "Built successfully") | | | |


roberto-apple:
The ios-site-isolation/TestExpectations entry for this test remains as [ Failure ]. The two sub-tests that still timeout on iOS (iframe-top-navigation-no-upgrade-1, iframe-top-navigation-no-upgrade-2) seem to be pre-existing iOS site isolation failures. 

From what I gathered, opener.postMessage doesn't route correctly after a popup's top frame undergoes a process swap. This is unrelated to this fix and needs to be addressed separately.

Source/WebCore/loader/LocalFrameLoaderClient.h
```diff
@@ -397,4 +398,6 @@ class WEBCORE_EXPORT LocalFrameLoaderClient : public FrameLoaderClient {
     WeakRef<FrameLoader> m_loader;
 };
 
+inline void LocalFrameLoaderClient::didChangeInsecureNavigationRequestsToUpgrade(HashSet<SecurityOriginData>&&) { }
```
achristensen07: I think this would be better on line 178 like dispatchDidNavigateWithinPage.

Source/WebKit/UIProcess/WebFrameProxy.h
```diff
@@ -37,6 +37,7 @@
 #include <WebCore/PageIdentifier.h>
 #include <WebCore/ReferrerPolicy.h>
 #include <WebCore/ScriptExecutionContextIdentifier.h>
+#include <WebCore/SecurityOriginData.h>
```
achristensen07, line 40: This can be a forward declaration instead of including a header from this widely-included header, which slows down the build.

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -7586,6 +7599,12 @@ void WebPageProxy::updateOpener(IPC::Connection& connection, WebCore::FrameIdent
     });
 }
 
+void WebPageProxy::didChangeInsecureNavigationRequestsToUpgrade(WebCore::FrameIdentifier frameID, HashSet<WebCore::SecurityOriginData>&& insecureNavigationRequestsToUpgrade)
```
achristensen07: It seems like we could just send IPC to the WebFrameProxy instead of the WebPageProxy.

Source/WebCore/page/csp/ContentSecurityPolicy.cpp
```diff
@@ -1234,6 +1238,20 @@ HashSet<SecurityOriginData> ContentSecurityPolicy::takeNavigationRequestsToUpgra
...
+{
+    RefPtr document = dynamicDowncast<Document>(m_scriptExecutionContext.get());
+    if (!document)
+        return;
+    if (!document->settings().siteIsolationEnabled())
+        return;
+    RefPtr frame = document->frame();
+    if (!frame)
+        return;
+    frame->loader().client().didChangeInsecureNavigationRequestsToUpgrade(HashSet<SecurityOriginData> { m_insecureNavigationRequestsToUpgrade });
```
achristensen07: This looks like an unnecessary HashSet copy.  If we pass a const HashSet& then we can just send it to IPC without copying it.

Source/WebKit/WebProcess/WebCoreSupport/WebLocalFrameLoaderClient.h
```diff
@@ -110,6 +110,7 @@ class WebLocalFrameLoaderClient final : public WebCore::LocalFrameLoaderClient,
 #endif
     void dispatchDidChangeMainDocument() final;
     void dispatchWillChangeDocument(const URL& currentUrl, const URL& newUrl) final;
+    void didChangeInsecureNavigationRequestsToUpgrade(HashSet<WebCore::SecurityOriginData>&&) final;
```
achristensen07: This naming is a little odd.  It's not clear what requests need upgrading when we are sending a set of origins.  The naming leads me to expect a set of requests.  Maybe having "CSP" in the name would make it more clear what this is.

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -5434,6 +5434,19 @@ void WebPageProxy::receivedNavigationActionPolicyDecision(WebProcessProxy& proce
...
 
+    // Apply CSP upgrade-insecure-requests before computing Site. Only needed for remote-frame
+    // navigations. Same-process navigations are already upgraded in the WebProcess.
+    if (navigationAction->data().originatingFrameIsRemote) {
+        if (auto& upgradeSet = frame.insecureNavigationRequestsToUpgrade(); !upgradeSet.isEmpty()) {
+            URL url = navigation.currentRequest().url();
+            if (upgradeSet.contains(WebCore::SecurityOriginData::fromURL(url)) && ResourceRequestBase::upgradeInsecureRequest(url)) {
+                ResourceRequest upgraded = navigation.currentRequest();
+                upgraded.setURL(WTF::move(url));
+                navigation.setCurrentRequest(WTF::move(upgraded), std::nullopt);
```
achristensen07: It looks like all this can be replaced by ResourceRequest.upgradeInsecureRequest

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -5434,6 +5434,19 @@ void WebPageProxy::receivedNavigationActionPolicyDecision(WebProcessProxy& proce
     m_isLockdownModeExplicitlySet = (websitePolicies && websitePolicies->isLockdownModeExplicitlySet()) || m_configuration->isLockdownModeExplicitlySet();
     auto lockdownMode = (websitePolicies ? websitePolicies->lockdownModeEnabled() : shouldEnableLockdownMode()) ? WebProcessProxy::LockdownMode::Enabled : WebProcessProxy::LockdownMode::Disabled;
 
+    // Apply CSP upgrade-insecure-requests before computing Site. Only needed for remote-frame
+    // navigations. Same-process navigations are already upgraded in the WebProcess.
+    if (navigationAction->data().originatingFrameIsRemote) {
```
achristensen07: Can we use processInitiatingNavigation to determine this instead of passing this over IPC?  The caller also has processInitiatingNavigation

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -5434,6 +5434,19 @@ void WebPageProxy::receivedNavigationActionPolicyDecision(WebProcessProxy& proce
...
+        if (auto& upgradeSet = frame.insecureNavigationRequestsToUpgrade(); !upgradeSet.isEmpty()) {
+            URL url = navigation.currentRequest().url();
+            if (upgradeSet.contains(WebCore::SecurityOriginData::fromURL(url)) && ResourceRequestBase::upgradeInsecureRequest(url)) {
```
achristensen07: An empty set will never contain anything.  The isEmpty check seems redundant to me.

Source/WebKit/UIProcess/WebFrameProxy.h
```diff
@@ -370,6 +374,7 @@ class WebFrameProxy : public API::ObjectImpl<API::Object::Type::Frame>, public I
     WebCore::ScrollbarMode m_scrollingMode;
     std::optional<WebCore::DocumentSecurityPolicy> m_documentSecurityPolicy;
     RefPtr<WebCore::SecurityOrigin> m_documentSecurityOrigin;
+    HashSet<WebCore::SecurityOriginData> m_insecureNavigationRequestsToUpgrade;
```
achristensen07: I think this needs a more clear name.  See my later feedback on adding "CSP" or something.

achristensen07 (changes requested):

Source/WebCore/loader/LocalFrameLoaderClient.h
roberto-apple: I had it the other way because the style checker flagged me with `error: [build/webcore_export] Inline functions should not be in classes annotated with WEBCORE_EXPORT. Remove the macro from the class and apply it to each appropriate method, or move the inline function definition out-of-line.` when uploading PR. But it seems that many more other inline definitions get flagged as well.

Source/WebKit/UIProcess/WebFrameProxy.h
roberto-apple: Since I've added `HashSet<WebCore::SecurityOriginData> m_cspOriginsThatUpgradeInsecureNavigations;` member, full definition is needed.

achristensen07: That shouldn't be the case.  If we put setCSPOriginsThatUpgradeInsecureNavigations in the cpp it ought to work.  If not, we can worry about this later.  It's not terribly important for this PR.

Source/WebKit/WebProcess/WebCoreSupport/WebLocalFrameLoaderClient.cpp
```diff
@@ -671,8 +679,9 @@ void WebLocalFrameLoaderClient::dispatchDidCommitLoad(std::optional<HasInsecureC
 #endif
 
     RefPtr<Frame> coreLocalFrame = m_localFrame.ptr();
+    auto cspOriginsThatUpgradeInsecureNavigations = protect(protect(m_localFrame->document())->contentSecurityPolicy())->insecureNavigationRequestsToUpgrade();
```
achristensen07: I think this is also an unnecessary set copy.  If you change auto to auto& and remove the WTF::move when sending this IPC, it should remove the copy.

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -5434,6 +5434,18 @@ void WebPageProxy::receivedNavigationActionPolicyDecision(WebProcessProxy& proce
...
+    // Apply CSP upgrade-insecure-requests before computing Site. Only needed for remote-frame
+    // navigations. Same-process navigations are already upgraded in the WebProcess.
+    if (&processInitiatingNavigation != &frame.process()) {
+        auto& upgradeSet = frame.cspOriginsThatUpgradeInsecureNavigations();
+        URL url = navigation.currentRequest().url();
+        if (upgradeSet.contains(WebCore::SecurityOriginData::fromURL(url)) && ResourceRequestBase::upgradeInsecureRequest(url)) {
+            ResourceRequest upgraded = navigation.currentRequest();
+            upgraded.setURL(WTF::move(url));
+            navigation.setCurrentRequest(WTF::move(upgraded), std::nullopt);
+        }
```
achristensen07: I was thinking this could be written like this:
if (upgradeSet.contains(WebCore::SecurityOriginData::fromURL(url)))
    navigation.upgradeCurrentInsecureRequest()
Then add something like this:
void API::Navigation::upgradeCurrentInsecureRequest()
{
    m_currentRequest.upgradeInsecureRequest()
}
I think that would be slightly more clear.

achristensen07 (approved):

PR #63935, opened 2026-04-30
GlyphGeometryCache: allow caching width results that used fallback fonts if possible

vitorroriz:
#### 9bf2adeec6e951b643b52ed52663d6b35a20e5fb
<pre>
GlyphGeometryCache: allow caching width results that used fallback fonts if possible
<a href="https://rdar.apple.com/175896996">rdar://175896996</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=313699">https://bugs.webkit.org/show_bug.cgi?id=313699</a>

Reviewed by Alan Baradlay.

Previously, FontCascade::width() refused to cache into GlyphGeometryCache
when fallback fonts were used during text measurement. This could cause
repeated full shaping costs for text that triggers font fallback (e.g.
when the primary font doesn&apos;t cover all scripts in the content), even
when the caller never needed the fallback font set.

Canonical link: <a href="https://commits.webkit.org/312912@main">https://commits.webkit.org/312912@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/37abf7fa04176d8f94570fb08c033005f5643fb4

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/161217 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/34690 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/27791 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/170120 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/117588 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/effad7d8-6fc6-41c2-a85c-e59c8c582ada) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/163086 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/34791 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/34691 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/125054 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/88151 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b4d0263f-4ad5-494f-9a4d-80f3ac6fb3f1) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/164176 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/27339 "Passed tests") | [❌ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/144814 "Found 1 new API test failure: TestWebKitAPI.WKNavigation.HTTPSFirstRedirectNoHTTPDowngradeRedirect (failure)") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/105651 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/a134ff76-ca68-47c9-94b9-797e06adf308) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/26387 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/24889 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/17840 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/136081 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/22575 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/172603 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/19624 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/24216 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/133333 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/34364 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/28982 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/133342 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36073 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/34349 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/144370 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/96214 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/27893 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/21188 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/33859 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/101284 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/33358 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/33602 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/33507 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/platform/graphics/FontCascade.cpp
+++ b/Source/WebCore/platform/graphics/FontCascade.cpp
@@ -317,13 +317,20 @@ float FontCascade::width(const TextRun& run, SingleThreadWeakHashSet<const Font>
     }
 
     auto* cacheEntry = fonts()->glyphGeometryCache().add(run, { }, TextShapingContext { *this });
+    bool callerNeedsFallbackFonts = fallbackFonts;
+    bool canUseFallbackFontCacheEntry = !callerNeedsFallbackFonts && !run.rtl();
 
     if (cacheEntry && cacheEntry->width) {
-        if (!glyphOverflow)
-            return *cacheEntry->width;
-        if (cacheEntry->glyphOverflow && cacheEntry->glyphOverflow->computeBounds == glyphOverflow->computeBounds) {
-            *glyphOverflow = *cacheEntry->glyphOverflow;
-            return *cacheEntry->width;
+        // The cache key doesn't include inline direction. For primary-font-only text this
+        // is fine (same total advance regardless of direction), but fallback-font text in
+        // vertical writing mode with RTL inline direction can produce different widths.
+        if (!cacheEntry->usedFallbackFonts || canUseFallbackFontCacheEntry) {
+            if (!glyphOverflow)
+                return *cacheEntry->width;
+            if (cacheEntry->glyphOverflow && cacheEntry->glyphOverflow->computeBounds == glyphOverflow->computeBounds) {
+                *glyphOverflow = *cacheEntry->glyphOverflow;
+                return *cacheEntry->width;
+            }
         }
     }
 
@@ -332,10 +339,15 @@ float FontCascade::width(const TextRun& run, SingleThreadWeakHashSet<const Font>
         fallbackFonts = &localFallbackFonts;
 
     float result = width(codePathToUse, run, fallbackFonts, glyphOverflow);
-    if (cacheEntry && fallbackFonts->isEmptyIgnoringNullReferences()) {
-        cacheEntry->width = result;
-        if (glyphOverflow)
-            cacheEntry->glyphOverflow = *glyphOverflow;
+    bool hasFallbackFonts = !fallbackFonts->isEmptyIgnoringNullReferences();
+
+    if (cacheEntry) {
+        if (!hasFallbackFonts || canUseFallbackFontCacheEntry) {
+            cacheEntry->width = result;
+            if (glyphOverflow)
+                cacheEntry->glyphOverflow = *glyphOverflow;
+            cacheEntry->usedFallbackFonts = hasFallbackFonts;
+        }
     }
     return result;
 }
--- a/Source/WebCore/platform/graphics/FontCascadeFonts.h
+++ b/Source/WebCore/platform/graphics/FontCascadeFonts.h
@@ -70,6 +70,7 @@ struct GlyphOverflow {
 struct GlyphGeometryCacheEntry {
     Markable<float> width;
     Markable<GlyphOverflow> glyphOverflow;
+    bool usedFallbackFonts { false };
 };
 
 namespace ShapedTextCacheDefaults {


Source/WebCore/platform/graphics/FontCascade.cpp
```diff
@@ -317,13 +317,21 @@ float FontCascade::width(const TextRun& run, SingleThreadWeakHashSet<const Font>
...
 
     if (cacheEntry && cacheEntry->width) {
-        if (!glyphOverflow)
-            return *cacheEntry->width;
-        if (cacheEntry->glyphOverflow && cacheEntry->glyphOverflow->computeBounds == glyphOverflow->computeBounds) {
-            *glyphOverflow = *cacheEntry->glyphOverflow;
-            return *cacheEntry->width;
+        // The cache key doesn't include direction. For primary-font-only text this is fine
+        // (same total advance in both directions), but fallback-font text in vertical writing
+        // mode can produce different widths per direction.
```
alanbaradlay: > but fallback-font text in vertical writing mode can produce different widths per direction

Could you expand on this (either here in the comment or in the commit message). I don't see any reference to orthogonal content here.

Andrew-Fryer: I don't understand why vertical writing mode makes the direction matter.
I'm guessing the direction matters because the boundary between fonts has different spacing for AB vs. BA; is that right?
Why not include direction in the cache key when using fallback-fonts instead of banning this case?

Andrew-Fryer:
Ok, I think I understand everything except for the vertical different direction with fallback fonts weirdness.
LGTM except for that issue.

Source/WebCore/platform/graphics/FontCascade.cpp
vitorroriz: Sorry, I haven't provided enough context.
Historically, WebKit didn't allow the WidthCache (now GeometryCache) to cache content when it detected use of font fallback (any font rather than the primary font was is used for that text run being measured).
The reason for that seems to be that at some places, the font fallback set is used to compute things like line height (?), so callers pass an output pointer for ::width to fill wit the fallback set of fonts being used. IIUC, in case of a hit we wouldn't fill that output parameter, so the caller wouldn't know what fallback is being used, and this was never cached. So we would have 2 options, either also storing the fallback set in the cache, or just caching when the fallback is not necessary.

 However, IFC passes a null pointer as output parameter, it doesn't need the set to be used because it seems it always compute the fallback set.
 
Ok, now back to your original question. Our GeometryCache key does not account for direction as the sum of the width is always the same I believe. Conceptually I believe that should be the same for vertical-mode. However, in my first iterations of the patch a set of vertical tests were failing layout dump by 1 pixel:
```
  fast/css/vertical-text-overflow-ellipsis-text-align-center-mixed.html
  fast/css/vertical-text-overflow-ellipsis-text-align-justify-mixed.html
  fast/css/vertical-text-overflow-ellipsis-text-align-left-mixed.html
  fast/css/vertical-text-overflow-ellipsis-text-align-right-mixed.html
  ```

These are vertical tests with: 
```
 direction: rtl;
 -webkit-writing-mode: vertical-rl;
 -webkit-text-orientation: mixed;
```
For some reason that I don't understand, the direction seems to affect width in this case. Maybe a font variant is used by CT? I'm not sure, but for simplicity I simply excluded RTL content from being cached in the presence of font fallback. Another alternative would be to add direction to the key of the GeometryCache but I guess this would be better as a follow-up.

TLDR: just to be clear RTL/LTR is cached in the general case, it is just excluded in the presence of fallback set.

Andrew-Fryer: Ok, sounds like it's safe to land this now and then we can visit this logic in the future if we see a need to understand the 1 pixel differences or have the cache hit on this case.

alanbaradlay: when you say "direction" do you mean block or inline direction? This PR seems to handle inline direction (isLTR) but the comment refers to block direction which is somewhat confusing -that’s what my original question was about.

vitorroriz: @alanbaradlay It is kind of both but I need to update the comment: I'm referring to inline direction in respect to what needs to be considered by the cache. However, the only failing tests (1-pixel layout dump) are the ones that have vertical-rl block direction and inline direction RTL. I'll adjust the comment.

alanbaradlay: maybe figure out where that 1px diff comes from?

vitorroriz: The difference comes from shaping. In the presence of `unicode-bidi: bidi-override` inline direction makes into a TextRun and affects how we shape: 'Vi' will be shaped differently from 'iV'. That's a pre-existent general issue not related to this patch. The reason why it fails for this specific tests is that block direction vertical makes us pick a variant of the primary font, which for us essentially is a fallback. So in the future we should consider making inline direction into the GeometryCache key but that would probably require a lot of rebaselining of layout dump tests.

alanbaradlay (approved):

PR #63952, opened 2026-04-30
Preserve sticky user activation across same-origin navigations

kate-k-lee:
#### db9860c223cb0105828527821ac1e317f41cf924
<pre>
Preserve sticky user activation across same-origin navigations
<a href="https://bugs.webkit.org/show_bug.cgi?id=313716">https://bugs.webkit.org/show_bug.cgi?id=313716</a>

Reviewed by NOBODY (OOPS!).

Implement sticky user activation carry-over per whatwg/html#11454,
centralized in LocalDOMWindow::carryOverStickyActivationFromPreviousWindow
and applied at two entry points:

* DocumentWriter::begin: carry sticky activation onto a same-origin
  destination, suppressed when the redirect chain crossed origins
  (guards A -&gt; B -&gt; A confused-deputy patterns).
* FrameLoader::open(CachedFrameBase&amp;): apply the same rule on bfcache
  reactivation; no network load, so the redirect suppressor is N/A.

Migrate sticky and history-action activation from timestamp-derived
flags to explicit booleans on LocalDOMWindow, per the spec&apos;s updated
data model. m_lastActivationTimestamp continues to drive transient
activation only. notifyActivated, its propagation, consume-history-
action, and WebPage::updateUserActivationTimestamps update the
booleans alongside the timestamp.

Gated behind StickyUserActivationAcrossSameOriginNavigationsEnabled
(testable, off by default, auto-enabled in WebKitTestRunner). The
data-model migration is observationally equivalent and not gated.

navigation-state-reset-sameorigin.html now contains FAIL because
sticky activation is preserved across same-origin subframe navigations
per the spec change; revert the expected.txt once the WPT is imported.
</pre>
https://github.com/WebKit/WebKit/commit/5db185b96acf5e208750c492d6691a921bb99fb8

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/173458 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/47390 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/41317 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/183031 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/131326 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/175323 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/48683 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/47072 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/134874 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/95222 "4 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/176416 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/38302 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/155802 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/115350 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/36943 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/35296 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/31516 "Built successfully") | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/4313 "Passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/147367 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/35296 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/185456 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/34720 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/38327 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/39497 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/143741 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/47058 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/155358 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/143607 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/47737 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/31912 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/112849 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/38545 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/119629 "Built successfully") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/210491 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/46243 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/10251 "Passed tests") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/54911 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/45645 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/45876 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/45702 "Built successfully") | | |

### DIFF
--- a/LayoutTests/imported/w3c/web-platform-tests/html/user-activation/navigation-state-reset-sameorigin-expected.txt
+++ b/LayoutTests/imported/w3c/web-platform-tests/html/user-activation/navigation-state-reset-sameorigin-expected.txt
@@ -5,5 +5,5 @@ Tests that navigating a same-origin child frame resets its activation states.
 Click inside the yellow area.
 
 
-FAIL Post-navigation state reset. assert_true: child-two sticky user activation state after loaded expected true got false
+PASS Post-navigation state reset.
 
--- a/LayoutTests/platform/ios/TestExpectations
+++ b/LayoutTests/platform/ios/TestExpectations
@@ -6899,6 +6899,8 @@ imported/w3c/web-platform-tests/html/user-activation/navigation-state-reset-cros
 imported/w3c/web-platform-tests/html/user-activation/navigation-state-reset-sameorigin.html [ Skip ]
 imported/w3c/web-platform-tests/html/user-activation/user-activation-interface.html [ Skip ]
 imported/w3c/web-platform-tests/html/user-activation/activation-same-and-cross-origin.sub.html [ Skip ]
+imported/w3c/web-platform-tests/html/user-activation/navigate-to-crossorigin-redirect.html [ Skip ]
+imported/w3c/web-platform-tests/html/user-activation/navigate-to-sameorigin.html [ Skip ]
 
 # WebGL in OffscreenCanvas requires the GPUP.
 imported/w3c/web-platform-tests/html/canvas/offscreen/manual/the-offscreen-canvas/offscreencanvas.getcontext.worker.html [ Skip ]
--- a/Source/WebCore/loader/DocumentWriter.cpp
+++ b/Source/WebCore/loader/DocumentWriter.cpp
@@ -47,6 +47,7 @@
 #include "LocalFrameView.h"
 #include "MIMETypeRegistry.h"
 #include "Navigation.h"
+#include "NetworkLoadMetrics.h"
 #include "Page.h"
 #include "PluginDocument.h"
 #include "RawDataDocumentParser.h"
@@ -207,6 +208,21 @@ bool DocumentWriter::begin(const URL& urlReference, bool dispatch, Document* own
 
     frameLoader->setOutgoingReferrer(url);
     frame->setDocument(document.copyRef());
+
+    // whatwg/html#11454: carry sticky activation over same-origin navigations,
+    // except when the redirect chain crossed origins (A -> B -> A).
+    if (previousWindow && existingDocument) {
+        bool hadCrossOriginRedirect = false;
+        if (RefPtr documentLoader = frameLoader->documentLoader()) {
+            if (auto* metrics = documentLoader->response().deprecatedNetworkLoadMetricsOrNull())
+                hadCrossOriginRedirect = metrics->hasCrossOriginRedirect;
+        }
+        if (RefPtr newWindow = document->window()) {
+            Ref previousOrigin = existingDocument->securityOrigin();
+            Ref targetOrigin = SecurityOrigin::create(url);
+            newWindow->carryOverStickyActivationFromPreviousWindow(*previousWindow, previousOrigin, targetOrigin, hadCrossOriginRedirect);
+        }
+    }
 
     if (m_decoder)
         document->setDecoder(m_decoder.copyRef());
--- a/LayoutTests/TestExpectations
+++ b/LayoutTests/TestExpectations
@@ -845,8 +845,6 @@ imported/w3c/web-platform-tests/html/user-activation/activation-trigger-keyboard
 imported/w3c/web-platform-tests/html/user-activation/activation-trigger-keyboard-escape.html [ Skip ]
 imported/w3c/web-platform-tests/html/user-activation/activation-trigger-mouse-left.html [ Skip ]
 imported/w3c/web-platform-tests/html/user-activation/activation-trigger-mouse-right.html [ Skip ]
-webkit.org/b/313716 imported/w3c/web-platform-tests/html/user-activation/navigate-to-crossorigin-redirect.html [ Skip ]
-webkit.org/b/313716 imported/w3c/web-platform-tests/html/user-activation/navigate-to-sameorigin.html [ Skip ]
 imported/w3c/web-platform-tests/html/webappapis/scripting/processing-model-2/unhandled-promise-rejections/allow-crossorigin.html [ Skip ]
 imported/w3c/web-platform-tests/html/webappapis/scripting/processing-model-2/unhandled-promise-rejections/disallow-crossorigin.html [ Skip ]
 imported/w3c/web-platform-tests/html/webappapis/scripting/processing-model-2/unhandled-promise-rejections/promise-rejection-events.html [ Skip ]
--- a/LayoutTests/imported/w3c/web-platform-tests/html/user-activation/navigate-to-crossorigin-redirect-expected.txt
+++ b/LayoutTests/imported/w3c/web-platform-tests/html/user-activation/navigate-to-crossorigin-redirect-expected.txt
@@ -1,0 +1,5 @@
+Non-zero-sized body to allow clicks.
+
+
+PASS User activation propagation across a same-origin navigation
+
--- a/LayoutTests/imported/w3c/web-platform-tests/html/user-activation/navigate-to-sameorigin-expected.txt
+++ b/LayoutTests/imported/w3c/web-platform-tests/html/user-activation/navigate-to-sameorigin-expected.txt
@@ -1,0 +1,5 @@
+Non-zero-sized body to allow clicks.
+
+
+PASS User activation propagation across a same-origin navigation
+
--- a/Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
+++ b/Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
@@ -8114,6 +8114,20 @@ StandardFontFamily:
     WebCore:
       default: '""_str'
 
+StickyUserActivationAcrossSameOriginNavigationsEnabled:
+  type: bool
+  status: testable
+  category: dom
+  humanReadableName: "Sticky user activation across same-origin navigations"
+  humanReadableDescription: "Preserve sticky user activation when navigating to a same-origin page (per whatwg/html#11454)"
+  defaultValue:
+    WebKitLegacy:
+      default: false
+    WebKit:
+      default: false
+    WebCore:
+      default: false
+
 StorageAPIEnabled:
   type: bool
   status: stable
--- a/Source/WebCore/loader/FrameLoader.cpp
+++ b/Source/WebCore/loader/FrameLoader.cpp
@@ -2824,6 +2824,13 @@ void FrameLoader::open(CachedFrameBase& cachedFrame)
     Ref document = *cachedFrame.document();
     ASSERT(document->window());
 
+    // Snapshot the previous page before clear() detaches it; used for
+    // sticky-activation carry-over below.
+    RefPtr previousWindow = m_frame->window();
+    RefPtr<SecurityOrigin> previousOrigin;
+    if (RefPtr previousDocument = m_frame->document())
+        previousOrigin = &previousDocument->securityOrigin();
+
     clear(document.ptr(), true, true, cachedFrame.isMainFrame());
 
     document->attachToCachedFrame(cachedFrame);
@@ -2860,6 +2867,16 @@ void FrameLoader::open(CachedFrameBase& cachedFrame)
     frame->setDocument(document.copyRef());
 
     protect(document->window())->resumeFromBackForwardCache();
+
+    // whatwg/html#11454 bfcache reactivation: carry sticky activation from a
+    // same-origin previous page onto the cached window. No network load, so
+    // the redirect-chain suppressor does not apply.
+    if (previousWindow && previousOrigin && previousWindow.get() != document->window()) {
+        if (RefPtr cachedWindow = document->window()) {
+            Ref targetOrigin = document->securityOrigin();
+            cachedWindow->carryOverStickyActivationFromPreviousWindow(*previousWindow, *previousOrigin, targetOrigin, /* hadCrossOriginRedirect */ false);
+        }
+    }
 
     updateFirstPartyForCookies();
 
--- a/Source/WebCore/page/LocalDOMWindow.cpp
+++ b/Source/WebCore/page/LocalDOMWindow.cpp
@@ -1605,6 +1605,20 @@ void LocalDOMWindow::consumeLastActivationIfNecessary()
         m_lastActivationTimestamp = -MonotonicTime::infinity();
 }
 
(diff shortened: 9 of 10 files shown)


Source/WebCore/loader/DocumentWriter.cpp
```diff
@@ -207,6 +207,20 @@ bool DocumentWriter::begin(const URL& urlReference, bool dispatch, Document* own
     frameLoader->setOutgoingReferrer(url);
     frame->setDocument(document.copyRef());
 
+    // Preserves the sticky user activation state after a page navigates to
+    // another same-origin page, unless the navigation involved a redirect,
+    // to prevent confused deputy attacks (A -> B -> A via redirect).
+    // https://bugs.webkit.org/show_bug.cgi?id=313716
+    if (previousWindow && previousWindow->hasStickyActivation() && existingDocument) {
+        bool wasRedirected = false;
+        if (auto* documentLoader = frameLoader->documentLoader())
```
lukewarlow: I think this needs to use RefPtr rather than auto*.

LayoutTests/imported/w3c/web-platform-tests/html/user-activation/navigate-to-sameorigin.html
```diff
@@ -1,5 +1,7 @@
 <!DOCTYPE html>
```
lukewarlow, line 1: Are these new changes to this test? If so they'll need to be exported via something like `export-w3c-test-changes -b 313716`, once that's done you can copy the WPT PR link into the see also of the webkit bugzilla entry (I think this is needed for the bots).

lukewarlow: If they're already upstream it's worth importing the test change separately to make it super clear where the changes come from.

lukewarlow: Same applies to the new test too.

Source/WebCore/loader/DocumentWriter.cpp
```diff
@@ -207,6 +207,20 @@ bool DocumentWriter::begin(const URL& urlReference, bool dispatch, Document* own
...
+            wasRedirected = documentLoader->originalURL() != documentLoader->responseURL();
+        if (!wasRedirected && existingDocument->securityOrigin().isSameOriginAs(SecurityOrigin::create(url))) {
+            if (auto* newWindow = document->window())
```
lukewarlow: The SaferCPP bot didn't seem to mind this one so I think it can stay as auto*.

Source/WebCore/loader/DocumentWriter.cpp
kate-k-lee: Fixed. Thanks.

Source/WebCore/loader/DocumentWriter.cpp
kate-k-lee: I'm not sure why the bot didn't flag it, but I checked how document->window() is used elsewhere in the loader code (FrameLoader.cpp, DocumentLoader.cpp, etc.) and they all use RefPtr. Updated this one to match for consistency.

LayoutTests/imported/w3c/web-platform-tests/html/user-activation/navigate-to-sameorigin.html
kate-k-lee: These tests are already upstream (wpt/wpt#[59293](https://github.com/web-platform-tests/wpt/pull/59293/)). I've imported them in a separate PR (#63976, Bug 313752) and removed them from this PR. This PR now only contains the DocumentWriter.cpp change and the feature flag.

LayoutTests/platform/ios/TestExpectations
```diff
@@ -7032,6 +7028,8 @@ imported/w3c/web-platform-tests/html/user-activation/navigation-state-reset-cros
 imported/w3c/web-platform-tests/html/user-activation/navigation-state-reset-sameorigin.html [ Skip ]
 imported/w3c/web-platform-tests/html/user-activation/user-activation-interface.html [ Skip ]
 imported/w3c/web-platform-tests/html/user-activation/activation-same-and-cross-origin.sub.html [ Skip ]
+imported/w3c/web-platform-tests/html/user-activation/navigate-to-crossorigin-redirect.html [ Skip ]
+imported/w3c/web-platform-tests/html/user-activation/navigate-to-sameorigin.html [ Skip ]
```
annevk, line 6903: Why does it fail on iOS?

LayoutTests/imported/w3c/web-platform-tests/html/user-activation/navigation-state-reset-sameorigin-expected.txt
```diff
@@ -5,5 +5,5 @@ Tests that navigating a same-origin child frame resets its activation states.
 Click inside the yellow area.
 
 
-PASS Post-navigation state reset.
+FAIL Post-navigation state reset. assert_false: expected false got true
```
annevk: This needs some kind of explanation as to why this is not a regression.

Source/WebCore/loader/DocumentWriter.cpp
```diff
@@ -208,6 +208,21 @@ bool DocumentWriter::begin(const URL& urlReference, bool dispatch, Document* own
...
 
+    // Preserves the sticky user activation state after a page navigates to
+    // another same-origin page, unless the navigation involved a redirect,
+    // to prevent confused deputy attacks (A -> B -> A via redirect).
+    // https://bugs.webkit.org/show_bug.cgi?id=313716
+    if (frame->settings().stickyUserActivationAcrossSameOriginNavigationsEnabled()
+        && previousWindow && previousWindow->hasStickyActivation() && existingDocument) {
+        bool wasRedirected = false;
+        if (RefPtr documentLoader = frameLoader->documentLoader())
+            wasRedirected = documentLoader->originalURL() != documentLoader->responseURL();
```
annevk: This is not robust. In theory the URL could be the same, but a redirect could have taken place due to a cookie or some such.

I'm also not sure this is the correct place to make this overall check. Maybe @youennf can give some pointers.

LayoutTests/imported/w3c/web-platform-tests/html/user-activation/navigation-state-reset-sameorigin-expected.txt
kate-k-lee: Thank you for the careful review.
                                                                                                                                                                         
My reading of the spec PR discussion is that sticky activation preservation is intended to apply per-frame, including iframes. In particular, the response in [LINK](https://github.com/whatwg/html/pull/11454#issuecomment-3306143534) to the question of restricting it to top-level only leans against that restriction:                                                                                                                                                                       
_"I'm less sure about this. My instinct was to just do whatever was easiest to spec/implement, which in this case was to allow it to work in iframes."_     
and in [LINK](https://github.com/whatwg/html/pull/11454#issuecomment-3824322153)                      
_"On iframes, it seems simpler to just allow those navigations to carry over sticky activation."_

Based on those, my implementation in DocumentWriter::begin() applies uniformly to both top-level and iframe navigations.                                                                  

About navigation-state-reset-sameorigin.html: this test exercises a same-origin iframe navigation (**not top-level**) and asserts that sticky activation is reset — which I read as reflecting the pre-spec-change behavior. Under my reading of the new spec, sticky activation should be preserved in iframe navigations as well, so the test's existing assertion no longer holds. The expected.txt update to FAIL documents this transitional state until WPT upstream is updated to match the new spec.                                                      

If you read the spec discussion differently, please let us know — happy to revisit the scope. 
Thanks again!

LayoutTests/platform/ios/TestExpectations
kate-k-lee: @annevk 

Both tests rely on test_driver.click() to give the popup window user activation before navigating it. PR #[63430](https://github.com/WebKit/WebKit/pull/63430/) enabled popup-window click dispatch via eventSender, but only for the non-iOS path. the iOS WK2 path goes through iController.[singleTapAtPoint](https://github.com/WebKit/WebKit/blob/2a7f5b392764c317809700eeebf3d69b8a7e44e6/LayoutTests/resources/testdriver-vendor.js#L382), which targets the main test view only and has no popup-view-aware equivalent, so it was not covered by that fix. As a result, on iOS the test's precondition never holds. The popup never reaches the activated state, and the assertions about preserving sticky activation across navigation can't be exercised. The tests fail at the activation setup step, not at the behavior we're shipping.

Also, this matches a pre-existing convention in LayoutTests/platform/ios/TestExpectations, which already groups several user-activation WPTs under the [comment](https://github.com/WebKit/WebKit/blob/7332074f0ccefdd29a2dc32d146846adc60acfa7/LayoutTests/platform/ios/TestExpectations#L7028):                                 
_# UserActivation tests rely on test_driver.click() which is not supported on iOS_                                                                                                          
             
The two new tests (navigate-to-sameorigin.html, navigate-to-crossorigin-redirect.html) are skipped under that same group for the same reason.

Source/WebCore/loader/DocumentWriter.cpp
kate-k-lee: Thanks for catching this. I hadn't accounted for that case. I'll revisit the redirect detection and follow up with an updated patch.

Thanks.

kate-k-lee: @annevk 

Addressed both concerns in the latest push:

1. **Redirect detection** is no longer based on URL equality between originalURL() and responseURL(). The check now uses NetworkLoadMetrics::hasCrossOriginRedirect, which is set to true whenever any step in the redirect chain crosses an origin boundary (regardless of whether the final URL happens to match the start URL). This correctly handles A → B → A patterns and  stops blocking legitimate same-origin internal redirects.

2. **Additionally** added a check that the navigation is renderer-initiated. Browser-initiated navigations (address bar typing, bookmarks, history navigation, embedder API loads) now  suppress carry-over via DocumentLoader::isRequestFromClientOrUserInput(). This aligns with Chromium's behavior.

Regarding your second comment (whether DocumentWriter::begin() is the right place for this check) — with the URL-equality removed, the check now reads only from the navigation's own DocumentLoader (initiator flag, redirect metrics) and the previous Document/Window (origin, sticky activation). All of this is per-Document state that is naturally available at navigation commit time. Happy to revisit the location if you still see a better fit.                                                              

**Not in this PR (planned follow-up):** The spec discussion also calls for the carry-over to apply when a document is reactivated from the back/forward cache, so that behavior remains consistent between bfcache hits and misses. This PR only updates the "create new document" path (DocumentWriter::begin()); WebKit's bfcache restore path is a separate code site that I plan to address in a follow-up once the spec text for the reactivate algorithm is finalized.

kate-k-lee:
Moving this to draft. Plan to (1) add bfcache restore coverage, and (2) drop the Chromium-aligned bits that aren't actually in the spec text (whatwg/html#11454), keeping the implementation strictly spec-faithful. Will re-request review once both are sorted. Thanks!

kate-k-lee:
@annevk 
This PR was temporarily moved back to draft while I addressed the review feedback and made additional changes. It is now ready for review again.

Thanks,

LayoutTests/imported/w3c/web-platform-tests/html/user-activation/navigation-state-reset-sameorigin-expected.txt
kate-k-lee, line 1: This WPT now requires preserving sticky activation in subframes too — will update WebKit accordingly:
https://github.com/web-platform-tests/wpt/pull/60008/changes

kate-k-lee:
Hi @youennf 
could you please take a look at the implementation changes for the proposed HTML spec update in https://github.com/whatwg/html/pull/11454 ?

The spec PR has already been reviewed by @annevk and is expected to be merged soon. I’ve updated the implementation accordingly, and the new behavior is currently guarded as default-off/testable, so it should not affect existing behavior by default.

Your review would be appreciated when you have a chance. Thanks!

PR #63962, opened 2026-04-30
Fix SVG pathLength="0" and negative pathLength handling for stroke dashing

Ahmad-S792:
#### 40238a317bef41bf4fb231e3d830f1aff3914a89
<pre>
Fix SVG pathLength=&quot;0&quot; and negative pathLength handling for stroke dashing
<a href="https://bugs.webkit.org/show_bug.cgi?id=313729">https://bugs.webkit.org/show_bug.cgi?id=313729</a>
<a href="https://rdar.apple.com/175928827">rdar://175928827</a>

Reviewed by Rob Buis.

Per <a href="https://w3c.github.io/svgwg/svg2-draft/paths.html#PathLengthAttribute">https://w3c.github.io/svgwg/svg2-draft/paths.html#PathLengthAttribute</a>:

  &quot;A value of zero is valid and must be treated as a scaling factor of
  infinity. A value of zero scaled infinitely must remain zero, while any
  non-percentage value greater than zero must become +Infinity.&quot;

  &quot;A negative value is an error (see Error handling).&quot;

The pathLength attribute was checked via `if (float pathLength = ...)` which
treats zero as falsy/unspecified, and negative values were used for scaling
instead of being ignored.

Use hasAttribute() to distinguish specified-zero from unspecified, then
handle zero and negative values per spec. Note: hasAttribute() (not
hasAttributeWithoutSynchronization()) is required because pathLength is
an SVG animated property that needs synchronization before attribute lookup.

* LayoutTests/TestExpectations: Unskip Progressions
* Source/WebCore/rendering/svg/SVGRenderSupport.cpp:
(WebCore::SVGRenderSupport::applyStrokeStyleToContext):
* Source/WebCore/rendering/svg/SVGTextLayoutEngine.cpp:
(WebCore::SVGTextLayoutEngine::beginTextPathLayout):

Canonical link: <a href="https://commits.webkit.org/314001@main">https://commits.webkit.org/314001@main</a>
</pre>


nullhook:
when dash array contains zero-length entries the early return`setStrokeStyle(SolidStroke);` seems incorrect. it seems to cause a regression too. can you add a test that exercises `pathLength=0` and `dasharray="0 10"` this should produce `[0, Infinity]` and nothing should render

rwlbuis (approved):
Nice, LGTM.

PR #64019, opened 2026-05-01
[WebDriver BiDi] Validate script.evaluate serialization options and parse stack traces

sergedeh:
[WebDriver BiDi] Validate script.evaluate serialization options and parse stack traces
https://bugs.webkit.org/show_bug.cgi?id=288060

Reviewed by NOBODY (OOPS!).

`script.evaluate` previously ignored `serializationOptions`, always serialized
objects to one level, returned empty stack traces for JavaScript exceptions,
and could confuse page-thrown exceptions with internal evaluation failures.

Validate nullable `maxObjectDepth` and `maxDomDepth` values as non-negative,
finite integers representable by the current IPC depth type. Use the generated
protocol enum parser for `includeShadowTree`. Carry `maxObjectDepth` through IPC
as `std::optional<int>` and pass JavaScript `null` for an unlimited depth,
instead of using a numeric sentinel.

Parse supported V8- and JSC/SpiderMonkey-style stack strings into BiDi call
frames. Use the first parsed frame for the top-level line and column, and share
the exception-details construction used by `script.callFunction` and
`script.evaluate`.

Keep user JavaScript exceptions on an explicit result channel so they become
`script.EvaluateResultException`. Keep serialization, bridge, malformed-result,
and browsing-context failures on the BiDi command-error channel. Sanitize
JavaScript callback errors before crossing the native boundary and preserve a
useful diagnostic when normalizing internal errors.

This patch validates `maxDomDepth` and `includeShadowTree`, but does not carry
them to the WebProcess or implement their DOM/shadow-tree behavior. It also
does not cover the full WebDriver BiDi `js-uint` range because depth is still
stored as `int`.

The existing serializer still changes recursive values to a generic object at
the depth boundary and does not serialize DOM nodes, shared identifiers,
Proxy/generator types, or exact non-Error thrown and rejected values. Those
serialization semantics require separate WebProcess, IPC, and exception-value
work.

* Source/WebKit/UIProcess/Automation/BidiScriptAgent.cpp:
(validateDepthParameter): Validate nullable depth parameters while keeping
invalid input distinct from an unlimited depth.
(WebKit::BidiScriptAgent::callFunction): Reuse common empty exception details.
(WebKit::BidiScriptAgent::evaluate): Validate serialization options and forward
object depth.
(WebKit::BidiScriptAgent::finishEvaluateBidiScriptResult): Separate JavaScript
exception results from command failures and reject malformed results.
(WebKit::BidiScriptAgent::parseStackTrace): Parse supported stack formats.
(WebKit::BidiScriptAgent::buildExceptionDetailsFromExceptionObject): Build Error
details with parsed stack frames.
(WebKit::BidiScriptAgent::createEmptyStackTrace):
(WebKit::BidiScriptAgent::createExceptionDetails):
Share construction for exception details without JavaScript stack data.
* Source/WebKit/UIProcess/Automation/BidiScriptAgent.h:
Declare parsed-stack state and non-null exception-details helpers.
* Source/WebKit/UIProcess/Automation/WebAutomationSession.cpp:
(normalizedBidiScriptEvaluationError): Normalize bridge failures to BiDi command
errors while preserving diagnostic details.
(WebKit::WebAutomationSession::evaluateBidiScript): Carry optional object depth
and preserve callback error details.

* Source/WebKit/UIProcess/Automation/WebAutomationSession.h:
Update the optional object-depth interface.
* Source/WebKit/WebProcess/Automation/WebAutomationSessionProxy.cpp:
(WebKit::WebAutomationSessionProxy::evaluateBidiScript): Pass an absent object
depth to JavaScript as null.
* Source/WebKit/WebProcess/Automation/WebAutomationSessionProxy.h:
Update the optional object-depth interface.
* Source/WebKit/WebProcess/Automation/WebAutomationSessionProxy.js:
(safelyReadStringProperty): Safely copy diagnostics from arbitrary thrown values.
(createBidiScriptError): Build fixed callback error records.
(AutomationSessionProxy.evaluateBidiScript): Separate result serialization from
the callback error channel.
(AutomationSessionProxy._executeBidiScript): Resolve page exceptions as explicit
exception results and reject serializer failures as internal errors.
(AutomationSessionProxy.serializeBidiRemoteValue): Treat null object depth as
unlimited.
* Source/WebKit/WebProcess/Automation/WebAutomationSessionProxy.messages.in:
Carry object depth as an optional IPC value.
Remove stale expectations for serialization-option validation and mark the
Error exception-detail cases that now pass.

https://github.com/WebKit/WebKit/commit/016942dbcabd73a09bc19c511c399a667c035338

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/184963 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/58883 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/52711 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/195297 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/149918 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/60240 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/58609 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/144537 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/100299 "Exiting early after 60 failures. 60 tests run. 61 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/187918 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/48213 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/165133 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/124825 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/46940 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/45034 "Passed tests") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/157309 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/44699 "Passed tests") | [✅ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/6349 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/51546 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/49385 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/197245 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/58550 "Built successfully") | | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/154022 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/59259 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/41302 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/153679 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/48196 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/131550 "Built successfully") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/128179 "The change is no longer eligible for processing.") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/57752 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/19715 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/57111 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/57360 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/57188 "Built successfully") | | |


Source/WebKit/UIProcess/Automation/BidiScriptAgent.cpp
```diff
@@ -61,6 +63,31 @@ using EvaluateResultType = Inspector::Protocol::BidiScript::EvaluateResultType;
 static RefPtr<Inspector::Protocol::BidiScript::RemoteValue> deserializeRemoteValue(const JSON::Value*);
 static Ref<JSON::Value> deserializeLocalValue(const JSON::Value&);
 
+// Per W3C BiDi spec, maxObjectDepth/maxDomDepth default to null (no limit). We use -1 as sentinel for null.
+static constexpr int unlimitedDepth = -1;
+
+static std::optional<int> validatedDepthParameter(const JSON::Object& options, ASCIILiteral key)
```
burg: I don't like using -1 as sentinel. I would prefer this helper have the return type `CommandResult<std::optional<int>>` so that the callsite can check for an error generically rather than potentially passing it along.

burg: Naming: to match the return type `CommandResult<T>` I prefer that this be `validateDepthParameter` to reflect that it is fallible. Otherwise, `validatedDepthParameter` is quietly a `validatedDepthParameterOrError`.

Source/WebKit/UIProcess/Automation/BidiScriptAgent.cpp
```diff
@@ -444,9 +454,29 @@ void BidiScriptAgent::evaluate(const String& expression, bool awaitPromise, Ref<
...
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(!validatedDomDepth, InvalidParameter);
+        // FIXME: Propagate maxDomDepth once the WebProcess serializer supports DOM depth limiting.
+        // https://bugs.webkit.org/show_bug.cgi?id=288060
+
+        auto validatedObjDepth = validatedDepthParameter(serializationOptions, "maxObjectDepth"_s);
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(!validatedObjDepth, InvalidParameter);
+        maxObjectDepth = *validatedObjDepth;
+
+        if (auto includeShadowTreeJSON = serializationOptions.getValue("includeShadowTree"_s)) {
+            String includeShadowTreeValue;
```
burg, line 470: It is better to use the generated protocol enum value parsing function (and rely on its error handling) rather than listing out the enum raw values here.

Source/WebKit/UIProcess/Automation/BidiScriptAgent.cpp
```diff
@@ -457,108 +487,35 @@ void BidiScriptAgent::evaluate(const String& expression, bool awaitPromise, Ref<
...
         // IPC or internal error - NOT a JavaScript exception.
-        // Include diagnostic information from the underlying error.
         String errorText = makeString("Internal error: script evaluation failed"_s);
         if (!result.error().isEmpty())
             errorText = makeString("Internal error: "_s, result.error());
 
-        // Construct error object structure for RemoteValue.value per BiDi spec.
-        auto errorObject = JSON::Object::create();
-        errorObject->setString("message"_s, errorText);
         auto exceptionRemote = BidiScript::RemoteValue::create()
```
burg: Naming: I would call this `exceptionMirror` or `exceptionValue`. In the other cases it was `exceptionAsRemoteValue` which is also clearer.

Source/WebKit/UIProcess/Automation/BidiScriptAgent.cpp
```diff
@@ -1040,6 +997,140 @@ std::optional<String> BidiScriptAgent::browsingContextForRealm(RealmIdentifier r
...
+}
+
+Ref<Inspector::Protocol::BidiScript::StackTrace> BidiScriptAgent::createEmptyStackTrace()
+{
+    return Inspector::Protocol::BidiScript::StackTrace::create()
+        .setCallFrames(JSON::ArrayOf<Inspector::Protocol::BidiScript::StackFrame>::create())
+        .release();
+}
+
+RefPtr<Inspector::Protocol::BidiScript::ExceptionDetails> BidiScriptAgent::createExceptionDetails(const String& text, Ref<Inspector::Protocol::BidiScript::RemoteValue>&& exceptionValue, unsigned lineNumber, unsigned columnNumber)
```
burg: This should return `Ref<>`?

Source/WebKit/UIProcess/Automation/BidiScriptAgent.cpp
```diff
@@ -1040,6 +997,140 @@ std::optional<String> BidiScriptAgent::browsingContextForRealm(RealmIdentifier r
...
+            .setFunctionName(functionName.isEmpty() ? emptyString() : functionName)
+            .setUrl(urlPart.isEmpty() ? emptyString() : urlPart)
+            .release();
+        callFrames->addItem(WTF::move(stackFrame));
+    }
+
+    return result;
+}
+
+RefPtr<Inspector::Protocol::BidiScript::ExceptionDetails> BidiScriptAgent::buildExceptionDetailsFromExceptionValue(RefPtr<JSON::Value> exceptionValue)
```
burg: This should return `Ref<>`?

Source/WebKit/UIProcess/Automation/BidiScriptAgent.cpp
```diff
@@ -457,108 +487,35 @@ void BidiScriptAgent::evaluate(const String& expression, bool awaitPromise, Ref<
 
 void BidiScriptAgent::finishEvaluateBidiScriptResult(const String& realmID, const String& expression, Inspector::CommandResult<String>&& result, Inspector::CommandCallbackOf<Inspector::Protocol::BidiScript::EvaluateResultType, String, RefPtr<Inspector::Protocol::BidiScript::RemoteValue>, RefPtr<Inspector::Protocol::BidiScript::ExceptionDetails>>&& callback)
 {
-    // FIXME: Implement full BiDi exception details (stack traces, line/column numbers, exception type)
-    // https://bugs.webkit.org/show_bug.cgi?id=304548
     using namespace Inspector::Protocol;
 
     if (!result.has_value()) {
```
burg: In the `callFunction` path, there is special casing of `JavaScriptError;` for internal errors. As written, won't this code report user JS exceptions as internal errors?

burg (changes requested):
This is great work, apologies for the delayed review. Can you double check this is doing the right thing for user JS exceptions?

Source/WebKit/UIProcess/Automation/BidiScriptAgent.cpp
sergedeh: Yes, you are right. The current code is treating any failed CommandResult as an internal failure. I fixed it at the JS boundary to make it more BiDi-compliant: user JS exceptions and awaited rejections now become script.EvaluateResultException, while actual internal failures stay as command errors. I will create a follow-up ticket to upgrade callFunction to use the same BiDi-compliant flow too.

sergedeh:
Reviewer orientation: https://gist.github.com/sergedeh/11525e8c76a8d0a85c0b021bade67586

PR #64041, opened 2026-05-01
[GStreamer][MSE] Tear down the pipeline upon EOS notification and re-create it when the MediaSource un-marks its EOS flag

philn:
#### 96f2c38a56fb640ae23a1c60b9f0e1ff58b8c551
<pre>
[GStreamer][MSE] Tear down the pipeline upon EOS notification and re-create it when the MediaSource un-marks its EOS flag
<a href="https://bugs.webkit.org/show_bug.cgi?id=313814">https://bugs.webkit.org/show_bug.cgi?id=313814</a>

Without this patch the following sequence of events would lead to a GStreamer error in the MSE
playback pipeline where we would attempt to push a buffer downstream after EOS was previously
pushed:

- appendBuffer()
- let the segment play until EOS
- appendBuffer()
- play

The proposed solution is to dispose the internal MSE playback pipeline after EOS has been received.
If playback is requested again, the pipeline is re-created from scratch. The previous workarounds
related with &quot;setEosWithNoBuffers&quot; are then no longer needed and can be removed.

Test: media/media-source/media-source-real-play-after-eos.html

* LayoutTests/media/media-source/media-source-real-play-after-eos-expected.txt: Added.
* LayoutTests/media/media-source/media-source-real-play-after-eos.html: Added.
* Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp:
(WebCore::MediaPlayerPrivateGStreamer::tearDown):
(WebCore::MediaPlayerPrivateGStreamer::load):
(WebCore::MediaPlayerPrivateGStreamer::timeChanged):
(WebCore::MediaPlayerPrivateGStreamer::handleMessage):
(WebCore::MediaPlayerPrivateGStreamer::createGSTPlayBin):
(WebCore::MediaPlayerPrivateGStreamer::cdmInstanceAttached):
(WebCore::MediaPlayerPrivateGStreamer::cdmInstanceDetached):
(WebCore::MediaPlayerPrivateGStreamer::handleSyncErrorMessage): Deleted.
* Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.h:
* Source/WebCore/platform/graphics/gstreamer/TrackPrivateBaseGStreamer.cpp:
(WebCore::TrackPrivateBaseGStreamer::~TrackPrivateBaseGStreamer):
* Source/WebCore/platform/graphics/gstreamer/TrackPrivateBaseGStreamer.h:
* Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.cpp:
(WebCore::MediaPlayerPrivateGStreamerMSE::load):
(WebCore::MediaPlayerPrivateGStreamerMSE::play):
(WebCore::MediaPlayerPrivateGStreamerMSE::pause):
(WebCore::MediaPlayerPrivateGStreamerMSE::seekToTarget):
(WebCore::MediaPlayerPrivateGStreamerMSE::doSeek):
(WebCore::MediaPlayerPrivateGStreamerMSE::didPreroll):
(WebCore::MediaPlayerPrivateGStreamerMSE::didEnd):
(WebCore::MediaPlayerPrivateGStreamerMSE::updateStates):
(WebCore::MediaPlayerPrivateGStreamerMSE::rebuildPipeline):
(WebCore::MediaPlayerPrivateGStreamerMSE::setEosWithNoBuffers): Deleted.
* Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.h:
* Source/WebCore/platform/graphics/gstreamer/mse/MediaSourcePrivateGStreamer.cpp:
(WebCore::MediaSourcePrivateGStreamer::markEndOfStream):
(WebCore::MediaSourcePrivateGStreamer::unmarkEndOfStream):
* Source/WebCore/platform/graphics/gstreamer/mse/SourceBufferPrivateGStreamer.cpp:
(WebCore::SourceBufferPrivateGStreamer::flush):
* Source/WebCore/platform/graphics/gstreamer/mse/WebKitMediaSourceGStreamer.cpp:
(webKitMediaSrcLoop):

Canonical link: <a href="https://commits.webkit.org/312998@main">https://commits.webkit.org/312998@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/6f3ed8a39ca5be64496b4d550a86a11aebe39565

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/161783 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/35257 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/28360 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/170686 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/118154 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/163652 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/35358 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/35258 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/125522 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/88438 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/164742 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/27835 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/145318 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/106118 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/26884 "Passed tests") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/25399 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/18407 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/136577 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/23073 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/173175 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/20215 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/24714 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/133818 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/34931 "Built successfully") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/29488 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/133823 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36142 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/34915 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/144868 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/96963 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/28356 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/21691 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/34425 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/101870 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/33925 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/34168 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/34074 "Built successfully") | | |

### DIFF
--- a/LayoutTests/media/media-source/media-source-real-play-after-eos.html
+++ b/LayoutTests/media/media-source/media-source-real-play-after-eos.html
@@ -1,0 +1,92 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <title>media-source-real-play-after-eos (inspired from media-source-real-seek-and-play)</title>
+        <script src="mp4-generator.js"></script>
+        <script src="../video-test.js"></script>
+        <script>
+         // Tests sequential append of contiguous segments with real MP4 data.
+
+         // Verifies that playback of a second segment works after playback of the first segment
+         // ended with an EOS event at the platform level.
+
+         var source;
+         var videoSourceBuffer;
+
+         async function runTest() {
+             findMediaElement();
+
+             if (!MediaSource.isTypeSupported(MP4.VIDEO_TYPE)) {
+                 failTest(`${MP4.VIDEO_TYPE} is not supported`);
+                 return;
+             }
+
+             source = new MediaSource();
+             video.src = URL.createObjectURL(source);
+             await waitFor(source, 'sourceopen');
+
+             videoSourceBuffer = source.addSourceBuffer(MP4.VIDEO_TYPE);
+
+             var videoInitData = MP4.initSegment({
+                 timescale: 1000,
+                 tracks: [{ id: 1, type: 'video', timescale: 1000 }]
+             });
+
+             // First segment: 4 seconds
+             var videoSegment1 = MP4.mediaSegment({
+                 sequenceNumber: 1,
+                 tracks: [{ id: 1, type: 'video', baseDecodeTime: 0, samples: [
+                     { duration: 1000, compositionTimeOffset: 0, isSync: true },
+                     { duration: 1000, compositionTimeOffset: 0, isSync: false },
+                     { duration: 1000, compositionTimeOffset: 0, isSync: false },
+                     { duration: 1000, compositionTimeOffset: 0, isSync: false },
+                 ]}]
+             });
+
+             // Second segment: 4 more seconds (contiguous)
+             var videoSegment2 = MP4.mediaSegment({
+                 sequenceNumber: 2,
+                 tracks: [{ id: 1, type: 'video', baseDecodeTime: 4000, samples: [
+                     { duration: 1000, compositionTimeOffset: 0, isSync: true },
+                     { duration: 1000, compositionTimeOffset: 0, isSync: false },
+                     { duration: 1000, compositionTimeOffset: 0, isSync: false },
+                     { duration: 1000, compositionTimeOffset: 0, isSync: false },
+                 ]}]
+             });
+
+             videoSourceBuffer.appendBuffer(videoInitData);
+             await waitFor(videoSourceBuffer, 'updateend');
+
+             videoSourceBuffer.appendBuffer(videoSegment1);
+             await waitFor(videoSourceBuffer, 'updateend');
+
+             run('source.endOfStream()');
+             testExpected('source.readyState', 'ended');
+
+             // Play the first segment until EOS.
+             video.play();
+             video.currentTime = 3.5;
+             await waitFor(video, 'seeked');
+             await waitFor(video, 'ended');
+
+             videoSourceBuffer.appendBuffer(videoSegment2);
+             await waitFor(videoSourceBuffer, 'updateend');
+
+             run('source.endOfStream()');
+             testExpected('source.readyState', 'ended');
+             testExpectedEqualWithTolerance('source.duration', 8, 0.01);
+
+             // Play the second segment, the seek towards its end should work as well.
+             video.play();
+             video.currentTime = 7.5;
+             await waitFor(video, 'seeked');
+             await waitFor(video, 'ended');
+
+             endTest();
+         }
+        </script>
+    </head>
+    <body onload="runTest()">
+        <video></video>
+    </body>
+</html>
--- a/Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp
+++ b/Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp
@@ -226,14 +226,9 @@ void MediaPlayerPrivateGStreamer::tearDown(bool clearMediaPlayer)
 
     m_sinkTaskQueue.startAborting();
 
-    for (auto& track : m_audioTracks.values())
-        track->disconnect();
-
-    for (auto& track : m_textTracks.values())
-        track->disconnect();
-
-    for (auto& track : m_videoTracks.values())
-        track->disconnect();
+    m_audioTracks.clear();
+    m_videoTracks.clear();
+    m_textTracks.clear();
 
     if (m_fillTimer.isActive())
         m_fillTimer.stop();
@@ -270,6 +265,8 @@ void MediaPlayerPrivateGStreamer::tearDown(bool clearMediaPlayer)
     if (m_pipeline) {
         unregisterPipeline(m_pipeline);
         gst_element_set_state(m_pipeline.get(), GST_STATE_NULL);
+
+        m_source = nullptr;
 
         auto bus = adoptGRef(gst_pipeline_get_bus(GST_PIPELINE(m_pipeline.get())));
         gst_bus_disable_sync_message_emission(bus.get());
@@ -386,13 +383,27 @@ void MediaPlayerPrivateGStreamer::load(const String& urlString)
         m_isDelayingLoad = true;
     }
 
-    // Reset network and ready states. Those will be set properly once
-    // the pipeline pre-rolled.
-    m_networkState = m_isDelayingLoad ? MediaPlayer::NetworkState::Idle : MediaPlayer::NetworkState::Loading;
-    player->networkStateChanged();
-    m_readyState = MediaPlayer::ReadyState::HaveNothing;
-    player->readyStateChanged();
+    // Reset network and ready states. Those will be set properly once the pipeline pre-rolled.
+    // Those states are managed directly by the MSE player sub-class.
+    if (!isMediaSource()) {
+        m_networkState = m_isDelayingLoad ? MediaPlayer::NetworkState::Idle : MediaPlayer::NetworkState::Loading;
+        player->networkStateChanged();
+        m_readyState = MediaPlayer::ReadyState::HaveNothing;
+        player->readyStateChanged();
+    }
+
     m_areVolumeAndMuteInitialized = false;
+    m_isEndReached = false;
+    m_isCachedPositionValid = false;
+    m_cachedDuration = MediaTime::invalidTime();
+    m_isSeeking = false;
+    m_isSeekPending = false;
+    m_seekTarget = { };
+
+#if ENABLE(ENCRYPTED_MEDIA)
+    if (m_cdmContext)
+        gst_element_set_context(GST_ELEMENT(m_pipeline.get()), m_cdmContext.get());
+#endif
 
     if (!m_isDelayingLoad)
         commitLoad();
@@ -1521,7 +1532,8 @@ void MediaPlayerPrivateGStreamer::loadStateChanged()
 
 void MediaPlayerPrivateGStreamer::timeChanged(const MediaTime& seekedTime)
 {
-    updateStates();
+    if (!isMediaSource())
+        updateStates();
     GST_DEBUG_OBJECT(pipeline(), "Emitting timeChanged notification (seekCompleted:%d)", seekedTime.isValid());
     if (RefPtr player = m_player.get()) {
         if (seekedTime.isValid())
@@ -1940,13 +1952,6 @@ bool MediaPlayerPrivateGStreamer::handleNeedContextMessage(GstMessage* message)
     return false;
 }
 
-void MediaPlayerPrivateGStreamer::handleSyncErrorMessage(GstMessage*)
-{
-    GST_DEBUG_OBJECT(pipeline(), "Accounting queued sync error");
-    // This attribute is decremented later from handleMessage() in the main thread.
-    m_queuedSyncErrors.exchangeAdd(1);
-}
-
 // Returns the size of the video.
 FloatSize MediaPlayerPrivateGStreamer::naturalSize() const
 {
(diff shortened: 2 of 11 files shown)


Source/WebCore/platform/graphics/gstreamer/mse/MediaSourcePrivateGStreamer.cpp
```diff
@@ -234,7 +228,7 @@ void MediaSourcePrivateGStreamer::unmarkEndOfStream()
     if (!player)
         return;
 
-    player->setEosWithNoBuffers(false);
+    player->restart();
```
calvaris: With this defaulting to PLAYING, won't we go spuriously to PLAYING in the case the application appends new data after endOfStream() without having called play() again? The new layout test happens to call video.play() right after, so it doesn't catch this.

Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.h
calvaris: This seems to be dead code now, right?

Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp
```diff
@@ -4600,8 +4594,8 @@ void MediaPlayerPrivateGStreamer::cdmInstanceDetached(CDMInstance& instance)
     ASSERT(m_cdmInstance == &instance);
     GST_DEBUG_OBJECT(m_pipeline.get(), "detaching CDM instance %p, setting empty context", m_cdmInstance.get());
     m_cdmInstance = nullptr;
-    GRefPtr<GstContext> context = adoptGRef(gst_context_new("drm-cdm-proxy", FALSE));
-    gst_element_set_context(GST_ELEMENT(m_pipeline.get()), context.get());
+    m_cdmContext = adoptGRef(gst_context_new("drm-cdm-proxy", TRUE));
+    gst_element_set_context(GST_ELEMENT(m_pipeline.get()), m_cdmContext.get());
```
calvaris: I think I understand what you're doing here but this will break a detach + attach sequence as you'll not attach the new context because you already have one. I think the following code solves this:

```suggestion
    m_cdmContext = nullptr;

    auto emptyContext = adoptGRef(gst_context_new("drm-cdm-proxy", TRUE));
    gst_element_set_context(GST_ELEMENT(m_pipeline.get()), emptyContext.get());
```

Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.cpp
calvaris:
```suggestion
if (!pipeline())
    restart(GST_STATE_PAUSED);
m_isPaused = false;
if (!m_playbackRate && m_playbackRatePausedState == PlaybackRatePausedState::ManuallyPaused)
    m_playbackRatePausedState = PlaybackRatePausedState::RatePaused;
updateStates();
```

And let the updateStates machine do its work...

Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.h
```diff
@@ -129,6 +130,7 @@ class MediaPlayerPrivateGStreamerMSE : public MediaPlayerPrivateGStreamer {
 
     void emitStreams(const Vector<RefPtr<MediaSourceTrackGStreamer>>&);
 
+    String m_url;
```
calvaris: You're shadowing the m_url from the parent. I think you can make it protected and reuse it here.

Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.cpp
```diff
@@ -581,32 +589,16 @@ void MediaPlayerPrivateGStreamerMSE::emitStreams(const Vector<RefPtr<MediaSource
...
-            // by the main thread. That's not a problem, since the player itself will be destructed and nobody will care about the
-            // error count integrity anymore.
-        } else
-            m_ignoreErrors = false;
-    }
+    GST_DEBUG("Re-starting, target state: %s", gst_state_get_name(targetState));
+    m_isEndReached = false;
+    MediaPlayerPrivateGStreamer::load(m_url);
+    gst_element_set_state(m_pipeline.get(), targetState);
+    m_isPipelinePlaying = targetState == GST_STATE_PLAYING;
```
calvaris: To align with `changePipelineState` it would be better something like:

```suggestion
auto change = gst_element_set_state(m_pipeline.get(), targetState);
if (change == GST_STATE_CHANGE_FAILURE) {
    GST_WARNING_OBJECT(pipeline(), "restart: state change to %s failed", gst_state_get_name(targetState));
    return;
}
m_isPipelinePlaying = targetState == GST_STATE_PLAYING;
```

Or even more importantly: why are you not using `changePipelineState`?

LayoutTests/media/media-source/media-source-real-play-after-eos.html
```diff
@@ -0,0 +1,92 @@
...
+             if (!MediaSource.isTypeSupported(MP4.VIDEO_TYPE)) {
+                 failTest(`${MP4.VIDEO_TYPE} is not supported`);
+                 return;
+             }
+
+             source = new MediaSource();
+             video.src = URL.createObjectURL(source);
+             await waitFor(source, 'sourceopen');
+
+             sourceBuffer = source.addSourceBuffer(MP4.VIDEO_TYPE);
```
calvaris: I think we should do this multitrack and at least make an audio source.

Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.cpp
```diff
@@ -581,32 +589,16 @@ void MediaPlayerPrivateGStreamerMSE::emitStreams(const Vector<RefPtr<MediaSource
...
-    if (eosWithNoBuffers) {
-        // On GStreamer, EOS with no buffers causes a parsebin error here:
-        // https://github.com/GStreamer/gst-plugins-base/blob/1.18.6/gst/playback/gstparsebin.c#L3495
-
-        m_ignoreErrors = true;
-
-        GST_DEBUG_OBJECT(pipeline(), "EOS with no buffers, setting pipeline to READY state.");
-        changePipelineState(GST_STATE_READY);
+    if (pipeline())
+        return;
```
calvaris, line 592:
```suggestion
        return;
        
    // restart() relies on sourceSetup() seeing hasAllTracks() == true, otherwise
    // emitStreams() is skipped and the new WebKitMediaSrc never publishes pads,
    // permanently stalling preroll. We don't expect to ever rebuild without all
    // tracks present, but degrade gracefully in release builds rather than hang.
    if (!m_mediaSourcePrivate || !m_mediaSourcePrivate->hasAllTracks()) {
        GST_ERROR("restart() invoked without all tracks ready; aborting rebuild.");
        ASSERT_NOT_REACHED();
        return;
    }
        
```

Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.cpp
calvaris:
```suggestion
    if (!player || !m_source)
```

Source/WebCore/platform/graphics/gstreamer/mse/MediaSourcePrivateGStreamer.cpp
philn: `unmarkEndOfStream()` is called in the context of `MediaSource::openIfInEndedState()`. I don't really understand the question, sorry.

Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.cpp
philn: `changePipelineState` does more things that we don't need to do here, so i went for the KISS approach.

LayoutTests/media/media-source/media-source-real-play-after-eos.html
philn: I think this would complicate the test without any other benefit.

The current test times out without this patch and passes with the patch applied and that should be sufficient to prevent future regressions.

Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.cpp
philn: I don't think this is needed, this method is called from `MediaSource::openIfInEndedState()` which exits early already if the source is not in ended state.

Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.cpp
philn: I think a release assert on `m_source` makes more sense, actually.

Source/WebCore/platform/graphics/gstreamer/mse/MediaSourcePrivateGStreamer.cpp
calvaris: It's easy though I might not have explained it properly. Imagine:

1. you're playing
2. you get to the end of the stream
3. as a result you are paused
4. you append more data
5. as a result `unmarkEndOfStream` is called

If the default parameter is `restart(GST_STATE_PLAYING)` you would begin an unrequested playback. Am I wrong?

Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.cpp
calvaris: Let's try `changePipelineState`. I haven't seen any `gst_element_set_state` in this file and I guess there's a reason for it.

LayoutTests/media/media-source/media-source-real-play-after-eos.html
calvaris: I suspect we might be running into track races. That's why I suggested it.

Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.cpp
calvaris: We can keep them both, you can change the order and short-circuit will make the optimization.

Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.cpp
calvaris: It's called from more places and lots of track issues can happen while we are at EOS. It does not hurt and you can mark is as unlikely if you want.

Source/WebCore/platform/graphics/gstreamer/mse/MediaSourcePrivateGStreamer.cpp
philn: let's just always restart in paused then.

LayoutTests/media/media-source/media-source-real-play-after-eos.html
philn: What races? The test is written in a synchronous sequence of `await` calls.

calvaris: The awaits make the JS side deterministic, but the parts of this patch that I'm worried about run on the per‑stream streaming threads inside WebKitMediaSrc, and those are concurrent in multi‑track. Two concrete cases:

The new EOS filter at webKitMediaSrcLoop:625 drops EOS when a stream never pushed a buffer. In single‑track that's harmless (only one sink, no aggregation). In audio+video, if the application appends to only one track before endOfStream() (a normal moment in adaptive playback), the empty track's EOS is silently dropped, the corresponding sink never reports EOS, GstBin never aggregates pipeline EOS, didEnd never fires and 'ended' never dispatche, await waitFor(video, 'ended') hangs.

The asserts on STREAM_COLLECTION / SEGMENT push were softened in this patch with the comment "can fail during PAUSED→READY". Single‑track has one pad, retry just delays the single ordered sequence. Multi‑track has two pads racing on the same collection; if A's push fails and V's succeeds, V's first buffer can reach playbin before A's STREAM_COLLECTION, scrambling the documented event ordering.

Both are silent in single‑track and observable in multi‑track. The cost of audio + video in the test is one extra addSourceBuffer and a parallel appendBuffer chain.

philn: > The new EOS filter at webKitMediaSrcLoop:625 drops EOS when a stream never pushed a buffer. In single‑track that's harmless (only one sink, no aggregation). In audio+video, if the application appends to only one track before endOfStream() (a normal moment in adaptive playback), the empty track's EOS is silently dropped, the corresponding sink never reports EOS, GstBin never aggregates pipeline EOS, didEnd never fires and 'ended' never dispatche, await waitFor(video, 'ended') hangs.
> 

Sorry but that's not a valid use-case. If you add an audio sourcebuffer and a video sourcebuffer to a mediasource and then append data to only one sourcebuffer and request playback on the associated video element, the pipeline will just hang forever. You can try it with current main, even without this PR.

> The asserts on STREAM_COLLECTION / SEGMENT push were softened in this patch with the comment "can fail during PAUSED→READY". Single‑track has one pad, retry just delays the single ordered sequence. Multi‑track has two pads racing on the same collection; if A's push fails and V's succeeds, V's first buffer can reach playbin before A's STREAM_COLLECTION, scrambling the documented event ordering.
> 

The version I'm about to push has those changes reverted, they were not needed in the end.

calvaris (approved):

aperezdc:
Backported into `webkitglib/2.52` as commit 17ef6524f4883c8bb24eec0667b97637dfeedf89

PR #64046, opened 2026-05-01
[OMG] Tail call shuffler should be able to handle registers

kmiller68:
#### 62596b85906b6d10e97da86663eadc5ea8f41e46
<pre>
[OMG] Tail call shuffler should be able to handle registers
<a href="https://bugs.webkit.org/show_bug.cgi?id=313820">https://bugs.webkit.org/show_bug.cgi?id=313820</a>
<a href="https://rdar.apple.com/176022434">rdar://176022434</a>

Reviewed by Yusuke Suzuki.

prepareForTailCallImpl shuffles arguments, callee-save restores, and the
return PC into position for Wasm tail calls. The old algorithm sorted
stack moves by source offset and used a &quot;danger zone&quot; heuristic, but it
could only handle stack-to-stack moves -- register-destination arguments
had to be skipped entirely, relying on B3 patchpoint constraints to place
them. This made the patchpoint early-clobber all callee-save registers,
preventing B3 from using callee-saves for any patchpoint input.

This meant that we were *extremely* limited on registers for X86_64. Now
that B3 is free to assign values to callee saves, it should reduce the
constraints on the tail call patchpoint. This will be more important with
the lazy restore frame in <a href="https://bugs.webkit.org/show_bug.cgi?id=313732">https://bugs.webkit.org/show_bug.cgi?id=313732</a>

Replace it with the same parallel move algorithm used by BBQ&apos;s
emitShuffle. The algorithm resolves dependencies between moves
recursively and breaks cycles by spilling to scratch slots. This handles
all source/destination combinations (GPR, FPR, Stack) uniformly. There
were a few changes needed for OMG though:

1 Callee-save restoration is now part of the parallel move rather than a
  separate emitRestore() call before the shuffle. The algorithm correctly
  orders reads from callee-save registers relative to their restores, so
  B3 is free to place inputs in callee-save registers.

2 The patchpoint early-clobber is reduced from all callee-saves to just
  macroClobberedGPRs (the scratch register). The late clobber is removed
  entirely since the patchpoint is terminal. This gives B3 more freedom
  in register allocation.

3 On ARM/ARM64/RISCV64, the return PC is shuffled directly into the link
  register as part of the parallel move, eliminating the separate
  load+move after the shuffle. ARM64E untagging still happens before the
  SP adjustment.

4 The boxed callee (for indirect and direct tail calls) is now placed
  into its frame slot as part of the parallel move. Since B3 can now
  place patchpoint inputs in callee-save registers, the boxed callee
  must be read before callee-save restores clobber it. The
  boxedCalleeCallee Value* is passed through createTailCallPatchpoint
  constrained to LateColdAny so prepareForTailCallImpl can include it
  in the shuffle targeting CallFrameSlot::callee in the new frame.

5 V128 stack entries are split into two Width64 entries to satisfy the
  algorithm&apos;s atomic-location invariant (a write to location X must only
  conflict with reads from exactly X). FPR sources are pre-spilled to
  scratch before splitting.

6 Width32 stack sources at non-8-byte-aligned offsets are pre-spilled
  when they conflict with a destination. B3/Air can allocate 4-byte spill
  slots at 4-byte alignment, so a Width32 source may sit in the upper
  half of an 8-byte slot that a Width64 destination write covers.

7 The patchArgs parameter, wasmCallerInfoAsCallee parameter, and
  tmpNeedsSaving machinery are removed since the parallel move handles
  everything directly. The duplicate calleeCode patchpoint arg
  (previously passed in the initial patchArgs at wasmScratchGPR0) is
  also removed.

I went with splitting V128 stack arguments because I expect them to be
rare and we can&apos;t reasonably handle V128s as the atomic size. For 4-byte
values copied them since I expect 8-byte values to be common and
splitting them to significantly slow down the shuffle, so I went with
the spill approach.

I added a load/store/transferWidth set of instructions to MacroAssembler.
In the future we can adopt this in other dynamic width situations like
BBQ.

No new tests, covered by existing tests.

* Source/JavaScriptCore/assembler/MacroAssembler.h:
(JSC::MacroAssembler::loadWidth):
(JSC::MacroAssembler::storeWidth):
(JSC::MacroAssembler::transferWidth):
* Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp:
(JSC::Wasm::OMGIRGenerator::emitIndirectCall):
(JSC::Wasm::prepareForTailCallImpl):
(JSC::Wasm::OMGIRGenerator::createTailCallPatchpoint):
(JSC::Wasm::OMGIRGenerator::emitDirectCall):

Canonical link: <a href="https://commits.webkit.org/312691@main">https://commits.webkit.org/312691@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/bd59ff55ae917ed4d5bc0d8a57c022865a1c8c79

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/160715 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/34188 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/27289 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/169618 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/115781 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/a5b5daca-c104-4258-b64d-be3ecfb4f0ef) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/34289 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/34188 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/124640 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/88005 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/a230e402-58eb-4354-af12-367956ceaaca) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/163674 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/26884 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/144363 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/105237 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b2ce3ece-e5b0-4f9f-8159-a8f2425622a8) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/25936 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/24443 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/17338 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/152771 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/135630 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/22126 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/172102 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/21553 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/19038 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/23766 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/132906 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/33862 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/28535 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/132919 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/33846 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/143921 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/95655 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/24465 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/27507 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/20736 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/193114 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/33357 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/100648 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/49735 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/32856 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/33100 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/33004 "Built successfully") | | | |

### DIFF
--- a/Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
+++ b/Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
@@ -90,6 +90,7 @@
 #include <limits>
 #include <wtf/CheckedArithmetic.h>
 #include <wtf/FastMalloc.h>
+#include <wtf/RecursableLambda.h>
 #include <wtf/StdLibExtras.h>
 #include <wtf/TZoneMallocInlines.h>
 #include <wtf/text/MakeString.h>
@@ -120,6 +121,7 @@ static constexpr bool verboseTailCalls = false;
 #if ASSERT_ENABLED
 static constexpr bool traceExecutionIncludesConstructionSite = false;
 #endif
+
 }
 }
 
@@ -858,7 +860,7 @@ public:
 
     Vector<ConstrainedValue> createCallConstrainedArgs(BasicBlock*, const CallInformation& wasmCalleeInfo, const ArgumentList&);
     auto createCallPatchpoint(BasicBlock*, const RTT&, const CallInformation&, const ArgumentList& tmpArgs) -> CallPatchpointData;
-    auto createTailCallPatchpoint(BasicBlock*, const RTT&, const CallInformation& wasmCallerInfoAsCallee, const CallInformation& wasmCalleeInfoAsCallee, const ArgumentList& tmpArgSourceLocations, Vector<B3::ConstrainedValue> patchArgs) -> CallPatchpointData;
+    auto createTailCallPatchpoint(BasicBlock*, const RTT&, const CallInformation& wasmCallerInfoAsCallee, const CallInformation& wasmCalleeInfoAsCallee, const ArgumentList& tmpArgSourceLocations, Value* boxedCalleeCallee = nullptr) -> CallPatchpointData;
 
     InliningNode* canInline(FunctionSpaceIndex functionIndexSpace, unsigned callProfileIndex) const;
     [[nodiscard]] PartialResult emitInlineDirectCall(InliningNode*, FunctionCodeIndex calleeIndex, const RTT&, const ArgumentList& args, ValueResults&);
@@ -2000,15 +2002,12 @@ auto OMGIRGenerator::emitIndirectCall(Value* calleeInstance, Value* calleeCode, 
         const RTT& callerType = m_info.rtt(callerTypeSignatureIndex);
         CallInformation wasmCallerInfoAsCallee = callingConvention.callInformationFor(callerType, CallRole::Callee);
 
-        auto [patchpoint, _, prepareForCall] = createTailCallPatchpoint(m_currentBlock, signature, wasmCallerInfoAsCallee, wasmCalleeInfoAsCallee, args, { { calleeCode, ValueRep(GPRInfo::wasmScratchGPR0) } });
+        auto [patchpoint, _, prepareForCall] = createTailCallPatchpoint(m_currentBlock, signature, wasmCallerInfoAsCallee, wasmCalleeInfoAsCallee, args, boxedCalleeCallee);
         unsigned patchArgsIndex = patchpoint->reps().size();
         patchpoint->append(calleeCode, ValueRep(GPRInfo::nonPreservedNonArgumentGPR0));
-        patchpoint->append(boxedCalleeCallee, ValueRep::SomeRegister);
         patchArgsIndex += m_proc.resultCount(patchpoint->type());
         patchpoint->setGenerator([prepareForCall = prepareForCall, patchArgsIndex](CCallHelpers& jit, const B3::StackmapGenerationParams& params) {
             prepareForCall->run(jit, params);
-            jit.storeWasmCalleeToCalleeCallFrame(params[patchArgsIndex + 1].gpr(), sizeof(CallerFrameAndPC) - prologueStackPointerDelta());
-            // Allow scratch after the callee is stored, which could be in the scratch register.
             AllowMacroScratchRegisterUsage allowScratch(jit);
             jit.farJump(params[patchArgsIndex].gpr(), WasmEntryPtrTag);
         });
@@ -5306,10 +5305,9 @@ auto OMGIRGenerator::createCallPatchpoint(BasicBlock* block, const RTT& signatur
 }
 
 // See createTailCallPatchpoint for the setup before this.
-static inline void prepareForTailCallImpl(unsigned functionIndex, CCallHelpers& jit, const B3::StackmapGenerationParams& params, const RTT& signature, const CallInformation& wasmCallerInfoAsCallee, const CallInformation& wasmCalleeInfoAsCallee, unsigned firstPatchArg, unsigned lastPatchArg, int32_t newFPOffsetFromFP)
+static inline void prepareForTailCallImpl(unsigned functionIndex, CCallHelpers& jit, const B3::StackmapGenerationParams& params, const RTT& signature, const CallInformation& wasmCalleeInfoAsCallee, unsigned firstPatchArg, unsigned lastPatchArg, int32_t newFPOffsetFromFP, std::optional<unsigned> boxedCalleeArgIndex)
 {
     const RTT& functionSignature = signature;
-    const Checked<int32_t> offsetOfFirstSlotFromFP = WTF::roundUpToMultipleOf<stackAlignmentBytes()>(wasmCallerInfoAsCallee.headerAndArgumentStackSizeInBytes);
     JIT_COMMENT(jit, "Set up tail call, new FP offset from FP: ", newFPOffsetFromFP);
 
     const unsigned frameSize = params.code().frameSize();
@@ -5323,22 +5321,24 @@ static inline void prepareForTailCallImpl(unsigned functionIndex, CCallHelpers& 
     auto newReturnPCOffset = fpOffsetToSPOffset(checkedSum<intptr_t>(CallFrame::returnPCOffset(), newFPOffsetFromFP).value());
 
     // We requested some extra stack space below via requestCallArgAreaSize
-    // ... FP [initial safe area][caller stack space ] [callArgSpace                    ] SP ...
+    // ... FP [caller stack space ] [callArgSpace ] SP ...
     // becomes
-    // ... FP [safe area growing ->    ] [danger           ] [ scratch                  ] SP ...
+    // ... FP [danger             ] [ scratch     ] SP ...
     // This scratch space sits at the very bottom of the stack, near sp.
     // AirLowerStackArgs takes care of adding callArgSpace to our total caller frame size.
-    // BUT, even though we have this extra space, the new frame might be bigger, so we can't
-    // use the new frame as scratch. The new return pc represents the lowest offset from SP we can use.
+    // BUT, even though we have this extra space, we need to be careful not to write into the
+    // live frame while shuffling. If for some reason we found that the scratch space wasn't big
+    // enough for our needs we could always ask B3 for more. We only use it for 32-bit / v128
+    // stack arguments to our caller that can't be parallel moved into place directly.
     int spillPointer = 0;
     const int scratchAreaUpperBound = std::min(
         safeCast<int>(WTF::roundUpToMultipleOf<stackAlignmentBytes()>(static_cast<int>(wasmCalleeInfoAsCallee.headerAndArgumentStackSizeInBytes))),
         newReturnPCOffset);
     auto allocateSpill = [&] (Width width) -> int {
-        int offset = spillPointer;
-        spillPointer += bytesForWidth(width);
-        ASSERT(spillPointer <= scratchAreaUpperBound);
-        ASSERT(offset < scratchAreaUpperBound);
+        int offset = WTF::roundUpToMultipleOf(bytesForWidth(width), spillPointer);
+        spillPointer = offset + bytesForWidth(width);
+        RELEASE_ASSERT(spillPointer <= scratchAreaUpperBound);
+        RELEASE_ASSERT(offset < scratchAreaUpperBound);
         return offset;
     };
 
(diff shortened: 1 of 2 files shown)


Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -5459,167 +5408,344 @@ static inline void prepareForTailCallImpl(unsigned functionIndex, CCallHelpers&
...
+        // so B3 places them directly and src == dst (no move needed). Only stack
+        // destinations reach here and need splitting for the parallel move.
+        if (dstType.width() >= Width128 && dst.kind == ShuffleLocation::Stack) {
+            if (src.kind == ShuffleLocation::FPR) {
+                // It's annoying to figure out if the full width is available
+                // for a store so just write it to a scratch slot and handle it
+                // from there.
+                // FIXME: We could probably do something better here, especially
+                // if the FPR isn't a callee-save.
+                int scratchOffset = allocateSpill(Width128);
```
Constellation, line 5529: is spill space enough for extreme condition? IIRC, there are upper bound `scratchAreaUpperBound` and if we are allocating more spills than that, we start overriding the other content.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -5459,167 +5408,344 @@ static inline void prepareForTailCallImpl(unsigned functionIndex, CCallHelpers&
...
+            }
+        }
+    };
 
-    std::ranges::sort(argsToMove, [](const auto& left, const auto& right) {
-        return std::get<0>(left) > std::get<0>(right);
-    });
+    auto spillToScratch = [&](unsigned index) {
+        auto& e = entries[index];
+        int scratchOffset = allocateSpill(Width64);
```
Constellation, line 5700: Ditto.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
kmiller68: Yeah, I turned it into a release assert to check that. I also got rid of a bunch of unnecessary spill allocations for FP constants so hopefully that makes it a non-issue.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -5451,167 +5390,351 @@ static inline void prepareForTailCallImpl(unsigned functionIndex, CCallHelpers&
...
+        } else if (src.kind == ShuffleLocation::FPR) {
+            if (dst.kind == ShuffleLocation::FPR) {
+                if (e.width == Width128) {
+                    ASSERT(e.isVector);
+                    jit.moveVector(src.fpr, dst.fpr);
+                } else
+                    jit.moveDouble(src.fpr, dst.fpr);
+            } else {
+                ASSERT(dst.kind == ShuffleLocation::Stack);
+                jit.storeDouble(src.fpr, CCallHelpers::Address(MacroAssembler::stackPointerRegister, dst.stackOffset));
```
dhecht: In the GPR -> stack case, we write 32 or 64 bits depending on width. Is there a reason we don't do that for FPR -> stack too? I don't think this is a bug but wondering if there's a reason for the inconsistency.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -5451,167 +5390,351 @@ static inline void prepareForTailCallImpl(unsigned functionIndex, CCallHelpers&
...
+            dst = ShuffleLocation::fromGPR(dstParam.location.jsr().payloadGPR());
+        else if (dstParam.location.isFPR())
+            dst = ShuffleLocation::fromFPR(dstParam.location.fpr());
+        else {
+            ASSERT(dstParam.location.isStack());
+            dst = ShuffleLocation::fromStack(fpOffsetToSPOffset(checkedSum<int32_t>(dstParam.location.offsetFromFP(), newFPOffsetFromFP).value()));
+        }
+
+        // V128 args with register destinations have exact patchpoint constraints,
+        // so B3 places them directly and src == dst (no move needed). Only stack
```
dhecht, line 5520: worth having an assert in the else clause proving this?

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -5451,167 +5390,351 @@ static inline void prepareForTailCallImpl(unsigned functionIndex, CCallHelpers&
...
+        if (dstParam.location.isGPR())
+            dst = ShuffleLocation::fromGPR(dstParam.location.jsr().payloadGPR());
+        else if (dstParam.location.isFPR())
+            dst = ShuffleLocation::fromFPR(dstParam.location.fpr());
+        else {
+            ASSERT(dstParam.location.isStack());
+            dst = ShuffleLocation::fromStack(fpOffsetToSPOffset(checkedSum<int32_t>(dstParam.location.offsetFromFP(), newFPOffsetFromFP).value()));
+        }
+
+        // V128 args with register destinations have exact patchpoint constraints,
```
dhecht, line 5519: is this particular to V128? i.e. the "args with register destination"? If not I think it'd be clearer if only the second sentence talks about V128 specifically since that's the special handling, IIUC.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -5451,167 +5390,351 @@ static inline void prepareForTailCallImpl(unsigned functionIndex, CCallHelpers&
...
+                // from there.
+                // FIXME: We could probably do something better here, especially
+                // if the FPR isn't a callee-save.
+                int scratchOffset = allocateSpill(Width128);
+                auto scratchAddr = CCallHelpers::Address(MacroAssembler::stackPointerRegister, scratchOffset);
+                jit.storeVector(src.fpr, scratchAddr);
+                src = ShuffleLocation::fromStack(scratchOffset);
+            }
+
+            ASSERT(src.kind == ShuffleLocation::Stack);
```
dhecht: is this because v128 won't be constants? maybe worth clarifying.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -5451,167 +5390,351 @@ static inline void prepareForTailCallImpl(unsigned functionIndex, CCallHelpers&
...
+                if (e.width <= Width32)
+                    jit.store32(src.gpr, dstAddr);
+                else
+                    jit.storePtr(src.gpr, dstAddr);
+            }
+        } else if (src.kind == ShuffleLocation::FPR) {
+            if (dst.kind == ShuffleLocation::FPR) {
+                if (e.width == Width128) {
+                    ASSERT(e.isVector);
+                    jit.moveVector(src.fpr, dst.fpr);
```
dhecht: I thought this case isn't possible (i.e. v128 with src.fpr != dst.fpr)? i.e. should this just be ASSERT(e.width != Width128) and delete this case? And then not sure if isVector is still needed?

dhecht: Actually, isn't that true for all reg, reg cases (due to patchpoint constraint you mention above), so it seems all these move cases are dead code. Or am I misunderstanding something?

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
dhecht: is this danger thing still relevant?

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -5451,167 +5390,351 @@ static inline void prepareForTailCallImpl(unsigned functionIndex, CCallHelpers&
...
+        using enum Kind;
+
+        static ShuffleLocation none() { return { .stackOffset = 0, .kind = None }; }
+        static ShuffleLocation fromGPR(GPRReg r) { return { .gpr = r, .kind = GPR }; }
+        static ShuffleLocation fromFPR(FPRReg r) { return { .fpr = r, .kind = FPR }; }
+        static ShuffleLocation fromStack(int32_t offset) { return { .stackOffset = offset, .kind = Stack }; }
+
+        bool operator==(const ShuffleLocation& other) const
+        {
+            if (kind != other.kind || kind == None)
```
dhecht: why the kind == None return false case? I think a quick comment explaining that would help.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
```diff
@@ -5451,167 +5390,351 @@ static inline void prepareForTailCallImpl(unsigned functionIndex, CCallHelpers&
...
+        switch (e.src.kind) {
+        case ShuffleLocation::GPR:
+            jit.storePtr(e.src.gpr, scratchAddr);
+            break;
+        case ShuffleLocation::FPR:
+            jit.storeDouble(e.src.fpr, scratchAddr);
+            break;
+        case ShuffleLocation::Stack: {
+            auto srcAddr = CCallHelpers::Address(MacroAssembler::stackPointerRegister, e.src.stackOffset);
+            if (e.width <= Width32)
```
dhecht: nit: is there a reason we sometimes use == Width32 (like line 5665) and sometimes <= ? If the inconsistency isn't meaningful it'd be better to be consistent to remove distractions.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
kmiller68: Not really. might as well in case we ever move this code to somewhere shared and it actually matters though. I'll change.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
kmiller68: Today, not really. It might make sense to have some other dispatch data come e.g. the code target come in as a `ColdAny` then shuffled into place. But I was planning that for a follow up. I guess it could apply to FPRs generally though.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
kmiller68: Yeah, because we generate spill slots for V128s / 32-bit stack values that are not 8-byte aligned, so in theory if there's an issue we could spill into the live frame. Although, I think the frame layout in the comment is inaccurate now that I think about it. Will update.

Source/JavaScriptCore/wasm/WasmOMGIRGenerator.cpp
kmiller68: Sure made them all ==

Constellation (approved):

PR #64063, opened 2026-05-01
[JSC][WASM][Debugger] Fix STW deadlocks when VM blocks in memory.atomic.wait or WebCore operations

hyjorc1:
#### f426a395257bfc31f28152f4d194feb6628c5437
<pre>
[JSC][WASM][Debugger] Fix STW deadlocks when VM blocks in memory.atomic.wait or WebCore operations
<a href="https://rdar.apple.com/176042013">rdar://176042013</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=313838">https://bugs.webkit.org/show_bug.cgi?id=313838</a>

Reviewed by NOBODY (OOPS!).

Worker threads that block while the VM is entered (vm.isEntered()==true)
prevent the WASM debugger&apos;s Stop-The-World protocol from completing: the VM
never reaches a JSC trap check point to call notifyVMStop(), hanging the
STW count indefinitely.

This patch fixes two categories of blocking operations:

1. memory.atomic.wait32/64 (JSC/WASM)

WaiterListManager::waitForSync() now polls every kDebuggerSTWCheckInterval
(50ms) and calls notifyVMStop(WasmAtomicsWaitBlocked) if NeedStopTheWorld
is set. WasmAtomicsWaitBlocked skips clearStop() so stop data persists
across multiple STW cycles; waitForSync() calls clearStop() on exit.
Stop data (callee/cfr/PC/MC/stack) is threaded from the asm through the
slow path stack frame. stepAtBytecode() handles the atomics-wait case
with resumeAll so notifier threads can run.

2. WebCore blocking operations (six sites)

Adds waitWithSTWParticipation() (WorkerSTWParticipation.h): a drop-in
replacement for BinarySemaphore::wait() that polls every 50ms and calls
notifyVMStop(VMStopped) if NeedStopTheWorld is set. Applied to:
- SubtleCrypto.wrapKey / unwrapKey (WorkerGlobalScope)
- Notification.permission check (WorkerNotificationClient)
- IndexedDB blob disk write, worker path (SerializedScriptValue)
- WebSocket peer initialization (WorkerThreadableWebSocketChannel)
- FileSystemSyncAccessHandle close/resize (WorkerFileSystemStorageConnection)
WorkerDedicatedRunLoop::runInMode() caps timeoutDelay to 50ms and checks
NeedStopTheWorld after each wait, covering sync XHR and any operation
that drives the worker run loop synchronously.

Also fixes DebugServer::start() idempotency: worker VMs in the jsc shell
[...]

PR #64177, opened 2026-05-04
[JSC][WASM][Debugger] Fix STW deadlocks when VM blocks in memory.atomic.wait or WebCore operations

hyjorc1:
#### da1f31b2f4318b5a53ecf09aadafb85fbc8413d4
<pre>
[JSC][WASM][Debugger] Fix STW deadlocks when VM blocks in memory.atomic.wait or WebCore operations
<a href="https://rdar.apple.com/176042013">rdar://176042013</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=313838">https://bugs.webkit.org/show_bug.cgi?id=313838</a>

Reviewed by Mark Lam.

Worker threads that block while the VM is entered (vm.isEntered()==true)
prevent the WASM debugger&apos;s Stop-The-World protocol from completing: the VM
never reaches a JSC trap check point to call notifyVMStop(), hanging the
STW count indefinitely.

This patch fixes two categories of blocking operations:

1. memory.atomic.wait32/64 (JSC/WASM)

WaiterListManager::waitForSync() now polls every kDebuggerSTWCheckInterval
(50ms) and calls notifyVMStop(WasmAtomicsWaitBlocked) if NeedStopTheWorld
is set. WasmAtomicsWaitBlocked skips clearStop() so stop data persists
across multiple STW cycles; waitForSync() calls clearStop() on exit.
Stop data (callee/cfr/PC/MC/stack) is threaded from the asm through the
slow path stack frame. stepAtBytecode() handles the atomics-wait case
with resumeAll so notifier threads can run.

2. WebCore blocking operations (six sites)

Adds waitWithSTWParticipation() (WorkerSTWParticipation.h): a drop-in
replacement for BinarySemaphore::wait() that polls every 50ms and calls
notifyVMStop(VMStopped) if NeedStopTheWorld is set. Applied to:
- SubtleCrypto.wrapKey / unwrapKey (WorkerGlobalScope)
- Notification.permission check (WorkerNotificationClient)
- IndexedDB blob disk write, worker path (SerializedScriptValue)
- WebSocket peer initialization (WorkerThreadableWebSocketChannel)
- FileSystemSyncAccessHandle close/resize (WorkerFileSystemStorageConnection)
WorkerDedicatedRunLoop::runInMode() caps timeoutDelay to 50ms and checks
NeedStopTheWorld after each wait, covering sync XHR and any operation
that drives the worker run loop synchronously.

Also fixes DebugServer::start() idempotency: worker VMs in the jsc shell
re-enter start() via jsc.cpp&apos;s per-VM init path, corrupting m_serverSocket;
the isInService() guard prevents this.

Two new debugger tests: MemoryAtomicWaitTestCase (1ns timeout, breakpoint
+ step) and MemoryAtomicWaitNoTimeoutTestCase (infinite wait, process
interrupt verifies the STW fix).

Canonical link: <a href="https://commits.webkit.org/312813@main">https://commits.webkit.org/312813@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/9f97a6aeefd24f47236e0f8e12748225906df61a

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/161013 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/34486 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/27587 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/169916 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/117273 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/162882 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/34587 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/34486 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/124895 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/117273 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/163972 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/27160 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/144634 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/105492 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/26209 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/24709 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/17636 "Built successfully") | | 
| [✅ 🛠 🧪 jsc](https://ews-build.webkit.org/#/builders/20/builds/153069 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/135903 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/22396 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/172399 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/21851 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/19420 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/24037 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/133172 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/34160 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/28944 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/133182 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/34144 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/144191 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/95983 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/24504 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/27747 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/21009 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/193412 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/33655 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/101080 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/49809 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/33154 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/33398 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/33303 "Built successfully") | | | |

### DIFF
--- a/Source/JavaScriptCore/runtime/StopTheWorldCallback.h
+++ b/Source/JavaScriptCore/runtime/StopTheWorldCallback.h
@@ -28,10 +28,19 @@
 #include <cstdint>
 #include <utility>
 #include <wtf/IterationStatus.h>
+#include <wtf/Seconds.h>
 
 namespace JSC {
 
 class VM;
+
+#if ENABLE(WEBASSEMBLY_DEBUGGER)
+// How often a VM blocked in a non-JSC operation (memory.atomic.wait, sync XHR, etc.)
+// wakes to check NeedStopTheWorld and participate in STW.
+// 50 ms is chosen to be as long as possible while remaining below the ~100 ms threshold
+// for human perception of responsiveness in a debugger driven by human interaction.
+inline constexpr Seconds DebuggerSTWCheckInterval { 0.05 };
+#endif
 
 using StopTheWorldStatus = std::pair<IterationStatus, VM*>;
 
@@ -56,6 +65,10 @@ enum class StopTheWorldEvent : uint8_t {
     //
     // The WASM program itself triggered a stop (breakpoint hit, trap, or dynamic module load).
     WasmProgramStop,
+    // A thread blocked in memory.atomic.wait32/64 cooperatively participating in STW.
+    // Unlike WasmProgramStop, clearStop() is skipped on resume so the stop data persists
+    // across multiple STW cycles while the wait is still in progress.
+    WasmAtomicsWaitBlocked,
     // Not a real stop: fired when the mutator reaches a call/throw site during step-into.
     // Signals the debugger thread to check whether a callee entry breakpoint was installed;
     // the mutator does not pause and wait for a debugger command.
--- a/Source/JavaScriptCore/runtime/WaiterListManager.cpp
+++ b/Source/JavaScriptCore/runtime/WaiterListManager.cpp
@@ -32,10 +32,17 @@
 #include "JSLock.h"
 #include "JSObjectInlines.h"
 #include "ObjectConstructor.h"
+#include "Options.h"
+#include "VMManager.h"
+#include "VMTraps.h"
 #include <wtf/DataLog.h>
 #include <wtf/NeverDestroyed.h>
 #include <wtf/RawPointer.h>
 #include <wtf/TZoneMallocInlines.h>
+
+#if ENABLE(WEBASSEMBLY_DEBUGGER)
+#include "WasmDebugServerUtilities.h"
+#endif
 
 WTF_ALLOW_UNSAFE_BUFFER_USAGE_BEGIN
 
@@ -72,6 +79,39 @@ WaiterListManager& WaiterListManager::singleton()
     return manager;
 }
 
+static void waitForSync(Locker<Lock>& listLocker, VM& vm, Ref<Waiter>& syncWaiter, Ref<WaiterList>& list, MonotonicTime time)
+{
+    listLocker.assertIsHolding(list->lock);
+    while (syncWaiter->isOnList() && time.now() < time && !vm.hasTerminationRequest()) {
+#if ENABLE(WEBASSEMBLY_DEBUGGER)
+        // FIXME: rdar://176407534 This is a workaround. Ideally VMManager would maintain a registry of all
+        // blocking operations (atomics.wait, futex, etc.) and signal them directly on STW
+        // instead of relying on each site to poll NeedStopTheWorld at a fixed interval.
+        // Implementing that correctly is non-trivial due to registration races, lock ordering
+        // between VMManager and each waiter's list lock, and unregistration races on wait
+        // completion — worth a dedicated follow-up patch.
+        if (Options::enableWasmDebugger()) [[unlikely]] {
+            if (vm.traps().hasTrapBit(VMTraps::NeedStopTheWorld)) {
+                // Unlock to participate in STW. The waiter stays on the list —
+                // isOnList() and time guards handle all outcomes on the next iteration.
+                list->lock.unlock();
+                VMManager::singleton().notifyVMStop(vm, StopTheWorldEvent::WasmAtomicsWaitBlocked);
+                list->lock.lock();
+            }
+            auto cap = MonotonicTime::now() + DebuggerSTWCheckInterval;
+            syncWaiter->condition().waitUntil(list->lock, std::min(time, cap).approximate<WallTime>());
+            continue;
+        }
+#endif
+        syncWaiter->condition().waitUntil(list->lock, time.approximate<WallTime>());
+    }
+
+#if ENABLE(WEBASSEMBLY_DEBUGGER)
+    if (Options::enableWasmDebugger()) [[unlikely]]
+        vm.debugState()->clearStop();
+#endif
+}
+
 template <typename ValueType>
 WaiterListManager::WaitSyncResult WaiterListManager::waitSyncImpl(VM& vm, ValueType* ptr, ValueType expectedValue, Seconds timeout)
 {
@@ -89,8 +129,7 @@ WaiterListManager::WaitSyncResult WaiterListManager::waitSyncImpl(VM& vm, ValueT
         list->addLast(listLocker, syncWaiter);
         dataLogLnIf(WaiterListsManagerInternal::verbose, "<WaiterListManager> <Thread:", Thread::currentSingleton(), "> added a new SyncWaiter=", syncWaiter.get(), " to a waiterList for ptr ", RawPointer(ptr));
 
-        while (syncWaiter->isOnList() && time.now() < time && !vm.hasTerminationRequest())
-            syncWaiter->condition().waitUntil(list->lock, time.approximate<WallTime>());
+        waitForSync(listLocker, vm, syncWaiter, list, time);
 
         // At this point, syncWaiter should be either notified (dequeued) or timeout (not dequeued).
         bool didGetDequeued = !syncWaiter->isOnList();
@@ -351,7 +390,7 @@ Ref<WaiterList> WaiterListManager::findOrCreateList(void* ptr)
 {
     Locker waiterListsLocker { m_waiterListsLock };
     return m_waiterLists.ensure(ptr, [] {
-        return adoptRef(*new WaiterList()); 
+        return adoptRef(*new WaiterList());
     }).iterator->value.get();
 }
 
--- a/Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.cpp
+++ b/Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.cpp
@@ -1232,10 +1232,20 @@ WASM_IPINT_EXTERN_CPP_DECL(get_global_64, unsigned index)
 WASM_IPINT_EXTERN_CPP_DECL(memory_atomic_wait32, IPIntStackEntry* args)
 {
 #if CPU(ARM64) || CPU(X86_64)
-    uint8_t memoryIndex = args[0].i32;
-    uint64_t timeout = args[1].i64;
-    uint32_t value = args[2].i32;
-    uint64_t pointerWithOffset = args[3].i64;
+    uint8_t memoryIndex = args[4].i32;
+    uint64_t timeout = args[5].i64;
+    uint32_t value = args[6].i32;
+    uint64_t pointerWithOffset = args[7].i64;
+#if ENABLE(WEBASSEMBLY_DEBUGGER)
+    if (Options::enableWasmDebugger()) [[unlikely]] {
+        auto* callee = std::bit_cast<Wasm::IPIntCallee*>(args[0].i64);
+        auto* callFrame = std::bit_cast<CallFrame*>(args[1].i64);
+        auto* pc = std::bit_cast<uint8_t*>(args[2].i64);
+        auto* mc = std::bit_cast<uint8_t*>(args[3].i64);
+        auto* stack = args + 4; // wasm expression stack: [memoryIndex, timeout, value, pointer+offset]
+        instance->vm().debugState()->setAtomicsWaitStopData(callee, instance, callFrame, pc, mc, stack);
+    }
+#endif
     int32_t result = Wasm::memoryAtomicWait32(instance, pointerWithOffset, value, timeout, memoryIndex);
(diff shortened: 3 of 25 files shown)


hyjorc1:
See the previous discussion at https://github.com/WebKit/WebKit/pull/64063#discussion_r3177633801.

Source/JavaScriptCore/runtime/WaiterListManager.cpp
```diff
@@ -89,8 +129,7 @@ WaiterListManager::WaitSyncResult WaiterListManager::waitSyncImpl(VM& vm, ValueT
         list->addLast(listLocker, syncWaiter);
         dataLogLnIf(WaiterListsManagerInternal::verbose, "<WaiterListManager> <Thread:", Thread::currentSingleton(), "> added a new SyncWaiter=", syncWaiter.get(), " to a waiterList for ptr ", RawPointer(ptr));
 
-        while (syncWaiter->isOnList() && time.now() < time && !vm.hasTerminationRequest())
-            syncWaiter->condition().waitUntil(list->lock, time.approximate<WallTime>());
+        waitForSync(listLocker, vm, syncWaiter.get(), list.get(), time);
```
MenloDorian: Instead of passing `syncWaiter.get()` and `list.get()`, I suggest passing references to the `Ref<Waiter>` and `Ref<WaiterList>`, and let `waitForSync` call `get` itself.  This communicates that these data structures came from a `Ref`s that are keeping them alive.

Source/JavaScriptCore/runtime/StopTheWorldCallback.h
```diff
@@ -28,11 +28,18 @@
...
+#include <wtf/Seconds.h>
 
 namespace JSC {
 
 class VM;
 
+#if ENABLE(WEBASSEMBLY_DEBUGGER)
+// How often a VM blocked in a non-JSC operation (memory.atomic.wait, sync XHR, etc.)
+// wakes to check NeedStopTheWorld and participate in STW.
+inline constexpr Seconds kDebuggerSTWCheckInterval { 0.05 };
```
MenloDorian: Might want to add a comment here that we're choosing 50 ms because we want this interval to be as long as possible.  For a debugger which deals with human interaction, 100 ms is the typical threshold human perception of responsiveness.  So, 50 ms is a good tradeoff between responsiveness while not interrupting the wait too soon.

Source/JavaScriptCore/wasm/debugger/WasmDebugServer.cpp
```diff
@@ -83,6 +83,15 @@ DebugServer::DebugServer()
...
 bool DebugServer::start()
 {
+    // In production (browser/WebContent) and the jsc shell, start() is called exactly once
+    // from the main context — workers go through a different initialization path and never
+    // reach here. The one exception is the jsc shell with $.agent.start(): those workers
+    // share this singleton but also call start() via jsc.cpp's per-VM path. The main thread
+    // always runs first, so isInService() is already true and no concurrent start() races
+    // are possible.
+    if (isInService())
+        return true;
```
MenloDorian, line 93: This is not very thread safe.  But neither is the `createAndBindServerSocket()` below.  We can make this more robust by adding a static `initLock`, and using a `Locker` to acquire it at the top, that will completely mitigate this is thread safety issue.

Source/WebCore/workers/WorkerSTWParticipation.h
```diff
@@ -23,34 +23,37 @@
...
+// at any such site so the VM cooperatively participates in STW while blocked.
 
-#ifdef __OBJC__
-#include "WebGPU.h"
-#include "WebGPUExt.h"
-#include <Metal/Metal.h>
-#endif
+namespace WebCore {
+
+inline void waitWithSTWParticipation(WTF::BinarySemaphore& semaphore, JSC::VM& vm)
```
MenloDorian: Does this really have to be an inline function?  It seems like it doesn't need to be since this is a `wait` operation and `wait`s are notoriously slow, and not used in hot paths.

Source/JavaScriptCore/wasm/debugger/WasmDebugServerUtilities.h
```diff
@@ -217,11 +223,20 @@ struct DebugState {
     }
     bool isStoppedAtBytecode() const
     {
-        bool result = stopData && stopData->pc;
+        bool result = stopData && !!stopData->pc;
         if (result)
-            RELEASE_ASSERT(stopReason == Reason::Breakpoint || stopReason == Reason::Step || stopReason == Reason::WasmTrap);
+            RELEASE_ASSERT(stopReason == Reason::Breakpoint || stopReason == Reason::Step || stopReason == Reason::WasmTrap || stopReason == Reason::Interrupted);
```
MenloDorian, line 231: The only reason we add `|| stopReason == Reason::Interrupted` is to handle the case for `setAtomicsWaitStopData`.   Other places that use `Reason::Interrupted` are NOT at bytecode.  Maybe it's worth just introducing a `Reason::InterruptedAtomicWait` reason to keep them distinct.

Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.cpp
```diff
@@ -1227,10 +1227,20 @@ WASM_IPINT_EXTERN_CPP_DECL(get_global_64, unsigned index)
...
+    uint64_t pointerWithOffset = args[7].i64;
+#if ENABLE(WEBASSEMBLY_DEBUGGER)
+    if (Options::enableWasmDebugger()) [[unlikely]] {
+        auto* callee = std::bit_cast<Wasm::IPIntCallee*>(args[0].i64);
+        auto* callFrame = std::bit_cast<CallFrame*>(args[1].i64);
+        auto* pc = std::bit_cast<uint8_t*>(args[2].i64);
+        auto* mc = std::bit_cast<uint8_t*>(args[3].i64);
+        auto* stack = args + 4; // wasm expression stack: [memoryIndex, timeout, value, pointer+offset]
+        instance->vm().debugState()->setAtomicsWaitStopData(callee, instance, callFrame, pc, mc, stack);
+    }
```
MenloDorian, line 1247: I wish there's another way we can achieve this without making the IPInt pass this all the time.  But for now, this is probably OK.

Source/JavaScriptCore/wasm/WasmIPIntSlowPaths.cpp
```diff
@@ -1243,10 +1253,20 @@ WASM_IPINT_EXTERN_CPP_DECL(memory_atomic_wait32, IPIntStackEntry* args)
...
```
MenloDorian, line 1273: Ditto.

Source/WebCore/workers/WorkerRunLoop.cpp
```diff
@@ -301,12 +304,30 @@ WorkerDedicatedRunLoop::RunInModeResult WorkerDedicatedRunLoop::runInMode(Worker
     if (predicate.isDefaultMode() && m_sharedTimer->isActive())
         timeoutDelay = std::min(timeoutDelay, m_sharedTimer->fireTimeDelay());
 
+#if ENABLE(WEBASSEMBLY_DEBUGGER)
+    // Apply the STW check cap last so it overrides all platform-specific timeouts.
+    // Zero overhead when the WASM debugger is disabled.
+    if (script && JSC::Options::enableWasmDebugger()) [[unlikely]]
+        timeoutDelay = std::min(timeoutDelay, JSC::kDebuggerSTWCheckInterval);
+#endif
+
```
MenloDorian: You can put this inside the `if (script)` block below.  No need to check the `script` condition twice.

Source/WebCore/workers/WorkerRunLoop.cpp
```diff
@@ -301,12 +304,30 @@ WorkerDedicatedRunLoop::RunInModeResult WorkerDedicatedRunLoop::runInMode(Worker
...
+#if ENABLE(WEBASSEMBLY_DEBUGGER)
+    // Heap access is released above; participate in STW if needed while it remains released.
+    // Check regardless of result — NeedStopTheWorld may be set even when a message arrived.
+    if (script && JSC::Options::enableWasmDebugger()) [[unlikely]] {
+        JSC::VM& vm = script->vm();
+        if (vm.traps().hasTrapBit(JSC::VMTraps::NeedStopTheWorld))
+            JSC::VMManager::singleton().notifyVMStop(vm, JSC::StopTheWorldEvent::VMStopped);
+    }
+#endif
```
MenloDorian, line 316: Ditto.  You can put this inside the `if (script)` block below.

MenloDorian (approved):
r=me.  This PR does improve the code, but is definitely ideal.  We should implement a fix soon to remove the need for polling, and use a signaling solution to service the VMManager STW.

Source/JavaScriptCore/wasm/debugger/WasmDebugServerUtilities.h
hyjorc1: Good catch. I've added a FIXME to track this. I'd prefer not to introduce Reason::AtomicsWaitBlocked in this patch. As it would require auditing all stopReason switch statements and helper predicates across the execution handler, which is enough scope creep to risk regressions in an already-large patch. The FIXME captures the intent for a clean follow-up.

PR #64180, opened 2026-05-04
Animated GIF freezes when presentation attributes are changed on SVG image referenced by use element

nullhook:
#### 0eadba6f98dc0665039ff9641b0e764492c261a2
<pre>
Animated GIF freezes when presentation attributes are changed on SVG image referenced by use element
<a href="https://bugs.webkit.org/show_bug.cgi?id=314093">https://bugs.webkit.org/show_bug.cgi?id=314093</a>
<a href="https://rdar.apple.com/96837306">rdar://96837306</a>

Reviewed by Said Abou-Hallawa.

When an SVG &lt;image&gt; element displaying an animated GIF is referenced by a
&lt;use&gt; element, rapidly changing a presentation attribute like opacity causes
the animation to freeze.

Each setAttribute call triggers invalidateInstances(), which rebuilds the
&lt;use&gt; shadow tree. The clone&apos;s renderer destruction calls
RenderImageResource::willBeDestroyed(), which unconditionally calls
stopAnimation() on the shared BitmapImage. Since the animation timer is
killed faster than the frame duration, it never fires to advance frames.

* LayoutTests/svg/custom/resources/animated-gif-for-use-test.gif: Added.
* LayoutTests/svg/custom/use-animated-image-opacity-change-expected.txt: Added.
* LayoutTests/svg/custom/use-animated-image-opacity-change.html: Added.
* Source/WebCore/rendering/RenderImageResource.cpp:
(WebCore::RenderImageResource::willBeDestroyed):

Canonical link: <a href="https://commits.webkit.org/313531@main">https://commits.webkit.org/313531@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/6085746bc0e2cfddd6276c1cdd866c7169185d92

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/163316 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/36799 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/30014 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/172255 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/119763 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/65d01f0f-3704-49d8-a0b0-523c520238e1) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/165185 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/37126 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/36799 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/126823 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/89407 "3 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/929c8a9d-0f75-43bf-a19e-3cf787f49641) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/166275 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/29094 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/146813 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/107390 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/bed23cb5-568e-46c2-bc99-71e6d44077f7) 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/28140 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/26771 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/19957 "Built successfully") | | 
| | [❌ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/138035 "Found 1 new API test failure: TestWebKitAPI.WKAttachmentTestsIOS.InsertPastedContactAsAttachment (failure)") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/24537 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/174759 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/27344 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/26192 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/135128 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/36467 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/30870 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/135120 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36432 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/37254 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/146332 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/99202 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/30422 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/23177 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/35963 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/108775 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/35462 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/1202 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/35710 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/35610 "Built successfully") | | | |


LayoutTests/svg/custom/use-animated-image-opacity-change.html
```diff
@@ -0,0 +1,49 @@
...
+description("Animated GIF in SVG image referenced by use element should keep animating when presentation attributes are rapidly changed.");
+jsTestIsAsync = true;
+
+var svgImage = document.getElementById("gif");
+
+window.onload = function() {
+    if (!window.internals)
+        return;
+
+    internals.settings.setWebkitImageReadyEventEnabled(true);
```
shallawa: I do not think you need this. This is only need when you want to receive the event `webkitImageFrameReady` for an image.

LayoutTests/svg/custom/use-animated-image-opacity-change.html
```diff
@@ -0,0 +1,49 @@
...
+        window.initialFrameIndex = internals.imageFrameIndex(svgImage);
+
+        var opacity = 1;
+        var frameCount = 0;
+
+        function changeOpacity() {
+            opacity = opacity >= 1 ? 0 : opacity + 0.005;
+            svgImage.setAttribute("opacity", opacity + "");
+            frameCount++;
+            if (frameCount < 120)
```
shallawa: Why do you need all these frames? This will make the test take at least 2 seconds.

LayoutTests/svg/custom/use-animated-image-opacity-change.html
```diff
@@ -0,0 +1,49 @@
...
+                requestAnimationFrame(changeOpacity);
+            else {
+                window.finalFrameIndex = internals.imageFrameIndex(svgImage);
```
shallawa: I think this should work also:
```
var finalFrameIndex = ...
```

LayoutTests/svg/custom/use-animated-image-opacity-change.html
```diff
@@ -0,0 +1,49 @@
+<!DOCTYPE html>
+<html>
+<head>
+<script src="../../resources/js-test-pre.js"></script>
```
shallawa: Please use 
```
<script src="../../resources/js-test.js"></script>
```

LayoutTests/svg/custom/use-animated-image-opacity-change.html
```diff
@@ -0,0 +1,49 @@
...
+                shouldBeTrue("finalFrameIndex > initialFrameIndex");
+                finishJSTest();
+            }
+        }
+
+        requestAnimationFrame(changeOpacity);
+    });
+};
+</script>
+<script src="../../resources/js-test-post.js"></script>
```
shallawa: You do not need this if you reference `js-test.js`.

LayoutTests/svg/custom/use-animated-image-opacity-change.html
```diff
@@ -0,0 +1,49 @@
...
+
+        function changeOpacity() {
+            opacity = opacity >= 1 ? 0 : opacity + 0.005;
+            svgImage.setAttribute("opacity", opacity + "");
+            frameCount++;
+            if (frameCount < 120)
+                requestAnimationFrame(changeOpacity);
+            else {
+                window.finalFrameIndex = internals.imageFrameIndex(svgImage);
+                shouldBeTrue("finalFrameIndex > initialFrameIndex");
```
shallawa: What are we verifying here?
```
    if (frameCount < 3) {
        opacity = opacity + 0.005;
        svgImage.setAttribute("opacity", opacity + "");

        // Verify we are advancing the animation every frame.
        var frameIndex = internals.imageFrameIndex(svgImage);
        shouldBeTrue("frameIndex > initialFrameIndex");

        frameCount++;
        requestAnimationFrame(changeOpacity);
    }
```

Source/WebCore/rendering/RenderImageResource.cpp
```diff
@@ -60,7 +62,9 @@ void RenderImageResource::initialize(RenderElement& renderer)
 
 void RenderImageResource::willBeDestroyed()
 {
-    image()->stopAnimation();
+    RefPtr cached = cachedImage();
```
shallawa:
```
RefPtr cachedImage = this->cachedImage();
```

Source/WebCore/testing/Internals.cpp
```diff
@@ -1182,6 +1185,16 @@ static Image* imageFromImageElement(HTMLImageElement& element)
     return cachedImage ? cachedImage->image() : nullptr;
 }
 
+static Image* imageFromElement(Element& element)
+{
+    CachedImage* cachedImage = nullptr;
+    if (auto* htmlImage = dynamicDowncast<HTMLImageElement>(element))
```
shallawa: htmlImage -> imageElement.

Source/WebCore/testing/Internals.cpp
```diff
@@ -1182,6 +1185,16 @@ static Image* imageFromImageElement(HTMLImageElement& element)
...
+        cachedImage = htmlImage->cachedImage();
+    else if (auto* svgImage = dynamicDowncast<SVGImageElement>(element))
```
shallawa: svgImage -> imageElement.

Source/WebCore/testing/Internals.cpp
```diff
@@ -1182,6 +1185,16 @@ static Image* imageFromImageElement(HTMLImageElement& element)
     return cachedImage ? cachedImage->image() : nullptr;
 }
 
+static Image* imageFromElement(Element& element)
+{
+    CachedImage* cachedImage = nullptr;
+    if (auto* htmlImage = dynamicDowncast<HTMLImageElement>(element))
+        cachedImage = htmlImage->cachedImage();
```
shallawa: This is awkward. We still have the other function `imageFromImageElement(HTMLImageElement& element)`. Since the code is repeated for `HTMLImageElement` and `SVGImageElement`, cannot we use templates?

nullhook:
One thing I've observed is that a single `` element’s ImageLoader and its Renderer both act as `CachedImage` clients. As a result, when reading `numberOfClients`, there’s no way to distinguish between them. 

This matters because if you remove a single `` that's animating, our guard `cached->numberOfClients() <= 1` is false (ImageLoader + Renderer = 2) and we'd incorrectly skip `stopAnimation()`

edit: I've solved this by removing the renderer as client first and then checking `numberOfClients()` - It still seems like an approximation but the right fix would be to return number of renderers only

LayoutTests/svg/custom/use-animated-image-opacity-change.html
nullhook: `finalFrameIndex` needs to be global scoped for `shouldBeTrue` to see.

LayoutTests/svg/custom/use-animated-image-opacity-change.html
nullhook: I was testing with another GIF earlier and left it there. I think we just need enough frames for the GIF's frame timer to fire. I've lowered this number.

Source/WebCore/testing/Internals.cpp
nullhook: I don't think templates help here. However, we can replace `imageFromImageElement` and have everything go through the new `imageFromElement`

LayoutTests/svg/custom/use-animated-image-opacity-change.html
nullhook: We can't assert per-frame because the GIF frame duration is longer (100ms) than the rAF interval here.

So, checking `frameIndex > initialFrameIndex` per frame would fail immediately on the first frame. I can perhaps use a GIF with much shorter duration (within 16ms) but that would be unrealistically fast to reproduce the bug.

LayoutTests/svg/custom/use-animated-image-opacity-change.html
```diff
@@ -0,0 +1,39 @@
...
+    shouldBecomeEqual("internals.isImageAnimating(svgImage)", "true", function() {
+        window.initialFrameIndex = internals.imageFrameIndex(svgImage);
+
+        var opacity = 1;
+        var intervalID = setInterval(function() {
+            opacity = opacity + 0.005;
+            svgImage.setAttribute("opacity", opacity + "");
+        }, 20);
+
+        shouldBecomeEqual("internals.imageFrameIndex(svgImage) > initialFrameIndex", "true", function() {
```
shallawa: This is definitely a flaky test. There is no guarantee that a new frame is going to be drawn before hitting this `shouldBecomeEqual()`. And if this does not happen the test will timeout.

shallawa: I think this should work

```
<!DOCTYPE html>
<html>
<head>
<script src="../../resources/js-test.js"></script>
</head>
<body>
    <svg width="100" height="200">
        <image id="gif" opacity="1" width="100" height="100"></image>
        <use href="#gif" transform="translate(0, 100)"></use>
    </svg>
    <script>
        function drawFrame(image, frame) {
            return new Promise((resolve) => {
                // Force layout and display so the image frame starts decoding
                document.body.offsetHeight;
                testRunner.display();

                image.addEventListener("webkitImageFrameReady", function listener() {
                    image.removeEventListener("webkitImageFrameReady", listener, true);
                    resolve(frame + 1);
                }, true);
            });
        }

        function drawImage(image, frameCount) {
            let promise = drawFrame(image, 0);
            for (let i = 1; i < frameCount; ++i) {
                promise = promise.then((frame) => {
                    return drawFrame(image, frame);
                });
            }
            return promise;
        }

        function loadAndDrawImage(image, href, frameCount) {
            return new Promise((resolve) => {
                image.onload = (() => {
                    drawImage(image, frameCount).then(resolve);
                });
                image.href.baseVal = href;
            });
        }

        window.jsTestIsAsync = true;

        if (window.internals) {
            internals.clearMemoryCache();
            internals.settings.setWebkitImageReadyEventEnabled(true);
            internals.settings.setAnimatedImageAsyncDecodingEnabled(true);
        }

        var image = document.getElementById("gif");

        var opacity = 1;
        var intervalID = setInterval(function() {
            opacity = opacity + 0.005;
            image.setAttribute("opacity", opacity + "");
        }, 20);

        loadAndDrawImage(image, "resources/animated-gif-for-use-test.gif", 10).then(() => {
            debug("All frames of the animated image were drawn in order.");
            clearInterval(intervalID);
            finishJSTest();
        });
    </script>
</body>
</html>
```

Source/WebCore/testing/Internals.cpp
```diff
@@ -216,6 +216,8 @@
 #include "ResourceLoadObserver.h"
 #include "SMILTimeContainer.h"
 #include "SVGDocumentExtensions.h"
+#include "SVGElementTypeHelpers.h"
+#include "SVGImageElement.h"
```
shallawa: Do we still need the changes in this file? The test above does not use `imageFrameIndex()` or `isImageAnimating()`.

Source/WebCore/rendering/RenderImageResource.cpp
```diff
@@ -60,9 +60,11 @@ void RenderImageResource::initialize(RenderElement& renderer)
 
 void RenderImageResource::willBeDestroyed()
 {
-    image()->stopAnimation();
+    RefPtr cachedImage = this->cachedImage();
     if (m_styleImage && m_renderer)
         m_styleImage->removeClient(*m_renderer);
+    if (cachedImage && cachedImage->numberOfClients() <= 1)
+        image()->stopAnimation();
```
shallawa: I think this is incorrect. Usually the `ImageLoader` is a `CachedImageClient` for the `CachedImage`. So I expect this condition will always fail and we will never stop the animation for any image even if it is not displayed in the page.

Also after we `removeClient(*m_renderer)` and the numberOfClients() is 1, why do we `stopAnimation()`?

nullhook: This is fragile for sure. The real issue is `numberOfClients` mixes renderer and non-renderer clients, so there's no clean way to check this.

nullhook: I think a proper fix would be to _only_ stop animation when there are no clients.

Source/WebCore/rendering/RenderImageResource.cpp
```diff
@@ -60,9 +60,11 @@ void RenderImageResource::initialize(RenderElement& renderer)
 
 void RenderImageResource::willBeDestroyed()
 {
-    image()->stopAnimation();
+    RefPtr cachedImage = this->cachedImage();
     if (m_styleImage && m_renderer)
         m_styleImage->removeClient(*m_renderer);
+    if (cachedImage && !cachedImage->numberOfRendererClients())
+        image()->stopAnimation();
```
shallawa: I think we should use `CachedImage::isVisibleInViewport()` instead. If `CachedImage` has one or more client after removing `m_renderer` but all of them are outside the viewport, then we still need stop the animation.

```
if (cachedImage && m_renderer && !cachedImage->isVisibleInViewport(protect(m_renderer->document())))
```

`CachedImage::isVisibleInViewport()` is expensive but it is more accurate in this context.

shallawa (approved):

Source/WebCore/rendering/RenderImageResource.cpp
nullhook: This approach results in a failure `fast/images/stopped-animation-deleted-image.html` and also seems fragile. There seems to be a timing issue here. I think `isVisibleInViewport` is asking a in-flight removed renderer here.

nullhook: I think the renderer count approach is more reliable.

shallawa: This test is fragile since it uses two setTimeout(50). Admittedly I wrote it myself. If you can rewrite it to use `webkitImageFrameReady` this will be great and it allow using `isVisibleInViewport()`.

PR #64393, opened 2026-05-06
[JSC] Rewrite test262 runner

heimskr:
#### 01efee2dab1400e07d101e81c47012c0a4f9f500
<pre>
[JSC] Rewrite test262 runner
<a href="https://bugs.webkit.org/show_bug.cgi?id=314232">https://bugs.webkit.org/show_bug.cgi?id=314232</a>
<a href="https://rdar.apple.com/176394927">rdar://176394927</a>

The old test262 runner is a bit slow and could use some new features.
This patch rewrites the runner in Python and adds:
1. Incremental expectations updating: don&apos;t affect other entries in expectations.yaml
   if you use --save with a subset.
2. OS-/build configuration-based skipping: new skip.conditions section in config.yaml
   with `if: {os, os_version, build, sanitizer}` clauses, plus a --sanitizer CLI flag
   for declaring the current sanitizer
3. Regex test filter: `--filter &lt;regex&gt;` to run tests whose paths match a regex.
4. Exit codes as a means of determining test success/failure, rather than text output.

* JSTests/test262/expectations-linux.yaml:
* JSTests/test262/expectations.yaml:
* Tools/CISupport/build-webkit-org/steps.py:
(RunTest262Tests):
* Tools/CISupport/build-webkit-org/steps_unittest.py:
* Tools/Scripts/filter-test-logs:
* Tools/Scripts/test262-import:
* Tools/Scripts/test262-runner:
* Tools/Scripts/test262/Import.pm: Removed.
* Tools/Scripts/test262/README.md:
* Tools/Scripts/test262/Runner.pm: Removed.
* Tools/Scripts/test262/__init__.py: Copied from Tools/Scripts/test262-runner.
* Tools/Scripts/test262/cpanfile: Removed.
* Tools/Scripts/test262/cpanfile.snapshot: Removed.
* Tools/Scripts/test262/importer.py: Added.
(_die):
(_git):
(_read_revision):
(_get_new_revision):
(_write_revision):
(_normalize_summary):
(_get_summary):
(_write_summary):
(_transfer_files):
(_clone_remote):
(_validate_source_dir):
(_parse_args):
(main):
* Tools/Scripts/test262/importer_unittest.py: Added.
(TestDefaultPaths):
(TestDefaultPaths.test_test262_dir_resolves_into_jstests):
(TestDefaultPaths.test_revision_and_summary_live_beside_tests):
(TestReadRevision):
(TestReadRevision._write):
(TestReadRevision.test_parses_revision):
(TestReadRevision.test_missing_file_returns_none):
(TestReadRevision.test_no_revision_line_returns_none):
(TestWriteRevision):
(TestWriteRevision.test_exact_format):
(TestWriteRevision.test_round_trips_through_read):
(TestNormalizeSummary):
(TestNormalizeSummary.test_strips_prefixes_and_trailing_newline):
(TestNormalizeSummary.test_no_trailing_newline_added):
(TestNormalizeSummary.test_rename_two_paths):
(TestNormalizeSummary.test_paths_with_regex_metacharacters):
(TestNormalizeSummary.test_empty_stays_empty):
(TestWriteSummary):
(TestWriteSummary.test_writes_verbatim_without_trailing_newline):
(TestGetSummary):
(TestGetSummary.test_combines_harness_and_test_diffs):
(TestGetSummary.test_uses_no_index_and_diff_filter):
(TestGetNewRevision):
(TestGetNewRevision.test_returns_revision_and_tracking):
(TestGetNewRevision.test_empty_revision_dies):
(TestValidateSourceDir):
(TestValidateSourceDir._make_valid_source):
(TestValidateSourceDir.test_valid_source_passes):
(TestValidateSourceDir.test_missing_subdirs_dies):
(TestValidateSourceDir.test_same_as_current_dies):
(TestTransferFiles):
(TestTransferFiles.test_replaces_existing_trees):
(TestGit):
(TestGit.test_strips_output):
(TestGit.test_nonzero_exit_dies):
(TestMainFlow):
(TestMainFlow.test_same_revision_skips_import):
(TestMainFlow.test_new_revision_transfers_and_records):
(TestMainFlow.test_dirty_local_source_dies):
* Tools/Scripts/test262/local/bin/instmodsh: Removed.
* Tools/Scripts/test262/local/cache/authors/id/B/BI/BINGOS/ExtUtils-MakeMaker-7.32.tar.gz: Removed.
* Tools/Scripts/test262/local/cache/authors/id/T/TI/TINITA/YAML-1.24.tar.gz: Removed.
* Tools/Scripts/test262/local/cache/authors/id/Y/YA/YANICK/Parallel-ForkManager-1.19.tar.gz: Removed.
* Tools/Scripts/test262/local/cache/modules/02packages.details.txt: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/Command.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/Command/MM.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/Liblist.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/Liblist/Kid.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_AIX.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_Any.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_BeOS.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_Cygwin.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_DOS.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_Darwin.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_MacOS.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_NW5.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_OS2.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_QNX.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_UWIN.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_Unix.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_VMS.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_VOS.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_Win32.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MM_Win95.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MY.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MakeMaker.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MakeMaker/Config.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MakeMaker/FAQ.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MakeMaker/Locale.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MakeMaker/Tutorial.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MakeMaker/version.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MakeMaker/version/regex.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/MakeMaker/version/vpp.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/Mkbootstrap.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/Mksymlists.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/ExtUtils/testlib.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/Parallel/ForkManager.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Any.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Any.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Dumper.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Dumper.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Dumper/Base.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Dumper/Base.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Error.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Error.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Loader.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Loader.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Loader/Base.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Loader/Base.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Marshall.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Marshall.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Mo.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Node.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Node.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Tag.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Tag.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Types.pm: Removed.
* Tools/Scripts/test262/local/lib/perl5/YAML/Types.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/darwin-2level/.meta/ExtUtils-MakeMaker-7.32/MYMETA.json: Removed.
* Tools/Scripts/test262/local/lib/perl5/darwin-2level/.meta/ExtUtils-MakeMaker-7.32/install.json: Removed.
* Tools/Scripts/test262/local/lib/perl5/darwin-2level/.meta/Parallel-ForkManager-1.19/MYMETA.json: Removed.
* Tools/Scripts/test262/local/lib/perl5/darwin-2level/.meta/Parallel-ForkManager-1.19/install.json: Removed.
* Tools/Scripts/test262/local/lib/perl5/darwin-2level/.meta/YAML-1.24/MYMETA.json: Removed.
* Tools/Scripts/test262/local/lib/perl5/darwin-2level/.meta/YAML-1.24/install.json: Removed.
* Tools/Scripts/test262/local/lib/perl5/darwin-2level/auto/ExtUtils/MakeMaker/.packlist: Removed.
* Tools/Scripts/test262/local/lib/perl5/darwin-2level/auto/Parallel/ForkManager/.packlist: Removed.
* Tools/Scripts/test262/local/lib/perl5/darwin-2level/auto/YAML/.packlist: Removed.
* Tools/Scripts/test262/local/lib/perl5/darwin-2level/perllocal.pod: Removed.
* Tools/Scripts/test262/local/lib/perl5/unicore/Exact.pl: Removed.
* Tools/Scripts/test262/local/lib/perl5/unicore/PVA.pl: Removed.
* Tools/Scripts/test262/local/lib/perl5/unicore/lib/gc_sc/SpacePer.pl: Removed.
* Tools/Scripts/test262/local/lib/perl5/unicore/lib/gc_sc/Word.pl: Removed.
* Tools/Scripts/test262/runner.py: Added.
(_YamlDumper):
(_YamlDumper.check_simple_key):
(_YamlDumper.choose_scalar_style):
(_worker_setup):
(_run_batch):
(_die):
(_load_yaml):
(_read_file):
(_parse_metadata):
(_parse_error):
(_detect_platform):
(_evaluate_condition):
(_should_skip):
(_get_scenarios):
(_compile_harness):
(_compile_includes):
(_get_feature_flags):
(_run_jsc):
(_process_test_file):
(_classify_results):
(_discover_tests):
(_load_import_file):
(_find_failing):
(_find_jsc):
(_save_expectations):
(_save_results):
(_summarize_results):
(_generate_html_index):
(_parse_args):
(main):
(main._print_progress):
(main._on_batch_done):
* Tools/Scripts/test262/runner_unittest.py: Added.
(TestParseMetadata):
(TestParseMetadata.test_basic_metadata):
(test_no_metadata):
(test_empty_metadata):
(test_malformed_yaml):
(test_multiple_features):
(TestParseError):
(TestParseError.test_none_input):
(TestParseError.test_empty_input):
(TestParseError.test_exception_format):
(TestParseError.test_exception_multiline):
(TestParseError.test_no_exception_returns_first_line):
(TestDetectPlatform):
(TestDetectPlatform.test_macos_debug):
(TestDetectPlatform.test_release_with_sanitizer):
(TestDetectPlatform.test_linux):
(TestEvaluateCondition):
(TestEvaluateCondition.test_os_match):
(TestEvaluateCondition.test_os_mismatch):
(TestEvaluateCondition.test_os_case_insensitive):
(TestEvaluateCondition.test_os_version):
(TestEvaluateCondition.test_os_version_missing):
(TestEvaluateCondition.test_os_version_invalid):
(TestEvaluateCondition.test_build_match):
(TestEvaluateCondition.test_sanitizer_match):
(TestEvaluateCondition.test_multiple_conditions_all_must_match):
(TestEvaluateCondition.test_empty_condition):
(TestShouldSkip):
(TestShouldSkip._skip):
(TestShouldSkip.test_skip_by_file):
(TestShouldSkip.test_skip_by_path_pattern):
(TestShouldSkip.test_no_skip_without_config):
(TestShouldSkip.test_skip_by_feature):
(TestShouldSkip.test_filter_features_overrides_skip):
(TestShouldSkip.test_filter_features_skips_non_matching):
(TestShouldSkip.test_filter_features_without_config):
(TestShouldSkip.test_conditional_skip_by_file):
(TestShouldSkip.test_conditional_skip_not_matching_platform):
(TestShouldSkip.test_conditional_skip_by_path):
(TestShouldSkip.test_conditional_skip_by_feature):
(TestShouldSkip.test_conditional_skip_feature_overridden_by_filter):
(TestGetScenarios):
(TestGetScenarios.test_no_flags):
(TestGetScenarios.test_none_flags):
(TestGetScenarios.test_raw):
(TestGetScenarios.test_no_strict):
(TestGetScenarios.test_only_strict):
(TestGetScenarios.test_module):
(TestGetScenarios.test_unrecognized_flag):
(TestGetScenarios.test_raw_takes_priority):
(TestGetFeatureFlags):
(TestGetFeatureFlags.test_basic_feature_flag):
(TestGetFeatureFlags.test_multiple_features):
(TestGetFeatureFlags.test_unmapped_feature):
(TestGetFeatureFlags.test_no_features):
(TestGetFeatureFlags.test_no_config):
(TestGetFeatureFlags.test_can_block_is_false):
(TestGetFeatureFlags.test_can_block_with_features):
(TestClassifyResults):
(TestClassifyResults._result):
(TestClassifyResults.test_expected_pass):
(TestClassifyResults.test_expected_pass_no_expectations):
(TestClassifyResults.test_unexpected_pass):
(TestClassifyResults.test_expected_fail_matching_exit_code):
(TestClassifyResults.test_unexpected_fail_different_exit_code):
(TestClassifyResults.test_unexpected_fail_no_expectations):
(TestClassifyResults.test_unexpected_fail_no_entry):
(TestClassifyResults.test_crash_can_match_expectation):
(TestClassifyResults.test_skipped):
(TestClassifyResults.test_strict_mode_independent):
(TestClassifyResults.test_pass_with_expected_zero_is_expected):
(TestCompileHarness):
(TestCompileHarness.test_concatenates_files):
(TestCompileIncludes):
(TestCompileIncludes.test_creates_temp_file):
(TestCompileIncludes.test_missing_include_returns_none):
(TestCompileIncludes.test_multiple_includes):
(TestDiscoverTests):
(TestDiscoverTests.test_finds_js_files):
(TestDiscoverTests.test_missing_directory):
(TestDiscoverTests.test_sorted_output):
(TestLoadImportFile):
(TestLoadImportFile.test_parses_added_and_modified):
(TestSaveExpectations):
(TestSaveExpectations._make_result):
(TestSaveExpectations.test_full_save):
(TestSaveExpectations.test_documents_exit_codes):
(TestSaveExpectations.test_incremental_preserves_unrun):
(TestSaveExpectations.test_incremental_adds_new_failure):
(TestSaveExpectations.test_incremental_removes_now_passing):
(TestSaveExpectations.test_incremental_skipped_not_counted_as_run):
(TestRunJsc):
(TestRunJsc._base_env):
(TestRunJsc.test_pass_returns_none_output):
(TestRunJsc.test_failure_returns_output):
(TestRunJsc.test_uses_env_and_fork_exec):
(TestRunJsc.test_strict_mode_uses_strict_file):
(TestRunJsc.test_module_uses_module_file):
(TestRunJsc.test_raw_skips_harness):
(TestRunJsc.test_async_includes_async_harness):
(TestRunJsc.test_negative_exception):
(TestRunJsc.test_watchdog_timeout):
(TestRunJsc.test_feature_flags_in_command):
(TestRunJsc.test_timeout_exception):
(TestRunJsc.test_no_dyld_without_darwin):
(TestRunJsc.test_signal_exit_code_normalized):
(TestCrashDetection):
(TestCrashDetection.test_exit_zero_no_crash):
(TestCrashDetection.test_exit_three_special_cased):
(TestCrashDetection.test_sigsegv_detected):
(TestCrashDetection.test_sigabrt_detected):
(TestCrashDetection.test_sigbus_detected):
(TestSaveResults):
(TestSaveResults.test_round_trip):
(TestFindJsc):
(TestFindJsc.test_explicit_path):
(TestFindJsc.test_explicit_path_missing):
(TestFindJsc.test_port_build_dir):
(TestFindJsc.test_port_build_dir_missing):
(TestFindJsc.test_port_does_not_fall_back_to_path):
(TestTestFileRegex):
(TestTestFileRegex.test_matches_js):
(TestTestFileRegex.test_rejects_fixture):
(TestTestFileRegex.test_rejects_non_js):
(TestParseArgs):
(TestParseArgs._parse):
(TestParseArgs.test_build_config_optional_defaults_to_release):
(TestParseArgs.test_explicit_release):
(TestParseArgs.test_explicit_debug):
(TestParseArgs.test_release_and_debug_are_mutually_exclusive):
(TestEndToEnd):
(TestEndToEnd.setUp):
(TestEndToEnd._write_yaml):
(TestEndToEnd._run):
(TestEndToEnd.test_failure_without_expectations):
(TestEndToEnd.test_pass_without_expectations):
(TestEndToEnd.test_newly_failing):
(TestEndToEnd.test_newly_passing):
(TestEndToEnd.test_expected_failure_is_not_new):
(TestEndToEnd.test_crash_where_failure_was_expected_is_new):
(TestEndToEnd.test_test_only_accepts_a_single_file):
(TestEndToEnd.test_no_tests_found_is_an_error):
(TestEndToEnd.test_skipped_files_reports_newly_passing):
(TestEndToEnd.test_skipped_files_does_not_report_expected_failures):
(TestEndToEnd.test_save_records_exit_codes):
* Tools/Scripts/webkitperl/test262_unittest/README: Removed.
* Tools/Scripts/webkitperl/test262_unittest/fixtures/expectations-compare.yaml: Removed.
* Tools/Scripts/webkitperl/test262_unittest/fixtures/expectations.yaml: Removed.
* Tools/Scripts/webkitperl/test262_unittest/fixtures/mock-jsc-fail-new-error.pl: Removed.
* Tools/Scripts/webkitperl/test262_unittest/fixtures/mock-jsc-fail.pl: Removed.
* Tools/Scripts/webkitperl/test262_unittest/fixtures/test/expected-to-fail-now-failing-with-new-error.js: Removed.
* Tools/Scripts/webkitperl/test262_unittest/fixtures/test/expected-to-fail-now-failing.js: Removed.
* Tools/Scripts/webkitperl/test262_unittest/fixtures/test/expected-to-fail-now-passing.js: Removed.
* Tools/Scripts/webkitperl/test262_unittest/fixtures/test/expected-to-pass-now-failing.js: Removed.
* Tools/Scripts/webkitperl/test262_unittest/fixtures/test/fail.js: Removed.
* Tools/Scripts/webkitperl/test262_unittest/fixtures/test/pass.js: Removed.
* Tools/Scripts/webkitperl/test262_unittest/test262-runner-tests.pl: Removed.
* Tools/Scripts/webkitpy/test/main.py:
(main):

Canonical link: <a href="https://commits.webkit.org/318622@main">https://commits.webkit.org/318622@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/2a29fcfa04532221b735f4b1190740c3a753b630

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/178504 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/52442 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/46237 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/188120 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/141753 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/180367 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/53736 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/52152 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/139169 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/96640 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/25174df3-7a26-4acd-aebf-929930952920) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/181461 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/42726 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/159916 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/119623 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/3acd2884-a8a0-4aa2-92e4-e1d876ea8055) 
| [❌ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/177827 "Found 1 webkitpy test failure: webkitpy.style.checker_unittest.GlobalVariablesTest.test_webkit_base_filter_rules") | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/41392 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/39742 "Passed tests") | [✅ 🧪 jsc-wpe](https://ews-build.webkit.org/#/builders/251/builds/1592 "Passed tests") | | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/8405 "Passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/151792 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/39415 "Passed tests") | [⏳ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/GTK-GTK3-GCC-Build-EWS "Waiting in queue, processing has not started yet") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/39777 "Built successfully and passed tests") | | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/43984 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/190547 "Built successfully") | | 
| [✅ 🧪 services](https://ews-build.webkit.org/#/builders/28/builds/177907 "Passed tests") | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/52111 "Built successfully") | | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/148328 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/52792 "Built successfully") | [❌ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/36039 "Found 1 new test failure: imported/w3c/web-platform-tests/web-locks/held.https.any.sharedworker.html (failure)") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/148098 "Passed tests") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/27138 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/42807 "Passed tests") | | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/119695 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/51310 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/14390 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/50695 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/50935 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/50757 "Built successfully") | | | |


Tools/Scripts/test262/runner_unittest.py
```diff
@@ -0,0 +1,1037 @@
+#!/usr/bin/env python3
```
gsnedders, line 1: This should be `runner_unittest.py`, based on how we normally name unittests, and run them with `test-webkitpy` (and we should make sure this gets included).

Tools/Scripts/test262/compare.py
```diff
@@ -0,0 +1,291 @@
...
+import sys
+import tempfile
+import time
+from pathlib import Path
+
+try:
+    import yaml
+except ImportError:
+    print("Error: PyYAML required. pip3 install pyyaml", file=sys.stderr)
+    sys.exit(1)
```
gsnedders: By and large, we use `webkitcorepy`'s autoinstaller to handle Python dependencies, rather than relying on packages being installed in the environment — and we should do this here too.

Tools/Scripts/test262/compare.py
```diff
@@ -0,0 +1,291 @@
...
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
+# OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR
+# TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF
+# THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+# SUCH DAMAGE.
+
+"""Diff the Python test262-runner against the legacy Perl test262-runner.pl.
```
gsnedders: Do we actually expect to keep both runners around for long? Clearly a script like this is helpful while developing the Python rewrite, but it's not clear to me that we're likely to actually _keep_ wanting to run both?

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+
+
+class _YamlDumper(yaml.SafeDumper):
+    # PyYAML's SafeDumper (like libyaml) refuses to emit an inline key
+    # (`key: value`) when the key scalar is longer than 128 chars, and falls
+    # back to the explicit `? key\n: value` form. Our long test paths blow
+    # past that and the Perl consumer can't parse the explicit form. Override
+    # to allow any length simple key.
+    def check_simple_key(self):
+        from yaml.events import ScalarEvent, AliasEvent
```
gsnedders: Imports should be all the top of the file, unless we have a strong reason not to.

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+
+def _worker_run_file(filepath):
+    return _process_test_file(filepath, _ctx)
+
+
+def _worker_run_batch(filepaths):
+    results = []
+    for fp in filepaths:
+        results.extend(_process_test_file(fp, _ctx))
+    return results
```
gsnedders, line 150: What's the purpose of all of this? Are we just trying to avoid passing `ctx` for each test?

At the very least, we should be storing things on `multiprocessing.current_process()` rather than the module-global (because that could get confusing on Linux where multiprocessing forks by default).

As a very silly example, you could do something like:

```python
class _Worker:
    def __init__(self, ctx):
        self.ctx = ctx
        signal.signal(signal.SIGINT, signal.SIG_IGN)

    def run(self, filepath):
        return _process_test_file(filepath, self.ctx)


def _init_worker(ctx):
    multiprocessing.current_process()._run = _Worker(ctx).run

def _call(filepath):
    return multiprocessing.current_process()._run(filepath)
```

But really, I'd think that we probably want `_process_test_file` as a method on the worker?

Or, if we want to have the ability to run in-process, have a parent class containing almost everything, with a Worker subclass just additionally setting the signal handler.

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+    print(
+        "Error: PyYAML is required. Install with: pip3 install pyyaml",
+        file=sys.stderr,
+    )
+    sys.exit(1)
+
+try:
+    _YamlLoader = yaml.CSafeLoader
+except AttributeError:
+    _YamlLoader = yaml.SafeLoader
```
gsnedders, line 71: Let's be explicit these are the safe loaders:

```suggestion
try:
    _SafeYamlLoader = yaml.CSafeLoader
except AttributeError:
    _SafeYamlLoader = yaml.SafeLoader
```

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+
+
+def _load_yaml(path):
+    with open(path, "r", encoding="utf-8") as f:
+        return yaml.load(f, Loader=_YamlLoader)
+
+
+def _read_file(path):
+    with open(path, "r", encoding="utf-8", errors="replace") as f:
+        return f.read()
```
gsnedders: I'd strongly recommend against single-use utility functions.

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+    t = t + (0,) * (max_len - len(t))
+    ops = {
+        ">=": a >= t,
+        "<=": a <= t,
+        ">": a > t,
+        "<": a < t,
+        "==": a == t,
+        "!=": a != t,
+    }
+    return ops.get(op, False)
```
gsnedders: We should just use `webkitcorepy.Version`, not reimplement this.

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+            with open(path, "r", encoding="utf-8", errors="replace") as f:
+                content += f.read()
+        except FileNotFoundError:
+            pass
+    if not content:
+        return None
+    fd, path = tempfile.mkstemp(dir=temp_dir, suffix=".js")
+    with os.fdopen(fd, "w", encoding="utf-8") as f:
+        f.write(content)
+    return path
```
gsnedders, line 334: I'd probably favour `tempfile.NamedTemporaryFile` here.

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+            content += f.read()
+    return content
+
+
+def _compile_includes(includes, harness_dir, temp_dir):
+    content = ""
+    for name in includes:
+        path = os.path.join(harness_dir, name)
+        try:
+            with open(path, "r", encoding="utf-8", errors="replace") as f:
```
gsnedders, line 325: Do we actually want to allow invalid UTF-8?

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+
+
+# --- Harness ---
+
+
+def _compile_harness(filenames, harness_dir):
+    content = ""
+    for name in filenames:
+        path = os.path.join(harness_dir, name)
+        with open(path, "r", encoding="utf-8", errors="replace") as f:
```
gsnedders, line 313: Do we actually want to allow invalid UTF-8?

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+    sys.exit(1)
+
+
+def _load_yaml(path):
+    with open(path, "r", encoding="utf-8") as f:
+        return yaml.load(f, Loader=_YamlLoader)
+
+
+def _read_file(path):
+    with open(path, "r", encoding="utf-8", errors="replace") as f:
```
gsnedders, line 167: Do we actually want to allow invalid UTF-8?

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+                "time": 0,
+                "passed": False,
+                "error": f"Read error: {e}",
+                "output": str(e),
+                "features": [],
+                "exit_code": 1,
+                "skipped": False,
+                "is_crash": True,
+            }
+        ]
```
gsnedders, line 456: This feels like an exception we should never be hitting, and treating it as a fatal error for the entire test run seems not unreasonable?

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+                "mode": scenario,
+                "time": exec_time,
+                "passed": passed,
+                "error": error,
+                "output": output if not passed else None,
+                "features": features,
+                "exit_code": exit_code,
+                "skipped": False,
+                "is_crash": is_crash,
+            }
```
gsnedders, line 546: Using a dataclass rather than a dictionary would be nicer.

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+        )
+        if result.returncode == 0 and result.stdout.strip():
+            return result.stdout.strip()
+    except Exception:
+        pass
+
+    _die(
+        f"Cannot find jsc in {build_base / config_name}.\n"
+        "Try --release or specify with --jsc <path>."
+    )
```
gsnedders: I feel like we have this logic in webkitpy already — duplicating this, finding the correct path for the build, seems unfortunate.

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+
+    return parser.parse_args()
+
+
+# --- Main ---
+
+
+def main():
+    args = _parse_args()
+    results_dir = os.path.join(os.getcwd(), "test262-results")
```
gsnedders, line 1177: cwd is slightly surprising here — in most other WebKit scripts we use something relative to the repo root or the build directory!

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+            "skip_set": skip_set,
+            "skip_paths": skip_paths,
+            "skip_features": skip_features,
+            "filter_features": filter_features,
+            "skipped_only": args.skipped_files,
+            "temp_dir": temp_dir,
+            "platform_info": platform_info,
+            "conditional_skips": conditional_skips,
+            "verbose": args.verbose,
+        }
```
gsnedders, line 1378: This should be a dataclass, if not part of an actual class?

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+            # balancing headroom.
+            batch_size = max(1, min(50, total_files // (num_processes * 4) or 1))
+            batches = [
+                files[i:i + batch_size]
+                for i in range(0, total_files, batch_size)
+            ]
+            futures = {
+                executor.submit(_worker_run_batch, batch): batch
+                for batch in batches
+            }
```
gsnedders: Why not just use `executor.map` here and use its own `chunksize` parameter rather than re-implementing this?

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+                except KeyboardInterrupt:
+                    if not args.no_progress:
+                        print()
+                    print(
+                        "\nInterrupted. Exiting...",
+                        file=sys.stderr,
+                    )
+                    # Kill workers so their JSC subprocesses die.
+                    try:
+                        import multiprocessing
```
gsnedders: Imports should be at the top.

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+                        for proc in multiprocessing.active_children():
+                            try:
+                                proc.kill()
+                            except (OSError, ProcessLookupError):
+                                pass
```
gsnedders: On Python 3.14 and later, we should just call `ProcessPoolExecutor.kill_workers()`; on earlier versions, maybe see the implementation of that (https://github.com/python/cpython/blob/8732053c92fdf3120e94ea2b8fc2d39b372f0602/Lib/concurrent/futures/process.py#L930) to see what you should probably be doing (including shutting down the pool first!).

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+                    # Kill multiprocessing's resource_tracker helper so it
+                    # doesn't print a leaked-semaphore warning after we exit.
+                    # The OS reclaims the semaphores when the process dies.
+                    try:
+                        from multiprocessing import resource_tracker
+                        rt_pid = resource_tracker._resource_tracker._pid
+                        if rt_pid is not None:
+                            os.kill(rt_pid, signal.SIGKILL)
+                    except Exception:
+                        pass
```
gsnedders: IIRC, this is specific to some older Python versions, and it would be nice to avoid this?

JSTests/test262/test262-Revision.txt
```diff
@@ -1,2 +1,2 @@
 test262 remote url: https://github.com/tc39/test262.git
-test262 revision: 578794183cf5f51a453ff9a793416f2ae20d3458
+test262 revision: d0c1b4555b03dd404873fd6422a4b5da00136500
```
gsnedders: It would be nice to not change the revision as part of the rewrite, because this potentially makes it harder to know that all the expectation changes are from the change of format.

JSTests/test262/expectations.yaml
```diff
@@ -1,949 +1,952 @@
 ---
 test/annexB/language/function-code/block-decl-func-skip-arguments.js:
-  default: 'Test262Error: Expected SameValue(«"function arguments() {}"», «"[object Arguments]"») to be true'
+  default: 3
```
gsnedders, line 21: How does this not break `compare.py`? Is the Perl code not actually reading the default?

gsnedders (changes requested):
While I'm sure some of my comments can be easily dismissed with "this just matches the Perl, we should land this then fix it to be more like the rest of everything else later", there are some more notable comments here. There's also just a decent amount of "why not use the existing library code in webkitcorepy/webkitpy for this", because we don't need to re-invent the wheel.

Beyond my comments below, I do question some of the code structure here — it feels like we should split this into multiple files, and likely use classes in some places to reduce the amount of closely-related functions we have, especially with passing around large numbers of arguments.

Additionally, I'd _really_ like us to not add a new Python script without having type annotations in it — test runners are always full of almost-never-taken branches, and type-checking does at least give us some confidence of correctness there.

And finally: I do wonder how much we actually need multiple processes here? I'd hope/expect basically all the execution time is in `jsc`, rather than the Python harness. If so, then we could presumably just use `asyncio` to run all of this, and avoid the IPC/pickle overhead entirely.

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1498 @@
...
+        skip_section = config["skip"]
+        skip_set = set(skip_section.get("files") or [])
+        skip_paths = skip_section.get("paths") or []
+        skip_features_raw = skip_section.get("features") or []
+        skip_features = set()
+        for f in skip_features_raw:
+            m = re.match(r"(\S+)", str(f))
+            if m:
+                skip_features.add(m.group(1))
+        conditional_skips = skip_section.get("conditions") or []
```
gsnedders, line 1252: Given this is new: I think there's reasonable questions about whether we want this to work in different and novel ways to how we have conditions everywhere else — in `webkitpy`, in `webkitexpectationspy`, etc.

I think having it in `expectations.yaml` isn't silly, but it is yet another variant of this. :)

JSTests/test262/expectations.yaml
heimskr: compare.py was used before I switched to exit codes and probably should be removed from the PR.

heimskr:
> And finally: I do wonder how much we actually need multiple processes here? I'd hope/expect basically all the execution time is in `jsc`, rather than the Python harness. If so, then we could presumably just use `asyncio` to run all of this, and avoid the IPC/pickle overhead entirely.

The test runner has/had a lot of overhead. A full run of the entire suite with a release build of jsc used to take about 30 minutes on my machine with the old Perl version, vs. about 12 minutes on the same machine with the Python version.

clopez:
I tried to test this but I couldn't because it doesn't support `--platform wpe` or `--wpe` (or `--gtk` or `--jsc-only`). And if I don't pass any of that platform parameters then it uses the /bin/jsc binary installed from my system distribution (Debian) rather than the one I have built at WebKitBuild.

This test262 script is run on the Linux GTK and WPE JS bots. See: https://build.webkit.org/#/builders?tags=%2BLinux&tags=%2BJS

On Linux, on a given WebKit checkout is possible to have three different ports built at the same time of jsc: WPE, GTK and JSCOnly

```
WebKit $ ls -1 WebKitBuild/*/*/*/jsc
WebKitBuild/GTK/Release/bin/jsc 
WebKitBuild/JSCOnly/Release/bin/jsc
WebKitBuild/WPE/Release/bin/jsc
```

So the script needs to keep supporting passing a platform parameter to select which jsc program to run.

clopez:
> 4. Exit codes as a means of determining test success/failure, rather than text output.

That is an interesting decision

That means we will miss checking when a WebKit commit causes a test in Test262 to give different text output as long as the failure code is the same.

But is also true that the same happens for many of the layout tests, we just mark them as Fail and we don't notice if the text diff changes as long as it is different.

I see the codes are 0 (pass), 3 (fail), and 128+${CRASHSIGNALNUMBER} (crash), but i had to read the code to understand this. It would be nice to document this in the yaml file with a comment so is straightforward for anyone trying to change the expectations file what those numbers mean.

Tools/Scripts/test262/runner.py
```diff
@@ -0,0 +1,1473 @@
...
+            em, es = divmod(int(elapsed), 60)
+            etm, ets = divmod(int(eta), 60)
+            end_char = "\n" if final else ""
+            print(
+                f"\r[{completed}/{total_files}] "
+                f"{rate:5.1f}/s  {em}m{es:02d}s elapsed  "
+                f"ETA {etm}m{ets:02d}s   ",
+                end=end_char,
+                flush=True,
+            )
```
clopez, line 1411: This 1-line progress looks nice on a interactive TTY, but when this runs on buildbot and the output gets redirected to a log file it is going to look pretty bad.

Suggested change:

```diff
--- a/Tools/Scripts/test262/runner.py
+++ b/Tools/Scripts/test262/runner.py
@@ -1319,8 +1319,9 @@ def main() -> None:
             em, es = divmod(int(elapsed), 60)
             etm, ets = divmod(int(eta), 60)
             end_char = "\n" if final else ""
+            line_char = "\r" if sys.stdout.isatty() else "\n"
             print(
-                f"\r[{completed}/{total_files}] "
+                f"{line_char}[{completed}/{total_files}] "
                 f"{rate:5.1f}/s  {em}m{es:02d}s elapsed  "
                 f"ETA {etm}m{ets:02d}s   ",
                 end=end_char,

```

heimskr: Good idea. I implemented the change and also changed the progress update interval so that at least 5 seconds must pass between each line of non-TTY output. Otherwise, with the normal 0.1 seconds minimum, you'd probably get hundreds of lines of log output per run.

heimskr:
> That is an interesting decision

The rationale is that the text output of test262 tests is generally not very important; if the test fails, it'll throw. I don't think there are many cases where a regression will be indicated by a test starting to have different text output without outright failing.

Tools/Scripts/test262/runner.py
clopez, line 1064: There is an issue here. Both are False by default. That shouldn't happen. You need to pass `required=True` to `parser.add_mutually_exclusive_group()` so the user is forced to either specify `--release` or `--debug`, otherwise both default to false and cause issues later.

For example, when I was testing this happened:

```
$ Tools/Scripts/test262-runner --wpe 
Error: Cannot find jsc in /home/clopez/webkit/webkit/WebKitBuild/WPE/Debug.
Build it, or specify an existing binary with --jsc <path>.
```

It took a default of debug because later the code does `config_name = "Release" if release else "Debug"` but `args.debug` was False.

Also if it has to default to something it should default to release and not to debug.

heimskr: Fixed.

clopez:
Ok, is looking good to me.

Just some final fixes to make it run well on Linux.

1. Please add this patch on top of yours.
```diff
diff --git a/Tools/Scripts/test262-runner b/Tools/Scripts/test262-runner
index 0771e8e550be..e9c86148a9d8 100755
--- a/Tools/Scripts/test262-runner
+++ b/Tools/Scripts/test262-runner
@@ -31,6 +31,10 @@
 import os
 import sys
 
+from webkitpy.port.linux_container_sdk_utils import maybe_enter_webkit_container_sdk, maybe_use_container_sdk_root_dir
+maybe_enter_webkit_container_sdk()
+maybe_use_container_sdk_root_dir()
+
 script_dir = os.path.dirname(os.path.realpath(__file__))
 sys.path.insert(0, os.path.join(script_dir, "test262"))
 
```
This is needed to enable the auto-enter mechanism inside the WebKit container SDK on Linux that is used on the GTK/WPE bots.

2. When you rebase it on top of main you are going to find a conflict with https://commits.webkit.org/318294@main -- to resolve it simply do something like is done on that patch. Suggestion:
```diff
diff --git a/Tools/Scripts/test262/runner.py b/Tools/Scripts/test262/runner.py
index 9a5352e7e562..ffaea8942e7a 100644
--- a/Tools/Scripts/test262/runner.py
+++ b/Tools/Scripts/test262/runner.py
@@ -98,7 +98,7 @@ class _YamlDumper(yaml.SafeDumper):
 
 _DEFAULT_TEST262_DIR = _WEBKIT_ROOT / "JSTests" / "test262"
 _DEFAULT_CONFIG = _DEFAULT_TEST262_DIR / "config.yaml"
-_DEFAULT_EXPECTATIONS = _DEFAULT_TEST262_DIR / "expectations.yaml"
+_DEFAULT_EXPECTATIONS = _DEFAULT_TEST262_DIR / "expectations-linux.yaml" if sys.platform.startswith("linux") else "expectations.yaml"
 _REPORT_CSS = _MODULE_DIR / "report.css"
 
 DEFAULT_HARNESS_FILES = ["sta.js", "assert.js"]
```
And then on the `expectations-linux.yaml` file change the string failures for the exit codes.

Thanks

Tools/Scripts/test262/runner.py
clopez: I have noticed that this introduces a undesired side effect: now `--release` or `--debug` is required even for `--stats` or when an explicit `--jsc` path is supplied.

On the original runner specifying `--release` or `--debug` was not mandatory, but it defaulted to `--release` instead of debug. So if you want, revert this but change the default to be `--release` so it matches the original runner behavior

clopez:
I have pushed to the branch [`bug-314232-review` on my fork](https://github.com/clopez/WebKit/tree/bug-314232-review) at Github [a commit on top of yours](https://github.com/clopez/WebKit/commit/aa5156e2eb3cd1eab4fab99a4537d95d095bc22a) ([.patch file here](https://github.com/clopez/WebKit/commit/aa5156e2eb3cd1eab4fab99a4537d95d095bc22a.patch)) with a review done by Claude comparing this python rewrite against the original perl runner to ensure no regressions are introduced.

There are 6 issues flagged there. Please check those (the details are on the commit log) and feel free to apply it partially or as a whole (no need for credit on the commit log).

Also, on top of that, I think is needed to clear all the leftovers from the old runner:

 - `Tools/Scripts/test262/Import.pm`
 - `Tools/Scripts/test262/Runner.pm`
 - `Tools/Scripts/test262/cpanfile*`
 - directory `Tools/Scripts/test262/local`
 - `Tools/Scripts/test262/README.md` (needs update)

heimskr:
@clopez Thanks for the changes, they look good! I applied them just now.

clopez:
cool, BTW; please remove also the lefotovers from the previous runner; i think all this files/directories are not longer needed and can be simply removed:

- Tools/Scripts/test262/Import.pm
- Tools/Scripts/test262/Runner.pm
- Tools/Scripts/test262/cpanfile*
- directory Tools/Scripts/test262/local
- Tools/Scripts/test262/README.md (needs update)

also check https://github.com/WebKit/WebKit/pull/64393#discussion_r3703053264 

and please add this patch so filter-test-logs also prints to stdout on the bots the final summary
```diff
diff --git a/Tools/Scripts/filter-test-logs b/Tools/Scripts/filter-test-logs
index a6bd446b8044..60081647fa9f 100755
--- a/Tools/Scripts/filter-test-logs
+++ b/Tools/Scripts/filter-test-logs
@@ -82,7 +82,7 @@ FILTERS = {
         'start_printing': lambda line: 'tests ran as expected' in line,
     },
     'test262': {
-        'start_printing': lambda line: 'TESTS SUMMARY---' in line,
+        'start_printing': lambda line: 'TESTS SUMMARY---' in line or 'tests run' in line,
         'always_print': lambda line: line.startswith('! NEW FAIL'),
     },
     'minibrowser': {
```

clopez (approved):
Thanks!
This looks good to me and is an improvement over the original perl runner.

Please coordinate landing this with a restart of build.webkit.org buildbot server.

Because once you land this the bots that run test262 tests are going to break (they will try to use perl to run the runner).

Buildbot server needs to be restarted to make the changes effective, ask for that on the #dev channel of webkit slack // cc @aj062

PR #64488, opened 2026-05-07
Cache Intl.DateTimeFormat

justinmichaud:
#### 8d261de036d46e2f57473eb3c8c65f43bbf77324
<pre>
Cache Intl.DateTimeFormat
<a href="https://bugs.webkit.org/show_bug.cgi?id=314337">https://bugs.webkit.org/show_bug.cgi?id=314337</a>

Reviewed by Keith Miller.

We cache new Intl.DateTimeFormat like other engines. A LRU cache keeps
the underlying const impl object cached, and JS wrappers share a reference
to it now.

We also register a new language change observer to invalidate our cache.

We restructure the resetting code a bit so that it is clear when
we are resetting due to TZ or language, even if the actual reset
logic is the same.

date-time-complex.js goes 200-1000x faster (simple vs complex case).

While touching this code, we also fix a related bug: canonicalizeLocaleList
should defineOwnProperty on the resulting array, we shouldn&apos;t observe
this. I added a test for this case too, and now we match v8&apos;s behaviour
with and without this new cache.

Finally, I have included some llm-generated tests to make sure we hit
a bunch of these edge cases. They are a bit verbose for my tastes,
but I figured that including them could still be useful for the fuzzer.

Canonical link: <a href="https://commits.webkit.org/317628@main">https://commits.webkit.org/317628@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/76fe350cb82a16ad0377d307db14725802adbc19

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/175647 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/49580 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/43363 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/185239 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/137825 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/50872 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/49262 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/136782 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/137825 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/178603 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/40207 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/157545 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/117260 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/38849 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/37232 "Passed tests") | [⏳ 🧪 jsc-wpe](https://ews-build.webkit.org/#/builders/JSC-Tests-WPE-EWS "Waiting to run tests") | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/6054 "Passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/149268 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/37037 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/33709 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/36911 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/41271 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/41438 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/187660 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/49247 "Built successfully") | | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/145751 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/39258 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/49927 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/33661 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/145589 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/40406 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/122122 "Built successfully") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/115948 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/48434 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/12014 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/47836 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/48068 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/47893 "Built successfully") | | |

### DIFF
--- a/Source/JavaScriptCore/runtime/IntlCache.cpp
+++ b/Source/JavaScriptCore/runtime/IntlCache.cpp
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2020-2023 Apple Inc. All rights reserved.
+ * Copyright (C) 2026 Igalia S.L.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -25,14 +26,37 @@
 
 #include "config.h"
 #include "IntlCache.h"
-#include "IntlObject.h"
 
+#include "IntlDateTimeFormat.h"
+#include "IntlObject.h"
+#include <atomic>
+#include <mutex>
+#include <wtf/Language.h>
 #include <wtf/TZoneMallocInlines.h>
 #include <wtf/Vector.h>
 
 namespace JSC {
 
 WTF_MAKE_TZONE_ALLOCATED_IMPL(IntlCache);
+
+std::atomic<uint64_t> IntlCache::s_languagesEpoch { 1 };
+
+void IntlCache::ensureLanguageChangeObserver()
+{
+    static std::once_flag onceKey;
+    std::call_once(onceKey, [] {
+        WTF::addLanguageChangeObserver(&s_languagesEpoch, [](void*) {
+            s_languagesEpoch.fetch_add(1, std::memory_order_release);
+        });
+    });
+}
+
+IntlCache::IntlCache()
+    : m_lastSeenLanguagesEpoch(s_languagesEpoch.load(std::memory_order_acquire))
+{
+}
+
+IntlCache::~IntlCache() = default;
 
 UDateTimePatternGenerator* IntlCache::cacheSharedPatternGenerator(const CString& locale, UErrorCode& status)
 {
--- a/Source/JavaScriptCore/runtime/IntlCache.h
+++ b/Source/JavaScriptCore/runtime/IntlCache.h
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2020-2023 Apple Inc. All rights reserved.
+ * Copyright (C) 2026 Igalia S.L.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
@@ -25,10 +26,16 @@
 
 #pragma once
 
+#include "IntlDateTimeFormat.h"
+#include <atomic>
+#include <optional>
 #include <unicode/udatpg.h>
 #include <wtf/HashMap.h>
 #include <wtf/Noncopyable.h>
+#include <wtf/RefPtr.h>
 #include <wtf/TZoneMalloc.h>
+#include <wtf/TinyLRUCache.h>
+#include <wtf/Vector.h>
 #include <wtf/text/CString.h>
 #include <wtf/text/StringHash.h>
 #include <wtf/text/WTFString.h>
@@ -36,18 +43,55 @@
 
 namespace JSC {
 
+struct IntlDateTimeFormatImplKey {
+    std::optional<String> locales;
+    IntlDateTimeFormat::RequiredComponent required { IntlDateTimeFormat::RequiredComponent::Any };
+    IntlDateTimeFormat::Defaults defaults { IntlDateTimeFormat::Defaults::All };
+
+    friend bool operator==(const IntlDateTimeFormatImplKey&, const IntlDateTimeFormatImplKey&) = default;
+};
+
 class IntlCache {
     WTF_MAKE_NONCOPYABLE(IntlCache);
     WTF_MAKE_TZONE_ALLOCATED(IntlCache);
 public:
-    IntlCache() = default;
+    IntlCache();
+    ~IntlCache();
+
+    static void ensureLanguageChangeObserver();
 
     Vector<char16_t, 32> getBestDateTimePattern(const CString& locale, std::span<const char16_t> skeleton, UErrorCode&);
     Vector<char16_t, 32> getFieldDisplayName(const CString& locale, UDateTimePatternField, UDateTimePGDisplayWidth, UErrorCode&);
 
     String canonicalizeUnicodeLocaleID(const String& languageTag);
 
+    // The cache is reset only at the VM entry service scope; get and insert never check.
+    RefPtr<const IntlDateTimeFormatImpl> findCachedDateTimeFormatImpl(const IntlDateTimeFormatImplKey& key)
+    {
+        if (auto cached = m_cachedDateTimeFormatImpls.findIfCached(key))
+            return *cached;
+        return nullptr;
+    }
+
+    void cacheDateTimeFormatImpl(const IntlDateTimeFormatImplKey& key, Ref<const IntlDateTimeFormatImpl>&& impl)
+    {
+        m_cachedDateTimeFormatImpls.insert(key, RefPtr<const IntlDateTimeFormatImpl>(WTF::move(impl)));
+    }
+
+    void clearForTimeZoneChange() { clearDateTimeFormatImplCache(); }
+    bool hasLanguageChange() const { return m_lastSeenLanguagesEpoch != s_languagesEpoch.load(std::memory_order_acquire); }
+    void clearForLanguageChange()
+    {
+        m_lastSeenLanguagesEpoch = s_languagesEpoch.load(std::memory_order_acquire);
+        clearDateTimeFormatImplCache();
+    }
+
 private:
+    void clearDateTimeFormatImplCache() { m_cachedDateTimeFormatImpls.clear(); }
+
+    static constexpr size_t dateTimeFormatImplCacheCapacity = 4;
+    using DateTimeFormatImplCache = WTF::TinyLRUCache<IntlDateTimeFormatImplKey, RefPtr<const IntlDateTimeFormatImpl>, dateTimeFormatImplCacheCapacity>;
+
     UDateTimePatternGenerator* getSharedPatternGenerator(const CString& locale, UErrorCode& status)
     {
         if (m_cachedDateTimePatternGenerator) {
@@ -62,6 +106,11 @@ private:
     std::unique_ptr<UDateTimePatternGenerator, ICUDeleter<udatpg_close>> m_cachedDateTimePatternGenerator;
     CString m_cachedDateTimePatternGeneratorLocale;
     UncheckedKeyHashMap<String, String> m_cachedCanonicalizedLocaleIDs;
+
+    static std::atomic<uint64_t> s_languagesEpoch;
+
+    DateTimeFormatImplCache m_cachedDateTimeFormatImpls;
+    uint64_t m_lastSeenLanguagesEpoch { 0 };
 };
 
 } // namespace JSC
--- a/Source/JavaScriptCore/runtime/IntlDateTimeFormat.cpp
+++ b/Source/JavaScriptCore/runtime/IntlDateTimeFormat.cpp
@@ -1,6 +1,7 @@
 /*
  * Copyright (C) 2015 Andy VanWagoner (andy@vanwagoner.family)
  * Copyright (C) 2016-2025 Apple Inc. All rights reserved.
+ * Copyright (C) 2026 Igalia S.L.
  *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions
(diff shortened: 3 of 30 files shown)


Constellation (commented):
Let's split timezone change notification part as a separate change to make it easier to be reviewed and landed (and monitored via perf dashboard etc.). That's currently only exists on Cocoa platform and it is generally useful.

Source/JavaScriptCore/runtime/IntlCache.cpp
```diff
@@ -25,15 +26,38 @@
...
+#include <mutex>
+#include <wtf/Language.h>
 #include <wtf/TZoneMallocInlines.h>
 #include <wtf/Vector.h>
 
 namespace JSC {
 
 WTF_MAKE_TZONE_ALLOCATED_IMPL(IntlCache);
 
+std::atomic<uint64_t> IntlCache::s_languagesGeneration { 1 };
```
kmiller68: Nit: I think epoch is the more common term of art although I suppose that could be confusing with the Unix Epoch.

Source/JavaScriptCore/runtime/IntlCache.h
```diff
@@ -25,29 +26,78 @@
...
+        return nullptr;
+    }
+
+    void cacheDateTimeFormatImpl(const IntlDateTimeFormatImplKey& key, Ref<const IntlDateTimeFormatImpl>&& impl)
+    {
+        // An impl built while a language change is unprocessed may be based on the old
+        // language list; it must not outlive the clear that change will trigger.
+        if (hasLanguageChange()) [[unlikely]]
+            return;
+        m_cachedDateTimeFormatImpls.insert(key, RefPtr<const IntlDateTimeFormatImpl>(WTF::move(impl)));
```
kmiller68: Is there not a TOCTOU issue here? AFAICT, the langue change notification can happen on a background thread. Even if it always happened on the main thread this could be a problem for Workers/API users.

Source/JavaScriptCore/runtime/IntlCache.cpp
```diff
@@ -25,15 +26,38 @@
...
+
+void IntlCache::ensureLanguageChangeObserver()
+{
+    static std::once_flag onceKey;
+    std::call_once(onceKey, [] {
+        WTF::addLanguageChangeObserver(&s_languagesGeneration, [](void*) {
```
kmiller68: Nit: 
```suggestion
        addLanguageChangeObserver(&s_languagesGeneration, [](void*) {
```

kmiller68: Also, can you add a `RELEASE_ASSERT(isMainThread());` to `addLanguageChangeObserver`/`removeLanguageChangeObserver`?

Source/JavaScriptCore/runtime/IntlDateTimeFormat.cpp
```diff
@@ -738,6 +739,22 @@ void IntlDateTimeFormat::initializeDateTimeFormat(JSGlobalObject* globalObject,
     VM& vm = globalObject->vm();
     auto scope = DECLARE_THROW_SCOPE(vm);
 
+    std::optional<IntlDateTimeFormatImplKey> cacheKey;
+    const bool canCache = originalOptions.isUndefined() && (locales.isUndefined() || locales.isString()) && toLocaleStringTimeZone.isNull() && !globalObject->isHavingABadTime();
```
kmiller68: Why is `isHavingABadTime` relevant here?

Source/JavaScriptCore/runtime/IntlCache.h
```diff
@@ -25,29 +26,78 @@
...
+    }
+
+    void clearForTimeZoneChange() { clearDateTimeFormatImplCache(); }
+    bool hasLanguageChange() const { return m_lastSeenLanguagesGeneration != s_languagesGeneration.load(std::memory_order_relaxed); }
+    void clearForLanguageChange()
+    {
+        m_lastSeenLanguagesGeneration = s_languagesGeneration.load(std::memory_order_relaxed);
```
kmiller68: I think this needs to be acquire. Otherwise the following total scheduling is valid: 

// initializeDateTimeFormat: 
// findCachedDateTimeFormatImpl: 
if (hasLanguageChange()) // true
// m_lastSeenLanguagesGeneration = s_languagesGeneration.load(std::memory_order_relaxed); is reordered below
clearDateTimeFormatImplCache();
...
// create a new cache
// language changes again and s_languagesGeneration++
cacheDateTimeFormatImpl();
m_lastSeenLanguagesGeneration = s_languagesGeneration.load(std::memory_order_relaxed);

Source/JavaScriptCore/runtime/VM.cpp
```diff
@@ -1734,15 +1734,25 @@ NativeExecutable* VM::promiseAnySlowRejectFunctionExecutableSlow()
...
     }
 
-    if (dateCache.hasTimeZoneChange()) [[unlikely]]
+    if (dateCache.hasTimeZoneChange()) [[unlikely]] {
+        intlCache().clearForTimeZoneChange();
         dateCache.clearForTimeZoneChange();
+    }
+
+    if (intlCache().hasLanguageChange()) [[unlikely]]
+        intlCache().clearForLanguageChange();
```
kmiller68, line 1764: Why is any of this necessary if we check on access?

Source/WTF/wtf/Language.cpp
```diff
@@ -65,11 +67,13 @@ static ObserverMap& NODELETE observerMap()
 
 void addLanguageChangeObserver(void* context, LanguageChangeObserverFunction customObserver)
 {
+    RELEASE_ASSERT(isMainThread());
```
sosukesuzuki: This assertions causes following failure on EWS Linux:

```
ASSERTION FAILED: isMainThread()
../../../Source/WTF/wtf/Language.cpp(70) : void WTF::addLanguageChangeObserver(...)
4  JSC::IntlCache::ensureLanguageChangeObserver()
8  JSC::initializeWithOptionsCustomization(...)
9  JSC::initialize()
10 WebKit::InitializeWebKit2()
```

Source/JavaScriptCore/runtime/IntlCache.h
```diff
@@ -25,29 +26,73 @@
...
 #include <wtf/text/CString.h>
 #include <wtf/text/StringHash.h>
 #include <wtf/text/WTFString.h>
 #include <wtf/unicode/icu/ICUHelpers.h>
 
 namespace JSC {
 
+struct IntlDateTimeFormatImplKey {
+    // std::monostate means no locales argument; a null String cannot mark that because it compares equal to the empty string.
+    Variant<std::monostate, String> locales;
```
sosukesuzuki: `std::optional<String>` isn't enough?

sosukesuzuki (commented):
Overall looks good to me, with comments

Source/JavaScriptCore/runtime/IntlCache.cpp
justinmichaud: I added a lock instead, since it does appear to be called off the main thread in testAPI, and this code is infrequently ran. Now we can be extra sure nothing will break.

Source/JavaScriptCore/runtime/IntlRelativeTimeFormat.cpp
```diff
@@ -365,7 +366,7 @@ JSValue IntlRelativeTimeFormat::formatToParts(JSGlobalObject* globalObject, doub
     if (numberStart >= 0) {
         if (numberStart > 0) {
             JSObject* part = createIntlPartObject(globalObject, literalString, jsSubstring(vm, formattedRelativeTime, 0, numberStart));
-            parts->push(globalObject, part);
+            parts->putDirectIndex(globalObject, parts->length(), part);
```
kmiller68, line 369: Why are these changes here? I think these should be the same?

Source/JavaScriptCore/runtime/VMEntryScope.cpp
```diff
@@ -51,7 +51,7 @@ void VMEntryScope::setUpSlow()
 #endif
     }
 
-    if (m_vm.hasAnyEntryScopeServiceRequest() || m_vm.hasTimeZoneChange()) [[unlikely]]
+    if (m_vm.hasAnyEntryScopeServiceRequest() || m_vm.hasTimeZoneChange() || m_vm.hasLanguageChange()) [[unlikely]]
```
kmiller68: Nit: Can you add a FIXME to roll time zone and language change into the entry service requests? The various observers would then just set the request.

kmiller68 (approved):
r=me with questions/nits

Source/JavaScriptCore/runtime/IntlRelativeTimeFormat.cpp
justinmichaud: Since I encountered one bug of this kind, I fixed the nearby cases too. The included tests catch if we revert this, and this does appear to me to be the intended spec behaviour.

Source/JavaScriptCore/runtime/VMEntryScope.cpp
justinmichaud: This turned out to be such an easy fix that I just did it here.

aperezdc:
This is already part of the 2.54 branch. This landed as `317628@main`, which is earlier than the branch point at `317695@main`.

PR #64510, opened 2026-05-08
[Digital Credentials] Add WebDriver BiDi digitalCredentials module for automated testing

marcoscaceres:
#### fc22dff35d852f9e43f4d3f40924fd368b7a29e3
<pre>
[Digital Credentials] Add WebDriver BiDi digitalCredentials module for automated testing
<a href="https://rdar.apple.com/168941907">rdar://168941907</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=306292">https://bugs.webkit.org/show_bug.cgi?id=306292</a>

Reviewed by BJ Burg.

Add a WebDriver BiDi digitalCredentials module that allows automated testing
of the Digital Credentials API by simulating wallet behavior (respond, decline,
wait) without requiring a real credential provider.

The simulated wallet behavior is looked up for the requesting frame&apos;s browsing
context, falling back to the session&apos;s default behavior, matching the Digital
Credentials specification&apos;s &quot;handle virtual wallet behavior&quot; algorithm. To do
this, the requesting frame&apos;s identifier is threaded from the credential request
coordinator down to the UIProcess chooser entry point.

A &quot;wait&quot; behavior parks the chooser completion handler so the request stays
pending. The handler is now released when the request is dismissed on abort or
document teardown, not only when the session ends, so a waiting request cannot
leak past the document that made it. The completion callback also tolerates a
reply that arrives after the request was already aborted.

* Source/WebCore/Modules/identity/CredentialRequestCoordinator.cpp:
(WebCore::CredentialRequestCoordinator::initiateTheCredentialRequest):
(WebCore::CredentialRequestCoordinator::processCredentialChooserResponse):
* Source/WebCore/Modules/identity/CredentialRequestCoordinatorClient.h:
* Source/WebCore/Modules/identity/dummy/DummyCredentialRequestCoordinatorClient.cpp:
(WebCore::DummyCredentialRequestCoordinatorClient::showDigitalCredentialsChooser):
* Source/WebCore/Modules/identity/dummy/DummyCredentialRequestCoordinatorClient.h:
* Source/WebCore/loader/EmptyClients.cpp:
* Source/WebKit/DerivedSources-input.xcfilelist:
* Source/WebKit/DerivedSources.make:
* Source/WebKit/Sources.txt:
* Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.cpp: Added.
(WebKit::toPresentationProtocol):
(WebKit::BidiDigitalCredentialsAgent::BidiDigitalCredentialsAgent):
(WebKit::BidiDigitalCredentialsAgent::~BidiDigitalCredentialsAgent):
(WebKit::BidiDigitalCredentialsAgent::setVirtualWalletBehavior):
(WebKit::BidiDigitalCredentialsAgent::behaviorForContext const):
(WebKit::BidiDigitalCredentialsAgent::abortPendingHandler):
(WebKit::BidiDigitalCredentialsAgent::holdPendingHandler):
(WebKit::BidiDigitalCredentialsAgent::releasePendingHandler):
(WebKit::BidiDigitalCredentialsAgent::cancelAllPendingRequests):
* Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.h: Added.
(WebKit::WebDriverBidiProcessor::WebDriverBidiProcessor):
* Source/WebKit/UIProcess/Automation/WebDriverBidiProcessor.h:
* Source/WebKit/UIProcess/Automation/protocol/BidiDigitalCredentials.json: Added.
(WebKit::WebPageProxy::showDigitalCredentialsChooser):
(WebKit::WebPageProxy::dismissDigitalCredentialsChooser):
* Source/WebKit/WebProcess/DigitalCredentials/DigitalCredentialsCoordinator.cpp:
(WebKit::DigitalCredentialsCoordinator::showDigitalCredentialsChooser):
* Source/WebKit/WebProcess/DigitalCredentials/DigitalCredentialsCoordinator.h:
(WebKit::WebChromeClient::showDigitalCredentialsChooser):
(WebKit::WebPage::showDigitalCredentialsChooser):
* WebDriverTests/imported/w3c/tools/webdriver/webdriver/bidi/client.py:
(BidiSession.__init__):
* WebDriverTests/imported/w3c/tools/webdriver/webdriver/bidi/modules/__init__.py:
* WebDriverTests/imported/w3c/tools/webdriver/webdriver/bidi/modules/digital_credentials.py: Added.
(DigitalCredentials):
(DigitalCredentials.set_virtual_wallet_behavior):
* WebDriverTests/imported/w3c/webdriver/tests/bidi/external/digital_credentials/__init__.py: Added.
* WebDriverTests/imported/w3c/webdriver/tests/bidi/external/digital_credentials/set_virtual_wallet_behavior/__init__.py: Added.
* WebDriverTests/imported/w3c/webdriver/tests/bidi/external/digital_credentials/set_virtual_wallet_behavior/set_virtual_wallet_behavior.py: Added.
(test_set_virtual_wallet_behavior_respond):
(test_set_virtual_wallet_behavior_decline):
(test_set_virtual_wallet_behavior_wait):
(test_set_virtual_wallet_behavior_clear):
(test_set_virtual_wallet_behavior_invalid_respond_no_protocol):
(test_set_virtual_wallet_behavior_invalid_decline_with_response):
(test_set_virtual_wallet_behavior_respond_end_to_end):
(test_set_virtual_wallet_behavior_wait_is_aborted):

Canonical link: <a href="https://commits.webkit.org/316435@main">https://commits.webkit.org/316435@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/578e63dad21a045f453efae624b64ea908f473f8

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/172137 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/46054 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/39472 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/181580 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/129691 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/af37004e-348c-4d0e-8cf1-6523b5a10c31) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/174002 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/47341 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/45694 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/133896 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/94453 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/0d78fc76-8f59-485d-bf5b-9debfcf7141e) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/175095 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/37327 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/154432 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/114369 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/995f450a-a1c1-49d0-b027-bf18ac90919d) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/35958 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/34226 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/30190 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/146384 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/33944 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/184111 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/36725 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/38416 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/142646 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/45721 "Built successfully") | | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/142535 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/38520 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/46401 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/30791 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/111079 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/37614 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/118177 "Built successfully") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/44919 "Built successfully") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/8881 "") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/44319 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/44551 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/44378 "Built successfully") | | | |

### DIFF
--- a/Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.cpp
+++ b/Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.cpp
@@ -1,0 +1,140 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+ * THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "BidiDigitalCredentialsAgent.h"
+
+#if ENABLE(WEBDRIVER_BIDI)
+
+#include "WebAutomationSession.h"
+#include "WebAutomationSessionMacros.h"
+#include "WebDriverBidiProtocolObjects.h"
+#include <WebCore/DigitalCredentialsResponseData.h>
+#include <WebCore/ExceptionData.h>
+#include <wtf/TZoneMallocInlines.h>
+
+namespace WebKit {
+
+using namespace Inspector;
+
+WTF_MAKE_TZONE_ALLOCATED_IMPL(BidiDigitalCredentialsAgent);
+
+static WebCore::DigitalCredentialPresentationProtocol toPresentationProtocol(Inspector::Protocol::BidiDigitalCredentials::DigitalCredentialProtocol protocol)
+{
+    switch (protocol) {
+    case Inspector::Protocol::BidiDigitalCredentials::DigitalCredentialProtocol::OrgIsoMdoc:
+        return WebCore::DigitalCredentialPresentationProtocol::OrgIsoMdoc;
+    }
+    RELEASE_ASSERT_NOT_REACHED();
+}
+
+BidiDigitalCredentialsAgent::BidiDigitalCredentialsAgent(WebAutomationSession& session, BackendDispatcher& backendDispatcher)
+    : m_session(session)
+    , m_digitalCredentialsDomainDispatcher(BidiDigitalCredentialsBackendDispatcher::create(backendDispatcher, this))
+{
+}
+
+BidiDigitalCredentialsAgent::~BidiDigitalCredentialsAgent()
+{
+    cancelAllPendingRequests();
+}
+
+void BidiDigitalCredentialsAgent::setVirtualWalletBehavior(Inspector::Protocol::BidiDigitalCredentials::VirtualWalletAction action, const String& optionalContext, std::optional<Inspector::Protocol::BidiDigitalCredentials::DigitalCredentialProtocol>&& optionalProtocol, RefPtr<JSON::Object>&& optionalResponse, CommandCallback<void>&& callback)
+{
+    RefPtr session = m_session.get();
+    ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(!session, InternalError);
+
+    using VirtualWalletAction = Inspector::Protocol::BidiDigitalCredentials::VirtualWalletAction;
+
+    auto applyBehavior = [&](VirtualWalletBehavior&& behavior) {
+        if (!optionalContext.isEmpty())
+            m_browsingContextToWalletBehaviors.set(optionalContext, WTF::move(behavior));
+        else
+            m_defaultBehavior = WTF::move(behavior);
+    };
+
+    switch (action) {
+    case VirtualWalletAction::Respond: {
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_AND_DETAILS_IF(!optionalProtocol, InvalidParameter, "Action 'respond' requires a 'protocol' parameter."_s);
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_AND_DETAILS_IF(!optionalResponse, InvalidParameter, "Action 'respond' requires a 'response' parameter."_s);
+
+        VirtualWalletBehavior behavior { action, toPresentationProtocol(*optionalProtocol), optionalResponse->toJSONString() };
+        applyBehavior(WTF::move(behavior));
+        break;
+    }
+    case VirtualWalletAction::Decline:
+    case VirtualWalletAction::Wait:
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_AND_DETAILS_IF(optionalProtocol, InvalidParameter, "Only action 'respond' accepts a 'protocol' parameter."_s);
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_AND_DETAILS_IF(optionalResponse, InvalidParameter, "Only action 'respond' accepts a 'response' parameter."_s);
+        applyBehavior(VirtualWalletBehavior { action, WebCore::DigitalCredentialPresentationProtocol::OrgIsoMdoc, String() });
+        break;
+    case VirtualWalletAction::Clear:
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_AND_DETAILS_IF(optionalProtocol, InvalidParameter, "Only action 'respond' accepts a 'protocol' parameter."_s);
+        ASYNC_FAIL_WITH_PREDEFINED_ERROR_AND_DETAILS_IF(optionalResponse, InvalidParameter, "Only action 'respond' accepts a 'response' parameter."_s);
+        if (!optionalContext.isEmpty())
+            m_browsingContextToWalletBehaviors.remove(optionalContext);
+        else
+            m_defaultBehavior = std::nullopt;
+        break;
+    }
+
+    callback({ });
+}
+
+std::optional<VirtualWalletBehavior> BidiDigitalCredentialsAgent::behaviorForContext(const String& browsingContextID) const
+{
+    auto it = m_browsingContextToWalletBehaviors.find(browsingContextID);
+    if (it != m_browsingContextToWalletBehaviors.end())
+        return it->value;
+    return m_defaultBehavior;
+}
+
+void BidiDigitalCredentialsAgent::abortPendingHandler(const String& contextID, ASCIILiteral message)
+{
+    if (auto handler = m_browsingContextToPendingWaitHandlers.take(contextID))
+        handler(makeUnexpected(WebCore::ExceptionData { WebCore::ExceptionCode::AbortError, message }));
+}
+
+void BidiDigitalCredentialsAgent::holdPendingHandler(const String& contextID, DigitalCredentialsPickerCompletionHandler&& handler)
+{
+    abortPendingHandler(contextID, "Superseded by a new credential request."_s);
+    m_browsingContextToPendingWaitHandlers.set(contextID, WTF::move(handler));
+}
+
+void BidiDigitalCredentialsAgent::releasePendingHandler(const String& contextID)
+{
+    abortPendingHandler(contextID, "The credential request was aborted."_s);
+}
+
+void BidiDigitalCredentialsAgent::cancelAllPendingRequests()
+{
+    auto handlers = std::exchange(m_browsingContextToPendingWaitHandlers, { });
+    for (auto& handler : handlers.values())
+        handler(makeUnexpected(WebCore::ExceptionData { WebCore::ExceptionCode::AbortError, "Automation session ended."_s }));
+}
+
+} // namespace WebKit
+
+#endif // ENABLE(WEBDRIVER_BIDI)
--- a/Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.h
+++ b/Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.h
@@ -1,0 +1,83 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS''
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO,
+ * THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS
+ * BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+ * THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#pragma once
+
+#if ENABLE(WEBDRIVER_BIDI)
+
+#include "WebDriverBidiBackendDispatchers.h"
+#include "WebDriverBidiProtocolObjects.h"
+#include <WebCore/DigitalCredentialPresentationProtocol.h>
+#include <wtf/CompletionHandler.h>
+#include <wtf/Expected.h>
+#include <wtf/HashMap.h>
+#include <wtf/TZoneMalloc.h>
+#include <wtf/WeakPtr.h>
+#include <wtf/text/WTFString.h>
+
+namespace WebCore {
+struct DigitalCredentialsResponseData;
+struct ExceptionData;
+}
+
+namespace WebKit {
+
+class WebAutomationSession;
+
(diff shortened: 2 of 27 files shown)


Source/WebKit/UIProcess/Automation/protocol/BidiDigitalCredentials.json
```diff
@@ -0,0 +1,30 @@
...
+    ],
+    "commands": [
+        {
+            "name": "setVirtualWalletBehavior",
+            "description": "Sets the virtual wallet behavior for the session or a specific browsing context.",
+            "spec": "https://w3c-fedid.github.io/digital-credentials/#command-digitalCredentials-setVirtualWalletBehavior",
+            "wpt": "https://github.com/web-platform-tests/wpt/tree/master/webdriver/tests/bidi/external/digital_credentials",
+            "parameters": [
+                { "name": "action", "$ref": "BidiDigitalCredentials.VirtualWalletAction", "description": "The action to take when a credential request occurs." },
+                { "name": "context", "type": "string", "optional": true, "description": "The browsing context ID to scope this behavior to." },
```
burg: Nit: please crossreference the alias type in BidiBrowsingContext domain.

Source/WebKit/UIProcess/Automation/protocol/BidiDigitalCredentials.json
```diff
@@ -0,0 +1,30 @@
...
+                { "name": "protocol", "type": "string", "optional": true, "description": "The protocol string for the synthetic response (required for 'respond')." },
```
burg: Protocol string is uh.. overloaded in this context. Is this a protocol version? or a protocol specification?

Source/WebKit/UIProcess/Automation/protocol/BidiDigitalCredentials.json
```diff
@@ -0,0 +1,30 @@
...
+                { "name": "response", "type": "object", "optional": true, "description": "The response data for the synthetic credential (required for 'respond')." }
```
burg: This pattern of optional-with-required comment is  for type-dependent fields. The description shouldn't use a parenthetical to state the requirement, see other command examples.

(It seems common enough in WebDriver BiDi that i ought to make this part of the protocol generator..)

Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.cpp
```diff
@@ -0,0 +1,127 @@
...
+using namespace Inspector;
+
+WTF_MAKE_TZONE_ALLOCATED_IMPL(BidiDigitalCredentialsAgent);
+
+static std::optional<WebCore::DigitalCredentialPresentationProtocol> protocolFromString(const String& protocol)
+{
+    if (protocol == "org-iso-mdoc"_s)
+        return WebCore::DigitalCredentialPresentationProtocol::OrgIsoMdoc;
+    return std::nullopt;
+}
```
burg, line 51: Ah I see.. you'll want to create an enum type in the .json spec file that matches the DigitalCredentialPresentationProtocol values, then use a conversion method so that this is enum->enum rather than becoming untyped.

Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.cpp
```diff
@@ -0,0 +1,127 @@
...
+{
+    cancelAllPendingRequests();
+}
+
+void BidiDigitalCredentialsAgent::setVirtualWalletBehavior(Inspector::Protocol::BidiDigitalCredentials::VirtualWalletAction action, const String& optionalContext, const String& optionalProtocol, RefPtr<JSON::Object>&& optionalResponse, CommandCallback<void>&& callback)
+{
+    RefPtr session = m_session.get();
+    ASYNC_FAIL_WITH_PREDEFINED_ERROR_IF(!session, InternalError);
+
+    using Action = Inspector::Protocol::BidiDigitalCredentials::VirtualWalletAction;
```
burg: Hmm, can we keep the most inner part of the qualified name? Like 'VirtualWalletAction'? Otherwise it becomes harder to grep for uses..

Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    if (action == Action::Respond) {
```
burg: This would be a lot easier to follow / verify if it switched on the Action value. There might be a few redundant checks but it will be so much easier to audit against the spec.

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -147,6 +147,11 @@
 #include "ViewWindowCoordinates.h"
 #include "WKContextPrivate.h"
 #include "WebAutomationSession.h"
+#if ENABLE(WEBDRIVER_BIDI)
+#include "BidiDigitalCredentialsAgent.h"
+#include "WebDriverBidiProcessor.h"
+#include "WebDriverBidiProtocolObjects.h"
+#endif
```
burg: Sorting seems off here. This is probably down below?

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -10534,6 +10539,32 @@ void WebPageProxy::showDigitalCredentialsPicker(IPC::Connection& connection, con
                 completionHandler(makeUnexpected(WebCore::ExceptionData { WebCore::ExceptionCode::SecurityError, "Digital credentials feature is disabled by preference."_s }))
             );
 
+#if ENABLE(WEBDRIVER_BIDI)
+            if (isControlledByAutomation()) {
+                if (RefPtr automationSession = configuration().processPool().automationSession()) {
+                    auto& agent = automationSession->bidiProcessor().digitalCredentialsAgent();
+                    const auto contextID = automationSession->handleForWebPageProxy(*this);
+                    const auto behavior = agent.behaviorForContext(contextID);
```
burg: Can `behavior` become `walletBehavior` either in the local variable name or the method name, so that this line is somewhat understandable without all the other lines?

Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.h
```diff
@@ -0,0 +1,80 @@
...
+}
+
+namespace WebKit {
+
+class WebAutomationSession;
+
+struct VirtualWalletBehavior {
+    Inspector::Protocol::BidiDigitalCredentials::VirtualWalletAction action;
+    WebCore::DigitalCredentialPresentationProtocol protocol { WebCore::DigitalCredentialPresentationProtocol::OrgIsoMdoc };
+    String responseJSON;
```
burg, line 52: Should this be validated/parsed with `JSONValue`?

Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.h
```diff
@@ -0,0 +1,80 @@
...
+
+    void holdPendingHandler(const String& contextID, DigitalCredentialsPickerCompletionHandler&&);
+    void cancelAllPendingRequests();
+
+private:
+    WeakPtr<WebAutomationSession> m_session;
+    Ref<Inspector::BidiDigitalCredentialsBackendDispatcher> m_digitalCredentialsDomainDispatcher;
+    HashMap<String, VirtualWalletBehavior> m_contextBehaviors;
+    std::optional<VirtualWalletBehavior> m_defaultBehavior;
+    HashMap<String, DigitalCredentialsPickerCompletionHandler> m_pendingWaitHandlers;
```
burg: Nit: `browsingContextToPendingWaitHandlers`

Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.h
```diff
@@ -0,0 +1,80 @@
...
+
+    std::optional<VirtualWalletBehavior> behaviorForContext(const String& browsingContextID) const;
+
+    void holdPendingHandler(const String& contextID, DigitalCredentialsPickerCompletionHandler&&);
+    void cancelAllPendingRequests();
+
+private:
+    WeakPtr<WebAutomationSession> m_session;
+    Ref<Inspector::BidiDigitalCredentialsBackendDispatcher> m_digitalCredentialsDomainDispatcher;
+    HashMap<String, VirtualWalletBehavior> m_contextBehaviors;
```
burg: Nit: `m_browsingContextToWalletBehaviors`

Source/WebKit/UIProcess/WebPageProxy.cpp
burg: Design question: WebDriver has User Contexts (~profiles / WKWebsiteDataStore)  and Browsing Contexts (~browsing contexts). should wallet behavior be per-browsing context, per toplevel browsing context, per user context, or per session?

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -10534,6 +10539,32 @@ void WebPageProxy::showDigitalCredentialsPicker(IPC::Connection& connection, con
...
+                    if (behavior) {
+                        using Action = Inspector::Protocol::BidiDigitalCredentials::VirtualWalletAction;
```
burg: ditto other comment about `using Action`

burg (commented):
This is in pretty good shape!

Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.cpp
marcoscaceres: Done above.

Source/WebKit/UIProcess/Automation/BidiDigitalCredentialsAgent.h
marcoscaceres: It arrives already parsed/validated as a JSON::Object from the dispatcher, so there's no unvalidated input here. So, it should be ok to just stay as String.

Source/WebKit/UIProcess/Automation/protocol/BidiDigitalCredentials.json
marcoscaceres: Yeah, I see what you mean... it's an identifier. "protocol identifier" https://www.w3.org/TR/digital-credentials/#dfn-protocol-identifier

marcoscaceres: hopefully the description makes more sense now.

Source/WebKit/UIProcess/WebPageProxy.cpp
marcoscaceres: On scoping: per the spec's "handle virtual wallet behavior" algorithm (§13.2) it's per browsing context with a per-session default. `setVirtualWalletBehavior` stores per the optional `context` (a browsing-context id), or as the session default when omitted, and lookup uses the requesting global's browsing context, else the default. The lookup was previously keyed off the top-level page handle, so a per-iframe `context` override never matched; fixed here by threading the requesting frame's `FrameIdentifier` to `showDigitalCredentialsChooser` and resolving via `effectiveHandleForWebFrameProxy` (same id space `setVirtualWalletBehavior` stores under).

WebDriverTests/imported/w3c/webdriver/tests/bidi/external/digital_credentials/set_virtual_wallet_behavior/set_virtual_wallet_behavior.py
```diff
@@ -0,0 +1,75 @@
...
+    )
+
+    await bidi_session.digital_credentials.set_virtual_wallet_behavior(
+        action="respond",
+        protocol="org-iso-mdoc",
+        response={"documents": [{"docType": "org.iso.18013.5.1.mDL"}]},
+    )
+
+    # Verify the command was accepted (no error thrown)
+    # Full end-to-end credential flow test requires page JS execution
```
burg: Not exactly sure what prevents this. bidi_session has a way to eval JS.

WebDriverTests/imported/w3c/webdriver/tests/bidi/external/digital_credentials/set_virtual_wallet_behavior/set_virtual_wallet_behavior.py
```diff
@@ -0,0 +1,75 @@
...
+    test_url = url("/common/blank.html", protocol="https")
+    await bidi_session.browsing_context.navigate(
+        context=new_tab["context"],
+        url=test_url,
+        wait="complete",
+    )
+
+    await bidi_session.digital_credentials.set_virtual_wallet_behavior(
+        action="wait",
+    )
```
burg, line 70: Yeah uh this is going to wait forever right?

burg (changes requested):
It isn't clear to me that a `wait` action will ever recover as written. Should `m_browsingContextToPendingWaitHandlers` be emptied out at times other than the session ending?

WebDriverTests/imported/w3c/webdriver/tests/bidi/external/digital_credentials/set_virtual_wallet_behavior/set_virtual_wallet_behavior.py
marcoscaceres: Correct. Wait forever.

WebDriverTests/imported/w3c/webdriver/tests/bidi/external/digital_credentials/set_virtual_wallet_behavior/set_virtual_wallet_behavior.py
marcoscaceres: Added an end-to-end test that drives `navigator.credentials.get()` through `bidi_session.script.evaluate` and asserts the returned credential, with a feature-detect skip where the DC API isn't compiled in.

WebDriverTests/imported/w3c/webdriver/tests/bidi/external/digital_credentials/set_virtual_wallet_behavior/set_virtual_wallet_behavior.py
marcoscaceres: It still waits forever on its own, but an aborted signal or navigation now rejects it with `AbortError`.

marcoscaceres:
The parked `wait` handler is now released on abort or document teardown via a new `releasePendingHandler` (called from `dismissDigitalCredentialsChooser`), not just at session end, and I reordered the spec's "stops being fully active" abort so it arms for `wait` too.

burg (approved):
Consider hooking up contextDestroyed / similar hooks for page and frame so that test state gets removed rather than leaked until session ends.

PR #64512, opened 2026-05-08
AX: AXTextMarker::findLine can hang when a replaced element (e.g. an image) shares a containing block with normal inline content

twilco:
#### 803f725dbc584e54cf6b6c81a2e7de7666a622b1
<pre>
AX: AXTextMarker::findLine can hang when a replaced element (e.g. an image) shares a containing block with normal inline content
<a href="https://bugs.webkit.org/show_bug.cgi?id=314359">https://bugs.webkit.org/show_bug.cgi?id=314359</a>
<a href="https://rdar.apple.com/176509656">rdar://176509656</a>

Reviewed by Dominic Mazzoni.

AccessibilityRenderObject::textRuns() hardcoded lineIndex = 0 for every replaced
element (image, SVG, video, etc.). When such an element was a DOM sibling of a
soft-wrapping inline text in the same containing block, the hardcoded 0 produced
an AXTextRunLineID identical to the text&apos;s first wrap line. AXTextMarker::findLine
then returned its input marker on every call, and AXTextMarker::lineIndex spun forever
asking nextLineEnd() to advance, hanging the AX thread.

This commit fixes the issue in a few ways:

1. For in-flow inline replaced elements, compute the actual lineIndex via
 InlineIterator::boxFor(), matching the existing pattern in the RenderLineBreak
 branch directly above. The line layout already knows which line it placed the
 element on, we just needed to ask.

2. For out-of-flow replaced elements (float / position:absolute), no line box
 exists. Use the renderer pointer in the lineID&apos;s containing-block slot so the
 resulting (renderer*, 0) lineID is unique to that element and can never
 collide with an in-flow text run&apos;s lineID. Apply the same hardening to the
 RenderLineBreak no-box fallback for symmetry.

3. AXTextRuns(): change the constructor parameter to void*. The field has
 always been void*, the constructor was the only inconsistency. Removes the
 need for a reinterpret_cast at the new call sites that pass a non-RenderBlock
 pointer for the unique-per-element lineID case.

4. AXTextMarker::lineIndex(): add a non-progress guard so this class of bug
 degrades to &quot;wrong number&quot; instead of hanging the AX thread.

* LayoutTests/accessibility-isolated-tree/TestExpectations:
* LayoutTests/accessibility/mac/replaced-element-line-index-hang-expected.txt: Added.
* LayoutTests/accessibility/mac/replaced-element-line-index-hang.html: Added.
* Source/WebCore/accessibility/AXTextMarker.cpp:
(WebCore::AXTextMarker::lineIndex const):
* Source/WebCore/accessibility/AXTextRun.h:
(WebCore::AXTextRuns::AXTextRuns):
(WebCore::AccessibilityRenderObject::textRuns):

Canonical link: <a href="https://commits.webkit.org/312968@main">https://commits.webkit.org/312968@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/1a526f474f77ee707a144f62aa293bffc74f4bb6

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/161293 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/34766 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/27867 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/170196 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/117664 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/163162 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/34867 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/34767 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/125114 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/88176 "1 flakes 1 failures") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/164252 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/27399 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/144878 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/105711 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/26447 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/24957 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/17916 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/136141 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/22638 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/172679 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/19700 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/24279 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/133396 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/34440 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/29050 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/133404 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36125 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/34425 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/144433 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/96290 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/27946 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/21251 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/33935 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/101360 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/33434 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/33678 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/33583 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/accessibility/AXTextRun.h
+++ b/Source/WebCore/accessibility/AXTextRun.h
@@ -126,7 +126,7 @@ public:
     bool containsOnlyASCII { true };
 
     AXTextRuns() = default;
-    AXTextRuns(RenderBlock* containingBlock, Vector<AXTextRun>&& textRuns, String&& text, bool containsOnlyASCII = true)
+    AXTextRuns(void* containingBlock, Vector<AXTextRun>&& textRuns, String&& text, bool containsOnlyASCII = true)
         : text(WTF::move(text))
         , containingBlock(containingBlock)
         , runs(WTF::move(textRuns))
--- a/Source/WebCore/accessibility/AccessibilityRenderObject.cpp
+++ b/Source/WebCore/accessibility/AccessibilityRenderObject.cpp
@@ -1451,9 +1451,12 @@ AXTextRuns AccessibilityRenderObject::textRuns()
     CheckedPtr renderer = this->renderer();
     if (CheckedPtr renderLineBreak = dynamicDowncast<RenderLineBreak>(renderer.get())) {
         auto box = InlineIterator::boxFor(*renderLineBreak);
-
+        // If we have a real line box, use its containing block + lineIndex. Otherwise (no box),
+        // use the renderer pointer in the containing-block slot so the resulting (renderer*, 0)
+        // lineID is unique and can't collide with in-flow content. Mirrors the replaced-element
+        // branch below.
         return AXTextRuns(
-            renderLineBreak->containingBlock(),
+            box ? static_cast<void*>(renderLineBreak->containingBlock()) : renderLineBreak.get(),
             { AXTextRun(box ? box->lineIndex() : 0, /* startIndex */ 0, /* endIndex */ 1, { lengthOneDomOffsets }, { 0 }, 0, 0) },
             makeString('\n').isolatedCopy()
         );
@@ -1466,15 +1469,29 @@ AXTextRuns AccessibilityRenderObject::textRuns()
 
     if (isReplacedElement()) {
         CheckedPtr containingBlock = renderer ? renderer->containingBlock() : nullptr;
-        FloatRect rect = localRect();
-        uint16_t width = static_cast<uint16_t>(rect.width());
-        uint16_t height = static_cast<uint16_t>(rect.height());
         if (!containingBlock)
             return { };
 
+        FloatRect rect = localRect();
+        uint16_t width = static_cast<uint16_t>(rect.width());
+        uint16_t height = static_cast<uint16_t>(rect.height());
+
+        size_t lineIndex = 0;
+        void* lineIDContainingBlock = containingBlock.get();
+        if (CheckedPtr renderBox = dynamicDowncast<RenderBox>(renderer.get())) {
+            if (renderBox->isFloatingOrOutOfFlowPositioned()) {
+                // Out-of-flow (float / position:absolute) replaced elements don't sit on a
+                // line box. Use the renderer pointer in the lineID's containing-block slot
+                // so the resulting (renderer*, 0) lineID is unique to this element and can't
+                // collide with any in-flow text run's real (containingBlock, 0) lineID.
+                lineIDContainingBlock = renderer.get();
+            } else if (auto box = InlineIterator::boxFor(*renderBox))
+                lineIndex = box->lineIndex();
+        }
+
         return AXTextRuns(
-            containingBlock.get(),
-            { AXTextRun(0, /* startIndex */ 0, /* endIndex */ 1, { lengthOneDomOffsets }, { width }, height, 0) },
+            lineIDContainingBlock,
+            { AXTextRun(lineIndex, /* startIndex */ 0, /* endIndex */ 1, { lengthOneDomOffsets }, { width }, height, 0) },
             String(span(objectReplacementCharacter))
         );
     }
--- a/LayoutTests/accessibility-isolated-tree/TestExpectations
+++ b/LayoutTests/accessibility-isolated-tree/TestExpectations
@@ -68,6 +68,9 @@ accessibility/mac/text-marker-word-nav.html [ Timeout ]
 accessibility/mac/text-operation/text-operation-replace-across-multiple-fields.html [ Failure ]
 accessibility/textarea-insertion-point-line-number.html [ Failure ]
 
+# We compute the wrong line index relative to what is actually visually laid out.
+accessibility/mac/replaced-element-line-index-hang.html [ Failure ]
+
 accessibility/mac/html5-input-number.html [ Timeout ]
 
 # Fails because we don't get paint calls for SVG shapes and thus can't cache a path for them.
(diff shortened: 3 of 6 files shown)


Source/WebCore/accessibility/AXTextRun.h
minorninth, line 129: I don't get why this is void* now

Source/WebCore/accessibility/AccessibilityRenderObject.cpp
minorninth, line 1459: Could we use RenderObject* as a base class here?

Source/WebCore/accessibility/AXTextRun.h
twilco: Because the pointer is never supposed to be dereferenced anyways, so void* is a stronger signal of that (though poorer documentation than RenderBlock* was). A line ID in our text system is two things: a containing block pointer, and a line index. The containing block pointer is only for comparing equality, i.e. are these two pieces of text on the same line?

The reason for it changing now is that we now have a "containing block" that is actually a RenderImage*, not a RenderBlock*. As the reason this bug was caused is because we got the containing block for a RenderImage* to use for its line ID, but hardcoded line index of zero because images do not always participate in actual line layout, but we want to represent them in the text system, e.g. because images are a "stop" in character-by-character navigation.

Because we hardcoded the line index to zero, it collided with text that was _actually_ laid out on line index 0 of the shared containing block (between the text and the image). This caused the infinite loop.

So for images that do not participate in line layout, making the RenderImage* its own "containing block" guarantees it is treated as a unique line, as it should be.

All that to say: I think we could type is as RenderObject*. But it is genuinely not safe to dereference these raw pointers, so I chose void* to decrease the chances of that happening since the type needed to change anyways. If you'd prefer RenderObject*, I'm fine with that too.

minorninth: Thanks for the explanation!

It looks like WebKit has ObjectIdentifier<T>, that would be a very strong signal that you want to identify the RenderObject but not dereference it.

If it were me I'd probably use RenderObject* and name the variable opaque* or something like that. I'd be fine with void* and a name like opaque with a brief comment somewhere.

I think it depends in part on whether you think it'd be dangerous or unsafe to dereference the RenderObject, or just that you don't think it should be necessary in order to compute the correct result.

minorninth (approved):

Source/WebCore/accessibility/AXTextRun.h
twilco: Indeed that was originally my thought as well, but sadly RenderObject doesn't expose a method to get an ObjectIndentifier.

It would be unsafe to dereference the pointer since it's raw, so my hope was void* was extra encouragement not to do that :)

There is a comment explaining this choice (in fact `AXTextRunLineID` has always stored a void*!):

```
struct AXTextRunLineID {
    // Do not dereference, for comparison to other AXTextRunLineIDs only.
    void* containingBlock { nullptr };
    size_t lineIndex { 0 };
    ...
};
```

PR #64560, opened 2026-05-08
[GFC][Crash] Grid items with auto height triggering a crash in blockContentSizeSuggestion()

yulunwu2:
#### e9a9f1b34ebcfffc202353153c28a419fa8f63c6
<pre>
[GFC][Crash] Grid items with auto height triggering a crash in blockContentSizeSuggestion()
<a href="https://bugs.webkit.org/show_bug.cgi?id=314371">https://bugs.webkit.org/show_bug.cgi?id=314371</a>
&lt;<a href="https://rdar.apple.com/176527830">rdar://176527830</a>&gt;

Reviewed by Sammy Gill.

Grid items with auto height are hitting ASSERT_NOT_IMPLEMENTED_YET
in blockSpecifiedSizeSuggestion(). This assert is not needed.
blockSpecifiedSizeSuggestion() and inlineSpecifiedSizeSuggestion()
are implemented per spec as indefinite preferred size is undefined.

This PR explicitly allows preferredSize() auto and updates the
documentation explaining why this is the case.

Fixing that exposed another crash in blockContentSizeSuggestion() which
called integrationUtils::minContentHeight().

integrationUtils::minContentHeight() is problematic here because it is fetching
a cached intrinsic content logical height that was saved by flex layout.

This isn&apos;t applicable for RenderBox() as finding the height for a RenderBox()
requires a width constraint to resolve. We should instead run layout() on the
grid item with the inlineAxisConstraint to resolve the block size for the grid item.

This PR introduces an overloaded minContentHeight() method that takes
inline constraints and supports grid items.

* LayoutTests/fast/grid/grid-item-block-content-size-suggestion-crash-expected.txt: Added.
* LayoutTests/fast/grid/grid-item-block-content-size-suggestion-crash.html: Added.
* Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp:
(WebCore::Layout::GridLayout::layoutGridItems const):
* Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.cpp:
(WebCore::Layout::GridLayoutUtils::inlineSpecifiedSizeSuggestion):
(WebCore::Layout::GridLayoutUtils::blockSpecifiedSizeSuggestion):
(WebCore::Layout::GridLayoutUtils::blockContentSizeSuggestion):
(WebCore::Layout::GridLayoutUtils::automaticMinimumBlockSize):
(WebCore::Layout::GridLayoutUtils::usedBlockSizeForGridItem):
(WebCore::Layout::GridLayoutUtils::usedBlockMinimumSize):
(WebCore::Layout::GridLayoutUtils::inlineAxisGridItemSizingFunctions):
(WebCore::Layout::GridLayoutUtils::blockAxisGridItemSizingFunctions):
* Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.h:
* Source/WebCore/layout/formattingContexts/grid/TrackSizingAlgorithm.cpp:
(WebCore::Layout::minimumContributions):
* Source/WebCore/layout/formattingContexts/grid/TrackSizingAlgorithm.h:
(WebCore::Layout::GridItemSizingFunctions::GridItemSizingFunctions):
* Source/WebCore/layout/integration/LayoutIntegrationUtils.cpp:
(WebCore::Layout::IntegrationUtils::minContentHeight const):
* Source/WebCore/layout/integration/LayoutIntegrationUtils.h:

Canonical link: <a href="https://commits.webkit.org/313478@main">https://commits.webkit.org/313478@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/5afb326693902d5179fa1419701d6d8adec5d7a5

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/163135 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/36616 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/29833 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/172074 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/119582 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/056c9a55-679a-42bb-81b2-d23694083242) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/165004 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/36920 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/36616 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/126657 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/89259 "3 flakes 1 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/8122a1d3-0cdb-468a-b5d7-90bce10b68a5) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/166094 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/28920 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/146655 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/107227 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/df83041d-8498-4664-bc8b-eddc38a2da48) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/27966 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/26603 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/19774 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/137859 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/24379 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/174577 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/26734 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/26034 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/134967 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/36286 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/30707 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/134959 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36405 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/37072 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/146174 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/98922 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/30261 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/23018 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/35782 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/108143 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/35281 "Built successfully") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/1041 "") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/35528 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/35428 "Built successfully") | | | |

### DIFF
--- a/LayoutTests/fast/grid/grid-item-block-content-size-suggestion-crash.html
+++ b/LayoutTests/fast/grid/grid-item-block-content-size-suggestion-crash.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<style>
+.grid {
+    display: grid;
+    width: 200px;
+    grid-template-columns: 1fr;
+    grid-template-rows: auto;
+}
+</style>
+<div class="grid">
+    <div style="grid-row: 1 / 2">PASS if no crash.</div>
+</div>
+<script>
+    window.testRunner?.dumpAsText();
+    document.body.offsetHeight;
+</script>
--- a/Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
+++ b/Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
@@ -545,7 +545,8 @@ std::pair<UsedInlineSizes, UsedBlockSizes> GridLayout::layoutGridItems(const Pla
         auto usedInlineSizeForGridItem = GridLayoutUtils::usedInlineSizeForGridItem(gridItem, gridItem.usedInlineBorderAndPadding(), columnTrackSizingFunctions, gridAreaInlineSize, integrationUtils);
         usedInlineSizes.append(usedInlineSizeForGridItem);
 
-        auto usedBlockSizeForGridItem = GridLayoutUtils::usedBlockSizeForGridItem(gridItem, gridItem.usedBlockBorderAndPadding(), rowTrackSizingFunctions, gridAreaBlockSize, integrationUtils);
+        // FIXME: investigate to check if we should use the usedInlineSize or the size of the grid are in the inline direction.
+        auto usedBlockSizeForGridItem = GridLayoutUtils::usedBlockSizeForGridItem(gridItem, gridItem.usedBlockBorderAndPadding(), rowTrackSizingFunctions, gridAreaBlockSize, formattingContext, usedInlineSizeForGridItem);
         usedBlockSizes.append(usedBlockSizeForGridItem);
 
         auto& layoutBox = gridItem.layoutBox();
--- a/Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.cpp
+++ b/Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.cpp
@@ -64,11 +64,15 @@ static bool NODELETE spansFlexMaxTrackSizingFunction(WTF::Range<size_t> spannedT
     return false;
 }
 
+// https://www.w3.org/TR/css-grid-2/#specified-size-suggestion
+// If the item's preferred size in the relevant axis is definite, then the specified size suggestion is that size. It is otherwise undefined.
 static std::optional<LayoutUnit> inlineSpecifiedSizeSuggestion(const PlacedGridItem& gridItem, LayoutUnit borderAndPadding, LayoutUnit containingBlockSize)
 {
     auto& preferredSize = gridItem.inlineAxisSizes().preferredSize;
     if (preferredSize.isFixed() || preferredSize.isPercent() || preferredSize.isCalculated())
         return Style::evaluate<LayoutUnit>(preferredSize, containingBlockSize, gridItem.usedZoom()) + borderAndPadding;
+    if (preferredSize.isAuto())
+        return { };
     ASSERT_NOT_IMPLEMENTED_YET();
     return { };
 }
@@ -85,11 +89,15 @@ static LayoutUnit inlineContentSizeSuggestion(const PlacedGridItem& gridItem, co
     return integrationUtils.minContentWidth(gridItem.layoutBox());
 }
 
+// https://www.w3.org/TR/css-grid-2/#specified-size-suggestion
+// If the item's preferred size in the relevant axis is definite, then the specified size suggestion is that size. It is otherwise undefined.
 static std::optional<LayoutUnit> blockSpecifiedSizeSuggestion(const PlacedGridItem& gridItem, LayoutUnit borderAndPadding, LayoutUnit containingBlockSize)
 {
     auto& preferredSize = gridItem.blockAxisSizes().preferredSize;
     if (preferredSize.isFixed() || preferredSize.isPercent() || preferredSize.isCalculated())
         return Style::evaluate<LayoutUnit>(preferredSize, containingBlockSize, gridItem.usedZoom()) + borderAndPadding;
+    if (preferredSize.isAuto())
+        return { };
     ASSERT_NOT_IMPLEMENTED_YET();
     return { };
 }
@@ -100,10 +108,16 @@ static std::optional<LayoutUnit> NODELETE blockTransferredSizeSuggestion(const P
     return { };
 }
 
-static LayoutUnit blockContentSizeSuggestion(const PlacedGridItem& gridItem, const IntegrationUtils& integrationUtils)
+// https://drafts.csswg.org/css-grid-1/#min-size-auto
+// The content size suggestion is the min-content size in the relevant axis, clamped, if it has a preferred aspect ratio,
+// by any definite opposite-axis minimum and maximum sizes converted through the aspect ratio.
+// https://drafts.csswg.org/css-sizing-3/#sizing-values
+// For a box’s block size, unless otherwise specified, this [min-content] is equivalent to its automatic size.
+static LayoutUnit blockContentSizeSuggestion(const PlacedGridItem& gridItem, LayoutUnit inlineAxisConstraint, const GridFormattingContext& formattingContext)
 {
+    // FIXME: Clamp by opposite-axis min/max sizes converted through the aspect ratio.
     ASSERT(!gridItem.preferredAspectRatio(), "Grid items with preferred aspect ratio not supported yet.");
-    return integrationUtils.minContentHeight(gridItem.layoutBox());
+    return formattingContext.integrationUtils().minContentHeight(gridItem.layoutBox(), inlineAxisConstraint);
 }
 
 // https://drafts.csswg.org/css-overflow-3/#overflow-properties
@@ -225,7 +239,7 @@ static LayoutUnit automaticMinimumInlineSize(const PlacedGridItem& gridItem, Lay
 
 // https://drafts.csswg.org/css-grid-1/#min-size-auto
 static LayoutUnit automaticMinimumBlockSize(const PlacedGridItem& gridItem, LayoutUnit borderAndPadding, const TrackSizingFunctionsList& trackSizingFunctions,
-    LayoutUnit containingBlockSize, const IntegrationUtils& integrationUtils)
+    LayoutUnit containingBlockSize, const GridFormattingContext& formattingContext, LayoutUnit inlineAxisConstraint)
 {
     auto& blockAxisSizes = gridItem.blockAxisSizes();
     ASSERT(blockAxisSizes.minimumSize.isAuto());
@@ -262,7 +276,7 @@ static LayoutUnit automaticMinimumBlockSize(const PlacedGridItem& gridItem, Layo
                 return *transferredSizeSuggestion;
         }
         // else its content size suggestion
-        return blockContentSizeSuggestion(gridItem, integrationUtils);
+        return blockContentSizeSuggestion(gridItem, inlineAxisConstraint, formattingContext);
     };
 
     // In all cases, the size suggestion is additionally clamped by the maximum size in
@@ -274,7 +288,7 @@ static LayoutUnit automaticMinimumBlockSize(const PlacedGridItem& gridItem, Layo
 }
 
 LayoutUnit usedBlockSizeForGridItem(const PlacedGridItem& placedGridItem, LayoutUnit borderAndPadding,
-    const TrackSizingFunctionsList& trackSizingFunctions, LayoutUnit rowsSize, const IntegrationUtils& integrationUtils)
+    const TrackSizingFunctionsList& trackSizingFunctions, LayoutUnit rowsSize, const GridFormattingContext& formattingContext, LayoutUnit inlineAxisConstraint)
 {
     auto& blockAxisSizes = placedGridItem.blockAxisSizes();
     ASSERT(blockAxisSizes.maximumSize.isFixed() || blockAxisSizes.maximumSize.isNone());
@@ -301,7 +315,7 @@ LayoutUnit usedBlockSizeForGridItem(const PlacedGridItem& placedGridItem, Layout
             && !marginStart.isAuto() && !marginEnd.isAuto()) {
             auto& usedZoom = placedGridItem.usedZoom();
 
-            auto minimumSize = GridLayoutUtils::usedBlockMinimumSize(placedGridItem, trackSizingFunctions, borderAndPadding, rowsSize, integrationUtils);
+            auto minimumSize = GridLayoutUtils::usedBlockMinimumSize(placedGridItem, trackSizingFunctions, borderAndPadding, rowsSize, formattingContext, inlineAxisConstraint);
             auto maximumSize = [&blockAxisSizes, &usedZoom] {
                 auto& computedMaximumSize = blockAxisSizes.maximumSize;
                 if (computedMaximumSize.isNone())
@@ -344,7 +358,7 @@ LayoutUnit usedInlineMinimumSize(const PlacedGridItem& gridItem, const TrackSizi
 }
 
 LayoutUnit usedBlockMinimumSize(const PlacedGridItem& gridItem, const TrackSizingFunctionsList& trackSizingFunctions,
-    LayoutUnit borderAndPadding, LayoutUnit rowsSize, const IntegrationUtils& integrationUtils)
+    LayoutUnit borderAndPadding, LayoutUnit rowsSize, const GridFormattingContext& formattingContext, LayoutUnit inlineAxisConstraint)
 {
     auto& minimumSize = gridItem.blockAxisSizes().minimumSize;
     return WTF::switchOn(minimumSize,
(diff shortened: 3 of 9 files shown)


LayoutTests/fast/grid/grid-item-block-content-size-suggestion-crash.html
sammygill, line 1: I think you could name this grid-item-block-content-size-suggestion-crash.html and then I don't think you would need the -expected file or the `dumpAsText` in the test

Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.cpp
```diff
@@ -100,10 +102,11 @@ static std::optional<LayoutUnit> NODELETE blockTransferredSizeSuggestion(const P
     return { };
 }
 
-static LayoutUnit blockContentSizeSuggestion(const PlacedGridItem& gridItem, const IntegrationUtils& integrationUtils)
+static LayoutUnit blockContentSizeSuggestion(const PlacedGridItem& gridItem, const GridFormattingContext& formattingContext, LayoutUnit inlineAxisConstraint)
 {
     ASSERT(!gridItem.hasPreferredAspectRatio(), "Grid items with preferred aspect ratio not supported yet.");
-    return integrationUtils.minContentHeight(gridItem.layoutBox());
+    formattingContext.integrationUtils().layoutWithFormattingContextForBox(gridItem.layoutBox(), inlineAxisConstraint);
+    return formattingContext.geometryForGridItem(gridItem.layoutBox()).borderBoxHeight();
```
sammygill: Why this change? the The spec says:

The [content size suggestion](https://drafts.csswg.org/css-grid-1/#content-size-suggestion) is the [min-content size](https://www.w3.org/TR/css-sizing-3/#min-content) in the relevant axis, clamped, if it has a [preferred aspect ratio](https://www.w3.org/TR/css-sizing-4/#preferred-aspect-ratio), by any [definite](https://www.w3.org/TR/css-sizing-3/#definite) opposite-axis [minimum](https://www.w3.org/TR/css-sizing-3/#min-width) and [maximum sizes](https://www.w3.org/TR/css-sizing-3/#max-width) converted through the aspect ratio.

Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.cpp
```diff
@@ -64,12 +64,13 @@ static bool NODELETE spansFlexMaxTrackSizingFunction(WTF::Range<size_t> spannedT
...
 }
 
+// https://www.w3.org/TR/css-grid-2/#specified-size-suggestion
+// If the item's preferred size in the relevant axis is definite, then the specified size suggestion is that size. It is otherwise undefined.
 static std::optional<LayoutUnit> inlineSpecifiedSizeSuggestion(const PlacedGridItem& gridItem, LayoutUnit borderAndPadding, LayoutUnit containingBlockSize)
 {
     auto& preferredSize = gridItem.inlineAxisSizes().preferredSize;
     if (preferredSize.isFixed() || preferredSize.isPercent() || preferredSize.isCalculated())
         return Style::evaluate<LayoutUnit>(preferredSize, containingBlockSize, gridItem.usedZoom()) + borderAndPadding;
-    ASSERT_NOT_IMPLEMENTED_YET();
```
sammygill, line 72: I presume we are hitting this because the `preferredSize` is a specific type that is allowed through the eligibility logic. Can we instead handle that check here and return the appropriate value before the assert? IMO we should only consider removing this once this has been determined to be dead code because otherwise we are claiming that any other type is supported and correct here. Even in that case it might make sense to keep this assert since we will become aware of any other changes to the implementation of `preferredSize`

LayoutTests/fast/grid/grid-item-block-content-size-suggestion-crash.html
yulunwu2: Renamed + removed expectation file, 


0 tests ran as expected, 1 didn't:
Regressions: Unexpected missing results (1)
  fast/grid/grid-item-block-content-size-suggestion-crash.html [ Missing ]


From what I can see searching our test suite for *crash.html, see

https://searchfox.org/wubkat/source/LayoutTests/css3/unicode-bidi-float-crash.html

I think we do need these?

sammygill: oh might just be for WPTs then... see https://commits.webkit.org/311552@main

Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.cpp
yulunwu2: As of right now, we do not handle preferred aspect ratios. In that case, we can directly use the min-content-size which should be the borderBoxHeight() right?

Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.cpp
yulunwu2: SGTM, updated to only allow auto() through.

LayoutTests/fast/grid/grid-item-block-content-size-suggestion-crash.html
yulunwu2: ah gocha makes sense

yulunwu2: This does remind me, we should export more of our tests as WPTs sometime.

Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.cpp
sammygill: My first inclination would be that if the spec calls for the min-content size (height in this case) and we have an API that is called the same thing then we should use that...  

I haven't looked into whether using `layoutWithFormattingContextForBox` is the right thing to do but could you elaborate on what you mean by "This function had a dependency on integrationUtils::minContentHeight()
which does not support grid items"

What would it take to support grid items there? We might need to figure out what `RenderGrid` is doing and fill in the holes in the relevant portion of the integration logic.

yulunwu2: https://searchfox.org/wubkat/source/Source/WebCore/layout/integration/LayoutIntegrationFormattingContextLayout.cpp#238-257

I think the limitation (per comment) is that we can't ask RenderBox for content height

yulunwu2: I think we *could* add a cache to hash (grid item, layout constraints) -> height but that might be beyond the scope of this change? I think this would be something worth doing as I suspect having a cache will be a performance gain.

yulunwu2: Lets discuss Monday!

yulunwu2: After thinking it over, I'd like to land this as-is and follow up with a rename.

The issue with integrationUtils.minContentHeight() isn't doing what its name suggests: it is fetching a cached intrinsic content logical height that was saved by flex layout.

https://searchfox.org/wubkat/source/Source/WebCore/layout/integration/LayoutIntegrationUtils.cpp#59-63

I don't think this is applicable for RenderBox. Finding the height of a render box is dependent on the inline size to have an answer.

integrationUtils.minContentHeight(gridItem.layoutBox());

is unable to resolve the height because it doesn't have any inline constraints. I think we should land the patch as-is and rename integrationUtils.minContentHeight to indicate that it only supports flex items and not generic RenderBox.

Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
```diff
@@ -545,7 +545,7 @@ std::pair<UsedInlineSizes, UsedBlockSizes> GridLayout::layoutGridItems(const Pla
         auto usedInlineSizeForGridItem = GridLayoutUtils::usedInlineSizeForGridItem(gridItem, gridItem.usedInlineBorderAndPadding(), columnTrackSizingFunctions, gridAreaInlineSize, integrationUtils);
         usedInlineSizes.append(usedInlineSizeForGridItem);
 
-        auto usedBlockSizeForGridItem = GridLayoutUtils::usedBlockSizeForGridItem(gridItem, gridItem.usedBlockBorderAndPadding(), rowTrackSizingFunctions, gridAreaBlockSize, integrationUtils);
+        auto usedBlockSizeForGridItem = GridLayoutUtils::usedBlockSizeForGridItem(gridItem, gridItem.usedBlockBorderAndPadding(), rowTrackSizingFunctions, gridAreaBlockSize, formattingContext, usedInlineSizeForGridItem);
```
sammygill, line 549: Should this be the used inline size or the size of the grid area in the inline direction? If we do the former that seems to put a dependency between finding the inline and block size which may or may not be weird (I honestly don't know). If we aren't sure we should add a FIXME so we can investigate this later and help us remember

Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.cpp
```diff
@@ -409,19 +424,21 @@ GridItemSizingFunctions inlineAxisGridItemSizingFunctions(const IntegrationUtils
         [&integrationUtils](const PlacedGridItem& gridItem, LayoutUnit blockAxisConstraint) {
             return inlineAxisMaxContentContribution(gridItem, blockAxisConstraint, integrationUtils);
         },
-        [&integrationUtils](const PlacedGridItem& gridItem, const TrackSizingFunctionsList& trackSizingFunctions, LayoutUnit borderAndPadding, LayoutUnit availableSpace) {
+        [&integrationUtils](const PlacedGridItem& gridItem, const TrackSizingFunctionsList& trackSizingFunctions, LayoutUnit borderAndPadding, LayoutUnit availableSpace, LayoutUnit) {
```
sammygill: Maybe consider naming this along with an ASSERT_UNUSED_PARAM

Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.cpp
sammygill: Could you elaborate on finding the height of a RenderBox? RenderBox is just some generic type so I don't really get the connection between that and grid/flex layout. Flex item's also inherit from RenderBox in the same way so it would seem the issue would apply to them as well? 

I think it's probably fine to use `layoutWithFormattingContextForBox` but I would really want us to see if we can have a dedicated API for this and avoid coupling this logic with that function. I don't necessarily think that `minContenentHeight` is doing the wrong thing but it sounds like we might want an overloaded version that takes in an inline axis constraint that grid can use.

sammygill (approved):

Source/WebCore/layout/formattingContexts/grid/GridLayout.cpp
yulunwu2: Added FIXME

Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.cpp
yulunwu2: done, thanks for the review

Source/WebCore/layout/formattingContexts/grid/GridLayoutUtils.cpp
yulunwu2: Flex is stashing values during the legacy layout pass so heights are accessible in

    if (CheckedPtr flexContainer = dynamicDowncast<RenderFlexibleBox>(renderer->parent()))
        return flexContainer->cachedFlexItemIntrinsicContentLogicalHeight(renderer.get());

I think overloading could be done:

    LayoutUnit minContentHeight(const ElementBox&) const;
    // Flex items: returns cached intrinsic content logical height.

    LayoutUnit minContentHeight(const ElementBox&, LayoutUnit inlineConstraint) const;
    // Grid items: lays out at inlineConstraint and returns the
    // resulting block size from the geometry.

would do drastically different things behind the scenes but I think we can ensure the right one is being used by asserting.

yulunwu2: updated to use overload

PR #64591, opened 2026-05-09
[HDR] Use accelerated backing stores to apply the gain-map image in GPUProcess

shallawa:
#### 479c91177422567b22378a1781c15e944eeb1b09
<pre>
[HDR] Use accelerated backing stores to apply the gain-map image in GPUProcess
<a href="https://bugs.webkit.org/show_bug.cgi?id=314452">https://bugs.webkit.org/show_bug.cgi?id=314452</a>
<a href="https://rdar.apple.com/176605566">rdar://176605566</a>

Reviewed by Cameron McCormack.

DecodingDestination::BaseAndGainMap will be chosen to decode an ImageFrame
destination only if the ImageSource has gain-map and the context is remote.

ImageDecoder::createNativeImageAtIndex() can now create a NativeImage with
BaseAndGainMap. This NativeImage will return a calculated Headroom rather
than returning the Headroom of the baseImage which should be Headroom::None.

Decoding a frame may return a destination different from what was preferred.
BitmapImageSource has to account for this and do not assume the returned
NativeImage has to be the requested DecodingDestination.

When comparing DecodingDestinations in DecodingOptions for compatibility,
they have to be equal except one exception: A fully decoded HDR image is
compatible with DecodingDestination::BaseAndGainMap request.

* Source/WebCore/platform/graphics/BitmapImageSource.cpp:
(WebCore::BitmapImageSource::preferredDecodingDestination const):
(WebCore::BitmapImageSource::compatibleDecodingDestinationWithOptionsAtIndex const):
(WebCore::BitmapImageSource::decode):
(WebCore::BitmapImageSource::requestNativeImageAtIndexIfNeeded):
(WebCore::BitmapImageSource::nativeImageAtIndexCacheIfNeeded):
(WebCore::BitmapImageSource::nativeImageAtIndexRequestIfNeeded):
(WebCore::BitmapImageSource::isCompatibleWithOptionsAtIndex const): Deleted.
* Source/WebCore/platform/graphics/BitmapImageSource.h:
* Source/WebCore/platform/graphics/DecodingOptions.h:
(WebCore::DecodingOptions::isCompatibleWith const):
* Source/WebCore/platform/graphics/GraphicsContext.h:
(WebCore::GraphicsContext::renderingMethod const):
* Source/WebCore/platform/graphics/ImageDecoder.cpp:
(WebCore::ImageDecoder::createNativeImageAtIndex):
* Source/WebCore/platform/graphics/ImageDecoder.h:
* Source/WebCore/platform/graphics/ImageFrame.cpp:
(WebCore::ImageFrame::clearImage):
(WebCore::ImageFrame::compatibleDecodingDestinationWithOptions const):
(WebCore::ImageFrame::hasDecodedNativeImageCompatibleWithOptions const): Deleted.
* Source/WebCore/platform/graphics/ImageFrame.h:
(WebCore::ImageFrame::decodingDestinationIfExists const):
* Source/WebCore/platform/graphics/ImageFrameWorkQueue.cpp:
(WebCore::ImageFrameWorkQueue::start):
* Source/WebCore/platform/graphics/ImageFrameWorkQueue.h:
(WebCore::ImageFrameWorkQueue::Request::isCompatibleWith const):
* Source/WebCore/platform/graphics/NativeImage.cpp:
(WebCore::NativeImage::NativeImage):
(WebCore::NativeImage::replacePlatformImage const):
* Source/WebCore/platform/graphics/NativeImage.h:
(WebCore::NativeImage::hasHDRGainMap const):
(WebCore::NativeImage::baseImageHeadroom const):
(WebCore::NativeImage::headroom const):
* Source/WebCore/platform/graphics/ShareableBitmap.cpp:
(WebCore::ShareableBitmapConfiguration::ShareableBitmapConfiguration):
(WebCore::ShareableBitmap::create):
(WebCore::ShareableBitmap::ShareableBitmap):
* Source/WebCore/platform/graphics/ShareableBitmap.h:
(WebCore::ShareableBitmapConfiguration::baseImageHeadroom const):
(WebCore::ShareableBitmapConfiguration::headroom const): Deleted.
* Source/WebCore/platform/graphics/cairo/NativeImageCairo.cpp:
(WebCore::NativeImage::headroom const): Deleted.
* Source/WebCore/platform/graphics/cg/ImageDecoderCG.cpp:
(WebCore::ImageDecoderCG::frameGainMapAtIndex):
* Source/WebCore/platform/graphics/cg/NativeImageCG.cpp:
(WebCore::NativeImage::computeHeadroom const):
(WebCore::NativeImage::headroom const): Deleted.
* Source/WebCore/platform/graphics/cg/ShareableBitmapCG.mm:
(WebCore::ShareableBitmapConfiguration::ShareableBitmapConfiguration):
(WebCore::ShareableBitmap::createBasePlatformImage):
(WebCore::ShareableBitmap::createPlatformImage):
* Source/WebCore/platform/graphics/skia/NativeImageSkia.cpp:
(WebCore::NativeImage::headroom const): Deleted.
* Source/WebKit/WebProcess/GPU/graphics/RemoteGraphicsContextProxy.h:
* Source/WebKit/WebProcess/GPU/graphics/RemoteResourceCacheProxy.cpp:
(WebKit::createShareableBitmapForNativeImage):

Canonical link: <a href="https://commits.webkit.org/313339@main">https://commits.webkit.org/313339@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/ac15039421bc7e34056c6a5d8b415179b1b488c0

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/162710 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/36186 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/29332 "Built successfully") | [loading 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/171648 "Build is in progress. Recent messages:Printed configuration; Running apply-patch; Checked out pull request; Compiled WebKit") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/119156 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/d53d69c6-4cb9-4de4-ac1f-fc0bb1cd28a1) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/164579 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/36290 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/36190 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/126291 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/88936 "2 flakes 17 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/606ee2d8-05d6-42f5-86b9-2a0460b1b492) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/165669 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/28640 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/146198 "Passed tests") | [loading 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/106868 "Build is in progress. Recent messages:Printed configuration; Skipping applying patch since patch_id isn't provided; Checked out pull request; run-api-tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/3e75b8e9-f56b-4068-99bd-59f98284d99c) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/27686 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/26217 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/19348 "Built successfully") | | 
| [⏳ 🛠 🧪 jsc ](https://ews-build.webkit.org/#/builders/JSC-Tests-EWS "Waiting in queue, processing has not started yet") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/137394 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/23927 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/174152 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/21465 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/25572 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/134600 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/35860 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/30314 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/134596 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36336 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/35844 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/145723 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/98226 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/29136 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/22559 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/35354 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/103105 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/34855 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/35100 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/35003 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/platform/graphics/BitmapImageSource.cpp
+++ b/Source/WebCore/platform/graphics/BitmapImageSource.cpp
@@ -335,10 +335,13 @@ bool BitmapImageSource::hasEverAnimated() const
     return m_frameAnimator && m_frameAnimator->hasEverAnimated();
 }
 
-DecodingDestination BitmapImageSource::preferredDecodingDestination(GraphicsContext&, ImagePaintingOptions options) const
+DecodingDestination BitmapImageSource::preferredDecodingDestination(GraphicsContext& context, ImagePaintingOptions options) const
 {
     if (options.drawsHDRContent() == DrawsHDRContent::No || options.dynamicRangeLimit() == PlatformDynamicRangeLimit::standard() || !hasHDRGainMap())
         return DecodingDestination::Base;
+
+    if (options.allowAcceleratedApplyGainMap() == AllowAcceleratedApplyGainMap::Yes && !context.renderingMethod())
+        return DecodingDestination::BaseAndGainMap;
 
     return DecodingDestination::ShouldDecodeToHDR;
 }
@@ -372,9 +375,9 @@ bool BitmapImageSource::isPendingDecodingAtIndex(unsigned index, SubsamplingLeve
     return m_workQueue->isPendingDecodingAtIndex(index, subsamplingLevel, options);
 }
 
-bool BitmapImageSource::isCompatibleWithOptionsAtIndex(unsigned index, SubsamplingLevel subsamplingLevel, const DecodingOptions& options) const
+std::optional<DecodingDestination> BitmapImageSource::compatibleDecodingDestinationWithOptionsAtIndex(unsigned index, SubsamplingLevel subsamplingLevel, const DecodingOptions& options) const
 {
-    return frameAtIndex(index).hasDecodedNativeImageCompatibleWithOptions(options, subsamplingLevel);
+    return frameAtIndex(index).compatibleDecodingDestinationWithOptions(options, subsamplingLevel);
 }
 
 void BitmapImageSource::decode(Function<void(DecodingStatus)>&& decodeCallback)
@@ -389,10 +392,10 @@ void BitmapImageSource::decode(Function<void(DecodingStatus)>&& decodeCallback)
 
     // FIXME: HTMLImageElement.decode() needs a parameter to control whether it should decode SDR or HDR image.
     auto preferredDecodingDestination = hasHDRContent() ? DecodingDestination::ShouldDecodeToHDR : DecodingDestination::Base;
-    auto isCompatibleNativeImage = isCompatibleWithOptionsAtIndex(index, SubsamplingLevel::Default, { DecodingMode::Asynchronous, preferredDecodingDestination });
+    auto compatibleDecodingDestination = compatibleDecodingDestinationWithOptionsAtIndex(index, SubsamplingLevel::Default, { DecodingMode::Asynchronous, preferredDecodingDestination });
 
     RefPtr frameAnimator = this->frameAnimator();
-    if (frameAnimator && (frameAnimator->hasEverAnimated() || isCompatibleNativeImage)) {
+    if (frameAnimator && (frameAnimator->hasEverAnimated() || compatibleDecodingDestination)) {
         // startAnimation() always decodes the nextFrame which is currentFrameIndex + 1.
         // If primaryFrameIndex = 0, then the sequence of decoding is { 1, 2, .., n, 0, 1, ...}.
         if (startAnimation(SubsamplingLevel::Default, DecodingMode::Asynchronous)) {
@@ -401,7 +404,7 @@ void BitmapImageSource::decode(Function<void(DecodingStatus)>&& decodeCallback)
         }
     }
 
-    if (!isCompatibleNativeImage) {
+    if (!compatibleDecodingDestination) {
         LOG(Images, "BitmapImageSource::%s - %p - url: %s. Decoding for frame at index = %d will be requested.", __FUNCTION__, this, sourceUTF8().data(), index);
         requestNativeImageAtIndex(index, SubsamplingLevel::Default, ImageAnimatingState::No, { DecodingMode::Asynchronous, preferredDecodingDestination });
         return;
@@ -557,23 +560,23 @@ DecodingStatus BitmapImageSource::requestNativeImageAtIndex(unsigned index, Subs
     return DecodingStatus::Decoding;
 }
 
-DecodingStatus BitmapImageSource::requestNativeImageAtIndexIfNeeded(unsigned index, SubsamplingLevel subsamplingLevel, ImageAnimatingState animatingState, const DecodingOptions& options)
+Expected<DecodingDestination, DecodingStatus> BitmapImageSource::requestNativeImageAtIndexIfNeeded(unsigned index, SubsamplingLevel subsamplingLevel, ImageAnimatingState animatingState, const DecodingOptions& options)
 {
     if (index >= m_frames.size())
-        return DecodingStatus::Invalid;
+        return makeUnexpected(DecodingStatus::Invalid);
 
     // Never decode the same frame from two different threads.
     if (isPendingDecodingAtIndex(index, subsamplingLevel, options)) {
         LOG(Images, "BitmapImageSource::%s - %p - url: %s. Frame at index = %d is being decoded.", __FUNCTION__, this, sourceUTF8().data(), index);
         ++m_blankDrawCountForTesting;
-        return DecodingStatus::Decoding;
+        return makeUnexpected(DecodingStatus::Decoding);
     }
 
-    // isCompatibleWithOptionsAtIndex() returns true only if the frame is complete.
-    if (isCompatibleWithOptionsAtIndex(index, subsamplingLevel, options))
-        return DecodingStatus::Complete;
+    // compatibleDecodingDestination() returns a DecodingDestination only if the frame is complete.
+    if (auto compatibleDecodingDestination = compatibleDecodingDestinationWithOptionsAtIndex(index, subsamplingLevel, options))
+        return *compatibleDecodingDestination;
 
-    return requestNativeImageAtIndex(index, subsamplingLevel, animatingState, options);
+    return makeUnexpected(requestNativeImageAtIndex(index, subsamplingLevel, animatingState, options));
 }
 
 Expected<Ref<NativeImage>, DecodingStatus> BitmapImageSource::nativeImageAtIndexCacheIfNeeded(unsigned index, SubsamplingLevel subsamplingLevel, const DecodingOptions& options)
@@ -592,18 +595,25 @@ Expected<Ref<NativeImage>, DecodingStatus> BitmapImageSource::nativeImageAtIndex
         return makeUnexpected(DecodingStatus::Decoding);
     }
 
-    if (!isCompatibleWithOptionsAtIndex(index, subsamplingLevel, options)) {
-        DecodingOptions decodingOptions = { DecodingMode::Synchronous, options.decodingDestination() };
-        PlatformImagePtr platformImage = m_decoder->createFrameImageAtIndex(index, subsamplingLevel, decodingOptions);
+    auto decodingDestination = options.decodingDestination();
 
-        RefPtr nativeImage = NativeImage::create(WTF::move(platformImage));
-        if (!nativeImage)
+    if (auto compatibleDecodingDestination = compatibleDecodingDestinationWithOptionsAtIndex(index, subsamplingLevel, options))
+        decodingDestination = *compatibleDecodingDestination;
+    else {
+        auto decodingOptions = DecodingOptions { DecodingMode::Synchronous, decodingDestination };
+
+        auto result = m_decoder->createNativeImageAtIndex(index, subsamplingLevel, decodingOptions);
+        if (!result)
             return makeUnexpected(DecodingStatus::Invalid);
 
-        cacheNativeImageAtIndex(index, subsamplingLevel, decodingOptions, nativeImage.releaseNonNull());
+        Ref nativeImage = WTF::move(std::get<Ref<NativeImage>>(*result));
+        decodingDestination = std::get<DecodingDestination>(*result);
+
+        decodingOptions = { DecodingMode::Synchronous, decodingDestination };
+        cacheNativeImageAtIndex(index, subsamplingLevel, decodingOptions, WTF::move(nativeImage));
     }
 
-    if (RefPtr nativeImage = frameAtIndex(index).nativeImage(options.decodingDestination()))
+    if (RefPtr nativeImage = frameAtIndex(index).nativeImage(decodingDestination))
         return nativeImage.releaseNonNull();
 
     return makeUnexpected(DecodingStatus::Invalid);
@@ -616,11 +626,12 @@ Expected<Ref<NativeImage>, DecodingStatus> BitmapImageSource::nativeImageAtIndex
 
     ASSERT(!isAnimated());
 
-    auto status = requestNativeImageAtIndexIfNeeded(index, subsamplingLevel, ImageAnimatingState::No, options);
-    if (status == DecodingStatus::Invalid || status == DecodingStatus::Decoding)
-        return makeUnexpected(status);
+    auto decodingDestinationOrStatus = requestNativeImageAtIndexIfNeeded(index, subsamplingLevel, ImageAnimatingState::No, options);
+    if (!decodingDestinationOrStatus)
+        return makeUnexpected(decodingDestinationOrStatus.error());
 
-    if (RefPtr nativeImage = frameAtIndex(index).nativeImage(options.decodingDestination()))
+    auto decodingDestination = decodingDestinationOrStatus.value();
+    if (RefPtr nativeImage = frameAtIndex(index).nativeImage(decodingDestination))
         return nativeImage.releaseNonNull();
 
     return makeUnexpected(DecodingStatus::Invalid);
--- a/Source/WebCore/platform/graphics/ImageFrameWorkQueue.cpp
+++ b/Source/WebCore/platform/graphics/ImageFrameWorkQueue.cpp
@@ -76,8 +76,15 @@ void ImageFrameWorkQueue::start()
             if (minimumDecodingDuration > 0_s)
                 startingTime = MonotonicTime::now();
 
-            PlatformImagePtr platformImage = protectedDecoder->createFrameImageAtIndex(request.index, request.subsamplingLevel, request.options);
-            RefPtr nativeImage = NativeImage::create(WTF::move(platformImage));
+            RefPtr<NativeImage> nativeImage;
+            DecodingDestination decodingDestination = request.options.decodingDestination();
+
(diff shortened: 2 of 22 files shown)


Source/WebCore/platform/graphics/cg/ShareableBitmapCG.mm
```diff
@@ -263,7 +262,7 @@
     CGImageDumpToFile(outputPlatformImage.get(), "*/output-image.br2");
 #endif
 
-    return outputPlatformImage;
+    return WTF::move(outputPlatformImage);
```
heycam: I think it is unlikely `WTF::move` is needed here, due to Named Return Value Optimization.

Source/WebCore/platform/graphics/BitmapImage.cpp
```diff
@@ -136,8 +136,7 @@ ImageDrawResult BitmapImage::draw(GraphicsContext& context, const FloatRect& des
         else {
 #endif
             auto headroom = options.headroom();
-
-            if (headroom == Headroom::FromImage)
+            if (headroom == Headroom::FromImage && !nativeImage->hasHDRGainMap())
                 headroom = nativeImage->headroom();
```
heycam: It's not obvious to me why we only do this for non-gain-map images. Can you make it clear why that is?

Source/WebCore/platform/graphics/BitmapImageSource.cpp
```diff
@@ -335,11 +335,14 @@ bool BitmapImageSource::hasEverAnimated() const
...
 }
 
-DecodingDestination BitmapImageSource::preferredDecodingDestination(GraphicsContext&, ImagePaintingOptions options) const
+DecodingDestination BitmapImageSource::preferredDecodingDestination(GraphicsContext& context, ImagePaintingOptions options) const
 {
     if (options.drawsHDRContent() == DrawsHDRContent::No || options.dynamicRangeLimit() == PlatformDynamicRangeLimit::standard() || !hasHDRGainMap())
         return DecodingDestination::Base;
 
+    if (options.allowAcceleratedApplyGainMap() == AllowAcceleratedApplyGainMap::Yes && !context.renderingMethod())
+        return DecodingDestination::BaseAndGainMap;
```
heycam, line 344: Why `!context.renderingMethod()`? And indeed, what does `RenderingMethod` represent? It currently has only a single enumerated value.

Source/WebCore/platform/graphics/BitmapImageSource.cpp
```diff
@@ -557,23 +560,23 @@ DecodingStatus BitmapImageSource::requestNativeImageAtIndex(unsigned index, Subs
...
 
     // Never decode the same frame from two different threads.
     if (isPendingDecodingAtIndex(index, subsamplingLevel, options)) {
         LOG(Images, "BitmapImageSource::%s - %p - url: %s. Frame at index = %d is being decoded.", __FUNCTION__, this, sourceUTF8().data(), index);
         ++m_blankDrawCountForTesting;
-        return DecodingStatus::Decoding;
+        return makeUnexpected(DecodingStatus::Decoding);
     }
 
     // isCompatibleWithOptionsAtIndex() returns true only if the frame is complete.
```
heycam: Comment needs rewording for the new function.

Source/WebCore/platform/graphics/BitmapImageSource.cpp
```diff
@@ -557,23 +560,23 @@ DecodingStatus BitmapImageSource::requestNativeImageAtIndex(unsigned index, Subs
     return DecodingStatus::Decoding;
 }
 
-DecodingStatus BitmapImageSource::requestNativeImageAtIndexIfNeeded(unsigned index, SubsamplingLevel subsamplingLevel, ImageAnimatingState animatingState, const DecodingOptions& options)
+Expected<DecodingDestination, DecodingStatus> BitmapImageSource::requestNativeImageAtIndexIfNeeded(unsigned index, SubsamplingLevel subsamplingLevel, ImageAnimatingState animatingState, const DecodingOptions& options)
```
heycam, line 563: Does `requestNativeImageAtIndexIfNeeded` ever return `DecodingStatus::Partial`? I am wondering what it means if we have partially decoded an image: don't we need a `DecodingDestination` in that case too, to paint the partial image?

Also, using `Expected<DecodingDestination, DecodingStatus>` leaves us with the possibility of returning `DecodingStatus::Complete`, but would we always want to return a `DecodingDestination` in that case?

I am just concerned this type doesn't represent the exact set of values that can be returned from this function.

Source/WebCore/platform/graphics/ImageFrameWorkQueue.cpp
```diff
@@ -95,7 +102,7 @@ void ImageFrameWorkQueue::start()
                 }
 
                 // The DecodeQueue may have been cleared before the frame was decoded.
-                if (protectedThis->decodeQueue().isEmpty() || protectedThis->decodeQueue().first() != request) {
+                if (protectedThis->decodeQueue().isEmpty() || !request.isCompatibleWith(protectedThis->decodeQueue().first())) {
```
heycam, line 105: Why do we do this now instead of comparing `request` exactly? If we happen to clear the decode queue, then request a different but compatible `DecodingDestination`, don't we still want to log and return early?

Source/WebKit/WebProcess/GPU/graphics/RemoteGraphicsContextProxy.cpp
```diff
@@ -320,7 +320,7 @@ void RemoteGraphicsContextProxy::drawNativeImage(const NativeImage& image, const
 {
 #if HAVE(SUPPORT_HDR_DISPLAY_APIS)
     auto headroom = options.headroom();
-    if (headroom == Headroom::FromImage)
+    if (headroom == Headroom::FromImage && !image.hasHDRGainMap())
         headroom = image.headroom();
```
heycam: Why don't we want to get the headroom of a non-gain map HDR image? Don't we want to take its headroom into account when computing the tile headroom?

Source/WebCore/platform/graphics/BitmapImageSource.cpp
shallawa: RenderingMethod was added in https://commits.webkit.org/219626@main like this
```
enum class RenderingMethod : uint8_t {
    Default,
    Local,
    DisplayList
};
```
It was added to control the backend type which  `GraphicsContext::createImageBuffer()` and its siblings create. `RenderingMethod::Default` means the backend will similar to the underlying `GraphicsContext`.

https://commits.webkit.org/268193@main removed the Default option to be like this
```
enum class RenderingMethod : bool { Local, DisplayList };
```
The commit message says:
>Another cleaning is to pass RenderingMethod as an std::optional argument to the three methods. This means we do not need the value "Default" since passing std::nullopt will have the same meaning.

https://commits.webkit.org/268193@main changed it to be 
```
enum class RenderingMethod : bool { Local };
```
The commit message says:
>Does not remove RenderingMethod enum because currently it affects the "local or remote" decision via holding it with std::optional. This will be addressed in the future.

So currently `!context.renderingMethod()` means the context is **remote**. All this work has to happen **only in GPUProcess**. If the context is local, e.g. PDF context, the image has to be fully decoded in WebContent process.

shallawa: And this is why I had to add this override in RemoteGraphicsContextProxy
```
std::optional<WebCore::RenderingMethod> renderingMethod() const final { return std::nullopt; }
```
Because the base class `GraphicsContext` has this virtual method:
```
virtual std::optional<RenderingMethod> renderingMethod() const { return RenderingMethod::Local; }
```

Source/WebCore/platform/graphics/cg/ShareableBitmapCG.mm
shallawa: This was removed.

Source/WebCore/platform/graphics/BitmapImageSource.cpp
shallawa: >Does requestNativeImageAtIndexIfNeeded ever return DecodingStatus::Partial? I am wondering what it means if we have partially decoded an image: don't we need a DecodingDestination in that case too, to paint the partial image?

No requestNativeImageAtIndexIfNeeded() never returns `DecodingStatus::Partial` even if the ImageFrame has a cached DecodingDestination with `DecodingStatus::Partial`. We always want to re-decode in the case.

Source/WebCore/platform/graphics/BitmapImageSource.cpp
shallawa: The comment is fixed.

Source/WebCore/platform/graphics/BitmapImageSource.cpp
shallawa: >Also, using Expected<DecodingDestination, DecodingStatus> leaves us with the possibility of returning DecodingStatus::Complete, but would we always want to return a DecodingDestination in that case?

No this function will not return `DecodingStatus::Complete`. The possible return values of this functions are:
1. error: `DecodingStatus::Invalid`
2. error: `DecodingStatus::Decoding`
3. value: `DecodingDestination`

Source/WebKit/WebProcess/GPU/graphics/RemoteGraphicsContextProxy.cpp
shallawa: This change was removed.

Source/WebCore/platform/graphics/ImageFrameWorkQueue.cpp
shallawa: The scenario is the following:
1. Decoding a `NativeImage` with `DecodingDestination::BaseAndGainMap` is requested.
2. Getting the gain-map image fails because of sandboxing. So we fall back to decoding with `ShouldDecodeToHDR`
3. `request.decodingDestination` is changed to from `BaseAndGainMap` to `ShouldDecodeToHDR`
4. We `callOnMainThread(..., request, ...)`
5. When comparing `request` and `protectedThis->decodeQueue().first()`, we find the first has `ShouldDecodeToHDR` and the second has `BaseAndGainMap`. But we still need to cache the decoded frame via `imageFrameDecodeAtIndexHasFinished()`. If we log and return early, this frame will never be displayed.

Source/WebCore/platform/graphics/BitmapImageSource.cpp
heycam: Right, my comment is about whether `Expected<DecodingDestination, DecodingStatus>` is the right type to be returning here, since it allows values that we will never return. Which makes it less clear at the calling site.

Source/WebCore/platform/graphics/ImageFrameWorkQueue.cpp
heycam: OK. This is very non-obvious. I think it is worth a comment to say why we use `isCompatibleWith` here, to support this scenario.

Source/WebCore/platform/graphics/cg/NativeImageCG.cpp
```diff
@@ -95,15 +95,22 @@ DestinationColorSpace NativeImage::colorSpace() const
...
+#else
+    m_baseImageHeadroom = Headroom::None;
 #endif
-}
 
-Headroom NativeImage::headroom() const
-{
-    return m_headroom;
+    if (hasHDRGainMap()) {
+        m_headroom = gainMapImageHeadroom;
```
heycam, line 110: Do gain maps always have this fixed headroom?

heycam (approved):
r=me with comments but please consider a followup where we use a different type to return from `requestNativeImageAtIndexIfNeeded`, which does not allow values that can never be returned.

PR #60911, opened 2026-03-19
[API Tests] Reduce run-api-tests startup overhead when running specific tests

rr-codes:
#### a80da14addae7e5ec036c259da7b3f739b678a56
<pre>
[API Tests] Reduce run-api-tests startup overhead when running specific tests
<a href="https://bugs.webkit.org/show_bug.cgi?id=310247">https://bugs.webkit.org/show_bug.cgi?id=310247</a>
<a href="https://rdar.apple.com/172887928">rdar://172887928</a>

Reviewed by Jonathan Bedard.

This patch reduces the startup overhead of run-api-tests by addressing
several bottlenecks in the Python script&apos;s initialization path.

1. Defer default_configuration() in configuration_options() (~500ms)

The -t/--target option eagerly created a Config object and called
default_configuration(), which spawns a Perl subprocess
(webkit-build-directory) to determine the build directory. This
happened unconditionally at option-parse time, even when --debug or
--release was explicitly passed. Changed the default to None; the
port base class (base.py:168-169) already has a lazy fallback that
calls default_configuration() only when no configuration was set.

2. Make SDK resolution lazy in embedded port modules (~1-1.5s)

Six embedded port modules (ios_simulator, ios_device,
watch_simulator, watch_device, visionos_simulator, visionos_device)
called apple_additions().get_sdk() at class-definition time, which
triggered xcodebuild -showsdks (~1s+) the moment any of these
modules was imported by PortFactory.get(). Fixed by converting SDK
from a class-level attribute to a lazy @property that defers the
lookup until first access. Each class now has a _DEFAULT_SDK class
attribute for the base SDK name and a @property SDK that calls
apple_additions().get_sdk() only when needed. This aligns with the
EmbeddedPort base class, which already declares SDK as
@property @abstractmethod. Also updated EmbeddedDevicePort&apos;s
determine_full_port_name() classmethod to use cls._DEFAULT_SDK
instead of cls.SDK, since properties don&apos;t work on class references.
Additionally, PORT_CLASSES is restored to alphabetical order and
the import logic is extracted into a _load_port_class() helper.

3. Defer configuration_for_upload until needed (~100-200ms)

configuration_for_upload was computed unconditionally before test
collection, but is only needed when --report-urls is specified.
Moved it into the upload block.

4. Skip test collection when exact test names are provided.

Previously, `_collect_tests` always copied the test binary to &quot;ToBeListed&quot; and
spawned it with --gtest_list_tests to enumerate every available test, even when
[...]

PR #64594, opened 2026-05-09
Reduce run-webkit-tests startup overhead for faster iteration

aproskuryakov:
#### 7d4b9ad5fc390e75e6391417c74bff8a5ba74da1
<pre>
Reduce run-webkit-tests startup overhead for faster iteration
<a href="https://bugs.webkit.org/show_bug.cgi?id=314459">https://bugs.webkit.org/show_bug.cgi?id=314459</a>
<a href="https://rdar.apple.com/176618242">rdar://176618242</a>

Reviewed by Sam Sneddon and Jonathan Bedard.

`run-webkit-tests fast/dom/ChildNode-replaceWith.html` wall clock drops from ~5s
to ~2.8s. WebKitTestRunner itself still takes ~1s; the rest is setup overhead
that this change trims.

Each of these improvements was independently measurable:
- LayoutTestHelper&apos;s ~440ms color-profile install now overlaps with test discovery
  instead of blocking _set_up_run.
- configuration_for_upload (which pulls in requests, ~80ms) runs only when
  --report is passed.
- webkit-build-directory perl calls: 4 → 1, via a class-level cache and a lazy
  --target default (~500ms).
- WebKitTestRunner warmup is skipped when child_processes == 1.
- Several production import paths no longer pull in unittest.mock,
  webkitcorepy.mocks submodules, the whole requests stack, or the embedded-port
  chain on a Mac run.
- Misc: cached inner finder, precompiled regex, lazy port-factory table.

* Tools/Scripts/libraries/webkitcorepy/webkitcorepy/mocks/__init__.py:
Load only ContextStack eagerly. Time / Subprocess / Requests / Terminal /
FileLock / Environment are exposed via PEP 562 module-level __getattr__ +
CallByNeed, so callers that only need ContextStack don&apos;t pull in their
dependencies (unittest.mock, requests, etc.). Resolved attributes are written
back into globals() so later access (and pickle/isinstance) skip the proxy.

* Tools/Scripts/libraries/webkitcorepy/webkitcorepy/partial_proxy.py:
* Tools/Scripts/libraries/webkitcorepy/webkitcorepy/timeout.py:
Defer importing unittest.mock to first use via CallByNeed. These modules are
imported on every webkitcorepy load but rarely call into mock at production
runtime.

* Tools/Scripts/libraries/webkitscmpy/webkitscmpy/__init__.py:
Drop the eager `from webkitscmpy import mocks` import; tests that need it import
it explicitly. Cuts an unconditional load of mocks.* from production paths.

* Tools/Scripts/webkitpy/common/version_name_map.py:
Add DARWIN_CURRENT_VERSION = Version(26) as the single source of truth for the
current Darwin major; the ios/tvos/watchos/visionos PUBLIC_TABLE entries already
share Darwin&apos;s value through this constant.

* Tools/Scripts/webkitpy/layout_tests/controllers/layout_test_finder_legacy.py:
(LayoutTestFinder.__init__):
(LayoutTestFinder._make_inner_finder):
(LayoutTestFinder.find_tests_by_path):
(LayoutTestFinder._is_test_file):
(LayoutTestFinder._is_w3c_resource_file):
Memoize the inner LayoutTestFinder_New per device_type with functools.cache
bound to the instance, so repeated lookups don&apos;t redo baseline_search_path()
work. Bound on the instance (not the class) so the cache doesn&apos;t outlive the
LayoutTestFinder.

* Tools/Scripts/webkitpy/layout_tests/controllers/manager.py:
(Manager.run):
Kick start_helper_async off before _collect_tests so the helper warmup overlaps
with test discovery and expectations parsing.
(Manager._set_up_run):
Join on the helper via wait_for_helper_ready instead of start_helper.
(Manager._update_worker_count):
After clamping options.child_processes, sync the value back onto the port via
set_option. Port has its own copy of options (_create_port_for_driver does
optparse.Values copy.copy), so the warmup gate and embedded-port device-count
logic see the effective worker count.
(Manager._end_test_run):
Compute configuration_for_upload only when --report is passed, avoiding an
unconditional import of requests on every run.

* Tools/Scripts/webkitpy/layout_tests/models/test_expectations.py:
(TestExpectationParser._tokenize_line):
Collapse whitespace with `&quot; &quot;.join(s.split())` instead of an `re.sub`, and
precompile _BUG_TOKEN_RE.

* Tools/Scripts/webkitpy/port/base.py:
(Port._helper_process):
(Port._helper_started):
(Port._helper_ready):
Class-level state for the LayoutTestHelper. _helper_started records that the
spawn was attempted, so a None _helper_process is interpreted as a failed spawn
rather than no spawn at all.
(Port.start_helper_async):
(Port.wait_for_helper_ready):
New no-op defaults; MacPort overrides them. Splitting the old start_helper into
a kick-off + join lets the caller overlap helper warmup with other work.
(Port.stop_helper):
Reset _helper_started and _helper_ready so a later run isn&apos;t mistaken for one
whose spawn failed.

* Tools/Scripts/webkitpy/port/config.py:
(Config._build_directories):
(Config._clear_cache_for_testing):
(Config.build_directory):
Cache is class-level now (invariant within a process and shareable across
Config instances) and keyed on
(port_implementation, configuration, for_host, use_cmake).

* Tools/Scripts/webkitpy/port/base_unittest.py:
* Tools/Scripts/webkitpy/port/config_unittest.py:
(PortTest.setUp):
(ConfigTest.setUp):
Clear the class-level Config cache between tests and between phases within
test_build_directory so it doesn&apos;t leak across cases.

* Tools/Scripts/webkitpy/port/darwin.py:
(DarwinPort.CURRENT_VERSION):
Import DARWIN_CURRENT_VERSION from version_name_map instead of duplicating
Version(26).

* Tools/Scripts/webkitpy/port/factory.py:
(_load_port_class):
(PortFactory.PORT_CLASSES):
(PortFactory.get):
Replace the eager loop over all 14 port modules with a table mapping port-name
prefix to a &quot;module.Class&quot; path string. Only the matched port module is
imported, via importlib.import_module + getattr.
(configuration_options):
--target now defaults to None. Port.__init__ already calls
set_option_default(&apos;configuration&apos;, self.default_configuration()) after
parsing, so the old eager Config.default_configuration() call at
option-definition time (one perl invocation every run) is redundant.

* Tools/Scripts/webkitpy/port/factory_unittest.py:
(FactoryTest.test_port_classes_table_consistency):
Assert each prefix in PORT_CLASSES matches the imported class&apos;s port_name so
the duplication cannot silently drift.

* Tools/Scripts/webkitpy/port/mac.py:
(MacPort.setup_test_run):
Skip the WebKitTestRunner warmup when child_processes == 1; the warmup exists
to avoid first-run binary-verification thrash with N parallel workers
(webkit.org/b/242106), which doesn&apos;t apply to a single worker.
(MacPort.start_helper):
Now just start_helper_async + wait_for_helper_ready.
(MacPort.start_helper_async):
(MacPort.wait_for_helper_ready):
Split into non-blocking popen + blocking readline. Track helper state on
Port._helper_process / _helper_started / _helper_ready so wait returns False
for a helper that failed to spawn.

Canonical link: <a href="https://commits.webkit.org/314080@main">https://commits.webkit.org/314080@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/f8ceb8e9e4df0883fd445729827b0d2cd56d6411

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/165049 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/38540 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/32117 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/174091 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/122306 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/ca3d46fa-96e5-430b-8a23-a6a09d30f351) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/166917 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/38874 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/38541 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/128263 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/122306 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/94e643f1-31d0-48a9-b966-14ce69460d47) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/168008 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/162/builds/30573 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/148311 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/108789 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/6f3d85a0-9e8e-40fb-8b0f-b3bc9f434d44) 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/164369 "Passed tests") | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/29620 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/28241 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/21846 "Built successfully") | | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/139515 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/26010 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/176614 "Built successfully") | | 
| | | [  ~~🧪 mac-wk2~~](https://ews-build.webkit.org/#/builders/169/builds/27715 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/1/builds/136581 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | | 
| [✅ 🧪 services](https://ews-build.webkit.org/#/builders/28/builds/164453 "Passed tests") | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/38608 "Built successfully") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/32377 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/136525 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36797 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/39298 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/147806 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/101597 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/31797 "Passed tests") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/24673 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/37808 "Built successfully") | | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/37205 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/2748 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/37451 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/37349 "Built successfully") | | | |

### DIFF
--- a/Tools/Scripts/libraries/webkitcorepy/webkitcorepy/mocks/__init__.py
+++ b/Tools/Scripts/libraries/webkitcorepy/webkitcorepy/mocks/__init__.py
@@ -20,11 +20,44 @@
 # (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-from webkitcorepy.mocks.context_stack import ContextStack
-from webkitcorepy.mocks.time_ import Time
-from webkitcorepy.mocks.subprocess import ProcessCompletion, Subprocess
+import importlib
 
-from webkitcorepy.mocks.requests_ import Response, Requests
-from webkitcorepy.mocks.terminal import Terminal
-from webkitcorepy.mocks.file_lock import FileLock
-from webkitcorepy.mocks.environment import Environment
+from webkitcorepy.call_by_need import CallByNeed
+
+# ContextStack is a base class for OutputCapture, OutputDuplicate, and PartialProxy, all
+# of which webkitcorepy/__init__.py loads eagerly, so it's effectively never lazy.
+from webkitcorepy.mocks.context_stack import ContextStack
+
+
+def _lazy(module_path, attr_name):
+    return CallByNeed(lambda: getattr(importlib.import_module(module_path), attr_name))
+
+
+# Other symbols load lazily — see __getattr__ — so callers importing only ContextStack
+# don't pay for time_/subprocess/requests_/etc.
+_LAZY = {
+    'Time': _lazy('webkitcorepy.mocks.time_', 'Time'),
+    'ProcessCompletion': _lazy('webkitcorepy.mocks.subprocess', 'ProcessCompletion'),
+    'Subprocess': _lazy('webkitcorepy.mocks.subprocess', 'Subprocess'),
+    'Response': _lazy('webkitcorepy.mocks.requests_', 'Response'),
+    'Requests': _lazy('webkitcorepy.mocks.requests_', 'Requests'),
+    'Terminal': _lazy('webkitcorepy.mocks.terminal', 'Terminal'),
+    'FileLock': _lazy('webkitcorepy.mocks.file_lock', 'FileLock'),
+    'Environment': _lazy('webkitcorepy.mocks.environment', 'Environment'),
+}
+
+
+def __getattr__(name):
+    proxy = _LAZY.get(name)
+    if proxy is None:
+        raise AttributeError("module 'webkitcorepy.mocks' has no attribute {!r}".format(name))
+    # Resolve through CallByNeed and store the real object in globals(), so future
+    # accesses skip __getattr__ and so pickle/isinstance see the underlying class
+    # rather than the proxy.
+    value = proxy.value
+    globals()[name] = value
+    return value
+
+
+def __dir__():
+    return sorted(set(globals()) | set(_LAZY))
--- a/Tools/Scripts/libraries/webkitcorepy/webkitcorepy/partial_proxy.py
+++ b/Tools/Scripts/libraries/webkitcorepy/webkitcorepy/partial_proxy.py
@@ -20,9 +20,12 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-from unittest import mock
+import importlib
 
 from webkitcorepy import mocks
+from webkitcorepy.call_by_need import CallByNeed
+
+mock = CallByNeed(lambda: importlib.import_module('unittest.mock'))
 
 
 class PartialProxy(mocks.ContextStack):
--- a/Tools/Scripts/libraries/webkitcorepy/webkitcorepy/timeout.py
+++ b/Tools/Scripts/libraries/webkitcorepy/webkitcorepy/timeout.py
@@ -22,6 +22,7 @@
 
 import bisect
 import collections
+import importlib
 import math
 import os
 import signal
@@ -29,6 +30,9 @@ import threading
 import time
 
 from webkitcorepy import log, string_utils
+from webkitcorepy.call_by_need import CallByNeed
+
+mock = CallByNeed(lambda: importlib.import_module('unittest.mock'))
 
 ORIGINAL_SLEEP = time.sleep
 
--- a/Tools/Scripts/webkitpy/layout_tests/controllers/layout_test_finder_legacy.py
+++ b/Tools/Scripts/webkitpy/layout_tests/controllers/layout_test_finder_legacy.py
@@ -27,6 +27,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import errno
+import functools
 import logging
 import re
 
@@ -60,6 +61,16 @@ class LayoutTestFinder(object):
         self._filesystem = self._port.host.filesystem
         self.LAYOUT_TESTS_DIRECTORY = 'LayoutTests'
         self._test_list_expectations = []
+        # Cache the inner finder per device_type. Bound here, not module-level, so the
+        # cache doesn't outlive this LayoutTestFinder instance.
+        self._inner_finder = functools.cache(self._make_inner_finder)
+
+    def _make_inner_finder(self, device_type=None):
+        return LayoutTestFinder_New(
+            self._port.host.filesystem,
+            self._port.layout_tests_dir(),
+            self._port.baseline_search_path(device_type),
+        )
 
     def find_tests(self, options, args, device_type=None, with_expectations=False):
         paths = self._strip_test_dir_prefixes(args)
@@ -95,30 +106,16 @@ class LayoutTestFinder(object):
 
     def find_tests_by_path(self, paths, device_type=None, with_expectations=False):
         """Return the list of tests found. Both generic and platform-specific tests matching paths should be returned."""
-        finder = LayoutTestFinder_New(
-            self._port.host.filesystem,
-            self._port.layout_tests_dir(),
-            self._port.baseline_search_path(device_type),
-        )
-
-        return list(finder.get_tests(paths))
+        return list(self._inner_finder(device_type).get_tests(paths))
 
     def _is_test_file(self, filesystem, dirname, filename):
-        finder = LayoutTestFinder_New(
-            self._port.host.filesystem,
-            self._port.layout_tests_dir(),
-            self._port.baseline_search_path(),
-        )
+        finder = self._inner_finder()
         return finder.is_test_file(
             dirname, filename
         ) and not self._is_w3c_resource_file(filesystem, dirname, filename)
 
     def _is_w3c_resource_file(self, filesystem, dirname, filename):
-        finder = LayoutTestFinder_New(
-            self._port.host.filesystem,
-            self._port.layout_tests_dir(),
-            self._port.baseline_search_path(),
-        )
+        finder = self._inner_finder()
 
         if dirname:
             dirname = self._filesystem.relpath(dirname, self._port.layout_tests_dir())
--- a/Tools/Scripts/webkitpy/layout_tests/controllers/manager.py
+++ b/Tools/Scripts/webkitpy/layout_tests/controllers/manager.py
@@ -345,12 +345,16 @@ class Manager(object):
         new_test_inputs = self._multiply_test_inputs(test_inputs, self._options.repeat_each, self._options.iterations)
         worker_count = self._runner.get_worker_count(new_test_inputs, int(self._options.child_processes))
         self._options.child_processes = worker_count
+        # Port has a copy of options (see _create_port_for_driver), so sync the effective
+        # worker count for code paths that read it via self.get_option (mac warmup gate,
+        # embedded port device count).
+        self._port.set_option('child_processes', worker_count)
 
     def _set_up_run(self, test_inputs, device_type):
-        # This must be started before we check the system dependencies,
-        # since the helper may do things to make the setup correct.
+        # The helper was kicked off in run() so its ~440 ms warmup overlaps with
+        # _collect_tests; join on it now since check_sys_deps and setup_test_run may use it.
         self._printer.write_update("Starting helper ...")
-        if not self._port.start_helper(pixel_tests=self._options.pixel_tests, prefer_integrated_gpu=self._options.prefer_integrated_gpu):
+        if not self._port.wait_for_helper_ready():
             return False
 
         self._update_worker_count(test_inputs)
@@ -390,6 +394,10 @@ class Manager(object):
 
             self._results_directory = self._port.results_directory()
             self._finder = LayoutTestFinder(self._port, self._options)
+
+            # Kick the helper off in parallel with test discovery and expectations
+            # parsing; _set_up_run will join on it. Saves ~440 ms on Mac.
+            self._port.start_helper_async(pixel_tests=self._options.pixel_tests, prefer_integrated_gpu=self._options.prefer_integrated_gpu)
 
             device_type_list = self._port.supported_device_types()
             tests_to_run_by_device, aggregate_tests_to_skip = self._collect_tests(args, device_type_list, driver_name)
@@ -517,9 +525,6 @@ class Manager(object):
                 if not self._set_up_run(test_inputs, device_type=device_type):
                     return test_run_results.RunDetails(exit_code=-1)
 
-                configuration = self._port.configuration_for_upload(self._port.target_host(0))
-                if not configuration.get('flavor', None):  # The --result-report-flavor argument should override wk1/wk2
-                    configuration['flavor'] = 'wk1' if self._port.is_webkitlegacy() else 'wk2'
                 temp_initial_results, temp_retry_results, temp_enabled_pixel_tests_in_retry = self._run_test_subset(test_inputs, device_type=device_type)
 
                 skipped_results = TestRunResults(self._expectations[(self._current_driver_name, device_type)], len(aggregate_tests_to_skip))
@@ -530,6 +535,9 @@ class Manager(object):
                 temp_initial_results = temp_initial_results.merge(skipped_results)
 
                 if self._options.report_urls:
+                    configuration = self._port.configuration_for_upload(self._port.target_host(0))
+                    if not configuration.get('flavor', None):  # The --result-report-flavor argument should override wk1/wk2
+                        configuration['flavor'] = 'wk1' if self._port.is_webkitlegacy() else 'wk2'
                     self._printer.writeln('\n')
                     self._printer.write_update('Preparing upload data ...')
 
--- a/Tools/Scripts/webkitpy/layout_tests/models/test_expectations.py
+++ b/Tools/Scripts/webkitpy/layout_tests/models/test_expectations.py
@@ -40,6 +40,8 @@ from webkitpy.layout_tests.models.test_configuration import (
 )
 from webkitpy.port.base import Port
 
+_BUG_TOKEN_RE = re.compile(r'Bug\((\w+)\)$')
+
 MYPY = False
 if MYPY:
     # MYPY is True when running Mypy; typing is stdlib only in Py3
@@ -317,7 +319,7 @@ class TestExpectationParser(object):
         else:
             expectation_line.comment = expectation_string[comment_index + 1:]
 
-        remaining_string = re.sub(r"\s+", " ", expectation_string[:comment_index].strip())
+        remaining_string = " ".join(expectation_string[:comment_index].split())
         if len(remaining_string) == 0:
             return expectation_line
 
@@ -344,7 +346,7 @@ class TestExpectationParser(object):
                 if token.startswith(WEBKIT_BUG_PREFIX):
                     bugs.append(token.replace(WEBKIT_BUG_PREFIX, 'BUGWK'))
                 else:
-                    match = re.match(r'Bug\((\w+)\)$', token)
+                    match = _BUG_TOKEN_RE.match(token)
                     if not match:
                         warnings.append('unrecognized bug identifier "%s"' % token)
                         break
--- a/Tools/Scripts/webkitpy/port/config.py
+++ b/Tools/Scripts/webkitpy/port/config.py
@@ -54,41 +54,50 @@ class Config(object):
         "Release": "--release",
     }
 
+    # Shared across all Config instances: webkit-build-directory is a ~110ms
+    # perl invocation whose result is invariant for given inputs within a run.
+    _build_directories = {}
+
     def __init__(self, executive, filesystem, port_implementation=None, use_cmake=False):
         self._executive = executive
         self._filesystem = filesystem
         self._webkit_finder = webkit_finder.WebKitFinder(self._filesystem)
         self._default_configuration = None
-        self._build_directories = {}
         self._port_implementation = port_implementation
         self._use_cmake = use_cmake
 
+    @classmethod
+    def _clear_cache_for_testing(cls):
+        cls._build_directories = {}
+
     def build_directory(self, configuration, for_host=False):
         """Returns the path to the build directory for the configuration."""
+        port_impl = self._port_implementation if not for_host else None
+        cache_key = (port_impl, configuration or "", for_host, self._use_cmake)
+        if self._build_directories.get(cache_key):
+            return self._build_directories[cache_key]
+
         if configuration:
             flags = ["--configuration", self.flag_for_configuration(configuration)]
         else:
-            configuration = ""
             flags = []
-
-        if self._port_implementation and not for_host:
-            flags.append('--' + self._port_implementation)
-
+        if port_impl:
+            flags.append('--' + port_impl)
         if self._use_cmake:
             flags.append('--cmake')
 
-        cache_key = (configuration, self._use_cmake)
-        if not self._build_directories.get(cache_key):
-            args = ["perl", self._webkit_finder.path_to_script("webkit-build-directory")] + flags
-            output = self._executive.run_command(args, cwd=self._webkit_finder.webkit_base(), return_stderr=False).rstrip()
-            parts = output.split("\n")
-            self._build_directories[cache_key] = parts[0]
+        args = ["perl", self._webkit_finder.path_to_script("webkit-build-directory")] + flags
+        output = self._executive.run_command(args, cwd=self._webkit_finder.webkit_base(), return_stderr=False).rstrip()
+        parts = output.split("\n")
+        self._build_directories[cache_key] = parts[0]
 
-            if len(parts) == 2:
-                default_configuration = parts[1][len(parts[0]):]
-                if default_configuration.startswith("/"):
-                    default_configuration = default_configuration[1:]
-                self._build_directories[(default_configuration, self._use_cmake)] = parts[1]
+        # With no --configuration flag, webkit-build-directory also prints the
+        # default-configuration dir on a second line; cache it for later hits.
+        if len(parts) == 2:
+            default_configuration = parts[1][len(parts[0]):]
+            if default_configuration.startswith("/"):
+                default_configuration = default_configuration[1:]
+            self._build_directories[(port_impl, default_configuration, for_host, self._use_cmake)] = parts[1]
 
         return self._build_directories[cache_key]
 
--- a/Tools/Scripts/webkitpy/port/factory.py
+++ b/Tools/Scripts/webkitpy/port/factory.py
@@ -31,6 +31,7 @@
 """Factory method to retrieve the appropriate port implementation."""
 
 import fnmatch
+import importlib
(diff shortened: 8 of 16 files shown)


rr-codes:
I tried doing a similar thing for api-tests here, which ended up getting reverted, but maybe there's some parts of that change which might be helpful to roll into this change maybe? https://github.com/WebKit/WebKit/pull/60911

aproskuryakov:
I feel like this is already close to being too big, and not sure if any of these would be a speedup for this hero case. Worth looking into separately.

JonWBedard (approved):

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/mocks/__init__.py
```diff
@@ -21,10 +21,26 @@
...
+    'Subprocess': ('webkitcorepy.mocks.subprocess', 'Subprocess'),
+    'Response': ('webkitcorepy.mocks.requests_', 'Response'),
+    'Requests': ('webkitcorepy.mocks.requests_', 'Requests'),
+    'Terminal': ('webkitcorepy.mocks.terminal', 'Terminal'),
+    'FileLock': ('webkitcorepy.mocks.file_lock', 'FileLock'),
+    'Environment': ('webkitcorepy.mocks.environment', 'Environment'),
+}
+
+
+def __getattr__(name):
```
gsnedders, line 50: You should also provide `__dir__` in the module: https://docs.python.org/3/reference/datamodel.html#customizing-module-attribute-access

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/mocks/__init__.py
```diff
@@ -21,10 +21,26 @@
 # SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 from webkitcorepy.mocks.context_stack import ContextStack
```
gsnedders, line 29: Why is this _not_ included in the lazy-load set? If it's the one thing that `run-webkit-tests` needs, that doesn't seem like the strongest reason to special-case it when we have enough other code that also uses `webkitcorepy`.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/mocks/__init__.py
```diff
@@ -21,10 +21,26 @@
...
+
+
+def __getattr__(name):
+    target = _LAZY.get(name)
+    if target is None:
+        raise AttributeError("module 'webkitcorepy.mocks' has no attribute {!r}".format(name))
+    import importlib
+    value = getattr(importlib.import_module(target[0]), target[1])
+    globals()[name] = value
+    return value
```
gsnedders, line 59: Is there any reason to _not_ use `CallByNeed` here, rather than reimplementing the logic here?

Tools/Scripts/webkitpy/layout_tests/controllers/layout_test_finder_legacy.py
```diff
@@ -60,6 +60,16 @@ def __init__(self, port, options):
...
+        self._inner_finder_cache = {}
+
+    def _inner_finder(self, device_type=None):
+        if device_type not in self._inner_finder_cache:
+            self._inner_finder_cache[device_type] = LayoutTestFinder_New(
+                self._port.host.filesystem,
+                self._port.layout_tests_dir(),
+                self._port.baseline_search_path(device_type),
+            )
+        return self._inner_finder_cache[device_type]
```
gsnedders: What's actually expensive here?

Things like `self._port.layout_tests_dir()` and `self._port.baseline_search_path(device_type)` should be memoized, and the only notable thing which `LayoutTestFinder_New` does is loading the `resource-files.json` which isn't a large file and we should probably just make that lazy?

Rather than having our own `_inner_finder_cache` we should be using either `functools.cache` or `webkitcorepy.memoized` around this function?

Tools/Scripts/webkitpy/layout_tests/controllers/manager.py
```diff
@@ -345,12 +345,16 @@ def _update_worker_count(self, test_inputs):
         new_test_inputs = self._multiply_test_inputs(test_inputs, self._options.repeat_each, self._options.iterations)
         worker_count = self._runner.get_worker_count(new_test_inputs, int(self._options.child_processes))
         self._options.child_processes = worker_count
+        # Port has a copy of options (see _create_port_for_driver), so sync the effective
+        # worker count for code paths that read it via self.get_option (e.g. mac warmup gate,
+        # embedded port device count).
+        self._port.set_option('child_processes', worker_count)
```
gsnedders, line 351: Huh? `Port.set_option` just does the same as the above line; see: https://github.com/WebKit/WebKit/blob/d53865edf63b4dc3620a616e60cbd9d48c6345bc/Tools/Scripts/webkitpy/port/base.py#L709-L711

Tools/Scripts/webkitpy/port/mac.py
```diff
@@ -255,7 +255,15 @@ def default_child_processes(self, **kwargs):
         return min(supportable_instances, default_count)
 
     def start_helper(self, pixel_tests=False, prefer_integrated_gpu=False):
+        if not self.start_helper_async(pixel_tests=pixel_tests, prefer_integrated_gpu=prefer_integrated_gpu):
+            return False
+        return self.wait_for_helper_ready()
+
+    def start_helper_async(self, pixel_tests=False, prefer_integrated_gpu=False):
```
gsnedders, line 300: I don't think I expect any change here, but I think it's maybe worthwhile having a discussion:

I'd suggest making it return a `concurrent.futures.Future` rather than implementing our own logic for waiting etc., but I think @JonWBedard had strong opinions against using threads which is the obvious approach here.

Tools/Scripts/webkitpy/port/mac.py
```diff
@@ -267,9 +275,21 @@ def start_helper(self, pixel_tests=False, prefer_integrated_gpu=False):
...
-        is_ready = Port.helper.stdout.readline()
-        if not is_ready.startswith(b'ready'):
-            _log.error("LayoutTestHelper could not start")
+        Port.helper._ready = False
+        return True
+
+    def wait_for_helper_ready(self):
+        if Port.helper:
+            if getattr(Port.helper, '_ready', False):
+                return True
```
gsnedders: I really dislike the option of setting a `_ready` attribute on a different object (and class instance), in case it becomes impossible to do that; it feels a lot better to just set something on `Port`.

Tools/Scripts/webkitpy/layout_tests/models/test_expectations.py
```diff
@@ -317,7 +319,7 @@ def _tokenize_line(cls, filename, expectation_string, line_number):
         else:
             expectation_line.comment = expectation_string[comment_index + 1:]
 
-        remaining_string = re.sub(r"\s+", " ", expectation_string[:comment_index].strip())
+        remaining_string = " ".join(expectation_string[:comment_index].split())
```
gsnedders, line 322: I'm genuinely surprised this is faster, but I agree it is!

Tools/Scripts/webkitpy/port/config.py
```diff
@@ -54,38 +54,49 @@ class Config(object):
...
+    @classmethod
+    def _clear_cache_for_testing(cls):
+        cls._build_directories = {}
+
     def build_directory(self, configuration, for_host=False):
         """Returns the path to the build directory for the configuration."""
+        port_impl = self._port_implementation if not for_host else None
+        cache_key = (port_impl, configuration or "", for_host)
+        if self._build_directories.get(cache_key):
+            return self._build_directories[cache_key]
```
gsnedders: Can we just use `webkitcorepy.memoize` here? If we're caching with `self`, then we effectively have the `_port_implementation` in the cache key already?

I'd be curious about how much better the cache-hit rate is here versus just using `webkitcorepy.memoize`.

Tools/Scripts/webkitpy/port/factory.py
```diff
@@ -137,23 +149,28 @@ def get(self, port_name=None, options=None, **kwargs):
...
-            cls = module.__dict__.get('port').__dict__.get(module_name).__dict__.get(class_name)
-            if cls:
-                classes.append(cls)
+
+        for prefix, lazy_cls in self.PORT_CLASSES:
+            if not port_name.startswith(prefix):
+                continue
+            cls = lazy_cls.value
+            if not cls:
+                continue
```
gsnedders: Given this, are we actually gaining anything by using `_lazy_port` versus just doing the import here? Because the only meaningful change from the above code is the `if not port_name.startswith(prefix):` branch above, I think?

gsnedders (changes requested):
None of these are particularly large; mostly just questioning using custom lazy loading versus `CallByNeed` and custom caching versus `@webkitcorepy.memoize`.

There's also one or two questions about where exactly the time is going today.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/mocks/__init__.py
aproskuryakov: Claude tells me: 

ContextStack is the parent class for OutputCapture, OutputDuplicate, PartialProxy, all of which load eagerly via `webkitcorepy/__init__.py`.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/mocks/__init__.py
aproskuryakov: Yes, this was attempted, but broke regression tests, because it ends up in a pickle and produces a type mismatch.

Tools/Scripts/webkitpy/layout_tests/controllers/layout_test_finder_legacy.py
aproskuryakov: Claude says: The expensive bit is `_load_w3c_resource_data` (reads + parses `imported/w3c/resources/resource-files.json`) plus `baseline_search_path(device_type)` resolution.

And yes, worth converting.

Tools/Scripts/webkitpy/layout_tests/controllers/manager.py
aproskuryakov: I don't think that it's the same. `Manager` and `Port` both have an `_options` member, and it's a copy, not shared. The comment attempts to explain that, would you like it clarified in some way?

Tools/Scripts/webkitpy/port/mac.py
aproskuryakov: Right, starting a thread seems unnecessary here, and I also prefer doing without when possible.

Tools/Scripts/webkitpy/port/config.py
aproskuryakov: Claude says: `webkitcorepy.memoize` doesn't exist (I checked); they may mean `webkitpy.common.memoized.memoized`, which keys on `(self, *args)` so the cache is per-instance. That's exactly what was there before my change — and the reason multiple `Config` instances each shelled out to perl. The class-level dict is the change. As for hit rate: with the new lazy `--target` default in `factory.py`, only port-side Configs call build_directory. Single-driver Mac run = 1 Config = no benefit. Multi-driver run (e.g., wk1+wk2) = 2 Configs = saves one perl call (~110ms). Worth keeping for that case, but the reviewer is right that the win is narrower than I implied. Will note.

Tools/Scripts/webkitpy/port/factory.py
aproskuryakov: Without lazy loading, it loads all 14 port modules, which is what's being avoided here. This measured as a 150ms  win.

aproskuryakov: Perhaps I misunderstood an alternative that you are suggesting?

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/mocks/__init__.py
gsnedders: I was thinking more `return _LAZY.get(name).value` — if you explicitly return the value, you should be explicitly returning the module object, rather than the `CallByNeed` wrapper, and should be pickleable.

I do wonder if we should make `CallByNeed` pickleable, but that's really a separate task altogether. 🙂

Tools/Scripts/webkitpy/layout_tests/controllers/manager.py
gsnedders: Oh, no, I somehow misread where in the code we were, in what file.

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/mocks/__init__.py
gsnedders: Maybe add a comment?

Tools/Scripts/webkitpy/port/factory.py
gsnedders: Huh? `cls = lazy_cls.value` _does_ eagerly load them, it's the conditional that's the load-bearing part here, per my original comment.

Like sure, moving to `CallByNeed` is not unreasonable, but that's really unrelated refactoring.

Plus, we also deliberately were building up `classes` to avoid duplicating the body of the loop, which I guess is the more notable change.

Tools/Scripts/webkitpy/port/factory.py
```diff
@@ -99,26 +101,36 @@ def configuration_options():
     ]
 
 
+def _lazy_port(port_class_path):
+    def _load():
+        module_name, class_name = port_class_path.rsplit('.', 1)
+        module = __import__('webkitpy.port.{}'.format(module_name), globals(), locals(), [], 0)
+        return module.__dict__.get('port').__dict__.get(module_name).__dict__.get(class_name)
```
gsnedders: Nit, given this is just moved from where it previously lived, but isn't this equivalent?

```suggestion
        module_name, class_name = port_class_path.rsplit('.', 1)
        module = importlib.import_module(f'webkitpy.port.{module_name}')
        return getattr(module, class_name)
```

(The use of `__dict__.get` is a bit sus — because we absolutely don't handle it returning `None`.)

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/partial_proxy.py
```diff
@@ -20,9 +20,10 @@
 # OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-from unittest import mock
-
 from webkitcorepy import mocks
+from webkitcorepy.call_by_need import CallByNeed
+
+mock = CallByNeed(lambda: __import__('unittest.mock', fromlist=['_']))
```
gsnedders: I was wondering why the `fromlist=['_']`, but per the [docs](https://docs.python.org/3/library/functions.html#import__):

> The fromlist gives the names of objects or submodules that should be imported from the module given by name.
> […]
> When the name variable is of the form package.module, normally, the top-level package (the name up till the first dot) is returned, not the module named by name. However, when a non-empty fromlist argument is given, the module named by name is returned.

This is the exact sort of strange edge-case that leads the docs to recommend using `importlib.import_module()` instead, which seems like a worthwhile change, giving us:

```suggestion
mock = CallByNeed(lambda: importlib.import_module('unittest.mock'))
```

(And, again, the `importlib` import should be free — because it'll have been imported at interpreter startup time, as part of loading all the import machinery.)

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/mocks/__init__.py
```diff
@@ -20,11 +20,40 @@
...
-from webkitcorepy.mocks.time_ import Time
-from webkitcorepy.mocks.subprocess import ProcessCompletion, Subprocess
 
-from webkitcorepy.mocks.requests_ import Response, Requests
-from webkitcorepy.mocks.terminal import Terminal
-from webkitcorepy.mocks.file_lock import FileLock
-from webkitcorepy.mocks.environment import Environment
+
+def _lazy(module_path, attr_name):
+    return CallByNeed(lambda: getattr(__import__(module_path, fromlist=[attr_name]), attr_name))
```
gsnedders: Given the [docs](https://docs.python.org/3/library/functions.html#import__) recommend `importlib.import_module()` instead, we should probably just use that? It has a no weird edge-cases, unlike `__import__`.

(The `importlib` import should be a near-free lookup in the `sys.modules` `dict` — because it'll have been imported at interpreter startup time, as part of loading all the import machinery.)

Tools/Scripts/libraries/webkitcorepy/webkitcorepy/timeout.py
```diff
@@ -29,6 +29,9 @@
 import time
 
 from webkitcorepy import log, string_utils
+from webkitcorepy.call_by_need import CallByNeed
+
+mock = CallByNeed(lambda: __import__('unittest.mock', fromlist=['_']))
```
gsnedders: I was wondering why the `fromlist=['_']`, but per the [docs](https://docs.python.org/3/library/functions.html#import__):

> The fromlist gives the names of objects or submodules that should be imported from the module given by name.
> […]
> When the name variable is of the form package.module, normally, the top-level package (the name up till the first dot) is returned, not the module named by name. However, when a non-empty fromlist argument is given, the module named by name is returned.

This is the exact sort of strange edge-case that leads the docs to recommend using `importlib.import_module()` instead, which seems like a worthwhile change, giving us:

```suggestion
mock = CallByNeed(lambda: importlib.import_module('unittest.mock'))
```

(And, again, the `importlib` import should be free — because it'll have been imported at interpreter startup time, as part of loading all the import machinery.)

Tools/Scripts/webkitpy/layout_tests/controllers/layout_test_finder_legacy.py
```diff
@@ -60,6 +61,18 @@ def __init__(self, port, options):
         self._filesystem = self._port.host.filesystem
         self.LAYOUT_TESTS_DIRECTORY = 'LayoutTests'
         self._test_list_expectations = []
+        # Cache the inner finder per device_type — constructing it parses
+        # imported/w3c/resources/resource-files.json (~11 ms) and resolves
+        # baseline_search_path. Bound here (not module-level) so the cache
+        # doesn't outlive this LayoutTestFinder.
+        self._inner_finder = functools.lru_cache(maxsize=None)(self._make_inner_finder)
```
gsnedders: I guess this comment is technically true, but we don't worry about this elsewhere where we use `@webkitcorepy.decorators.Memoize()` or `@webkitpy.common.memoized`!

We also require Python 3.9 nowadays, so we can just use `functools.cache` and avoid `maxsize=None` and the confusing `lru_cache` name.

gsnedders (approved):

Tools/Scripts/webkitpy/layout_tests/controllers/layout_test_finder_legacy.py
gsnedders: Filed https://bugs.webkit.org/show_bug.cgi?id=315373 and https://bugs.webkit.org/show_bug.cgi?id=315374 as a followups there.

Tools/Scripts/webkitpy/port/factory.py
gsnedders: (Happy landing with this unaddressed, per offline discussion.)

PR #64762, opened 2026-05-12
[Site Isolation] Enable same-site BFCache with cross-site iframes via UIProcess coordination

basuke:
#### 444bd55f430068ef071504ce4cebeaaae92333df
<pre>
[Site Isolation] Enable same-site BFCache with cross-site iframes via UIProcess coordination
<a href="https://bugs.webkit.org/show_bug.cgi?id=313657">https://bugs.webkit.org/show_bug.cgi?id=313657</a>
<a href="https://rdar.apple.com/175857874">rdar://175857874</a>

Reviewed by Sihui Liu.

Builds on the plumbing added in &quot;[BFCache] Plumb ShouldRestoreFromBackForwardCache
and RestoredFromBackForwardCache signals as preparation&quot; (bug 315606). Building on bug 315104
and bug 315105, enable same-site back/forward cache for pages with cross-site
iframes under Site Isolation. The UIProcess orchestrates iframe-process
suspend and restore in parallel with the main WebProcess, gated by the
MultiProcessBackForwardCacheEnabled feature flag.

On cache, the UIProcess walks its live frame tree, dispatches
SuspendWithFrameItem to each iframe process, and detaches the iframe
WebFrameProxy subtree into the cache entry.

On restore, the UIProcess is the sole driver of the take action for every
in-process BFCache entry — not just orchestrated SI multi-process entries.
WebPageProxy::goToBackForwardItem takes the UI-side mirror synchronously
for any non-suspended-page cache entry that the UIProcess tracks, removes
the forward-list entry, and sets shouldRestoreFromBackForwardCache =
ShouldRestoreFromBackForwardCache::Yes on GoToBackForwardItemParameters
to signal the target WebProcess that the take happened upstream. The
WebProcess restores from cache only when its own
BackForwardCache::singleton().get() lookup succeeds; ::Yes with no cached
page logs an error and falls through to a normal load. Code paths without
an authoritative UIProcess decision (e.g. unsuspend via SuspendedPageProxy,
JavaScript history.back() inside the WebProcess, WebKitLegacy WebView
back/forward) pass ::Unspecified, which preserves the legacy
hasCachedPage-driven restore semantics. ::No is reserved for explicit
suppression — when received together with a cached page, FrameLoader
evicts the orphan WP-side entry to keep UIProcess and WebProcess in sync.

For orchestrated SI entries (cachedChildren non-empty) the inner block also
dispatches RestoreWithFrameItem to each iframe process in parallel and
reattaches the iframe subtree on main-frame commit, gated by the new
RestoredFromBackForwardCache parameter on DidCommitLoadForFrame so cache
misses do not reattach onto a fresh-loaded page. Aggregator failure
triggers reload(ExpiredOnly). WebBackForwardCacheEntry::takeForRestoration()
returns the cached children together with the deduplicated iframe
processes in a single atomic call, eliminating the order-dependence
between collecting processes and consuming the children. Live-tree
iframe-process enumeration in didCacheBackForwardItem goes through a new
WebPageProxy::activeRemoteFrameProcesses() helper that returns the
de-duplicated set of remote frame processes hosting iframes.

didTakeBackForwardItemForRestoration (WP→UI) is retained as a dual-path
fallback notification. Its UIProcess handler bails via the existing
`if (!item-&gt;backForwardCacheEntry()) return;` guard once UI has taken the
entry, but its IPC fire-site in commitProvisionalLoad is load-bearing for
several layout-test scenarios (multiple-back-forward-navigations,
page-cache-navigate-during-restore, main-resource-delegates-on-back-navigation,
websocket BFCache CCNS tests), so the path is intentionally preserved.

The destructor-suppression contract for UI-driven take — takeForRestoration()
and takeSuspendedPage() both call markAsTakenForRestoration() to clear
m_backForwardFrameItemID so ~WebBackForwardCacheEntry skips its
ClearCachedPage IPC and leaves the WP-side cached Page intact for the actual
restore in commitProvisionalLoad — is documented at the destructor site in
WebBackForwardCacheEntry.cpp.

A race guard in didCacheBackForwardItem (review feedback) covers the case
where the WebProcess fires DidCacheBackForwardItem when its outgoing page
enters the local back/forward cache, but the user navigates back to the
same item before the IPC arrives. The WebProcess has (or will) consume
its own cache for the restoration, so creating a UI-side mirror would be
stale. The guard skips cache-entry creation when the item URL matches
pageLoadState.pendingAPIRequestURL(), which is set when the UI dispatches
a goToBackForwardItem and cleared once the WebProcess starts the
provisional load. A currentItem-based check is not used: in the common
forward-navigation flow addIfCacheable runs before BackForwardAddItem in
commitProvisionalLoad, so when DidCacheBackForwardItem arrives the
outgoing (cached) item is still the current item — that check would
falsely reject every legitimate cache add.

Tests:
- MultiProcessBFCacheSameSiteWithCrossSiteIframe[MultipleCycles]: main
  and iframe markers survive goBack/goForward.
- MultiProcessBFCacheSameSiteWithDifferentCrossSiteIframes: cached a1
  iframe does not pollute a2&apos;s live tree.

* Source/WebCore/history/BackForwardCache.cpp:
(WebCore::BackForwardCache::addIfCacheable):
* Source/WebCore/loader/FrameLoader.cpp:
(WebCore::FrameLoader::continueLoadAfterNavigationPolicy):
* Source/WebKit/UIProcess/WebBackForwardCacheEntry.cpp:
(WebKit::WebBackForwardCacheEntry::~WebBackForwardCacheEntry):
(WebKit::WebBackForwardCacheEntry::takeSuspendedPage):
(WebKit::WebBackForwardCacheEntry::iframeProcesses):
(WebKit::WebBackForwardCacheEntry::takeForRestoration):
(WebKit::WebBackForwardCacheEntry::markAsTakenForRestoration):
* Source/WebKit/UIProcess/WebBackForwardCacheEntry.h:
(WebKit::WebPageProxy::goToBackForwardItem):
(WebKit::WebPageProxy::didCommitLoadForFrame):
(WebKit::WebPageProxy::activeRemoteFrameProcesses):
(WebKit::WebPageProxy::didCacheBackForwardItem):
(WebKit::WebPageProxy::didTakeBackForwardItemForRestoration):
(WebKit::WebPageProxy::discardBackForwardCacheEntry):
* Source/WebKit/UIProcess/WebPageProxyInternals.h:
(TEST(SiteIsolation, MultiProcessBFCacheSameSiteWithCrossSiteIframe)):
(TEST(SiteIsolation, MultiProcessBFCacheSameSiteWithCrossSiteIframeMultipleCycles)):
(TEST(SiteIsolation, MultiProcessBFCacheSameSiteWithDifferentCrossSiteIframes)):

Canonical link: <a href="https://commits.webkit.org/314050@main">https://commits.webkit.org/314050@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/5b8612e5b42cbc7bcd983f28d372d520314a45c5

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/164971 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/38462 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/32039 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/174013 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/122227 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/6ad498f5-9876-4fda-8465-ca0f93607b77) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/166839 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/38796 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/38463 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/128198 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/90238 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/680e01a8-5eb5-4c6d-8f6c-8f52e9f70f6d) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/167930 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/30501 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/148236 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/108724 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/52c6fa56-000b-48c6-8df1-75cad25d01da) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/29552 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/28166 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/21768 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/139447 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/25938 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/176536 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/29388 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/27642 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/136517 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/38530 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/32302 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/136463 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36776 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/39220 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/147734 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/101508 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/31735 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/24599 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/37730 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/110801 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/37127 "Built successfully") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/2674 "") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/37373 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/37271 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/loader/FrameLoader.cpp
+++ b/Source/WebCore/loader/FrameLoader.cpp
@@ -4132,7 +4132,7 @@ void FrameLoader::executeJavaScriptURL(const URL& url, const NavigationAction& a
     m_quickRedirectComing = false;
 }
 
-void FrameLoader::continueLoadAfterNavigationPolicy(const ResourceRequest& request, const FormSubmission* formSubmission, NavigationPolicyDecision navigationPolicyDecision, AllowNavigationToInvalidURL allowNavigationToInvalidURL, ShouldRestoreFromBackForwardCache)
+void FrameLoader::continueLoadAfterNavigationPolicy(const ResourceRequest& request, const FormSubmission* formSubmission, NavigationPolicyDecision navigationPolicyDecision, AllowNavigationToInvalidURL allowNavigationToInvalidURL, ShouldRestoreFromBackForwardCache shouldRestoreFromBackForwardCache)
 {
     // If we loaded an alternate page to replace an unreachableURL, we'll get in here with a
     // nil policyDataSource because loading the alternate page will have passed
@@ -4253,12 +4253,19 @@ void FrameLoader::continueLoadAfterNavigationPolicy(const ResourceRequest& reque
 
     if (isBackForwardLoadType(type)) {
         CheckedRef diagnosticLoggingClient = protect(frame->page())->diagnosticLoggingClient();
-        if (RefPtr provisionalItem = history().provisionalItem(); provisionalItem && BackForwardCache::singleton().get(*provisionalItem, protect(frame->page()).get())) {
+        RefPtr provisionalItem = history().provisionalItem();
+        bool hasCachedPage = provisionalItem && BackForwardCache::singleton().get(*provisionalItem, protect(frame->page()).get());
+        if (hasCachedPage && shouldRestoreFromBackForwardCache != ShouldRestoreFromBackForwardCache::No) {
             diagnosticLoggingClient->logDiagnosticMessageWithResult(DiagnosticLoggingKeys::backForwardCacheKey(), DiagnosticLoggingKeys::retrievalKey(), DiagnosticLoggingResultPass, ShouldSample::Yes);
             loadProvisionalItemFromCachedPage();
             FRAMELOADER_RELEASE_LOG(ResourceLoading, "continueLoadAfterNavigationPolicy: can't continue loading frame because it will be loaded from cache");
             return;
         }
+        if (hasCachedPage) {
+            FRAMELOADER_RELEASE_LOG_ERROR(ResourceLoading, "continueLoadAfterNavigationPolicy: evicting orphan cached page (UIProcess said No)");
+            BackForwardCache::singleton().remove(*provisionalItem);
+        } else if (shouldRestoreFromBackForwardCache == ShouldRestoreFromBackForwardCache::Yes)
+            FRAMELOADER_RELEASE_LOG_ERROR(ResourceLoading, "continueLoadAfterNavigationPolicy: expected to restore from back/forward cache but no cached page");
         diagnosticLoggingClient->logDiagnosticMessageWithResult(DiagnosticLoggingKeys::backForwardCacheKey(), DiagnosticLoggingKeys::retrievalKey(), DiagnosticLoggingResultFail, ShouldSample::Yes);
     }
 
--- a/Source/WebKit/UIProcess/WebPageProxy.cpp
+++ b/Source/WebKit/UIProcess/WebPageProxy.cpp
@@ -1751,7 +1751,7 @@ RefPtr<API::Navigation> WebPageProxy::launchProcessForReload()
     auto publicSuffix = WebCore::PublicSuffixStore::singleton().publicSuffix(URL(currentItem->url()));
 
     // We allow stale content when reloading a WebProcess that's been killed or crashed.
-    send(Messages::WebPage::GoToBackForwardItem({ navigation->navigationID(), copyFrameStateForBackForwardNavigation(protect(currentItem->mainFrameItem())), FrameLoadType::IndexedBackForward, ShouldTreatAsContinuingLoad::No, std::nullopt, m_lastNavigationWasAppInitiated, ShouldRestoreFromBackForwardCache::No, std::nullopt, publicSuffix, { }, WebCore::ProcessSwapDisposition::None }));
+    send(Messages::WebPage::GoToBackForwardItem({ navigation->navigationID(), copyFrameStateForBackForwardNavigation(protect(currentItem->mainFrameItem())), FrameLoadType::IndexedBackForward, ShouldTreatAsContinuingLoad::No, std::nullopt, m_lastNavigationWasAppInitiated, ShouldRestoreFromBackForwardCache::Unspecified, std::nullopt, publicSuffix, { }, WebCore::ProcessSwapDisposition::None }));
 
     Ref legacyMainFrameProcess = m_legacyMainFrameProcess;
     legacyMainFrameProcess->startResponsivenessTimer();
@@ -2741,6 +2741,48 @@ RefPtr<API::Navigation> WebPageProxy::goToBackForwardItem(WebBackForwardListFram
     if (process.ptr() != m_legacyMainFrameProcess.ptr())
         WEBPAGEPROXY_RELEASE_LOG_ERROR(ProcessSwapping, "goToBackForwardItem: processForTheFrameItem selected a different process pid=%d (main frame process pid=%d), frameID=%" PRIu64, process->processID(), m_legacyMainFrameProcess->processID(), frameItem.frameID() ? frameItem.frameID()->toUInt64() : 0);
 
+    // Cross-site SuspendedPageProxy entries follow the unsuspend() path; skip them here.
+    auto shouldRestoreFromBackForwardCache = ShouldRestoreFromBackForwardCache::Unspecified;
+    if (protect(preferences())->multiProcessBackForwardCacheEnabled()) {
+        RefPtr entry = item->backForwardCacheEntry();
+        shouldRestoreFromBackForwardCache = entry ? ShouldRestoreFromBackForwardCache::Yes : ShouldRestoreFromBackForwardCache::No;
+
+        if (entry && !item->suspendedPage()) {
+            auto [cachedChildren, iframeProcesses] = entry->takeForRestoration();
+            auto itemID = item->identifier();
+
+            protect(backForwardCache())->removeEntry(*item);
+
+            if (!cachedChildren.isEmpty()) {
+                auto mainFrameItemID = item->mainFrameItem().identifier();
+
+                internals().pendingBackForwardCachedChildren.set(itemID, WTF::move(cachedChildren));
+
+                auto aggregator = MainRunLoopSuccessCallbackAggregator::create(
+                    [weakThis = WeakPtr { *this }, itemID](bool success) {
+                        if (success)
+                            return;
+                        RefPtr page = weakThis.get();
+                        if (!page)
+                            return;
+                        page->internals().pendingBackForwardCachedChildren.remove(itemID);
+                        RELEASE_LOG_ERROR(ProcessSwapping, "WebPageProxy::goToBackForwardItem: iframe restore failed, reloading");
+                        page->reload(ReloadOption::ExpiredOnly);
+                    });
+
+                for (Ref iframeProcess : iframeProcesses) {
+                    if (!protect(browsingContextGroup())->remotePageInProcess(*this, iframeProcess)) {
+                        RELEASE_LOG_ERROR(ProcessSwapping, "WebPageProxy::goToBackForwardItem: no RemotePageProxy for pid %i, signaling failure", iframeProcess->processID());
+                        aggregator->failed();
+                        continue;
+                    }
+                    RELEASE_LOG(ProcessSwapping, "WebPageProxy::goToBackForwardItem: dispatching RestoreWithFrameItem to pid %i", iframeProcess->processID());
+                    iframeProcess->sendWithAsyncReply(Messages::WebPage::RestoreWithFrameItem(mainFrameItemID), aggregator->chain(), webPageIDInProcess(iframeProcess));
+                }
+            }
+        }
+    }
+
     Ref navigation = m_navigationState->createBackForwardNavigation(process->coreProcessIdentifier(), frameItem, protect(backForwardList().currentItem()), frameLoadType);
     Ref pageLoadState = internals().pageLoadState;
     auto transaction = pageLoadState->transaction();
@@ -2749,7 +2791,7 @@ RefPtr<API::Navigation> WebPageProxy::goToBackForwardItem(WebBackForwardListFram
     process->markProcessAsRecentlyUsed();
 
     auto publicSuffix = WebCore::PublicSuffixStore::singleton().publicSuffix(URL(item->url()));
-    process->send(Messages::WebPage::GoToBackForwardItem({ navigation->navigationID(), copyFrameStateForBackForwardNavigation(frameItem), frameLoadType, ShouldTreatAsContinuingLoad::No, std::nullopt, m_lastNavigationWasAppInitiated, ShouldRestoreFromBackForwardCache::No, std::nullopt, WTF::move(publicSuffix), { }, WebCore::ProcessSwapDisposition::None }), webPageIDInProcess(process));
+    process->send(Messages::WebPage::GoToBackForwardItem({ navigation->navigationID(), copyFrameStateForBackForwardNavigation(frameItem), frameLoadType, ShouldTreatAsContinuingLoad::No, std::nullopt, m_lastNavigationWasAppInitiated, shouldRestoreFromBackForwardCache, std::nullopt, WTF::move(publicSuffix), { }, WebCore::ProcessSwapDisposition::None }), webPageIDInProcess(process));
     process->startResponsivenessTimer();
 
     return RefPtr<API::Navigation> { WTF::move(navigation) };
@@ -5878,7 +5920,7 @@ void WebPageProxy::continueNavigationInNewProcess(API::Navigation& navigation, W
                 preventProcessShutdownScope = newProcess->shutdownPreventingScope()
             ] (std::optional<PageIdentifier> pageID) mutable {
                 if (pageID)
-                    newProcess->send(Messages::WebPage::GoToBackForwardItem({ navigationID, frameState.releaseNonNull(), FrameLoadType::IndexedBackForward, shouldTreatAsContinuingLoad, std::nullopt, lastNavigationWasAppInitiated, ShouldRestoreFromBackForwardCache::No, std::nullopt, WTF::move(publicSuffix), { }, WebCore::ProcessSwapDisposition::None }), *pageID);
+                    newProcess->send(Messages::WebPage::GoToBackForwardItem({ navigationID, frameState.releaseNonNull(), FrameLoadType::IndexedBackForward, shouldTreatAsContinuingLoad, std::nullopt, lastNavigationWasAppInitiated, ShouldRestoreFromBackForwardCache::Unspecified, std::nullopt, WTF::move(publicSuffix), { }, WebCore::ProcessSwapDisposition::None }), *pageID);
             });
             return;
         }
@@ -8100,6 +8142,19 @@ void WebPageProxy::didCommitLoadForFrame(IPC::Connection& connection, FrameIdent
         }
         if (RefPtr websitePolicies = navigation->websitePolicies(); websitePolicies && !m_provisionalPage)
             m_mainFrameWebsitePolicies = websitePolicies->copy();
+    }
+
+    // Reattach iframe subtree from BFCache restore. Always drain the pending entry to avoid
+    // leaking iframe processes when the BFCache restore fails (RestoredFromBackForwardCache::No).
+    if (frame->isMainFrame() && navigation) {
+        RefPtr targetItem = navigation->targetItem();
+        if (!targetItem) {
+            if (restoredFromBackForwardCache == RestoredFromBackForwardCache::Yes)
+                WEBPAGEPROXY_RELEASE_LOG_ERROR(ProcessSwapping, "didCommitLoadForFrame: BFCache commit but navigation has no target item");
+        } else if (auto pendingChildren = internals().pendingBackForwardCachedChildren.take(targetItem->identifier()); !pendingChildren.isEmpty()) {
+            if (restoredFromBackForwardCache == RestoredFromBackForwardCache::Yes)
+                frame->adoptChildFrames(WTF::move(pendingChildren));
+        }
     }
 
     m_hasCommittedAnyProvisionalLoads = true;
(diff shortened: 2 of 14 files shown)


Source/WebCore/loader/FrameLoader.cpp
```diff
@@ -2449,7 +2449,14 @@ void FrameLoader::commitProvisionalLoad()
     if (!frame->tree().parent() && history().currentItem() && (!history().provisionalItem() || history().currentItem()->itemID() != history().provisionalItem()->itemID())) {
         // Check to see if we need to cache the page we are navigating away from into the back/forward cache.
         // We are doing this here because we know for sure that a new page is about to be loaded.
-        BackForwardCache::singleton().addIfCacheable(*protect(history().currentItem()), protect(frame->page()).get());
+        // Capture hasRemoteFrameDescendant() BEFORE addIfCacheable() because the CachedFrame
+        // constructor detaches child frames from the tree.
+        bool hasRemoteFrameDescendants = frame->tree().hasRemoteFrameDescendant();
```
szewai: Why is didCachePageWithRemoteFrames conditional on hasRemoteFrameDescendants? 
UIProcess should mirror WebProcess's BackForwardCache 1:1 — every successful addIfCacheable should produce a DidCacheItem(itemID) IPC, regardless of whether there are remote frames, so UIProcess's WebBackForwardCacheEntry lifecycle is driven by an explicit signal rather than inferred from frameState->hasCachedPage flipping. 
UIProcess can walk the tree itself if it needs to know whether iframe coordination is required.

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -18112,6 +18076,117 @@ void WebPageProxy::drawFrameToSnapshot(FrameIdentifier frameID, const IntRect& r
...
+            return;
+
+        Ref process = remotePage.siteIsolatedProcess();
+
+        // Re-verify that this RemotePageProxy's process actually hosts one of
+        // our live iframe LocalFrames. forEachRemotePage already scopes to this
+        // WebPageProxy, but a RemotePageProxy can exist without a matching
+        // LocalFrame in our frame tree (e.g. during navigation tear-down).
+        bool hasSubframe = false;
+        for (RefPtr frame = mainFrame->traverseNext().frame; frame; frame = frame->traverseNext().frame) {
```
szewai: The complexity is high as we are walking the tree for all remote pages -- it's O(remotePages × frames); while we only need to walk the tree once and collect the processes of the remote frames --O(frames).

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -18112,6 +18076,117 @@ void WebPageProxy::drawFrameToSnapshot(FrameIdentifier frameID, const IntRect& r
...
+        WEBPAGEPROXY_RELEASE_LOG(ProcessSwapping, "suspendRemoteFrames: no backForwardCacheEntry for itemID %" PUBLIC_LOG_STRING ", skipping stash", itemID.toString().utf8().data());
+        return;
+    }
+
+    // Remove the dispatched iframe processes' RemotePageProxies from the
+    // BrowsingContextGroup so forEachWebContentProcess and forEachRemotePage
+    // stop reaching these frozen processes while the page is cached.
+    Ref bcg = browsingContextGroup();
+    for (auto& remotePage : cachedRemotePages)
+        bcg->takeRemotePageInProcessForProvisionalPage(*this, remotePage->process());
```
szewai: suspendRemoteFrames takes RemotePageProxys out of BCG and restoreRemoteFramesFromBackForwardCacheEntry adds them back. I don't think this is necessary or safe.
The cached WebFrameProxy already keeps the FrameProcess alive, which keeps the RemotePageProxy in BCG alive. 
Manually taking the RemotePageProxy out and re-adding it on restore can produce a duplicate when an intervening page reuses the same iframe process: BCG sees no entry for that process, creates a new RemotePageProxy, and on restore the old one is added back, breaking the per-(WebPageProxy, process) invariant.

Source/WebKit/UIProcess/WebBackForwardList.cpp
```diff
@@ -788,8 +788,17 @@ void WebBackForwardList::backForwardUpdateItem(IPC::Connection& connection, Ref<
...
-            else if (!item->suspendedPage())
+            else if (!item->suspendedPage()) {
+                // This item is transitioning out of the BFCache. If the entry
+                // carries same-site cached iframe WebFrameProxys, reattach
+                // them to the live main frame and dispatch RestoreWithFrameItem
+                // to the corresponding iframe processes before the entry is
+                // destroyed. Doing it here (rather than from a WebProcess-sent
+                // RestoreRemoteFrames IPC) avoids the race where the entry is
+                // already gone by the time the notification reaches UIProcess.
+                webPageProxy->restoreRemoteFramesFromBackForwardCacheEntry(*item);
```
szewai: restoreRemoteFramesFromBackForwardCacheEntry is called from backForwardUpdateItem, which fires after WebProcess has already begun restoring locally. This means restore vs. web-process eviction cannot be distinguished at this point.
Can we make web process explicitly tell UI process it is restoring from cache? E.g. UI process in GoToBackForwardItem tells main frame process to load from cache, and the process could tell UI process whether the restoration succeeds or not, and if succeeds, UI process will go ahead and tell other cached processes; or not then this will be treated as a new load (and cache entry will be thrown).

szewai (commented):
The patch fixes m_childFrames pollution for pages with cross-site iframes, but UIProcess is still structured as an observer reacting to WebProcess state changes rather than as the driver.
WebProcess can initiate caching, but UIProcess should own the entry lifecycle and the cross-process coordination decisions on top of that (e.g. UI process can reply web process "I am not going to add cache or send message to other subframe process because client disables cache, and please discard your cache").

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -18190,10 +18285,48 @@ void WebPageProxy::didCacheBackForwardItem(BackForwardItemIdentifier itemID, Com
...
+
+    RefPtr entry = item->backForwardCacheEntry();
+    if (!entry) {
+        WEBPAGEPROXY_RELEASE_LOG_ERROR(ProcessSwapping, "didCacheBackForwardItem: cache entry missing after addEntry for itemID %" PUBLIC_LOG_STRING, itemID.toString().utf8().data());
+        return completionHandler(false);
+    }
+
+    Vector<std::pair<Ref<WebProcessProxy>, Ref<RemotePageProxy>>> iframeTargets;
+    forEachRemoteIframeProcess([&](WebProcessProxy& process, RemotePageProxy& remotePage) {
+        iframeTargets.append({ process, remotePage });
```
szewai: I think we only need `WebProcessProxy` to send SuspendWithFrameItem message -- we could get page id with webPageIDInProcess.
Also, the only call site of `forEachRemoteIframeProcess` collects results into a vector, so we might just add a function like `activeRemoteFrameProcess` to return Vector<Ref<WebProcessProxy>> directly.

Source/WebKit/UIProcess/WebPageProxyInternals.h
```diff
@@ -280,6 +281,9 @@ struct WebPageProxy::Internals final : WebPopupMenuProxy::Client
     WeakHashSet<WebPageProxy> m_openedPages;
     HashMap<WebCore::SleepDisablerIdentifier, std::unique_ptr<WebCore::SleepDisabler>> sleepDisablers;
 
+    // Iframe subtrees stashed in goToBackForwardItem, reattached in didCommitLoadForFrame.
+    HashMap<WebCore::BackForwardItemIdentifier, Vector<Ref<WebFrameProxy>>> pendingRestoreOrchestration;
```
szewai: Nit:`pendingRestoreOrchestration` is hard to parse — consider ·pendingBackForwardCachedChildren· or similar, so the name communicates what's stored (cached child frames awaiting reattach on commit) rather than an abstract concept.

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -2741,6 +2741,72 @@ RefPtr<API::Navigation> WebPageProxy::goToBackForwardItem(WebBackForwardListFram
     if (process.ptr() != m_legacyMainFrameProcess.ptr())
         WEBPAGEPROXY_RELEASE_LOG_ERROR(ProcessSwapping, "goToBackForwardItem: processForTheFrameItem selected a different process pid=%d (main frame process pid=%d), frameID=%" PRIu64, process->processID(), m_legacyMainFrameProcess->processID(), frameItem.frameID() ? frameItem.frameID()->toUInt64() : 0);
 
+    // UIP-driven take for in-process BFCache entries (both orchestrated SI
```
szewai: There comment density is quite high. Well-scoped and well-named functions and variables should carry most of the explanation; please try only keeping necessary comments in code. The other design choice could go to commit message.

Source/WebCore/loader/FrameLoader.cpp
```diff
@@ -4132,7 +4132,7 @@ void FrameLoader::executeJavaScriptURL(const URL& url, const NavigationAction& a
     m_quickRedirectComing = false;
 }
 
-void FrameLoader::continueLoadAfterNavigationPolicy(const ResourceRequest& request, const FormSubmission* formSubmission, NavigationPolicyDecision navigationPolicyDecision, AllowNavigationToInvalidURL allowNavigationToInvalidURL, ShouldRestoreFromBackForwardCache)
+void FrameLoader::continueLoadAfterNavigationPolicy(const ResourceRequest& request, const FormSubmission* formSubmission, NavigationPolicyDecision navigationPolicyDecision, AllowNavigationToInvalidURL allowNavigationToInvalidURL, ShouldRestoreFromBackForwardCache shouldRestoreFromBackForwardCache)
```
szewai, line 4135: Why is `ShouldRestoreFromBackForwardCache` needed? Doesn't web process always prioritize cache if it is usable?

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -2741,6 +2741,72 @@ RefPtr<API::Navigation> WebPageProxy::goToBackForwardItem(WebBackForwardListFram
...
+    // WP-side cached Page intact for the actual restore in
+    // commitProvisionalLoad. Cross-site SuspendedPageProxy entries are skipped
+    // here — they follow their own unsuspend() path.
+    //
+    // ShouldRestoreFromBackForwardCache::Yes is sent to WP so FrameLoader can
+    // drive loadProvisionalItemFromCachedPage even when its own
+    // BackForwardCache::singleton().get() lookup races and returns null —
+    // UI's take is authoritative.
+    //
+    // didTakeBackForwardItemForRestoration (WP→UI) remains as a dual-path
```
szewai: I am not sure we need to mention `didTakeBackForwardItemForRestoration` at all given there's no change.
If you want to indicate we should probably drop the function since it's redundant, then we could add a FIXME comment in corresponding place.

Source/WebCore/loader/FrameLoader.cpp
basuke: As you requested for UIP-driven BFCache restore, what UIP knows about cache should be matched with the actual state in WP. So following code that uses `shouldRestoreFromBackForwardCache` is wrong. I'll fix them.

Source/WebCore/loader/FrameLoader.cpp
```diff
@@ -4253,12 +4253,19 @@ void FrameLoader::continueLoadAfterNavigationPolicy(const ResourceRequest& reque
 
     if (isBackForwardLoadType(type)) {
         CheckedRef diagnosticLoggingClient = protect(frame->page())->diagnosticLoggingClient();
-        if (RefPtr provisionalItem = history().provisionalItem(); provisionalItem && BackForwardCache::singleton().get(*provisionalItem, protect(frame->page()).get())) {
+        RefPtr provisionalItem = history().provisionalItem();
+        bool hasCachedPage = provisionalItem && BackForwardCache::singleton().get(*provisionalItem, protect(frame->page()).get());
+        if (hasCachedPage && shouldRestoreFromBackForwardCache != ShouldRestoreFromBackForwardCache::No) {
```
szewai, line 4258: The commit message says "When that flag is set, `continueLoadAfterNavigationPolicy` calls `loadProvisionalItemFromCachedPage()` even if `BackForwardCache::singleton().get()` returns null in the WP — UI's take is authoritative." 
This doesn't  match the code: the restore branch requires `hasCachedPage` to be true, and Yes with no cached page just logs an error and falls through to a normal load. 
Could you update the commit message to reflect the actual behavior?

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -8102,6 +8136,15 @@ void WebPageProxy::didCommitLoadForFrame(IPC::Connection& connection, FrameIdent
             m_mainFrameWebsitePolicies = websitePolicies->copy();
     }
 
+    // Reattach iframe subtree from BFCache restore.
+    if (restoredFromBackForwardCache == RestoredFromBackForwardCache::Yes && frame->isMainFrame() && navigation) {
```
szewai: I think we might need to clear `pendingBackForwardCachedChildren` regardless of `restoredFromBackForwardCache` value, otherwise it's leaked.
(If RestoredFromBackForwardCache::Yes we reattach children, otherwise we just throw it away)

Source/WebCore/loader/FrameLoader.cpp
```diff
@@ -4253,12 +4253,19 @@ void FrameLoader::continueLoadAfterNavigationPolicy(const ResourceRequest& reque
...
             diagnosticLoggingClient->logDiagnosticMessageWithResult(DiagnosticLoggingKeys::backForwardCacheKey(), DiagnosticLoggingKeys::retrievalKey(), DiagnosticLoggingResultPass, ShouldSample::Yes);
             loadProvisionalItemFromCachedPage();
             FRAMELOADER_RELEASE_LOG(ResourceLoading, "continueLoadAfterNavigationPolicy: can't continue loading frame because it will be loaded from cache");
             return;
         }
+        if (hasCachedPage) {
+            FRAMELOADER_RELEASE_LOG_ERROR(ResourceLoading, "continueLoadAfterNavigationPolicy: evicting orphan cached page (UIProcess said No)");
+            BackForwardCache::singleton().remove(*provisionalItem);
```
szewai, line 4266: There is an edge case where UI sends ShouldRestoreFromBackForwardCache::No but WP has a cached page, WP evicts it. 
The intended trigger: WP fires DidCacheBackForwardItem (in flight), user immediately back-navigates, UI's entry doesn't exist yet.

In the current implementation, it seems UI process would try suspending iframes after receiving DidCacheBackForwardItem. I think we might make WebPageProxy::didCacheBackForwardItem avoid doing that if the item is the current item or there's a pending navigation to it. Otherwise it'll leave inconsistent states where UI process has a cached entry but web process already evicts it.

(Also we probably want to follow up to make use of the cached page instead of evicting it in this edge case.)

Source/WebKit/UIProcess/WebPageProxy.cpp
```diff
@@ -2741,6 +2741,40 @@ RefPtr<API::Navigation> WebPageProxy::goToBackForwardItem(WebBackForwardListFram
...
+                    if (!page)
+                        return;
+                    page->internals().pendingBackForwardCachedChildren.remove(itemID);
+                    RELEASE_LOG_ERROR(ProcessSwapping, "WebPageProxy::goToBackForwardItem: iframe restore failed, reloading");
+                    page->reload(ReloadOption::ExpiredOnly);
+                });
+
+            for (Ref iframeProcess : iframeProcesses) {
+                RELEASE_LOG(ProcessSwapping, "WebPageProxy::goToBackForwardItem: dispatching RestoreWithFrameItem to pid %i", iframeProcess->processID());
+                iframeProcess->sendWithAsyncReply(Messages::WebPage::RestoreWithFrameItem(mainFrameItemID), aggregator->chain(), webPageIDInProcess(iframeProcess));
```
szewai: We might need to sanity check if `RemotePageProxy` still exists in the process, otherwise the message might be dropped by the receiver (since it does not have `WebPage` for processing `RestoreWithFrameItem`) and UI process might be stuck:
E.g. 
```
if (!browsingContextGroup().remotePageInProcess(*this, iframeProcess))
          aggregator->failed();
```

szewai (approved):
The commit message says "replacing processes(), which previously returned main + iframes", but the PR does not touch `processes()`. Please update commit message to reflect the actual changes.

PR #64773, opened 2026-05-12
AX: Cache unignored children on AXIsolatedTree to eliminate redundant subtree walks

twilco:
#### 15c92bda16a3291a3ff5d40d140330f810280acb
<pre>
AX: Cache unignored children on AXIsolatedTree to eliminate redundant subtree walks
<a href="https://bugs.webkit.org/show_bug.cgi?id=314655">https://bugs.webkit.org/show_bug.cgi?id=314655</a>
<a href="https://rdar.apple.com/176904705">rdar://176904705</a>

Reviewed by Dominic Mazzoni.

Cache stitchedUnignoredChildren() results in a HashMap&lt;AXID, Vector&lt;Ref&lt;AXCoreObject&gt;&gt;&gt;
owned by AXIsolatedTree. The cache is populated lazily on first access and cleared
conditionally in applyPendingChangesFromSnapshot when the incoming snapshot carries a
structural change (appends, subtreeRemovals, childrenUpdates, parentUpdates) or a property
update that affects the unignored-children walk (IsIgnored, IsExposableTable, StitchGroups).

Based on logging I added while testing this change, the hit rate is extremely good:
75% on one webpage, 90% on another.

Computing unignored children is by far our most expensive operation, made worse by the introduction
of text stitching which adds more work on top, so saving work here is extremely valuable.
On a large GitHub page, holding VO-Right to move to a pre-defined spot in the middle of the page took
1min 11s before this change, and 18 seconds faster (a 3.94x speedup).

This is especially beneficial because within the scope of a single AppKit selector, AppKit may call
multiple methods on our wrapper that used to each compute the unignored children from scratch.

Additional optimizations built on the cache:

  - stitchedUnignoredChildrenCount(): returns count from cache without copying the vector.

  - cachedStitchedUnignoredChildren(): returns a pointer to the cached vector for callers
    that can iterate without copying (accessibilityIndexOfChild, handleRowsAttribute,
    handleVisibleRowsAttribute, hasUnignoredChild).

  - crossFrameUnignoredChildrenInRange(): builds only the requested subrange from the cache
    with per-element cross-frame patching, used by _accessibilityChildrenFromIndex:. This
    used to build the entire children and then pick the subrange.

  - accessibilityArrayAttributeCount: now handles AXChildrenInNavigationOrder alongside
    AXChildren, avoiding a fallthrough to AppKit&apos;s default implementation that rebuilds
    the full children NSArray just to count it. This showed up in samples.

(WebCore::AXCoreObject::tabChildren):
(WebCore::AXCoreObject::unignoredChildren):
(WebCore::AXCoreObject::hasUnignoredChild):
(WebCore::AXCoreObject::stitchedUnignoredChildrenCount):
(WebCore::AXCoreObject::crossFrameUnignoredChildrenInRange):
(WebCore::AXCoreObject::selectedRadioButton):
(WebCore::AXCoreObject::selectedTabItem):
(WebCore::AXCoreObject::selectedChildren):
(WebCore::AXCoreObject::listboxSelectedChildren):
(WebCore::AXCoreObject::selectedListItems):
(WebCore::AXCoreObject::ariaTreeRows):
(WebCore::AXCoreObject::columnHeader):
(WebCore::AXCoreObject::rowHeader):
(WebCore::AXCoreObject::appendRadioButtonDescendants const):
(WebCore::AXCoreObject::cachedStitchedUnignoredChildren):
(WebCore::AXCoreObject::unignoredChildrenFromCacheIfAvailable):
(WebCore::AXIsolatedObject::ensureCachedStitchedUnignoredChildren):
(WebCore::AXIsolatedObject::stitchedUnignoredChildren):
(WebCore::AXIsolatedObject::stitchedUnignoredChildrenCount):
(WebCore::AXIsolatedObject::cachedStitchedUnignoredChildren):
(WebCore::AXIsolatedObject::crossFrameUnignoredChildrenInRange):
(WebCore::AXIsolatedObject::approximateHitTest const):
(WebCore::AXIsolatedObject::relativeFrame const):
(WebCore::AXIsolatedObject::relativeFrameFromChildren const):
(WebCore::AXIsolatedObject::tableHeaderContainer):
* Source/WebCore/accessibility/isolatedtree/AXIsolatedObject.h:
(WebCore::AXIsolatedTree::applyPendingChangesFromSnapshot):
(WebCore::AXIsolatedTree::lastMarker):
(WebCore::AXIsolatedTree::cachedStitchedUnignoredChildren):
(handleRowsAttribute):
(handleVisibleRowsAttribute):
(-[WebAccessibilityObjectWrapper accessibilityIndexOfChild:]):
(-[WebAccessibilityObjectWrapper accessibilityArrayAttributeCount:]):
(-[WebAccessibilityObjectWrapper _accessibilityChildrenFromIndex:maxCount:returnPlatformElements:]):

Canonical link: <a href="https://commits.webkit.org/313177@main">https://commits.webkit.org/313177@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/3faa9f7c5282d5ba700e7e0f8635d309f91a89f9

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/162230 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/35706 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/28822 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/171138 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/118603 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/164099 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/35810 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/35707 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/125902 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/88757 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/165189 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/28240 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/145730 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/106480 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/27287 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/25801 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/18859 "Built successfully") | | 
| [⏳ 🛠 🧪 jsc ](https://ews-build.webkit.org/#/builders/JSC-Tests-EWS "Waiting in queue, processing has not started yet") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/136989 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/23469 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/173632 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/20985 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/25112 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/134199 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/35380 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/29889 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/134200 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36248 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/35363 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/145265 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/97577 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/28741 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/22094 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/34873 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/102625 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/34374 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/34619 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/34522 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/accessibility/AXCoreObject.h
+++ b/Source/WebCore/accessibility/AXCoreObject.h
@@ -1115,7 +1115,7 @@ public:
 
 #if ENABLE(INCLUDE_IGNORED_IN_CORE_AX_TREE)
     bool onlyAddsUnignoredChildren() const { return isTableColumn() || role() == AccessibilityRole::TableHeaderContainer; }
-    AccessibilityChildrenVector unignoredChildren(bool updateChildrenIfNeeded = true);
+    virtual AccessibilityChildrenVector unignoredChildren(bool updateChildrenIfNeeded = true);
     bool hasUnignoredChild();
 #else
     const AccessibilityChildrenVector& unignoredChildren(bool updateChildrenIfNeeded = true) LIFETIME_BOUND { return children(updateChildrenIfNeeded); }
@@ -1125,7 +1125,11 @@ public:
         return !children.isEmpty();
     }
 #endif // ENABLE(INCLUDE_IGNORED_IN_CORE_AX_TREE)
-    AccessibilityChildrenVector stitchedUnignoredChildren();
+    virtual AccessibilityChildrenVector stitchedUnignoredChildren();
+    virtual size_t stitchedUnignoredChildrenCount();
+    virtual const AccessibilityChildrenVector* cachedUnignoredChildren() { return nullptr; }
+    virtual const AccessibilityChildrenVector* cachedStitchedUnignoredChildren() { return nullptr; }
+    virtual AccessibilityChildrenVector crossFrameUnignoredChildrenInRange(size_t start, size_t maxCount);
 
     virtual bool isBlockFlow() const { return false; }
     bool hasStitchableRole() const
--- a/Source/WebCore/accessibility/isolatedtree/AXIsolatedObject.cpp
+++ b/Source/WebCore/accessibility/isolatedtree/AXIsolatedObject.cpp
@@ -323,6 +323,140 @@ const AXCoreObject::AccessibilityChildrenVector& AXIsolatedObject::children(bool
     return m_children;
 }
 
+AXIsolatedTree::CachedUnignoredChildren& AXIsolatedObject::ensureCachedUnignoredChildren()
+{
+    auto& cache = tree().cachedUnignoredChildrenMap();
+    auto result = cache.ensure(objectID(), [&] {
+        auto children = AXCoreObject::unignoredChildren();
+        bool hasPotentialStitchable = false;
+        for (const auto& child : children) {
+            if (child->hasStitchableRole()) {
+                hasPotentialStitchable = true;
+                break;
+            }
+        }
+        return AXIsolatedTree::CachedUnignoredChildren { WTF::move(children), hasPotentialStitchable };
+    });
+    return result.iterator->value;
+}
+
+#if ENABLE(INCLUDE_IGNORED_IN_CORE_AX_TREE)
+AXCoreObject::AccessibilityChildrenVector AXIsolatedObject::unignoredChildren(bool)
+{
+    AX_ASSERT(!isMainThread());
+    return ensureCachedUnignoredChildren().children;
+}
+#endif
+
+AXCoreObject::AccessibilityChildrenVector AXIsolatedObject::stitchedUnignoredChildren()
+{
+    AX_ASSERT(!isMainThread());
+    auto& entry = ensureCachedUnignoredChildren();
+    if (!entry.hasPotentialStitchable)
+        return entry.children;
+    auto copy = entry.children;
+    copy.removeAllMatching([] (const auto& child) {
+        if (!child->hasStitchableRole())
+            return false;
+        std::optional stitchedIntoID = child->stitchedIntoID();
+        return stitchedIntoID && *stitchedIntoID != child->objectID();
+    });
+    return copy;
+}
+
+size_t AXIsolatedObject::stitchedUnignoredChildrenCount()
+{
+    AX_ASSERT(!isMainThread());
+    auto& entry = ensureCachedUnignoredChildren();
+    if (!entry.hasPotentialStitchable)
+        return entry.children.size();
+    size_t count = 0;
+    for (const auto& child : entry.children) {
+        if (!child->hasStitchableRole()) {
+            ++count;
+            continue;
+        }
+        std::optional stitchedIntoID = child->stitchedIntoID();
+        if (!stitchedIntoID || *stitchedIntoID == child->objectID())
+            ++count;
+    }
+    return count;
+}
+
+const AXCoreObject::AccessibilityChildrenVector* AXIsolatedObject::cachedUnignoredChildren()
+{
+    AX_ASSERT(!isMainThread());
+    return &ensureCachedUnignoredChildren().children;
+}
+
+const AXCoreObject::AccessibilityChildrenVector* AXIsolatedObject::cachedStitchedUnignoredChildren()
+{
+    AX_ASSERT(!isMainThread());
+    auto& entry = ensureCachedUnignoredChildren();
+    if (!entry.hasPotentialStitchable)
+        return &entry.children;
+    return nullptr;
+}
+
+AXCoreObject::AccessibilityChildrenVector AXIsolatedObject::crossFrameUnignoredChildrenInRange(size_t start, size_t maxCount)
+{
+    AX_ASSERT(!isMainThread());
+    auto& entry = ensureCachedUnignoredChildren();
+
+    if (entry.children.isEmpty()) {
+#if ENABLE_ACCESSIBILITY_LOCAL_FRAME
+        if (!start && maxCount) {
+            if (RefPtr crossFrameChild = crossFrameChildObject())
+                return { Ref { *crossFrameChild } };
+        }
+#endif
+        return { };
+    }
+
+    if (!entry.hasPotentialStitchable) {
+        if (start >= entry.children.size())
+            return { };
+        size_t end = std::min(start + maxCount, entry.children.size());
+
+        AccessibilityChildrenVector result;
+        result.reserveInitialCapacity(end - start);
+        for (size_t i = start; i < end; i++) {
+            Ref child = entry.children[i];
+#if ENABLE_ACCESSIBILITY_LOCAL_FRAME
+            if (RefPtr crossFrameChild = child->crossFrameChildObject())
+                result.append(crossFrameChild.releaseNonNull());
+            else
+#endif
+                result.append(WTF::move(child));
+        }
+        return result;
+    }
+
+    // Has potential stitchable children — need to skip stitched-away elements
+    // while computing the range relative to the stitched result.
+    AccessibilityChildrenVector result;
+    size_t stitchedIndex = 0;
+    for (const auto& child : entry.children) {
+        if (child->hasStitchableRole()) {
+            std::optional stitchedIntoID = child->stitchedIntoID();
+            if (stitchedIntoID && *stitchedIntoID != child->objectID())
+                continue;
+        }
+        if (stitchedIndex >= start + maxCount)
+            break;
+        if (stitchedIndex >= start) {
+#if ENABLE_ACCESSIBILITY_LOCAL_FRAME
+            if (RefPtr crossFrameChild = child->crossFrameChildObject())
+                result.append(*crossFrameChild);
+            else
+#endif
+                result.append(child);
+        }
+        ++stitchedIndex;
+    }
+    return result;
+}
+
 void AXIsolatedObject::setSelectedChildren(const AccessibilityChildrenVector& selectedChildren)
 {
     AX_ASSERT(selectedChildren.isEmpty() || selectedChildren[0]->isAXIsolatedObjectInstance());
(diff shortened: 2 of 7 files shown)


Source/WebCore/accessibility/isolatedtree/AXIsolatedObject.cpp
```diff
@@ -716,7 +774,7 @@ RefPtr<AXIsolatedObject> AXIsolatedObject::approximateHitTest(const IntPoint& po
     }
 
     AXIsolatedObject* mutableThis = const_cast<AXIsolatedObject*>(this);
-    auto children = mutableThis->unignoredChildren();
+    auto children = mutableThis->stitchedUnignoredChildren();
```
minorninth: Should this have been stitchedUnignoredChildren before semantically, or is this just because stitchedUnignoredChildren is cached?

Alternatively could you just have unignoredChildren return stitchedUnignoredChildren?

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1646,14 +1656,33 @@ void AXIsolatedTree::applyPendingChangesFromSnapshot(PendingChanges&& snapshot)
             object->setChildrenIDs(WTF::move(update.second));
     }
 
+    bool hasUnignoredChildrenAffectingPropertyChange = false;
     for (auto& change : snapshot.propertyChanges) {
         if (RefPtr object = objectForID(change.axID)) {
-            for (auto& property : change.properties)
+            for (auto& property : change.properties) {
+                if (!hasTreeStructureChange && !hasUnignoredChildrenAffectingPropertyChange) {
```
minorninth, line 1677: Can !hasTreeStructureChange come out of this loop?

Also is !hasUnignoredChildrenAffectingPropertyChange necessary since you break after setting to true?

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1646,14 +1656,33 @@ void AXIsolatedTree::applyPendingChangesFromSnapshot(PendingChanges&& snapshot)
...
+                    }
+                }
                 object->setProperty(property.first, WTF::move(property.second));
+            }
             object->shrinkPropertiesAfterUpdates();
         }
     }
 
+    if (hasUnignoredChildrenAffectingPropertyChange && !hasTreeStructureChange) {
+        // Something has changed that could effect any object's computed unignored children,
```
minorninth: nit: effect -> affect

I'm curious what percentage of calls to applyPendingChangesFromSnapshot end up clearing the cache. If it's the majority, then I wonder if it'd be cheaper or less complex to just always clear it.

Then again, maybe things like typing in an <input> or just tabbing and moving focus call applyPendingChanges but don't affect the tree structure. Maybe scrolling too.

Source/WebCore/accessibility/mac/WebAccessibilityObjectWrapperMac.mm
```diff
@@ -4409,7 +4363,14 @@ - (NSUInteger)accessibilityIndexOfChild:(id)targetChild
     if (backingObject->isTree())
         return [super accessibilityIndexOfChild:targetChild];
 
-    const auto& children = backingObject->stitchedUnignoredChildren();
+    const auto* childrenPointer = backingObject->cachedStitchedUnignoredChildren();
```
minorninth, line 4412: Why is all of this logic necessary here, can't it call stitchedUnignoredChildren and it will recompute if necessary?

Source/WebCore/accessibility/AXCoreObject.h
```diff
@@ -1115,7 +1115,21 @@ class AXCoreObject : public RefCountedAndCanMakeWeakPtr<AXCoreObject> {
         return !children.isEmpty();
     }
 #endif // ENABLE(INCLUDE_IGNORED_IN_CORE_AX_TREE)
-    AccessibilityChildrenVector stitchedUnignoredChildren();
+    virtual AccessibilityChildrenVector stitchedUnignoredChildren();
```
minorninth, line 1128: I wish we didn't have to declare these in AXCoreObject just in order to use them in AXIsolatedTree.

It feels like we should have a pure virtual interface that represents the functions needed by the platform wrappers which is implemented by both AccessibilityObject and by AXIsolatedObject, but that should be separate than the base class that both inherit from.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: > Can !hasTreeStructureChange come out of this loop?

I don't think so — we still need to process the property update. But maybe I'm misunderstanding.

> Also is !hasUnignoredChildrenAffectingPropertyChange necessary since you break after setting to true?

We break out of the switch, not the loop, so I think we do need it. It prevents us re-entering the switch unnecessarily.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: On 30s of VoiceOver usage on gmail.com, we keep the cache 1054 times, and clear it 111 times. So seems worth it!

Source/WebCore/accessibility/isolatedtree/AXIsolatedObject.cpp
twilco: It was because stitchedUnignoredChildren is cached. I considered making unignoredChildren return stitchedUnignoredChildren, but that's tricky to make work since the latter depends on the former.

I have decided to simplify this by caching unignored-children (not stitched-unignored-children), plus a flag indicating whether there may be something to stitch. This avoids a lot of the clunkiness the previous iteration of the PR had (like switching to `stitchedUnignoredChildren` simply because it was cached) without sacrificing meaningful performance, and if we build more variants of children on top of unignoredChildren, they benefit automatically.

Source/WebCore/accessibility/mac/WebAccessibilityObjectWrapperMac.mm
twilco: This saves a copy of the vector — we operate on a pointer to the vector inside the hashmap. So it's a small optimization, but open to removing it and keeping this simpler too.

Source/WebCore/accessibility/AXCoreObject.h
twilco: Agree! This seems like a bigger refactor that would bring lots of benefits outside of making this cleaner. Let's talk about a plan for making this happen.

minorninth (approved):

PR #64837, opened 2026-05-13
VP9 with transparency plays back incorrectly

stepancar:
#### d20ee4ad8b9cb36831f5647c86981c0ae353eeab
<pre>
VP9 with transparency plays back incorrectly
<a href="https://bugs.webkit.org/show_bug.cgi?id=275908">https://bugs.webkit.org/show_bug.cgi?id=275908</a>

Reviewed by NOBODY (OOPS!).

WebKit does not support WebM VP9 videos with an alpha channel: the alpha
plane was dropped by the parser and such videos played fully opaque.
This change adds support for them. Chrome and Firefox already play
these files natively.

Came across this while looking at cross-browser transparent video
playback. I&apos;d been working around it in
<a href="https://github.com/stepancar/alpha-video-kit/">https://github.com/stepancar/alpha-video-kit/</a> following Jake
Archibald&apos;s writeup (<a href="https://jakearchibald.com/2024/video-with-transparency/).">https://jakearchibald.com/2024/video-with-transparency/).</a>

Tests: media/video-with-alpha.html
       media/video-with-alpha-misaligned-keyframes.html

* LayoutTests/media/content/video-with-alpha-misaligned-keyframes.webm: Added.
* LayoutTests/media/video-with-alpha-misaligned-keyframes-expected.txt: Added.
* LayoutTests/media/video-with-alpha-misaligned-keyframes.html: Added.
* LayoutTests/platform/glib/TestExpectations:
* Source/WebCore/platform/MediaSamplesBlock.h:
* Source/WebCore/platform/cocoa/CoreVideoSoftLink.cpp:
* Source/WebCore/platform/cocoa/CoreVideoSoftLink.h:
* Source/WebCore/platform/graphics/cocoa/CMUtilities.h:
* Source/WebCore/platform/graphics/cocoa/CMUtilities.mm:
* Source/WebCore/platform/graphics/cocoa/SourceBufferParserWebM.cpp:
* Source/WebCore/platform/graphics/cocoa/VP9UtilitiesCocoa.h:
* Source/WebCore/platform/graphics/cocoa/VP9UtilitiesCocoa.mm:
* Source/WebCore/platform/graphics/cocoa/WebCoreDecompressionSession.h:
* Source/WebCore/platform/graphics/cocoa/WebCoreDecompressionSession.mm:
</pre>

https://github.com/WebKit/WebKit/commit/ec43b6be1bb6386be4de41b0de11bc37d10b0a94

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/181952 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/55899 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/49703 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/192177 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/146281 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/183814 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/57213 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/55622 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/142053 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/98622 "2 flakes 17 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/184907 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/45734 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/162787 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/122488 "Passed tests") | 
| | [❌ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/44419 "Found 1 new test failure: imported/w3c/web-platform-tests/css/css-anchor-position/scroll-to-anchored-fixed-002.html (failure)") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/42685 "Passed tests") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/154816 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/42312 "Passed tests") | [✅ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/3342 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/48530 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/46941 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/194105 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/55570 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/162283 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/151467 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/56261 "Built successfully") | [❌ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/38937 "Found 3 new test failures: http/tests/resourceLoadStatistics/dont-count-third-party-image-as-third-party-script.html http/tests/security/contentSecurityPolicy/1.1/scriptnonce-allowed-by-enforced-policy-and-blocked-by-report-policy2.py http/tests/storageAccess/deny-storage-access-under-opener-ephemeral.html (failure)") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/151171 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/45737 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/128543 "Built successfully") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/124322 "The change is no longer eligible for processing.") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/54772 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/17307 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/54153 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/54392 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/54220 "Built successfully") | | |


stepancar:
Hi @jyavenard  @jernoble @eric-carlson . Could you look at it? Thanks!

stepancar:
Gentle ping - this has been open for ~3 weeks and all EWS queues are now  green (the earlier reds were unrelated flaky).  @jyavenard @jernoble - would either of you have a chance to review? Happy to address any feedback. Thanks!

veedowen:
@stepancar does your PR effectively add alpha support for webm+vp9?

stepancar:
@veedowen Yes, correct. It adds native alpha support for VP9 in WebM (previously the alpha plane was dropped at the parser)

veedowen:
@youennf can you review or tag an appropriate reviewer?

stepancar:
Hi @veedowen @youennf, could you take a look at this PR when you get a chance? Thanks!

veedowen:
@stepancar i'm not a maintainer so i can't sign off on this, i'm just interested in following

stepancar:
@veedowen, I see, thanks!

jyavenard:
sorry, didn't get any notification. Will look now

Source/WebCore/platform/graphics/cocoa/CMUtilities.mm
```diff
@@ -555,6 +555,7 @@ static CFStringRef convertToCMYCbCRMatrix(PlatformVideoMatrixCoefficients coeffi
             // Attach HDR10+ (aka SMPTE ST 2094-40) metadata, if present:
             if (samples[i].hdrMetadataType == PlatformMediaCapabilitiesHdrMetadataType::SmpteSt209440 && samples[i].hdrMetadata)
                 CFDictionarySetValue(attachments, PAL::kCMSampleAttachmentKey_HDR10PlusPerFrameData, Ref { *samples[i].hdrMetadata }->createCFData().get());
+
```
jyavenard: unneeded

Source/WebCore/platform/graphics/cocoa/CMUtilities.mm
```diff
@@ -595,7 +596,7 @@ static CFStringRef convertToCMYCbCRMatrix(PlatformVideoMatrixCoefficients coeffi
     return adoptCF(rawSampleBuffer);
 }
 
-UniqueRef<MediaSamplesBlock> samplesBlockFromCMSampleBuffer(CMSampleBufferRef cmSample, const TrackInfo* trackInfo)
+UniqueRef<MediaSamplesBlock> samplesBlockFromCMSampleBuffer(CMSampleBufferRef cmSample, const TrackInfo* trackInfo, RefPtr<FragmentedSharedBuffer>&& alphaData)
```
jyavenard: when will this ever be a FragmentedSharedBuffer? should be a SharedBuffer if possible

jyavenard (commented):
Thank you for this.

Two issues with this:
- VideoMediaSampleRenderer  shouldn't be used for the task. It really should be using HW decoding and merge the two with vImage ; All the bricks are there. 
- there's no guarantee with webm that the two tracks (YUV / A) are aligned when it comes to keyframes. See https://bugzilla.mozilla.org/show_bug.cgi?id=1393087 for an example of a video. They are more common that I had first expected when I implemented it for firefox. 

The VideoMediaSampleRenderer job is to call WebCoreDecompressionSession as and when needed and submit it to the AVSampleBufferVideoRenderer.
So the task to decode and handle a CMSampleBuffer with alpha should be done by the WebCoreDecompressionSession itself, or an abstraction layer in between. The CMSampleBuffer could contain the alpha plane as an attachment.

For the decoding, I am unsure if the HW VP9 decoder can directly decode into RGB, may have to go through YUV first; you may prefer using vImage instead for CoreImage here 
vImageConvert_420YpCbCr8ToARGB8888 for the conversion, vImageOverwriteChannelsWithScalar_ARGB8888 to merge the two YUV/Alpha channels

jyavenard:
Actually, I think using a sample buffer backed by a CMTaggedBufferGroup may be a better fit.
It is an object specifically designed to group multiple related buffer.
Mind you the only reason we are considering this is because the VideoMediaSampleRenderer and WebCoreDecompressionSession take a CMSampleBuffer. 

If we were to replace it all and only use MediaSamplesBlock then we can do whatever.

But it would allow a more narrow integration/support in the WebCoreDecompressionSession class.

Source/WebCore/platform/graphics/cocoa/CMUtilities.mm
stepancar: Yeah, you are right, my bad

stepancar:
Hi @jyavenard! Thank you for review!
> The VideoMediaSampleRenderer job is to call WebCoreDecompressionSession as and when needed and submit it to the AVSampleBufferVideoRenderer.
So the task to decode and handle a CMSampleBuffer with alpha should be done by the WebCoreDecompressionSession itself, or an abstraction layer in between.

Yeah, okay, this would keep VideoMediaSampleRenderer clean; I agree. I considered putting this into WebCoreDecompressionSession, but self-composition didn't look great, so I moved the responsibility up a level. I will spend more time trying to move it to WebCoreDecompressionSession.

> you may prefer using vImage instead for CoreImage here
vImageConvert_420YpCbCr8ToARGB8888 for the conversion, vImageOverwriteChannelsWithScalar_ARGB8888 to merge the two YUV/Alpha channels

I wanted to avoid CPU blending to support high resolution. The vImage documentation states that it is CPU-based https://developer.apple.com/documentation/accelerate/vimage-library, which is why I sought an alternative and experimented with Core Image. @jyavenard, what is your main concern here?

stepancar:
> there's no guarantee with webm that the two tracks (YUV / A) are aligned when it comes to keyframes. See https://bugzilla.mozilla.org/show_bug.cgi?id=1393087 for an example of a video. They are more common that I had first expected when I implemented it for firefox.

Interesting, I didn't expect this. @jyavenard  Are you suggesting we stick to the Firefox approach? It seems like Chromium uses a slightly different approach

jyavenard:
> Interesting, I didn't expect this. @jyavenard Are you suggesting we stick to the Firefox approach? It seems like Chromium uses a slightly different approach

what is chrome approach ?

stepancar:
@jyavenard , Firstly, I thought it was just feeding the decoder incorrect delta frames, but I just tested it, and it throws a media error

```
{
  code: 3
  message: "PipelineStatus::PIPELINE_ERROR_DECODE: video decode error!"
}
```

stepancar:
@jyavenard https://stepancar.github.io/articles/articles/vp9-alpha-keyframe-misalignment/demo/ you can test it on this demo

Source/WebCore/platform/graphics/cocoa/CMUtilities.mm
stepancar: fixed

stepancar:
While I was testing browser differences, I noticed that the loop for WebM does not work the same way in my Safari. It seems that it is not related to my changes; it looks like it is connected to this PR https://github.com/WebKit/WebKit/pull/67171

stepancar:
> The VideoMediaSampleRenderer job is to call WebCoreDecompressionSession as and when needed and submit it to the AVSampleBufferVideoRenderer.
So the task to decode and handle a CMSampleBuffer with alpha should be done by the WebCoreDecompressionSession itself, or an abstraction layer in between. The CMSampleBuffer could contain the alpha plane as an attachment.

@jyavenard I added alphaData attachment to CMSampleBuffer, moved logic out of VideoMediaSampleRenderer, it sits in WebCoreDecompressionSession now,

stepancar:
> Actually, I think using a sample buffer backed by a CMTaggedBufferGroup may be a better fit. It is an object specifically designed to group multiple related buffer. Mind you the only reason we are considering this is because the VideoMediaSampleRenderer and WebCoreDecompressionSession take a CMSampleBuffer.
> 
> If we were to replace it all and only use MediaSamplesBlock then we can do whatever.
> 
> But it would allow a more narrow integration/support in the WebCoreDecompressionSession class.

@jyavenard , I looked at examples where we use CMTaggedBufferGroup and for me it feels like using it in alpha context is not ideal fit. 
In stere case - groups peer buffers are self-contained images that are each meaningful on their own, like the left/right views of a stereo pair.
The compressed alpha plane is the opposite: it's subordinate side data that has no meaning outside the color frame it belongs to.That puts it in the same category as HDR10+ per-frame metadata, which already travels as a sample attachment. So, attachment fits better as for me.

stepancar:
@jyavenard I think I addressed all your comments, what you suggested simplified the solution a lot comparing to my first approach, I cleaned up the PR, will be happy to hear your feedback!

LayoutTests/platform/ios/TestExpectations
```diff
@@ -329,6 +329,10 @@ http/tests/site-isolation/scrolling/scrolling-after-cross-origin-navigation.html
 # This test requires alpha-channel video support.
 compositing/video/video-background-color.html [ WontFix ]
 
+# webkit.org/b/275908 The iOS software VP9 pipeline stalls on the alpha plane and the tests time out.
```
jyavenard: this is the wrong webkit bug, can you please create a new one specifically about the issue?

Source/WebCore/platform/graphics/cocoa/WebCoreDecompressionSession.mm
```diff
@@ -581,6 +694,110 @@ static bool NODELETE isNonRecoverableError(OSStatus status)
...
+    }
+    if (!alphaSession)
+        return decodeSampleInternal(sample, flags);
+
+    // Decode directly on the current dispatcher turn so a flush cannot slip in between the alpha and color decodes.
+    Ref<DecodingPromise> alphaPromise = alphaSession->decodeSampleInternal(alphaSampleBuffer.get(), flags);
+    Ref<DecodingPromise> colorPromise = decodeSampleInternal(sample, flags);
+
+    DecodingPromise::Producer producer;
+    auto promise = producer.promise();
```
jyavenard:
```suggestion
    Ref promise = producer.promise();
```

Source/WebCore/platform/graphics/cocoa/CMUtilities.mm
```diff
@@ -606,6 +616,8 @@ static CFStringRef convertToCMYCbCRMatrix(PlatformVideoMatrixCoefficients coeffi
             if (samples[i].hdrMetadataType == PlatformMediaCapabilitiesHdrMetadataType::SmpteSt209440 && samples[i].hdrMetadata)
                 CFDictionarySetValue(attachments, PAL::kCMSampleAttachmentKey_HDR10PlusPerFrameData, Ref { *samples[i].hdrMetadata }->createCFData().get());
         }
+        if (RefPtr alphaData = samples[0].alphaData)
```
jyavenard, line 619: Why would only the first sample of the block hold the alphaData ? it is possible for multiple sample with the same timestamp to be held. Probably not a concern in practice, but this feels incorrect.

jyavenard: wore, that alphaData is used for all samples rather than just the first.

Source/WebCore/platform/graphics/cocoa/CMUtilities.mm
```diff
@@ -606,6 +616,8 @@ static CFStringRef convertToCMYCbCRMatrix(PlatformVideoMatrixCoefficients coeffi
...
+            PAL::CMSetAttachment(rawSampleBuffer, compressedAlphaDataAttachmentKey(), alphaData->createCFData().get(), kCMAttachmentMode_ShouldPropagate);
```
jyavenard, line 620: `kCMAttachmentMode_ShouldPropagate` is unlikely to be necessary as it's only designed to transfer from the demuxer to the decoder, no need to leak that sample further and not in every decoded output

Source/WebCore/platform/graphics/cocoa/WebCoreDecompressionSession.mm
```diff
@@ -405,10 +514,14 @@ static bool NODELETE isNonRecoverableError(OSStatus status)
...
-            protectedThis->decodeSampleInternal(sample.get(), flags)->chainTo(WTF::move(producer));
-        else
+        if (flushId != protectedThis->m_flushId) {
             producer.reject(noErr);
+            return;
+        }
+        if (RetainPtr alphaData = compressedAlphaData(sample.get()))
+            protectedThis->decodeSampleWithAlpha(sample.get(), alphaData.get(), flags)->chainTo(WTF::move(producer));
+        else
+            protectedThis->decodeSampleInternal(sample.get(), flags)->chainTo(WTF::move(producer));
```
jyavenard, line 575: The alpha channel will be given to the alpha WebCoreDecompressionSession, we may want to assert there that no alpha channels is given otherwise we could find ourselves with an infinite recursion

Source/WebCore/platform/graphics/cocoa/WebCoreDecompressionSession.mm
```diff
@@ -581,6 +694,110 @@ static bool NODELETE isNonRecoverableError(OSStatus status)
...
+#endif
+    RetainPtr alphaSampleBuffer = createAlphaSampleBuffer(alphaData, formatDescription.get(), PAL::CMSampleBufferGetPresentationTimeStamp(sample), PAL::CMSampleBufferGetDuration(sample), isSync);
+    if (!alphaSampleBuffer)
+        return decodeSampleInternal(sample, flags);
+
+    RefPtr<WebCoreDecompressionSession> alphaSession;
+    {
+        Locker lock { m_lock };
+        if (!m_alphaDecompressionSession && !isInvalidated())
+            m_alphaDecompressionSession = WebCoreDecompressionSession::createOpenGL(m_dispatcher.ptr());
```
jyavenard: You need to call setResourceOwner and provide the data for it, otherwise those frames will be allocated as the GPU process and it has a low memory threshold on iOS

Source/WebCore/platform/graphics/cocoa/WebCoreDecompressionSession.mm
```diff
@@ -581,6 +694,110 @@ static bool NODELETE isNonRecoverableError(OSStatus status)
...
+            int32_t height = colorPixelBuffer ? static_cast<int32_t>(CVPixelBufferGetHeight(colorPixelBuffer.get())) : 0;
+            if (!m_mergedPixelBufferPool || width != m_mergedPoolWidth || height != m_mergedPoolHeight) {
+                m_mergedPixelBufferPool = createIOSurfaceCVPixelBufferPool(width, height, kCVPixelFormatType_32BGRA, 4).value_or(nullptr);
+                m_mergedFormatDescription = nullptr;
+                m_mergedPoolWidth = width;
+                m_mergedPoolHeight = height;
+            }
+            if (!m_ciContext) {
+                m_ciContext = [CIContext contextWithOptions:nil];
+                m_alphaBlendFilter = [CIFilter filterWithName:@"CIBlendWithRedMask"];
```
jyavenard: Is this correct/intended ?

Source/WebCore/platform/graphics/cocoa/WebCoreDecompressionSession.mm
```diff
@@ -57,6 +61,105 @@
...
+    }
+
+    auto outputBuffer = createCVPixelBufferFromPool(pool);
+    if (!outputBuffer) {
+        RELEASE_LOG_ERROR(Media, "WebM alpha merge: CVPixelBufferPool allocation failed (status=%d)", int(outputBuffer.error()));
+        return nullptr;
+    }
+
+    CVBufferPropagateAttachments(colorBuffer, outputBuffer->get());
+    CVBufferSetAttachment(outputBuffer->get(), kCVImageBufferAlphaChannelModeKey, kCVImageBufferAlphaChannelMode_PremultipliedAlpha, kCVAttachmentMode_ShouldPropagate);
```
jyavenard: I'd like to understand the premultiplication here. WebM's alpha is straight, not premultiplied, but we're tagging the output PremultipliedAlpha. Is the reasoning that CIBlendWithRedMask composites over the clearColor background set on line 779, and that the result comes back premultiplied? I haven't verified that it does — if that's the intent, a comment saying so would help, since the two call sites are far enough apart that the connection isn't visible from either one.
If it turns out not to hold, every semi-transparent pixel would be wrong and we wouldn't notice: the test content is fully transparent on one half and fully opaque on the other, and those two cases look identical under either convention. Could you add a mid-range alpha — a gradient, or a third region at ~50% — and check the composited value against what you'd expect? That settles it either way, and guards us if CoreImage's behaviour shifts later.

Source/WebCore/platform/graphics/cocoa/WebCoreDecompressionSession.mm
```diff
@@ -581,6 +694,110 @@ static bool NODELETE isNonRecoverableError(OSStatus status)
...
+    DecodingPromise::Producer producer;
+    auto promise = producer.promise();
+
+    colorPromise->whenSettled(m_dispatcher, [protectedThis = Ref { *this }, this, producer = WTF::move(producer), alphaPromise = WTF::move(alphaPromise)](auto&& colorResult) mutable {
+        assertIsCurrent(m_dispatcher.get());
+        if (!colorResult) {
+            producer.reject(colorResult.error());
+            return;
+        }
+        alphaPromise->whenSettled(m_dispatcher, [protectedThis, this, producer = WTF::move(producer), colorFrames = WTF::move(*colorResult)](auto&& alphaResult) mutable {
```
jyavenard: Instead of doing the promises in series, you could use NativePromise::all() instead and let them run in paralle.

And use the completion promise to run compositeAlphaOntoColorFrames

jyavenard (commented):
looking good, some comments.

Source/WebCore/platform/graphics/cocoa/WebCoreDecompressionSession.mm
stepancar: Do you mean allSettled?

stepancar: when alpha failed I need to show frame without alpha

Source/WebCore/rendering/RenderVideo.cpp
stepancar, line 549: @jyavenard 
I was testing the PR using this video https://stepancar.github.io/alpha-video-kit/original-alpha.webm
which contains colored frames, and mask which represents a moving circle.
In my demo I had a background underneath. Sometimes I could see the circle (which means alpha was applied, but background underneeth disappeared). It was hard to reproduce, at some point, I was able to reproduce it by scrolling the page multiple times. Forcing foregroundIsKnownToBeOpaqueInRect to true solved this issue. But, honestly, I'm not sure about this solution.

Image

PR #64929, opened 2026-05-14
Web Inspector: Show a Node associated with layout and rendering events in a separate column

obyknovenius:
#### f7dd925a68af3c39406f6942a728e1428d00f802
<pre>
Web Inspector: Show a Node associated with layout and rendering events in a separate column
<a href="https://bugs.webkit.org/show_bug.cgi?id=314816">https://bugs.webkit.org/show_bug.cgi?id=314816</a>

Reviewed by Devin Rousso.

`Layout` and `Largest Contentful Paint` events in the Timeline panel
have associated Nodes. For these events, a little arrow is shown next
to the name of the event. While this lets navigate to the Node in the
Elements panel, it is not very discoverable.

This patch adds a separate column for the Node in the Timeline panel,
making it more visible and navigatable. It also adds a Node to other
layout and rendering events.

* LayoutTests/inspector/timeline/resources/timeline-event-utilities.js:
(TestPage.registerInitializer.InspectorTest.TimelineEvent.captureTimelineWithScript):
(TestPage.registerInitializer.InspectorTest.TimelineEvent.captureTimelineWithMarker):
(TestPage.registerInitializer):
* LayoutTests/inspector/timeline/timeline-layout-events-expected.txt: Added.
* LayoutTests/inspector/timeline/timeline-layout-events-iframe-expected.txt: Added.
* LayoutTests/inspector/timeline/timeline-layout-events-iframe.html: Added.
* LayoutTests/inspector/timeline/timeline-layout-events.html: Added.
(WebCore::InspectorInstrumentation::didInvalidateLayoutImpl):
(WebCore::InspectorInstrumentation::didCompositeImpl):
(WebCore::InspectorInstrumentation::didRecalculateStyleImpl):
(WebCore::InspectorInstrumentation::didScheduleStyleRecalculationImpl):
(WebCore::InspectorInstrumentation::didCompleteRenderingFrameImpl):
* Source/WebCore/inspector/InspectorInstrumentation.h:
(WebCore::InspectorInstrumentation::didInvalidateLayout):
(WebCore::InspectorInstrumentation::didComposite):
(WebCore::InspectorInstrumentation::didRecalculateStyle):
(WebCore::InspectorInstrumentation::didCompleteRenderingFrame):
* Source/WebCore/inspector/TimelineRecordFactory.cpp:
(WebCore::TimelineRecordFactory::appendLayoutRoot):
(WebCore::TimelineRecordFactory::appendNodeId):
* Source/WebCore/inspector/TimelineRecordFactory.h:
* Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp:
(WebCore::PageTimelineAgent::internalStart):
(WebCore::PageTimelineAgent::didInvalidateLayout):
(WebCore::PageTimelineAgent::didLayout):
(WebCore::PageTimelineAgent::didScheduleStyleRecalculation):
(WebCore::PageTimelineAgent::didRecalculateStyle):
(WebCore::PageTimelineAgent::didComposite):
(WebCore::PageTimelineAgent::didPaint):
(WebCore::PageTimelineAgent::didCompleteRenderingFrame):
(WebCore::PageTimelineAgent::nodeIdForDocument const):
(WebCore::PageTimelineAgent::nodeIdForRenderer const):
* Source/WebCore/inspector/agents/page/PageTimelineAgent.h:
* Source/WebCore/page/LocalFrameViewLayoutContext.cpp:
(WebCore::LocalFrameViewLayoutContext::scheduleLayout):
(WebCore::LocalFrameViewLayoutContext::scheduleSubtreeLayout):
* Source/WebInspectorUI/UserInterface/Controllers/TimelineManager.js:
(WI.TimelineManager.prototype._processRecord):
* Source/WebInspectorUI/UserInterface/Views/LayoutTimelineDataGridNode.js:
(WI.LayoutTimelineDataGridNode.prototype.get data):
(WI.LayoutTimelineDataGridNode.prototype.createCellContent):
(WI.LayoutTimelineDataGridNode):
* Source/WebInspectorUI/UserInterface/Views/LayoutTimelineView.js:
(WI.LayoutTimelineView):

Canonical link: <a href="https://commits.webkit.org/314460@main">https://commits.webkit.org/314460@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/140f05d49c80e0c8d65f86ce995e0e394394efa7

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/165895 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/39385 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/32966 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/174938 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/123150 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/167761 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/39811 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/39389 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/128932 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/90817 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/168854 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/31300 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/149040 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/109459 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/30277 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/28957 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/23082 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/140245 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/26739 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/177471 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/30377 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/28445 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/137271 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/39454 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/33104 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/137198 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/37027 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/40142 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/148534 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/102708 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/32485 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/25401 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/38653 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/111726 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/38050 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/3483 "Passed tests") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/38295 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/38193 "Built successfully") | | |

### DIFF
--- a/LayoutTests/inspector/timeline/resources/timeline-event-utilities.js
+++ b/LayoutTests/inspector/timeline/resources/timeline-event-utilities.js
@@ -5,7 +5,7 @@ function savePageData(data) {
 TestPage.registerInitializer(() => {
     InspectorTest.TimelineEvent = {};
 
-    InspectorTest.TimelineEvent.captureTimelineWithScript = function({expression, eventType, timelineType}) {
+    InspectorTest.TimelineEvent.captureTimelineWithScript = function({expression, eventType, timelineType, silent = false}) {
         let savePageDataPromise = InspectorTest.awaitEvent("SavePageData").then((event) => event.data);
 
         let promise = new Promise((resolve, reject) => {
@@ -21,11 +21,13 @@ TestPage.registerInitializer(() => {
 
                         timeline.removeEventListener(WI.Timeline.Event.RecordAdded, recordAddedListener);
 
-                        InspectorTest.log("Stopping Capture...");
+                        if (!silent)
+                            InspectorTest.log("Stopping Capture...");
                         WI.timelineManager.stopCapturing();
                     });
 
-                    InspectorTest.log("Evaluating...");
+                    if (!silent)
+                        InspectorTest.log("Evaluating...");
                     InspectorTest.evaluateInPage(expression).catch(reject);
                     return;
                 }
@@ -39,14 +41,15 @@ TestPage.registerInitializer(() => {
             });
         });
 
-        InspectorTest.log("Starting Capture...");
+        if (!silent)
+            InspectorTest.log("Starting Capture...");
         const newRecording = true;
         WI.timelineManager.startCapturing(newRecording);
 
         return promise;
     }
 
-    InspectorTest.TimelineEvent.captureTimelineWithMarker = function({expression, markerType, timelineType}) {
+    InspectorTest.TimelineEvent.captureTimelineWithMarker = function({expression, markerType, timelineType, silent = false}) {
         let savePageDataPromise = InspectorTest.awaitEvent("SavePageData").then((event) => event.data);
 
         let promise = new Promise((resolve, reject) => {
@@ -61,11 +64,13 @@ TestPage.registerInitializer(() => {
 
                         recording.removeEventListener(WI.TimelineRecording.Event.MarkerAdded, markerAddedListener);
 
-                        InspectorTest.log("Stopping Capture...");
+                        if (!silent)
+                            InspectorTest.log("Stopping Capture...");
                         WI.timelineManager.stopCapturing();
                     });
 
-                    InspectorTest.log("Evaluating...");
+                    if (!silent)
+                        InspectorTest.log("Evaluating...");
                     InspectorTest.evaluateInPage(expression).catch(reject);
                     return;
                 }
@@ -79,7 +84,8 @@ TestPage.registerInitializer(() => {
             });
         });
 
-        InspectorTest.log("Starting Capture...");
+        if (!silent)
+            InspectorTest.log("Starting Capture...");
         const newRecording = true;
         WI.timelineManager.startCapturing(newRecording);
 
--- a/Source/WebCore/inspector/TimelineRecordFactory.cpp
+++ b/Source/WebCore/inspector/TimelineRecordFactory.cpp
@@ -177,10 +177,14 @@ Ref<JSON::Object> TimelineRecordFactory::createScreenshotData(const String& imag
     return data;
 }
 
-void TimelineRecordFactory::appendLayoutRoot(JSON::Object& data, Inspector::Protocol::DOM::NodeId nodeId, const FloatQuad& quad)
+void TimelineRecordFactory::appendLayoutRoot(JSON::Object& data, const FloatQuad& quad)
+{
+    data.setArray("root"_s, createQuad(quad));
+}
+
+void TimelineRecordFactory::appendNodeId(JSON::Object& data, Inspector::Protocol::DOM::NodeId nodeId)
 {
     data.setInteger("nodeId"_s, nodeId);
-    data.setArray("root"_s, createQuad(quad));
 }
 
 } // namespace WebCore
--- a/Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
+++ b/Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
@@ -148,8 +148,10 @@ void PageTimelineAgent::internalStart(std::optional<int>&& maxCallStackDepth)
             checkedThis->pushCurrentRecord(TimelineRecordFactory::createRenderingFrameData(name), TimelineRecordType::RenderingFrame, false);
             break;
         case RunLoop::Event::DidDispatch:
-            if (checkedThis->m_startedComposite)
-                checkedThis->didComposite();
+            if (checkedThis->m_startedComposite) {
+                ASSERT(checkedThis->m_inspectedPage->localMainFrame());
+                checkedThis->didComposite(*checkedThis->m_inspectedPage->localMainFrame());
+            }
             checkedThis->didCompleteCurrentRecord(TimelineRecordType::RenderingFrame);
             break;
         }
@@ -190,9 +192,14 @@ Inspector::Protocol::ErrorStringOr<void> PageTimelineAgent::setAutoCaptureEnable
     return { };
 }
 
-void PageTimelineAgent::didInvalidateLayout()
+void PageTimelineAgent::didInvalidateLayout(const RenderElement& layoutRoot)
 {
-    appendRecord(JSON::Object::create(), TimelineRecordType::InvalidateLayout, true);
+    auto data = JSON::Object::create();
+
+    if (auto nodeId = nodeIdForRenderer(layoutRoot))
+        TimelineRecordFactory::appendNodeId(data.get(), nodeId);
+
+    appendRecord(WTF::move(data), TimelineRecordType::InvalidateLayout, true);
 }
 
 void PageTimelineAgent::willLayout()
@@ -207,30 +214,24 @@ void PageTimelineAgent::didLayout(const RenderElement& layoutRoot, const Vector<
         return;
 
     ASSERT(entry->type == TimelineRecordType::Layout);
-    ASSERT(!layoutAreas.isEmpty());
 
-    RefPtr<Node> node = layoutRoot.element();
+    if (!layoutAreas.isEmpty())
+        TimelineRecordFactory::appendLayoutRoot(entry->data.get(), layoutAreas[0]);
 
-    if (layoutRoot.isRenderView())
-        node = &layoutRoot.document();
-    else if (layoutRoot.isBeforeOrAfterContent())
-        node = layoutRoot.generatingElement();
-    else if (layoutRoot.isAnonymous())
-        node = layoutRoot.parent()->element();
-
-    Inspector::Protocol::DOM::NodeId nodeID = 0;
-    if (CheckedPtr domAgent = Ref { m_instrumentingAgents.get() }->persistentDOMAgent())
-        nodeID = domAgent->pushNodeToFrontend(node);
-
-    if (nodeID && !layoutAreas.isEmpty())
-        TimelineRecordFactory::appendLayoutRoot(entry->data.get(), nodeID, layoutAreas[0]);
+    if (auto nodeId = nodeIdForRenderer(layoutRoot))
+        TimelineRecordFactory::appendNodeId(entry->data.get(), nodeId);
 
     didCompleteCurrentRecord(TimelineRecordType::Layout);
 }
 
-void PageTimelineAgent::didScheduleStyleRecalculation()
+void PageTimelineAgent::didScheduleStyleRecalculation(Document& document)
 {
-    appendRecord(JSON::Object::create(), TimelineRecordType::ScheduleStyleRecalculation, true);
+    auto data = JSON::Object::create();
+
+    if (auto nodeId = nodeIdForDocument(document))
+        TimelineRecordFactory::appendNodeId(data.get(), nodeId);
+
+    appendRecord(WTF::move(data), TimelineRecordType::ScheduleStyleRecalculation, true);
 }
 
 void PageTimelineAgent::willRecalculateStyle()
@@ -238,8 +239,17 @@ void PageTimelineAgent::willRecalculateStyle()
     pushCurrentRecord(JSON::Object::create(), TimelineRecordType::RecalculateStyles, true);
 }
 
-void PageTimelineAgent::didRecalculateStyle()
+void PageTimelineAgent::didRecalculateStyle(Document& document)
 {
+    auto* entry = lastRecordEntry();
+    if (!entry)
+        return;
+
+    ASSERT(entry->type == TimelineRecordType::RecalculateStyles);
+
+    if (auto nodeId = nodeIdForDocument(document))
+        TimelineRecordFactory::appendNodeId(entry->data.get(), nodeId);
+
     didCompleteCurrentRecord(TimelineRecordType::RecalculateStyles);
 }
 
@@ -250,10 +260,22 @@ void PageTimelineAgent::willComposite()
     m_startedComposite = true;
 }
 
-void PageTimelineAgent::didComposite()
+void PageTimelineAgent::didComposite(const LocalFrame& frame)
 {
-    if (m_startedComposite)
+    if (m_startedComposite) {
+        auto* entry = lastRecordEntry();
+        if (!entry)
+            return;
+
+        ASSERT(entry->type == TimelineRecordType::Composite);
+
+        ASSERT(frame.document());
+        Ref document = *frame.document();
+        if (auto nodeId = nodeIdForDocument(document.get()))
+            TimelineRecordFactory::appendNodeId(entry->data.get(), nodeId);
+
         didCompleteCurrentRecord(TimelineRecordType::Composite);
+    }
     m_startedComposite = false;
 
     if (instruments().contains(Inspector::Protocol::Timeline::Instrument::Screenshot))
@@ -281,6 +303,9 @@ void PageTimelineAgent::didPaint(RenderObject& renderer, const LayoutRect& clipR
 
     auto clipQuadInRootView = protect(renderer.view().frameView())->contentsToRootView(renderer.localToAbsoluteQuad({ clipRect }));
     entry->data = TimelineRecordFactory::createPaintData(clipQuadInRootView);
+
+    if (auto nodeId = nodeIdForRenderer(renderer))
+        TimelineRecordFactory::appendNodeId(entry->data.get(), nodeId);
 
     didCompleteCurrentRecord(TimelineRecordType::Paint);
 }
@@ -317,7 +342,7 @@ void PageTimelineAgent::mainFrameNavigated()
     }
 }
 
-void PageTimelineAgent::didCompleteRenderingFrame()
+void PageTimelineAgent::didCompleteRenderingFrame(const LocalFrame& frame)
 {
 #if PLATFORM(COCOA)
     if (!tracking() || protect(environment())->debugger()->isPaused())
@@ -329,9 +354,11 @@ void PageTimelineAgent::didCompleteRenderingFrame()
         return;
 
     if (m_startedComposite)
-        didComposite();
+        didComposite(frame);
 
     didCompleteCurrentRecord(TimelineRecordType::RenderingFrame);
+#else
+    UNUSED_PARAM(frame);
 #endif
 }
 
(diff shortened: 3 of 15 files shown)


obyknovenius:
Screenshot 2026-05-14 at 15 38 33

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
```diff
@@ -148,8 +148,11 @@ void PageTimelineAgent::internalStart(std::optional<int>&& maxCallStackDepth)
             checkedThis->pushCurrentRecord(TimelineRecordFactory::createRenderingFrameData(name), TimelineRecordType::RenderingFrame, false);
             break;
         case RunLoop::Event::DidDispatch:
-            if (checkedThis->m_startedComposite)
-                checkedThis->didComposite();
+            if (checkedThis->m_startedComposite) {
+                Ref inspectedPage = checkedThis->m_inspectedPage.get();
+                if (RefPtr localMainFrame = inspectedPage->localMainFrame())
```
dcrousso: i think we can simplify this
```suggestion
                if (RefPtr localMainFrame = Ref { checkedThis->m_inspectedPage.get() }->localMainFrame())
```

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
```diff
@@ -207,39 +213,40 @@ void PageTimelineAgent::didLayout(const RenderElement& layoutRoot, const Vector<
         return;
 
     ASSERT(entry->type == TimelineRecordType::Layout);
-    ASSERT(!layoutAreas.isEmpty());
 
-    RefPtr<Node> node = layoutRoot.element();
+    if (!layoutAreas.isEmpty())
```
dcrousso, line 218: why did this change from an `ASSERT` to a runtime check?

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
```diff
@@ -207,39 +213,40 @@ void PageTimelineAgent::didLayout(const RenderElement& layoutRoot, const Vector<
...
     didCompleteCurrentRecord(TimelineRecordType::Layout);
 }
 
-void PageTimelineAgent::didScheduleStyleRecalculation()
+void PageTimelineAgent::didScheduleStyleRecalculation(Document& document)
 {
-    appendRecord(JSON::Object::create(), TimelineRecordType::ScheduleStyleRecalculation, true);
+    auto data = JSON::Object::create();
+    if (CheckedPtr domAgent = Ref { m_instrumentingAgents.get() }->persistentDOMAgent())
+        TimelineRecordFactory::appendNodeId(data.get(), domAgent->pushNodeToFrontend(&document));
```
dcrousso: unfortunately i dont think this is actually a valid thing to do

the current workflow for the DOM agent is that the frontend _must_ call `DOM.getDocument` before anything else as `InspectorDOMAgent::pushNodeToFrontend` assumes that the `Document` has already been assigned a unique `DOM.NodeId`

for more info see <[https://webkit.org/b/213499](https://webkit.org/b/213499)> (_Web Inspector: allow DOM nodes to be instrumented at any point, regardless of whether the main document has also been instrumented_)

as such, i think you'd instead need to use `InspectorDOMAgent::boundNodeId`

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
```diff
@@ -207,39 +213,40 @@ void PageTimelineAgent::didLayout(const RenderElement& layoutRoot, const Vector<
...
+void PageTimelineAgent::didRecalculateStyle(Document& document)
 {
+    auto* entry = lastRecordEntry();
+    if (!entry)
+        return;
+
+    ASSERT(entry->type == TimelineRecordType::RecalculateStyles);
+
+    if (CheckedPtr domAgent = Ref { m_instrumentingAgents.get() }->persistentDOMAgent())
+        TimelineRecordFactory::appendNodeId(entry->data.get(), domAgent->pushNodeToFrontend(&document));
```
dcrousso: ditto

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
```diff
@@ -250,10 +257,22 @@ void PageTimelineAgent::willComposite()
...
+    if (m_startedComposite) {
+        auto* entry = lastRecordEntry();
+        if (!entry)
+            return;
+
+        ASSERT(entry->type == TimelineRecordType::Composite);
+
+        if (CheckedPtr domAgent = Ref { m_instrumentingAgents.get() }->persistentDOMAgent()) {
+            if (RefPtr document = frame.document())
+                TimelineRecordFactory::appendNodeId(entry->data.get(), domAgent->pushNodeToFrontend(document.get()));
```
dcrousso: ditto

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
```diff
@@ -364,4 +388,23 @@ void PageTimelineAgent::captureScreenshot()
...
+        if (renderElement->isBeforeOrAfterContent())
+            node = renderElement->generatingElement();
+    } else if (renderer.isAnonymous())
+        node = renderer.parent()->element();
+
+    Inspector::Protocol::DOM::NodeId nodeId = 0;
+    if (CheckedPtr domAgent = Ref { m_instrumentingAgents.get() }->persistentDOMAgent())
+        nodeId = domAgent->pushNodeToFrontend(node);
+
+    return nodeId;
```
dcrousso: rather than save to a variable i think we should early `return` to simplify things
```suggestion
    if (CheckedPtr domAgent = Ref { m_instrumentingAgents.get() }->persistentDOMAgent())
        return domAgent->pushNodeToFrontend(node);

    return 0;
```
id even suggest putting the check for if the `domAgent` exists first as if that's not true (which admittedly shouldn't be possible) then none of the other logic above would actually do anything since we'd just `return 0;`

Source/WebCore/page/LocalFrameViewLayoutContext.cpp
```diff
@@ -531,7 +531,7 @@ void LocalFrameViewLayoutContext::scheduleLayout()
         LOG(Layout, "LocalFrameView %p layout timer scheduled at %.3fs", this, document->timeSinceDocumentCreation().value());
 #endif
 
-    InspectorInstrumentation::didInvalidateLayout(protect(frame()));
+    InspectorInstrumentation::didInvalidateLayout(protect(frame()), *renderView());
```
dcrousso, line 681: what guarantee is there that `renderView()` is not `nullptr`?

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
```diff
@@ -250,10 +257,22 @@ void PageTimelineAgent::willComposite()
     m_startedComposite = true;
 }
 
-void PageTimelineAgent::didComposite()
+void PageTimelineAgent::didComposite(const LocalFrame& frame)
 {
-    if (m_startedComposite)
+    if (m_startedComposite) {
+        auto* entry = lastRecordEntry();
```
dcrousso, line 266: i wonder if we should modify `didCompleteCurrentRecord` to accept a lambda that lets the caller modify the `entry` right before it calls `didCompleteRecordEntry` so that we don't have to fetch `lastRecordEntry` and verify it's `type` more than once

that could be used in a bunch of spots (e.g. here, above, `PageTimelineAgent::didPaint`, `InspectorTimelineAgent::didDispatchEvent`, etc.)

dcrousso (commented):
very cool!  i think there's only one issue that needs to be addressed

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
obyknovenius: The runtime check was there before my changes: https://github.com/WebKit/WebKit/blob/6d7499c6a698ecab5e7e00d609fa722a8384f479/Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp#L207-L209 The `ASSERT` looks redundant. But I can bring it back.

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
obyknovenius: Isn't it already handled in `pushNodeToFrontend()` (not to be confused with `pushNodePathToFrontend()`)?
https://github.com/WebKit/WebKit/blob/703a0299474b6ec7303fcd26870af6201d3529e8/Source/WebCore/inspector/agents/InspectorDOMAgent.cpp#L610-L619

Source/WebCore/page/LocalFrameViewLayoutContext.cpp
obyknovenius: Added a runtime check.

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
obyknovenius: I like the idea. I can try to do it in a follow-up PR.

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
dcrousso: technically yes this will handle the scenario where the `DOM.getDocument` hasn't been called yet as it'll fail inside `InspectorDOMAgent::assertDocument`

in the case that this is the main `m_document` then it's doing a bunch of extra work because it has to fetch the ID for it multiple times only to eventually return the already bound ID in `InspectorDOMAgent::pushNodePathToFrontend`

i suppose this does improve things for `<iframe>` since `DOM.getDocument` is only meant to get the main top-level `Document` and therefore this will push the entire path to the frontend

can we add some tests for this just to make sure (both main document and `<iframe>`)?  i think it should be as easy as making sure _not_ to invoke `WI.DOMManager.prototype.requestDocument` before starting a timeline recording (and then having the page do whatever it needs to trigger this codepath)

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
```diff
@@ -149,7 +149,9 @@ void PageTimelineAgent::internalStart(std::optional<int>&& maxCallStackDepth)
             break;
         case RunLoop::Event::DidDispatch:
             if (checkedThis->m_startedComposite)
-                checkedThis->didComposite();
+                if (RefPtr localMainFrame = checkedThis->m_inspectedPage->localMainFrame())
+                    checkedThis->didComposite(*localMainFrame);
```
dcrousso: is there a concern that this would potentially not run if `localMainFrame` is `nullptr`?  wouldn't we still want it to fire even if that's not the case? (i.e. the frontend would still show this event in the Timelines Tab but just without a node)

Source/WebCore/page/LocalFrameViewLayoutContext.cpp
```diff
@@ -554,7 +531,9 @@ void LocalFrameViewLayoutContext::scheduleLayout()
         LOG(Layout, "LocalFrameView %p layout timer scheduled at %.3fs", this, document->timeSinceDocumentCreation().value());
 #endif
 
-    InspectorInstrumentation::didInvalidateLayout(protect(frame()));
+    if (CheckedRef renderView = *this->renderView())
+        InspectorInstrumentation::didInvalidateLayout(protect(frame()), renderView.get());
```
dcrousso: ditto about us potentially always wanting to do this (i.e. maybe we just change `InspectorInstrumentation::didInvalidateLayout` to accept a `const RenderElement*` instead of a `const RenderElement&`?)

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
dcrousso: ah in that case it's fine to remove :)

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
obyknovenius: Added `PageTimelineAgent::nodeIdForDocument()` that calls `boundNodeId()` for the main document and `pushNodePathToFrontend()` for an iframe document and two new tests to verify that it works as expected.

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
obyknovenius: I don't think it can happen. Added an `ASSERT` to highlight this assumption.

Source/WebCore/page/LocalFrameViewLayoutContext.cpp
obyknovenius: Same here. Added `ASSERT`s like in `LocalFrameViewLayoutContext::scheduleSubtreeLayout()`.

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
```diff
@@ -364,4 +391,45 @@ void PageTimelineAgent::captureScreenshot()
...
+        return 0;
+
+    if (renderer.isRenderView()) {
+        RefPtr document = renderer.document();
+        if (document->isTopDocument())
+            return domAgent->boundNodeId(document);
+        return domAgent->pushNodePathToFrontend(document);
+    }
+
+    if (CheckedPtr<const RenderElement> renderElement = dynamicDowncast<RenderElement>(renderer)) {
```
dcrousso: could you add a brief explanation of what each of these scenarios represent?  it's a bit more complex and branched than i expected (though it does look pretty good)

LayoutTests/inspector/timeline/resources/timeline-event-utilities.js
```diff
@@ -5,7 +5,7 @@ function savePageData(data) {
 TestPage.registerInitializer(() => {
     InspectorTest.TimelineEvent = {};
 
-    InspectorTest.TimelineEvent.captureTimelineWithScript = function({expression, eventType, timelineType}) {
+    InspectorTest.TimelineEvent.captureTimelineWithScript = function({expression, eventType, timelineType, silent = false}) {
```
dcrousso, line 8: id strongly recommend not having the `silent` option as timelines tests are often flaky and having extra logging about the state of things are incredibly helpful to know where in the test things failed

Source/WebCore/inspector/TimelineRecordFactory.cpp
```diff
@@ -177,10 +177,14 @@ Ref<JSON::Object> TimelineRecordFactory::createScreenshotData(const String& imag
...
-void TimelineRecordFactory::appendLayoutRoot(JSON::Object& data, Inspector::Protocol::DOM::NodeId nodeId, const FloatQuad& quad)
+void TimelineRecordFactory::appendLayoutRoot(JSON::Object& data, const FloatQuad& quad)
 {
-    data.setInteger("nodeId"_s, nodeId);
     data.setArray("root"_s, createQuad(quad));
 }
 
+void TimelineRecordFactory::appendNodeId(JSON::Object& data, Inspector::Protocol::DOM::NodeId nodeId)
+{
+    data.setInteger("nodeId"_s, nodeId);
```
dcrousso, line 187: NIT: given that `Inspector::Protocol::DOM::NodeId` can be `0` i kinda feel like we should move the validity check inside this method
```cpp
if (nodeId)
    data.setInteger("nodeId"_s, nodeId);
```
instead of having it wrap every caller
```cpp
TimelineRecordFactory::appendNodeId(data.get(), nodeIdForRenderer(layoutRoot));
```

dcrousso (approved):
nice work!  thanks for iterating with me!

Source/WebCore/inspector/agents/page/PageTimelineAgent.cpp
obyknovenius: I modified this method to be more self-explanatory. It covers the following cases:
- Frames - main document frame and `<iframe>`s
- Anonymous nodes including elements
- Pseudo- and regular elements
- Regular nodes

These are all the cases I could think of, but there's still a chance that I missed something.

LayoutTests/inspector/timeline/resources/timeline-event-utilities.js
obyknovenius: I agree but  `InspectorTest` logs messages by appending a text node to the results `<pre>` element causing layout events to be fired. These events interfere with the layout events of the test itself making it fail. I couldn't find a better solution.

Source/WebCore/inspector/TimelineRecordFactory.cpp
obyknovenius: You're right, but it can be confusing for the caller that the node id returned from `nodeIdForDocument()` or `nodeIdForRenderer()` was not actually appended. I prefer to be more explicit that value `0` is invalid as node id.

obyknovenius:
@dcrousso Thank you for your review. Merging it.

PR #65035, opened 2026-05-16
[AppKit Gestures] Disambiguate drag-and-drop from text selection during press-and-drag

aprotyas:
#### 26475c7bd8ca76d414f8723243d61a7637907301
<pre>
[AppKit Gestures] Disambiguate drag-and-drop from text selection during press-and-drag
<a href="https://bugs.webkit.org/show_bug.cgi?id=314870">https://bugs.webkit.org/show_bug.cgi?id=314870</a>
<a href="https://rdar.apple.com/176383341">rdar://176383341</a>

Reviewed by Richard Robinson.

Currently, any draggable element (link, image, `draggable=&quot;true&quot;`, etc.)
is siltently ignored in favor of text selection extensions when starting
a press-and-drag interaction. This is because NSTextSelectionManager
gestures recognize before WebKit can disambiguate this interaction.

This patch builds on top of 312982@main, which ported over async gesture
gates for AppKit gesture recognizers. We use this gesture to defer the
text selection gestures until the web process has reported the hit-test
content. On response, the deferring gesture either fails and lets the
text selection gestures proceed or transitions to .Ended, excluding the
text selection gestures and letting our drag press gesture win. We allow
our drag press gesture to win when the hit test reports draggable
content the press point.

In service of this patch, we also refactor text selection drag-and-drops.
Concretely, we let WKTextSelectionController launder the gesture it is
provided by NSTextSelectionManager via setTextSelectionDragGesture:...,
whose state is then read to perform the same mouseDown/Dragged sequences
we do elsewhere (to invoke drag-and-drop).

Non-exhaustive list of important changes, all made to solve this gesture
disambiguation problem:

- WKAppKitGestureController gains a position-information cache modeled
after the one in WKContentViewInteraction. Pending handlers are paired
with the request they were enqueued for, and only fire when the arriving
info matches.

- WebViewImpl::cancelDrag() consolidates the drag-cancel cleanup
(canceling the WebProcess drag and clearing the gesture controller&apos;s
drag state) so each early-return path in WebViewImpl::startDrag clears
both consistently.

Test: AppKitGesturesTests.pressDragOverTextCreatesSelection
      AppKitGesturesTests.pressDragOnLinkInitiatesDragAndDrop
      AppKitGesturesTests.pressDragOnExistingSelectionDoesNotExtendSelection

* Source/WebKit/UIProcess/mac/PageClientImplMac.mm:
(WebKit::PageClientImpl::didCommitLoadForMainFrame):
(WebKit::PageClientImpl::positionInformationDidChange):
* Source/WebKit/UIProcess/mac/WKAppKitGestureController.h:
* Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm:
(representsDraggableElement):
(-[WKAppKitGestureController setUpGestureRecognizers]):
(-[WKAppKitGestureController setUpDragPressGestureRecognizer]):
(-[WKAppKitGestureController setUpDragDeferringGestureRecognizer]):
(-[WKAppKitGestureController addGesturesToWebView]):
(-[WKAppKitGestureController enableGesturesIfNeeded]):
(-[WKAppKitGestureController deferringGestureRecognizer:shouldDeferOtherGestureRecognizer:]):
(-[WKAppKitGestureController deferringGestureRecognizer:shouldDeferGesturesForEventThatWillBeginAction:]):
(-[WKAppKitGestureController deferringGestureRecognizer:didEndActionWithEvent:]):
(-[WKAppKitGestureController deferringGestureRecognizer:didTransitionToState:]):
(-[WKAppKitGestureController _invalidateCurrentPositionInformation]):
(-[WKAppKitGestureController didCommitLoadForMainFrame]):
(-[WKAppKitGestureController requestAsynchronousPositionInformationUpdate:]):
(-[WKAppKitGestureController doAfterPositionInformationUpdate:forRequest:]):
(-[WKAppKitGestureController _currentPositionInformationIsValidForRequest:]):
(-[WKAppKitGestureController _hasValidOutstandingPositionInformationRequest:]):
(-[WKAppKitGestureController _invokeAndRemovePendingHandlersValidForCurrentPositionInformation]):
(-[WKAppKitGestureController positionInformationDidChange:]):
(-[WKAppKitGestureController _dragPressShouldBeginAtLocation:]):
(-[WKAppKitGestureController activeDragGestureRecognizer]):
(-[WKAppKitGestureController setTextSelectionDragGesture:completionHandler:]):
(-[WKAppKitGestureController setGestureDraggingSession:]):
(-[WKAppKitGestureController clearGestureDragState]):
(-[WKAppKitGestureController reset]):
(-[WKAppKitGestureController gestureRecognizer:shouldRecognizeSimultaneouslyWithGestureRecognizer:]):
(-[WKAppKitGestureController gestureRecognizer:shouldBeRequiredToFailByGestureRecognizer:]):
(-[WKAppKitGestureController gestureRecognizerShouldBegin:]):
(-[WKAppKitGestureController _gestureRecognizer:canPreventGestureRecognizer:]):
* Source/WebKit/UIProcess/mac/WKTextSelectionController.swift:
(WKTextSelectionController.showContextMenu(at:)):
(WKTextSelectionController.dragSelection(withGesture:completionHandler:)):
(WKTextSelectionController.textSelectionDragGestureUpdated(_:)):
* Source/WebKit/UIProcess/mac/WebViewImpl.h:
* Source/WebKit/UIProcess/mac/WebViewImpl.mm:
(WebKit::WebViewImpl::cancelDrag):
(WebKit::WebViewImpl::startDrag):
(WebKit::WebViewImpl::setTextSelectionDragGesture):
(WebKit::WebViewImpl::invalidateCachedPositionInformation):
(WebKit::WebViewImpl::positionInformationDidChange):
* Tools/TestWebKitAPI/Helpers/cocoa/JavaScriptMessages.swift:
(GetSelection.expression):
  Address a minor latest issue with the selection JS helper. It is
  possible for non-editable content to have _no_ selection at all (i.e.
  anchorNode === null and rangeCount === 0). In these cases, we were
  still unconditionally accessing selection.acnhorNode.parentElement,
  which threw a JS eval exception. Instead, we want to encode this as a
  `.none` case in JavaScriptSelection, which we can then incorporate in
  the pressDragOnLinkInitiatesDragAndDrop test.
* Tools/TestWebKitAPI/Helpers/cocoa/JavaScriptTypes.swift:
(JavaScriptSelection.encoded):
* Tools/TestWebKitAPI/Tests/WebKit/WebPage/AppKitGesturesTests.swift:
(AppKitGesturesTests.pressDragOverTextCreatesSelection(_:)):
(AppKitGesturesTests.pressDragOnLinkInitiatesDragAndDrop):
(AppKitGesturesTests.pressDragOnExistingSelectionDoesNotExtendSelection):

Canonical link: <a href="https://commits.webkit.org/313504@main">https://commits.webkit.org/313504@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/61281480175000c5963bc83dbc22b1208035c4de

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/163240 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/36721 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/29938 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/172179 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/119687 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/03b833e3-14b4-4636-a001-a74a95ddc2c6) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/165109 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/37049 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/36722 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/126756 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/89352 "5 failures") | [❌ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b922fc25-ba89-4078-976c-e2801cb836f2) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/166199 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/29028 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/146753 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/107323 "Passed tests") | | [loading 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/c8b56182-f219-49a5-8b1c-ca1c5bf13a5c) 
| [loading 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/162561 "Build is in progress. Recent messages:") | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/28074 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/26702 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/19880 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/137967 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/24477 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/174682 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/26868 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/26132 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/135061 "Passed tests") | | 
| [⏳ 🧪 services ](https://ews-build.webkit.org/#/builders/Services-EWS "Waiting in queue, processing has not started yet") | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/36391 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/30804 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/135053 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/37177 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/146272 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/99090 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/24852 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/30356 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/23116 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/35887 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/108274 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/35386 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/1141 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/35633 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/35533 "Built successfully") | | | |

### DIFF
--- a/Source/WebKit/UIProcess/mac/PageClientImplMac.mm
+++ b/Source/WebKit/UIProcess/mac/PageClientImplMac.mm
@@ -46,6 +46,7 @@
 #import "ViewGestureController.h"
 #import "ViewSnapshotStore.h"
 #import "WKAPICast.h"
+#import "WKAppKitGestureController.h"
 #import "WKFullScreenWindowController.h"
 #import "WKStringCF.h"
 #import "WKViewInternal.h"
@@ -315,6 +316,9 @@ void PageClientImpl::didCommitLoadForMainFrame(const String&, bool)
     impl->dismissContentRelativeChildWindowsWithAnimation(true);
     impl->clearPromisedImageDragData();
     impl->pageDidScroll({ 0, 0 });
+#if HAVE(APPKIT_GESTURES_SUPPORT)
+    impl->invalidateCachedPositionInformation();
+#endif
 #if ENABLE(WRITING_TOOLS)
     impl->hideTextAnimationView();
 #endif
@@ -1224,8 +1228,15 @@ void PageClientImpl::showCaptionDisplaySettings(WebCore::HTMLMediaElementIdentif
     protect(m_impl)->showCaptionDisplaySettings(identifier, options, WTF::move(completionHandler));
 }
 
-void PageClientImpl::positionInformationDidChange(const InteractionInformationAtPosition&)
+void PageClientImpl::positionInformationDidChange(const InteractionInformationAtPosition& info)
 {
+    CheckedPtr impl = m_impl.get();
+    if (!impl)
+        return;
+
+#if HAVE(APPKIT_GESTURES_SUPPORT)
+    impl->positionInformationDidChange(info);
+#endif
 }
 
 } // namespace WebKit
--- a/Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
+++ b/Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
@@ -30,10 +30,13 @@
 
 #import "AppKitSPI.h"
 #import "IdentifierTypes.h"
+#import "InteractionInformationAtPosition.h"
+#import "InteractionInformationRequest.h"
 #import "NativeWebWheelEvent.h"
 #import "RemoteLayerTreeDrawingAreaProxy.h"
 #import "ScrollingAccelerationCurve.h"
 #import "ViewGestureController.h"
+#import "WKDeferringGestureRecognizer.h"
 #import "WKWebView.h"
 #import "WebEventModifier.h"
 #import "WebEventType.h"
@@ -53,6 +56,7 @@
 #import <WebCore/PointerID.h>
 #import <WebCore/Scrollbar.h>
 #import <source_location>
+#import <wtf/BlockPtr.h>
 #import <wtf/CheckedPtr.h>
 #import <wtf/Markable.h>
 #import <wtf/MonotonicTime.h>
@@ -150,7 +154,19 @@ static WebCore::FloatSize toRawPlatformDelta(WebCore::FloatSize delta)
     return -delta;
 }
 
-@interface WKAppKitGestureController () <NSGestureRecognizerDelegatePrivate>
+using InteractionInformationCallback = Function<void(const WebKit::InteractionInformationAtPosition&)>;
+
+struct InteractionInformationRequestAndCallback {
+    WebKit::InteractionInformationRequest request;
+    InteractionInformationCallback callback;
+};
+
+static bool representsDraggableElement(const WebKit::InteractionInformationAtPosition& info)
+{
+    return info.isLink || info.isImage || info.isAttachment || info.isDHTMLDraggable || info.isColorInput || info.prefersDraggingOverTextSelection;
+}
+
+@interface WKAppKitGestureController () <NSGestureRecognizerDelegatePrivate, WKDeferringGestureRecognizerDelegate>
 @end
 
 @implementation WKAppKitGestureController {
@@ -162,6 +178,10 @@ static WebCore::FloatSize toRawPlatformDelta(WebCore::FloatSize delta)
     RetainPtr<NSPressGestureRecognizer> _singleClickGestureRecognizer;
     RetainPtr<NSClickGestureRecognizer> _doubleClickGestureRecognizer;
     RetainPtr<NSPressGestureRecognizer> _secondaryClickGestureRecognizer;
+
+    // Auxiliary gesture recognizers to support drag-and-drop.
+    RetainPtr<NSGestureRecognizer> _textSelectionDragGesture;
+    RetainPtr<WKDeferringGestureRecognizer> _dragDeferringGestureRecognizer;
 
     bool _isMomentumActive;
 
@@ -181,7 +201,14 @@ static WebCore::FloatSize toRawPlatformDelta(WebCore::FloatSize delta)
 
     RetainPtr<NSPressGestureRecognizer> _dragPressGestureRecognizer;
     RetainPtr<NSDraggingSession> _gestureDraggingSession;
+    BlockPtr<void(NSDraggingSession *)> _textSelectionDragCompletionHandler;
     bool _dragGestureHasSentMouseDown;
+
+    WebKit::InteractionInformationAtPosition _positionInformation;
+    std::optional<WebKit::InteractionInformationRequest> _lastOutstandingPositionInformationRequest;
+    Vector<std::optional<InteractionInformationRequestAndCallback>> _pendingPositionInformationHandlers;
+    uint64_t _positionInformationCallbackDepth;
+    bool _hasValidPositionInformation;
 }
 
 #if __has_include(<WebKitAdditions/WKAppKitGestureControllerAdditionsImpl.mm>)
@@ -213,6 +240,7 @@ static WebCore::FloatSize toRawPlatformDelta(WebCore::FloatSize delta)
     [self setUpDoubleClickGestureRecognizer];
     [self setUpSecondaryClickGestureRecognizer];
     [self setUpDragPressGestureRecognizer];
+    [self setUpDragDeferringGestureRecognizer];
 }
 
 - (void)setUpPanGestureRecognizer
@@ -259,6 +287,22 @@ static WebCore::FloatSize toRawPlatformDelta(WebCore::FloatSize delta)
     [_secondaryClickGestureRecognizer setName:@"WKSecondaryClickGesture"];
 }
 
+- (void)setUpDragPressGestureRecognizer
+{
+    _dragPressGestureRecognizer = adoptNS([[NSPressGestureRecognizer alloc] initWithTarget:self action:@selector(dragPressGestureRecognized:)]);
+    [self configureForDragPress:_dragPressGestureRecognizer];
+    [_dragPressGestureRecognizer setDelegate:self];
+    [_dragPressGestureRecognizer setName:@"WKDragPressGesture"];
+}
+
+- (void)setUpDragDeferringGestureRecognizer
+{
+    _dragDeferringGestureRecognizer = adoptNS([[WKDeferringGestureRecognizer alloc] initWithDeferringGestureDelegate:self]);
+    [self configureForDragDeferral:_dragDeferringGestureRecognizer];
+    [_dragDeferringGestureRecognizer setDelegate:self];
+    [_dragDeferringGestureRecognizer setName:@"WKDragDeferringGesture"];
+}
+
 - (void)addGesturesToWebView
 {
     CheckedPtr viewImpl = _viewImpl.get();
@@ -274,8 +318,8 @@ static WebCore::FloatSize toRawPlatformDelta(WebCore::FloatSize delta)
     [webView addGestureRecognizer:_singleClickGestureRecognizer.get()];
     [webView addGestureRecognizer:_doubleClickGestureRecognizer.get()];
     [webView addGestureRecognizer:_secondaryClickGestureRecognizer.get()];
-
-    // FIXME: Add drag press gesture after rdar://problem/176383341 is resolved.
+    [webView addGestureRecognizer:_dragPressGestureRecognizer.get()];
+    [webView addGestureRecognizer:_dragDeferringGestureRecognizer.get()];
 }
 
 - (void)enableGesturesIfNeeded
@@ -285,8 +329,7 @@ static WebCore::FloatSize toRawPlatformDelta(WebCore::FloatSize delta)
     [self enableGestureIfNeeded:_singleClickGestureRecognizer.get()];
     [self enableGestureIfNeeded:_doubleClickGestureRecognizer.get()];
     [self enableGestureIfNeeded:_secondaryClickGestureRecognizer.get()];
-
-    // FIXME: Enable drag press gesture after rdar://problem/176383341 is resolved.
+    [self enableGestureIfNeeded:_dragPressGestureRecognizer.get()];
 }
 
 - (void)enableGestureIfNeeded:(NSGestureRecognizer *)gesture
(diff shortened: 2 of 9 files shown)


Source/WebKit/UIProcess/mac/PageClientImplMac.mm
```diff
@@ -315,6 +316,9 @@ - (NSCursor *)_cursorRectCursor;
     impl->dismissContentRelativeChildWindowsWithAnimation(true);
     impl->clearPromisedImageDragData();
     impl->pageDidScroll({ 0, 0 });
+#if HAVE(APPKIT_GESTURES_SUPPORT)
+    [impl->appKitGestureController() didCommitLoadForMainFrame];
```
rr-codes: optional nit: I feel like for encapsulation purposes it's slightly better if other types don't directly access appKitGestureController and instead go through WebViewImpl, but doesn't really matter too much

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
```diff
@@ -181,7 +196,14 @@ @implementation WKAppKitGestureController {
 
     RetainPtr<NSPressGestureRecognizer> _dragPressGestureRecognizer;
     RetainPtr<NSDraggingSession> _gestureDraggingSession;
+    BlockPtr<void(NSDraggingSession *)> _textSelectionDragCompletionHandler;
```
rr-codes, line 204: optional nit: this would probably be more readable and potentially less bug-prone too if done in Swift especially with all the async-y stuff

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
```diff
@@ -573,16 +611,191 @@ - (void)mouseTrackingGestureRecognized:(NSGestureRecognizer *)gesture
...
+}
+
+- (BOOL)_hasValidOutstandingPositionInformationRequest:(const WebKit::InteractionInformationRequest&)request
+{
+    return _lastOutstandingPositionInformationRequest.and_then([&request](const auto& outstandingRequest) -> std::optional<bool> {
+        return outstandingRequest.isValidForRequest(request);
+    }).value_or(false);
+}
+
+- (void)_invokeAndRemovePendingHandlersValidForCurrentPositionInformation
```
rr-codes, line 757: I wonder if there's a way to abstract and factor out all this deferring/async/position information logic into an agnostic separate helper type that WKAppKitGestureController can then just use?

rr-codes:
(will review in depth later)

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
rr-codes: I actually think I feel slightly strongly about this, especially because of how complex the logic is, and the fact that it's pretty agnostic/applicable to things in general

Tools/TestWebKitAPI/Tests/WebKit/WebPage/AppKitGesturesTests.swift
```diff
@@ -318,6 +318,172 @@ struct AppKitGesturesTests {
...
+        let dragInitiated = Future()
+
+        let implementation: @convention(block) (NSView, NSArray, NSGestureRecognizer, AnyObject) -> NSDraggingSession? = { _, _, _, _ in
+            dragInitiated.signal()
+            return NSDraggingSession()
+        }
+
+        try await withSwizzledObjectiveCInstanceMethod(
+            replacing: NSView.self,
+            name: #selector(NSView.beginDraggingSession(items:gesture:source:)),
```
rr-codes, line 447: why does this need swizzling?

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
```diff
@@ -573,16 +611,191 @@ - (void)mouseTrackingGestureRecognized:(NSGestureRecognizer *)gesture
...
+        return NO;
+
+    if (deferringGestureRecognizer == _dragDeferringGestureRecognizer) {
+        NSPoint locationInView = [webView convertPoint:[event locationInWindow] fromView:nil];
+        if (viewImpl->isTextSelectedAtPoint(locationInView)) {
+            WK_APPKIT_GESTURE_CONTROLLER_RELEASE_LOG(RefPtr { _page.get() }->logIdentifier(), "Not deferring; text already selected at %@", NSStringFromPoint(locationInView));
+            return NO;
+        }
+
+        WK_APPKIT_GESTURE_CONTROLLER_RELEASE_LOG(RefPtr { _page.get() }->logIdentifier(), "Deferring; awaiting position info at %@", NSStringFromPoint(locationInView));
```
pxlcoder: Should these logs mention "drag" somewhere? Right now they just say "deferring" and "not deferring" without mentioning what.

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
```diff
@@ -573,16 +611,191 @@ - (void)mouseTrackingGestureRecognized:(NSGestureRecognizer *)gesture
...
+
+    for (int index = _pendingPositionInformationHandlers.size() - 1; index >= 0; --index) {
+        if (!_pendingPositionInformationHandlers[index])
+            _pendingPositionInformationHandlers.removeAt(index);
+    }
+}
+
+- (void)positionInformationDidChange:(const WebKit::InteractionInformationAtPosition&)info
+{
+    // Handlers could tear down `self`.
```
pxlcoder: How? Isn't the ownership `WKWebView` -> `WebViewImpl` -> `AppKitGestureController`?

Is it that the handlers could call into client code that tears down the web view?

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
aprotyas: Yes, they probably should, given drag deferral is only one use case.

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
aprotyas: That was my idea, yes, I do not have a practical example but you can imagine us calling into some UI delegate method from the `doAfterPositionInformationUpdate:` callback that ends up tearing down the web view.

Source/WebKit/UIProcess/mac/PageClientImplMac.mm
aprotyas: Sure, I will try to reword this and route through the web view instead.

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
```diff
@@ -573,16 +611,191 @@ - (void)mouseTrackingGestureRecognized:(NSGestureRecognizer *)gesture
...
+        [self doAfterPositionInformationUpdate:[weakSelf = WeakObjCPtr<WKAppKitGestureController>(self), weakDeferring = WeakObjCPtr<WKDeferringGestureRecognizer>(deferringGestureRecognizer)](const auto& info) {
+            RetainPtr strongSelf = weakSelf.get();
+            RetainPtr strongDeferring = weakDeferring.get();
+            if (!strongSelf || !strongDeferring)
+                return;
+
+            auto deferralState = [strongDeferring state];
+            if (deferralState != NSGestureRecognizerStatePossible) {
+                WK_APPKIT_GESTURE_CONTROLLER_RELEASE_LOG(RefPtr { strongSelf->_page.get() }->logIdentifier(),
+                    "Position info arrived after deferring gesture exited Possible (state=%ld); skipping resolution", (long)deferralState);
```
rr-codes: not that it really matters, but static_cast instead of C-style?

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
```diff
@@ -573,16 +611,191 @@ - (void)mouseTrackingGestureRecognized:(NSGestureRecognizer *)gesture
...
+                return;
+            }
+            bool isDraggableElement = info.isLink || info.isImage || info.isAttachment || info.isDHTMLDraggable || info.isColorInput || info.prefersDraggingOverTextSelection;
```
rr-codes: can this be extracted into a small static helper method maybe?

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
rr-codes: if only there was a language that would check this at compile time lol

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
```diff
@@ -1093,9 +1337,32 @@ - (BOOL)gestureRecognizerShouldBegin:(NSGestureRecognizer *)gestureRecognizer
...
+            bool requestIsValid = _hasValidPositionInformation && _positionInformation.request.isApproximatelyValidForRequest(request, radius);
+            bool shouldDrag = requestIsValid && isDraggableElement;
+            WK_APPKIT_GESTURE_CONTROLLER_RELEASE_LOG(RefPtr { _page.get() }->logIdentifier(),
+                "Drag-press shouldBegin → %d (hasInfo=%d link=%d image=%d attachment=%d dhtml=%d color=%d prefersDrag=%d radius=%d)",
+                shouldDrag, _hasValidPositionInformation,
+                _positionInformation.isLink, _positionInformation.isImage, _positionInformation.isAttachment,
+                _positionInformation.isDHTMLDraggable, _positionInformation.isColorInput, _positionInformation.prefersDraggingOverTextSelection, radius);
+            if (!requestIsValid)
+                [self _invalidateCurrentPositionInformation];
+            return shouldDrag;
```
rr-codes: can all this be extracted into its own method?

Source/WebKit/UIProcess/mac/WKTextSelectionController.swift
```diff
@@ -216,11 +216,106 @@ extension WKTextSelectionController {
 
     @objc(dragSelectionWithGesture:completionHandler:)
     func dragSelection(withGesture gesture: NSGestureRecognizer, completionHandler: @escaping @Sendable (NSDraggingSession) -> Void) {
```
rr-codes, line 202: this should be made async instead of a completion handler

Source/WebKit/UIProcess/mac/WKTextSelectionController.swift
```diff
@@ -216,11 +216,106 @@ extension WKTextSelectionController {
...
 
         Logger.viewGestures.log("[pageProxyID=\(page.logIdentifier())] \(#function) gesture: \(String(reflecting: gesture))")
+
+        let locationInWindow = gesture.location(in: nil)
+        let windowNumber = impl.windowNumber()
+        let timestamp = GetCurrentEventTime()
+        let modifierFlags = gesture.modifierFlags
+
+        guard
+            let mouseDown = NSEvent.mouseEvent(
```
rr-codes: this should probably make the variables first and then just do

```
guard let mouseDown, let mouseDragged else ...
```

Source/WebKit/UIProcess/mac/WKTextSelectionController.swift
```diff
@@ -216,11 +216,106 @@ extension WKTextSelectionController {
...
+                modifierFlags: modifierFlags,
+                timestamp: timestamp,
+                windowNumber: windowNumber,
+                context: nil,
+                eventNumber: 0,
+                clickCount: 1,
+                pressure: 1
+            )
+        else {
+            return
```
rr-codes, line 230: this should be `assertionFailure`'d since these should never really be nil

Source/WebKit/UIProcess/mac/WKTextSelectionController.swift
```diff
@@ -216,11 +216,106 @@ extension WKTextSelectionController {
...
+        }
+
+        impl.setTextSelectionDragGesture(gesture) { session in
```
rr-codes, line 233: async instead of completion handler

Source/WebKit/UIProcess/mac/WKTextSelectionController.swift
```diff
@@ -216,11 +216,106 @@ extension WKTextSelectionController {
...
+
+        let locationInWindow = gesture.location(in: nil)
+        let windowNumber = impl.windowNumber()
+        let timestamp = GetCurrentEventTime()
+        let modifierFlags = gesture.modifierFlags
+
+        switch gesture.state {
+        case .changed:
+            guard
+                let mouseDragged = NSEvent.mouseEvent(
```
rr-codes: maybe an extension function on `NSEvent` would be nice to avoid all this repetitive boilerplate

Tools/TestWebKitAPI/Tests/WebKit/WebPage/AppKitGesturesTests.swift
```diff
@@ -318,6 +318,172 @@ struct AppKitGesturesTests {
...
+                )
+            }
+
+            await dragInitiated.wait()
+        }
+
+        await page.waitForNextPresentationUpdate()
+
+        let selection = try? await page.callJavaScript(JavaScriptMessages.GetSelection())
+        if let selection {
```
rr-codes: no need for if let selection

Tools/TestWebKitAPI/Tests/WebKit/WebPage/AppKitGesturesTests.swift
```diff
@@ -318,6 +318,172 @@ struct AppKitGesturesTests {
...
+                    pressAndWait: .seconds(1.0)
+                )
+            }
+
+            await dragInitiated.wait()
+        }
+
+        await page.waitForNextPresentationUpdate()
+
+        let selection = try? await page.callJavaScript(JavaScriptMessages.GetSelection())
```
rr-codes:
```suggestion
        let selection = try await page.callJavaScript(JavaScriptMessages.GetSelection())
```

Tools/TestWebKitAPI/Tests/WebKit/WebPage/AppKitGesturesTests.swift
```diff
@@ -318,6 +318,172 @@ struct AppKitGesturesTests {
...
+
+        await page.waitForNextPresentationUpdate()
+
+        let selection = try await page.callJavaScript(JavaScriptMessages.GetSelection())
+
+        #expect(
+            selection
+                != .range(
+                    base: .init(in: "div", at: crazyRange.lowerBound),
+                    extent: .init(in: "div", at: try #require(Self.text.utf16Range(of: "ones")).upperBound)
```
rr-codes: `try #require(Self.text.utf16Range(of: "ones"))` should probably just be stored in a separate variable before this line

rr-codes (approved):
Looks good, but do consider https://github.com/WebKit/WebKit/pull/65035#discussion_r3253290612 if you want!

Source/WebKit/UIProcess/mac/WKTextSelectionController.swift
aprotyas: Right, did you mean something like this?

```swift
extension NSEvent {
    fileprivate static func syntheticMouseEvent(
        type: NSEvent.EventType,
        location: NSPoint,
        modifierFlags: NSEvent.ModifierFlags = [],
        timestamp: TimeInterval,
        windowNumber: Int,
        pressure: Float
    ) -> NSEvent? {
        NSEvent.mouseEvent(
            with: type,
            location: location,
            modifierFlags: modifierFlags,
            timestamp: timestamp,
            windowNumber: windowNumber,
            context: nil,
            eventNumber: 0,
            clickCount: 1,
            pressure: pressure
        )
    }
}
```

rr-codes: yup! (could go even further and just inline the CGGetCurrentEventTime() thing too but that's fine)

rr-codes: (also, it can be `_ type` instead of `type` since having a parameter label is kinda redundant in the case of this parameter

Tools/TestWebKitAPI/Tests/WebKit/WebPage/AppKitGesturesTests.swift
aprotyas: So this test tries to do a little better than "selection doesn't change" for the drag-and-drop case, by checking that a drag initiation happens (i.e. we get as far as `-[NSView beginDraggingSession:...]`). Actually starting an OS-level drag from a unit test is tricky because the system enters a modal drag tracking loop that blocks the main thread -- see workaround for this same issue in [313084@main](https://commits.webkit.org/313084@main). Generally other drag-and-drop tests avoid this problem because they "fake" their drag session by overriding `beginDraggingSession:...` (both `DragAndDropSimulator` as well as `TestRunnerWKWebView` do this). Swizzling this selector is our way of doing the same thing, and stubbing a code to signal a `Future` lets the test deterministically assert that the system did intend to initiate drag.

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
aprotyas: Yes, I strongly agree that it is the right direction to go. For this PR, I simply lifted the logic from WKContentViewInteraction.

I will follow up and extract a shared `PositionInformationCache` type that's used from both components. I'd rather not bundle it in with this PR since that's a sizable refactor on its own.

Source/WebKit/UIProcess/mac/WKTextSelectionController.swift
aprotyas: Bundling together a reply for this comment, https://github.com/WebKit/WebKit/pull/65035#discussion_r3253285596, and https://github.com/WebKit/WebKit/pull/65035#discussion_r3261392404.

I considered the async path, but the protocol declaration in NSTextSelectionManager annotates the block param (and the NSDraggingSession in the reply) as non-null. This makes async-ifying quite awkward, because when `WebViewImpl::cancelDrag()` gets called, AppKit never produced an NSDraggingSession for us to hand back, and we can't fabricate one.

An async -> NSDraggingSession path would force `withCheckedContinuation` to resume on the cancel paths, but there's no non-nil NSDraggingSession I can supply. And I can't make the path async -> NSDraggingSession? because Swift will warn about a nullability mismatch (rightfully so, we'd be in violation of the annotation).

Source/WebKit/UIProcess/mac/WKAppKitGestureController.mm
aprotyas: I filed:

> [Cocoa] Share position information caching across iOS and macOS interaction controllers
https://bugs.webkit.org/show_bug.cgi?id=315088

Tools/TestWebKitAPI/Tests/WebKit/WebPage/AppKitGesturesTests.swift
aprotyas: The `try?` was masking a latent issue. Check my new `.none` case

aprotyas:
Style guide error is a false positive. Moving forward with the merge, thank you for the reviews!

PR #65122, opened 2026-05-18
[Swift GTK] Preprocess Platform*.h to make Swift -D flags

adetaylor:
#### 8c249e411563bcc9390f06e0fbf1b00584b2e1c6
<pre>
[Swift GTK] Preprocess Platform*.h to make Swift -D flags
<a href="https://bugs.webkit.org/show_bug.cgi?id=315038">https://bugs.webkit.org/show_bug.cgi?id=315038</a>

Reviewed by Geoffrey Garen.

CMake rules for the Mac currently hard-code an option HAVE_MATERIAL_HOSTING.
They should not hard-code any such options; instead the options to be passed to
the Swift compiler should be worked out the same way as on Xcode builds, which
is by preprocessing Platform* and noting the resulting list of definitions.
That&apos;s what this change does. This will ensure the list of HAVE and USE symbols
is appropriate for each different platform.

This also deduplicates the way that -Xcc -D options are passed to Swift based
on the CMake configuration variables.

* Source/WTF/Scripts/generate-platform-args:
(platform_defines_from_dM):
(run_xcode_mode):
(run_cmake_mode):
(main):
* Source/WebKit/PlatformIOS.cmake:
* Tools/Scripts/swift/swiftc-wrapper.sh:

Canonical link: <a href="https://commits.webkit.org/314527@main">https://commits.webkit.org/314527@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/f5c0f51d0ad20e2b43857d88c7e4ae78cbabf95e

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/166357 "5 style errors") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/39847 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/33428 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/175405 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/123612 "Built successfully") | [❌ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/cea1af31-e7a7-4937-8fc7-7dc7fe9571e9) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/168223 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/40273 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/39855 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/129316 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/91071 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [❌ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/58fe6aab-d540-430d-a072-eb0ee338022a) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/169316 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/31695 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/149472 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/109841 "Passed tests") | | [❌ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/5327d5d0-b1ea-40bf-a7dc-8f1289b4acce) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/30673 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/29376 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/23545 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/20/builds/158514 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/140642 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/27169 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/177937 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/27251 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/30839 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/28875 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/137673 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/39917 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/33522 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/137590 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/37075 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/40605 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/148966 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/103260 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/32883 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/25832 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/199036 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/39115 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/112190 "Built successfully") | [  ~~🧪 jsc-armv7-tests~~](https://ews-build.webkit.org/#/builders/25/builds/53432 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/38512 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/3918 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/38757 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/38655 "Built successfully") | | | |


adetaylor:
Next steps here: this change can't be a configure-time step, because on initial build `wtf/Platform.h` won't exist in the desired location. So we need to change it to a build step. This was previously deemed to be difficult, but if we merge #65175 first, then we have a new `_SwiftRebuildTrigger.swift` target which we may be able to plumb in to allow the right sort of dependency tracking.

Source/WTF/Scripts/generate-platform-args
```diff
@@ -98,13 +53,138 @@ for arch in archs:
...
+def run_xcode_mode():
+    # For build settings which may be undefined, behave like a shell and implicitly
+    # fall back to the empty string.
+    os.environ.setdefault('FRAMEWORK_SEARCH_PATHS', '')
+    os.environ.setdefault('HEADER_SEARCH_PATHS', '')
+    os.environ.setdefault('SYSTEM_FRAMEWORK_SEARCH_PATHS', '')
+    os.environ.setdefault('SYSTEM_HEADER_SEARCH_PATHS', '')
+    os.environ.setdefault('GCC_PREPROCESSOR_DEFINITIONS', '')
+
+    framework_search_paths = shlex.split(
```
adetaylor, line 70: I considered getting CMake to set all these environment variables so that there would be less of a need for a "Xcode mode" vs "cmake mode" in this script, but it added a couple of complexities:
* This Xcode flow hard-codes `xcrun` so would need changes for non-Apple platforms
* In the `cmake` flow we also generate a load of `-Xcc -D` arguments which are not relevant in the Xcode flow. We could add a separate `cmake` custom build command for that
Overall it seems cleanest to keep these two flows, and share the common `-dM` parsing logic.

Source/WTF/Scripts/generate-platform-args
```diff
@@ -98,13 +53,138 @@ for arch in archs:
...
+        sys.exit('clang command is required after --')
+
+    result = subprocess.run(cmd, capture_output=True, text=True)
+    if result.returncode != 0:
+        sys.stderr.write(result.stderr)
+        sys.exit(result.returncode)
+
+    swift_defines = platform_defines_from_dM(result.stdout)
+
+    cmake_seeds = []
```
adetaylor, line 149: This bit isn't in the pre-existing Xcode flow: it's transcribing the contents of `cmakeconfig.h` into `-Xcc -D` arguments for Swift, because clang modules don't support `-include` (ugh)

Source/WTF/Scripts/generate-platform-args
```diff
@@ -1,82 +1,42 @@
...
-         *(arg for path in system_header_search_paths for arg in ('-isystem', path)),
-         '-include', 'wtf/Platform.h', '/dev/zero'), text=True)
-
-    # Concatenate all the depfiles, in case a header is only imported on specific architectures.
-    shutil.copyfileobj(open(depfile), combined_depfile_fd)
+# Two modes:
+#
+#   Xcode build phase (default, no args): driven by Xcode env vars
+#   ($SCRIPT_OUTPUT_FILE_0, $ARCHS, $DERIVED_FILES_DIR, ...). Iterates over
+#   $ARCHS and writes one resp per arch.
```
kmiller68, line 17: Nit: Might be worth changing the xcodebuild phase to pass those arguments rather than parse them from the environment. That's a gross way to do things generally IMO.

Source/cmake/WebKitMacros.cmake
```diff
@@ -813,6 +818,83 @@ function(_webkit_setup_swift_header_deps _target _stamp _header)
...
+# ${_resp_path}, a swiftc @-response file containing the platform-derived
+# flags. Mirrors the Xcode "Generate Swift platform args" build phase.
+#
+# Output format (one token per line, swiftc-response-file convention):
+#   -DNAME              for every truthy HAVE_/USE_/ENABLE_/WTF_PLATFORM_/ASSERT_
+#                       macro that wtf/Platform.h derives (Swift #if conditions)
+#   -Xcc -DNAME=VALUE   for every entry in cmakeconfig.h (clang importer seed)
+#
+# Build-time (rather than configure-time) is required so the staged WTF
+# headers exist when the preprocessor runs (clean builds otherwise fail).
```
kmiller68: It seems like the real reason to do it at build time is that it would be confusing to make changing a header require re-configuring. Normally, you only have to reconfigure when adding/removing files or something in the system changed (SDK/toolchain/etc).

kmiller68 (commented):
Seems mostly reasonable. A few comments.

Source/WTF/Scripts/generate-platform-args
adetaylor: (as discussed on chat) Prototyped here - https://github.com/WebKit/WebKit/compare/main...adetaylor:WebKit:eng/generate-platform-args-uses-parameters-from-xcode - but it results in some duplication between pbxproj files.

Source/cmake/WebKitMacros.cmake
adetaylor: Yeah. I believe cmake can be encouraged to automatically reconfigure when certain files change, and we probably could make that work.

But IMO doing it at build-time rather than configure-time is the obvious thing to do for many reasons, including parallelism and correct dependency tracking; it just makes things a bit more complicated.

In any case I think this bit of the comment is over-verbose and I'm going to shrink it.

adetaylor:
@kmiller68 also sent me a bunch of AI-generated feedback via chat, which I've addressed too. Specifically,

>   1. Race condition: `_webkit_generate_platform_swift_args` now appends WTF, ${_target}, and any $`{_target}_FRAMEWORKS _CopyHeaders / _CopyPrivateHeaders` targets to its `DEPENDS`, guarded by `if (TARGET …)`
> 2. Spurious Swift rebuilds: `run_cmake_mode` now reads the existing `resp` and skips the write when contents match.
> 3. Stale-script trap: Dropped `../Scripts/generate-platform-args` from the `file(COPY)` in `Source/WTF/wtf/CMakeLists.txt`; the cmake macro now references `${CMAKE_SOURCE_DIR}/Source/WTF/Scripts/generate-platform-args` directly.
> 4. Duplicate regex: Removed `CMAKECONFIG_DEFINE_RE`

Source/cmake/WebKitMacros.cmake
```diff
@@ -813,6 +818,97 @@ function(_webkit_setup_swift_header_deps _target _stamp _header)
...
+#   -Xcc -DNAME=VALUE   for every entry in cmakeconfig.h (clang importer seed)
+#
+# The resp is rebuilt whenever any tracked Platform header changes thanks to
+# DEPFILE.
+function(_webkit_generate_platform_swift_args _target _resp_path _ordering_dep)
+    set(_depfile "${_resp_path}.d")
+    # /dev/null can't be used as the clang input because its mtime is "now",
+    # so the depfile-driven dirty check would always fire. Use a stable empty
+    # file in the build tree instead.
+    set(_empty_input "${CMAKE_BINARY_DIR}/CMakeFiles/empty.cpp")
```
geoffreygaren: This should probably go in DerivedSources.

geoffreygaren (approved):
I don't think this is a viable long-term approach, but it seems like an improvement in the short term, at least, to get CMake to match Xcode.

PR #65241, opened 2026-05-20
Auto-pick last-built tree on Mac for scripts that use local build products, add --xcode flag

zakariaridouh:
#### b3f1f7efedad5b6f50e1acd35a4a400e2393d120
<pre>
Auto-pick last-built tree on Mac for scripts that use local build products, add --xcode flag
<a href="https://bugs.webkit.org/show_bug.cgi?id=315146">https://bugs.webkit.org/show_bug.cgi?id=315146</a>
<a href="https://rdar.apple.com/177478344">rdar://177478344</a>

Reviewed by Pascoe.

When both WebKitBuild/&lt;Config&gt;/ (Xcode) and WebKitBuild/cmake-mac/&lt;Config&gt;/
(CMake) exist, webkit-build-directory now defaults to whichever tree was built
most recently instead of always preferring Xcode. The choice is logged on
STDERR (&quot;Using last-built tree: cmake-mac/Debug&quot; or &quot;Using last-built tree:
Debug&quot;).

Tools that benefit (anything that shells out to webkit-build-directory):
run-api-tests, run-webkit-tests, run-perf-tests, run-webdriver,
run-minibrowser, run-swiftbrowser, run-jsc-stress-tests, run-testmem,
dump-class-layout, install-built-product, measure-build-time,
extract-tbds-from-internal-sdk.

Build drivers (build-webkit and its wrappers, set-webkit-configuration) call
webkitdirs.pm directly without opting into the tiebreaker and keep today&apos;s
behavior, so they never auto-flip the build target.

Add --xcode as a counterpart to --cmake for explicit pinning.

The tiebreaker is opt-in via enableLastBuiltTiebreaker(), which
webkit-build-directory calls for itself before any product-directory query. It
compares each build system&apos;s own build log rather than a specific product
binary, so a partial build (e.g. a JSC-only build that never relinks
WebKit.framework) still moves the marker: the CMake/Ninja tree&apos;s
cmake-mac/&lt;Config&gt;/.ninja_log versus the Xcode/XCBuild tree&apos;s
WebKitBuild/XCBuildData/build.db, read via File::stat&apos;s -&gt;mtime. A missing log
resolves to mtime 0, so an absent log degrades to the pre-existing &quot;Xcode wins
when both exist&quot; default. build.db lives at the WebKitBuild root and is shared
across Xcode configurations, so with two Xcode configs it reflects the most
recent Xcode build of any of them rather than this specific configuration; that
only ever biases back toward the historical Xcode-wins default. Non-Cocoa ports
are gated out via the existing isAppleCocoaWebKit() guard.

* Tools/Scripts/webkit-build-directory:
* Tools/Scripts/webkitdirs.pm:
(determineIsCMakeBuild):
(enableLastBuiltTiebreaker): Added.
* Tools/Scripts/webkitperl/webkitdirs_unittest/determineIsCMakeBuild.pl:
* Tools/Scripts/webkitperl/webkitdirs_unittest/determineIsCMakeBuildXcode.pl: Added.
* Tools/Scripts/webkitperl/webkitdirs_unittest/determineIsCMakeBuildLastBuiltCMake.pl: Added.
* Tools/Scripts/webkitperl/webkitdirs_unittest/determineIsCMakeBuildLastBuiltXcode.pl: Added.
* Tools/Scripts/webkitpy/port/base.py:
(Port.__init__):
* Tools/Scripts/webkitpy/port/config.py:
(Config.__init__):
(Config.build_directory):
* Tools/Scripts/webkitpy/port/config_unittest.py:
(ConfigTest.make_config):
(ConfigTest.test_build_directory_passes_xcode_flag):
(ConfigTest.test_build_directory_passes_xcode_flag.mock_run_command):
* Tools/Scripts/webkitpy/port/factory.py:
(configuration_options):

Canonical link: <a href="https://commits.webkit.org/316853@main">https://commits.webkit.org/316853@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/610f23a72a10ed18d77c4861630b3f47c3898d4e

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/173569 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/47502 "Built successfully") | [  ~~🛠 mac~~](https://ews-build.webkit.org/#/builders/168/builds/41187 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/183142 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/131437 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [loading 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/22439290-f75c-4b16-9b37-6382443e197f) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/175434 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/48794 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/47184 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/134967 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/131437 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/176527 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/38395 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/155681 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/115444 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/46ad3f31-84ef-42d0-a813-aaf4a6479742) 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/172890 "Passed tests") | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/37036 "Passed tests") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/35397 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/31627 "Built successfully") | | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/4193 "Passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/147460 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/35176 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/185568 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/34831 "Built successfully and passed tests") | | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/39597 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/143847 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/47169 "Built successfully") | | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/143704 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/38777 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/47848 "Built successfully") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/168/builds/41187 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/113022 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/38640 "Passed tests") | | [  ~~🛠 jsc-armv7~~](https://ews-build.webkit.org/#/builders/35/builds/210602 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/46354 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/10129 "Passed tests") | [  ~~🧪 jsc-armv7-tests~~](https://ews-build.webkit.org/#/builders/35/builds/210602 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/45756 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/45987 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/45815 "Built successfully") | | | |

### DIFF
--- a/Tools/Scripts/webkit-build-directory
+++ b/Tools/Scripts/webkit-build-directory
@@ -37,6 +37,11 @@ use Getopt::Long;
 use lib $FindBin::Bin;
 use webkitdirs;
 
+# Opt this read-only path resolver into the last-built tiebreaker. Build
+# drivers (build-webkit, set-webkit-configuration) never call this, so they
+# keep the "Xcode wins when both trees exist" default.
+enableLastBuiltTiebreaker();
+
 my $showConfigurationDirectory = 0;
 my $showExecutablePath = 0;
 my $showHelp = 0;
--- a/Tools/Scripts/webkitdirs.pm
+++ b/Tools/Scripts/webkitdirs.pm
@@ -104,6 +104,7 @@ BEGIN {
        &determineCurrentSVNRevision
        &determineCrossTarget
        &determineXcodeSDK
+       &enableLastBuiltTiebreaker
        &executableProductDir
        &exitStatus
        &extractNonMacOSHostConfiguration
@@ -279,6 +280,7 @@ my $osXVersion;
 my $iosVersion;
 my $generateDsym;
 my $isCMakeBuild;
+my $shouldPickLastBuilt = 0;
 my $isGenerateProjectOnly;
 my $shouldBuild32Bit;
 my $isInspectorFrontend;
@@ -3111,20 +3113,61 @@ sub determineIsCMakeBuild()
     return if defined($isCMakeBuild);
     $isCMakeBuild = checkForArgumentAndRemoveFromARGV("--cmake");
     return if $isCMakeBuild;
+    if (checkForArgumentAndRemoveFromARGV("--xcode")) {
+        $isCMakeBuild = 0;
+        return;
+    }
 
-    # Auto-detect a CMake macOS build when no Xcode build is present at the
-    # expected path. The CMake macOS presets place artifacts under
-    # WebKitBuild/cmake-mac/<Configuration>; an Xcode build, if present,
-    # always wins to preserve existing workflows.
+    # Auto-detect a CMake macOS build. The CMake presets place artifacts under
+    # WebKitBuild/cmake-mac/<Configuration>; when both trees exist, Xcode wins
+    # by default unless a caller opts into the last-built tiebreaker via
+    # enableLastBuiltTiebreaker() (read-only path resolvers do; build drivers
+    # like build-webkit do not, so they are never auto-flipped).
     if (isAppleCocoaWebKit()) {
         determineBaseProductDir();
         determineConfiguration();
         my $cmakeMacBuild = File::Spec->catdir($baseProductDir, "cmake-mac", $configuration);
         my $xcodeBuild = File::Spec->catdir($baseProductDir, $configuration);
+
         if (-f File::Spec->catfile($cmakeMacBuild, "CMakeCache.txt") && !-d $xcodeBuild) {
             $isCMakeBuild = 1;
+            return;
+        }
+
+        # Heuristic: prefer whichever tree the developer built most recently,
+        # comparing each build system's own build log rather than a specific
+        # product binary. A product like WebKit.framework only relinks when
+        # WebKit itself changes, so a JSC-only (or WTF-only, tool-only) build
+        # would leave it stale and flip the choice to the wrong tree. Both build
+        # systems instead touch a log on every build of any target:
+        #   - CMake/Ninja:   cmake-mac/<Configuration>/.ninja_log
+        #   - Xcode/XCBuild: WebKitBuild/XCBuildData/build.db
+        # A missing log resolves to mtime 0, so an absent log degrades to the
+        # pre-existing "Xcode wins when both exist" default. Note build.db lives
+        # at the WebKitBuild root and is shared across Xcode configurations, so
+        # with two Xcode configs it reflects the most recent Xcode build of any
+        # of them rather than this specific configuration.
+        if ($shouldPickLastBuilt && -d $cmakeMacBuild && -d $xcodeBuild) {
+            my $cmakeMarker = File::Spec->catfile($cmakeMacBuild, ".ninja_log");
+            my $xcodeMarker = File::Spec->catfile($baseProductDir, "XCBuildData", "build.db");
+            my $cmakeMtime = -f $cmakeMarker ? stat($cmakeMarker)->mtime : 0;
+            my $xcodeMtime = -f $xcodeMarker ? stat($xcodeMarker)->mtime : 0;
+            if ($cmakeMtime > $xcodeMtime) {
+                $isCMakeBuild = 1;
+                print STDERR "Using last-built tree: cmake-mac/$configuration (CMake)\n";
+            } elsif ($xcodeMtime && $cmakeMtime) {
+                print STDERR "Using last-built tree: $configuration (Xcode)\n";
+            }
         }
     }
+}
+
+# Opt a read-only path resolver (e.g. webkit-build-directory) into the
+# last-built tiebreaker in determineIsCMakeBuild(). Must be called before the
+# first isCMakeBuild()/product-directory query, since the result is cached.
+sub enableLastBuiltTiebreaker
+{
+    $shouldPickLastBuilt = 1;
 }
 
 sub isCMakeBuild()
--- a/Tools/Scripts/webkitperl/webkitdirs_unittest/determineIsCMakeBuildLastBuiltCMake.pl
+++ b/Tools/Scripts/webkitperl/webkitdirs_unittest/determineIsCMakeBuildLastBuiltCMake.pl
@@ -1,0 +1,75 @@
+#!/usr/bin/env perl
+#
+# Copyright (C) 2026 Apple Inc. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1.  Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+# 2.  Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in the
+#     documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS ``AS IS'' AND ANY
+# EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS BE LIABLE FOR ANY
+# DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+# (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON
+# ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+# SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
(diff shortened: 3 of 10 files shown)


Tools/Scripts/webkit-build-directory
```diff
@@ -37,6 +37,9 @@ use Getopt::Long;
 use lib $FindBin::Bin;
 use webkitdirs;
 
+# Opt this read-only path resolver into the last-built tiebreaker.
+$ENV{WEBKIT_PICK_LAST_BUILT} = 1 unless defined $ENV{WEBKIT_PICK_LAST_BUILT};
```
pascoej: If I understand correctly this is setting an environment variable. Could we avoid that or make it's purpose more clear (ex. only set if want opt out?)

zakariaridouh: Its not an env var in the usual sense, it's set inside the `webkit-build-directory` at startup, and lives only for the duration of this perl process.

I'm using it as a one way signal from this script to `webkitdirs.pm` where we read it to enable the mtime check and "tie breaker". `build-webkit` and `set-webkit-configuration` use webkitdirs directly and never enter this script, so they never set the flag and the new code path is silently skipped for them since we check if its set before we do the mtime checks.

We could switch it to a module scoped flag with a setter (something like setPickLastBuiltMode(1)) if we want to skip the whole ENV thing.

Tools/Scripts/webkitdirs.pm
```diff
@@ -3161,18 +3161,38 @@ sub determineIsCMakeBuild()
...
         my $xcodeBuild = File::Spec->catdir($baseProductDir, $configuration);
+
         if (-f File::Spec->catfile($cmakeMacBuild, "CMakeCache.txt") && !-d $xcodeBuild) {
             $isCMakeBuild = 1;
+            return;
+        }
+
+        if ($ENV{WEBKIT_PICK_LAST_BUILT} && -d $cmakeMacBuild && -d $xcodeBuild) {
+            my $cmakeMarker = File::Spec->catfile($cmakeMacBuild, ".ninja_log");
+            my $xcodeMarker = File::Spec->catfile($xcodeBuild, "WebKit.framework", "WebKit");
```
geoffreygaren: Are we sure that the Xcode build will always touch this path? It seems like an inefficiency that we might some day remove.

Tools/Scripts/webkitdirs.pm
```diff
@@ -3161,18 +3161,38 @@ sub determineIsCMakeBuild()
...
         my $cmakeMacBuild = File::Spec->catdir($baseProductDir, "cmake-mac", $configuration);
         my $xcodeBuild = File::Spec->catdir($baseProductDir, $configuration);
+
         if (-f File::Spec->catfile($cmakeMacBuild, "CMakeCache.txt") && !-d $xcodeBuild) {
             $isCMakeBuild = 1;
+            return;
+        }
+
+        if ($ENV{WEBKIT_PICK_LAST_BUILT} && -d $cmakeMacBuild && -d $xcodeBuild) {
+            my $cmakeMarker = File::Spec->catfile($cmakeMacBuild, ".ninja_log");
```
geoffreygaren, line 3151: Does ninja touch this file in a no-op build?

Tools/Scripts/webkit-build-directory
geoffreygaren: I tend to agree that passing an argument by environment variable is a bit hard to follow. 

Can callers pass an argument to determineIsCMakeBuild instead?

Tools/Scripts/webkitdirs.pm
zakariaridouh: Verified that a no-op build (no work to do) does not touch .ninja_log; it only advances when >= 1 command runs. Regardless, I've changed this approach so we no longer rely on `.ninja_log`

Tools/Scripts/webkitdirs.pm
zakariaridouh: I'm not entirely sure, but I agree that relying on that specific path was pretty finicky and hack. Changed it so both trees now compare `WebKit.framework/Versions/A/WebKit` (which the rest of `webkitdirs.pm` uses). 

If either marker is missing/stale it resolves to mtime 0 and we fall through to the current "Xcode wins when both exist" default, so the "worst case scenario" is current behavior.

Tools/Scripts/webkit-build-directory
zakariaridouh: Fair enough, I removed the env var. It's now an opt-in module flag with a setPickLastBuiltMode() setter that webkit-build-directory calls before any product-directory query.

Tools/Scripts/webkit-build-directory
```diff
@@ -37,6 +37,11 @@ use Getopt::Long;
 use lib $FindBin::Bin;
 use webkitdirs;
 
+# Opt this read-only path resolver into the last-built tiebreaker. Build
+# drivers (build-webkit, set-webkit-configuration) never call this, so they
+# keep the "Xcode wins when both trees exist" default.
+setPickLastBuiltMode(1);
```
pascoej: This 1 seems magical, maybe just because the name doesn't seem flag-like.

zakariaridouh: changed it to a flag like var.

pascoej (approved):

Tools/Scripts/webkitperl/webkitdirs_unittest/determineIsCMakeBuildLastBuiltCMake.pl
```diff
@@ -0,0 +1,70 @@
...
+# The tiebreaker is gated on isAppleCocoaWebKit(); force it on so the test
+# is platform-independent (webkitperl runs on non-Cocoa bots too).
+no warnings qw(redefine prototype);
+*webkitdirs::isAppleCocoaWebKit = sub () { 1 };
+use warnings qw(redefine prototype);
+
+my $configuration = "Debug";
+my $base = tempdir(CLEANUP => 1);
+my $cmakeMarker = File::Spec->catfile($base, "cmake-mac", $configuration, "WebKit.framework", "Versions", "A", "WebKit");
+my $xcodeMarker = File::Spec->catfile($base, $configuration, "WebKit.framework", "Versions", "A", "WebKit");
```
pascoej: non-blocking: Might choose a different file that is always built? (thinking of JSC engineers who don't build WebKit after)

zakariaridouh: changed it to `.ninja_log`

PR #65283, opened 2026-05-20
[GTK][WPE] ThreadedCompositor can composite a half-committed set of scrolling-tree layer positions

alexgcastro:
#### d2e8005925d80b352d9adba5edeaab8d6669da8c
<pre>
[GTK][WPE] ThreadedCompositor can composite a half-committed set of scrolling-tree layer positions
<a href="https://bugs.webkit.org/show_bug.cgi?id=315185">https://bugs.webkit.org/show_bug.cgi?id=315185</a>

Reviewed by NOBODY (OOPS!).

Fixed, sticky, positioned and frame-scrolling layer positions are corrupted by two
races during async scrolling.

The scrolling thread applies layer positions while ThreadedCompositor copies the
same layers into its TextureMapperLayer targets on the compositor thread. Nothing
serializes the whole apply pass against the whole flush, so the compositor can copy
a frame node&apos;s new position together with a fixed/sticky child&apos;s old one. Serialize
them with a per-scene CoordinatedSceneState lock, held around the apply pass via
willSetPositionsForScrolling()/didSetPositionsForScrolling() and around the
flushCompositingState() copy loop.

GraphicsLayerCoordinated::updateGeometry also writes m_position from the main thread;
during a programmatic scroll it lays out one frame ahead of the offset the scrolling
thread applied, so its setPosition() lands a stale, un-scroll-compensated position
over the scroll-adjusted one. Mark the layer as owned by the scrolling tree
(m_ownedByScrollingTree, set in setPositionForScrolling) and suppress setPosition()
while owned; resetScrollingTreeOwnership() returns it to main-thread control on
hand-off.

* Source/WebCore/page/scrolling/coordinated/ScrollingTreeCoordinated.cpp:
(WebCore::ScrollingTreeCoordinated::applyLayerPositionsInternal):
* Source/WebCore/page/scrolling/coordinated/ScrollingTreeFixedNodeCoordinated.cpp:
(WebCore::ScrollingTreeFixedNodeCoordinated::commitStateBeforeChildren):
* Source/WebCore/page/scrolling/coordinated/ScrollingTreeFrameScrollingNodeCoordinated.cpp:
(WebCore::resetScrollingTreeOwnership):
(WebCore::ScrollingTreeFrameScrollingNodeCoordinated::commitStateBeforeChildren):
* Source/WebCore/page/scrolling/coordinated/ScrollingTreePositionedNodeCoordinated.cpp:
(WebCore::ScrollingTreePositionedNodeCoordinated::commitStateBeforeChildren):
* Source/WebCore/page/scrolling/coordinated/ScrollingTreeStickyNodeCoordinated.cpp:
(WebCore::ScrollingTreeStickyNodeCoordinated::commitStateBeforeChildren):
* Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp:
(WebCore::CoordinatedPlatformLayer::setPosition):
(WebCore::CoordinatedPlatformLayer::setPositionForScrolling):
(WebCore::CoordinatedPlatformLayer::resetScrollingTreeOwnership):
(WebCore::CoordinatedPlatformLayer::willSetPositionsForScrolling):
(WebCore::CoordinatedPlatformLayer::didSetPositionsForScrolling):
* Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.h:
* Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/CoordinatedSceneState.h:
(WebKit::CoordinatedSceneState::stateLock):
* Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/LayerTreeHost.cpp:
(WebKit::LayerTreeHost::willSetPositionsForScrolling):
(WebKit::LayerTreeHost::didSetPositionsForScrolling):
* Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/LayerTreeHost.h:
* Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/LayerTreeHostPlayStation.cpp:
(WebKit::LayerTreeHost::willSetPositionsForScrolling):
(WebKit::LayerTreeHost::didSetPositionsForScrolling):
* Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/LayerTreeHostPlayStation.h:
* Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/ThreadedCompositor.cpp:
(WebKit::ThreadedCompositor::flushCompositingState):
</pre>
https://github.com/WebKit/WebKit/commit/d2e8005925d80b352d9adba5edeaab8d6669da8c

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/167645 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/41142 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/34737 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/176702 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/125326 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/169511 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/41577 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/41155 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/130436 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/91685 "1 flakes 5 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/170604 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/32858 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/150637 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/110953 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/31840 "Passed tests") | [❌ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/30540 "Found 1 new API test failure: TestWebKitAPI.SiteIsolation.MultiProcessBFCacheSameSiteWithCrossSiteIframe (failure)") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/25435 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/141827 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/28332 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/179242 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/32283 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/30050 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/138840 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/41214 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/34697 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/138725 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/41902 "Built successfully") | | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/105064 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/33985 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/27003 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/40406 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/113652 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/39803 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/5104 "Passed tests") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/40054 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/39948 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/page/scrolling/coordinated/ScrollingTreeCoordinated.cpp
+++ b/Source/WebCore/page/scrolling/coordinated/ScrollingTreeCoordinated.cpp
@@ -85,7 +85,13 @@ void ScrollingTreeCoordinated::applyLayerPositionsInternal()
     if (!rootScrollingNode)
         return;
 
+    auto rootContentsLayer = static_cast<ScrollingTreeFrameScrollingNodeCoordinated*>(rootScrollingNode)->rootContentsLayer();
+    if (!rootContentsLayer)
+        return;
+
+    rootContentsLayer->willSetPositionsForScrolling();
     ThreadedScrollingTree::applyLayerPositionsInternal();
+    rootContentsLayer->didSetPositionsForScrolling();
 
     if (ScrollingThread::isCurrentThread()) {
         auto rootContentsLayer = static_cast<ScrollingTreeFrameScrollingNodeCoordinated*>(rootScrollingNode)->rootContentsLayer();
--- a/Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
+++ b/Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
@@ -176,6 +176,8 @@ void CoordinatedPlatformLayer::notifyCompositionRequired()
 void CoordinatedPlatformLayer::setPosition(FloatPoint&& position)
 {
     ASSERT(m_lock.isHeld());
+    if (m_ownedByScrollingTree)
+        return;
     if (m_position == position)
         return;
 
@@ -187,12 +189,19 @@ void CoordinatedPlatformLayer::setPosition(FloatPoint&& position)
 void CoordinatedPlatformLayer::setPositionForScrolling(const FloatPoint& position, ForcePositionSync forceSync)
 {
     Locker locker { m_lock };
+    m_ownedByScrollingTree = true;
     if (m_position == position && forceSync == ForcePositionSync::No)
         return;
 
     m_position = position;
     m_pendingChanges.add(Change::Position);
     notifyCompositionRequired();
+}
+
+void CoordinatedPlatformLayer::resetScrollingTreeOwnership()
+{
+    Locker locker { m_lock };
+    m_ownedByScrollingTree = false;
 }
 
 const FloatPoint& CoordinatedPlatformLayer::position() const
@@ -900,6 +909,18 @@ void CoordinatedPlatformLayer::requestComposition(CompositionReason reason)
 {
     if (m_client)
         m_client->requestComposition(reason);
+}
+
+void CoordinatedPlatformLayer::willSetPositionsForScrolling()
+{
+    if (m_client)
+        m_client->willSetPositionsForScrolling();
+}
+
+void CoordinatedPlatformLayer::didSetPositionsForScrolling()
+{
+    if (m_client)
+        m_client->didSetPositionsForScrolling();
 }
 
 RunLoop* CoordinatedPlatformLayer::compositingRunLoop() const
--- a/Source/WebCore/page/scrolling/coordinated/ScrollingTreeFixedNodeCoordinated.cpp
+++ b/Source/WebCore/page/scrolling/coordinated/ScrollingTreeFixedNodeCoordinated.cpp
@@ -56,8 +56,11 @@ bool ScrollingTreeFixedNodeCoordinated::commitStateBeforeChildren(const Scrollin
         return false;
 
     auto& fixedStateNode = downcast<ScrollingStateFixedNode>(stateNode);
-    if (fixedStateNode.hasChangedProperty(ScrollingStateNode::Property::Layer))
+    if (fixedStateNode.hasChangedProperty(ScrollingStateNode::Property::Layer)) {
+        if (m_layer)
+            m_layer->resetScrollingTreeOwnership();
         m_layer = static_cast<CoordinatedPlatformLayer*>(fixedStateNode.layer());
+    }
 
     return ScrollingTreeFixedNode::commitStateBeforeChildren(stateNode);
 }
--- a/Source/WebCore/page/scrolling/coordinated/ScrollingTreeFrameScrollingNodeCoordinated.cpp
+++ b/Source/WebCore/page/scrolling/coordinated/ScrollingTreeFrameScrollingNodeCoordinated.cpp
@@ -40,6 +40,12 @@
 
 namespace WebCore {
 
+static void resetScrollingTreeOwnership(CoordinatedPlatformLayer* layer)
+{
+    if (layer)
+        layer->resetScrollingTreeOwnership();
+}
+
 Ref<ScrollingTreeFrameScrollingNode> ScrollingTreeFrameScrollingNodeCoordinated::create(ScrollingTree& scrollingTree, ScrollingNodeType nodeType, ScrollingNodeID nodeID)
 {
     return adoptRef(*new ScrollingTreeFrameScrollingNodeCoordinated(scrollingTree, nodeType, nodeID));
@@ -60,26 +66,47 @@ ScrollingTreeScrollingNodeDelegateCoordinated& ScrollingTreeFrameScrollingNodeCo
 
 bool ScrollingTreeFrameScrollingNodeCoordinated::commitStateBeforeChildren(const ScrollingStateNode& stateNode)
 {
+    RefPtr<CoordinatedPlatformLayer> previousScrolledContentsLayer;
+    if (auto* frameStateNode = dynamicDowncast<ScrollingStateFrameScrollingNode>(stateNode)) {
+        if (frameStateNode->hasChangedProperty(ScrollingStateNode::Property::ScrolledContentsLayer) && scrolledContentsLayer())
+            previousScrolledContentsLayer = static_cast<CoordinatedPlatformLayer*>(scrolledContentsLayer());
+    }
+
     if (!ScrollingTreeFrameScrollingNode::commitStateBeforeChildren(stateNode))
         return false;
+
+    if (previousScrolledContentsLayer)
+        previousScrolledContentsLayer->resetScrollingTreeOwnership();
 
     if (!is<ScrollingStateFrameScrollingNode>(stateNode))
         return false;
 
     const auto& scrollingStateNode = downcast<ScrollingStateFrameScrollingNode>(stateNode);
 
-    if (scrollingStateNode.hasChangedProperty(ScrollingStateNode::Property::RootContentsLayer))
+    if (scrollingStateNode.hasChangedProperty(ScrollingStateNode::Property::RootContentsLayer)) {
+        resetScrollingTreeOwnership(m_rootContentsLayer.get());
         m_rootContentsLayer = static_cast<CoordinatedPlatformLayer*>(scrollingStateNode.rootContentsLayer());
-    if (scrollingStateNode.hasChangedProperty(ScrollingStateNode::Property::CounterScrollingLayer))
+    }
+    if (scrollingStateNode.hasChangedProperty(ScrollingStateNode::Property::CounterScrollingLayer)) {
+        resetScrollingTreeOwnership(m_counterScrollingLayer.get());
         m_counterScrollingLayer = static_cast<CoordinatedPlatformLayer*>(scrollingStateNode.counterScrollingLayer());
-    if (scrollingStateNode.hasChangedProperty(ScrollingStateNode::Property::InsetClipLayer))
+    }
+    if (scrollingStateNode.hasChangedProperty(ScrollingStateNode::Property::InsetClipLayer)) {
+        resetScrollingTreeOwnership(m_insetClipLayer.get());
         m_insetClipLayer = static_cast<CoordinatedPlatformLayer*>(scrollingStateNode.insetClipLayer());
-    if (scrollingStateNode.hasChangedProperty(ScrollingStateNode::Property::ContentShadowLayer))
+    }
+    if (scrollingStateNode.hasChangedProperty(ScrollingStateNode::Property::ContentShadowLayer)) {
+        resetScrollingTreeOwnership(m_contentShadowLayer.get());
         m_contentShadowLayer = static_cast<CoordinatedPlatformLayer*>(scrollingStateNode.contentShadowLayer());
-    if (scrollingStateNode.hasChangedProperty(ScrollingStateNode::Property::HeaderLayer))
+    }
+    if (scrollingStateNode.hasChangedProperty(ScrollingStateNode::Property::HeaderLayer)) {
+        resetScrollingTreeOwnership(m_headerLayer.get());
         m_headerLayer = static_cast<CoordinatedPlatformLayer*>(scrollingStateNode.headerLayer());
-    if (scrollingStateNode.hasChangedProperty(ScrollingStateNode::Property::FooterLayer))
+    }
+    if (scrollingStateNode.hasChangedProperty(ScrollingStateNode::Property::FooterLayer)) {
+        resetScrollingTreeOwnership(m_footerLayer.get());
         m_footerLayer = static_cast<CoordinatedPlatformLayer*>(scrollingStateNode.footerLayer());
+    }
 
     m_delegate->updateFromStateNode(scrollingStateNode);
     return true;
--- a/Source/WebCore/page/scrolling/coordinated/ScrollingTreePositionedNodeCoordinated.cpp
+++ b/Source/WebCore/page/scrolling/coordinated/ScrollingTreePositionedNodeCoordinated.cpp
@@ -52,8 +52,11 @@ ScrollingTreePositionedNodeCoordinated::~ScrollingTreePositionedNodeCoordinated(
 
 bool ScrollingTreePositionedNodeCoordinated::commitStateBeforeChildren(const ScrollingStateNode& stateNode)
 {
-    if (stateNode.hasChangedProperty(ScrollingStateNode::Property::Layer))
+    if (stateNode.hasChangedProperty(ScrollingStateNode::Property::Layer)) {
+        if (m_layer)
+            m_layer->resetScrollingTreeOwnership();
         m_layer = static_cast<CoordinatedPlatformLayer*>(stateNode.layer());
+    }
 
     return ScrollingTreePositionedNode::commitStateBeforeChildren(stateNode);
 }
--- a/Source/WebCore/page/scrolling/coordinated/ScrollingTreeStickyNodeCoordinated.cpp
+++ b/Source/WebCore/page/scrolling/coordinated/ScrollingTreeStickyNodeCoordinated.cpp
@@ -54,8 +54,11 @@ ScrollingTreeStickyNodeCoordinated::ScrollingTreeStickyNodeCoordinated(Scrolling
 
 bool ScrollingTreeStickyNodeCoordinated::commitStateBeforeChildren(const ScrollingStateNode& stateNode)
 {
-    if (stateNode.hasChangedProperty(ScrollingStateNode::Property::Layer))
+    if (stateNode.hasChangedProperty(ScrollingStateNode::Property::Layer)) {
+        if (m_layer)
+            m_layer->resetScrollingTreeOwnership();
         m_layer = static_cast<CoordinatedPlatformLayer*>(stateNode.layer());
+    }
 
     return ScrollingTreeStickyNode::commitStateBeforeChildren(stateNode);
 }
--- a/Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.h
+++ b/Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.h
@@ -78,6 +78,8 @@ public:
         virtual bool isCompositionRequiredOrOngoing() const = 0;
         virtual void requestComposition(CompositionReason) = 0;
         virtual RunLoop* compositingRunLoop() const = 0;
+        virtual void willSetPositionsForScrolling() = 0;
+        virtual void didSetPositionsForScrolling() = 0;
         virtual int maxTextureSize() const = 0;
         virtual void willPaintTile() = 0;
         virtual void didPaintTile() = 0;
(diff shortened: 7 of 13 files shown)


Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
```diff
@@ -180,41 +180,65 @@ void CoordinatedPlatformLayer::setPosition(FloatPoint&& position)
...
+}
+
 const FloatPoint& CoordinatedPlatformLayer::position() const
 {
     ASSERT(m_lock.isHeld());
-    return m_position;
+    return m_scrollingTreePosition ? *m_scrollingTreePosition : m_position;
+}
+
+void CoordinatedPlatformLayer::commitPosition()
```
carlosgcampos: commitPositionFromScrollingTree? for consistency and to clarify what this is for

alexgcastro: Good point! I'll change it, thanks! :pray:

Source/WebCore/page/scrolling/coordinated/ScrollingTreeCoordinated.cpp
```diff
@@ -79,6 +79,21 @@ Ref<ScrollingTreeNode> ScrollingTreeCoordinated::createScrollingTreeNode(Scrolli
...
+static void commitLayerPositionsRecursive(ScrollingTreeNode& node)
+{
+    if (node.isFixedNode())
+        static_cast<ScrollingTreeFixedNodeCoordinated&>(node).commitLayerPositions();
+    else if (node.isStickyNode())
+        static_cast<ScrollingTreeStickyNodeCoordinated&>(node).commitLayerPositions();
+    else if (node.isFrameScrollingNode())
+        static_cast<ScrollingTreeFrameScrollingNodeCoordinated&>(node).commitLayerPositions();
+    else if (node.isPositionedNode())
+        static_cast<ScrollingTreePositionedNodeCoordinated&>(node).commitLayerPositions();
```
carlosgcampos: I think we could add a pure virtual to ScrollingTreeNode and we don't need these ifs

Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
```diff
@@ -180,41 +180,65 @@ void CoordinatedPlatformLayer::setPosition(FloatPoint&& position)
...
     m_pendingChanges.add(Change::Position);
     notifyCompositionRequired();
 }
 
-void CoordinatedPlatformLayer::setPositionForScrolling(const FloatPoint& position, ForcePositionSync forceSync)
+void CoordinatedPlatformLayer::setPositionFromScrollingTree(const FloatPoint& position, ForcePositionSync forceSync)
 {
     Locker locker { m_lock };
-    if (m_position == position && forceSync == ForcePositionSync::No)
+    if (m_scrollingTreePosition == position && forceSync == ForcePositionSync::No)
```
carlosgcampos: I would call the member m_pendingPositionFromScrollingTree

Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
```diff
@@ -180,41 +180,65 @@ void CoordinatedPlatformLayer::setPosition(FloatPoint&& position)
...
         return;
 
-    m_position = position;
+    m_scrollingTreePosition = position;
     m_pendingChanges.add(Change::Position);
     notifyCompositionRequired();
 }
 
+void CoordinatedPlatformLayer::resetPositionFromScrollingTree()
```
carlosgcampos:
```suggestion
void CoordinatedPlatformLayer::resetPendingPositionFromScrollingTree()
```

Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
```diff
@@ -180,41 +180,65 @@ void CoordinatedPlatformLayer::setPosition(FloatPoint&& position)
...
+    m_scrollingTreePosition = std::nullopt;
+    m_committedPosition = m_position;
+    m_pendingChanges.add(Change::Position);
+}
+
 const FloatPoint& CoordinatedPlatformLayer::position() const
 {
     ASSERT(m_lock.isHeld());
-    return m_position;
+    return m_scrollingTreePosition ? *m_scrollingTreePosition : m_position;
```
carlosgcampos: I'm not sure I understand this, I guess we want either the committed or position, but we don't want to return the pending one no?

Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
```diff
@@ -180,41 +180,65 @@ void CoordinatedPlatformLayer::setPosition(FloatPoint&& position)
...
+
+void CoordinatedPlatformLayer::commitPosition()
+{
+    Locker locker { m_lock };
+    if (!m_scrollingTreePosition)
+        return;
+    if (m_committedPosition == *m_scrollingTreePosition)
+        return;
+    m_committedPosition = *m_scrollingTreePosition;
+    m_pendingChanges.add(Change::Position);
```
carlosgcampos: Don't we set the pending one to nullopt? Once committed we no longer have a pending position. And do we need to change the pending changes? If we have a pending position and it's different from the committed one is because setPositionFromScrollingThread had already been called, so the falg has already been added, no?

Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
```diff
@@ -996,7 +1020,7 @@ void CoordinatedPlatformLayer::flushCompositingStateOnTarget(const OptionSet<Com
 {
     if (reasons.containsAny({ CompositionReason::RenderingUpdate, CompositionReason::AsyncScrolling })) {
         if (m_pendingChanges.contains(Change::Position)) {
-            layer.setPosition(m_position);
+            layer.setPosition(m_committedPosition);
```
carlosgcampos: I'm not sure I understand why we need 3 positions. Can't m_position be the committed one? I wonder if we really need to explicitly commit the position after applying all positions, or they can be committed here while flushing the compositing state.

carlosgcampos: Does this work when disabling async scrolling? because here we use m_committedPosition which is only set from async scrolling if I've understood the patch correctly.

carlosgcampos (changes requested):

Source/WebCore/page/scrolling/coordinated/ScrollingTreeCoordinated.cpp
alexgcastro: Right! Good idea.

Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
alexgcastro: Right!

Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
alexgcastro: Right.

Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
alexgcastro: Right, I'm trying to remove the committed.

Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
alexgcastro: Ditto.

fujii:
You don't fix the race condition. I still see the problem by adding sleep.

~~~diff
diff --git a/Source/WebCore/page/scrolling/coordinated/ScrollingTreeCoordinated.cpp b/Source/WebCore/page/scrolling/coordinated/ScrollingTreeCoordinated.cpp
index d3f79535c159..1bdf62e336c3 100644
--- a/Source/WebCore/page/scrolling/coordinated/ScrollingTreeCoordinated.cpp
+++ b/Source/WebCore/page/scrolling/coordinated/ScrollingTreeCoordinated.cpp
@@ -83,6 +83,8 @@ static void commitLayerPositionsRecursive(ScrollingTreeNode& node)
 {
     node.commitLayerPositions();
 
+    sleep(10_ms);
+
     for (auto& child : node.children())
         commitLayerPositionsRecursive(child.get());
 }
diff --git a/Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/ThreadedCompositor.cpp b/Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/ThreadedCompositor.cpp
index 1b0d2d6ef20b..223752c90832 100644
--- a/Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/ThreadedCompositor.cpp
+++ b/Source/WebKit/WebProcess/WebPage/CoordinatedGraphics/ThreadedCompositor.cpp
@@ -442,6 +442,8 @@ static String reasonsToString(const OptionSet<CompositionReason>& reasons)
 
 void ThreadedCompositor::renderLayerTree()
 {
+    sleep(10_ms);
+
     ASSERT(m_sceneState);
     ASSERT(m_workQueue->runLoop().isCurrent());
 #if PLATFORM(GTK) || PLATFORM(WPE)
~~~

alexgcastro:
> You don't fix the race condition. I still see the problem by adding sleep.
> 

Thanks for the feedback! :pray: It is interesting because the current patch fixes my current problem for a race, but clearly in an scenario that is pushing it we need to lock probably. Now I wonder if we want to solve this one too in this PR or create a new one, let me try to adapt the patch to make it more transactional in the end. I think it would need locking but let's try to do it as smart as possible.

Source/WebCore/page/scrolling/coordinated/ScrollingTreeCoordinated.cpp
```diff
@@ -87,10 +95,15 @@ void ScrollingTreeCoordinated::applyLayerPositionsInternal()
 
     ThreadedScrollingTree::applyLayerPositionsInternal();
 
-    if (ScrollingThread::isCurrentThread()) {
-        auto rootContentsLayer = static_cast<ScrollingTreeFrameScrollingNodeCoordinated*>(rootScrollingNode)->rootContentsLayer();
+    auto rootContentsLayer = static_cast<ScrollingTreeFrameScrollingNodeCoordinated*>(rootScrollingNode)->rootContentsLayer();
+    if (rootContentsLayer)
+        rootContentsLayer->willCommitSceneState();
+    commitLayerPositionsRecursive(*rootScrollingNode);
```
fujii: Do we really need commitLayerPositionsRecursive?
I think just calling `ThreadedScrollingTree::applyLayerPositionsInternal()` here is enough.

fujii: This is [the minimal patch](https://github.com/user-attachments/files/28693382/simple.patch) of the locking approach.
You seem to try to fix multiple problems in this PR. But, I understand only a single race condition.
If you know more problems, I'd like to understand the problems. Let me know.

alexgcastro: Thanks for the review and the patch! :pray:  I'll review it, this patch was initially 2 patches for races of the 3 threads, but maybe it can be simplified after adding the last lock, I just wanted to do it as a follow up, so that we can more easily bisect with this more complete version if there is any issue. But I can go in the opposite direction and leave just the smaller and propose the next one if required.

I'll check the patch and decide if there is any good reason to keep it and test it.

Thanks again!

carlosgcampos: I first thought that it was a good idea to keep the pending/committed approach to keep it atomic and take the lock for less time, but that means we have to iterate twice. So maybe it's not worth it, it would be a lot simpler to just lock before applyLayerPositionsInternal() like @fujii suggests and iterate once.

alexgcastro: I agree, let's focus on this one first, I'm going to change the topic to explain the fix is for the compositor thread, and we can discuss later about main vs scrolling thread if there is a way to reproduce, like @fujii is proposing. That way we can progress more easily.

I'm going to upload the simplified patch for review and we can open a new bug to check if we can reproduce the other scenario realiably. 

Thanks for all the help! :pray:

Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
alexgcastro: Right, this is not required anymore. Thanks!

Source/WebCore/page/scrolling/coordinated/ScrollingTreeCoordinated.cpp
fujii: I'm not sure I understand your concern.
However, the minimal patch of the locking approach has a subtle race condition between the main thread and scrolling thread.
Also, the locking approach has a caveat that the scrolling thread blocks the compositor thread.
I think we should take the message passing approach.
The scroll thread applies scrolling layer positions and sends a message to the compositor thread. Then, the compositor thread commits the scrolling layer positions, and sends a message to the scrolling thread.

fujii: This is [a minimal patch](https://github.com/user-attachments/files/28781729/mini-msg-pass.patch) of the message passing approach.
But, this patch needs more works and revised. The compositor thread doesn't send a message.

alexgcastro: @fujii What do you if to help evolve all the work in smaller chunks, and progress in the discussion we land this one and  you can propose the improvement for the message passing as a different patch? I can add the bug explaining the other race with an example and that would speed up the work :pray: 😊

Thanks!

fujii: It makes difficult to review if we revert a previous approach.
I think we should understand the problem before deciding which direction we go.
If you have an example, let me see it.

alexgcastro: Ok, let's try it this html with this patch and you can see the main vs scrolling race, I can reproduce it even with the current lock patch, use the mouse whell with the minibrowser on desktop while it is automatically scrolling, I'm going to test your message passing patch to verify it:

```
diff --git a/Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp b/Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
index 4074da78d088..14c396605d3a 100644
--- a/Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
+++ b/Source/WebCore/platform/graphics/texmap/coordinated/CoordinatedPlatformLayer.cpp
@@ -179,6 +179,9 @@ void CoordinatedPlatformLayer::setPosition(FloatPoint&& position)
     if (m_position == position)
         return;

+    WTFLogAlways("RACE main write on layer %p main=(%.0f,%.0f) over scroll-set=(%.0f,%.0f)", this, position.x(), position.y(), m_position.x(), m_position.y());
+    WTF::sleep(WTF::Seconds::fromMilliseconds(15));
+
     m_position = WTF::move(position);
     m_pendingChanges.add(Change::Position);
     notifyCompositionRequired();
```
[raceB-repro.html](https://github.com/user-attachments/files/28806286/raceB-repro.html)

alexgcastro: > This is [a minimal patch](https://github.com/user-attachments/files/28781729/mini-msg-pass.patch) of the message passing approach. But, this patch needs more works and revised. The compositor thread doesn't send a message.

Tested this patch, it fixes the races but is has a major problem that it starves the async scrolling when I use the scrollTo webpage, it does not even render the scrollbar, probably this condition stops it because main is asking for it:

```
+    if (rootContentsLayer->isCompositionRequiredOrOngoing())
+        return;

alexgcastro:
New version in  d2e8005 adds to the lock and ownership concept, that simplifies our previous pending with just a boolean, fixing both races. Basically with the lock we just need the main thread not to change the position and just trust let the scrolling thread to drive the changes.

Still I like @fujii's proposal to try to avoid the locks, but it looks like a bigger change that is why I added this one. My proposal would be to land this one that fixes both races, and try other options that are more ambitious, that would provide a place to bisect if we break or not and differences, providing information in the history of the commits about this problem in the commits, for the work of trying to get the best solution.

Thanks for all the help! :pray:

Source/WebCore/page/scrolling/coordinated/ScrollingTreeCoordinated.cpp
fujii: Thank you for the test content. I understand the problems now. Attached your test content to the bugzilla ticket. https://bugs.webkit.org/show_bug.cgi?id=315185
Yes, my minimal patch of the message passing approach still has even more problems. The compositor thread has to post a message after committing layer positions.

fujii:
I don't think the ownership concept is the right approach to fix the problems.
I think we should take the message passing or transaction approach to fix the problems. However, I admit those approaches aren't simple.
I'd like to ask @carlosgcampos to approve this PR.

alexgcastro:
Closing in favor of PR #67062

PR #65297, opened 2026-05-20
Give JIT helper threads QOS_UTILITY on less powerful devices

nipunshukla21:
#### 371f7ad0ed89cd9ca8034690e3986ba7ddcc24ca
<pre>
Give JIT helper threads QOS_UTILITY on less powerful devices
<a href="https://bugs.webkit.org/show_bug.cgi?id=315203">https://bugs.webkit.org/show_bug.cgi?id=315203</a>
<a href="https://rdar.apple.com/177532305">rdar://177532305</a>

Reviewed by Marcus Plutowski and Dan Hecht.

On devices with only 2 P-cores, JIT worklist threads run at the same
QOS as the main threads, causing P-core contention. Sustained P-cluster
load also pushes less powerful devices into thermal throttling,
causing performance losses.

This change lowers JIT worklist thread priority to QOS::Utility on
small devices, defined as &lt;=2 P-cores or &lt;=6 logical cores. JIT threads
can then run on E-cores and free the P-cluster. This causes regression
on more powerful devices so we keep existing behavior there.

There are no functional changes associated with this commit.

* Source/JavaScriptCore/jit/JITWorklistThread.cpp:
(JSC::JITWorklistThread::threadDidStart):
* Source/WTF/wtf/NumberOfCores.cpp:
(WTF::numberOfPerformanceProcessorCores):
* Source/WTF/wtf/NumberOfCores.h:
* Source/WTF/wtf/Threading.cpp:
(WTF::Thread::setCurrentThreadIsUtility):
* Source/WTF/wtf/Threading.h:

Canonical link: <a href="https://commits.webkit.org/313672@main">https://commits.webkit.org/313672@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/b463b074aad0db1e871d3058622ed6c8554acd33

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/163618 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/37102 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/30321 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/172558 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/120067 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/37433 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/37101 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/127091 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/89600 "layout-tests (failure)") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/166577 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/29365 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/147100 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/107645 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/28414 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/27045 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/20261 "Built successfully") | 
| [✅ 🛠 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/20/builds/155769 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/138313 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/24817 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/175063 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/24509 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/27994 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/26480 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/135393 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/36772 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/31147 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/135375 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36526 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/37557 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/146612 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/99618 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/30689 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/23464 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/196020 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/36267 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/109413 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/51106 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/35765 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/1490 "Passed tests") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/36015 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/35912 "Built successfully") | | |

### DIFF
--- a/Source/JavaScriptCore/jit/JITWorklistThread.cpp
+++ b/Source/JavaScriptCore/jit/JITWorklistThread.cpp
@@ -33,6 +33,11 @@
 #include "VM.h"
 #include <wtf/TZoneMallocInlines.h>
 
+#if HAVE(QOS_CLASSES)
+#include <wtf/NumberOfCores.h>
+#include <wtf/Threading.h>
+#endif
+
 namespace JSC {
 
 WTF_MAKE_TZONE_ALLOCATED_IMPL(JITWorklistThread);
@@ -174,6 +179,15 @@ void JITWorklistThread::threadDidStart()
 {
     dataLogLnIf(Options::verboseCompilationQueue(), m_worklist, ": Thread started");
 
+#if HAVE(QOS_CLASSES)
+#if CPU(ARM64)
+    bool isSmall = WTF::numberOfPerformanceProcessorCores() <= static_cast<int>(Options::smallDeviceMaxPerformanceCores());
+#else
+    bool isSmall = WTF::numberOfProcessorCores() <= static_cast<int>(Options::smallDeviceMaxLogicalCores());
+#endif
+    if (isSmall)
+        Thread::setCurrentThreadIsUtility();
+#endif
 }
 
 void JITWorklistThread::threadIsStopping(const AbstractLocker&)
--- a/Source/WTF/wtf/NumberOfCores.cpp
+++ b/Source/WTF/wtf/NumberOfCores.cpp
@@ -104,4 +104,20 @@ int numberOfPhysicalProcessorCores()
 }
 #endif
 
+#if CPU(ARM64) && OS(DARWIN)
+int numberOfPerformanceProcessorCores()
+{
+    static int s_pcores = 0;
+    if (s_pcores > 0)
+        return s_pcores;
+
+    int32_t count = 0;
+    size_t valueSize = sizeof(count);
+    int result = sysctlbyname("hw.perflevel0.physicalcpu", &count, &valueSize, nullptr, 0);
+    RELEASE_ASSERT(result >= 0 && count > 0);
+    s_pcores = count;
+    return s_pcores;
+}
+#endif
+
 }
--- a/Source/JavaScriptCore/runtime/OptionsList.h
+++ b/Source/JavaScriptCore/runtime/OptionsList.h
@@ -299,6 +299,8 @@ bool hasCapacityToUseLargeGigacage();
     v(Int32, priorityDeltaOfDFGCompilerThreads, computePriorityDeltaOfWorkerThreads(-1, 0), Normal, nullptr) \
     v(Int32, priorityDeltaOfFTLCompilerThreads, computePriorityDeltaOfWorkerThreads(-2, 0), Normal, nullptr) \
     v(Int32, priorityDeltaOfWasmCompilerThreads, computePriorityDeltaOfWorkerThreads(-1, 0), Normal, nullptr) \
+    v(Unsigned, smallDeviceMaxPerformanceCores, 2, Normal, nullptr) \
+    v(Unsigned, smallDeviceMaxLogicalCores, 6, Normal, nullptr) \
     \
     v(Bool, useProfiler, false, Normal, nullptr) \
     v(Bool, dumpProfilerDataAtExit, false, Normal, nullptr) \
--- a/Source/WTF/wtf/NumberOfCores.h
+++ b/Source/WTF/wtf/NumberOfCores.h
@@ -32,4 +32,10 @@ WTF_EXPORT_PRIVATE int numberOfProcessorCores();
 WTF_EXPORT_PRIVATE int numberOfPhysicalProcessorCores();
 #endif
 
+#if CPU(ARM64) && OS(DARWIN)
+// Returns the number of performance ("P") cores on Apple silicon, queried via
+// the "hw.perflevel0.physicalcpu" sysctl.
+WTF_EXPORT_PRIVATE int numberOfPerformanceProcessorCores();
+#endif
+
 }
--- a/Source/WTF/wtf/Threading.cpp
+++ b/Source/WTF/wtf/Threading.cpp
@@ -393,6 +393,17 @@ void Thread::setCurrentThreadIsUserInitiated(int relativePriority)
 #endif
 }
 
+void Thread::setCurrentThreadIsUtility(int relativePriority)
+{
+#if HAVE(QOS_CLASSES)
+    ASSERT(relativePriority <= 0);
+    ASSERT(relativePriority >= QOS_MIN_RELATIVE_PRIORITY);
+    pthread_set_qos_class_self_np(adjustedQOSClass(QOS_CLASS_UTILITY), relativePriority);
+#else
+    UNUSED_PARAM(relativePriority);
+#endif
+}
+
 #if HAVE(QOS_CLASSES)
 static Thread::QOS NODELETE toQOS(qos_class_t qosClass)
 {
--- a/Source/WTF/wtf/Threading.h
+++ b/Source/WTF/wtf/Threading.h
@@ -199,6 +199,7 @@ public:
     // relativePriority is a value in the range [-15, 0] where a lower value indicates a lower priority.
     WTF_EXPORT_PRIVATE static void setCurrentThreadIsUserInteractive(int relativePriority = 0);
     WTF_EXPORT_PRIVATE static void setCurrentThreadIsUserInitiated(int relativePriority = 0);
+    WTF_EXPORT_PRIVATE static void setCurrentThreadIsUtility(int relativePriority = 0);
     WTF_EXPORT_PRIVATE static QOS currentThreadQOS();
     WTF_EXPORT_PRIVATE static bool currentThreadIsRealtime();
     bool isRealtime() const { return m_isRealtime; }


Source/WTF/wtf/NumberOfCores.cpp
```diff
@@ -104,4 +104,21 @@ int numberOfPhysicalProcessorCores()
 }
 #endif
 
+bool isSmallDevice(int maxPerformanceCores, int maxLogicalCoresFallback)
+{
+#if OS(DARWIN)
+    static int32_t pcores = [] {
+        int32_t count = 0;
+        size_t valueSize = sizeof(count);
+        int result = sysctlbyname("hw.perflevel0.physicalcpu", &count, &valueSize, nullptr, 0);
```
Andrew-Fryer: The name of the sysctl stuff is odd IMHO.
I think it might be worth a comment to say that "hw.perflevel0.physicalcpu" counts performance cores and "hw.physicalcpu_max" counts all cores (rather than needing to infer that from variable names).

Source/WTF/wtf/NumberOfCores.cpp
```diff
@@ -104,4 +104,23 @@ int numberOfPhysicalProcessorCores()
 }
 #endif
 
+bool isSmallDevice(int maxPerformanceCores, int maxLogicalCoresFallback)
```
Achierius: I'd rather this be "deviceHasSmallCoreCount" or even "isSmallCoreCountDevice" to be explicit

Achierius: Shouldn't the argument names be like "minPerformanceCores" &c? It's also unintuitive that only one of the options does anything on a given system. I would prefer that there's one argument, and that we treat the argument as #pcores on systems that have distinct p/e cores, and #cores on systems that don't (since there all cores can be said to be 'performant'). Then pass in a different argument depending on the system type -- that could be keyed off of whether the platform is ARM64.

Source/WTF/wtf/NumberOfCores.cpp
```diff
@@ -104,4 +104,23 @@ int numberOfPhysicalProcessorCores()
...
+{
+#if OS(DARWIN)
+    static int32_t pcores = [] {
+        int32_t count = 0;
+        size_t valueSize = sizeof(count);
+        int result = sysctlbyname("hw.perflevel0.physicalcpu", &count, &valueSize, nullptr, 0);
+        if (result < 0 || count <= 0)
+            return -1;
```
Achierius: This seems likely to cause "unexpected behavior" insofar as an error (e.g. a sandbox violation) would change our scheduler policies. That seems undesirable, I would softly prefer this to be an error instead so that we can catch that if it does happen

Achierius (changes requested):

Source/WTF/wtf/NumberOfCores.cpp
nipunshukla21: I think `maxCores` works better here as it is the maximum core count needed to satisfy the condition of whether it's a small core count device. Let me know if that makes sense and if not I can change it to min cores

nipunshukla21:
I also started one more round of A/B on different device configurations to ensure the gating works correctly. I will wait for that to come back as expected (neutral on powerful devices, progressed on weak ones) before attempting to land pending an r+.

Source/JavaScriptCore/jit/JITWorklistThread.cpp
```diff
@@ -173,7 +178,15 @@ auto JITWorklistThread::work() -> WorkResult
...
     dataLogLnIf(Options::verboseCompilationQueue(), m_worklist, ": Thread started");
-
+#if HAVE(QOS_CLASSES)
+#if CPU(ARM64)
+    static bool isSmall = WTF::isSmallCoreCountDevice(Options::smallDeviceMaxPerformanceCores());
+#else
+    static bool isSmall = WTF::isSmallCoreCountDevice(Options::smallDeviceMaxLogicalCores());
+#endif
+    if (isSmall)
+        pthread_set_qos_class_self_np(QOS_CLASS_UTILITY, 0);
```
dhecht: Should this use `Thread::adjustedQOSClass`?

Source/JavaScriptCore/jit/JITWorklistThread.cpp
```diff
@@ -173,7 +178,15 @@ auto JITWorklistThread::work() -> WorkResult
 void JITWorklistThread::threadDidStart()
 {
     dataLogLnIf(Options::verboseCompilationQueue(), m_worklist, ": Thread started");
-
+#if HAVE(QOS_CLASSES)
+#if CPU(ARM64)
+    static bool isSmall = WTF::isSmallCoreCountDevice(Options::smallDeviceMaxPerformanceCores());
```
dhecht: does this work correctly given we build with -fno-threadsafe-static?

Source/WTF/wtf/NumberOfCores.cpp
```diff
@@ -104,4 +104,22 @@ int numberOfPhysicalProcessorCores()
...
+        size_t valueSize = sizeof(count);
+        int result = sysctlbyname("hw.perflevel0.physicalcpu", &count, &valueSize, nullptr, 0);
+        RELEASE_ASSERT(result >= 0 && count > 0);
+        return count;
+    }();
+    return pcores <= maxCores;
+#endif
+    return numberOfProcessorCores() <= maxCores;
+}
+
```
dhecht, line 122: I think it's a bit confusing that the caller and callee ifdefs are not the same condition (so there are various combinations to think about across file boundaries). Is it on purpose that the caller does not have the DARWIN guard as well?

In any case, why not just implement a numberOfPerformanceProcessorCores in NumberOfCores.cpp (which keeps that file just an API to get at CPU topology) and then have JITWorklistThread.cpp have all the policy logic that uses the core counts as inputs?

Achierius (approved):
r=me

Source/JavaScriptCore/jit/JITWorklistThread.cpp
dhecht: Also, are there any issues with priority inversion, given the JIT threads share locks with the mutator? I think usually a boost would be given but not sure that will happen with our custom lock implementation.

dhecht: Or more generally, have a parallel to setCurrentThreadIsUserInitiated and setCurrentThreadIsUserInteractive for utility?

Source/WTF/wtf/NumberOfCores.cpp
nipunshukla21: >I think it's a bit confusing that the caller and callee ifdefs are not the same condition (so there are various combinations to think about across file boundaries). Is it on purpose that the caller does not have the DARWIN guard as well?

We define `HAVE(QOS_CLASSES)` as an alias for `PLATFORM(COCOA)`:
```
#if !defined(HAVE_QOS_CLASSES) && PLATFORM(COCOA)
#define HAVE_QOS_CLASSES 1
#endif
```
and `PLATFORM(COCOA)`  is defined as:
```
#if PLATFORM(MAC) || PLATFORM(IOS_FAMILY)
#define WTF_PLATFORM_COCOA 1
#endif
```
so implicit in `HAVE(QOS_CLASSES)`  is that we're satisfying `OS(DARWIN)` so we only need the extra `CPU(ARM64)` branch.  I felt that if a future engineer were to look at each bit of code in isolation if would be more immediately obvious what can be called where given this choice of ifdefs. The callee function is a generic WTF utility which could theoretically be invoked from non-Cocoa ARM64 builds like on Linux, where `sysctlbyname` isn't available so I added a specific `OS(DARWIN)` gate there to make it clear where and when it can be called.

That was my rationale but if you disagree I'm happy to change things to be more consistent.

dhecht: The refactoring to move the policy into the caller resolves this part of the concern as well since now the ifdefs are independent (would just be a compile-time issue if there's an inconsistency).

dhecht (approved):
r=me too

PR #65351, opened 2026-05-21
[iOS] Defer ElementDidFocus IPC to the next rendering update

NathanSolomon733:
#### 64ee64bad41e5d511964c29b8ab00ee77269821a
<pre>
[iOS] Defer ElementDidFocus IPC to the next rendering update
<a href="https://bugs.webkit.org/show_bug.cgi?id=315247">https://bugs.webkit.org/show_bug.cgi?id=315247</a>
<a href="https://rdar.apple.com/177468274">rdar://177468274</a>

Reviewed by Wenson Hsieh.

On iOS, WebPage::elementDidFocus runs layoutIfNeeded() from a JS click handler
dispatched via simulateClick() in order to populate interactionRect for the
ElementDidFocus IPC. But, this is very expensive for form-heavy interactions.
Defer the IPC to the next rendering update. elementDidFocus() now stores the
gesture state into a member variable on WebPage,
ensures the next rendering update is scheduled, and then returns. Utilize
the fact that the next rendering update will already do layoutIfNeeded(). So when
WebPage::updateRendering() runs during that next rendering update,
Page::updateRendering() has already run its layout pass for the frame so we send the
IPC message with correct geometry for the focused element while avoiding extra
layoutIfNeeded() calls for each user-gesture-driven focus that would fire before
the next rendering update. This also means we are now sending fewer IPC messages
to the UI Process than before.

Remove and update iOS-specific test baselines that captured existing failures caused by the
extra layoutIfNeeded() in the iOS focus path. The exec-command
baselines had a FAIL for undo not clearing after .value
assignment on input/textarea inside contenteditable. The content-visibility
ios baseline had an extra newline (for some currently unknown reason) now
matches the generic baseline that matches MacOS baseline as well.

* LayoutTests/platform/ios/imported/w3c/web-platform-tests/css/css-contain/content-visibility/content-visibility-053-expected.txt: Removed.
* LayoutTests/platform/ios/imported/w3c/web-platform-tests/editing/other/exec-command-with-text-editor.tentative_type=text-expected.txt:
* LayoutTests/platform/ios/imported/w3c/web-platform-tests/editing/other/exec-command-with-text-editor.tentative_type=textarea-expected.txt:
(WebKit::WebPage::completeSyntheticClick):
(WebKit::WebPage::executeEditCommandWithCallback):
(WebKit::WebPage::runJavaScriptInFrameInScriptWorld):
(WebKit::WebPage::updateRendering):
(WebKit::WebPage::resetFocusedElementForFrame):
(WebKit::WebPage::elementDidRefocus):
(WebKit::WebPage::elementDidFocus):
(WebKit::WebPage::elementDidBlur):
* Source/WebKit/WebProcess/WebPage/ios/WebPageIOS.mm:
(WebKit::WebPage::requestFocusedElementInformation):
(WebKit::WebPage::focusedElementInformation):
(WebKit::WebPage::focusedElementInformationWithoutLayout):
(WebKit::WebPage::emitDeferredFocusedElementUpdate):
(WebKit::WebPage::flushPendingFocusedElementUpdateIfNeeded):
(WebKit::WebPage::focusTextInputContextAndPlaceCaret):

Canonical link: <a href="https://commits.webkit.org/315733@main">https://commits.webkit.org/315733@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/6134353175996da4ffe5fe7036b5fc7b0de8fec2

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/169829 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/43751 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/37141 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/179188 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/127541 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/122c250c-cee0-445d-8f4f-f143dd038591) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/44817 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/43381 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/132356 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/93255 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/5233638d-0802-4d27-80c3-d6d045e3a464) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/172788 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/162/builds/35458 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/152765 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/112858 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/ae5e821e-bb23-4468-b6c2-dfafc3fa5574) 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/34142 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/32497 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/27886 "Built successfully") | | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/144532 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/32164 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/181787 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/34495 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/36663 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/140809 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/43407 "Built successfully") | | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/140640 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/37885 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/44096 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/29144 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/108391 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/35907 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/115861 "Built successfully") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/42607 "Built successfully") | [  ~~🧪 mac-site-isolation~~](https://ews-build.webkit.org/#/builders/185/builds/7244 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/41999 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/42254 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/42142 "Built successfully") | | | |

### DIFF
--- a/Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
+++ b/Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
@@ -3408,6 +3408,9 @@ void WebPage::completeSyntheticClick(std::optional<WebCore::FrameIdentifier> fra
     if ((!handledPress && !handledRelease) || !nodeRespondingToClick.isElementNode())
         send(Messages::WebPageProxy::DidNotHandleTapAsClick(roundedIntPoint(location)));
 
+#if PLATFORM(IOS_FAMILY)
+    flushPendingFocusedElementUpdateIfNeeded();
+#endif
     send(Messages::WebPageProxy::DidCompleteSyntheticClick());
 
 #if PLATFORM(IOS_FAMILY)
--- a/Source/WebKit/WebProcess/WebPage/WebPage.cpp
+++ b/Source/WebKit/WebProcess/WebPage/WebPage.cpp
@@ -1848,6 +1848,9 @@ void WebPage::changeFont(WebCore::FontChanges&& changes)
 void WebPage::executeEditCommandWithCallback(const String& commandName, const String& argument, CompletionHandler<void()>&& completionHandler)
 {
     executeEditCommand(commandName, argument);
+#if PLATFORM(IOS_FAMILY)
+    flushPendingFocusedElementUpdateIfNeeded();
+#endif
     completionHandler();
 }
 
@@ -4925,6 +4928,9 @@ void WebPage::runJavaScriptInFrameInScriptWorld(RunJavaScriptParameters&& parame
             WEBPAGE_RELEASE_LOG_ERROR(Process, "runJavaScriptInFrameInScriptWorld: Request to run JavaScript failed with error %" PRIVATE_LOG_STRING, result.error()->message.utf8().data());
         else
             WEBPAGE_RELEASE_LOG(Process, "runJavaScriptInFrameInScriptWorld: Request to run JavaScript succeeded");
+#if PLATFORM(IOS_FAMILY)
+        flushPendingFocusedElementUpdateIfNeeded();
+#endif
         completionHandler(WTF::move(result));
     });
 }
@@ -5397,6 +5403,9 @@ void WebPage::updateRendering()
     protect(corePage())->updateRendering();
 
 #if PLATFORM(IOS_FAMILY)
+    if (auto pendingUpdate = std::exchange(m_pendingFocusedElementUpdate, { }))
+        emitDeferredFocusedElementUpdate(WTF::move(*pendingUpdate));
+
     findController().redraw();
     foundTextRangeController().redraw();
 #endif
@@ -7599,6 +7608,7 @@ void WebPage::resetFocusedElementForFrame(WebFrame* frame)
 
     if (frame->isMainFrame() || m_focusedElement->document().frame() == frame->coreLocalFrame()) {
 #if PLATFORM(IOS_FAMILY)
+        m_pendingFocusedElementUpdate = std::nullopt;
         m_sendAutocorrectionContextAfterFocusingElement = false;
         send(Messages::WebPageProxy::ElementDidBlur());
 #elif PLATFORM(MAC)
@@ -7612,8 +7622,15 @@ void WebPage::elementDidRefocus(Element& element, const FocusOptions& options)
 {
     elementDidFocus(element, options);
 
-    if (m_userIsInteracting)
-        scheduleFullEditorStateUpdate();
+    if (!m_userIsInteracting)
+        return;
+
+#if PLATFORM(IOS_FAMILY)
+    if (m_pendingFocusedElementUpdate)
+        return;
+#endif
+
+    scheduleFullEditorStateUpdate();
 }
 
 bool WebPage::shouldDispatchUpdateAfterFocusingElement(const Element& element) const
@@ -7743,12 +7760,7 @@ void WebPage::elementDidFocus(Element& element, const FocusOptions& options)
         if (isChangingFocusedElement && (m_userIsInteracting || m_keyboardIsAttached))
             m_sendAutocorrectionContextAfterFocusingElement = true;
 
-        auto information = focusedElementInformation();
-        if (!information)
-            return;
-
         RefPtr<API::Object> userData;
-
         m_formClient->willBeginInputSession(this, &element, protect(WebFrame::fromCoreFrame(*protect(element.document().frame()))).get(), m_userIsInteracting, userData);
 
         if (!userData) {
@@ -7758,9 +7770,17 @@ void WebPage::elementDidFocus(Element& element, const FocusOptions& options)
                     userData = userDataFromJSONData(*data);
             }
         }
-
-        information->preventScroll = options.preventScroll;
-        send(Messages::WebPageProxy::ElementDidFocus(information.value(), m_userIsInteracting, m_recentlyBlurredElement, m_lastActivityStateChanges, UserData(WebProcess::singleton().transformObjectsToHandles(userData.get()).get())));
+        m_pendingFocusedElementUpdate = PendingFocusedElementUpdate {
+            .element = element,
+            .options = options,
+            .userIsInteracting = m_userIsInteracting,
+            .recentlyBlurredElementSnapshot = m_recentlyBlurredElement,
+            .activityStateChanges = m_lastActivityStateChanges,
+            .userData = WTF::move(userData),
+        };
+        if (RefPtr formControlElement = dynamicDowncast<HTMLFormControlElement>(element))
+            m_pendingFocusedElementUpdate->isFocusingWithValidationMessage = formControlElement->isFocusingWithValidationMessage();
+        scheduleFullEditorStateUpdate();
 #elif PLATFORM(MAC)
         // FIXME: This can be unified with the iOS code above by bringing ElementDidFocus to macOS.
         // This also doesn't take other noneditable controls into account, such as input type color.
@@ -7773,6 +7793,9 @@ void WebPage::elementDidFocus(Element& element, const FocusOptions& options)
 void WebPage::elementDidBlur(WebCore::Element& element)
 {
     if (m_focusedElement == &element) {
+#if PLATFORM(IOS_FAMILY)
+        m_pendingFocusedElementUpdate = std::nullopt;
+#endif
         m_recentlyBlurredElement = WTF::move(m_focusedElement);
         callOnMainRunLoop([protectedThis = Ref { *this }] {
             if (protectedThis->m_recentlyBlurredElement) {
--- a/Source/WebKit/WebProcess/WebPage/ios/WebPageIOS.mm
+++ b/Source/WebKit/WebProcess/WebPage/ios/WebPageIOS.mm
@@ -948,6 +948,7 @@ void WebPage::handleDoubleTapForDoubleClickAtPoint(const IntPoint& point, Option
 
 void WebPage::requestFocusedElementInformation(CompletionHandler<void(const std::optional<FocusedElementInformation>&)>&& completionHandler)
 {
+    flushPendingFocusedElementUpdateIfNeeded();
     std::optional<FocusedElementInformation> information;
     if (m_focusedElement)
         information = focusedElementInformation();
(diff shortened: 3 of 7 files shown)


Source/WebKit/WebProcess/WebPage/WebPage.cpp
smfr: A nicer way to write this would be
```
m_pendingFocusedElementUpdate = PendingFocusedElementUpdate {
    .element = element,
    .pending.options = options,
    ...
};
```

smfr: `tookDeferredPath` needs a better name.

NathanSolomon733: Done.

NathanSolomon733: I decided to just remove the `tookDeferredPath` local variable by reverting `elementDidFocus` back to void . On re-read I think it is now confusing that I made `elementDidFocus` return a bool that tells you if you deferred sending the IPC message. It has nothing to do with the name of that function. 

We are trying to decide if this function should early return here, so I changed it to instead check the state of WebPage's member variable `m_pendingFocusedElementUpdate` that gets set if we took the fast path or not. I think its clearer to read the state of that member based on its name instead of reading the bool result of `elementDidFocus`

Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
Andrew-Fryer, line 3412: IIUC, this is needed to preserve ordering of IPC message to UI Process.
Is there a way to verify that you've found all the places we need to do this?

Source/WebKit/WebProcess/WebPage/ios/WebPageIOS.mm
Andrew-Fryer, line 2641: if `focusedElement` is null, wouldn't the existing code break on `if (auto elementContext = contextForElement(*focusedElement))`?
Was static analysis upset without this check?

NathanSolomon733: It wasn't upset without the check. I put the check in because I deference focusedElement on the line right after in `focusedElementInformationAssumingLayoutClean(*focusedElement);`, but I realize now that this check is not needed. 
Before entering `focusedElementInformation()`, there is already an existing check for 
`    if (!m_focusedElement)
        return;`

There's just an assumed but not enforced assumption in the code right now that focusedElementInformation() is called before a nullptr check on m_focusedElement first

Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
NathanSolomon733: There's no way to guarantee this other than reasoning and testing what paths need focused element information eagerly. We could make the patch safer by only enabling the deferred path for if m_isUserInteracting is false, meaning if the request to focus the element comes from a page-driven/WCP-side-only driven click. 
For a single, manual real user initiated click, this patch would not be noticeable to performance. It would have a benefit if there is a programmatic focusing or cycling through traversing focusable elements on popular sites which might happen in the autofill implementation.

Source/WebKit/WebProcess/WebPage/ios/WebPageIOS.mm
whsieh:
```suggestion
    RefPtr document = focusedOrMainFrame->document();
```

whsieh: Maybe `focusedElementInformationWithoutLayout`, to be consistent with `boundingAbsoluteRectWithoutLayout`

whsieh, line 2653: Should this assert that `element` is the same as the current `m_focusedElement`?

Source/WebKit/WebProcess/WebPage/WebPage.cpp
whsieh:
```suggestion
if (auto pendingUpdate = std::exchange(m_pendingFocusedElementUpdate, { })) {
```

Source/WebKit/WebProcess/WebPage/Cocoa/WebPageCocoa.mm
whsieh: This also makes me a bit nervous — the relationship between FES and `EditorState` is pretty central to how keyboard focus works on iOS, for instance (FES arrives first, followed by ES in the next remote layer tree commit at the latest).

Is there a risk that we end up with FES in the UI process, without a corresponding `EditorState`? (This would manifest as failing to zoom to reveal the focused element after bringing up the software keyboard on iOS, for instance).

whsieh (approved):

Source/WebKit/WebProcess/WebPage/WebPage.cpp
NathanSolomon733: Done

Source/WebKit/WebProcess/WebPage/ios/WebPageIOS.mm
NathanSolomon733: Yes, I just added an ASSERT. There is already an early return right before the call sites of `focusedElementInformationWithoutLayout()` checking if `element != m_focusedElement`, but debug assert is better to enforce it.

NathanSolomon733: Done

NathanSolomon733:
> Is there a risk that we end up with FES in the UI process, without a corresponding EditorState? (This would manifest as failing to zoom to reveal the focused element after bringing up the software keyboard on iOS, for instance).

On the deferred path, I do not think it is possible to send the deferred ElementDidFocus IPC without an EditorState IPC to be sent because when first capturing the pending state of the focused element in `WebPage::elementDidFocus`, we also call `scheduleFullEditorStateUpdate`. So there will be an `EditorState` sent in the same rendering update as when `ElementDidFocus` is sent. And the code ordering in `RemoteLayerTreeDrawingArea::updateRendering()` is that we call `emitDeferredFocusedElementUpdate` via `webPage->updateRendering()`, then commit another `EditorState` via layer tree commit. 

It is possible that somehow an `EditorState` IPC gets sent first, and then some time later a pending `ElementDidFocus` IPC gets sent after, which would change the existing ordering. But I don't think this is an issue because if an immediate
` sendEditorStateUpdate()` call sent an `EditorState` message ahead of `ElementDidFocus`, that function calls `WebPageProxy::editorStateChanged`, which calls `WebPageProxy::dispatchDidUpdateEditorState`, and this function early returns if `m_waitingForPostLayoutEditorStateUpdateAfterFocusingElement` is not set. This bool is set if the UI process has processed `ElementDidFocus` because it is set in `_elementDidFocus` which is the UI Process handler. So nothing would happen if `EditorState` reached the UI process before `ElementDidFocus` did. It will also return early if `!editorState().hasPostLayoutData()`. So, a pre-focus `EditorState` is ignored, and the reveal of the focused element is then fulfilled by the next post-layout `EditorState` committed in the same rendering update as `ElementDidFocus` (which maintains the right order).

PR #65408, opened 2026-05-21
[Swift GTK] Switch from WebKit_Internal module to bridging header

adetaylor:
#### 85f07a8b5f6d6ad5858e90314c84990af0ee2701
<pre>
[Swift GTK] Switch from WebKit_Internal module to bridging header
<a href="https://bugs.webkit.org/show_bug.cgi?id=315278">https://bugs.webkit.org/show_bug.cgi?id=315278</a>

Reviewed by NOBODY (OOPS!).

To use project-internal C++ code from Swift, Swift provides two options - use a
modulemap or use an internal bridging header. This PR switches from the first
to the second approach. This is now generally regarded as best practice, but
there is a forcing function to adopt this change - Linux/GTK builds of Swift do
not work with the modulemap approach. This is because they reveal a
pre-existing problem with our modulemaps, in that there is a circular
dependency between the WebKit_Internal module and the WebKit_DerivedSources_IPC
module. This was previously invisible on Apple platforms because of the
inadvertent lack of a (non-private) modulemap for the latter module, which
meant it was largely being ignored in Xcode builds. Our options to rectify this
are to add a public empty modulemap, and resolve the circular dependency, or to
switch to a bridging header. Resolving the circular dependency is very hard
without splitting WebKit_Internal into multiple layers, so switching to the
bridging header approach is best at this point.

This change does not impact WTF, JSC, bmalloc, or public WebKit modulemaps.
It strictly affects the internal relationships between Swift and C++ in WebKit.

It switches to this new approach only on the most recent SDKs for Xcode builds
because bridging headers have only recently become supported for framework
targets.

Components of this change:

- Add Modules/Internal/WebKitInternalBridgingHeader.h. Xcode picks it up
  via SWIFT_OBJC_BRIDGING_HEADER and SWIFT_BRIDGING_HEADER_IS_INTERNAL=YES,
  on the most recent SDKs.

- Drop `import WebKit_Internal` from every Swift source and from the
  message-receiver template; bridging-header symbols are visible implicitly.
  Stop generating the WebKit_DerivedSources_IPC modulemap and its
  golden/unit-test, again only on the most recent SDKs.

- CMake: WEBKIT_SETUP_SWIFT_AND_GENERATE_SWIFT_CPP_INTEROP_HEADER takes an
  optional bridging-header path that becomes -internal-import-bridging-header.
  Add UIProcess/API/ios, Platform/spi/visionos, Shared/ios and
  ModelProcess/cocoa to the Cocoa private include dirs so the header&apos;s #imports
  resolve under CMake (Xcode handles this via header maps).

- Enable -enable-library-evolution on the CMake Mac port to match Xcode&apos;s
  BUILD_LIBRARY_FOR_DISTRIBUTION=YES. This avoids a warning.

* Source/WebKit/Configurations/WebKit.xcconfig:
* Source/WebKit/DerivedSources-output.xcfilelist:
* Source/WebKit/ModelProcess/cocoa/WKUSDStageConverter.swift:
* Source/WebKit/Modules/Internal/WebKitSwiftBridgingHeader.h: Added.
* Source/WebKit/Platform/cocoa/FloatRectCG.swift:
* Source/WebKit/Platform/cocoa/IntRectCG.swift:
* Source/WebKit/Platform/cocoa/WKMaterialHostingSupport.swift:
* Source/WebKit/Platform/spi/visionos/WKSurroundingsEffect.swift:
* Source/WebKit/Platform/spi/visionos/WKSurroundingsEffectView.swift:
* Source/WebKit/PlatformCocoa.cmake:
* Source/WebKit/PlatformIOS.cmake:
(generate_swift_message_handler):
* Source/WebKit/Scripts/webkit/tests/MessageArgumentDescriptions.cpp:
* Source/WebKit/Scripts/webkit/tests/TestWithSwiftConditionallyMessageReceiver.swift:
* Source/WebKit/Scripts/webkit/tests/TestWithSwiftMessageReceiver.swift:
* Source/WebKit/Shared/API/APIArray.swift:
* Source/WebKit/Shared/IPCTesterReceiver.swift:
* Source/WebKit/Shared/Model/WKStageModeOrbitSimulator.swift:
* Source/WebKit/Shared/WebKit-Swift.h:
* Source/WebKit/UIProcess/API/Cocoa/_WKTextExtraction.swift:
* Source/WebKit/UIProcess/API/Swift/URLSchemeHandler.swift:
* Source/WebKit/UIProcess/API/Swift/WebPage.swift:
* Source/WebKit/UIProcess/Cocoa/Separated/WKSeparatedImageView+Analysis.swift:
* Source/WebKit/UIProcess/Cocoa/Separated/WKSeparatedImageView+Generation.swift:
* Source/WebKit/UIProcess/Cocoa/Separated/WKSeparatedImageView+Rendering.swift:
* Source/WebKit/UIProcess/Cocoa/Separated/WKSeparatedImageView+Surface.swift:
* Source/WebKit/UIProcess/Cocoa/Separated/WKSeparatedImageView.swift:
* Source/WebKit/UIProcess/Cocoa/TextExtraction/WKWebView+TextExtraction.swift:
* Source/WebKit/UIProcess/Cocoa/WKDeferringGestureRecognizer.swift:
* Source/WebKit/UIProcess/Cocoa/WKScrollGeometryAdapter.swift:
* Source/WebKit/UIProcess/Cocoa/WKTextEffectManager+VersionCheck.swift:
* Source/WebKit/UIProcess/Cocoa/WKUIDelegateAdapter.swift:
* Source/WebKit/UIProcess/Cocoa/WebPageWebView.swift:
* Source/WebKit/UIProcess/PDF/WKPDFHUDView.swift:
* Source/WebKit/UIProcess/StdlibExtras.swift:
* Source/WebKit/UIProcess/SwiftDemoLogo.swift:
* Source/WebKit/UIProcess/WKMouseDeviceObserver.swift:
* Source/WebKit/UIProcess/WebBackForwardList.swift:
* Source/WebKit/UIProcess/WebPageProxy.swift:
* Source/WebKit/UIProcess/mac/WKTextSelectionController.swift:
</pre>
https://github.com/WebKit/WebKit/commit/221ca2374c8d330e99acc96959e57d61277a91f2

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/165700 "4 style errors") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/39190 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/32771 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/174742 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/122955 "Built successfully") | [❌ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/356c0e4c-91c9-4c5a-afdb-ae4fa09ce133) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/39654 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/39194 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/128766 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/90688 "Passed tests") | [❌ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/9eb85ff1-be97-4570-b094-b88d1c5343ad) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/168659 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/31190 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/148871 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/109290 "Passed tests") | | [❌ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/74277898-1f4a-41b3-93c9-bf937a4d1e46) 
| [❌ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/165018 "Found 1 webkitpy test failure: webkit.messages_unittest.GeneratedFileContentsTest.test_message_argument_description") | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/30166 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/28782 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/22823 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/20/builds/157857 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/140135 "Passed tests") | | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/177266 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/26593 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/30181 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/28275 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/137096 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/39259 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/32928 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/137025 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/39947 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/148365 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/102458 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/32315 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/25232 "Passed tests") | [  ~~🛠 jsc-armv7~~](https://ews-build.webkit.org/#/builders/35/builds/198849 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/38458 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/111531 "Built successfully") | [  ~~🧪 jsc-armv7-tests~~](https://ews-build.webkit.org/#/builders/35/builds/198849 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/37854 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/3313 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/38100 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/37998 "Built successfully") | | | |


Source/WebKit/Modules/Internal/WebKitSwiftBridgingHeader.h
```diff
@@ -0,0 +1,131 @@
...
+ */
+
+// Bridging header used by SWIFT_OBJC_BRIDGING_HEADER to expose WebKit's
+// internal Objective-C[++] and C++ types to the framework's own Swift sources.
+// Combined with SWIFT_BRIDGING_HEADER_IS_INTERNAL=YES, these declarations are
+// only visible inside the WebKit framework target.
+//
+// This file is the "main file" of the bridging-header precompilation, so no
+// `#pragma once` / include guard is needed (and adding one would trigger
+// -Wpragma-once-outside-header).
```
geoffreygaren, line 33: We are inside a header. Why would -Wpragma-once-outside-header trigger?

Source/WebKit/Configurations/WebKit.xcconfig
```diff
@@ -83,8 +83,14 @@ BUILD_LIBRARY_FOR_DISTRIBUTION = YES;
 SWIFT_INSTALL_OBJC_HEADER = NO;
 SWIFT_LIBRARY_LEVEL = api;
 
+// WebKit's internal C++/Objective-C[++] surface is exposed to Swift via a
+// bridging header. SWIFT_BRIDGING_HEADER_IS_INTERNAL keeps these symbols
+// visible only to the framework's own Swift sources.
+SWIFT_OBJC_BRIDGING_HEADER = $(SRCROOT)/Modules/Internal/WebKitInternalBridgingHeader.h;
```
geoffreygaren: Let's put Swift in the name, since Swift is not the only technology that has the concept of 'bridge'. Maybe `WebKitInternalSwiftBridgingHeader.h`.

Also, is this in fact called "WebKit Internal" anymore, or was that the module way of lookin gat the world? Maybe it's just `WebKitSwiftBridgingHeader.h`. That sounds nice.

Put another way, if there is no `WebKitExternalSwiftBridgingHeader.h`, now or in the future, then I would like to drop "Internal".

Source/WebKit/Modules/Internal/WebKitInternalBridgingHeader.h
```diff
@@ -0,0 +1,131 @@
...
+#import "WebPageProxyInternals.h"
+#import "WebPermissionControllerProxy.h"
+#import "WebProcessActivityState.h"
+#import "WebProcessProxy.h"
+#import "WebPushMessage.h"
+#import "WebsiteData.h"
+
+// Generated message-receiver headers.
+#import "IPCTesterReceiverMessages.h"
+#import "WebBackForwardListMessages.h"
```
geoffreygaren: Why are these separate?

Tools/Scripts/webkitpy/style/checker.py
```diff
@@ -254,6 +254,14 @@
...
      ["-build/include_order"]),
 
+    ([
+     # The bridging header groups its #imports by visibility/audit-region role,
+     # not alphabetically. Headers that open NS_HEADER_AUDIT_BEGIN(..., sendability)
+     # must precede the C++ headers that would otherwise put `enum class`es in
+     # scope before the audit pragma. See the in-file comment.
+     os.path.join('Source', 'WebKit', 'Modules', 'Internal', 'WebKitInternalBridgingHeader.h')],
+     ["-build/include_order"]),
+
```
geoffreygaren: It may be best to leave this out. The style checker only checks the code you edit, and it might be nice for future edits to retain alphabetical sorting locally, even if the file isn't sorted globally.

geoffreygaren (approved):

Tools/Scripts/webkitpy/style/checker.py
adetaylor: Will do!

Source/WebKit/Modules/Internal/WebKitInternalBridgingHeader.h
adetaylor: Historical - they're _generated_ code so were previously in a different module, but you're right, there's no real reason to separate them now - I'll lump them together.

Source/WebKit/Configurations/WebKit.xcconfig
adetaylor: OK will do

Source/WebKit/Modules/Internal/WebKitSwiftBridgingHeader.h
adetaylor: I don't know, but it does. I'll adjust the comment to be more despairing.

adetaylor:
Status here:
* The problems with open source CI are hopefully because they're using a version of the SDK before this support was added for framework bridging headers. I've made the change conditional so it only switches to a bridging header on the most recent SDK.
* The problems with Apple internal CI are unknown, and there's a bug in the CI so I can't figure out why. Working on it.
* I have yet to address code review comments.
* Conflicts need resolving too!
* Will fix all this next week

Source/WebKit/Modules/Internal/WebKitSwiftBridgingHeader.h
```diff
@@ -0,0 +1,136 @@
+/*
```
rr-codes, line 1: this file should be named `WebKit-Bridging-Header.h` since that's the convention. Additionally, it should be top level in WebKit instead of in Modules/Internal

Source/WebKit/Modules/Internal/WebKitSwiftBridgingHeader.h
```diff
@@ -0,0 +1,136 @@
...
+//     Clang demands every raw pointer in the whole TU also be annotated;
+//     that "virality" is a known limitation in WebKit.
+//   -Wpragma-clang-attribute and -Wignored-attributes — the audit-using
+//     headers in the first block below open
+//     NS_HEADER_AUDIT_BEGIN(..., sendability), whose attribute push
+//     complains about record/enum declarations already in scope from
+//     headers each SPI header itself transitively #imports. The push
+//     still applies correctly to the contents of the audit region.
+IGNORE_CLANG_WARNINGS_BEGIN("nullability-completeness")
+IGNORE_CLANG_WARNINGS_BEGIN("pragma-clang-attribute")
```
rr-codes, line 49: even with the comment, I'm still a bit confused by ignoring these? what errors happen otherwise?

Source/WebKit/Modules/Internal/WebKitSwiftBridgingHeader.h
```diff
@@ -0,0 +1,136 @@
...
+// Bridging header used by SWIFT_OBJC_BRIDGING_HEADER to expose WebKit's
+// internal Objective-C[++] and C++ types to the framework's own Swift sources.
+// Combined with SWIFT_BRIDGING_HEADER_IS_INTERNAL=YES, these declarations are
+// only visible inside the WebKit framework target.
+//
+// This file is the "main file" of the bridging-header precompilation, so no
+// `#pragma once` / include guard is needed (and adding one seems to trigger
+// -Wpragma-once-outside-header).
+
+#include <wtf/Compiler.h>
```
rr-codes, line 35: it is invalid for a bridging header's contents to include other frameworks/libraries

Source/WebKit/Modules/Internal/WebKitSwiftBridgingHeader.h
```diff
@@ -0,0 +1,136 @@
...
+#include <wtf/Platform.h>
+
+// Suppressed across all the #imports below:
+//   -Wnullability-completeness   — once any #imported header is annotated,
```
rr-codes, line 39: (I still think this shouldn't be ignored fwiw but that's just my opinion)

Source/WebKit/Modules/Internal/WebKitSwiftBridgingHeader.h
adetaylor: The requested name disagrees with @geoffreygaren 's prior request, but I'll move it as you suggest.

adetaylor:
Current situation: I now understand the problems on Apple internal CI, and they're not very easy to resolve. I'm working on it.

Source/WebKit/Modules/Internal/WebKitSwiftBridgingHeader.h
geoffreygaren: @rr-codes What is the nature of the convention, and where is it documented?

adetaylor:
Further update. I have worked around the *need* for this pull request to get GTK Swift builds working. Hooray.

This is good news because the problems with this bridging header approach are substantial. I'm going to spend some time tomorrow writing them up, then close this pull request, but point some Swift/module experts at my write-up to see if it's solvable. Fortunately, it's not urgent any longer.

adetaylor:
OK, here's the state of play here.

## Why I cared about this in the first place

There are two reasons:
* The Swift GTK build revealed a circular dependency between `WebKit_Internal` and `WebKit_DerivedSources_IPC` modules, which hadn't been revealed on Apple platforms due to an unrelated bug (that unrelated bug made `WebKit_DerivedSources_IPC` entirely ineffective). #66243 removes `WebKit_DerivedSources_IPC` and unblocks our ability to get Swift GTK builds further. (For what it's worth, the circular dependency was things like   `WebPageProxy.h`  ──includes──▶  `WebBackForwardListMessages.h`     (Internal → IPC), `WebPageProxyMessages.h`  ──includes──▶  `WebPageCreationParameters.h`   (IPC → Internal), but it was generally indicative of great coupling). So I _no longer_ need us to switch to a bridging header for the sake of GTK builds: this reason is now moot.
* Clang modules cause a certain amount of friction and we are keen to see if we can minimise further trouble - @geoffreygaren was especially keen that we move to bridging headers on that basis.

## The scope of this change

Specifically, this change was to discard just the `WebKit_Internal` and `WebKit_DerivedSources_IPC` modules, and replace them with a bridging header. It made no difference to other pre-existing modules (PAL, WTF, bmalloc, etc.)

## Why this change doesn't work as-is

Unfortunately, on the face of it, the bridging header approach doesn't work.

We get errors such as:

```
❯ DerivedSources/WebKit/WKWebView+SomeInternalOverlay.swift:37:11: error: cannot use class 'WKWebView' in an extension with public or '@usableFromInline'      
  members; it was imported via the internal bridging header                                                                                                                                                                                   
  extension WKWebView : AppKit.NSRefreshControlHosting {                                                                                                                                                                                      
            ^                                                                                                                                                                                                                                 
  Source/WebKit/UIProcess/API/Cocoa/WKWebView.h:73:12: note: class declared here                                                                                              
  @interface WKWebView : NSView                                                                                                                                                                                                               
             ^                                
```

and

```
Source/WebKit/UIProcess/API/Swift/WebPage+FrameInfo.swift:48:20: error: property cannot be declared public because its type uses an internal type                           
          public var securityOrigin: WKSecurityOrigin { wrapped.securityOrigin }                                                                                                                                                              
                     ^               ~~~~~~~~~~~~~~~~   
```

along with a variety of problems caused by interpreting ints as uints or vice-versa.

This is because the various headers from the _public_ `WebKit` module, are instead being pulled in via the bridging header. `WKWebView.h` should resolve from the public WebKit modulemap (as a `[system]`, public-visibility module type), but it's being textually absorbed into the internal bridging header instead. (The `[system]` attribute is responsible for the int vs uint issues).

## What ways forward there could be

* There was some suggestion that fiddling with `-import-underlying-module` might allow this to work, but I had no success so far
* Significant WebKit header restructuring, so that internal headers such as `WebExtensionController.h` cease to include public headers such as `WKWebExtensionController.h`
* Consult with the Swift and/or modules folks to figure out if this behavior is expected and to tell us definitively whether bridging headers can work or not.

I'm going to close this PR and ask a couple of folks to comment. It's quite possible I've done something daft - we'll see. The good news is that this no longer blocks the GTK work.

geoffreygaren:
> Significant WebKit header restructuring, so that internal headers such as WebExtensionController.h cease to include public headers such as WKWebExtensionController.h

FWIW, this option is probably un-possible.

PR #65493, opened 2026-05-22
Web Inspector: hovering over a node in a preview for a collection should highlight it

tangalice:
#### 20b0d2d5de071e54fb873a539241cae9eeac9b11
<pre>
Web Inspector: hovering over a node in a preview for a collection should highlight it
<a href="https://bugs.webkit.org/show_bug.cgi?id=143206">https://bugs.webkit.org/show_bug.cgi?id=143206</a>
<a href="https://rdar.apple.com/20341722">rdar://20341722</a>

Reviewed by Devin Rousso.

Hovering over a DOM node in a collection preview (Set, Map) didn&apos;t
highlight it on the page (works fine for arrays).

Arrays render through `_appendPropertyPreviews`, which already handled
node entries with a `remoteObjectAccessor`. Collection entries went
through `_appendValuePreview` instead, which only provided an accessor
when rendering the top level object, never for individual entry sub
previews. Without an accessor, `createElementForNodePreview` has no way
to resolve the node, so it skips attaching the hover listeners entirely.

The fix adds `_appendCollectionEntryNodePreview`, which handles node
entries directly by calling `getCollectionEntries({fetchStart:
entryIndex, fetchCount: 1})` to fetch the live `RemoteObject` for that
specific entry and pass it as the accessor. Non-node entries fall
through to `_appendPreview` as before. WeakMap and WeakSet are excluded
because `fetchStart` has no effect on them, so positional lookup is
not reliable.

* Source/WebInspectorUI/UserInterface/Views/ObjectPreviewView.js:
(WI.ObjectPreviewView.prototype._appendEntryPreviews):
(WI.ObjectPreviewView.prototype._appendCollectionEntryNodePreview):

Canonical link: <a href="https://commits.webkit.org/314589@main">https://commits.webkit.org/314589@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/8231659d800608be41fd7f760eec2e6e44ebc78a

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/166265 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/39755 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/33336 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/175313 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/123520 "Built successfully") 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/40181 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/39759 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/129232 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/91023 "Passed tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/169224 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/31619 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/149388 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/109757 "Passed tests") | 
| | [loading 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/30597 "Build is in progress. Recent messages:OS: Tahoe (26.2), Xcode: 26.2; Skipping applying patch since patch_id isn't provided; Checked out pull request; Skipped layout-tests; 4 new passes 2 flakes") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/29288 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/23453 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/140565 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/27085 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/177845 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/30747 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/28791 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/137584 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/39824 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/33433 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/137509 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/37118 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/40512 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/148881 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/103152 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/32806 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/25748 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/39023 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/112097 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/38420 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/3834 "Passed tests") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/38665 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/38563 "Built successfully") | | |

### DIFF
--- a/Source/WebInspectorUI/UserInterface/Views/ObjectPreviewView.js
+++ b/Source/WebInspectorUI/UserInterface/Views/ObjectPreviewView.js
@@ -166,11 +166,11 @@ WI.ObjectPreviewView = class ObjectPreviewView extends WI.Object
             var keyPreviewLossless = true;
             var entry = preview.collectionEntryPreviews[i];
             if (entry.keyPreview) {
-                keyPreviewLossless = this._appendPreview(element, entry.keyPreview);
+                keyPreviewLossless = this._appendCollectionEntryNodePreview(element, entry.keyPreview, preview, i, true);
                 element.append(" => ");
             }
 
-            var valuePreviewLossless = this._appendPreview(element, entry.valuePreview);
+            var valuePreviewLossless = this._appendCollectionEntryNodePreview(element, entry.valuePreview, preview, i, false);
 
             if (!keyPreviewLossless || !valuePreviewLossless)
                 lossless = false;
@@ -190,6 +190,24 @@ WI.ObjectPreviewView = class ObjectPreviewView extends WI.Object
         element.append(isIterator ? "]" : "}");
 
         return lossless;
+    }
+
+    _appendCollectionEntryNodePreview(element, entryPreview, collectionPreview, entryIndex, isKey)
+    {
+        if (entryPreview.subtype === "node" && collectionPreview === this._preview && this._object && !this._preview.isWeakCollection()) {
+            let options = {
+                remoteObjectAccessor: (callback) => {
+                    this._object.getCollectionEntries((entries) => {
+                        let remoteObject = isKey ? entries[0].key : entries[0].value;
+                        WI.consoleManager.releaseRemoteObjectWithConsoleClear(remoteObject);
+                        callback(remoteObject);
+                    }, {fetchStart: entryIndex, fetchCount: 1});
+                },
+            };
+            element.appendChild(WI.FormattedValue.createElementForNodePreview(entryPreview, options));
+            return false;
+        }
+        return this._appendPreview(element, entryPreview);
     }
 
     _appendPropertyPreviews(element, preview)
--- a/Source/WebInspectorUI/UserInterface/Models/ObjectPreview.js
+++ b/Source/WebInspectorUI/UserInterface/Models/ObjectPreview.js
@@ -70,4 +70,9 @@ WI.ObjectPreview = class ObjectPreview
     {
         return this._size !== undefined && (this._subtype === "array" || this._subtype === "set" || this._subtype === "map" || this._subtype === "weakmap" || this._subtype === "weakset");
     }
+
+    isWeakCollection()
+    {
+        return this._subtype === "weakmap" || this._subtype === "weakset";
+    }
 };


J0pan:
Please format the overflowing lines

Source/WebInspectorUI/UserInterface/Views/ObjectPreviewView.js
```diff
@@ -192,6 +192,28 @@ WI.ObjectPreviewView = class ObjectPreviewView extends WI.Object
...
+                remoteObjectAccessor: (callback) => {
+                    this._object.getCollectionEntries((entries) => {
+                        if (!entries.length)
+                            return;
+                        let remoteObject = isKey ? entries[0].key : entries[0].value;
+                        if (!remoteObject)
+                            return;
+                        WI.consoleManager.releaseRemoteObjectWithConsoleClear(remoteObject);
+                        callback(remoteObject);
+                    }, {fetchStart: entryIndex, fetchCount: 1});
```
J0pan, line 204: This would break for WeakMap / WeakSet. `Protocol/RemoteObject.js` asserts `fetchStart == 0` for weak collections, and weak collections are unordered (`InjectedScriptSource.js:_getWeakMapEntries` / `_getWeakSetEntries` ignore `fetchStart`). For `entryIndex > 0` you'll either trip the assertion or highlight a different node each hover. Add `this._object.isWeakCollection()` and fetch all entries, and then index into the result.

Source/WebInspectorUI/UserInterface/Views/ObjectPreviewView.js
```diff
@@ -192,6 +192,28 @@ WI.ObjectPreviewView = class ObjectPreviewView extends WI.Object
         return lossless;
     }
 
+    _appendCollectionEntryNodePreview(element, entryPreview, entryIndex, isKey)
+    {
+        if (entryPreview.subtype === "node" && this._object) {
```
J0pan: Missing `preview === this._preview`. `_appendEntryPreviews` is reachable recursively through `_appendPreview` at lines 169, 173, and 242. For nested collection preview (e.g. a `Map<key, Set>`), `this._object` is the outer obj, so `getCollectionEntries` would fetch from the wrong collection. `_appendPropertyPreviews `and `_appendValuePreview` already guard with `if (preview === this._preview && this._object)`, you can mirror that here.

Source/WebInspectorUI/UserInterface/Views/ObjectPreviewView.js
```diff
@@ -192,6 +192,28 @@ WI.ObjectPreviewView = class ObjectPreviewView extends WI.Object
...
+            let options = {
+                remoteObjectAccessor: (callback) => {
+                    this._object.getCollectionEntries((entries) => {
```
J0pan, line 200: The defense `if (!entries.length) return;` below is never ran because `RemoteObject.getCollectionEntries` will throw if the backend returns an error. Look at lines 397-399 in  `RemoteObject.js`, it calls `entries.map(...)` without checking whether `entries` is defined, so `.map` will throw a `TypeError` before this callback runs.

@burg should this be a separate rdar against `RemoteObject.js`?

J0pan:
Solid first pass, great work! 🎊

Source/WebInspectorUI/UserInterface/Views/ObjectPreviewView.js
```diff
@@ -192,6 +192,28 @@ WI.ObjectPreviewView = class ObjectPreviewView extends WI.Object
...
+                        if (!entries.length)
+                            return;
+                        let remoteObject = isKey ? entries[0].key : entries[0].value;
```
dcrousso, line 201: i think the issue with this approach is that there's no guarantee that the object hasn't been mutated since it was first previewed
```js
let map = new Map([["a", document.body.children[0]], ["b", document.body.children[1]]]);
console.log(map);
// user expands the preview of the value for "a" that highlights some <foo></foo> in the page
map.set("a", document.body.children[2]);
// user hovers the preview of the value for "a" again which now highlights some <bar></bar> in the page even though the preview shows <foo></foo>
```
ideally we'd get an actual `RemoteObjectId` for the value so that it's something directly tied to it instead of having to re-fetch it from the owner

perhaps we modify `Runtime.getPreview` and/or all the `Runtime` commands that have a `generatePreview` parameter to also have a `bindPreview` (name could be better) that will generate and add a `Runtime.RemoteObjectId` to the `Runtime.ObjectPreview` (which currently doesn't exist and as such is otherwise normally omitted) so that in these cases there's a way to immediately and directly jump to a `Runtime.RemoteObject`?  some part of me feels like that may not play super nicely with how `Runtime.RemoteObject` are created today, but i think we already allow for multiple `Runtime.RemoteObjectId` to map to the same value so maybe it's OK?

tangalice: Good point. I've filed bug [315583](https://bugs.webkit.org/show_bug.cgi?id=315583) to track the proper fix since the mutation race issue is a pre-existing one affecting all preview types. The fix in this PR just brings collections to the same level as arrays and plain objects rather than introducing a new class of issue

Source/WebInspectorUI/UserInterface/Views/ObjectPreviewView.js
dcrousso: we need another check that the collection is not a `WeakMap`/`WeakSet` as that does not guarantee any ordering (and as such does not support a `fetchStart` of anything other than `0`)

Source/WebInspectorUI/UserInterface/Views/ObjectPreviewView.js
```diff
@@ -192,6 +192,28 @@ WI.ObjectPreviewView = class ObjectPreviewView extends WI.Object
...
+                        if (!remoteObject)
+                            return;
```
dcrousso: is it actually possible for us to have an entry where the key/value is not representable as a `WI.RemoteObject`?  based on the protocol and other callers i dont think it is and as such we should drop this

dcrousso (approved):

J0pan (approved):

J0pan:
Due to ews outage yesterday, re-running `mac-wk2`, `mac-AS-debug-wk2`, `mac-intel-wk2`, this pr does not touch upon other suites that were impacted by the outage.

J0pan:
Both `mac-wk2` and `mac-AS-debug-wk2` have passed. Upon examination, the test failures in `mac-intel-wk2` are unrelated. Putting in merge-queue.

Source/WebInspectorUI/UserInterface/Views/ObjectPreviewView.js
dcrousso: along the lines of what @J0pan mentioned in <https://github.com/WebKit/WebKit/pull/65493#discussion_r3291576454>, we would need to pass along the owner `collectionPreview` (which should be available in the caller scope as `preview`) to this and use that instead of `this._preview` in the case of nested values (e.g. `new Map([["foo", new WeakMap([[document.body, {}]])]])`

also, rather than have this inline, lets create a `WI.ObjectPreview.prototype.isWeakCollection` just like the already existing `WI.RemoteObject.prototype.isWeakCollection`

dcrousso, line 200: along the lines of what @J0pan mentioned in <https://github.com/WebKit/WebKit/pull/65493#discussion_r3291576454>, we can't use `this._object` for anything other than the immediate children (as opposed to further down descendants) in the case of nested values (e.g. `new Map([["foo", new WeakMap([[document.body, {}]])]])`

as mentioned in that comment, one solution would be to add a `collectionPreview === this._preview` in the `if` at the top of this method (which is what we already do in other similar places like `_appendPropertyPreviews` and `_appendValuePreview`)

more ideally tho we'd do like i suggested in <https://github.com/WebKit/WebKit/pull/65493/changes#r3291864236> and have some way of "upgrading" from a `Runtime.ObjectPreview` to a real `Runtime.RemoteObject` (or at least getting a `Runtime.RemoteObjectId`)

dcrousso (approved):

PR #65597, opened 2026-05-25
Implement the `application-context` CSS media feature

marcoscaceres:
#### 40004592acd35273b7dc211e579d62e5f05e76b1
<pre>
Implement the `application-context` CSS media feature
<a href="https://bugs.webkit.org/show_bug.cgi?id=314609">https://bugs.webkit.org/show_bug.cgi?id=314609</a>
<a href="https://rdar.apple.com/176851836">rdar://176851836</a>

Reviewed by NOBODY (OOPS!).

Implements the `application-context` CSS media feature, which exposes
whether the current top-level traversable is an installed application
context. Values: `application` and `none` (default).

A document only observes the application context when it and all of its
ancestors up to the top-level traversable are same origin; otherwise the
value is `none`. This is checked with a new
`Document::isSameOriginAsAllAncestors()`, promoted from the identical
strict-origin ancestor walk that was previously local to
DocumentStorageAccess (now migrated to the shared helper). The origin
comparison is strict and does not honor `document.domain`, so a frame
that widened its origin via `document.domain` does not gain access.
Non-screen media types also evaluate to `none`.

Toggling the `Standalone` setting invalidates active media queries and
schedules a rendering update; the `MediaQueryList` change events are then
reported during &quot;update the rendering&quot; (per the spec requirement to
evaluate media queries and report changes when a top-level traversable
becomes, or ceases to be, an application context).

Spec change:
<a href="https://github.com/w3c/manifest/pull/1218">https://github.com/w3c/manifest/pull/1218</a>
</pre>
https://github.com/WebKit/WebKit/commit/40004592acd35273b7dc211e579d62e5f05e76b1

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/176542 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/50477 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/44267 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/186143 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/138824 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/e96a4667-0050-4eaf-81d1-09b95af85fcc) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/178405 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/51769 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/50161 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/137524 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/96258 "5 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/d401f7a2-fae2-438f-9057-ec16dcecc27d) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/179498 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/41010 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/158294 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/117999 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/214cf7fb-821c-4b17-9b47-426771f3b8cb) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/39660 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/38025 "Passed tests") | [⏳ 🧪 jsc-wpe](https://ews-build.webkit.org/#/builders/JSC-Tests-WPE-EWS "Waiting to run tests") | | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/6803 "Passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/150075 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/37792 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/34611 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/37810 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/42210 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/42236 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/188566 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/50142 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/157839 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/146576 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/50826 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/34415 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/146385 "Passed tests") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/41144 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/123030 "Built successfully") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/117103 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/49330 "Built successfully") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/12767 "") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/48732 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/48966 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/48791 "Built successfully") | | | |


Source/WebCore/css/query/MediaQueryFeatures.cpp
```diff
@@ -761,6 +762,32 @@ static const IdentifierSchema& displayModeFeatureSchema()
     };
     return schema;
 }
+
+static const IdentifierSchema& applicationContextFeatureSchema()
+{
+    static MainThreadNeverDestroyed<IdentifierSchema> schema {
+        "application-context"_s,
+        FixedVector { CSSValueBrowser, CSSValueInstalled },
+        MediaQueryDynamicDependency::ApplicationContext,
```
dcrousso, line 779: based on the naming of this parameter i think this is meant to be the `dependencies` needed by this media feature, not the media feature itself

(also as an aside i only see references to `MediaQueryDynamicDependency::Viewport` elsewhere in the code so im not sure this would do anything anyways)

Source/WebCore/page/Page.cpp
```diff
@@ -4336,6 +4333,16 @@ void Page::useSystemAppearanceChanged()
     });
 }
 
+void Page::standaloneChanged()
+{
+    forEachDocument([](auto& document) {
+        document.styleScope().evaluateMediaQueriesForApplicationContextChange();
+        document.updateElementsAffectedByMediaQueries();
+        document.scheduleRenderingUpdate(RenderingUpdateStep::MediaQueryEvaluation);
+        document.evaluateMediaQueriesAndReportChanges();
```
dcrousso: this already happens during `updateRendering`, so why do we need to call this immediately?

Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
```diff
@@ -7967,6 +7921,7 @@ Standalone:
   type: bool
   status: embedder
   webKitLegacyPreferenceKey: WebKitStandalonePreferenceKey
+  webcoreOnChange: standaloneChanged
```
dcrousso, line 8454: is this actually something that can/should ever change at runtime?

i thought that when using "Add to Dock"/"Add to Home Screen" it would create a new app/view and therefore a new page with the ability to set this setting before anything ever gets loaded

Source/WebCore/css/query/MediaQueryFeatures.cpp
marcoscaceres: You're right that nothing reads that value directly. But marking the feature as a dependency here is what tells the engine the query can change at runtime. Without it, CSS `@media (application-context: application)` rules stop updating when the app context changes. I found that out by removing it: `matchMedia` still updated, but the styling no longer re-applied. The viewport reference you saw is a separate special case; the general "is this query dynamic?" check just looks at whether the dependency set is non-empty. So it's doing real work, and I've kept it.

Source/WebCore/page/Page.cpp
marcoscaceres: They're doing two different things. The immediate call re-applies the styling so `@media` rules take effect right away. The scheduled one is what fires the `change` events, which need to happen during rendering rather than synchronously from the setter. This follows the same pattern the appearance and accessibility settings already use.

Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
marcoscaceres: You're right: in Safari this is set once when the app window is created, before anything loads, and nothing changes it afterwards. I kept the change hook because the spec allows the value to change at runtime, and it lets the tests exercise that transition. If you'd rather not support runtime changes until something actually needs it, I'm happy to drop it. What do you think?

dcrousso: id definitely lean towards not having this as otherwise it could lead a future engineer to the wrong conclusion (i.e. this can update at runtime)

but i also dont think there's necessarily any harm, especially if it has tests ensuring the proper behavior :)

marcoscaceres: Kept it, and added a comment to hopefully avoid any future confusion.

PR #65670, opened 2026-05-26
[Skia] Parameterize SkiaCompositingLayer paint walk by mode

nikolaszimmermann:
#### 6811e00c4dcf82c76b4630e16a27355b40daf2f2
<pre>
[Skia] Parameterize SkiaCompositingLayer paint walk by mode
<a href="https://bugs.webkit.org/show_bug.cgi?id=315561">https://bugs.webkit.org/show_bug.cgi?id=315561</a>

Make recursivePaint addressable by mode so any given walk runs in exactly
one mode at a time. Add a private PaintMode on PaintContext with two
values: Paint (content + damage) and DebugIndicators (overlay only).

The public paint() signature is unchanged. Internally it now drives two
walks: a Paint walk for the prior all-in-one behavior minus debug
indicators, and - if the root layer carries indicators - a follow-up
DebugIndicators walk that recurses against the root canvas in tree
order. In DebugIndicators mode paintWithIntermediateSurface() and
paintSelfAndChildrenWithFilterAndMask() short-circuit: no GPU surfaces,
no mask rasterization, no filter application. This places indicators
above filters, masks, and intermediate surfaces of the content walk.

m_layerDamage is only consumed by the Paint walk - the DebugIndicators
walk has nothing to do with damage state and leaves it alone.

This sets up a follow-up that derives the next frame&apos;s damaged region
by walking the same paint recursion against an SkNoDrawCanvas. No
observable behavior change in this commit.

(WebCore::SkiaCompositingLayer::paint):
(WebCore::SkiaCompositingLayer::paintSelf):
(WebCore::SkiaCompositingLayer::paintContents):
(WebCore::SkiaCompositingLayer::collectFrameDamage):
(WebCore::SkiaCompositingLayer::paintDebugIndicators):
(WebCore::SkiaCompositingLayer::paintWithIntermediateSurface):
(WebCore::SkiaCompositingLayer::paintWithMaskAndBackdrop):
(WebCore::SkiaCompositingLayer::paintWithFilterAndMask):
* Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h:

Canonical link: <a href="https://commits.webkit.org/314181@main">https://commits.webkit.org/314181@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/adf67d9655ad9b2217f13f409cc85b44d6587b78

| [loading 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/165278 "Build is in progress. Recent messages:OS: Tahoe (26.2), Xcode: 26.2; Skipping applying patch since patch_id isn't provided; Checked out pull request; check-webkit-style") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/38769 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/32347 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/174321 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/122535 "Built successfully") 
| [loading 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/167146 "Build is in progress. Recent messages:OS: Tahoe (26.2), Xcode: 26.2; Skipping applying patch since patch_id isn't provided; Checked out pull request; Running bindings-tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/39104 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/38770 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/128432 "Passed tests") | [⏳ 🧪 win-tests](https://ews-build.webkit.org/#/builders/Win-Tests-EWS "Waiting to run tests") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/168237 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/30747 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/148493 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/108957 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/29795 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/28412 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/22396 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/139690 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/26191 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/176843 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/29733 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/27896 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/136752 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/38837 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/32550 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/136691 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/39527 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/147987 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/101862 "Built successfully") | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/25154 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/31975 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/24854 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/38037 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/111110 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/37434 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/2932 "Passed tests") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/37680 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/37578 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
+++ b/Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
@@ -426,7 +426,14 @@ bool SkiaCompositingLayer::paint(SkCanvas& canvas, std::optional<Damage>& damage
 {
     bool hasRunningAnimations = computeTransformsAndAnimations({ }, { }, MonotonicTime::now());
     PaintContext context(damage);
+
+    context.mode = PaintMode::Paint;
     recursivePaint(canvas, context);
+
+    if (hasDebugIndicators()) {
+        context.mode = PaintMode::DebugIndicators;
+        recursivePaint(canvas, context);
+    }
 
 #if ENABLE(DAMAGE_TRACKING)
     if (damage && frameDamagePropagationEnabled()) {
@@ -441,8 +448,10 @@ bool SkiaCompositingLayer::paint(SkCanvas& canvas, std::optional<Damage>& damage
 void SkiaCompositingLayer::paintSelf(SkCanvas& canvas, PaintContext& context)
 {
 #if ENABLE(DAMAGE_TRACKING)
+    const bool collectsDamage = context.mode == PaintMode::Paint;
     auto cleanup = WTF::makeScopeExit([&] {
-        m_layerDamage = std::nullopt;
+        if (collectsDamage)
+            m_layerDamage = std::nullopt;
     });
 #endif
 
@@ -455,6 +464,19 @@ void SkiaCompositingLayer::paintSelf(SkCanvas& canvas, PaintContext& context)
     canvas.save();
     canvas.concat(SkM44(transform));
 
+    if (context.mode == PaintMode::Paint) {
+        paintContents(canvas, context);
+#if ENABLE(DAMAGE_TRACKING)
+        collectFrameDamage(canvas, context, transform);
+#endif
+    } else
+        paintDebugIndicators(canvas, context);
+
+    canvas.restore();
+}
+
+void SkiaCompositingLayer::paintContents(SkCanvas& canvas, PaintContext& context)
+{
     SkPaint paint;
     paint.setStyle(SkPaint::kFill_Style);
     auto ctm = canvas.getLocalToDeviceAs3x3();
@@ -510,28 +532,35 @@ void SkiaCompositingLayer::paintSelf(SkCanvas& canvas, PaintContext& context)
                 SkSamplingOptions(SkFilterMode::kLinear, SkMipmapMode::kNone), &paint, SkCanvas::kFast_SrcRectConstraint);
         }
     }
+}
 
 #if ENABLE(DAMAGE_TRACKING)
-    if (frameDamagePropagationEnabled() && context.frameDamage) {
-        auto frameDamage = transform.mapRect(effectiveLayerRect());
-        auto clipBounds = FloatRect(this->clipBounds(canvas, context));
-        if (!clipBounds.isEmpty())
-            frameDamage.intersect(clipBounds);
+void SkiaCompositingLayer::collectFrameDamage(SkCanvas& canvas, PaintContext& context, const TransformationMatrix& transform)
+{
+    if (!frameDamagePropagationEnabled() || !context.frameDamage)
+        return;
 
-        m_previousLayerRectInFrameCoordinates.unite(frameDamage);
+    auto frameDamage = transform.mapRect(effectiveLayerRect());
+    auto clipBounds = FloatRect(this->clipBounds(canvas, context));
+    if (!clipBounds.isEmpty())
+        frameDamage.intersect(clipBounds);
 
-        if (m_layerDamage) {
-            for (const auto& rect : *m_layerDamage) {
-                auto damageRect = transform.mapRect(FloatRect(rect));
-                if (!clipBounds.isEmpty())
-                    damageRect.intersect(clipBounds);
-                context.frameDamage->add(damageRect);
-            }
-        } else if ((m_contentsSolidColor.isValid() && m_contentsSolidColor.isVisible()) || m_contentsBuffer || m_imageBackingStore)
-            context.frameDamage->add(frameDamage);
-    }
+    m_previousLayerRectInFrameCoordinates.unite(frameDamage);
+
+    if (m_layerDamage) {
+        for (const auto& rect : *m_layerDamage) {
+            auto damageRect = transform.mapRect(FloatRect(rect));
+            if (!clipBounds.isEmpty())
+                damageRect.intersect(clipBounds);
+            context.frameDamage->add(damageRect);
+        }
+    } else if ((m_contentsSolidColor.isValid() && m_contentsSolidColor.isVisible()) || m_contentsBuffer || m_imageBackingStore)
+        context.frameDamage->add(frameDamage);
+}
 #endif
 
+void SkiaCompositingLayer::paintDebugIndicators(SkCanvas& canvas, PaintContext&)
+{
     if (m_debugBorder) {
         SkPaint borderPaint;
         borderPaint.setStyle(SkPaint::kStroke_Style);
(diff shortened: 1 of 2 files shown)


Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -401,48 +401,81 @@ bool SkiaCompositingLayer::computeTransformsAndAnimations(const TransformationMa
             m_animatedBackingStoreClient->requestBackingStoreUpdateIfNeeded(m_transforms.futureCombined);
     }
 
+    m_subtreeHasDebugIndicators = m_debugBorder.has_value() || m_repaintCount.has_value();
```
carlosgcampos: debug indicators is a global setting, it's stored by the RenderLayerCompositor and applied to all the layers, so I don't think it's possible to enable it only for some layers. We could probably just check the root layer.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
```diff
@@ -107,7 +108,28 @@ class SkiaCompositingLayer final : public RefCountedAndCanMakeWeakPtr<SkiaCompos
...
+
+    // Aspects of the paint walk that can be enabled independently, so damage
+    // collection and content drawing share a single recursion that cannot drift.
+    enum class PaintMode : uint8_t {
+        Content,
+        Damage,
+        DebugIndicators,
+
+        HighestEnumValue = DebugIndicators
+    };
```
carlosgcampos: This should be private, from the caller point of view there shouldn't be any change.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
```diff
@@ -107,7 +108,28 @@ class SkiaCompositingLayer final : public RefCountedAndCanMakeWeakPtr<SkiaCompos
     const TransformationMatrix& toSurfaceTransform() const { return m_transforms.combined; }
     FloatRect effectiveLayerRect() const { return FloatRect({ }, m_size); }
 
-    bool paint(SkCanvas&, std::optional<Damage>&);
+    // Must be called exactly once per frame, before paint(). Advances animation
+    // state and emits transform/opacity-change damage, so repeated calls within
+    // a frame would double-advance and double-report.
+    bool prepareForPainting();
```
carlosgcampos: I removed this on purpose, it's a lot simpler for the caller to just call paint.

carlosgcampos (changes requested):
This was not exactly the idea. The caller should just call paint() and internally we iterate the tree once, twice or three times depending on what we want, but every iteration has always a single mode. The reason for this was also to ensure that debug borders and repaint counters are always painted on top, after all filters, masks, or whatever are applied. So, paint mode should be exactly the same as it is now, painting layers and collecting the damage for the ui process if needed, but without painting debug indicators, then if debug indicators are enabled, we change the paint context mode to debug indicators and iterate again only painting the debug indicators. In this iteration we don't need to create intermediate surfaces, nor apply filters, masks, etc. Then in the follow up we add a new mode for collecting feedback to be used during the painting mode that will iterate the tree before the painting mode. This was my idea, maybe it's not really possible to do it that way.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
nikolaszimmermann: ok.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -441,8 +448,10 @@ bool SkiaCompositingLayer::paint(SkCanvas& canvas, std::optional<Damage>& damage
 void SkiaCompositingLayer::paintSelf(SkCanvas& canvas, PaintContext& context)
 {
 #if ENABLE(DAMAGE_TRACKING)
+    const bool collectsDamage = context.mode == PaintMode::Paint;
```
carlosgcampos, line 451: Why not just using the mode?

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -719,44 +767,13 @@ void SkiaCompositingLayer::paintWithIntermediateSurface(SkCanvas& canvas, PaintC
...
-
-void SkiaCompositingLayer::paintWithMaskAndBackdrop(SkCanvas& canvas, PaintContext& context)
+void SkiaCompositingLayer::paintSelfAndChildrenWithFilterAndMask(SkCanvas& canvas, PaintContext& context)
 {
     const bool shouldClipPath = m_mask && m_mask->m_clipPath.has_value();
     if (shouldClipPath && m_mask->m_clipPath->isEmpty())
         return;
 
-    sk_sp<SkImage> maskImage = m_mask && !shouldClipPath ? m_mask->maskImage() : nullptr;
+    sk_sp<SkImage> maskImage = (m_mask && !shouldClipPath && context.mode == PaintMode::Paint) ? m_mask->maskImage() : nullptr;
```
carlosgcampos: should we clip path in debug border mode?

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -802,14 +811,33 @@ void SkiaCompositingLayer::paintWithFilterAndMask(SkCanvas& canvas, PaintContext
...
 #if ENABLE(DAMAGE_TRACKING)
     auto clipBounds = FloatRect(this->clipBounds(canvas, context));
+    const bool collectsDamage = context.mode == PaintMode::Paint;
+    const bool needToProcessFrameDamage = collectsDamage && frameDamagePropagationEnabled() && context.frameDamage;
 #endif
 
+    if (context.mode != PaintMode::Paint) {
+#if ENABLE(DAMAGE_TRACKING)
+        if (needToProcessFrameDamage) {
+            for (const auto& rect : overlapRects) {
```
carlosgcampos, line 854: I'm confused here, this is insise mode != paint, and then we check if needs to process fram damage that is only true when mode is paint

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
nikolaszimmermann: The follow-up patch (that you don't see here) extends this:

```
 void SkiaCompositingLayer::paintSelf(SkCanvas& canvas, PaintContext& context)
 {
 #if ENABLE(DAMAGE_TRACKING)
-    const bool collectsDamage = context.mode == PaintMode::Paint;
+    const bool collectsDamage = context.mode == PaintMode::CollectDamage || (context.mode == PaintMode::Paint && !context.damageAlreadyCollected);
     auto cleanup = WTF::makeScopeExit([&] {
```

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
nikolaszimmermann: A bit unfortunate how I split this, the follow-up patch does:

```
-    const bool collectsDamage = context.mode == PaintMode::Paint;
+    const bool collectsDamage = context.mode == PaintMode::CollectDamage || (context.mode == PaintMode::Paint && !context.damageAlreadyCollected);
     const bool needToProcessFrameDamage = collectsDamage && frameDamagePropagationEnabled() && context.frameDamage;
```

So we can end up with needsToProcessFrameDamage == true, but context.mode != Paint...

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
nikolaszimmermann: Good point, likely not. I'll have to adapt this... Agreed?

carlosgcampos: I think mask should be ignored when not painting, not matter if it's a mask image or a clip path

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -455,64 +464,66 @@ void SkiaCompositingLayer::paintSelf(SkCanvas& canvas, PaintContext& context)
...
-    } else if (m_contentsBuffer || m_imageBackingStore) {
-        bool shouldClipContents = m_contentsClippingRect.isRounded() || !m_contentsClippingRect.rect().contains(m_contentsRect);
-        SkAutoCanvasRestore autoRestore(&canvas, shouldClipContents);
-        if (shouldClipContents) {
-            if (m_contentsClippingRect.isRounded())
-                canvas.clipRRect(SkRRect(m_contentsClippingRect), true);
-            else
-                canvas.clipRect(SkRect(m_contentsClippingRect.rect()));
-        }
+    if (context.mode == PaintMode::Paint) {
```
carlosgcampos, line 467: I think we could add paintContents and paintDebugBorders or something like that and avoid the big if

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -752,11 +770,18 @@ void SkiaCompositingLayer::paintBackdrop(SkCanvas& canvas, PaintContext& context
...
 void SkiaCompositingLayer::paintWithMaskAndBackdrop(SkCanvas& canvas, PaintContext& context)
 {
-    const bool shouldClipPath = m_mask && m_mask->m_clipPath.has_value();
-    if (shouldClipPath && m_mask->m_clipPath->isEmpty())
-        return;
-
-    sk_sp<SkImage> maskImage = m_mask && !shouldClipPath ? m_mask->maskImage() : nullptr;
+    // The mask only affects the painted result, so ignore it entirely outside
+    // PaintMode::Paint, be it a clip path or a mask image. Damage collection and
+    // debug indicators walk the tree unmasked.
```
carlosgcampos: But the case of clip being empty should probably be handled, no? if a layer is not painted because it's fully clipped out, we don't want to see the border, no?

carlosgcampos (approved):

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
nikolaszimmermann: Yeah much cleaner.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
nikolaszimmermann: Right, I missed that, sorry.

PR #65681, opened 2026-05-26
Text fragment can match a prefix that isn't word-bounded on left near start of document

richwaters:
#### 208745c0c095eda0e5fa99542804b712d4fd26da
<pre>
Text fragment can match a prefix that isn&apos;t word-bounded on left near start of document
<a href="https://bugs.webkit.org/show_bug.cgi?id=315536">https://bugs.webkit.org/show_bug.cgi?id=315536</a>

Reviewed by Darin Adler.

The change makes the SearchBuffer::prependContext function live up to its name and prepend context instead of appending context. Prior to the change, that appending could leave the buffer sent to isWordStartMatch in the wrong order. That resulted in the wrong character being examined for a word boundary, which caused acceptance of a match that should&apos;ve been rejected. The fix properly inserts into the front of the buffer so that the characters sent to isWordStartMatch are in the correct order, and that fixes the boundary check so that it properly rejects that non-bounded pattern.

Test: http/tests/scroll-to-text-fragment/prefix-word-bounded-at-doc-start.html

* LayoutTests/http/tests/scroll-to-text-fragment/prefix-word-bounded-at-doc-start-expected.html: Added.
* LayoutTests/http/tests/scroll-to-text-fragment/prefix-word-bounded-at-doc-start.html: Added.
(WebCore::prepend):
(WebCore::SearchBuffer::prependContext): Insert the gathered context at front of buffer instead of appending to end.

Canonical link: <a href="https://commits.webkit.org/314323@main">https://commits.webkit.org/314323@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/d7a1d006f2ffbcf2881c24807d084006454ddf54

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/164995 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/38486 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/32063 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/174037 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/122251 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/166863 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/38820 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/38487 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/128215 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/90254 "1 flakes 1 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/167954 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/30522 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/148256 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/108741 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/29573 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/28187 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/21792 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/139468 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/25958 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/176560 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/29412 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/27662 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/136536 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/38554 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/32323 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/136481 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36943 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/39244 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/147754 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/101543 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/31755 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/24619 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/37754 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/110825 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/37151 "Built successfully") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/2694 "") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/37397 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/37295 "Built successfully") | | |

### DIFF
--- a/LayoutTests/http/tests/scroll-to-text-fragment/prefix-word-bounded-at-doc-start-expected.html
+++ b/LayoutTests/http/tests/scroll-to-text-fragment/prefix-word-bounded-at-doc-start-expected.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<meta charset="utf-8" />
+<title>Scroll to text fragment - prefix near document start must be word-bounded</title>
+<style>
+  span {
+    background-color: pink;
+  }
+</style>
+
+<p>Pass condition: only the word in the final sentence below is highlighted; the heading is not.</p>
+<h2>II</h2>
+<p>The Pool of Tears</p>
+<p>Let me think: was I <span>the</span> same when I got up this morning?</p>
--- a/Source/WebCore/editing/TextIterator.cpp
+++ b/Source/WebCore/editing/TextIterator.cpp
@@ -1898,6 +1898,16 @@ inline bool NODELETE SearchBuffer::needsMoreContext() const
     return m_needsMoreContext;
 }
 
+static void prepend(Vector<char16_t>& buffer, StringView text)
+{
+    size_t length = text.length();
+    size_t oldSize = buffer.size();
+    buffer.grow(oldSize + length);
+    if (oldSize)
+        WTF::memmoveSpan(buffer.mutableSpan().subspan(length), buffer.span().first(oldSize));
+    text.getCharacters(buffer.mutableSpan().first(length));
+}
+
 inline void SearchBuffer::prependContext(StringView text)
 {
     ASSERT(m_needsMoreContext);
@@ -1915,7 +1925,8 @@ inline void SearchBuffer::prependContext(StringView text)
     }
 
     size_t usableLength = std::min(m_buffer.capacity() - m_prefixLength, text.length() - wordBoundaryContextStart);
-    WTF::append(m_buffer, text.substring(text.length() - usableLength, usableLength));
+    auto suffix = text.substring(text.length() - usableLength, usableLength);
+    prepend(m_buffer, suffix);
     m_prefixLength += usableLength;
 
     if (wordBoundaryContextStart || m_prefixLength == m_buffer.capacity())
--- a/LayoutTests/http/tests/scroll-to-text-fragment/prefix-word-bounded-at-doc-start.html
+++ b/LayoutTests/http/tests/scroll-to-text-fragment/prefix-word-bounded-at-doc-start.html
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<meta charset="utf-8" />
+<title>Scroll to text fragment - prefix near document start must be word-bounded</title>
+<link rel="help" href="https://wicg.github.io/scroll-to-text-fragment/">
+<meta name="assert" content="A text directive prefix that is not at a word boundary must not match. The target should resolve to the later occurrence whose prefix is word-bounded, not the one at the start of the document.">
+<meta name="fuzzy" content="maxDifference=0-2; totalPixels=0-1500" />
+<style>
+  ::target-text { background-color: pink; }
+</style>
+
+<p>Pass condition: only the word in the final sentence below is highlighted; the heading is not.</p>
+<h2>II</h2>
+<p>The Pool of Tears</p>
+<p>Let me think: was I the same when I got up this morning?</p>
+
+<script>
+  location.href = "#:~:text=i-,the";
+</script>
+</html>


LayoutTests/http/tests/scroll-to-text-fragment/prefix-word-bounded-at-doc-start-expected.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+<meta charset="utf-8" />
+<title>Scroll to text fragment - prefix near document start must be word-bounded</title>
+<style>
+  span {
+    background-color: rgb(255, 238, 190);
```
darinadler: I think this should be:
```suggestion
    background-color: Highlight;
```

LayoutTests/http/tests/scroll-to-text-fragment/prefix-word-bounded-at-doc-start-expected.html
```diff
@@ -4,7 +4,7 @@
 <title>Scroll to text fragment - prefix near document start must be word-bounded</title>
 <style>
   span {
-    background-color: rgb(255, 238, 190);
+    background-color: Highlight;
```
darinadler: Oops, looks like `Highlight` was wrong and it should be `Mark`.
```suggestion
    background-color: Mark;
```
But I’m also not sure if `Mark` will do the right thing on all platforms, so maybe GTK and WPE tests will still fail?

richwaters: I'll try Mark;  If it still fails, should I just emulate the whitespace-collapse test and just setup a different color for the other platforms.

darinadler: Yes, if you are stuck, then that seems like a practical choice. Even better would be if _someone_ made sure we can make these kinds of tests by exposing the color for testing. I’m surprised we’d put up with this frustrating limitation since we want to be able to make reference tests.

Source/WebCore/editing/TextIterator.cpp
```diff
@@ -1915,7 +1915,12 @@ inline void SearchBuffer::prependContext(StringView text)
     }
 
     size_t usableLength = std::min(m_buffer.capacity() - m_prefixLength, text.length() - wordBoundaryContextStart);
-    WTF::append(m_buffer, text.substring(text.length() - usableLength, usableLength));
+    auto suffix = text.substring(text.length() - usableLength, usableLength);
+    size_t oldSize = m_buffer.size();
+    m_buffer.grow(oldSize + usableLength);
+    if (oldSize)
+        WTF::memmoveSpan(m_buffer.mutableSpan().subspan(usableLength), m_buffer.span().first(oldSize));
+    suffix.getCharacters(m_buffer.mutableSpan().first(usableLength));
```
darinadler: Why not put this into a function named `prepend`? We don’t have to put it into `StringView.h`, could keep it in this file, but it seems like eventually we might move it there.

darinadler (approved):
Bummer that we have to hardcode all the colors! That isn’t great for the future

Source/WebCore/editing/TextIterator.cpp
richwaters: I did find a possible alternative to the hard-coded colors that works locally. This is to specify the target color in the test document:

```html
<style>
    ::target-text { background-color: pink; }
</style>
```

Then match that 'pink' in the expected file. I could include this with the refactor to 'prepend' and see if it gets through the tests. If not, I can go back to the hard-coded per-platform colors.

darinadler: Good idea.

richwaters: It looks like the ::target-text color stylng worked for the platform tests, but a couple of seemingly unrelated tests failed. Do I need to do anything more on this?

darinadler: I don’t think you need to do more.

Source/WebCore/editing/TextIterator.cpp
```diff
@@ -1915,7 +1925,8 @@ inline void SearchBuffer::prependContext(StringView text)
     }
 
     size_t usableLength = std::min(m_buffer.capacity() - m_prefixLength, text.length() - wordBoundaryContextStart);
-    WTF::append(m_buffer, text.substring(text.length() - usableLength, usableLength));
+    auto suffix = text.substring(text.length() - usableLength, usableLength);
+    prepend(m_buffer, suffix);
```
darinadler, line 1929: I think I will allow this to land this as is, but I would have done it as a one-liner, just like before:
```suggestion
    prepend(m_buffer, text.substring(text.length() - usableLength, usableLength));
```

darinadler (approved):

PR #65705, opened 2026-05-26
AX: Add AXRelativeIndexForTextMarker parameterized attribute.

AndresGonzalezApple:
#### 98a71680bf36199e257bd8e195b4205c3fcd5fe9
<pre>
AX: Add AXRelativeIndexForTextMarker parameterized attribute.
<a href="https://bugs.webkit.org/show_bug.cgi?id=315589">https://bugs.webkit.org/show_bug.cgi?id=315589</a>
&lt;<a href="https://rdar.apple.com/problem/177973707">rdar://problem/177973707</a>&gt;

Reviewed by Tyler Wilcock.

AXIndexForTextMarker returns the position of a text marker as an offset
from the AX tree root. That contract is well defined and intentionally
preserved (an AT navigating across element boundaries needs document
positions, e.g. <a href="https://rdar.apple.com/144892240">rdar://144892240</a> for runless objects), but it forces
clients that want a *receiver*-relative offset — the position of the
marker within a specific element&apos;s text content, in its own 0..N
character space — to reconstruct one by querying AXIndexForTextMarker
on the receiver&apos;s range bounds and on the marker, then subtracting.
Each of those queries walks from the AX tree root and is therefore
O(document length), and the subtraction has to be repeated at every
client call site that needs receiver-relative positions.

This patch adds AXRelativeIndexForTextMarker, a parameterized
attribute that takes a text marker and returns its character offset
within the receiver&apos;s textMarkerRange (0..numberOfCharacters), or
NSNotFound when the marker does not lie within the receiver&apos;s range.
It is the receiver-relative counterpart to AXIndexForTextMarker.

The implementation lives on AXCoreObject so it works in both
isolated-tree and non-isolated-tree configurations via virtual
dispatch on textMarkerRange(). It composes:

- A fast path for the common case: when the supplied marker is
  anchored to the receiver and the receiver carries text runs, the
  marker&apos;s own characterOffset is already receiver-local, so the
  attribute returns it directly without any walk. O(1).

- A fallback path that subtracts the receiver&apos;s range start offset
  from the supplied marker&apos;s offset, both computed against the AX
  tree root. This is correct on both threads, returns std::nullopt
  when the marker falls outside the receiver, and is no slower than
  the existing client-side alternative (the client would have made
  the same root-walks itself, only multiplied across IPC). A FIXME
  notes that a forward walk bounded by the receiver&apos;s range end
  would make this O(receiver length) instead of O(document length).

The attribute is registered on both the editable-text and static-text
parameterized-attribute lists alongside AXIndexForTextMarker. There
is no SDK feature flag — this is a new attribute, not a contract
change, so older WebKit builds simply do not register it and clients
querying it on those builds get the AX server&apos;s standard
&quot;unknown attribute&quot; response.

Added relativeIndexForTextMarker binding to WebKitTestRunner mirroring
indexForTextMarker. The new layout test covers in-range markers (start/end/middle),
out-of-range markers (in a preceding sibling, in a following sibling), and confirms
the result is receiver-relative by querying the same marker on its own anchoring
element and getting 0.

* LayoutTests/accessibility/mac/bounds-for-range-expected.txt:
* LayoutTests/accessibility/mac/menu-parameterized-attributes-expected.txt:
* LayoutTests/accessibility/mac/relative-index-for-text-marker-expected.txt: Added.
* LayoutTests/accessibility/mac/relative-index-for-text-marker.html: Added.
* Source/WebCore/accessibility/cocoa/AXCoreObjectCocoa.mm:
(WebCore::offsetFromRootForMarker):
(WebCore::AXCoreObject::relativeIndexForTextMarker):
* Source/WebCore/accessibility/cocoa/CocoaAccessibilityConstants.h:
(-[WebAccessibilityObjectWrapper accessibilityParameterizedAttributeNames]):
(handleRelativeIndexForTextMarkerAttribute):
(createParameterizedAttributeHandlerMap):
* Tools/WebKitTestRunner/InjectedBundle/AccessibilityUIElement.cpp:
(WTR::AccessibilityUIElement::relativeIndexForTextMarker):
* Tools/WebKitTestRunner/InjectedBundle/AccessibilityUIElement.h:
* Tools/WebKitTestRunner/InjectedBundle/Bindings/AccessibilityUIElement.idl:
* Tools/WebKitTestRunner/InjectedBundle/mac/AccessibilityUIElementMac.h:
* Tools/WebKitTestRunner/InjectedBundle/mac/AccessibilityUIElementMac.mm:
(WTR::AccessibilityUIElementMac::relativeIndexForTextMarker):

Canonical link: <a href="https://commits.webkit.org/313956@main">https://commits.webkit.org/313956@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/6e28a74f11b940d3e389c4256c85d9c032020c31

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/164631 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/38115 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/31686 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/173666 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/121883 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/166500 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/38449 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/38117 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/127926 "Passed tests") | [✅ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/90045 "Passed tests") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/167590 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/30229 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/147967 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/108470 "Passed tests") | | [loading 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b0631ab3-e568-4196-af58-b6b3560686cf) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/29276 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/27898 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/21424 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/139180 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/25683 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/176189 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/29012 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/27352 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/136242 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/38184 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/32022 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/136206 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/36693 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/38874 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/147478 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/101036 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/31480 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/24334 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/37382 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/110426 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/36780 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/2397 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/37028 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/36928 "Built successfully") | | | |


LayoutTests/accessibility/mac/relative-index-for-text-marker.html
```diff
@@ -0,0 +1,54 @@
+<!DOCTYPE HTML>
+<html><!-- webkit-test-runner [ IsAccessibilityIsolatedTreeEnabled=true ] -->
```
twilco: The commit message mentions this new attribute works with both isolated tree mode off and on, but because we've specified `IsAccessibilityIsolatedTreeEnabled=true` here, we won't ever test the non-ITM configuration, right?

Source/WebCore/accessibility/cocoa/AXCoreObjectCocoa.mm
```diff
@@ -501,6 +501,47 @@ static void attributedStringSetStyle(NSMutableAttributedString *attributedString
...
+    // the receiver-local offset.
+    // FIXME: replace with a forward walk bounded by range.end() for
+    // O(receiver length) instead of O(document length). For now the fast
+    // path above covers the common case.
+    unsigned rangeStartFromRoot = offsetFromRootForMarker(range.start());
+    unsigned rangeEndFromRoot = offsetFromRootForMarker(range.end());
+    unsigned targetFromRoot = offsetFromRootForMarker(target);
+
+    if (targetFromRoot < rangeStartFromRoot || targetFromRoot > rangeEndFromRoot)
+        return std::nullopt;
```
twilco: Few optimization ideas:

1. We don't use `rangeEndFromRoot` if `targetFromRoot < rangeStartFromRoot`. Can we avoid computing it in this case, since `offsetFromRootForMarker` is O(document length)?
2. When computing `rangeEndFromRoot`, could we start the walk from `range.start()`? Since we're the ones producing the range, we should be able to safely assume start comes before end (unlike when ATs give us markers).

Source/WebCore/accessibility/cocoa/AXCoreObjectCocoa.mm
```diff
@@ -501,6 +501,47 @@ static void attributedStringSetStyle(NSMutableAttributedString *attributedString
...
+    return targetFromRoot - rangeStartFromRoot;
```
twilco: Feels like we should guard this subtraction in case we have a bug:

```cpp
if (rangeStartFromRoot > targetFromRoot) {
    AX_ASSERT_NOT_REACHED();
    return std::nullopt;
}
return targetFromRoot - rangeStartFromRoot;
```

LayoutTests/accessibility/mac/relative-index-for-text-marker.html
AndresGonzalezApple: Fixed, inadvertently left over.

Source/WebCore/accessibility/cocoa/AXCoreObjectCocoa.mm
AndresGonzalezApple: Fixed.
Added a FIXME for that since there is no existing function to do that on/off the main thread, and think adding that deserves its own patch and tests.

Source/WebCore/accessibility/cocoa/AXCoreObjectCocoa.mm
AndresGonzalezApple: Not possible since we are returning nullopt above if targetFromRoot < startFromRoot.

twilco (approved):

PR #65774, opened 2026-05-27
Web Inspector: CSS nesting + @media display incorrect within Styles inspector.

rcaliman-apple:
#### 9c43b4a10979deb1385bbf167da5f75113e1ce48
<pre>
Web Inspector: CSS nesting + @media display incorrect within Styles inspector.
<a href="https://bugs.webkit.org/show_bug.cgi?id=304635">https://bugs.webkit.org/show_bug.cgi?id=304635</a>
<a href="https://rdar.apple.com/167326791">rdar://167326791</a>

Reviewed by Qianlang Chen.

<a href="https://commits.webkit.org/283188@main">https://commits.webkit.org/283188@main</a> implemented the CSS Nesting spec&apos;s
interleaved declarations model via `StyleRuleNestedDeclarations`/`CSSNestedDeclarations`.
This changed the previous behavior where interleaved CSS properties would get moved
to the beginning of the parent CSS style rule, ahead of any nested CSS rules,
which caused confusing cascade issues.

Web Inspector did support that previous behavior with the changes in
<a href="https://commits.webkit.org/261329@main">https://commits.webkit.org/261329@main</a>, but it was not updated to support `CSSNestedDeclarations`.

CSS properties nested directly under an at-rule now go into a `CSSNestedDeclarations` which,
until this patch, were explicitly skipped when collecting style information to show in Web Inspector.

This patch also replaces the changes for the now obsolete beahvior of moving CSS properties
to the beginning of the parent style rule. Instead, it supports showing and editing properties
interleaved with CSS rules at their original position within the parent rule.

Tests: inspector/css/getMatchedStylesForNodeNestedAtRuleDeclarations.html
       inspector/css/setStyleText-interleaved.html
* LayoutTests/inspector/css/getMatchedStylesForNodeNestedAtRuleDeclarations-expected.txt: Added.
* LayoutTests/inspector/css/getMatchedStylesForNodeNestedAtRuleDeclarations.html: Added.

New test to verify the visibility of nested declarations across at-rules with deep nesting
and mixed explicit/implicit rules.

* LayoutTests/inspector/css/modify-css-property-expected.txt:
Update test expectation to change the now obsolete behavior of reordering CSS properties.

* LayoutTests/inspector/css/modify-css-property.html:
* LayoutTests/inspector/css/resources/modify-css-property.css:
(.rule-deep-nested-interleaved):
(@supports (display: grid)):
(background-color: lime;):
(background-color: green;):

Add a test case to check editing interleaved CSS properties within deeply nested CSS at-rules.

* LayoutTests/inspector/css/setStyleText-expected.txt:
* LayoutTests/inspector/css/setStyleText-interleaved-expected.txt: Added.
* LayoutTests/inspector/css/setStyleText-interleaved.html: Added.
* LayoutTests/inspector/css/setStyleText-no-declarations-expected.txt: Added.
* LayoutTests/inspector/css/setStyleText-no-declarations.html: Added.

New protocol-level test to verify that changes to CSS properties within nested at-rules
do not trample css text for sibling or parent rules.

* LayoutTests/inspector/css/setStyleText.html:
* Source/WebCore/css/CSSNestedDeclarations.h:
* Source/WebCore/css/parser/CSSParser.cpp:
(WebCore::CSSParser::supportsDeclaration):

Suppress CSS parser observer during `CSSParser::supportsDeclaration` to
prevent @supports conditions from leaking as observed properties into a
`CSSNestedDeclarations` when the @supports rule has CSS properties as direct children.

* Source/WebCore/inspector/InspectorStyleSheet.cpp:
(WebCore::flatteningStrategyForStyleRuleType):
Introduce a CommitSelfOnly flattening strategy for `CSSNestedDeclarations`
because they have no children to recurse into.

(WebCore::flattenSourceData):
(WebCore::StyleSheetHandler::endRuleBody):

The CSS parser observer reports all properties on the parent rule&apos;s source data, but the CSSOM
creates separate `CSSNestedDeclarations` at interleaved positions. The source data flat list must
match the CSSOM flat list entry-by-entry. To achieve this, this patch assigns properties to &quot;gaps&quot;
between child rules:
- For style rules: properties in gap 0 (before first child rule) stay with the parent (matches CSSStyleRule.style);
subsequent gaps become inferred entries interleaved after each child rule.
- For at-rules: gaps become inferred entries (at-rules have no .style property in the CSSOM).

(WebCore::InspectorStyleSheet::buildObjectForRule):
(WebCore::InspectorStyleSheet::setRuleStyleText):

The previous computeCanonicalRuleText approach replaced the entire rule body and reconstructed
it as properties first, then child rules. This reordered interleaved properties, changing
cascade semantics.

New strategy uses two paths:
1. Basic rules (no child rules, not implicitly nested): full rule body replacement.
2. Rules with interleaved content: targeted property-range replacement for the
exact character range of the properties being edited. The frontend sends formatted text with
indentation, so targeted replacements trim surrounding whitespace to avoid duplication.

(WebCore::InspectorStyleSheet::styleForId const):
(WebCore::InspectorStyleSheet::ruleIndexByStyle const):
(WebCore::InspectorStyleSheet::buildArrayForRuleList):
(WebCore::InspectorStyleSheet::collectFlatRules):
(WebCore::isNotSpaceOrTab): Deleted.
(WebCore::computeCanonicalRuleText): Deleted.
* Source/WebCore/inspector/InspectorStyleSheet.h:
* Source/WebCore/inspector/agents/InspectorCSSAgent.cpp:
(WebCore::InspectorCSSAgent::setRuleSelector):
(WebCore::InspectorCSSAgent::addRule):
(WebCore::InspectorCSSAgent::buildObjectForRule):
* Source/WebCore/inspector/agents/InspectorCSSAgent.h:
* Source/WebCore/inspector/agents/frame/FrameCSSAgent.cpp:
(WebCore::FrameCSSAgent::setRuleSelector):
(WebCore::FrameCSSAgent::addRule):
(WebCore::FrameCSSAgent::buildObjectForRule):
* Source/WebCore/inspector/agents/frame/FrameCSSAgent.h:
* Source/WebCore/style/InspectorCSSOMWrappers.cpp:
* Source/WebCore/style/InspectorCSSOMWrappers.cpp:
(WebCore::Style::InspectorCSSOMWrappers::collect):
(WebCore::Style::InspectorCSSOMWrappers::getWrapperForNestedDeclarationsRuleInSheets):
* Source/WebCore/style/InspectorCSSOMWrappers.h:

* Source/WebCore/style/StyleResolver.cpp:

(WI.CSSStyleDeclaration.prototype.generateFormattedText):
* Source/WebInspectorUI/UserInterface/Views/SpreadsheetCSSStyleDeclarationSection.css:
(.spreadsheet-css-declaration):
(.spreadsheet-css-declaration .header-groupings &gt; .grouping,):
(.spreadsheet-css-declaration &gt; .grouping-close-braces &gt; .grouping-close-brace):
(.spreadsheet-css-declaration .header-groupings &gt; .grouping &gt; .open-brace,):
(.spreadsheet-css-declaration :is(.selector, .grouping) &gt; .icon):
(.spreadsheet-css-declaration .header-groupings &gt; .grouping &gt; .icon):
(.spreadsheet-css-declaration .selector &gt; .icon): Deleted.
* Source/WebInspectorUI/UserInterface/Views/SpreadsheetCSSStyleDeclarationSection.js:
(WI.SpreadsheetCSSStyleDeclarationSection.prototype._renderGroupings):
(WI.SpreadsheetCSSStyleDeclarationSection.prototype.get _renderedGroupings):
(WI.SpreadsheetCSSStyleDeclarationSection.prototype._ensureIconElement):
(WI.SpreadsheetCSSStyleDeclarationSection.prototype._renderSelector):
(WI.SpreadsheetCSSStyleDeclarationSection.prototype._populateIconElementContextMenu):
(WI.SpreadsheetCSSStyleDeclarationSection.prototype._highlightNodesWithSelector):

Render nested declarations rules and groupings with surrounding brackets
and indented, mirroring how they&apos;d be authored in a stylesheet text.

Canonical link: <a href="https://commits.webkit.org/321329@main">https://commits.webkit.org/321329@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/ef9c355753bbf741f812dd7ce34aefbf624ab589

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/187637 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/61457 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/55182 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/197826 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/152198 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/atp/aa19f960-b29b-11f1-8cc6-22b673bbed6c) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/189499 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/62869 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/61186 "Built successfully") | [❌ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/146506 "Found 1 new test failure: fast/dom/lazy-image-loading-document-leak.html (failure)") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/101254 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/atp/aa19f960-b29b-11f1-2231-6fefa8fea95d) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/190592 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/50237 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/166979 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/127027 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/atp/aa19f960-b29b-11f1-e327-faf5e552aea7) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/48962 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/46800 "Passed tests") | [✅ 🧪 jsc-wpe](https://ews-build.webkit.org/#/builders/251/builds/8668 "Passed tests") | | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/15529 "Passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/159243 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/46599 "Passed tests") | [✅ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/8939 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/48768 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/53856 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/51375 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/199878 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/61063 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/166531 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/156195 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/41846 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/61783 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/43100 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/155847 "Passed tests") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/50159 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/133893 "Built successfully") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/131518 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/60159 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/21591 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/59573 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/59755 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/59719 "Built successfully") | | | |

### DIFF
--- a/LayoutTests/inspector/css/setStyleText-interleaved.html
+++ b/LayoutTests/inspector/css/setStyleText-interleaved.html
@@ -1,0 +1,129 @@
+<html>
+<head>
+<script src="../../http/tests/inspector/resources/protocol-test.js"></script>
+<style>
+    .case1 {
+        color: red;
+        @media (min-width: 0px) {
+            color: yellow;
+        }
+        color: green;
+    }
+
+    .case2 {
+        color: red;
+        @media (min-width: 0px) {
+            color: yellow;
+        }
+        &:not(.x) {
+            color: orange;
+        }
+        color: green;
+    }
+
+    .case3 {
+        color: red;
+        @media (min-width: 0px) {
+            color: yellow;
+            @supports (display: grid) {
+                color: orange;
+            }
+            color: violet;
+        }
+        color: green;
+    }
+</style>
+<script>
+function test() {
+    let suite = ProtocolTest.createAsyncSuite("CSS.setStyleText.Interleaved");
+
+    const colorReplacements = {
+        "red": "magenta",
+        "yellow": "cyan",
+        "orange": "tomato",
+        "green": "pink",
+        "violet": "navy",
+    };
+
+    async function getAuthorRulesForSelector(selector) {
+        let { root } = await InspectorProtocol.awaitCommand({ method: "DOM.getDocument" });
+        let { nodeId } = await InspectorProtocol.awaitCommand({
+            method: "DOM.querySelector",
+            params: { nodeId: root.nodeId, selector },
+        });
+        let { matchedCSSRules } = await InspectorProtocol.awaitCommand({
+            method: "CSS.getMatchedStylesForNode",
+            params: { nodeId },
+        });
+        return matchedCSSRules
+            .map(({ rule }) => rule)
+            .filter((rule) => rule.origin === "author");
+    }
+
+    function collectColorProperties(rules) {
+        let results = [];
+        for (let rule of rules) {
+            let style = rule.style;
+            for (let prop of style.cssProperties) {
+                if (prop.name === "color" && prop.status !== "disabled")
+                    results.push({ value: prop.value, styleId: style.styleId });
+            }
+        }
+        return results;
+    }
+
+
+    async function testCase(caseName, selector, expectedColors) {
+        suite.addTestCase({
+            name: `CSS.setStyleText.Interleaved.${caseName}`,
+            description: `Test editing each color property in ${selector} preserves other properties.`,
+            test: async () => {
+                for (let targetColor of expectedColors) {
+                    let rules = await getAuthorRulesForSelector(selector);
+                    let colorProps = collectColorProperties(rules);
+
+                    let target = colorProps.find((p) => p.value === targetColor);
+                    if (!target) {
+                        ProtocolTest.fail(`Could not find color property with value "${targetColor}".`);
+                        continue;
+                    }
+
+                    let newValue = colorReplacements[targetColor];
+                    ProtocolTest.log(`Editing "${targetColor}" to "${newValue}"...`);
+                    await InspectorProtocol.awaitCommand({
+                        method: "CSS.setStyleText",
+                        params: {
+                            styleId: target.styleId,
+                            text: `color: ${newValue};`,
+                        },
+                    });
+                }
+
+                const filteredValueReplacer = (key, value) => {
+                    if (["ruleId", "styleId", "range", "sourceLine", "sourceURL"].includes(key))
+                        return "<filtered>";
+                    return value;
+                };
+
+                ProtocolTest.log("");
+                ProtocolTest.log("Author CSS rules after all edits:");
+                ProtocolTest.json(await getAuthorRulesForSelector(selector), filteredValueReplacer);
+            },
+        });
+    }
+
+    testCase("BasicInteleaving", ".case1", ["red", "yellow", "green"]);
+    testCase("ImplicitAndExplicitInterleaving", ".case2", ["red", "yellow", "orange", "green"]);
+    testCase("NestedInteleaving", ".case3", ["red", "yellow", "orange", "violet", "green"]);
+
+    suite.runTestCasesAndFinish();
+}
+</script>
+</head>
+<body onload="runTest()">
+<p>Tests that CSS.setStyleText correctly edits individual properties in interleaved nested declarations without affecting other properties.</p>
+<div class="case1"></div>
+<div class="case2"></div>
+<div class="case3"></div>
+</body>
+</html>
--- a/Source/WebCore/inspector/InspectorStyleSheet.cpp
+++ b/Source/WebCore/inspector/InspectorStyleSheet.cpp
@@ -33,6 +33,7 @@
 #include "CSSLayerBlockRule.h"
 #include "CSSLayerStatementRule.h"
 #include "CSSMediaRule.h"
+#include "CSSNestedDeclarations.h"
 #include "CSSParser.h"
 #include "CSSParserEnum.h"
 #include "CSSParserObserver.h"
@@ -86,6 +87,7 @@ using namespace Inspector;
 
 enum class RuleFlatteningStrategy {
     Ignore,
+    CommitSelfOnly,
     CommitSelfThenChildren,
 };
 
@@ -108,6 +110,9 @@ static RuleFlatteningStrategy NODELETE flatteningStrategyForStyleRuleType(StyleR
         // - `InspectorCSSOMWrappers::collect` .
         return RuleFlatteningStrategy::CommitSelfThenChildren;
 
+    case StyleRuleType::NestedDeclarations:
+        return RuleFlatteningStrategy::CommitSelfOnly;
+
     // FIXME (webkit.org/b/284176): support @position-try in Web Inspector.
     case StyleRuleType::PositionTry:
 
@@ -126,7 +131,6 @@ static RuleFlatteningStrategy NODELETE flatteningStrategyForStyleRuleType(StyleR
     case StyleRuleType::FontPaletteValues:
     case StyleRuleType::Property:
     case StyleRuleType::ViewTransition:
-    case StyleRuleType::NestedDeclarations:
     case StyleRuleType::Function:
     case StyleRuleType::FunctionDeclarations:
         // These rule types do not contain rules that apply directly to an element (i.e. these rules should not appear
@@ -275,6 +279,10 @@ static void flattenSourceData(RuleSourceDataList& dataList, RuleSourceDataList& 
             flattenSourceData(data->childRules, target);
             break;
 
+        case RuleFlatteningStrategy::CommitSelfOnly:
+            target.append(data.copyRef());
+            break;
+
         case RuleFlatteningStrategy::Ignore:
             break;
         }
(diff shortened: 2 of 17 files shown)


rcaliman-apple:
Screenshots 

|before|after|
|---|---|
|01-before-basic|02-after-basic|
|03-before-nested|04-after-nested|

Source/WebCore/inspector/InspectorStyleSheet.cpp
the-chenergy, line 1541: I think the scope of this variable is large enough that we shouldn't abbreviate its name, plus abbreviating "declaration" into "decl" doesn't seem to be a widely-used tradition in WebKit:
```suggestion
    RefPtr nestedDeclaration = dynamicDowncast<CSSNestedDeclarations>(rule);
```

the-chenergy, line 1554: Setting the text to `&` here shows an `&` for an implicitly nested rule in the frontend. I see that this is our design choice, but I question whether this is appropriate, given it's a misrepresentation of [the spec:](https://drafts.csswg.org/css-nesting-1/#nested-declarations) 
> Unless otherwise specified, a nested declarations rule is a nested style rule, and acts identically to any other style rule. **It matches the exact same elements and pseudo-elements as its parent style rule**, with the same specificity behavior. *(This is similar to being a style rule with an `&` selector, but slightly more powerful, as explained above.)*

* The "as explained above" walks through *why* an `& {}` rule was rejected: the nesting selector cannot represent pseudo-elements (§4). For `.foo, .foo::before { @media { color: blue; } }`, the implicit nested declarations rule applies to both `.foo` *and* `.foo::before`. A literal `& { color: blue; }` would only apply to `.foo`, which is not the same

Therefore, surfacing `&` to the inspector incorrectly tells users the rule has matching semantics it doesn't. The `&` is fabricated by the inspector for the protocol payload, not surfaced from anywhere real. Instead, I think the correct UI is to surface the rule as having no selector 🤔

the-chenergy, line 1657: This early-return will skip the `m_parsedStyleSheet->setText` and `fireStyleSheetChanged` below. Why is that OK?

Source/WebCore/style/InspectorCSSOMWrappers.h
the-chenergy, line 25: Can this be a forward declaration instead?

LayoutTests/inspector/css/setStyleText-interleaved.html
the-chenergy, line 117: "Inteleaving" -> "Interleaving"
```suggestion
    testCase("BasicInterleaving", ".case1", ["red", "yellow", "green"]);
    testCase("ImplicitAndExplicitInterleaving", ".case2", ["red", "yellow", "orange", "green"]);
    testCase("NestedInterleaving", ".case3", ["red", "yellow", "orange", "violet", "green"]);
```

the-chenergy (commented):
Awesome fix to such an annoying, long-standing, happened-before-our-eyes bug! 🙏

Source/WebCore/inspector/InspectorStyleSheet.cpp
rcaliman-apple: This is a very good point. The "&" is artificial here, just used to represent _some selector_ for the implied rule in the frontend, but it's misleading in some cases, like nested under a pseudo-element. 

I'll change this "&" for implied rules to an empty string. This follows prior art in other developer tools to just show a nested rule without a selector. After I do that, I'll need to rebaseline a bunch of tests.

Source/WebCore/style/InspectorCSSOMWrappers.h
rcaliman-apple: Done. The header only uses it as a pointer and inside `HashMap<..., Ref<CSSNestedDeclarations>>`, both fine with an incomplete type.
 
Also added `CSSNestedDeclarations.h` to `StyleResolver.cpp`. `Ref<T>`'s destructor calls `T::deref()`, and `~Resolver()` there is where the map gets destroyed. That file already includes `CSSStyleRule.h` for the same reason.

Source/WebCore/inspector/InspectorStyleSheet.cpp
rcaliman-apple: Good catch! This would've been a regression.

`properties.isEmpty()` is the normal state of a rule with an empty body. `addRule` creates exactly that (`selector {}`), so the common "add a rule, type the first declaration" flow hit this. The CSSOM got the declaration but the cached sheet text kept `{}` forever and no `styleSheetChanged` fired, leaving every source range stale.

Fixed! Removed the early return. Only the new targeted-replacement branch needed it. Now, empty rules fall back to full-body replacement as before, and the targeted branch uses a zero-width insertion at the body start.

Added a new test `setStyleText-no-declarations.html` to cover this.

rcaliman-apple:
@the-chenergy I addressed the code review feedback. The significant changes since last time are:
- dropping `"&"` from implicit nested rules; re-baselined tests.
- introducing surrounding brackets and indentation for nested groupings and nested declaration rules (matching other browser developer tools behavior, like Firefox)
- addressing a regression mentioned in https://github.com/WebKit/WebKit/pull/65774/changes#r3376200206

Screenshots 

|Safari Before|Safari After|Firefox|
|---|---|---|
|safari-before|safari-after|firefox|

the-chenergy (approved):
Awesome! Let's address the remaining EWS failures before landing (new test's results needing rebaselining, compiler errors in Linux ports, and Safer-C++ failures)

PR #65882, opened 2026-05-28
Caches grow independently

Scony:
#### b45b044702d913334821112f1b218f9ef256be48
<pre>
Caches grow independently
<a href="https://bugs.webkit.org/show_bug.cgi?id=315752">https://bugs.webkit.org/show_bug.cgi?id=315752</a>

Reviewed by NOBODY (OOPS!).

This change introduces a simple cross-cache-instance coordination
mechanim so that caches of the same cache class are growing only
up to the shared limit.
</pre>
https://github.com/WebKit/WebKit/commit/b45b044702d913334821112f1b218f9ef256be48

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/184442 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/58363 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/52182 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/194767 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/148725 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/186305 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/59712 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/58096 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/143992 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/100061 "Exiting early after 60 failures. 60 tests run. 61 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/187397 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/47783 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/164757 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/124410 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/46493 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/44681 "Passed tests") | [✅ 🧪 jsc-wpe](https://ews-build.webkit.org/#/builders/251/builds/6373 "Passed tests") | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/13216 "Passed tests") | [loading 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/156877 "Build is in progress. Recent messages:OS: Tahoe (26.5), Xcode: 26.5; Checked out pull request; Running run-api-tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/44280 "Passed tests") | [✅ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/5840 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/45718 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/51026 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/48973 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/196710 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/58033 "Built successfully") | | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/153574 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/58737 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/40900 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/153238 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/47764 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/131028 "Built successfully") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/127444 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/57242 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/19295 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/56607 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/56851 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/56676 "Built successfully") | | |

### DIFF
--- a/Source/WTF/wtf/SharedCacheBudget.cpp
+++ b/Source/WTF/wtf/SharedCacheBudget.cpp
@@ -1,0 +1,49 @@
+/*
+ * Copyright (C) 2026 Igalia S.L.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL APPLE INC. OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ * OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include <wtf/SharedCacheBudget.h>
+
+#include <wtf/ThreadAssertions.h>
+
+namespace WTF {
+
+bool SharedCacheBudget::acquireCacheEntry()
+{
+    ASSERT(isMainThread());
+    if (m_entriesInUse < m_entriesLimit) {
+        m_entriesInUse++;
+        return true;
+    }
+    return false;
+}
+
+void SharedCacheBudget::releaseCacheEntries(size_t entries)
+{
+    ASSERT(isMainThread());
+    m_entriesInUse -= entries;
+}
+
+} // namespace WTF
--- a/Source/WebCore/platform/graphics/FontCascadeFonts.h
+++ b/Source/WebCore/platform/graphics/FontCascadeFonts.h
@@ -82,6 +82,11 @@ static constexpr int minInterval = -3; // After a hit, cache the next 3 attempts
 static constexpr int maxInterval = -3; // Never ramp up sampling, stay aggressive
 static constexpr unsigned maxSize = 3000; // Shaped text entries are large due to GlyphBuffer
 static constexpr unsigned maxTextLength = 128; // Larger than default to cache longer canvas text
+#if PLATFORM(WPE)
+static constexpr unsigned maxSharedSize = maxSize * 2; // All instances are limited to 2 * individual cache size.
+#else
+static constexpr unsigned maxSharedSize = maxSize * 1024; // Virtually infinity.
+#endif
 }
 
 } // namespace WebCore
@@ -213,6 +218,11 @@ private:
     std::optional<uint32_t> m_creationThreadID;
 #endif
 };
+
+template<>
+WEBCORE_EXPORT WTF::SharedCacheBudget FontCascadeFonts::GlyphGeometryCache::s_budget;
+template<>
+WEBCORE_EXPORT WTF::SharedCacheBudget FontCascadeFonts::ShapedTextCache::s_budget;
 
 inline bool FontCascadeFonts::isFixedPitch(const FontCascadeDescription& description, FontSelector* fontSelector)
 {
--- a/Source/WebCore/platform/graphics/TextMeasurementCache.h
+++ b/Source/WebCore/platform/graphics/TextMeasurementCache.h
@@ -27,11 +27,13 @@
 
 #include <WebCore/TextRun.h>
 #include <WebCore/TextSpacing.h>
+#include <wtf/CryptographicallyRandomNumber.h>
 #include <wtf/Forward.h>
 #include <wtf/HashFunctions.h>
 #include <wtf/HashSet.h>
 #include <wtf/Hasher.h>
 #include <wtf/MemoryPressureHandler.h>
+#include <wtf/SharedCacheBudget.h>
 #include <wtf/ZippedRange.h>
 #include <wtf/text/RapidHash.h>
 #include <wtf/text/StringCommon.h>
@@ -55,6 +57,11 @@ static constexpr int minInterval = -3; // A cache hit pays for about 3 cache mis
 static constexpr int maxInterval = 20; // Sampling at this interval has almost no overhead.
 static constexpr unsigned maxSize = 500000; // Just enough to guard against pathological growth.
 static constexpr unsigned maxTextLength = 64; // Maximum text length for SmallStringKey.
+#if PLATFORM(WPE)
+static constexpr unsigned maxSharedSize = maxSize * 2; // All instances are limited to 2 * individual cache size.
+#else
+static constexpr unsigned maxSharedSize = maxSize * 1024; // Virtually infinity.
+#endif
 }
 
 template <typename CachedType,
@@ -178,6 +185,7 @@ public:
     void clear()
     {
         ASSERT(isMainThread());
+        s_budget.releaseCacheEntries(m_singleCharMap.size() + m_map.size());
         m_singleCharMap.clear();
         m_map.clear();
     }
@@ -190,6 +198,20 @@ private:
     {
         if (MemoryPressureHandler::singleton().isUnderMemoryPressure())
             return nullptr;
+
+        if (!s_budget.acquireCacheEntry()) {
+            if (!m_singleCharMap.isEmpty() && (m_map.isEmpty() || WTF::cryptographicallyRandomNumber<unsigned>() % 2)) {
+                m_singleCharMap.remove(m_singleCharMap.random());
+                s_budget.releaseCacheEntries(1);
+            } else if (!m_map.isEmpty()) {
+                m_map.remove(m_map.random());
+                s_budget.releaseCacheEntries(1);
+            } else
+                return nullptr;
+
+            if (!s_budget.acquireCacheEntry())
+                return nullptr;
+        }
 
         unsigned length = text.length();
         bool isNewEntry;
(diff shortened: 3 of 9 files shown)


Source/WebCore/platform/graphics/TextMeasurementCache.h
```diff
@@ -189,6 +194,20 @@ class TextMeasurementCache {
         if (MemoryPressureHandler::singleton().isUnderMemoryPressure())
             return nullptr;
 
+        if (!WTF::SharedCacheBudget::singleton().acquireCacheEntry(CacheBudgetCategory)) {
+            if (m_singleCharMap.size() > 0 && WTF::cryptographicallyRandomNumber<unsigned>() % 2) {
```
nikolaszimmermann: Why is this gated with cryptoRandomNumber % 2, but the other if not?

Source/WebCore/platform/graphics/TextMeasurementCache.h
```diff
@@ -205,6 +224,9 @@ class TextMeasurementCache {
             value = &addResult.iterator->value;
         }
 
+        if (!isNewEntry)
+            WTF::SharedCacheBudget::singleton().releaseCacheEntries(1, CacheBudgetCategory);
```
nikolaszimmermann: Isn't this unnecessary churn? Couldn't you avoid the prior acquireCacheEntry, and only execute that if you are sure you added a new entry?

Source/WebCore/style/MatchedDeclarationsCache.cpp
```diff
@@ -182,33 +182,42 @@ void MatchedDeclarationsCache::add(const RenderStyle& style, const RenderStyle&
...
 void MatchedDeclarationsCache::remove(unsigned hash)
 {
-    m_entries.remove(hash);
+    if (m_entries.remove(hash))
+        WTF::SharedCacheBudget::singleton().releaseCacheEntries(1, WTF::SharedCacheBudget::Category::MatchedDeclarationsCache);
 }
 
 void MatchedDeclarationsCache::invalidate()
 {
+    WTF::SharedCacheBudget::singleton().releaseCacheEntries(m_entries.size(), WTF::SharedCacheBudget::Category::MatchedDeclarationsCache);
```
nikolaszimmermann: This should likely be swapped to hold the invariant: remove first, then update cache.

Source/WTF/wtf/SharedCacheBudget.cpp
```diff
@@ -0,0 +1,80 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include <wtf/SharedCacheBudget.h>
+
+namespace WTF {
+
+#if PLATFORM(WPE)
+static constexpr unsigned long long sharedCacheLimitBytes = 200 * MB;
```
nikolaszimmermann: I guess this needs a runtime configuration option for upstream?

Source/WTF/wtf/SharedCacheBudget.cpp
```diff
@@ -0,0 +1,80 @@
...
+#else
+static constexpr unsigned long long sharedCacheLimitBytes = 1 * GB;
+#endif
+
+unsigned long long SharedCacheBudget::approximateBytesPerCategoryEntry(Category category)
```
nikolaszimmermann: That seems fragile... how did you estimate that?

Source/WebCore/platform/graphics/TextMeasurementCache.h
```diff
@@ -189,6 +194,20 @@ class TextMeasurementCache {
         if (MemoryPressureHandler::singleton().isUnderMemoryPressure())
             return nullptr;
 
+        if (!WTF::SharedCacheBudget::singleton().acquireCacheEntry(CacheBudgetCategory)) {
```
nikolaszimmermann: Is it worth to drop an item and then try to allocate a new one, instead of just giving up as done under memory pressure?

Source/WebCore/platform/graphics/TextMeasurementCache.h
Scony: The idea is to randomly pick one of hash maps that we want to prune. `if (rand % 2) then A else B`

Source/WebCore/platform/graphics/TextMeasurementCache.h
Scony: If I'd do it, then I'd need to handle random removal logic here as well. And the problem is, `m_singleCharMap.remove(m_singleCharMap.random());` / `m_map.remove(m_map.random());` could by chances remove the just-added entry so we'd need to handle that as well. Not to mention that removing just-added entry is not efficient as it effectively removes the most hot object in the cache.

Source/WTF/wtf/SharedCacheBudget.cpp
Scony: So far, all the cache limits were specified in the code - do you think it's worth exposing this one via preferences?

Source/WTF/wtf/SharedCacheBudget.cpp
Scony: It's true, but the alternative would be to calculate entry size upon each addition which may be very costly... I did estimations based on comparing optimistic vs pessimistic sizes of key+value in each case (excluding memory that is shared anyway).

Source/WebCore/platform/graphics/TextMeasurementCache.h
Scony: Given that we don't have any hit/miss statistics as in this case, I think it's worth it. This way we allow fresh entries in while evicting potentially useless ones.

Source/WTF/wtf/SharedCacheBudget.cpp
nikolaszimmermann: Hmm, the problem is that with the unified preferences yaml we can only set bools at runtime, not values... (limitation of --internal-features=+/-Foo). Not sure, maybe we can keep it as-is for now.

Source/WTF/wtf/SharedCacheBudget.cpp
nikolaszimmermann: Need to think about that one. My concern is that someone changes the cache entries and this logic becomes stale - and the procedure to estimate those numbers is also not well-defined or reproducible...

Source/WebCore/platform/graphics/TextMeasurementCache.h
nikolaszimmermann: That makes sense..

Source/WebCore/platform/graphics/TextMeasurementCache.h
nikolaszimmermann: m_charMap empty, but m_singleCharMap not, don't you always want to pick the latter than to evict instead of gating this with the rand %2 check? I'm confused :-)

Source/WTF/wtf/SharedCacheBudget.cpp
nikolaszimmermann: So we're talking about this struct:

```
struct GlyphOverflow {
    // FIXME: May need clearer&safer storage and names. See webkit.org/b/307002
    LayoutUnit left;
    LayoutUnit right;
    LayoutUnit top;
    LayoutUnit bottom;
    bool computeBounds { false };
};
```

```
struct GlyphGeometryCacheEntry {
    Markable<float> width;
    Markable<GlyphOverflow> glyphOverflow;
    bool usedFallbackFonts { false };
};
```

```
using GlyphGeometryCache = TextMeasurementCache<GlyphGeometryCacheEntry>;
```

From TextMeasurementCache:
```
    using Map = HashMap<SmallStringKey, CachedType, DefaultHash<SmallStringKey>, SmallStringKeyHashTraits>;
    using SingleCharMap = HashMap<uint32_t, CachedType, DefaultHash<uint32_t>, HashTraits<uint32_t>>;

    SingleCharMap m_singleCharMap;
    Map m_map;
```

CachedType = GlyphGeometryCacheEntry
The variance between 32/135/160 comes only from the text-length, no?

And we know that at insertion time?

```
    CachedType* add(StringView text, CachedType&& entry)
    {
        if (!isMainThread())
            return nullptr;

        unsigned length = text.length();
...
```

So we should be able to compute this in a cheap O(1) way upon insertion, without having to guess? Or am I overlooking something?

nikolaszimmermann: So this caches `CachedTextShapingResultAndDisplayList` entries:
```
using CachedTextShapingResultAndDisplayList = std::unique_ptr<TextShapingResultAndDisplayList>
```
which is not more than:
```
    TextShapingResult textShapingResult;
    RefPtr<const DisplayList::DisplayList> displayList;
```

seems like should also be possible to keep track of the DisplayList size w/o too much overhead?

Scony:
PR reworked to simpler solution without approximations.

Source/WebCore/platform/graphics/FontCascadeFonts.h
```diff
@@ -81,6 +81,11 @@ static constexpr int minInterval = -3; // After a hit, cache the next 3 attempts
 static constexpr int maxInterval = -3; // Never ramp up sampling, stay aggressive
 static constexpr unsigned maxSize = 3000; // Shaped text entries are large due to GlyphBuffer
 static constexpr unsigned maxTextLength = 128; // Larger than default to cache longer canvas text
+#if PLATFORM(WPE)
```
nikolaszimmermann: What is the reason for the split here? Are you just trying to keep existing behavior for the other ports?

PR #65886, opened 2026-05-28
[GStreamer] Rework suspending the player when muted and hidden

vivienne-w:
#### 7a75b47064bbcbb0b68623b4a58e0b94902ccc90
<pre>
[GStreamer] Rework suspending the player when muted and hidden
<a href="https://bugs.webkit.org/show_bug.cgi?id=315755">https://bugs.webkit.org/show_bug.cgi?id=315755</a>

When the player is both muted and invisible, we try to suspend it (by just pausing the playback pipeline),
to save on resources. However so far this is happens quite inconsistently, with multiple issues:

First, suspends are tied too much to visibility changes - (un)muting can&apos;t suspend/resume,
so this refactors them out into `managePlayerSuspend()`, and also renames the relevant variables/methods
to reflect that they&apos;re about suspends, not player visibility.

Second, we are currently not able to suspend the player if it&apos;s already muted and hidden at creation:
Either we ignore it because we don&apos;t have a pipeline yet, or because it&apos;s in NULL.
In this scenario, we would also not yet have a target state to resume to, until e.g. a play() request later on,
so we need reintroduce a separate boolean flag (m_isSuspended) to make sure we know to delay that state change until we resume.

Finally, `m_isMuted` is currently updated inconsistently, and in some scenarios only from a notify::mute callback.
This presents a race condition, if we try to suspend after muting the player, but read m_isMuted before the callback
updates it. To avoid this, `isMuted()` now directly checks with the pipeline so we have a single source of truth.
Only `notifyPlayerOfMute()` still uses the flag to avoid sending redundant notifications.

* Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp:
(WebCore::MediaPlayerPrivateGStreamer::play):
(WebCore::MediaPlayerPrivateGStreamer::changePipelineState): Clear saved state if a change &lt; PLAYING is requested; so a sequence like (suspend -&gt; play() -&gt; pause() -&gt; unsuspend) doesn&apos;t end up in the wrong state.
(WebCore::MediaPlayerPrivateGStreamer::isMuted const):
(WebCore::MediaPlayerPrivateGStreamer::setMuted): Also check here if we need to suspend.
(WebCore::MediaPlayerPrivateGStreamer::createGSTPlayBin): After creating the pipeline, also check here if we need to suspend.
(WebCore::MediaPlayerPrivateGStreamer::setViewportVisibility):
(WebCore::MediaPlayerPrivateGStreamer::managePlayerSuspend): Added. Handles suspending/resuming the pipeline when needed.
(WebCore::MediaPlayerPrivateGStreamer::paint):
* Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.h:
(WebCore::MediaPlayerPrivateGStreamer::isSuspended const):
(WebCore::MediaPlayerPrivateGStreamer::isPausedByViewport const): Deleted.
* Source/WebCore/platform/graphics/gstreamer/mse/MediaPlayerPrivateGStreamerMSE.cpp:
(WebCore::MediaPlayerPrivateGStreamerMSE::updateStates): Go forward with the state change even when suspended; changePipelineState() needs to save the new target state

Canonical link: <a href="https://commits.webkit.org/314795@main">https://commits.webkit.org/314795@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/295449467a28d0934de02e2dc932b915af515050

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/167064 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/40555 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/34135 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/176112 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/124322 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/168930 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/40988 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/40564 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/129927 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/91531 "2 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/170023 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/32331 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/150106 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/110452 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/31306 "Passed tests") | [loading 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/30008 "Build is in progress. Recent messages:OS: Tahoe (26.5), Xcode: 26.5; Running apply-patch; Checked out pull request; run-api-tests") | [❌ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/24851 "Hash 29544946 for PR 65886 does not build (failure)") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/141308 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/27803 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/178650 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/31548 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/29510 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/138295 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/40626 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/34157 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/138206 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/37250 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/41313 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/149601 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/104261 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/33481 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/26466 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/39824 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/112897 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/39219 "Built successfully") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/4568 "") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/39465 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/39363 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp
+++ b/Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp
@@ -496,11 +496,12 @@ void MediaPlayerPrivateGStreamer::play()
         return;
     }
 
-    if (changePipelineState(GST_STATE_PLAYING) == ChangePipelineStateResult::Ok) {
+    auto result = changePipelineState(GST_STATE_PLAYING);
+    if (result == ChangePipelineStateResult::Ok || result == ChangePipelineStateResult::SavedUntilResume) {
         m_isEndReached = false;
         m_isDelayingLoad = false;
         m_preload = MediaPlayer::Preload::Auto;
-        GST_INFO_OBJECT(pipeline(), "Play");
+        GST_INFO_OBJECT(pipeline(), "Play%s", (result == ChangePipelineStateResult::Ok) ? "" : " (state change saved due to suspend)");
 #if ENABLE(MEDIA_TELEMETRY)
         MediaTelemetryReport::singleton().reportPlaybackState(MediaTelemetryReport::AVPipelineState::Play);
 #endif
@@ -1135,10 +1136,16 @@ MediaPlayerPrivateGStreamer::ChangePipelineStateResult MediaPlayerPrivateGStream
 {
     ASSERT(m_pipeline);
 
-    if (isPausedByViewport() && newState > GST_STATE_PAUSED) {
-        GST_DEBUG_OBJECT(pipeline(), "Saving state for when player becomes visible: %s", gst_state_get_name(newState));
-        m_stateToRestoreWhenVisible = newState;
-        return ChangePipelineStateResult::Ok;
+    if (isSuspended()) {
+        // Save requests to play until we resume.
+        if (newState > GST_STATE_PAUSED) {
+            GST_DEBUG_OBJECT(pipeline(), "Saving state for when player is resumed: %s", gst_state_get_name(newState));
+            m_stateToResume = newState;
+            return ChangePipelineStateResult::SavedUntilResume;
+        }
+
+        // Any other state changes are ok to execute right away.
+        m_stateToResume = GST_STATE_VOID_PENDING;
     }
 
     GstState currentState, pending;
@@ -1442,8 +1449,14 @@ MediaTime MediaPlayerPrivateGStreamer::platformDuration() const
 
 bool MediaPlayerPrivateGStreamer::isMuted() const
 {
-    GST_INFO_OBJECT(pipeline(), "Player is muted: %s", boolForPrinting(m_isMuted));
-    return m_isMuted;
+    bool isMuted = false;
+
+    auto streamVolume = GST_STREAM_VOLUME(m_pipeline.get());
+    if (streamVolume)
+        isMuted = gst_stream_volume_get_mute(streamVolume);
+
+    GST_INFO_OBJECT(pipeline(), "Player is muted: %s", boolForPrinting(isMuted));
+    return isMuted;
 }
 
 void MediaPlayerPrivateGStreamer::commitLoad()
@@ -2077,6 +2090,7 @@ void MediaPlayerPrivateGStreamer::setMuted(bool shouldMute)
     GST_INFO_OBJECT(pipeline(), "Setting muted state to %s", boolForPrinting(shouldMute));
     g_object_set(streamVolume, "mute", static_cast<gboolean>(shouldMute), nullptr);
     configureMediaStreamAudioTracks();
+    managePlayerSuspend();
 }
 
 void MediaPlayerPrivateGStreamer::notifyPlayerOfMute()
@@ -3594,6 +3608,8 @@ void MediaPlayerPrivateGStreamer::createGSTPlayBin(const URL& url)
         if (auto* scale = makeGStreamerElement("scaletempo"_s))
             g_object_set(m_pipeline.get(), "audio-filter", scale, nullptr);
     }
+
+    managePlayerSuspend();
 #if ENABLE(MEDIA_TELEMETRY)
     MediaTelemetryReport::singleton().reportDrmInfo(getDrm());
     MediaTelemetryReport::singleton().reportPlaybackState(MediaTelemetryReport::AVPipelineState::Create);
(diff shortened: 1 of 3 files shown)


Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp
eocanha, line 1454: Oh! GstPlaybin3 [implements the GStStreamVolume interface](https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-base/gst/playback/gstplaybin3.c?ref_type=heads#L514). I have to recognize that this cast puzzled me a bit in the begining, but it's fine 👍

eocanha: Not avoiding this notification when the muted value hasn't changed could be problematic, but if you're sure that the player tracks the mute state on its own and doesn't trigger event notifications to the JavaScript listeners, it's fine.

eocanha (commented):
`s/relect/reflect` in the commit message.

Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp
calvaris: !allowPlaybackOfInvisibleVideos is not good because people can write "0" or "false" and you taking that as true.

Besides and more important, this is a spec breakage so if we are adding this feature we should either:
1. compute time passed since we paused and seek later when we come back
2. make this default as disabled and opt-in if desired.

calvaris: I completelly agree. We are overnotifying with this change.

Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.h
calvaris: With the new scheme m_isSuspended can be true while the pipeline is still in GST_STATE_NULL (the "muted+hidden at creation" path you describe in the commit message), so saying "the pipeline was paused" is now misleading. Please update the comment to match the new semantics.

Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp
calvaris, line 1148: I think it is ok to have this checks in the changePipelineState function but I don't like the idea of lying and saying that the change was ok. I think it might be better to use Rejected and react properly on call sites (or maybe add Suspended if you think it adds semantics that help you on call sites).

calvaris: This unveiled a dormant issue to me and it is that it seems that we we are considering the playback pipelines as source of truth but we are not considering the case of volume aggregation developed by @philn . What happens if we change the volume in the aggregator? Should we notify it is muted? What happens if the aggregator is muted and we unmute it from the web?

calvaris (changes requested):

Source/WebCore/platform/graphics/gstreamer/MediaPlayerPrivateGStreamer.cpp
philn: Not sure what you mean by volume aggregation?

If this is about the webkit audio sink, it implements the stream volume interface, just like other audio sinks.

vivienne-w: Also, volume changes generally don't trigger mute notifications, volume 0 is in itself not considered to be muted.

vivienne-w: I didn't add this envvar, it's been around since https://github.com/WebKit/WebKit/commit/0218f0f88a9ca5670d3e9eea1c95903ef8152a68
That said, I did introduce a mistake here, thanks for catching that.

vivienne-w: I see your point, and I agree that it would make more sense to indicate if we are delaying the change for a suspend (though I wouldn't use `Rejected`, that's not really accurate either). I'd add a variant named `SavedUntilResume` or something similar.

Though currently barely anything checks for this result in the first place, so there won't be any difference besides code clarity, and some logging maybe.

vivienne-w: Normally, gstreamer already takes care of avoiding redundant notifications, but I've found it can initially send a few redundant mute=false ones, so I agree to revert this.

I would only use `m_isMuted` for this purpose though, because we can only update it from the callback, so reading it to decide about suspends creates a race condition.

calvaris: res -> result

calvaris (approved):

PR #66163, opened 2026-06-01
Run clip-path animations on the compositor

rwlbuis:
#### 7ea7c5f230d635ff4084103d13cb2ae8ed61f987
<pre>
Run clip-path animations on the compositor
<a href="https://bugs.webkit.org/show_bug.cgi?id=185816">https://bugs.webkit.org/show_bug.cgi?id=185816</a>
<a href="https://rdar.apple.com/97091889">rdar://97091889</a>

Reviewed by NOBODY (OOPS!).

Implement clip-path animations on the compositor for cases where
the size of the animated element is below a certain limit.

This functionality can be tested through the
AcceleratedClipPathAnimationEnabled feature flag.

Note that we only support accelerated clip-path on iOS, on MacOS it is
not supported due to lack of support for the &quot;path&quot; key in CAPresentationModifier.

* Source/WebCore/animation/KeyframeEffect.cpp:
(WebCore::acceleratedPropertyDidChange):
(WebCore::allAcceleratedAnimationProperties):
(WebCore::KeyframeEffect::computeAnimationIsAcceleratedAndAffectsAnchorGeometry):
(WebCore::KeyframeEffect::lastStyleChangeEventStyleDidChange):
* Source/WebCore/animation/WebAnimationTypes.h:
* Source/WebCore/platform/animation/AcceleratedEffect.cpp:
(WebCore::acceleratedPropertyFromCSSProperty):
(WebCore::cssPropertyFromAcceleratedProperty):
(WebCore::blend):
* Source/WebCore/platform/animation/AcceleratedEffectValues.cpp:
(WebCore::AcceleratedEffectValues::clone const):
(WebCore::AcceleratedEffectValues::AcceleratedEffectValues):
* Source/WebCore/platform/animation/AcceleratedEffectValues.h:
(WebCore::AcceleratedEffectValues::AcceleratedEffectValues):
* Source/WebCore/rendering/RenderLayerCompositor.cpp:
(WebCore::RenderLayerCompositor::requiresCompositingForAnimation const):
* Source/WebKit/UIProcess/RemoteLayerTree/RemoteAnimationStack.h:
* Source/WebKit/UIProcess/RemoteLayerTree/RemoteAnimationStack.mm:
(WebKit::RemoteAnimationStack::RemoteAnimationStack):
(WebKit::RemoteAnimationStack::initEffectsFromMainThread):
(WebKit::RemoteAnimationStack::applyEffectsFromMainThread const):
* Source/WebCore/style/StyleInterpolation.cpp:
(WebCore::Style::Interpolation::isAccelerated):
</pre>
https://github.com/WebKit/WebKit/commit/7ea7c5f230d635ff4084103d13cb2ae8ed61f987

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/183616 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/57540 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/51357 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/193838 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/147907 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/185479 "Passed tests") | [loading 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/58885 "Build is in progress. Recent messages:OS: Tahoe (26.5), Xcode: 26.5; Checked out pull request; Compiled WebKit (warnings)") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/57275 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/143310 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/99501 "Exiting early after 60 failures. 60 tests run. 61 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/186571 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/47071 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/164071 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/123733 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/45773 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/44006 "Passed tests") | [✅ 🧪 jsc-wpe](https://ews-build.webkit.org/#/builders/251/builds/5702 "Passed tests") | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/12538 "Passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/156166 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/43595 "Passed tests") | [✅ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/5016 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/44892 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/50195 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/48273 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/195776 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/57210 "Built successfully") | | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/152842 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/57914 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/40221 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/152524 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/47068 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/130193 "Built successfully") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/126505 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/56422 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/18605 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/55793 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/56032 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/55860 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/rendering/RenderLayerCompositor.cpp
+++ b/Source/WebCore/rendering/RenderLayerCompositor.cpp
@@ -3866,8 +3866,29 @@ bool RenderLayerCompositor::requiresCompositingForAnimation(RenderLayerModelObje
     if (auto styleable = Styleable::fromRenderer(renderer)) {
         if (styleable->hasRunningAcceleratedAnimations())
             return true;
+
         if (auto* effectsStack = styleable->keyframeEffectStack()) {
+            auto requiresCompositingForAcceleratedClipPath = [] (const RenderLayerModelObject& renderer, const KeyframeEffectStack& effectsStack) {
+                // We only support accelerated clip-path on iOS, on MacOS it is not supported due to lack of support for the "path" key in CAPresentationModifier.
+#if PLATFORM(IOS_FAMILY)
+                Ref settings = renderer.settings();
+                if (!settings->acceleratedClipPathAnimationEnabled())
+                    return false;
+#else
+                return false;
+#endif
+                CheckedPtr box = dynamicDowncast<RenderBox>(renderer);
+                if (!box)
+                    return false;
+                constexpr unsigned maxSupportedAcceleratedClipPathArea = 100 * 100;
+                if (!effectsStack.isCurrentlyAffectingProperty(CSSPropertyClipPath))
+                    return false;
+                float boxArea = box->borderBoxWidth() * box->borderBoxHeight();
+                return boxArea <= maxSupportedAcceleratedClipPathArea;
+            };
+
             return (effectsStack->isCurrentlyAffectingProperty(CSSPropertyOpacity) && (usesCompositing() || (m_compositingTriggers & ChromeClient::AnimatedOpacityTrigger)))
+                || requiresCompositingForAcceleratedClipPath(renderer, *effectsStack)
                 || effectsStack->isCurrentlyAffectingProperty(CSSPropertyFilter)
                 || effectsStack->isCurrentlyAffectingProperty(CSSPropertyBackdropFilter)
                 || effectsStack->isCurrentlyAffectingProperty(CSSPropertyWebkitBackdropFilter)
--- a/Source/WebKit/UIProcess/RemoteLayerTree/RemoteAnimationStack.mm
+++ b/Source/WebKit/UIProcess/RemoteLayerTree/RemoteAnimationStack.mm
@@ -32,6 +32,7 @@
 #import "RemoteProgressBasedTimeline.h"
 #import <pal/spi/cocoa/QuartzCoreSPI.h>
 #import <wtf/TZoneMallocInlines.h>
+#import <wtf/cocoa/TypeCastsCocoa.h>
 
 namespace WebKit {
 
@@ -50,19 +51,21 @@ RemoteAnimationStack::RemoteAnimationStack(RemoteAnimations&& animations, WebCor
     bool affectsFilter = false;
     bool affectsOpacity = false;
     bool affectsTransform = false;
+    bool affectsClipPath = false;
 
     for (auto& animation : m_animations) {
         auto& properties = animation->animatedProperties();
         affectsFilter = affectsFilter || properties.containsAny({ WebCore::AcceleratedEffectProperty::Filter, WebCore::AcceleratedEffectProperty::BackdropFilter });
         affectsOpacity = affectsOpacity || properties.contains(WebCore::AcceleratedEffectProperty::Opacity);
         affectsTransform = affectsTransform || properties.containsAny(WebCore::transformRelatedAcceleratedProperties);
+        affectsClipPath = affectsClipPath || properties.contains(WebCore::AcceleratedEffectProperty::ClipPath);
         m_hasProgressBasedAnimations = m_hasProgressBasedAnimations || animation->timeline().isProgressBased();
         m_hasTimeBasedAnimations = m_hasTimeBasedAnimations || animation->timeline().isMonotonic();
-        if (affectsFilter && affectsOpacity && affectsTransform && m_hasProgressBasedAnimations && m_hasTimeBasedAnimations)
+        if (affectsFilter && affectsOpacity && affectsTransform && affectsClipPath && m_hasProgressBasedAnimations && m_hasTimeBasedAnimations)
             break;
     }
 
-    ASSERT(affectsFilter || affectsOpacity || affectsTransform);
+    ASSERT(affectsFilter || affectsOpacity || affectsTransform || affectsClipPath);
 
     if (affectsFilter)
         m_affectedLayerProperties.add(LayerProperty::Filter);
@@ -70,6 +73,8 @@ RemoteAnimationStack::RemoteAnimationStack(RemoteAnimations&& animations, WebCor
         m_affectedLayerProperties.add(LayerProperty::Opacity);
     if (affectsTransform)
         m_affectedLayerProperties.add(LayerProperty::Transform);
+    if (affectsClipPath)
+        m_affectedLayerProperties.add(LayerProperty::ClipPath);
 }
 
 #if PLATFORM(MAC)
@@ -211,6 +216,13 @@ void RemoteAnimationStack::applyEffectsFromMainThread(PlatformLayer *layer, bool
     if (m_affectedLayerProperties.contains(LayerProperty::Transform)) {
         auto computedTransform = computedValues.computedTransformationMatrix(m_bounds);
         [layer setTransform:computedTransform];
+    }
+
+    if (computedValues.transformOperationData) {
+        if (auto path = tryPath(computedValues.clipPath, *computedValues.transformOperationData)) {
+            if (auto mask = dynamic_objc_cast<CAShapeLayer>(layer.mask))
+                mask.path = protect(path->platformPath()).get();
+        }
     }
 }
 
--- a/Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
+++ b/Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
@@ -128,6 +128,20 @@ AXCustomColorModeEnabled:
     WebCore:
       default: false
 
+AcceleratedClipPathAnimationEnabled:
+  type: bool
+  status: unstable
+  humanReadableName: "Accelerated clip-path animation"
+  humanReadableDescription: "Enable accelerated clip-path animations"
+  condition: PLATFORM(IOS_FAMILY)
+  defaultValue:
+    WebKitLegacy:
+      default: false
+    WebKit:
+      default: false
+    WebCore:
+      default: false
+
 AcceleratedCompositingEnabled:
   type: bool
   status: embedder
--- a/Source/WebCore/animation/KeyframeEffect.cpp
+++ b/Source/WebCore/animation/KeyframeEffect.cpp
@@ -2989,6 +2989,16 @@ void KeyframeEffect::computeHasReferenceFilter()
     }();
 }
 
+static HashSet<CSSPropertyID> allAcceleratedAnimationProperties([[maybe_unused]] const Settings& settings)
+{
+    HashSet<CSSPropertyID> allProperties { CSSProperty::allAcceleratedAnimationProperties(settings) };
+#if PLATFORM(IOS_FAMILY)
+    if (settings.acceleratedClipPathAnimationEnabled())
+        allProperties.add(CSSPropertyClipPath);
+#endif
+    return allProperties;
+}
+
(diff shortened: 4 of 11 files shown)


Source/WebCore/rendering/RenderLayerCompositor.cpp
```diff
@@ -3816,8 +3816,24 @@ bool RenderLayerCompositor::requiresCompositingForAnimation(RenderLayerModelObje
     if (auto styleable = Styleable::fromRenderer(renderer)) {
         if (styleable->hasRunningAcceleratedAnimations())
             return true;
+
         if (auto* effectsStack = styleable->keyframeEffectStack()) {
+            auto requiresCompositingForAcceleratedClipPath = [] (const RenderLayerModelObject& renderer, const KeyframeEffectStack& effectsStack) {
+                constexpr unsigned maxSupportedAcceleratedClipPathArea = 100 * 100;
```
weinig: Can add a comment explaining this constant, why its needed, how the value was picked?

Source/WebCore/style/values/shapes/StyleCircleFunction.cpp
```diff
@@ -183,7 +183,7 @@ AcceleratedEffectCircleFunction::RadialSize Evaluation<Circle::RadialSize, Accel
 
 AcceleratedEffectCircleFunction Evaluation<CircleFunction, AcceleratedEffectCircleFunction>::operator()(const CircleFunction& value, const TransformOperationData& data, ZoomFactor zoom)
 {
-    auto containingBlockSize = data.motionPathData->offsetRect().rect().size();
+    auto containingBlockSize = data.motionPathData ? data.motionPathData->offsetRect().rect().size() : data.boundingBox.size();
```
weinig: Basing what bounding rect is used based on whether motionPathData is null or not seems a bit fragile.

Is the logic here that when motionPathData is non-null. this circle is being used for an OffsetPath, and otherwise, it is being used for a ClipPath? 

My gut say we should probably be passing in basis size value directly (I think from Evaluation<OffsetPath, ...> and Evaluation<ClipPath, ...>) instead of performing this check here.

Source/WebCore/rendering/RenderLayerCompositor.cpp
```diff
@@ -3816,8 +3816,24 @@ bool RenderLayerCompositor::requiresCompositingForAnimation(RenderLayerModelObje
...
+                if (!effectsStack.isCurrentlyAffectingProperty(CSSPropertyClipPath))
+                    return false;
+                Ref settings = renderer.settings();
+                if (!settings->acceleratedClipPathAnimationEnabled())
+                    return false;
+                CheckedPtr box = dynamicDowncast<RenderBox>(renderer);
+                if (!box)
+                    return false;
```
graouts, line 3871: I expect these checks are cheaper than the one above, so perhaps move them around?

Source/WebKit/UIProcess/RemoteLayerTree/RemoteAnimationStack.mm
```diff
@@ -50,26 +50,30 @@
...
 
     for (auto& animation : m_animations) {
         auto& properties = animation->animatedProperties();
         affectsFilter = affectsFilter || properties.containsAny({ WebCore::AcceleratedEffectProperty::Filter, WebCore::AcceleratedEffectProperty::BackdropFilter });
         affectsOpacity = affectsOpacity || properties.contains(WebCore::AcceleratedEffectProperty::Opacity);
         affectsTransform = affectsTransform || properties.containsAny(WebCore::transformRelatedAcceleratedProperties);
+        affectsClipPath = affectsClipPath || properties.contains(WebCore::AcceleratedEffectProperty::ClipPath);
         m_hasProgressBasedAnimations = m_hasProgressBasedAnimations || animation->timeline().isProgressBased();
         m_hasTimeBasedAnimations = m_hasTimeBasedAnimations || animation->timeline().isMonotonic();
         if (affectsFilter && affectsOpacity && affectsTransform && m_hasProgressBasedAnimations && m_hasTimeBasedAnimations)
```
graouts: Add a check for `affectsClipPath` to return early here.

Source/WebKit/UIProcess/RemoteLayerTree/RemoteAnimationStack.mm
```diff
@@ -191,6 +210,13 @@
         [m_transformPresentationModifier setValue:transform.get()];
     }
 
+    if (RetainPtr layer = m_clipPathLayer.get()) {
+        if (computedValues.transformOperationData) {
+            if (auto path = tryPath(computedValues.clipPath, *computedValues.transformOperationData))
+                dynamic_objc_cast<CAShapeLayer>(layer.get().mask).path = protect(path->platformPath()).get();
+        }
+    }
```
graouts: This steps away from how threaded animations work on macOS where presentation modifiers are used to set animated values on layers when called off the main thread.

I'm not super familiar with how masking works in the remote layer tree, but I expect the `mask` layer is set to a `CAShapeLayer` and we create a presentation modifier targeting that layer and use `path` as its key path.

Perhaps @smfr can chime in on how the CSS `clip-path` property ends up affecting the structure of the remote layer tree.

Source/WebCore/style/values/shapes/StyleCircleFunction.cpp
rwlbuis: This part was a bit hacky indeed. I like the passing in idea, I'll create a separate PR for that since it is a bit independent of accelerating clip-path animation.

rwlbuis: I moved this to https://github.com/WebKit/WebKit/pull/66575.

Source/WebKit/UIProcess/RemoteLayerTree/RemoteAnimationStack.mm
rwlbuis: I would like to stick to  presentation modifiers myself, however I encountered a problem, see the comment above:

// Passing the path using initialValue to CAPresentationModifier constructor raises an exception.

@smfr could you or somebody else check if it is possible to use CGPath as value for the "path" key in presentation modifiers in the first place?

graouts: > // Passing the path using initialValue to CAPresentationModifier constructor raises an exception.

What key path and layer are you targeting?

rwlbuis: I am targeting key path "path" on CAShapeLayer. However, AFAICS the CAPresentationModifier constructor does not deal yet with layers, yet the exception is thrown just on the basis of key path "path" and CGPath being passed for initialValue in the CAPresentationModifier constructor, this is what is surprising me.

Source/WebKit/UIProcess/RemoteLayerTree/RemoteAnimationStack.mm
```diff
@@ -212,6 +219,13 @@
         auto computedTransform = computedValues.computedTransformationMatrix(m_bounds);
         [layer setTransform:computedTransform];
     }
+
+    if (computedValues.transformOperationData) {
+        if (auto path = tryPath(computedValues.clipPath, *computedValues.transformOperationData)) {
+            if (auto mask = dynamic_objc_cast<CAShapeLayer>(layer.mask))
+                mask.path = protect(path->platformPath()).get();
+        }
+    }
```
graouts, line 228: This code path can only work for blending values on iOS since on macOS you'd need to go through the scrolling thread and use a `CAPresentationModifier` which can't be used to update this kind of data. We probably need a separate flag to only ever opt keyframe effects into acceleration if the remote layer tree for that platform can support it.

Source/WebCore/rendering/RenderLayerCompositor.cpp
```diff
@@ -3847,8 +3847,28 @@ bool RenderLayerCompositor::requiresCompositingForAnimation(RenderLayerModelObje
...
+
         if (auto* effectsStack = styleable->keyframeEffectStack()) {
+            auto requiresCompositingForAcceleratedClipPath = [] (const RenderLayerModelObject& renderer, const KeyframeEffectStack& effectsStack) {
+#if PLATFORM(IOS_FAMILY)
+                Ref settings = renderer.settings();
+                if (!settings->acceleratedClipPathAnimationEnabled())
+                    return false;
+#else
+                return false;
+#endif
```
graouts, line 3868: I think we need a comment explaining why we have a specific iOS code path.

rwlbuis: Makes sense, I added a comment.

PR #66014, opened 2026-05-29
[GTK][WPE] Bump wkdev-sdk to a Swift-enabled version

nikolaszimmermann:
#### ec528c8f30662c82bf4131c9b7c6747f1131b49a
<pre>
[GTK][WPE] Bump wkdev-sdk to a Swift-enabled version
<a href="https://bugs.webkit.org/show_bug.cgi?id=315451">https://bugs.webkit.org/show_bug.cgi?id=315451</a>

Reviewed by Adrian Perez de Castro.

Reland 313985@main (that got reverted in 314013@main, after
adressing the build issues in non-SDK builds.

Switch to wkdev-sdk v6 image that ships the Swift toolchain, and
default-enable ENABLE_SWIFT_DEMO_URI_SCHEME on GTK/WPE, if possible.

The default is set in WEBKIT_OPTION_BEGIN (next to the other *_DEFAULT
options) and only flips on when the toolchain can actually build the
Swift/C++ interop: Clang (GCC links but crashes at runtime) with a Swift
new enough for the -emit-clang-header-min-access frontend flag (Swift 6.3+,
probed directly rather than by version string). On any other toolchain it
stays off, so a stock distro build still works. Requesting a Swift feature
explicitly (-DENABLE_SWIFT_DEMO_URI_SCHEME=ON or ENABLE_BACK_FORWARD_LIST_SWIFT)
against an unsupported toolchain fails the configure step with FATAL_ERROR
rather than being silently dropped.

Build fixes needed to make the new SDK image reach a working binary:

* vfs-overlay.yaml.in: hardcoded /usr/include/{glib-2.0,libsoup-3.0};
  the wkdev SDK installs them under /jhbuild/install/include/..., so
  the overlay never applied. Switch to @GLib_INCLUDE_DIR@ /
  @Soup3_INCLUDE_DIR@.

* swiftc-wrapper.sh: wrap clang-only include flags
  (-isystem / -iquote / -idirafter / -isysroot and their argument) as
  &apos;-Xcc &lt;flag&gt; -Xcc &lt;path&gt;&apos; so pkg-config flags reach the Clang importer.

* WebKitMacros.cmake: skip /usr/include/c++/N(/backward) (and
  /usr/lib/gcc/*) when forwarding CMAKE_CXX_IMPLICIT_INCLUDE_DIRECTORIES
  to swiftc - pinning a specific libstdc++ on the importer clashes with
  the one Swift&apos;s embedded clang finds.

[...]

PR #66200, opened 2026-06-01
[GTK][WPE] Enable Swift toolchain support in Yocto cross-builds

nikolaszimmermann:
#### 5405edf71fd57afeb156ff79b72ad3199276a924
<pre>
[GTK][WPE] Enable Swift toolchain support in Yocto cross-builds
<a href="https://bugs.webkit.org/show_bug.cgi?id=316032">https://bugs.webkit.org/show_bug.cgi?id=316032</a>

Wire up a Swift/C++ interop toolchain for the GTK/WPE Yocto cross-builds. The
Swift compiler and runtime come from a new meta-swift layer. nativesdk-swift
wraps the swift.org prebuilt bundle to provide the SDK host clang (clang-21)
that compiles WebKit, with CLANGSDK=0 keeping meta-clang&apos;s nativesdk-clang out
of the SDK host. GCC and clang stay selectable per build (CC=clang CXX=clang++
build-webkit ...).

cross-toolchain-helper rewrites the SDK env-setup to emit both a GCC and a clang
form of the compiler exports, and splices CMAKE_Swift_FLAGS_INIT into
BUILD_WEBKIT_ARGS to point swiftc at the target sysroot. The swiftc-wrapper
drops host -I /usr/include and forwards a bare -fuse-ld= so swiftc&apos;s link step
uses the Yocto cross-linker. WebKitMacros mirrors the bundle clang&apos;s flags into
the platform-swift-args preprocess and skips host implicit include dirs when
cross-compiling. WebKit&apos;s CMakeLists adds a WebKit_DerivedHeadersForSwift
ordering target so the interop sub-target can&apos;t race *Messages.h generation.

The Swift-enabled RPi *-mesa images build wpewebkit with clang, set in each
local.conf. For downstream meta-webkit recipe builds the patch adds an opt-in
PACKAGECONFIG[swift], off by default.

* CMakeLists.txt:
* Tools/Scripts/cross-toolchain-helper:
(YoctoCrossBuilder.build_toolchain):
* Tools/Scripts/swift/swiftc-wrapper.sh:
* Tools/yocto/README.md:
* Tools/yocto/meta-openembedded_and_meta-webkit.patch:
* Tools/yocto/rpi/bblayers.conf:
* Tools/yocto/rpi/local-rpi3-32bits-mesa.conf:
* Tools/yocto/rpi/local-rpi3-32bits-userland.conf:
* Tools/yocto/rpi/local-rpi3-64bits-mesa.conf:
* Tools/yocto/rpi/local-rpi4-32bits-mesa.conf:
* Tools/yocto/rpi/local-rpi4-64bits-mesa.conf:
* Tools/yocto/rpi/local-rpi5-64bits-mesa.conf:
* Tools/yocto/rpi/manifest.xml:

Canonical link: <a href="https://commits.webkit.org/315523@main">https://commits.webkit.org/315523@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/2b374eb7b2e4029927aee2528547f61a1afffbde

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/169313 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/43235 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/36500 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/178669 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/127025 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/171179 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/43299 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/42863 "Built successfully") | [  ~~🧪 wpe-wk2~~](https://ews-build.webkit.org/#/builders/34/builds/131881 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/92818 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/172272 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/162/builds/34373 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/152170 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/112385 "Passed tests") | 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/33356 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/32022 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/27369 "Built successfully") | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/588 "Passed tests") | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/143346 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/31570 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/181269 "Built successfully") | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/30568 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/33979 "Built successfully") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/36192 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/1/builds/140337 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/42891 "Built successfully") | [loading 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/151641 "Build is in progress. Recent messages:") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/140175 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/37712 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/43580 "Built successfully") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/28545 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/107545 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/35443 "Passed tests") | [❌ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/115345 "Found 5 new failures in dfg/DFGLoopUnrollingPhase.cpp") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/201992 "Built successfully") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/42090 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/6635 "Passed tests") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/54177 "Passed tests") | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/41483 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/41738 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/41626 "Built successfully") | | |

### DIFF
--- a/Tools/Scripts/cross-toolchain-helper
+++ b/Tools/Scripts/cross-toolchain-helper
@@ -546,52 +546,91 @@ class YoctoCrossBuilder():
             original_setup_env_path = self._get_cross_toolchain_env_path(check_toolchain_built=False)
             _log.info('Patching toolchain environment-setup file to remove optimizations by default: {original_setup_env_path}'.format(original_setup_env_path=original_setup_env_path))
             # This does the following:
-            # 1. Remove any line that exports CFLAGS/CXXFLAGS/LDFLAGS/CPPFLAGS
-            # 2. On the line that exports the compiler variables (export CC, export CXX and export CPP)
-            #    remove any compiler flag that is not machine related (-m*) or is '-E' (for cpp) or is for the sysroot/target.
-            # 3. Do not override CC/CXX with the cross-compiled GCC compiler if the user wants to use Clang, instead
-            #    set CC/CXX pointing to the cross-compiled Clang (assuming the support was built in the toolchain).
+            # 1. Drop the CFLAGS/CXXFLAGS/LDFLAGS/CPPFLAGS exports.
+            # 2. On the CC/CXX/CPP exports, keep only machine flags (-m*), '-E', and
+            #    the sysroot.
+            # 3. If the SDK ships clang (swift.org bundle), also emit a Clang form
+            #    swapping `<sys>-gcc/g++/cpp` for `clang/clang++/clang-cpp --target=<sys>`
+            #    (Swift/C++ interop is Clang-only), selected by the user via CC/CXX.
             patched_setup_env_path = original_setup_env_path + "-no-opt-flags"
+            # Anchor on the trailing '=' so we don't also strip unrelated exports
+            # like CPPFLAGS_FOR_BUILD. No whitespace before '=' since shell variable
+            # assignments don't allow it.
             flags_regex = re.compile(r'export\s+(C|CXX|LD|CPP)FLAGS=')
-            gcc_compiler_regex = re.compile(r'export\s+C(C|XX|PP)=')
-            clang_compiler_regex = re.compile(r'export\s+CLANGC(C|XX|PP)=')
+            compiler_regex = re.compile(r'export\s+C(C|XX|PP)=')
+            toolchain_binary_regex = re.compile(r'([\w-]+-)(gcc|g\+\+|cpp)$')
+            clang_for = {'gcc': 'clang', 'g++': 'clang++', 'cpp': 'clang-cpp'}
+            # Detect whether any SDK host sysroot has clang / swiftc binaries.
+            sysroots_dir = os.path.join(os.path.dirname(original_setup_env_path), 'sysroots')
+            sdk_has_clang = False
+            sdk_has_swiftc = False
+            if os.path.isdir(sysroots_dir):
+                for entry in os.listdir(sysroots_dir):
+                    host_bin = os.path.join(sysroots_dir, entry, 'usr', 'bin')
+                    if os.path.isfile(os.path.join(host_bin, 'clang')):
+                        sdk_has_clang = True
+                    if os.path.isfile(os.path.join(host_bin, 'swiftc')):
+                        sdk_has_swiftc = True
             gcc_lines = []
             clang_lines = []
             with open(original_setup_env_path, 'r') as fr:
                 with open(patched_setup_env_path, 'w') as fw:
                     for line in fr.readlines():
                         if not flags_regex.match(line):
-                            if gcc_compiler_regex.match(line) or clang_compiler_regex.match(line):
-                                newline = ""
+                            if compiler_regex.match(line):
+                                # GCC form always; Clang form too when the SDK ships clang.
+                                gcc_line = ""
+                                clang_line = ""
                                 for word in line.strip().split():
                                     if word.startswith('-'):
-                                        if word == '-E' or any(word.startswith(p) for p in ['-m', '--sysroot=', '--target=']):
-                                            newline += " " + word
+                                        if word.startswith('-m') or word == '-E' or word.startswith('--sysroot'):
+                                            gcc_line += " " + word
+                                            clang_line += " " + word
                                     else:
-                                        newline += " " + word
-                                if '="' in newline and not newline.endswith('"'):
-                                    newline += '"'
-                                newline = newline.strip() + '\n'
-                                (clang_lines if clang_compiler_regex.match(line) else gcc_lines).append(newline)
+                                        gcc_line += " " + word
+                                        m = toolchain_binary_regex.search(word)
+                                        if m and sdk_has_clang:
+                                            target = m.group(1).rstrip('-')
+                                            # --gcc-install-dir/-isystem point at the
+                                            # sysroot so crt*.o/libgcc.a and the c++
+                                            # headers resolve. Globs expand at source time.
+                                            gcc_dir = ('$(ls -d "$SDKTARGETSYSROOT/usr/lib/gcc/' + target
+                                                       + '"/* 2>/dev/null | head -1)')
+                                            cxx_inc = ('$(ls -d "$SDKTARGETSYSROOT/usr/include/c++"'
+                                                       '/* 2>/dev/null | head -1)')
+                                            clang_line += " " + word.replace(m.group(0),
+                                                clang_for[m.group(2)] + " --target=" + target
+                                                + " --gcc-install-dir=" + gcc_dir
+                                                + " -isystem " + cxx_inc
+                                                + " -isystem " + cxx_inc + "/" + target)
+                                        else:
+                                            clang_line += " " + word
+                                if '="' in gcc_line and not gcc_line.endswith('"'):
+                                    gcc_line += '"'
+                                if '="' in clang_line and not clang_line.endswith('"'):
+                                    clang_line += '"'
+                                gcc_lines.append(gcc_line.strip() + '\n')
+                                clang_lines.append(clang_line.strip() + '\n')
                             else:
                                 fw.write(line)
-                    # The script should check the CC/CXX env var from the user, if is set to clang then export
-                    # the CLANGCC and CLANGCXX as CC/CXX, otherwise keep the default (GCC)
+                    # Be strict: the setup-environment script is expected to export
+                    # exactly CC/CXX/CPP. If the count ever differs from 3 the file
+                    # changed shape and this has to be inspected to see what is going on.
                     if len(gcc_lines) != 3:
                         raise RuntimeError('Unexpected CC/CXX/CPP parameters in setup-environment script')
-                    if clang_lines:
+                    # Emit the compiler exports: Clang form when the SDK ships clang
+                    # and the user asked for it via CC/CXX, otherwise GCC.
+                    if sdk_has_clang and clang_lines:
                         fw.write('\nif echo "${CC} ${CXX}" | grep -qi clang; then\n')
-                        for line in clang_lines:
-                            # Remove the CLANG prefix, so for example CLANGCXX becomes CXX
-                            line = re.sub(r'(export\s+)CLANG(CC|CXX|CPP)(=)', r'\1\2\3', line)
-                            fw.write(f'    {line}')
+                        for _cl in clang_lines:
+                            fw.write('    ' + _cl)
                         fw.write('else\n')
-                        for line in gcc_lines:
-                            fw.write(f'    {line}')
+                        for _gl in gcc_lines:
+                            fw.write('    ' + _gl)
                         fw.write('fi\n')
                     else:
-                        for line in gcc_lines:
-                            fw.write(line)
+                        for _gl in gcc_lines:
+                            fw.write(_gl)
                     # When building cog, meson tries to load the wpe headers that are on the repo from the sysroot path.
                     # That is not really a bug on meson as system libraries should be loaded from the sysroot when cross-building.
                     # Workaround the issue by creating a symlink that also resolves this WebKit repository path inside the sysroot.
@@ -599,6 +638,54 @@ class YoctoCrossBuilder():
                              '    mkdir -p "${{OECORE_TARGET_SYSROOT}}/{webkit_dir_up}"\n'
                              '    ln -s "{webkit_dir}" "${{OECORE_TARGET_SYSROOT}}/{webkit_dir_up}"\n'
                              'fi\n'.format(webkit_dir=self._webkit_dir, webkit_dir_up=os.path.dirname(self._webkit_dir)))
+                    # When the SDK ships swiftc, splice -DCMAKE_Swift_FLAGS_INIT into
+                    # BUILD_WEBKIT_ARGS to wire swiftc to the Yocto sysroot:
+                    #   -sdk / -I        target headers + Swift modules.
+                    #   -Xcc ...         swiftc's *compile* step spawns clang -cc1,
+                    #                    which -sdk doesn't reach, so the sysroot and
+                    #                    gcc-install-dir have to be handed to it here.
+                    #   -Xclang-linker   swiftc's *link* step spawns a separate clang
+                    #                    driver that does NOT inherit the -Xcc flags,
+                    #                    so --sysroot/--gcc-install-dir are repeated
+                    #                    for it. -fuse-ld=<abs <sys>-ld> additionally
+                    #                    forces the Yocto cross-linker; otherwise that
+                    #                    clang driver defaults to the host's ld and
+                    #                    links the target binary with the wrong linker.
+                    # The compile and link steps are distinct clang invocations, which
+                    # is why the sysroot/gcc-install-dir flags appear once per step.
+                    # Paths resolve at env-setup source time via shell.
+                    if sdk_has_swiftc:
+                        # Yocto splits gcc-install across usr/lib/<sys> and
+                        # usr/lib/gcc/<sys>; Swift's libstdc++ probe needs crtbegin.o
+                        # in the gcc/ tree, so copy what's missing (symlinks aren't
+                        # followed by the clang driver). Idempotent.
+                        fw.write(
+                            '\nif [ -n "${OECORE_TARGET_SYSROOT:-}" ] && [ -d "${OECORE_TARGET_SYSROOT}/usr/lib/swift" ]; then\n'
+                            '    _target_prefix="${TARGET_PREFIX%-}"\n'
+                            '    _gcc_src=$(ls -d "${OECORE_TARGET_SYSROOT}/usr/lib/${_target_prefix}"/* 2>/dev/null | head -1)\n'
+                            '    if [ -n "${_gcc_src}" ]; then\n'
+                            '        _gcc_ver=$(basename "${_gcc_src}")\n'
+                            '        _gcc_dst="${OECORE_TARGET_SYSROOT}/usr/lib/gcc/${_target_prefix}/${_gcc_ver}"\n'
+                            '        if [ -d "${_gcc_dst}" ] && [ ! -e "${_gcc_dst}/crtbegin.o" ]; then\n'
+                            '            for _f in "${_gcc_src}"/*; do\n'
+                            '                _bn=$(basename "${_f}")\n'
+                            '                [ -e "${_gcc_dst}/${_bn}" ] || cp -rp "${_f}" "${_gcc_dst}/${_bn}"\n'
+                            '            done\n'
+                            '            unset _f _bn\n'
+                            '        fi\n'
+                            '        unset _gcc_ver _gcc_dst\n'
+                            '    fi\n'
+                            '    unset _target_prefix _gcc_src\n'
+                            'fi\n'
+                            '\nif [ -n "${BUILD_WEBKIT_ARGS:-}" ] && [ -d "${OECORE_TARGET_SYSROOT}/usr/lib/swift" ]; then\n'
+                            '    _target_prefix="${TARGET_PREFIX%-}"\n'
+                            '    _gcc_install_dir=$(ls -d "${OECORE_TARGET_SYSROOT}/usr/lib/gcc/${_target_prefix}"/* 2>/dev/null | head -1)\n'
+                            '    _cross_ld=$(command -v "${_target_prefix}-ld" 2>/dev/null)\n'
+                            '    _swift_flags="-sdk ${OECORE_TARGET_SYSROOT} -I ${OECORE_TARGET_SYSROOT}/usr/lib/swift/linux -I ${OECORE_TARGET_SYSROOT}/usr/lib/swift -Xcc --sysroot=${OECORE_TARGET_SYSROOT} -Xcc --gcc-install-dir=${_gcc_install_dir} -Xclang-linker --sysroot=${OECORE_TARGET_SYSROOT} -Xclang-linker --gcc-install-dir=${_gcc_install_dir} -Xclang-linker -fuse-ld=${_cross_ld}"\n'
+                            '    BUILD_WEBKIT_ARGS="${BUILD_WEBKIT_ARGS} --cmakeargs=\\"-DCMAKE_Swift_FLAGS_INIT=\'${_swift_flags}\'\\""\n'
+                            '    export BUILD_WEBKIT_ARGS\n'
+                            '    unset _swift_flags _target_prefix _gcc_install_dir _cross_ld\n'
+                            'fi\n')
 
             # Put the patched setup-environ file in place of the other
             # and save a copy of the original just for debugging purposes
--- a/Tools/Scripts/swift/swiftc-wrapper.sh
+++ b/Tools/Scripts/swift/swiftc-wrapper.sh
@@ -51,8 +51,23 @@ for arg in "$@"; do
         pass_next_verbatim=
         continue
     fi
+    # Drop bare host `-I /usr/include` (or /usr/local/include): in a cross-build
+    # they leak host glibc into swiftc's clang importer and break the libstdc++
+    # `std` module. Safe everywhere (clang searches /usr/include by default).
+    if [[ -n "$pending_dash_I" ]]; then
+        pending_dash_I=
+        case "$arg" in
+            "/usr/include"|"/usr/local/include") ;;
+            *) args+=("-I" "$arg") ;;
+        esac
+        continue
+    fi
     case "$arg" in
-        "-Xcc"|"-Xlinker"|"-Xfrontend")
+        "-I")
+            pending_dash_I=1
+            ;;
+        "-I/usr/include"|"-I/usr/local/include") ;;
+        "-Xcc"|"-Xlinker"|"-Xfrontend"|"-Xclang-linker")
             args+=("$arg")
             pass_next_verbatim=1
             ;;
@@ -70,7 +85,11 @@ for arg in "$@"; do
             ;;
         "-include") skip_next=1 ;;
         "-fuse-ld="*)
-            args+=("-Xcc" "$arg")
+            # Link-only flag leaked into CFLAGS. -Xclang-linker passes it to the
+            # clang invocation swiftc uses *when it is used for linking* (not to
+            # the linker itself), instead of letting swiftc reject it at compile
+            # time. See swiftc's -Xclang-linker option documentation.
+            args+=("-Xclang-linker" "$arg")
             ;;
         # swiftc does not understand clang-specific include flags like
         # -isystem / -iquote / -idirafter; wrap them (and their following
--- a/Tools/yocto/meta-openembedded_and_meta-webkit.patch
+++ b/Tools/yocto/meta-openembedded_and_meta-webkit.patch
@@ -714,7 +714,7 @@ diff --git a/sources/meta-webkit/conf/distro/webkitdevci.conf b/sources/meta-web
 index 1f952c2..d299d1f 100644
 --- a/sources/meta-webkit/conf/distro/webkitdevci.conf
 +++ b/sources/meta-webkit/conf/distro/webkitdevci.conf
-@@ -47,6 +47,17 @@ PACKAGECONFIG:append:pn-weston-init = " no-idle-timeout"
+@@ -47,6 +47,18 @@ PACKAGECONFIG:append:pn-weston-init = " no-idle-timeout"
  PACKAGECONFIG:append:pn-php = " apache2"
  # Disable mysql support as we don't need to build mysql/mariadb
  PACKAGECONFIG:remove:pn-php = "mysql"
@@ -730,8 +730,9 @@ index 1f952c2..d299d1f 100644
  # Whitelist license from some deps.
  LICENSE_FLAGS_ACCEPTED:append = " commercial synaptics-killswitch"
 +
-+# Optional support for clang if meta-clang is enabled
-+CLANGSDK ?= "1"
++# The SDK host clang comes from the swift.org bundle (nativesdk-swift); keep
++# meta-clang's nativesdk-clang out of the SDK host to avoid a conflict.
++CLANGSDK = "0"
 diff --git a/sources/meta-webkit/conf/layer.conf b/sources/meta-webkit/conf/layer.conf
 index 124c13d..1a38638 100644
 --- a/sources/meta-webkit/conf/layer.conf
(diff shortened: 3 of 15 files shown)


nikolaszimmermann:
@adetaylor / @geoffreygaren - can you look at the non-yocto specific changes (CMakeLists.txt / WebKitMacros.cmake)?

nikolaszimmermann:
@psaavedra / @clopez - I'd appreciate if you could check the yocto integration - there are likely cleaner/better ways to integrate this.

Tools/CISupport/built-product-archive
```diff
@@ -335,6 +382,16 @@ def archiveBuiltProduct(configuration, platform, fullPlatform, minify=False):
...
 
+        # Swift-enabled cross builds: stage the Swift runtime libs from the
+        # cross sysroot into <build>/lib/ so deployed binaries can load them
+        # (the sysroot isn't present on the deploy target).
+        if crossTarget:
+            crossSysroot = _findCrossTargetSysroot(crossTarget)
+            if crossSysroot:
+                stagedLibs = _stageSwiftRuntimeLibs(crossSysroot,
+                                                   os.path.join(_configurationBuildDirectory, 'lib'))
+                contents.extend(os.path.join('lib', name) for name in stagedLibs)
```
clopez: Is this really needed?
The built products generated on a cross-toolchain environment are supossed to run on the image that is generated via `Tools/Scripts/cross-toolchain-helper --build-image --cross-target=$crossTarget` and that image is supposed to already include the swift runtime libs, isn't it?

Tools/yocto/meta-openembedded_and_meta-webkit.patch
```diff
@@ -799,3 +793,256 @@ index bdd7fea..7005240 100644
...
++    done
++
++    # swift.org binaries have a 168-byte .interp sized for the x86_64 loader.
++    # Yocto's relocate_sdk.py aborts when the SDK install path is >= p_filesz,
++    # so grow the slot with patchelf. The placeholder must be long enough and
++    # must not start with /lib*, /usr/lib32, or /usr/lib64 (relocate_sdk.py
++    # skips those as external markers); anchor it with ${SDKPATH}/ so the
++    # relocator still processes the binary, then pad to ~500 bytes.
++    pad=$(printf '%440s' '' | tr ' ' 'x')
++    long_interp="${SDKPATH}/${pad}/lib/ld-linux-x86-64.so.2"
```
clopez: I think that hard-coding the name of the interpreter to the `x86_64` one (`ld-linux-x86-64.so.2`) will cause a failure when `SDK_ARCH = aarch64`

clopez: Suggested change:

```
def sdk_arch_interpreter(d):
    sdk_arch = d.getVar('SDK_ARCH')
    if sdk_arch == 'x86_64':
        return 'ld-linux-x86-64.so.2'
    if sdk_arch == 'aarch64':
        return 'ld-linux-aarch64.so.1'
    bb.fatal('Unsupported SDK_ARCH for nativesdk-swift: {}'.format(sdk_arch))

SWIFT_SDK_INTERPRETER = "${@sdk_arch_interpreter(d)}"

[...]
long_interp="${SDKPATH}/${pad}/lib/${SWIFT_SDK_INTERPRETER}"
```

Tools/yocto/meta-openembedded_and_meta-webkit.patch
```diff
@@ -799,3 +793,256 @@ index bdd7fea..7005240 100644
...
+new file mode 100644
+index 0000000..4d5e6f7
+--- /dev/null
++++ b/sources/meta-webkit/recipes-devtools/swift/nativesdk-swift.bb
+@@ -0,0 +1,96 @@
++SUMMARY = "Swift toolchain bundle for the populate_sdk host (swift.org prebuilt)"
++HOMEPAGE = "https://swift.org/install/"
++LICENSE = "Apache-2.0"
++LIC_FILES_CHKSUM = "file://${S}/usr/share/swift/LICENSE.txt;md5=f6c482a0548ea60d6c2e015776534035"
++
```
psaavedra, line 881: Could I suggest here?: `COMPATIBLE_HOST = "(x86_64|aarch64).*-linux*"` ?

Tools/Scripts/swift/swiftc-wrapper.sh
```diff
@@ -87,10 +121,30 @@ for arg in "$@"; do
...
+                *)
+                    args+=("$name_only")
+                    # Skip the importer mirror when the shared -Xcc -D set already
+                    # defines this macro, otherwise the quoted copy here and the
+                    # unquoted shared copy collide as "macro redefined".
+                    if [[ -z "${xcc_defined[$name_only]}" ]]; then
+                        args+=("-Xcc" "$arg")
+                    fi
+                    ;;
+            esac
```
mcatanzaro: Did you review [my feedback from your previous pull request](https://github.com/WebKit/WebKit/pull/66014#pullrequestreview-4391719503)? You are now even further undoing @geoffreygaren's work from [314049@main](https://commits.webkit.org/314049@main).

adetaylor: Also related work in this area: 313188@main which didn't work and had to get reverted (I have yet to investigate why); #65122 which is a really important step towards getting the set of `-D` flags right.

nikolaszimmermann: > Did you review [my feedback from your previous pull request](https://github.com/WebKit/WebKit/pull/66014#pullrequestreview-4391719503)? You are now even further undoing @geoffreygaren's work from [314049@main](https://commits.webkit.org/314049@main).

I didn't! Sorry missed that, and also your comments regarding the wrong merge conflict resolve *sigh*.
I have been working on these branches for ~ 2 weeks, and rebased two times, messed it up the second time :/

Will check in detail tomorrow, thanks for the heads-up.

mcatanzaro: You got pretty unlucky with the timing of the commits tbh. I only noticed because I happened to be working on the same code, and got a conflict.

Tools/CISupport/built-product-archive
nikolaszimmermann: I tested on a rpi4 that didn't have the swift-enabled image (so no Swift runtime libs) -- we can omit this from the final patch - I used it for testing....

Tools/yocto/meta-openembedded_and_meta-webkit.patch
nikolaszimmermann: Good point! adapted.

Tools/yocto/meta-openembedded_and_meta-webkit.patch
nikolaszimmermann: Sure! Will add.

Tools/Scripts/swift/swiftc-wrapper.sh
nikolaszimmermann: Indeed, bad luck here. It works without the -D hacks after all.

nikolaszimmermann:
@mcatanzaro @adetaylor @clopez @psaavedra have another look please.

Tools/Scripts/swift/swiftc-wrapper.sh
```diff
@@ -70,7 +85,9 @@ for arg in "$@"; do
             ;;
         "-include") skip_next=1 ;;
         "-fuse-ld="*)
-            args+=("-Xcc" "$arg")
+            # Bare link-only flag leaked into CFLAGS; route to the linker so
+            # swiftc doesn't reject it at compile time.
+            args+=("-Xclang-linker" "$arg")
```
mcatanzaro, line 92: The code change is fine, but the comment here is wrong.

The [documentation](https://github.com/swiftlang/swift/blob/c4dcb70dc32e703674b5d434379c91b915223c72/include/swift/Option/Options.td#L1582) says: "Pass <arg> to Clang when it is used for linking." Which is indeed what we want to do. But your comment says it will be passed to the linker. That wouldn't make sense, because it's a compiler flag, not a linker flag.

Source/WTF/wtf/CMakeLists.txt
```diff
@@ -789,6 +789,13 @@ elseif (CMAKE_SYSTEM_NAME MATCHES "Linux")
     )
 endif ()
 
+# wtf/module.modulemap references ios/WebCoreThread.h, which CMake builds never
+# copied. The Swift C++ interop path makes Clang parse the modulemap, so the wtf
+# module fails to build without it. Harmless off-iOS (gated on USE(WEB_THREAD)).
+list(APPEND WTF_PUBLIC_HEADERS
+    ios/WebCoreThread.h
+)
```
mcatanzaro: It's not harmless. You have to move the file to a non-ios directory, or it won't exist at all in release tarballs, since platform-specific directories are excluded.

EWS does not build from tarballs, and so will not notice. But distros do not build from git!

nikolaszimmermann: Oh good point - didn't take that into account... so another solution is needed.

mcatanzaro: (Um: this is not to say that moving it to a non-iOS directory is the _correct_ thing to do, only that you definitely cannot use files in iOS directories.)

clopez: Good point. I would suggest to move that files to the directory above.

Tools/yocto/meta-openembedded_and_meta-webkit.patch
```diff
@@ -799,3 +792,261 @@ index bdd7fea..7005240 100644
...
++EXTRA_OECMAKE:append:arm = " \
++    -DCMAKE_Swift_COMPILER=${S}/Tools/Scripts/swift/swiftc-wrapper.sh \
++    -DCMAKE_Swift_FLAGS_INIT='-sdk ${STAGING_DIR_TARGET} -resource-dir ${STAGING_DIR_TARGET}${libdir}/swift' \
++    "
++
++EXTRA_OECMAKE:append:riscv64 = " \
++    -DENABLE_SWIFT_DEMO_URI_SCHEME=OFF \
++    -DENABLE_BACK_FORWARD_LIST_SWIFT=OFF \
++    "
```
clopez: This chunk is not really needed here as it doesn't have any practical effect.
The wpewebkit recipe from meta-webkit is not used to build webkit via the script build-webkit.

If the intention is to pass `-DCMAKE_Swift_COMPILER=xxx` and `-DCMAKE_Swift_FLAGS_INIT=zzz` to the build when using `build-webkit --cross-target=yyy` then the place to put this flags is not here but in `Tools/yocto/targets.conf` (check where there the flags are set under the setting `environment[BUILD_WEBKIT_ARGS]` for each target.

clopez: However this chunk needs to be contributed back to meta-webkit. But i would suggest to redo it so instead of applying it by default to the recipe it applies conditionally (opt-in) via a PACKAGECONFIG option.

Example:

```
PACKAGECONFIG[swift] = "-DENABLE_SWIFT_DEMO_URI_SCHEME=ON DENABLE_BACK_FORWARD_LIST_SWIFT=ON -DCMAKE_Swift_COMPILER=${S}/Tools/Scripts/swift/swiftc-wrapper.sh  -DCMAKE_Swift_FLAGS_INIT='-sdk ${STAGING_DIR_TARGET} -resource-dir ${STAGING_DIR_TARGET}${libdir}/swift',-DENABLE_SWIFT_DEMO_URI_SCHEME=OFF DENABLE_BACK_FORWARD_LIST_SWIFT=OFF,swift-native swift-stdlib swift-foundation libdispatch"
```
And then we have to ensure that `Tools/Scripts/swift/swiftc-wrapper.sh` is shipped on the tarballs. So perhaps we need to move that script to a more canonical path. I suggest moving it to `Source/WTF/Scripts`

Tools/yocto/meta-openembedded_and_meta-webkit.patch
```diff
@@ -799,3 +792,261 @@ index bdd7fea..7005240 100644
...
+diff --git a/sources/meta-webkit/recipes-devtools/swift/nativesdk-swift.bb b/sources/meta-webkit/recipes-devtools/swift/nativesdk-swift.bb
+new file mode 100644
+index 0000000..4d5e6f7
+--- /dev/null
++++ b/sources/meta-webkit/recipes-devtools/swift/nativesdk-swift.bb
+@@ -0,0 +1,106 @@
++SUMMARY = "Swift toolchain bundle for the populate_sdk host (swift.org prebuilt)"
++HOMEPAGE = "https://swift.org/install/"
++LICENSE = "Apache-2.0"
++LIC_FILES_CHKSUM = "file://${S}/usr/share/swift/LICENSE.txt;md5=f6c482a0548ea60d6c2e015776534035"
```
clopez, line 880: Can we send a PR to meta-swift with this recipe? so if/when they accept it then we can stop carrying it ourselves.

Tools/yocto/meta-openembedded_and_meta-webkit.patch
```diff
@@ -799,3 +792,261 @@ index bdd7fea..7005240 100644
...
+++          endif()
++         else()
+diff --git a/sources/meta-swift/recipes-devtools/swift/swift-stdlib.bb b/sources/meta-swift/recipes-devtools/swift/swift-stdlib.bb
+--- a/sources/meta-swift/recipes-devtools/swift/swift-stdlib.bb
++++ b/sources/meta-swift/recipes-devtools/swift/swift-stdlib.bb
+@@ -17,6 +17,7 @@
+     git://github.com/swiftlang/swift-experimental-string-processing.git;protocol=https;name=stringproc;tag=${SWIFT_TAG};nobranch=1;destsuffix=swift-experimental-string-processing; \
+     git://github.com/swiftlang/swift-syntax.git;protocol=https;name=syntax;tag=${SWIFT_TAG};nobranch=1;destsuffix=swift-syntax; \
+     file://0003-add-fix-for-metadataaccessor-abi-mismatch.patch;striplevel=1; \
++    file://0004-allow-overriding-SWIFT_SDK_LINUX_ARCH_TRIPLE.patch;striplevel=1; \
```
clopez, line 1064: is this change upstream worth? should we send a PR to meta-swift with it?

Tools/yocto/meta-openembedded_and_meta-webkit.patch
```diff
@@ -799,3 +792,261 @@ index bdd7fea..7005240 100644
...
++    src=${STAGING_DIR_TARGET}/usr/lib/${TARGET_SYS}/${SWIFT_GCC_VERSION}
++    dst=${STAGING_DIR_TARGET}/usr/lib/gcc/${TARGET_SYS}/${SWIFT_GCC_VERSION}
++    [ -d "${src}" ] || return 0
++    mkdir -p "${dst}"
++    for f in "${src}"/*; do
++        bn=$(basename "${f}")
++        [ -e "${dst}/${bn}" ] || cp -rp "${f}" "${dst}/${bn}"
++    done
+ }
+
```
clopez, line 1020: also i wonder if this one is worth considering for sending a PR to meta-swift?

Tools/Scripts/cross-toolchain-helper
```diff
@@ -391,6 +391,12 @@ class YoctoCrossBuilder():
                 with open(extra_image_fragment_file, 'rb') as src, open(local_conf_path, 'ab') as dst:
                     dst.write(b'\n')
                     dst.write(src.read())
+            # Copy any sibling *.conf.inc fragments next to local.conf so bitbake's
+            # `require <name>.conf.inc` directives resolve in the build conf dir.
+            local_conf_srcdir = os.path.dirname(self._targets_config.get_conf_local(self._target))
+            for sibling in os.listdir(local_conf_srcdir):
+                if sibling.endswith('.conf.inc'):
+                    shutil.copy2(os.path.join(local_conf_srcdir, sibling), os.path.join(conf_dir, sibling))
```
clopez: Instead of doing this I think is better to include the contents of the .conf.inc inside all the local-rpi*-mesa.conf files that should enable it.

The main reason for that is that this patch is not adding the file .conf.inc to the list of the files to be hashed at `YoctoTargetsConfig.get_hash_for_target()`, so any change to the .conf.inc file will not trigger an automatic wipe&rebuild cycle, so that can cause issues with stale builds (for example on the bots).

My suggestion would be to simplify this, drop the .conf.inc file and instead of adding a require line add the the content of the .conf.inc inside the .conf files

Tools/Scripts/cross-toolchain-helper
```diff
@@ -546,59 +552,129 @@ class YoctoCrossBuilder():
...
+                            '\nif [ -n "${BUILD_WEBKIT_ARGS:-}" ] && [ -d "${OECORE_TARGET_SYSROOT}/usr/lib/swift" ]; then\n'
+                            '    _target_prefix="${TARGET_PREFIX%-}"\n'
+                            '    _gcc_install_dir=$(ls -d "${OECORE_TARGET_SYSROOT}/usr/lib/gcc/${_target_prefix}"/* 2>/dev/null | head -1)\n'
+                            '    _cross_ld=$(command -v "${_target_prefix}-ld" 2>/dev/null)\n'
+                            '    _swift_flags="-sdk ${OECORE_TARGET_SYSROOT} -I ${OECORE_TARGET_SYSROOT}/usr/lib/swift/linux -I ${OECORE_TARGET_SYSROOT}/usr/lib/swift -Xcc --sysroot=${OECORE_TARGET_SYSROOT} -Xcc --gcc-install-dir=${_gcc_install_dir} -Xclang-linker --sysroot=${OECORE_TARGET_SYSROOT} -Xclang-linker --gcc-install-dir=${_gcc_install_dir} -Xclang-linker -fuse-ld=${_cross_ld}"\n'
+                            '    BUILD_WEBKIT_ARGS="${BUILD_WEBKIT_ARGS} --cmakeargs=\\"-DCMAKE_Swift_FLAGS_INIT=\'${_swift_flags}\'\\""\n'
+                            '    export BUILD_WEBKIT_ARGS\n'
+                            '    unset _swift_flags _target_prefix _gcc_install_dir _cross_ld\n'
+                            'fi\n')
```
clopez, line 688: Why is needed to set here `fuse-ld=` ? Please add some comments on the code explaining why this is needed.

Also `-Xcc` and `-Xclang-linker` are repeated twice

Tools/Scripts/cross-toolchain-helper
```diff
@@ -546,59 +552,129 @@ class YoctoCrossBuilder():
...
+                                if '="' in clang_line and not clang_line.endswith('"'):
+                                    clang_line += '"'
+                                gcc_lines.append(gcc_line.strip() + '\n')
+                                clang_lines.append(clang_line.strip() + '\n')
                             else:
                                 fw.write(line)
-                    # The script should check the CC/CXX env var from the user, if is set to clang then export
-                    # the CLANGCC and CLANGCXX as CC/CXX, otherwise keep the default (GCC)
-                    if len(gcc_lines) != 3:
-                        raise RuntimeError('Unexpected CC/CXX/CPP parameters in setup-environment script')
```
clopez, line 581: Why removing this sanity check?

Tools/Scripts/cross-toolchain-helper
```diff
@@ -546,59 +552,129 @@ class YoctoCrossBuilder():
...
+            # 2. On the CC/CXX/CPP exports, keep only machine flags (-m*), '-E', and
+            #    the sysroot.
+            # 3. If the SDK ships clang (swift.org bundle), also emit a Clang form
+            #    swapping `<sys>-gcc/g++/cpp` for `clang/clang++/clang-cpp --target=<sys>`
+            #    (Swift/C++ interop is Clang-only), selected by the user via CC/CXX.
             patched_setup_env_path = original_setup_env_path + "-no-opt-flags"
-            flags_regex = re.compile(r'export\s+(C|CXX|LD|CPP)FLAGS=')
-            gcc_compiler_regex = re.compile(r'export\s+C(C|XX|PP)=')
-            clang_compiler_regex = re.compile(r'export\s+CLANGC(C|XX|PP)=')
+            flags_regex = re.compile('export +(C|CXX|LD|CPP)FLAGS')
```
clopez: why changing `flags_regex`? the original one was more defensive, this new version can silently drop something like `export CPPFLAGS_FOR_BUILD=...` or fail to filter if there is tab character between `export` and the flagname

geoffreygaren:
I'm super unfamiliar with the context here, but I don't see any obvious problems.

Tools/Scripts/swift/swiftc-wrapper.sh
nikolaszimmermann:
```
            # Link-only flag leaked into CFLAGS. -Xclang-linker passes it to the
            # clang invocation swiftc uses *when it is used for linking* (not to
            # the linker itself), instead of letting swiftc reject it at compile
            # time. See swiftc's -Xclang-linker option documentation.
```

Better?

Source/WTF/wtf/CMakeLists.txt
nikolaszimmermann: Dropped.

Resolved by this upstream:

```
commit e5aa433da9cf28f14181eba69cb5e56a20e892c1
Author: Claudio Saavedra <csaavedra@igalia.com>
Date:   Fri May 29 11:37:30 2026 -0700

    [GLIB] Build broken with clang and swift
    https://bugs.webkit.org/show_bug.cgi?id=315866
    
    Reviewed by Nikolas Zimmermann and Adrian Perez de Castro.
    
    Source/WTF/wtf/module.modulemap declares an explicit `WebCoreThread`
    submodule pointing at ios/WebCoreThread.h, which is staged only on
    Cocoa builds. Clang validates header existence at modulemap parse time
    even for unused explicit submodules, so Swift's clang importer fails
    with "header 'ios/WebCoreThread.h' not found" and the wtf module never
    loads.
    
    Mirror the existing wtf/spi/ stub: on non-Apple SWIFT_REQUIRED builds,
    write an empty wtf/ios/WebCoreThread.h into the forwarding headers dir.
    
    * Source/WTF/wtf/CMakeLists.txt:
```

Tools/yocto/meta-openembedded_and_meta-webkit.patch
nikolaszimmermann: Added the PACKAGECONFIG[swift] knob - but kept the script in the original location -- @bertogg fixed the release tarballs already to include that script.

Tools/yocto/meta-openembedded_and_meta-webkit.patch
nikolaszimmermann: Yes, but I want to get consensus here first, then try upstreaming.

Tools/yocto/meta-openembedded_and_meta-webkit.patch
nikolaszimmermann: Will attempt that.

Tools/Scripts/cross-toolchain-helper
nikolaszimmermann: Sounds good. Done.

Tools/Scripts/cross-toolchain-helper
nikolaszimmermann: Restored, indeed an oversight.

Tools/Scripts/cross-toolchain-helper
nikolaszimmermann:
```
-            flags_regex = re.compile(r'export\s+(C|CXX|LD|CPP)FLAGS=')
-            gcc_compiler_regex = re.compile(r'export\s+C(C|XX|PP)=')
-            clang_compiler_regex = re.compile(r'export\s+CLANGC(C|XX|PP)=')
+            # Anchor on the trailing '=' so we don't also strip unrelated exports
+            # like CPPFLAGS_FOR_BUILD, and allow tabs between 'export' and the name.
+            flags_regex = re.compile(r'export\s+(C|CXX|LD|CPP)FLAGS\s*=')
+            compiler_regex = re.compile(r'export\s+C(C|XX|PP)\s*=')
+            toolchain_binary_regex = re.compile(r'([\w-]+-)(gcc|g\+\+|cpp)$')
```

Restored.

nikolaszimmermann:
Ready for next review.

Tools/Scripts/swift/swiftc-wrapper.sh
mcatanzaro: Sure.

Tools/Scripts/cross-toolchain-helper
```diff
@@ -546,59 +546,145 @@ class YoctoCrossBuilder():
...
-                    # The script should check the CC/CXX env var from the user, if is set to clang then export
-                    # the CLANGCC and CLANGCXX as CC/CXX, otherwise keep the default (GCC)
-                    if len(gcc_lines) != 3:
-                        raise RuntimeError('Unexpected CC/CXX/CPP parameters in setup-environment script')
-                    if clang_lines:
+                    # Fail loudly if the setup-environment file changed shape and we
+                    # matched no CC/CXX/CPP exports: emitting a patched file with no
+                    # compiler exports would silently break the cross build later.
+                    if not gcc_lines:
+                        raise RuntimeError('No CC/CXX/CPP exports found in setup-environment script: {path}'.format(path=original_setup_env_path))
```
clopez: please keep the original check against 3, i prefer to be strict here, if the number changes from 3 this has to be inspected to see what is going on.

	
```
# The script should check the CC/CXX env var from the user, if is set to clang then export
# the CLANGCC and CLANGCXX as CC/CXX, otherwise keep the default (GCC)
if len(gcc_lines) != 3:
    raise RuntimeError('Unexpected CC/CXX/CPP parameters in setup-environment script')
```

Tools/Scripts/cross-toolchain-helper
```diff
@@ -546,59 +546,145 @@ class YoctoCrossBuilder():
...
+            #    swapping `<sys>-gcc/g++/cpp` for `clang/clang++/clang-cpp --target=<sys>`
+            #    (Swift/C++ interop is Clang-only), selected by the user via CC/CXX.
             patched_setup_env_path = original_setup_env_path + "-no-opt-flags"
-            flags_regex = re.compile(r'export\s+(C|CXX|LD|CPP)FLAGS=')
-            gcc_compiler_regex = re.compile(r'export\s+C(C|XX|PP)=')
-            clang_compiler_regex = re.compile(r'export\s+CLANGC(C|XX|PP)=')
+            # Anchor on the trailing '=' so we don't also strip unrelated exports
+            # like CPPFLAGS_FOR_BUILD, and allow tabs between 'export' and the name.
+            flags_regex = re.compile(r'export\s+(C|CXX|LD|CPP)FLAGS\s*=')
+            compiler_regex = re.compile(r'export\s+C(C|XX|PP)\s*=')
```
clopez: the last `\s*` before the `=` is wrong. In bash/shell you can't have a space or tab character before the `=` sign when declaring a variable. Please keep the original regexes.

Tools/Scripts/cross-toolchain-helper
nikolaszimmermann: Ouch, forgot to restore that -- good catch.

Tools/Scripts/cross-toolchain-helper
nikolaszimmermann: Good catch too, will restore.

Tools/Scripts/cross-toolchain-helper
clopez: This question remains unanswered.
But not a blocker, we can land it and then later try to check if this fuse-ld parameter or those repeated ones can be removed

clopez (approved):

Tools/Scripts/cross-toolchain-helper
nikolaszimmermann: The repetition is on purpose (see the logic for Xclang-linke here: https://app.semanticdiff.com/gh/swiftlang/swift/pull/20441/commits/219feaf1893f2bd281545975fa878a798ee4bd2c#test/Driver/linker.swift?ignore_comments=false)

When swiftc links the final target binary, it doesn't link directly - it spawns a clang driver to do the lniking. That clang driver picks a linker and by default it resolves ld from the PATH not the Yocto cross-binutils ld... I hope I remembered that correctly, it's a while ago and I didn't retest... Landing it now.

PR #66314, opened 2026-06-03
AX: Atomically transfer pending tree updates from main thread to AX thread

twilco:
#### 30446586d80c473d54cc344b85404af37e1f536c
<pre>
AX: Atomically transfer pending tree updates from main thread to AX thread
<a href="https://bugs.webkit.org/show_bug.cgi?id=316148">https://bugs.webkit.org/show_bug.cgi?id=316148</a>
<a href="https://rdar.apple.com/178573374">rdar://178573374</a>

Reviewed by Andres Gonzalez.

Before this change, every main-thread writer acquired m_changeLogLock briefly
to mutate m_pendingChanges directly. Because the lock was acquired and released
per-mutation rather than once per logical work cycle, the AX thread could take
a snapshot mid-update-cycle and observe partial state. As an example,
performDeferredCacheUpdate calls tree-&gt;setFocusedNodeID(X) (publishing focus
immediately), but X&apos;s append into m_pendingChanges-&gt;appends is still queued in
m_unresolvedPendingAppends and only materialized later by
processQueuedNodeUpdates. If the AX thread takes the snapshot between these two writes,
it observes focusedNodeID = X with no corresponding appends entry and reports
focus on a node that isn&apos;t in m_readerThreadNodeMap.

This commit reshapes main-thread writes so the AX thread can never observe a
partial update. Introduce a two-buffer model:

  - m_workingChanges: main-thread-only, no lock. Writes accumulate here via
    markDirtyAndGetWorkingChanges().
  - m_committedChanges: cross-thread, guarded by m_changeLogLock. The AX thread
    reads only this buffer, via takeCommittedChangesLocked().

commitWorkingChanges() merges working into committed in a single critical
section via PendingChanges::merge(), so a logical work cycle becomes one
indivisible state change. Field-specific merge semantics mirror how
applyCommittedChanges consumes each field. Vectors concatenate, sets and maps
union, and last-write-wins fields only overwrite when the source wrote this
cycle.

The one-shot scheduleQueuedNodeUpdateProcessing() timer drops from 100ms to 17ms, so a
batch is processed roughly one frame after its first change.

The eager-apply dedup flag (m_appliedOrApplyingCommittedChanges) becomes a plain bool
guarded by m_changeLogLock, and is modified inside the same lock-section that commits
the working changes.

* Source/WTF/wtf/SystemTracing.h:
(WebCore::AXCoreObject::unignoredChildren):
(WebCore::AXObjectCache::onAccessibilityPaintFinished):
(WebCore::AXObjectCache::dirtyIsolatedTreeRelations):
(WebCore::AXObjectCache::startUpdateTreeSnapshotTimer):
(WebCore::AXObjectCache::updateTreeSnapshotTimerFired):
* Source/WebCore/accessibility/AXObjectCache.h:
* Source/WebCore/accessibility/AXTreeStore.cpp:
(WebCore::AXTreeStore&lt;AXIsolatedTree&gt;::applyPendingChangesForAllIsolatedTrees):
(WebCore::AXIsolatedObject::approximateHitTest const):
(WebCore::AXIsolatedTree::createEmpty):
(WebCore::AXIsolatedTree::createEmptyContent):
(WebCore::AXIsolatedTree::create):
(WebCore::AXIsolatedTree::storeTree):
(WebCore::AXIsolatedTree::removeTreeForFrameID):
(WebCore::AXIsolatedTree::queueChange):
(WebCore::AXIsolatedTree::addUnconnectedNode):
(WebCore::AXIsolatedTree::queueRemovals):
(WebCore::AXIsolatedTree::queueAppendsAndRemovals):
(WebCore::AXIsolatedTree::updateNode):
(WebCore::AXIsolatedTree::updateNodeProperties):
(WebCore::AXIsolatedTree::overrideNodeProperties):
(WebCore::AXIsolatedTree::pendingRootNodeID):
(WebCore::AXIsolatedTree::setPendingRootNodeID):
(WebCore::AXIsolatedTree::setFocusedNodeID):
(WebCore::AXIsolatedTree::updateRelations):
(WebCore::AXIsolatedTree::setSelectedTextMarkerRange):
(WebCore::AXIsolatedTree::setFrameGeometry):
(WebCore::AXIsolatedTree::updateFrame):
(WebCore::AXIsolatedTree::applyPendingChanges):
(WebCore::AXIsolatedTree::applyPendingChangesUnlessQueuedForDestruction):
(WebCore::AXIsolatedTree::applyPendingChangesOrTearDown):
(WebCore::AXIsolatedTree::clearTreeContentsLocked):
(WebCore::AXIsolatedTree::commitWorkingChanges):
(WebCore::AXIsolatedTree::takeCommittedChangesLocked):
(WebCore::AXIsolatedTree::applyCommittedChanges):
(WebCore::AXIsolatedTree::sortedLiveRegionsDidChange):
(WebCore::AXIsolatedTree::sortedNonRootWebAreasDidChange):
(WebCore::AXIsolatedTree::queueNodeUpdate):
(WebCore::AXIsolatedTree::queueNodeRemoval):
(WebCore::AXIsolatedTree::processQueuedNodeUpdates):
(WebCore::AXIsolatedTree::queueRemovalsLocked): Deleted.
(WebCore::AXIsolatedTree::setPendingRootNodeIDLocked): Deleted.
(WebCore::AXIsolatedTree::takePendingChangesLocked): Deleted.
(WebCore::AXIsolatedTree::applyPendingChangesFromSnapshot): Deleted.
(WebCore::AXIsolatedTree::markRelationsDirty):
(WebCore::AXIsolatedTree::markMostRecentlyPaintedTextDirty):
(WebCore::AXIsolatedTree::PendingChanges::setFocusedNodeID):
(WebCore::AXIsolatedTree::markDirtyAndGetWorkingChanges):
(WebCore::AXIsolatedTree::PendingChangesAccessor::PendingChangesAccessor): Deleted.
(WebCore::AXIsolatedTree::PendingChangesAccessor::~PendingChangesAccessor): Deleted.
(WebCore::AXIsolatedTree::PendingChangesAccessor::operator-&gt;): Deleted.
(WebCore::AXIsolatedTree::WTF_REQUIRES_LOCK): Deleted.

Canonical link: <a href="https://commits.webkit.org/319200@main">https://commits.webkit.org/319200@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/4c8fa0de6f40967fc1a32256a5fe67ae6977d8de

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/180627 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/54572 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/48370 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/190838 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/144949 "Built successfully") | [❌ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/e3252316-50ed-4974-aa91-ce0834ac5368) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/182489 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/55870 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/54288 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/140883 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/97608 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [❌ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/601810d9-f7e0-4679-8852-df26d15b47cf) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/183582 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/44557 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/161659 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/121335 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/d14c070c-dd1d-4dab-8b81-6e6d4ec23afc) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/43230 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/41511 "Passed tests") | [✅ 🧪 jsc-wpe](https://ews-build.webkit.org/#/builders/251/builds/3302 "Passed tests") | | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/10144 "Passed tests") | [loading-orange 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/153634 "Build is in progress. Recent messages:OS: Tahoe (26.5), Xcode: 26.5; Checked out pull request; 1 api test failed or timed out") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/41188 "Passed tests") | [✅ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/1998 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/41901 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/47181 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/45766 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/192774 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/54238 "Built successfully") | | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/150259 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/40250 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/54926 "Built successfully") | [❌ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/37805 "Too many flaky failures: http/tests/cache/disk-cache/resource-becomes-uncacheable.html, http/tests/contentextensions/video-element-resource-type.html, http/tests/misc/resource-timing-navigation-in-restored-iframe-2.html, http/tests/navigation/javascriptlink-frames.html, http/tests/navigation/post-redirect-get-reload.py, http/tests/resourceLoadStatistics/input-autofilled-user-interaction.html, http/tests/security/cached-cross-origin-shared-css-stylesheet.html, http/tests/security/contentSecurityPolicy/1.1/scriptnonce-allowed-by-enforced-policy-and-blocked-by-report-policy2.py, http/tests/security/contentSecurityPolicy/frame-src-about-blank-allowed-by-scheme.html, http/tests/security/mixedContent/secure-redirect-to-insecure-redirect-to-basic-auth-secure-image-UpgradeMixedContent.https.html (failure)") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/149977 "Passed tests") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/44598 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/127191 "Built successfully") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/122406 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/53443 "Built successfully") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/16181 "") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/52825 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/53062 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/52890 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/accessibility/AXObjectCache.cpp
+++ b/Source/WebCore/accessibility/AXObjectCache.cpp
@@ -403,7 +403,10 @@ bool AXObjectCache::shouldServeInitialCachedFrame()
     return !clientIsInTestMode() || forceInitialFrameCaching();
 }
 
-static constexpr Seconds updateTreeSnapshotTimerInterval { 100_ms };
+// Just over one frame at 60Hz (16.67ms). The timer is scheduled on demand when the isolated
+// tree first queues work, so this is "publish one frame after the first change in this batch",
+// with any further changes in the window coalescing into the same commit.
+static constexpr Seconds updateTreeSnapshotTimerDuration { 17_ms };
 #endif
 
 AXObjectCache::AXObjectCache(LocalFrame& localFrame, Document* document)
@@ -2125,6 +2128,9 @@ void AXObjectCache::notificationPostTimerFired()
 
 #if ENABLE(ACCESSIBILITY_ISOLATED_TREE)
     updateIsolatedTree(notificationsToPost);
+    // We're going to post platform notifications, so make sure to publish isolated tree
+    // changes so we provide up-to-date information.
+    processQueuedIsolatedNodeUpdates();
 #endif
 
     for (const auto& note : notificationsToPost) {
@@ -3078,7 +3084,6 @@ void AXObjectCache::onAccessibilityPaintFinished()
     }
 
     tree->markMostRecentlyPaintedTextDirty();
-    startUpdateTreeSnapshotTimer();
 }
 
 bool AXObjectCache::onFontChange(Element& element, const Style::ComputedStyle* oldStyle, const Style::ComputedStyle* newStyle)
@@ -4294,7 +4299,6 @@ void AXObjectCache::dirtyIsolatedTreeRelations()
 #if ENABLE(ACCESSIBILITY_ISOLATED_TREE)
     if (RefPtr tree = AXIsolatedTree::treeForFrameID(m_frameID))
         tree->markRelationsDirty();
-    startUpdateTreeSnapshotTimer();
 #endif
 }
 
@@ -6294,7 +6298,7 @@ void AXObjectCache::updateIsolatedTree(AccessibilityObject& axObject, AXProperty
 void AXObjectCache::startUpdateTreeSnapshotTimer()
 {
     if (!m_updateTreeSnapshotTimer.isActive())
-        m_updateTreeSnapshotTimer.startOneShot(updateTreeSnapshotTimerInterval);
+        m_updateTreeSnapshotTimer.startOneShot(updateTreeSnapshotTimerDuration);
 }
 
 void AXObjectCache::onPaint(const RenderObject& renderer, IntRect&& paintRect) const
@@ -7302,12 +7306,6 @@ void AXObjectCache::selectedTextRangeTimerFired()
     }
 
     m_lastDebouncedTextRangeObject = std::nullopt;
-}
-
-void AXObjectCache::updateTreeSnapshotTimerFired()
-{
-    m_updateTreeSnapshotTimer.stop();
-    processQueuedIsolatedNodeUpdates();
 }
 
 void AXObjectCache::processQueuedIsolatedNodeUpdates()
--- a/Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
+++ b/Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
@@ -110,8 +110,18 @@ void AXIsolatedTree::queueForDestruction()
     AX_ASSERT(isMainThread());
 
     Locker locker { m_changeLogLock };
-    m_queuedForDestruction = true;
+    // Publish anything still in the working buffer before the AX thread tears this tree down.
+    // Teardown (clearTreeContentsLocked) only clears m_committedChanges, and every NodeChange
+    // left behind holds a Ref to this tree (IsolatedObjectData::tree), so stranding them would
+    // be a self-reference cycle that keeps the tree alive forever.
+    commitWorkingChangesLocked();
+    // Note that setting m_hasPendingChanges here is deliberately redundant (since commitWorkingChages
+    // also sets it). It's repeated here because teardown depends on it, and would fail silently without
+    // it, since applyPendingChangesOrTearDown early-exits if not set. Setting it here makes queueForDestruction()
+    // correct no matter what commitWorkingChanges() does, now or in the future.
     m_hasPendingChanges.store(true);
+
+    m_queuedForDestruction = true;
     s_anyTreeNeedsTearDown.store(true, std::memory_order_relaxed);
 }
 
@@ -131,12 +141,12 @@ Ref<AXIsolatedTree> AXIsolatedTree::createEmpty(AXObjectCache& axObjectCache)
     tree->updateLoadingProgress(axObjectCache.loadingProgress());
     tree->m_processingProgress = 0;
 
-    // Now that the tree is ready to take client requests, add it to the tree maps so that it can be found.
-    storeTree(axObjectCache, tree);
-
 #if ENABLE(ACCESSIBILITY_LOCAL_FRAME)
     tree->updateFrameGeometryAndScrollPositionIfNeeded(axObjectCache);
 #endif
+
+    // Now that the tree is ready to take client requests, add it to the tree maps so that it can be found.
+    storeTree(axObjectCache, tree);
 
     return tree;
 }
@@ -174,11 +184,8 @@ void AXIsolatedTree::createEmptyContent(AccessibilityObject& axRoot)
     m_nodeMap.set(axRoot.objectID(), ParentChildrenIDs { std::nullopt, { axWebArea->objectID() } });
     m_nodeMap.set(axWebArea->objectID(), ParentChildrenIDs { axRoot.objectID(), { } });
 
-    {
-        Locker locker { m_changeLogLock };
-        setPendingRootNodeIDLocked(axRoot.objectID());
-        mutablePendingChanges()->focusedNodeID = axWebArea->objectID();
-    }
+    setPendingRootNodeID(axRoot.objectID());
+    markDirtyAndGetWorkingChanges().focusedNodeID = Markable<AXID> { axWebArea->objectID() };
     Vector<NodeChange> appends;
     appends.reserveInitialCapacity(2);
     appends.append(WTF::move(rootAppend));
@@ -237,12 +244,12 @@ RefPtr<AXIsolatedTree> AXIsolatedTree::create(AXObjectCache& axObjectCache)
         addUnconnectedNodeIfNeeded(targetID);
     tree->updateRelations(WTF::move(relations));
 
-    // Now that the tree is ready to take client requests, add it to the tree maps so that it can be found.
-    storeTree(axObjectCache, tree);
-
 #if ENABLE(ACCESSIBILITY_LOCAL_FRAME)
     tree->updateFrameGeometryAndScrollPositionIfNeeded(axObjectCache);
 #endif
+
+    // Now that the tree is ready to take client requests, add it to the tree maps so that it can be found.
+    storeTree(axObjectCache, tree);
 
     if (AXObjectCache::isAppleInternalInstall()) [[unlikely]]
         WTFEndSignpostAlways(tree.ptr(), InitialAccessibilityIsolatedTreeBuild);
@@ -253,14 +260,25 @@ void AXIsolatedTree::storeTree(AXObjectCache& cache, const Ref<AXIsolatedTree>& 
 {
     AX_ASSERT(isMainThread());
 
+    // Atomically publish any working-buffer writes into m_committedChanges before the tree
+    // becomes externally visible.
+    auto shouldEagerlyApply = tree->commitWorkingChanges();
+
     // Once we've added this new tree to the AXTreeStore, clients will be able to use
     // it off the main-thread. Make any final state mutations while we are the only thread
     // that can touch this tree.
     cache.setIsolatedTree(tree);
     AXTreeStore::set(tree->treeID(), tree.ptr());
     tree->m_replacingTree = nullptr;
-    Locker locker { s_storeLock };
-    treeFrameCache().set(cache.frameID(), tree.copyRef());
+
+    {
+        Locker locker { s_storeLock };
+        treeFrameCache().set(cache.frameID(), tree.copyRef());
+    }
+    AX_ASSERT(!tree->m_workingChangesDirty);
+
+    // Now that the tree is reachable off the main thread, hand the initial contents over.
+    tree->dispatchEagerApplyIfNeeded(shouldEagerlyApply);
 }
 
 double AXIsolatedTree::loadingProgress()
@@ -366,13 +384,12 @@ std::optional<AXIsolatedTree::NodeChange> AXIsolatedTree::nodeChangeForObject(Re
 void AXIsolatedTree::queueChange(NodeChange&& nodeChange)
 {
     AX_ASSERT(isMainThread());
-    AX_ASSERT(m_changeLogLock.isLocked());
 
     AXID objectID = nodeChange.data.axID;
     Markable parentID = nodeChange.data.parentID;
     bool isTextControl = AXCoreObject::isTextControl(nodeChange.data.role);
-    auto pending = mutablePendingChanges();
-    pending->appends.append(WTF::move(nodeChange));
+    auto& pending = markDirtyAndGetWorkingChanges();
+    pending.appends.append(WTF::move(nodeChange));
 
     // An append carries a full, fresh copy of the node's properties, so it supersedes a SelectedTextRange
     // update already queued for this object in an earlier cycle. Drop that stale update; without this it
@@ -385,27 +402,27 @@ void AXIsolatedTree::queueChange(NodeChange&& nodeChange)
     // same-cycle dedup in processQueuedNodeUpdates() already assumes), but we limit this to SelectedTextRange
     // for now since other properties may intentionally be pushed as targeted updates. Consider generalizing.
     if (isTextControl) {
-        for (auto& change : pending->propertyChanges) {
+        for (auto& change : pending.propertyChanges) {
             if (change.axID == objectID) {
                 change.properties.removeAllMatching([] (const auto& property) {
                     return property.first == AXProperty::SelectedTextRange;
                 });
             }
         }
-        pending->propertyChanges.removeAllMatching([] (const auto& change) {
+        pending.propertyChanges.removeAllMatching([] (const auto& change) {
             return change.properties.isEmpty();
         });
     }
 
     if (parentID) {
         auto siblingsIDs = m_nodeMap.get(*parentID).childrenIDs;
-        pending->childrenUpdates.append({ *parentID, WTF::move(siblingsIDs) });
+        pending.childrenUpdates.append({ *parentID, WTF::move(siblingsIDs) });
     }
 
     ASSERT_WITH_MESSAGE(objectID != parentID, "object ID was the same as its parent ID (%s) when queueing a node change", objectID.loggingString().utf8().data());
     ASSERT_WITH_MESSAGE(m_nodeMap.contains(objectID), "node map should've contained objectID: %s", objectID.loggingString().utf8().data());
     auto childrenIDs = m_nodeMap.get(objectID).childrenIDs;
-    pending->childrenUpdates.append({ objectID, WTF::move(childrenIDs) });
+    pending.childrenUpdates.append({ objectID, WTF::move(childrenIDs) });
 }
 
 void AXIsolatedTree::addUnconnectedNode(Ref<AccessibilityObject> axObject)
@@ -431,33 +448,27 @@ void AXIsolatedTree::addUnconnectedNode(Ref<AccessibilityObject> axObject)
     m_unconnectedNodes.add(objectID);
 
     // Because we are queuing a change for an object not intended to be connected to the rest of the tree,
-    // we don't need to update m_nodeMap or m_pendingChanges.childrenUpdates for this object or its parent as is
+    // we don't need to update m_nodeMap or m_workingChanges.childrenUpdates for this object or its parent as is
     // done in AXIsolatedTree::nodeChangeForObject and AXIsolatedTree::queueChange.
     //
     // Instead, just directly create and queue the node change so m_readerThreadNodeMap can hold a reference
     // to it. It will be removed from m_readerThreadNodeMap when the corresponding DOM element, renderer, or
     // other entity is removed from the page.
     NodeChange nodeChange { createIsolatedObjectData(axObject, *this), axObject->wrapper(), nextChangeSequence() };
-    Locker locker { m_changeLogLock };
-    mutablePendingChanges()->appends.append(WTF::move(nodeChange));
+    markDirtyAndGetWorkingChanges().appends.append(WTF::move(nodeChange));
 }
 
 void AXIsolatedTree::queueRemovals(Vector<NodeAndParentID>&& subtreeRemovals)
 {
     AX_ASSERT(isMainThread());
 
-    Locker locker { m_changeLogLock };
-    queueRemovalsLocked(WTF::move(subtreeRemovals));
-}
+    // Bail before touching the working buffer when there's nothing to queue.
+    if (subtreeRemovals.isEmpty() && m_protectedFromDeletionIDs.isEmpty())
+        return;
 
-void AXIsolatedTree::queueRemovalsLocked(Vector<NodeAndParentID>&& subtreeRemovals)
-{
-    AX_ASSERT(isMainThread());
-    AX_ASSERT(m_changeLogLock.isLocked());
-
-    auto pending = mutablePendingChanges();
-    pending->subtreeRemovals.appendVector(WTF::move(subtreeRemovals));
-    pending->protectedFromDeletionIDs.addAll(std::exchange(m_protectedFromDeletionIDs, { }));
+    auto& pending = markDirtyAndGetWorkingChanges();
+    pending.subtreeRemovals.appendVector(WTF::move(subtreeRemovals));
+    pending.protectedFromDeletionIDs.addAll(std::exchange(m_protectedFromDeletionIDs, { }));
 }
 
 void AXIsolatedTree::queueRemovalsAndUnresolvedChanges()
@@ -515,16 +526,15 @@ void AXIsolatedTree::queueAppendsAndRemovals(Vector<NodeChange>&& appends, Vecto
     AX_ASSERT(isMainThread());
 
     auto parentUpdateIDs = std::exchange(m_needsParentUpdate, { });
-    Locker locker { m_changeLogLock };
     for (auto&& append : appends)
         queueChange(WTF::move(append));
 
     for (const auto& axID : parentUpdateIDs) {
         ASSERT_WITH_MESSAGE(m_nodeMap.contains(axID), "An object marked as needing a parent update should've had an entry in the node map by now. ID was %s", axID.loggingString().utf8().data());
-        mutablePendingChanges()->parentUpdates.set(axID, *m_nodeMap.get(axID).parentID);
+        markDirtyAndGetWorkingChanges().parentUpdates.set(axID, *m_nodeMap.get(axID).parentID);
     }
 
-    queueRemovalsLocked(WTF::move(subtreeRemovals));
+    queueRemovals(WTF::move(subtreeRemovals));
 }
 
 void AXIsolatedTree::collectNodeChangesForSubtree(AccessibilityObject& axObject)
@@ -609,7 +619,6 @@ void AXIsolatedTree::updateNode(AccessibilityObject& axObject)
     // In both cases, we can't attach the wrapper immediately on the main thread, since the wrapper could be in use
     // on the AX thread (because this function updates an existing node).
     if (auto change = nodeChangeForObject(axObject)) {
-        Locker locker { m_changeLogLock };
         queueChange(WTF::move(*change));
         return;
     }
@@ -622,10 +631,8 @@ void AXIsolatedTree::updateNode(AccessibilityObject& axObject)
     if (!axParent)
         return;
 
-    if (auto change = nodeChangeForObject(downcast<AccessibilityObject>(*axParent))) {
-        Locker locker { m_changeLogLock };
+    if (auto change = nodeChangeForObject(downcast<AccessibilityObject>(*axParent)))
         queueChange(WTF::move(*change));
-    }
 }
 
 void AXIsolatedTree::objectChangedIgnoredState(const AccessibilityObject& object)
@@ -986,8 +993,7 @@ void AXIsolatedTree::updateNodeProperties(AccessibilityObject& axObject, const A
     if (properties.isEmpty())
         return;
 
-    Locker locker { m_changeLogLock };
-    mutablePendingChanges()->propertyChanges.append({ axObject.objectID(), WTF::move(properties) });
+    markDirtyAndGetWorkingChanges().propertyChanges.append({ axObject.objectID(), WTF::move(properties) });
 }
 
 void AXIsolatedTree::overrideNodeProperties(AXID axID, AXPropertyVector&& properties)
@@ -997,8 +1003,7 @@ void AXIsolatedTree::overrideNodeProperties(AXID axID, AXPropertyVector&& proper
     if (properties.isEmpty())
         return;
 
-    Locker locker { m_changeLogLock };
-    mutablePendingChanges()->propertyChanges.append({ axID, WTF::move(properties) });
+    markDirtyAndGetWorkingChanges().propertyChanges.append({ axID, WTF::move(properties) });
 }
 
 void AXIsolatedTree::updateDependentProperties(AccessibilityObject& axObject)
@@ -1239,8 +1244,14 @@ bool AXIsolatedTree::unsafeHasObjectForID(AXID axID) const
 
 std::optional<AXID> AXIsolatedTree::pendingRootNodeID()
 {
+    // The root the main thread has published or is about to publish, regardless of which buffer
+    // currently holds it. This should only be used on the main-thread.
+    AX_ASSERT(isMainThread());
+
+    if (m_workingChanges.rootNodeID)
+        return m_workingChanges.rootNodeID;
     Locker locker { m_changeLogLock };
-    return m_pendingChanges.rootNodeID;
+    return m_committedChanges.rootNodeID;
 }
 
 RefPtr<AXIsolatedObject> AXIsolatedTree::rootWebArea()
@@ -1256,17 +1267,10 @@ RefPtr<AXIsolatedObject> AXIsolatedTree::rootWebArea()
 
 void AXIsolatedTree::setPendingRootNodeID(AXID axID)
 {
-    Locker locker { m_changeLogLock };
-    setPendingRootNodeIDLocked(axID);
-}
-
-void AXIsolatedTree::setPendingRootNodeIDLocked(AXID axID)
-{
-    AXTRACE("AXIsolatedTree::setPendingRootNodeIDLocked"_s);
+    AXTRACE("AXIsolatedTree::setPendingRootNodeID"_s);
     AX_ASSERT(isMainThread());
-    AX_ASSERT(m_changeLogLock.isLocked());
 
-    mutablePendingChanges()->rootNodeID = axID;
+    markDirtyAndGetWorkingChanges().rootNodeID = axID;
 }
 
 void AXIsolatedTree::setFocusedNodeID(std::optional<AXID> axID)
@@ -1275,8 +1279,7 @@ void AXIsolatedTree::setFocusedNodeID(std::optional<AXID> axID)
     AXLOG(makeString("axID "_s, axID ? axID->loggingString() : ""_str));
     AX_ASSERT(isMainThread());
 
-    Locker locker { m_changeLogLock };
-    mutablePendingChanges()->focusedNodeID = axID;
+    markDirtyAndGetWorkingChanges().focusedNodeID = Markable<AXID> { axID };
 }
 
 void AXIsolatedTree::updateRelations(HashMap<AXID, AXRelations>&& relations)
@@ -1285,8 +1288,7 @@ void AXIsolatedTree::updateRelations(HashMap<AXID, AXRelations>&& relations)
     AX_ASSERT(isMainThread());
 
     m_relationsNeedUpdate = false;
-    Locker locker { m_changeLogLock };
-    mutablePendingChanges()->relations = WTF::move(relations);
+    markDirtyAndGetWorkingChanges().relations = WTF::move(relations);
 }
 
 void AXIsolatedTree::setSelectedTextMarkerRange(AXTextMarkerRange&& range)
@@ -1294,17 +1296,15 @@ void AXIsolatedTree::setSelectedTextMarkerRange(AXTextMarkerRange&& range)
     AXTRACE("AXIsolatedTree::setSelectedTextMarkerRange"_s);
     AX_ASSERT(isMainThread());
 
-    Locker locker { m_changeLogLock };
-    mutablePendingChanges()->selectedTextMarkerRange = WTF::move(range);
+    markDirtyAndGetWorkingChanges().selectedTextMarkerRange = WTF::move(range);
 }
 
 #if ENABLE(ACCESSIBILITY_LOCAL_FRAME)
 void AXIsolatedTree::setFrameGeometry(AXFrameGeometry&& geometry, IntPoint viewOriginScrollPosition)
 {
-    Locker locker { m_changeLogLock };
-    auto pending = mutablePendingChanges();
-    pending->frameGeometry = WTF::move(geometry);
-    pending->frameViewOriginScrollPosition = viewOriginScrollPosition;
+    auto& pending = markDirtyAndGetWorkingChanges();
+    pending.frameGeometry = WTF::move(geometry);
+    pending.frameViewOriginScrollPosition = viewOriginScrollPosition;
 }
 
 void AXIsolatedTree::updateFrameGeometryAndScrollPositionIfNeeded(AXObjectCache& cache)
@@ -1349,8 +1349,7 @@ void AXIsolatedTree::updateFrame(AXID axID, IntRect&& newFrame)
     properties.append({ AXProperty::RelativeFrame, WTF::move(newFrame) });
     // We can clear the initially-cached rough frame, since the object's frame has been cached.
     properties.append({ AXProperty::InitialLocalRect, FloatRect() });
-    Locker locker { m_changeLogLock };
-    mutablePendingChanges()->propertyChanges.append({ axID, WTF::move(properties) });
+    markDirtyAndGetWorkingChanges().propertyChanges.append({ axID, WTF::move(properties) });
 }
 
 void AXIsolatedTree::updateRootScreenRelativePosition()
@@ -1440,15 +1439,15 @@ void AXIsolatedTree::applyPendingChanges()
     if (!hasPendingChanges())
         return;
 
-    PendingChanges snapshot;
+    PendingChanges committedChanges;
     {
         Locker locker { m_changeLogLock };
-        // Release the lock before applying changes in the snapshot,
-        // since doing so can take time, and we don't want to block
-        // the main-thread unnecessarily.
-        snapshot = takePendingChangesLocked();
+        // Release the lock before applying committedChanges locally, since
+        // doing so can take time, and we don't want to block the main-thread
+        // unnecessarily.
+        committedChanges = takeCommittedChangesLocked();
     }
-    applyPendingChangesFromSnapshot(WTF::move(snapshot));
+    applyCommittedChanges(WTF::move(committedChanges));
 }
 
 void AXIsolatedTree::applyPendingChangesUnlessQueuedForDestruction()
@@ -1458,14 +1457,14 @@ void AXIsolatedTree::applyPendingChangesUnlessQueuedForDestruction()
     if (!hasPendingChanges())
         return;
 
-    PendingChanges snapshot;
+    PendingChanges committedChanges;
     {
         Locker locker { m_changeLogLock };
         if (m_queuedForDestruction) [[unlikely]]
             return;
-        snapshot = takePendingChangesLocked();
+        committedChanges = takeCommittedChangesLocked();
     }
-    applyPendingChangesFromSnapshot(WTF::move(snapshot));
+    applyCommittedChanges(WTF::move(committedChanges));
 }
 
 DidTearDown AXIsolatedTree::applyPendingChangesOrTearDown()
@@ -1475,7 +1474,7 @@ DidTearDown AXIsolatedTree::applyPendingChangesOrTearDown()
     if (!hasPendingChanges())
         return DidTearDown::No;
 
-    PendingChanges snapshot;
+    PendingChanges committedChanges;
     {
         Locker locker { m_changeLogLock };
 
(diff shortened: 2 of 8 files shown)


Source/WebCore/accessibility/AXObjectCache.cpp
```diff
@@ -5847,7 +5848,7 @@ void AXObjectCache::updateIsolatedTree(AccessibilityObject& axObject, AXProperty
 void AXObjectCache::startUpdateTreeSnapshotTimer()
 {
     if (!m_updateTreeSnapshotTimer.isActive())
-        m_updateTreeSnapshotTimer.startOneShot(updateTreeSnapshotTimerInterval);
+        m_updateTreeSnapshotTimer.startRepeating(updateTreeSnapshotTimerInterval);
```
AndresGonzalezApple: Or just startTreeUpdateTicker.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1202,8 +1196,12 @@ bool AXIsolatedTree::unsafeHasObjectForID(AXID axID) const
 
 std::optional<AXID> AXIsolatedTree::pendingRootNodeID()
 {
+    // On the main thread, an in-flight write may live in m_workingChanges until
```
minorninth: It looks like the only place pendingRootNodeID is called is from AXObjectCache::treeData, so the thread check doesn't make sense. As a side point I'd argue that it'd be more clear for it to show the current isolated tree, not the pending one.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1907,29 +1974,15 @@ void AXIsolatedTree::processQueuedNodeUpdates()
...
+        mutableWorkingChanges().mostRecentlyPaintedText = cache ? cache->mostRecentlyPaintedText() : HashMap<AXID, LineRange> { };
     }
 
     queueRemovalsAndUnresolvedChanges();
 
+    commitWorkingChanges();
+
 #if ENABLE(ACCESSIBILITY_THREAD_DISPATCHING)
-    if (hasPendingChanges() && m_appliedOrApplyingMainThreadSnapshot.exchange(false, std::memory_order_relaxed)) {
+    if (hasPendingChanges() && m_appliedOrApplyingCommittedChanges.exchange(false, std::memory_order_relaxed)) {
```
minorninth: I know memory_order_relaxed is existing code, but I'm still suspicious that it's correct here. My understanding is that memory_order_relaxed says that setting the flag itself is atomic, but that no ordering guarantees with respect to other variables are guaranteed. That means that it'd be possible for takeCommittedChanges to be called, but for another thread to observe inconsistent values of m_appliedOrApplyingCommittedChanges and m_committedChanges.

The type of thing I'd be worried about:

1. takeCommittedChanges could write to m_committedChanges and set m_appliedOrApplyingCommittedChanges to true

2. processQueuedNodeUpdates could observe m_appliedOrApplyingCommittedChanges true but m_committedChanges empty, it then sets m_appliedOrApplyingCommittedChanges to false

3. Subsequent calls to processQueuedNodeUpdates would observe m_appliedOrApplyingCommittedChanges as false so wouldn't check again

Source/WebCore/accessibility/AXObjectCache.cpp
```diff
@@ -400,7 +400,10 @@ bool AXObjectCache::shouldServeInitialCachedFrame()
     return !clientIsInTestMode() || forceInitialFrameCaching();
 }
 
-static constexpr Seconds updateTreeSnapshotTimerInterval { 100_ms };
+// Just over one frame at 60Hz (16.67ms), with the intent being
+// that the accessibility timer fires shortly after the last
+// rendering update.
+static constexpr Seconds updateTreeSnapshotTimerInterval { 17_ms };
```
minorninth: I also wonder about this running every 17ms if the page is idle

AndresGonzalezApple: 60Hz wakeup for every isolated tree's lifetime, including idle tabs. Was previously a 100ms on-demand one-shot. WebKit cares about wakeup cost, did you measure on a long-lived idle tab? Alternative: keep startOneShot and have setFocusedNodeID and the other direct writers explicitly schedule a tick when called.

AndresGonzalezApple: setFocusedNodeID used to publish focus immediately under the lock; now it can sit in m_workingChanges for up to one tick. Worth confirming no regression on rapidly-changing focus, e.g. typeahead.

Source/WebCore/accessibility/AXObjectCache.cpp
AndresGonzalezApple: Now that this is a long-lived repeating timer rather than a self-terminating one-shot, the function name is a misnomer. Consider ensureUpdateTreeSnapshotTimerRunning.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -251,12 +248,21 @@ void AXIsolatedTree::storeTree(AXObjectCache& cache, const Ref<AXIsolatedTree>&
...
     // Once we've added this new tree to the AXTreeStore, clients will be able to use
     // it off the main-thread. Make any final state mutations while we are the only thread
     // that can touch this tree.
     cache.setIsolatedTree(tree);
     AXTreeStore::set(tree->treeID(), tree.ptr());
     tree->m_replacingTree = nullptr;
+
+    // Start the repeating snapshot timer now that the tree is published. The
+    // timer drives processQueuedNodeUpdates which commits the working buffer.
+    cache.startUpdateTreeSnapshotTimer();
```
AndresGonzalezApple: In the buildIsolatedTree replacement path, where removeTreeForFrameID is not called first, the cache's timer is shared between old and new trees. startUpdateTreeSnapshotTimer is a no-op when active so this works, but if a future refactor interleaves create() with removeTreeForFrameID for the same frame you would get an old-tree stop() racing with a new-tree start(). Worth a comment documenting the cache/tree 1:1 lifetime assumption.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -298,6 +304,11 @@ void AXIsolatedTree::removeTreeForFrameID(FrameIdentifier frameID)
     Locker locker { s_storeLock };
     if (RefPtr tree = treeFrameCache().take(frameID)) {
         tree->m_geometryManager = nullptr;
+        // Stop the cache's snapshot timer so it doesn't keep firing on a tree
+        // that's about to be torn down. If a replacement tree is created, its
+        // storeTree call will start the timer again.
+        if (CheckedPtr cache = tree->axObjectCache())
+            cache->stopUpdateTreeSnapshotTimer();
```
AndresGonzalezApple: Possible NodeChange leak / self-reference cycle. Stopping the timer prevents commitWorkingChanges from running again, so any uncommitted entries in m_workingChanges.appends remain. IsolatedObjectData holds Ref<AXIsolatedTree> for this tree, so each leftover NodeChange pins the tree's refcount above zero. clearTreeContentsLocked clears m_committedChanges.appends but never m_workingChanges, so ~AXIsolatedTree never runs. Pre-PR this could not happen because all writes went directly to m_pendingChanges under lock and clearTreeContentsLocked cleared it. Reproduces if anything wrote to m_workingChanges.appends since the last commit, e.g., AXObjectCache::performDeferredCacheUpdate at AXObjectCache.cpp:5338 calls tree->addUnconnectedNode which writes working without going through processQueuedNodeUpdates. Suggested fix: call tree->commitWorkingChanges() on the main thread before queueForDestruction, while we still own the working buffer. That also lets clearTreeContentsLocked stay AX-thread-only.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1202,8 +1196,12 @@ bool AXIsolatedTree::unsafeHasObjectForID(AXID axID) const
 
 std::optional<AXID> AXIsolatedTree::pendingRootNodeID()
 {
+    // On the main thread, an in-flight write may live in m_workingChanges until
+    // commitWorkingChanges publishes it. AX-thread callers see only m_committedChanges.
+    if (isMainThread() && m_workingChanges.rootNodeID)
+        return m_workingChanges.rootNodeID;
     Locker locker { m_changeLogLock };
-    return m_pendingChanges.rootNodeID;
+    return m_committedChanges.rootNodeID;
```
AndresGonzalezApple, line 1249: Returns committed rootNodeID even on the main thread when working has none. Worth documenting that this unifies a logical 'current pending root, regardless of which buffer holds it'. The only caller is a debug dump (AXObjectCache.cpp:6168), so the lock cost is fine here, but consider an unsynchronized fast-path keyed off m_hasPendingChanges if this becomes hot.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1587,62 +1573,141 @@ void AXIsolatedTree::deleteSubtree(Ref<AXCoreObject>&& coreObjectToDelete, const
     }
 }
 
-AXIsolatedTree::PendingChanges AXIsolatedTree::takePendingChangesLocked()
+void AXIsolatedTree::mergeInto(PendingChanges& destination, PendingChanges&& source)
+{
```
AndresGonzalezApple, line 1634: destination.merge(WTF::move(source)) as a member would colocate the per-field semantics with the struct definition where readers are likely to look first.

AndresGonzalezApple, line 1634: The optional-move pattern (`if (source.foo) destination.foo = WTF::move(source.foo);`) repeats six times for selectedTextMarkerRange, sortedLiveRegionIDs, sortedNonRootWebAreaIDs, mostRecentlyPaintedText, relations, frameGeometry, frameViewOriginScrollPosition. A small helper mirroring mergeVector would shrink the block by ~10 lines:
```cpp
auto mergeOptional = [](auto& dst, auto&& src) {
    if (src) dst = WTF::move(src);
};
```

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1755,26 +1820,21 @@ void AXIsolatedTree::applyPendingChangesFromSnapshot(PendingChanges&& snapshot)
...
 void AXIsolatedTree::sortedNonRootWebAreasDidChange(Vector<AXID> webAreaIDs)
 {
     AX_ASSERT(isMainThread());
 
-    Locker locker { m_changeLogLock };
-    // FIXME: m_pendingChanges.sortedLiveRegionIDs and m_pendingChanges.sortedNonRootWebAreaIDs should be synced in AXIsolatedTree::processQueuedNodeUpdates(),
-    // not ad-hoc whenever the main-thread changes them. Otherwise we could sync IDs to the accessibility thread that don't have isolated objects
-    // until the next actual tree-update cycle.
-    mutablePendingChanges()->sortedNonRootWebAreaIDs = WTF::move(webAreaIDs);
+    mutableWorkingChanges().sortedNonRootWebAreaIDs = WTF::move(webAreaIDs);
```
AndresGonzalezApple: The FIXME got dropped but the issue is unchanged by this PR. These IDs still get pushed to the AX thread without alignment to processQueuedNodeUpdates, so the AX thread can still observe IDs that don't have corresponding isolated objects yet. Two-buffer atomicity addresses focus/append divergence (the motivating example), but not this one. Either keep the FIXME or fix it now.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1854,17 +1913,25 @@ void AXIsolatedTree::queueNodeRemoval(const AccessibilityObject& axObject)
...
+        return;
+
     WeakPtr cache = axObjectCache();
-    if (!cache)
+    if (!cache) {
+        // Leave m_hasQueuedNodeUpdates set so a later tick (with a non-null
+        // cache) processes the queued work instead of stranding it.
         return;
+    }
+    m_hasQueuedNodeUpdates = false;
```
AndresGonzalezApple, line 2038: Subtle issue with the reset point. The m_mostRecentlyPaintedTextIsDirty branch below calls mutableWorkingChanges() (writing mostRecentlyPaintedText), and queueRemovalsAndUnresolvedChanges goes through queueChange / queueRemovals; both set m_hasQueuedNodeUpdates back to true. The next 17ms tick re-enters, sees the flag set, calls commitWorkingChanges() which no-ops because m_workingChangesDirty is false. One wasted tick per real cycle. Two cleanups: (a) reset the flag after the work below, or have the in-function writes skip flag-setting via a 'currentlyProcessing' guard; (b) merge m_hasQueuedNodeUpdates with m_workingChangesDirty since every working-buffer write currently sets both, and the mark*Dirty paths could also just set m_workingChangesDirty.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
```diff
@@ -515,22 +515,26 @@ class AXIsolatedTree : public ThreadSafeRefCountedAndCanMakeThreadSafeWeakPtr<AX
         objectChangedIgnoredState(object);
     }
 
-    // Both setPendingRootNodeLocked and setFocusedNodeID are called during the generation
-    // of the IsolatedTree.
-    // Focused node updates in AXObjectCache use setFocusNodeID.
     void setPendingRootNodeID(AXID);
-    void NODELETE setPendingRootNodeIDLocked(AXID) WTF_REQUIRES_LOCK(m_changeLogLock);
     void setFocusedNodeID(std::optional<AXID>);
```
AndresGonzalezApple, line 532: Replacement one-liner would help orient readers: 'Main-thread setters that route writes through m_workingChanges; published to the AX thread on the next commitWorkingChanges()'.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
```diff
@@ -515,22 +515,26 @@ class AXIsolatedTree : public ThreadSafeRefCountedAndCanMakeThreadSafeWeakPtr<AX
...
     void updateRelations(HashMap<AXID, AXRelations>&&);
 
     AXCoreObject::AccessibilityChildrenVector sortedLiveRegions();
     AXCoreObject::AccessibilityChildrenVector sortedNonRootWebAreas();
 
-    void markMostRecentlyPaintedTextDirty() { m_mostRecentlyPaintedTextIsDirty = true; }
+    void markMostRecentlyPaintedTextDirty()
+    {
+        m_mostRecentlyPaintedTextIsDirty = true;
+        m_hasQueuedNodeUpdates = true;
```
AndresGonzalezApple: Both methods now silently assume main-thread invocation (m_hasQueuedNodeUpdates is non-atomic and only safe on the main thread). Add AX_ASSERT(isMainThread()) to enforce.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
```diff
@@ -637,6 +641,14 @@ class AXIsolatedTree : public ThreadSafeRefCountedAndCanMakeThreadSafeWeakPtr<AX
 
     struct PendingChanges {
         Markable<AXID> focusedNodeID;
+        // True iff setFocusedNodeID was called this cycle. Distinguishes
+        // "not written" (preserve previous focus) from "explicitly cleared"
+        // (propagate the null Markable).
+        //
+        // Always update both fields via setFocusedNodeID(). Never assign
+        // focusedNodeID directly, or the flag will go stale.
+        bool didExplicitlySetFocusedNodeID { false };
```
AndresGonzalezApple: Compiler can't help enforce this invariant since both fields are public. Two ways to harden: (a) make focusedNodeID and didExplicitlySetFocusedNodeID private and expose only setFocusedNodeID plus a const getter; (b) replace both with std::optional<Markable<AXID>> (outer optional = 'set this cycle?', inner Markable = the value, possibly empty for an explicit clear). (b) is the simpler representation.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
```diff
@@ -653,43 +665,38 @@ class AXIsolatedTree : public ThreadSafeRefCountedAndCanMakeThreadSafeWeakPtr<AX
...
+    PendingChanges& mutableWorkingChanges()
     {
-        return { m_pendingChanges, m_hasPendingChanges };
+        AX_ASSERT(isMainThread());
+        m_workingChangesDirty = true;
+        // Any change to m_workingChanges is going to need processing
+        // in processQueuedNodeUpdates(), so set this flag too.
+        m_hasQueuedNodeUpdates = true;
+        return m_workingChanges;
     }
```
AndresGonzalezApple, line 701: Lost a useful safeguard from the deleted PendingChangesAccessor: the old design ASSERT'd in its destructor that the accessor was actually used (AX_ASSERT(m_wasAccessed)), catching false-positive dirty flags. The new accessor unconditionally sets two flags on entry, so any caller that takes a reference and never writes inflates both. If you keep this design, consider naming it markDirtyAndGetWorkingChanges() so the side-effect is in the name.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
```diff
@@ -752,9 +758,20 @@ class AXIsolatedTree : public ThreadSafeRefCountedAndCanMakeThreadSafeWeakPtr<AX
...
+    // Main-thread-only buffer that accumulates writes between commit points.
+    // commitWorkingChanges() merges it into m_committedChanges under lock.
+    PendingChanges m_workingChanges;
+    bool m_workingChangesDirty { false };
+
+    // Set to true whenever any work is queued (working buffer write, deferred
+    // node-update queues, relation/painted-text dirty marks). Checked at the
+    // top of processQueuedNodeUpdates so the periodic snapshot timer can early
+    // out cheaply when there is nothing to do.
+    bool m_hasQueuedNodeUpdates { false };
```
AndresGonzalezApple, line 797: Plain bool, single-thread-only (main). Worth a comment naming the thread invariant. m_workingChangesDirty above and this flag are largely redundant; every working-buffer write sets both, and mark*Dirty paths set both via the helpers. The two only diverge for 'deferred queue work that doesn't write to working', which I can't find in the diff: m_needsUpdateNode etc. all eventually go through queueChange / queueRemovals which write working. If you confirm there's no real divergence, drop m_workingChangesDirty and have commitWorkingChanges check m_hasQueuedNodeUpdates directly.

Source/WebCore/accessibility/AXObjectCache.cpp
AndresGonzalezApple: Andres Gonzalez <andresg_22@apple.com>: Or hold down the tab key through a page with multiple focusable elements.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -251,12 +248,21 @@ void AXIsolatedTree::storeTree(AXObjectCache& cache, const Ref<AXIsolatedTree>&
 {
     AX_ASSERT(isMainThread());
 
+    // Atomically publish any working-buffer writes into m_committedChanges
+    // before the tree becomes externally visible via AXTreeStore::set.
+    tree->commitWorkingChanges();
```
AndresGonzalezApple: Anything that writes to working between this commit and treeFrameCache().set below won't be visible to AX-thread readers until the next periodic commit. The three intervening lines look benign but it's a footgun. Consider committing immediately before treeFrameCache().set, since that's the actual point at which the AX thread can find the tree via the frame map.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -400,34 +410,24 @@ void AXIsolatedTree::addUnconnectedNode(Ref<AccessibilityObject> axObject)
     AXLOG(makeString("AXIsolatedTree::addUnconnectedNode creating isolated object from live object ID "_s, objectID.loggingString()));
 
     // Because we are queuing a change for an object not intended to be connected to the rest of the tree,
-    // we don't need to update m_nodeMap or m_pendingChanges.childrenUpdates for this object or its parent as is
+    // we don't need to update m_nodeMap or m_committedChanges.childrenUpdates for this object or its parent as is
```
AndresGonzalezApple: queueChange and nodeChangeForObject write to m_workingChanges.childrenUpdates, not m_committedChanges. Should say m_workingChanges.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1587,62 +1573,141 @@ void AXIsolatedTree::deleteSubtree(Ref<AXCoreObject>&& coreObjectToDelete, const
...
+        destination.parentUpdates = WTF::move(source.parentUpdates);
+    else {
+        for (auto& parentUpdate : source.parentUpdates)
+            destination.parentUpdates.set(parentUpdate.key, parentUpdate.value);
+    }
+
+    // Optional last-write-wins: only overwrite if source wrote a value this cycle.
+    // Otherwise, destination's existing value (which may be persistent state
+    // preserved across cycles, e.g., focusedNodeID) is left untouched.
+    if (source.didExplicitlySetFocusedNodeID) {
```
AndresGonzalezApple: Once didExplicitlySetFocusedNodeID flips true (the first explicit setFocusedNodeID call) it never goes back to false; every take preserves it, every subsequent merge with source.didExplicit=true keeps it true. So in steady state this flag is effectively everSetFocusedNodeID. The current name reads as per-cycle and will mislead. Either rename or expand the comment to explain the steady-state behavior.

AndresGonzalezApple: A simpler representation: std::optional<Markable<AXID>> focusedNodeID. Outer nullopt means 'not set this cycle', inner empty Markable means 'explicitly cleared'. Same expressiveness, one field, no invariant to maintain.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1587,62 +1573,141 @@ void AXIsolatedTree::deleteSubtree(Ref<AXCoreObject>&& coreObjectToDelete, const
...
+    if (source.frameViewOriginScrollPosition)
+        destination.frameViewOriginScrollPosition = WTF::move(source.frameViewOriginScrollPosition);
+#endif
+}
+
+void AXIsolatedTree::commitWorkingChanges()
+{
+    AX_ASSERT(isMainThread());
+
+    if (!m_workingChangesDirty)
```
AndresGonzalezApple: m_workingChangesDirty is set unconditionally in mutableWorkingChanges() even if the caller doesn't end up writing (same conservative behavior the deleted PendingChangesAccessor had via its destructor). A no-op merge can still pay the lock cost. Probably not worth optimizing unless commitWorkingChanges shows up hot.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1587,62 +1573,141 @@ void AXIsolatedTree::deleteSubtree(Ref<AXCoreObject>&& coreObjectToDelete, const
...
+        return;
+
+    {
+        Locker locker { m_changeLogLock };
+        mergeInto(m_committedChanges, std::exchange(m_workingChanges, { }));
+        m_hasPendingChanges.store(true);
```
AndresGonzalezApple: m_hasPendingChanges is std::atomic, so the .store(true) does not need the lock for memory ordering. The lock is necessary for the merge itself, but the .store(true) could move outside without changing correctness.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1587,62 +1573,141 @@ void AXIsolatedTree::deleteSubtree(Ref<AXCoreObject>&& coreObjectToDelete, const
...
+    m_appliedOrApplyingCommittedChanges.store(true, std::memory_order_relaxed);
 #endif
-    return snapshot;
+    return committedChanges;
 }
 
-void AXIsolatedTree::applyPendingChangesFromSnapshot(PendingChanges&& snapshot)
+void AXIsolatedTree::applyCommittedChanges(PendingChanges&& committedChanges)
 {
     AXTRACE("AXIsolatedTree::applyPendingChanges"_s);
```
AndresGonzalezApple: Stale trace string. The function is now applyCommittedChanges; the signpost name was renamed below, but the AXTRACE was not. Same pattern in setPendingRootNodeID's AXTRACE earlier.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
minorninth, line 1646: One scenario that's come up before is an append followed by a delete, or maybe even the other way around. I'm worried that naively merging isn't going to do the right thing.

Can you add a test that appends and removes before the merge happens?

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: We now use the strictest memory ordering for all of these operations.

Source/WebCore/accessibility/AXObjectCache.cpp
twilco: Resolved by making this a one-shot timer again.

Source/WebCore/accessibility/AXObjectCache.cpp
twilco: Resolved by making this a one-shot timer again.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: Good catch! Added your suggested commitWorkingChanges().

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: Readded the FIXME!

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
twilco: Took your suggested comment

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
twilco: Took (b)

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
twilco: Comment added!

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
minorninth, line 792: I wonder if we should start adding prefixes to member variables that are only allowed on one thread, would make auditing easier. Possibly for a follow-up to apply to all of them.

Source/WebCore/accessibility/AXObjectCache.cpp
minorninth: I still see this interval - is it actually an interval still or just a duration?

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
twilco: There is real divergence between the flags, so I'm not sure that it makes sense to collapse them. Specifically, queueNodeUpdate / queueNodeRemoval / the mark*Dirty helpers set m_hasQueuedNodeUpdates without touching the working buffer. That write only happens in the next cycle, so between those points, the flags hold different values.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: Fixed!

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: Fixed!

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: Good point! Fixed.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: Fixed!

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
twilco: Added!

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: m_workingChanges can only be accessed by the main-thread, so either an explicit check, or at least an assert feels like it makes sense.

The pending-root is shown only if there is no valid root, so we do prefer the current root as you suggest.

Source/WebCore/accessibility/AXObjectCache.cpp
twilco: Sorry, that's confusing indeed. Changed name to duration.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: The three intervening lines are part of the "publishing" process, we need to commit before any of them (not after) or else those places prior to treeFrameCache().set could observe an empty tree. I added an `AX_ASSERT(!tree->m_workingChangesDirty);` to make sure none of them dirty the working-changes.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: Fixed with your std::optional<Markable<AXID>> suggestion.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: We want to perform all writes to this flag under lock so that it's state is exactly paired with the take and release of the working changes. Otherwise the AX thread could race, see pending changes false, despite the lock critical section having queued changes.

I added a comment to `AXIsolatedTree::commitWorkingChangesLocked()` that clarifies this.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: Took your (a) suggestion with a `m_isProcessingQueuedNodeUpdates` guard!

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: I think this is a non-issue now that we're back to a one-shot instead of repeating timer. Specifically, storeTree no longer starts a timer, and removeTreeForFrameID no longer stops one, so there is no start/stop pair left to interleave.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: Added a new test: `accessibility/append-and-remove-in-one-cycle.html`. This moves a child to another DOM element in one atomic operation. It also creates a new node, appends it, and then immediately removes it.

Let me know if this covers what you were looking for!

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
twilco: Good idea! I filed this:

https://bugs.webkit.org/show_bug.cgi?id=320999

Source/WebCore/accessibility/AXObjectCache.cpp
twilco: Good callout. I created a testpage with 20 focusables, and found no issues after rapidly tabbing back and forth and announcing the currently focused element.

AndresGonzalezApple:
The commit message is still referring to the 60 hz timer.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -110,8 +110,15 @@ void AXIsolatedTree::queueForDestruction()
...
     Locker locker { m_changeLogLock };
-    m_queuedForDestruction = true;
+    // Publish anything still in the working buffer before the AX thread tears this tree down.
+    // Teardown (clearTreeContentsLocked) only clears m_committedChanges, and every NodeChange
+    // left behind holds a Ref to this tree (IsolatedObjectData::tree), so stranding them would
+    // be a self-reference cycle that keeps the tree alive forever.
+    commitWorkingChangesLocked();
+    // m_queuedForDestruction is a pending-change guarded by m_hasPendingChanges,
+    // so make sure to also set m_hasPendingChanges.
     m_hasPendingChanges.store(true);
```
AndresGonzalezApple, line 122: This m_hasPendingChanges.store(true) is redundant: commitWorkingChangesLocked() on the line just above already stores true unconditionally. The accompanying comment implies the store is needed here, which is misleading since the commit guarantees it. Consider dropping the store (and comment) or, if kept for documentation, note that it is already set.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
```diff
@@ -1802,26 +1918,25 @@ void AXIsolatedTree::applyPendingChangesFromSnapshot(PendingChanges&& snapshot)
...
 void AXIsolatedTree::sortedNonRootWebAreasDidChange(Vector<AXID> webAreaIDs)
 {
     AX_ASSERT(isMainThread());
 
-    Locker locker { m_changeLogLock };
-    // FIXME: m_pendingChanges.sortedLiveRegionIDs and m_pendingChanges.sortedNonRootWebAreaIDs should be synced in AXIsolatedTree::processQueuedNodeUpdates(),
-    // not ad-hoc whenever the main-thread changes them. Otherwise we could sync IDs to the accessibility thread that don't have isolated objects
-    // until the next actual tree-update cycle.
-    mutablePendingChanges()->sortedNonRootWebAreaIDs = WTF::move(webAreaIDs);
+    // FIXME: m_pendingChanges.sortedLiveRegionIDs and m_pendingChanges.sortedNonRootWebAreaIDs
```
AndresGonzalezApple: This FIXME still refers to the removed m_pendingChanges member, renamed to m_workingChanges / m_committedChanges by this PR.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
```diff
@@ -647,7 +655,10 @@ class AXIsolatedTree : public ThreadSafeRefCountedAndCanMakeThreadSafeWeakPtr<AX
     };
 
     struct PendingChanges {
-        Markable<AXID> focusedNodeID;
+        // std::nullopt means "focus wasn't written this cycle", so the destination's existing
```
AndresGonzalezApple, line 658: Should rootNodeID stay a plain Markable<AXID> or also become an optional? I think it is fine because setPendingRootNodeID never writes an empty root; but a stray call clearing the root would be indistinguishable from 'not written' and silently dropped by merge(). At least worth a one-line comment here that the root is intentionally never cleared.

AndresGonzalezApple (approved):

AndresGonzalezApple (approved):

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.h
twilco: Added a comment clarifying root is not cleared, and what needs to happen if we do desire to clear it in the future.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: Good catch! Fixed.

Source/WebCore/accessibility/isolatedtree/AXIsolatedTree.cpp
twilco: This redundancy is intentional so that `queueForDestruction()` doesn't implicitly lean on behavior of another function. Added a comment.

minorninth:
Claude helped come up with one scenario that might break, what do you think? Before this change it would have been a race. To reproduce this, you'd need to remove and then insert with the same parent, which the test you added didn't cover.

```
  // X exists under P in the isolated tree
  P.removeChild(X);
  axP.childrenCount;   // forces updateChildren(P) → queues {X, P} into subtreeRemovals
  P.appendChild(X);
  axP.childrenCount;   // forces updateChildren(P) → X is "new" (removeSubtreeFromNodeMap
                       //   dropped it from m_nodeMap at :1111) → append with parentID == P
```
Both now land in one commit. In applyCommittedChanges: the removal loop (1796) deletes X from m_readerThreadNodeMap, then removeStaleAppends (1520) sees removedParent == appendedParent == P, concludes isReparent == false, and drops the append (1566-1567). Net result: X is gone from the isolated tree while the live tree still has it under P, and P's childrenUpdates still lists X — so X becomes an unresolved child ID.

twilco:
> Claude helped come up with one scenario that might break, what do you think? Before this change it would have been a race. To reproduce this, you'd need to remove and then insert with the same parent, which the test you added didn't cover.
> 
> ```
>   // X exists under P in the isolated tree
>   P.removeChild(X);
>   axP.childrenCount;   // forces updateChildren(P) → queues {X, P} into subtreeRemovals
>   P.appendChild(X);
>   axP.childrenCount;   // forces updateChildren(P) → X is "new" (removeSubtreeFromNodeMap
>                        //   dropped it from m_nodeMap at :1111) → append with parentID == P
> ```
> 
> Both now land in one commit. In applyCommittedChanges: the removal loop (1796) deletes X from m_readerThreadNodeMap, then removeStaleAppends (1520) sees removedParent == appendedParent == P, concludes isReparent == false, and drops the append (1566-1567). Net result: X is gone from the isolated tree while the live tree still has it under P, and P's childrenUpdates still lists X — so X becomes an unresolved child ID.

Good catch! This is fixed by: https://github.com/WebKit/WebKit/pull/71214

PR #66440, opened 2026-06-04
[GTK][WPE] Skia Compositor: implement batched painting

carlosgcampos:
#### 14f3f90b25369f7cb674514274346ff440a20865
<pre>
[GTK][WPE] Skia Compositor: implement batched painting
<a href="https://bugs.webkit.org/show_bug.cgi?id=316294">https://bugs.webkit.org/show_bug.cgi?id=316294</a>

Using the experimental SkCanvas API to draw a set of images with a
single call, we can group draw operations in batches before passing them
to skia. This is more efficient when we have many layers painting
content that can be merged.

* Source/WebCore/platform/Skia.cmake:
* Source/WebCore/platform/SourcesSkia.txt:
* Source/WebCore/platform/graphics/skia/SkiaBackingStore.cpp:
(WebCore::SkiaBackingStore::buildImageSet const):
(WebCore::SkiaBackingStore::Tile::image const):
(WebCore::SkiaBackingStore::Tile::image): Deleted.
* Source/WebCore/platform/graphics/skia/SkiaBackingStore.h:
(WebCore::SkiaCompositingLayer::paint):
(WebCore::SkiaCompositingLayer::paintSelf):
(WebCore::SkiaCompositingLayer::paintContents):
(WebCore::SkiaCompositingLayer::collectFrameDamage):
(WebCore::SkiaCompositingLayer::paintDebugIndicators):
(WebCore::SkiaCompositingLayer::paintSelfAndChildren):
(WebCore::SkiaCompositingLayer::paintWithIntermediateSurface):
(WebCore::SkiaCompositingLayer::paintBackdrop):
(WebCore::SkiaCompositingLayer::paintWithMaskAndBackdrop):
(WebCore::SkiaCompositingLayer::paintWithOpacity):
(WebCore::SkiaCompositingLayer::paintWithBlendMode):
(WebCore::SkiaCompositingLayer::paintWith3DRenderingContext):
* Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h:
* Source/WebCore/platform/graphics/skia/SkiaCompositingLayerImageSetBatch.cpp: Added.
(WebCore::SkiaCompositingLayerImageSetBatch::updatePaintProperties):
(WebCore::SkiaCompositingLayerImageSetBatch::updateSamplingOptions):
(WebCore::SkiaCompositingLayerImageSetBatch::addImageSet):
(WebCore::SkiaCompositingLayerImageSetBatch::addImage):
(WebCore::SkiaCompositingLayerImageSetBatch::flushIfNeeded):
(WebCore::SkiaCompositingLayerImageSetBatch::ScopedFlush::ScopedFlush):
(WebCore::SkiaCompositingLayerImageSetBatch::ScopedFlush::~ScopedFlush):
* Source/WebCore/platform/graphics/skia/SkiaCompositingLayerImageSetBatch.h: Added.

Canonical link: <a href="https://commits.webkit.org/314626@main">https://commits.webkit.org/314626@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/6ae0827fc2e824ce911f383afacd9d3e5300c6c5

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/166673 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/40163 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/33744 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/175721 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/123930 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/168539 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/40596 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/40171 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/129590 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/91262 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/169632 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/31983 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/149760 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/110115 "Passed tests") | 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/30960 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/29662 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/23862 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/140931 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/27457 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/178255 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/31155 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/29164 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/137951 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/40233 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/33809 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/137868 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/40921 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/149255 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/103692 "Built successfully") | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/25366 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/33154 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/26120 "Passed tests") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/39432 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/112506 "Built successfully") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/38828 "Built successfully") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/4210 "") | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/39073 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/38971 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/platform/graphics/skia/SkiaBackingStore.cpp
+++ b/Source/WebCore/platform/graphics/skia/SkiaBackingStore.cpp
@@ -90,6 +90,26 @@ void SkiaBackingStore::paintToCanvas(SkCanvas& canvas, const SkPaint& paint)
     }
 }
 
+Vector<SkCanvas::ImageSetEntry> SkiaBackingStore::buildImageSet(size_t matrixIndex, float opacity, bool enableAntialias) const
+{
+    if (m_tiles.isEmpty())
+        return { };
+
+    FloatRect layerRect = { { }, m_size };
+
+    Vector<SkCanvas::ImageSetEntry> images;
+    for (auto& tile : m_tiles.values()) {
+        const auto& image = tile.image();
+        if (!image)
+            continue;
+
+        // FIXME: implement per edge antialiasing.
+        unsigned aaFlags = enableAntialias && allTileEdgesExposed(layerRect, tile.rect()) ? SkCanvas::kAll_QuadAAFlags : SkCanvas::kNone_QuadAAFlags;
+        images.append(SkCanvas::ImageSetEntry(image, SkRect::MakeWH(image->width(), image->height()), SkRect(tile.rect()), matrixIndex, opacity, aaFlags, false));
+    }
+    return images;
+}
+
 void SkiaBackingStore::drawDebugBorders(SkCanvas& canvas, const SkPaint& paint)
 {
     for (const auto& tile : m_tiles.values())
@@ -160,7 +180,7 @@ void SkiaBackingStore::Tile::update(const IntRect& dirtyRect, const IntRect& til
     WTFEndSignpost(this, SkiaBackingStoreTileUpdate);
 }
 
-sk_sp<SkImage> SkiaBackingStore::Tile::image()
+sk_sp<SkImage> SkiaBackingStore::Tile::image() const
 {
     if (!m_cachedImage && m_texture) {
         auto* grContext = PlatformDisplay::sharedDisplay().skiaGrContext();
--- a/Source/WebCore/platform/graphics/skia/SkiaBackingStore.h
+++ b/Source/WebCore/platform/graphics/skia/SkiaBackingStore.h
@@ -29,6 +29,7 @@
 #include "CoordinatedBackingStoreProxy.h"
 #include "FloatRect.h"
 WTF_IGNORE_WARNINGS_IN_THIRD_PARTY_CODE_BEGIN
+#include <skia/core/SkCanvas.h>
 #include <skia/core/SkSurface.h>
 WTF_IGNORE_WARNINGS_IN_THIRD_PARTY_CODE_END
 #include <wtf/HashMap.h>
@@ -47,6 +48,7 @@ public:
 
     void update(const FloatSize&, float scale, CoordinatedBackingStoreProxy::Update&&);
     void paintToCanvas(SkCanvas&, const SkPaint&);
+    Vector<SkCanvas::ImageSetEntry> buildImageSet(size_t matrixIndex, float opacity, bool enableAntialias) const;
     void drawDebugBorders(SkCanvas&, const SkPaint&);
 
 private:
@@ -62,7 +64,7 @@ private:
 
         void update(const IntRect& dirtyRect, const IntRect& tileRect, CoordinatedTileBuffer&);
         const FloatRect& rect() const LIFETIME_BOUND { return m_rect; }
-        sk_sp<SkImage> image();
+        sk_sp<SkImage> image() const;
 
     private:
         void ensureTexture(const IntSize&, CoordinatedTileBuffer&);
@@ -70,7 +72,7 @@ private:
         float m_scale { 1. };
         FloatRect m_rect;
         RefPtr<BitmapTexture> m_texture;
-        sk_sp<SkImage> m_cachedImage;
+        mutable sk_sp<SkImage> m_cachedImage;
     };
 
     HashMap<uint32_t, Tile> m_tiles;
--- a/Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
+++ b/Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
@@ -427,6 +427,7 @@ bool SkiaCompositingLayer::paint(SkCanvas& canvas, std::optional<Damage>& damage
 
     context.mode = PaintMode::Paint;
     recursivePaint(canvas, context);
+    context.imageSetBatch.flushIfNeeded(canvas);
 
     recursiveCleanUpAfterPaint();
 
@@ -464,45 +465,70 @@ void SkiaCompositingLayer::paintSelf(SkCanvas& canvas, PaintContext& context)
     if (m_size.isEmpty() || !m_visible || !m_contentsVisible || !hasVisualContent())
         return;
 
-    TransformationMatrix transform(context.accumulatedReplicaTransform);
-    transform.multiply(m_transforms.combined);
-
-    canvas.save();
-    canvas.concat(SkM44(transform));
-
     if (context.mode == PaintMode::Paint) {
         paintContents(canvas, context);
 #if ENABLE(DAMAGE_TRACKING)
-        collectFrameDamage(canvas, context, transform);
+        collectFrameDamage(canvas, context);
 #endif
     }
-
-    canvas.restore();
 }
 
 void SkiaCompositingLayer::paintContents(SkCanvas& canvas, PaintContext& context)
 {
-    SkPaint paint;
-    paint.setStyle(SkPaint::kFill_Style);
-    auto ctm = canvas.getLocalToDeviceAs3x3();
-    paint.setAntiAlias(!ctm.preservesAxisAlignment() && !ctm.preservesRightAngles());
-    paint.setAlphaf(context.opacity);
-    if (context.blendMode)
-        paint.setBlendMode(*context.blendMode);
-    if (context.colorFilter)
-        paint.setColorFilter(context.colorFilter);
+    TransformationMatrix transform(context.accumulatedReplicaTransform);
+    transform.multiply(m_transforms.combined);
+
+    auto ctm = SkM44(transform).asM33();
+    bool enableAntialias = !ctm.preservesAxisAlignment() && !ctm.preservesRightAngles();
+
+    context.imageSetBatch.updatePaintProperties(canvas, context.colorFilter, context.blendMode);
+
+    auto setupPaint = [&] -> SkPaint {
+        SkPaint paint;
+        paint.setStyle(SkPaint::kFill_Style);
+        paint.setAntiAlias(enableAntialias);
+        paint.setAlphaf(context.opacity);
+        if (context.blendMode)
+            paint.setBlendMode(*context.blendMode);
+        if (context.colorFilter)
+            paint.setColorFilter(context.colorFilter);
+        return paint;
+    };
 
     if (m_backingStore)
-        m_backingStore->paintToCanvas(canvas, paint);
+        context.imageSetBatch.addImageSet(canvas, *m_backingStore, ctm, context.opacity, enableAntialias);
 
     if (m_contentsSolidColor.isValid() && m_contentsSolidColor.isVisible()) {
+        ScopedFlush autoFlush(canvas, context.imageSetBatch, ScopedFlush::Mode::FlushBefore);
+        canvas.concat(SkM44(transform));
+        SkPaint paint = setupPaint();
         paint.setColor(SkColor(m_contentsSolidColor.colorWithAlphaMultipliedBy(context.opacity)));
         canvas.drawRect(m_contentsRect, paint);
     } else if (m_contentsBuffer || m_imageBackingStore) {
-        bool shouldClipContents = m_contentsClippingRect.isRounded() || !m_contentsClippingRect.rect().contains(m_contentsRect);
-        SkAutoCanvasRestore autoRestore(&canvas, shouldClipContents);
-        if (shouldClipContents)
-            clipRect(canvas, m_contentsClippingRect);
+        bool shouldPaintNow = [&] {
+            if (m_contentsClippingRect.isRounded())
+                return true;
+
+            if (!m_contentsBuffer && !m_contentsTiling.size.isEmpty())
+                return true;
+
+            if (is<CoordinatedPlatformLayerBufferHolePunch>(m_contentsBuffer))
+                return true;
+
+            // FIXME: clip is not correctly applied with batched painting.
+            if (!m_contentsClippingRect.rect().contains(m_contentsRect))
+                return true;
+
+            return false;
+        }();
+
+        ScopedFlush autoFlush(canvas, context.imageSetBatch, shouldPaintNow ? ScopedFlush::Mode::FlushBefore : ScopedFlush::Mode::DoNothing);
+        if (shouldPaintNow) {
+            canvas.concat(SkM44(transform));
+
+            if (m_contentsClippingRect.isRounded() || !m_contentsClippingRect.rect().contains(m_contentsRect))
+                clipRect(canvas, m_contentsClippingRect);
+        }
 
         sk_sp<SkImage> image;
 
@@ -512,6 +538,7 @@ void SkiaCompositingLayer::paintContents(SkCanvas& canvas, PaintContext& context
                 TransformationMatrix matrix = canvas.getLocalToDevice();
                 downcast<CoordinatedPlatformLayerBufferHolePunch>(*m_contentsBuffer).setHolePunchVideoRectangle(enclosingIntRect(matrix.mapRect(m_contentsRect)));
 #endif
+                SkPaint paint = setupPaint();
                 paint.setBlendMode(SkBlendMode::kClear);
                 canvas.drawRect(SkRect(m_contentsRect), paint);
             } else
@@ -523,24 +550,35 @@ void SkiaCompositingLayer::paintContents(SkCanvas& canvas, PaintContext& context
                 SkMatrix matrix;
                 matrix.setScale(m_contentsTiling.size.width() / tileImage->width(), m_contentsTiling.size.height() / tileImage->height());
                 matrix.postTranslate(-m_contentsTiling.phase.width(), -m_contentsTiling.phase.height());
+                SkPaint paint = setupPaint();
                 paint.setShader(tileImage->makeShader(SkTileMode::kRepeat, SkTileMode::kRepeat, SkSamplingOptions(SkFilterMode::kLinear, SkMipmapMode::kNone), matrix));
                 canvas.drawRect(m_contentsRect, paint);
             }
         }
 
         if (image) {
-            canvas.drawImageRect(image, SkRect::MakeSize(SkSize::Make(image->dimensions())), SkRect(m_contentsRect),
-                SkSamplingOptions(SkFilterMode::kLinear, SkMipmapMode::kNone), &paint, SkCanvas::kFast_SrcRectConstraint);
+            if (shouldPaintNow) {
+                SkPaint paint = setupPaint();
+                canvas.drawImageRect(image, SkRect::MakeSize(SkSize::Make(image->dimensions())), SkRect(m_contentsRect),
+                    SkSamplingOptions(SkFilterMode::kLinear, SkMipmapMode::kNone), &paint, SkCanvas::kFast_SrcRectConstraint);
+            } else {
+                auto clippingRect = m_contentsClippingRect.rect();
+                if (clippingRect.contains(m_contentsRect))
+                    clippingRect = { };
+                context.imageSetBatch.addImage(canvas, image, m_contentsRect, clippingRect, ctm, context.opacity, enableAntialias);
+            }
         }
     }
 }
 
 #if ENABLE(DAMAGE_TRACKING)
-void SkiaCompositingLayer::collectFrameDamage(SkCanvas& canvas, PaintContext& context, const TransformationMatrix& transform)
+void SkiaCompositingLayer::collectFrameDamage(SkCanvas& canvas, PaintContext& context)
 {
     if (!frameDamagePropagationEnabled() || !context.frameDamage)
         return;
 
+    TransformationMatrix transform(context.accumulatedReplicaTransform);
+    transform.multiply(m_transforms.combined);
     auto frameDamage = transform.mapRect(effectiveLayerRect());
     auto clipBounds = FloatRect(this->clipBounds(canvas, context));
     if (!clipBounds.isEmpty())
@@ -567,10 +605,11 @@ void SkiaCompositingLayer::paintDebugIndicators(SkCanvas& canvas, PaintContext& 
 
     TransformationMatrix transform(context.accumulatedReplicaTransform);
     transform.multiply(m_transforms.combined);
-    SkAutoCanvasRestore autoRestore(&canvas, true);
-    canvas.concat(SkM44(transform));
 
     if (m_debugBorder) {
+        SkAutoCanvasRestore autoRestore(&canvas, true);
+        canvas.concat(SkM44(transform));
+
         SkPaint borderPaint;
         borderPaint.setStyle(SkPaint::kStroke_Style);
         borderPaint.setColor(SkColor(m_debugBorder->color));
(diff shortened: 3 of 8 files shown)


Source/WebCore/platform/graphics/skia/SkiaBackingStore.cpp
```diff
@@ -90,6 +90,26 @@ void SkiaBackingStore::paintToCanvas(SkCanvas& canvas, const SkPaint& paint)
...
+
+    FloatRect layerRect = { { }, m_size };
+
+    Vector<SkCanvas::ImageSetEntry> images;
+    for (auto& tile : m_tiles.values()) {
+        const auto& image = tile.image();
+        if (!image)
+            continue;
+
+        // FIXME: implement per edge anitaliasing.
```
nikolaszimmermann: typo antialiasing

Source/WebCore/platform/graphics/skia/SkiaBackingStore.h
```diff
@@ -47,6 +48,7 @@ class SkiaBackingStore {
 
     void update(const FloatSize&, float scale, CoordinatedBackingStoreProxy::Update&&);
     void paintToCanvas(SkCanvas&, const SkPaint&);
+    Vector<SkCanvas::ImageSetEntry> imageSet(size_t matrixIndex, float opacity, bool enableAntialias);
```
nikolaszimmermann: Placed a bit odd here, I'd move it 2 lines down.

The function name is better called "buildImageSet", it implies it's a simple accessor as it its now.
What is matrixIndex?

Without any context this is quite hard to digest.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
```diff
@@ -123,6 +123,15 @@ class SkiaCompositingLayer final : public RefCountedAndCanMakeWeakPtr<SkiaCompos
 
     bool computeTransformsAndAnimations(const TransformationMatrix& parentTransform, const TransformationMatrix& futureParentTransform, MonotonicTime);
 
+    struct ImageSetBatch {
+        Vector<SkCanvas::ImageSetEntry> imageSet;
+        Vector<SkPoint> dstClips;
+        Vector<SkMatrix> preViewMatrices;
```
nikolaszimmermann: This doesn't really ring a bell, what is a preViewMatrix? "preview"? "before-view" ? ?

nikolaszimmermann (commented):
I will continue review, need to think a bit about it.

Source/WebCore/platform/graphics/skia/SkiaBackingStore.h
carlosgcampos: Yes, the Skia API is not easy:

This is an ImageEntry:

```
struct SK_API ImageSetEntry {
    ImageSetEntry(sk_sp<const SkImage> image, const SkRect& srcRect, const SkRect& dstRect,
                  int matrixIndex, float alpha, unsigned aaFlags, bool hasClip);

    ImageSetEntry(sk_sp<const SkImage> image, const SkRect& srcRect, const SkRect& dstRect,
                  float alpha, unsigned aaFlags);

    ImageSetEntry();
    ~ImageSetEntry();
    ImageSetEntry(const ImageSetEntry&);
    ImageSetEntry& operator=(const ImageSetEntry&);

    sk_sp<const SkImage> fImage;
    SkRect fSrcRect;
    SkRect fDstRect;
    int fMatrixIndex = -1; // Index into the preViewMatrices arg, or < 0                                                                                                                                        
    float fAlpha = 1.f;
    unsigned fAAFlags = kNone_QuadAAFlags; // QuadAAFlags                                                                                                                                                       
    bool fHasClip = false; // True to use next 4 points in dstClip arg as quad
}
```

And this is the call to draw a image set:

```
void experimental_DrawEdgeAAImageSet(const ImageSetEntry imageSet[], int cnt,
                                         const SkPoint dstClips[], const SkMatrix preViewMatrices[],
                                         const SkSamplingOptions&, const SkPaint* paint = nullptr,
                                         SrcRectConstraint constraint = kStrict_SrcRectConstraint);
```

As you can see it receives an array of SkMatrix, but this array doesn't need to be the same size as the image set, because multiple images could share the same transform. So when adding several images with the same transform we don't group the SkMatrix array and create the ImageEntry with the same index of the matrices array. Something similar happens with the clip, ImageEntry is marked as having a clip or not, and clips are stored in the array in order.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
carlosgcampos: I'm using the same names as the skia API.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -566,9 +659,18 @@ void SkiaCompositingLayer::collectFrameDamage(SkCanvas& canvas, PaintContext& co
...
 {
+    if (!m_debugBorder && !m_repaintCount)
+        return;
+
+    TransformationMatrix transform(context.accumulatedReplicaTransform);
+    transform.multiply(m_transforms.combined);
+
     if (m_debugBorder) {
+        SkAutoCanvasRestore autoRestore(&canvas, true);
+        canvas.concat(SkM44(transform));
```
nikolaszimmermann, line 611: Hmm, will this break the repaint counters below? Isn't that now missing the canvas.concat(...) operation?

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -469,46 +536,57 @@ void SkiaCompositingLayer::paintSelf(SkCanvas& canvas, PaintContext& context)
...
-            clipRect(canvas, m_contentsClippingRect);
+        bool shouldPaintNow = m_contentsClippingRect.isRounded() || (!m_contentsBuffer && !m_contentsTiling.size.isEmpty()) || is<CoordinatedPlatformLayerBufferHolePunch>(m_contentsBuffer);
+        if (shouldPaintNow)
+            paintBatchedImageSetIfNeeded(canvas, context);
+
+        SkAutoCanvasRestore autoRestore(&canvas, shouldPaintNow);
+        if (shouldPaintNow) {
+            canvas.concat(SkM44(transform));
+
+            if (m_contentsClippingRect.isRounded())
```
nikolaszimmermann: What about the previous clipping condition !m_contentsClippingRect.rect().contains(m_contentsRect) -- isn't that missing now?

Source/WebCore/platform/graphics/skia/SkiaBackingStore.cpp
```diff
@@ -90,6 +90,26 @@ void SkiaBackingStore::paintToCanvas(SkCanvas& canvas, const SkPaint& paint)
...
+        return { };
+
+    FloatRect layerRect = { { }, m_size };
+
+    Vector<SkCanvas::ImageSetEntry> images;
+    for (auto& tile : m_tiles.values()) {
+        const auto& image = tile.image();
+        if (!image)
+            continue;
+
```
nikolaszimmermann, line 105: We used to call paintToCanvas() directly before the call-sites switched to batching -- isn't the "quick cull" logic now missing?

```
        if (canvas.quickReject(tile.rect()))
            continue;
```

There is no canvas here, I just wonder if that's a perf-deficit now...

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
carlosgcampos: No, repaint counter reset the matrix, now we simply don't change it and we don't need to reset

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
carlosgcampos: it's handled when adding the clip below

```
FloatRect clippingRect = m_contentsClippingRect.rect();
    if (clippingRect.contains(m_contentsRect))
        clippingRect = { };
```

Source/WebCore/platform/graphics/skia/SkiaBackingStore.cpp
carlosgcampos: Yes, we can't do the same because canvas isn't transformed. It will be rejected when painting like texture mapper does. We could transform he canvas and pass it here and then restore, but I'm not sure it's worth it.

nikolaszimmermann:
IIUC paintBatchedImageSetIfNeeded must be manually invoked before every CTM/clip/surface boundary (twice in paintContents, twice in paintSelfAndChildren, twice in paintWithIntermediateSurface, etc..) - no RAII guards, no assertions, so it's easy to get the wrong in future -- any concat/clipRect/save added without the matching flush would silently render the batch through the wrong matrix/clip...

I think we could have a ScopedImageSetBatchFlush(SkCanvas&, ImageSetBatch&, bool active) replacing the whole SkAutoCanvasRestore -- so it does all the operations we need in the right order (when active == true).

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
carlosgcampos: You are right, it's still missing, because we might paint now for other reasons than clipping.

Source/WebCore/platform/graphics/skia/SkiaBackingStore.cpp
nikolaszimmermann: Ok.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
nikolaszimmermann: Aha ok.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
nikolaszimmermann: Then makes sense to leave it as-is, ok. Should have checked the API docs first.

Source/WebCore/platform/graphics/skia/SkiaBackingStore.h
nikolaszimmermann: Indeed, a very convoluted API...

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -445,6 +446,97 @@ bool SkiaCompositingLayer::paint(SkCanvas& canvas, std::optional<Damage>& damage
     return hasRunningAnimations;
 }
 
+SkiaCompositingLayer::ScopedImageSetBatchFlush::ScopedImageSetBatchFlush(SkCanvas& canvas, ImageSetBatch& imageSetBatch, Mode mode)
```
nikolaszimmermann: Can we move all of the ImageSetBatch into a dedicated SkiaImageSetBatch file? SkiaCompositingLayer is already large enough alreadye

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -445,6 +446,97 @@ bool SkiaCompositingLayer::paint(SkCanvas& canvas, std::optional<Damage>& damage
...
+    }
+}
+
+SkiaCompositingLayer::ScopedImageSetBatchFlush::~ScopedImageSetBatchFlush()
+{
+    if (m_mode == Mode::FlushBeforeAndAfter)
+        m_imageSetBatch.flushIfNeeded(m_canvas);
+
+    if (m_mode != Mode::DoNothing)
+        m_canvas.restoreToCount(m_saveCount);
```
nikolaszimmermann: I guess that comes from the SkAuto... canvas restore logic?

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -445,6 +446,97 @@ bool SkiaCompositingLayer::paint(SkCanvas& canvas, std::optional<Damage>& damage
...
+}
+
+void SkiaCompositingLayer::ImageSetBatch::updatePaintProperties(SkCanvas& canvas, const sk_sp<SkColorFilter> colorFilter, const std::optional<SkBlendMode>& blendMode)
```
nikolaszimmermann: Maybe pass the colorFilter by const ref?

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -469,46 +561,69 @@ void SkiaCompositingLayer::paintSelf(SkCanvas& canvas, PaintContext& context)
...
-    paint.setStyle(SkPaint::kFill_Style);
-    auto ctm = canvas.getLocalToDeviceAs3x3();
-    paint.setAntiAlias(!ctm.preservesAxisAlignment() && !ctm.preservesRightAngles());
-    paint.setAlphaf(context.opacity);
-    if (context.blendMode)
-        paint.setBlendMode(*context.blendMode);
-    if (context.colorFilter)
-        paint.setColorFilter(context.colorFilter);
+    TransformationMatrix transform(context.accumulatedReplicaTransform);
+    transform.multiply(m_transforms.combined);
```
nikolaszimmermann, line 479: A newline here wouldn't hurt between the multiply and the auto ctm

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -469,46 +561,69 @@ void SkiaCompositingLayer::paintSelf(SkCanvas& canvas, PaintContext& context)
...
+    auto ctm = SkM44(transform).asM33();
+    bool enableAntialias = !ctm.preservesAxisAlignment() && !ctm.preservesRightAngles();
+
+    context.imageSetBatch.updatePaintProperties(canvas, context.colorFilter, context.blendMode);
+
+    auto setupPaint = [&](SkPaint& paint) {
```
nikolaszimmermann: I'd just use `[&]() -> SkPaint { ... return paint; }` here, to shorten all call-sites since you always create a new SkPaint and never mutate an existing one.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
```diff
@@ -529,24 +646,37 @@ void SkiaCompositingLayer::paintContents(SkCanvas& canvas, PaintContext& context
...
         if (image) {
-            canvas.drawImageRect(image, SkRect::MakeSize(SkSize::Make(image->dimensions())), SkRect(m_contentsRect),
-                SkSamplingOptions(SkFilterMode::kLinear, SkMipmapMode::kNone), &paint, SkCanvas::kFast_SrcRectConstraint);
+            if (shouldPaintNow) {
+                SkPaint paint;
+                setupPaint(paint);
+                canvas.drawImageRect(image, SkRect::MakeSize(SkSize::Make(image->dimensions())), SkRect(m_contentsRect),
+                    SkSamplingOptions(SkFilterMode::kLinear, SkMipmapMode::kNone), &paint, SkCanvas::kFast_SrcRectConstraint);
+            } else {
+                FloatRect clippingRect = m_contentsClippingRect.rect();
```
nikolaszimmermann: auto possible?

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
```diff
@@ -123,6 +123,41 @@ class SkiaCompositingLayer final : public RefCountedAndCanMakeWeakPtr<SkiaCompos
 
     bool computeTransformsAndAnimations(const TransformationMatrix& parentTransform, const TransformationMatrix& futureParentTransform, MonotonicTime);
 
+    class ImageSetBatch {
```
nikolaszimmermann: Much nicer, but I'd move that into a SkiaImageSetBatch class in a dedicated file and _remove_ the dependency from SkiaBackingStore (that's in addImageSet right now).

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
```diff
@@ -123,6 +123,41 @@ class SkiaCompositingLayer final : public RefCountedAndCanMakeWeakPtr<SkiaCompos
...
+        Vector<SkPoint> m_dstClips;
+        Vector<SkMatrix> m_preViewMatrices;
+        sk_sp<SkColorFilter> m_colorFilter;
+        std::optional<SkBlendMode> m_blendMode;
+        SkSamplingOptions m_samplingOptions;
+    };
+
+    class ScopedImageSetBatchFlush {
+    public:
+        enum class Mode : uint8_t { DoNothing, FlushBefore, FlushBeforeAndAfter };
```
nikolaszimmermann: Hmm, usually that goes in multiple lines - why doesn't the style checker complain? :-)

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
```diff
@@ -174,6 +210,12 @@ class SkiaCompositingLayer final : public RefCountedAndCanMakeWeakPtr<SkiaCompos
     void collect3DRenderingContextLayers(Vector<Ref<SkiaCompositingLayer>>&);
     void recursiveCleanUpAfterPaint();
 
+    void updateImageSetBatchPaintProperties(SkCanvas&, PaintContext&);
```
nikolaszimmermann: These can be removed.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
```diff
@@ -123,6 +123,41 @@ class SkiaCompositingLayer final : public RefCountedAndCanMakeWeakPtr<SkiaCompos
 
     bool computeTransformsAndAnimations(const TransformationMatrix& parentTransform, const TransformationMatrix& futureParentTransform, MonotonicTime);
 
+    class ImageSetBatch {
+    public:
+        void updatePaintProperties(SkCanvas&, const sk_sp<SkColorFilter>, const std::optional<SkBlendMode>&);
+        void updateSamplingOptions(SkCanvas&, SkSamplingOptions);
+        void addImageSet(SkCanvas&, SkiaBackingStore&, const SkMatrix&, float opacity, bool enableAntialias);
```
nikolaszimmermann: If you move that out of SkiaCompositingLayer, you can pass a Vector<SkCanvas::ImageSetEntry> here instead of a SkiaBackingStore - I think that decouples the classes.

nikolaszimmermann (approved):
Looks much nicer, I'll already approve it (if EWS is happy) and got some additional comments.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
carlosgcampos: Yes

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
carlosgcampos: Yes, this was supposed to be a reference.

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
carlosgcampos: I thought about that, I guess there aren't copies, right?

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
carlosgcampos: I guess

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
carlosgcampos: How can I remove the dependency?

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
carlosgcampos: I've seen it this way for small enums

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
carlosgcampos: Indeed

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.h
carlosgcampos: The problem is that the backing store needs to know the matrix index when creating the ImageEntry

Source/WebCore/platform/graphics/skia/SkiaCompositingLayer.cpp
carlosgcampos: Well, this is not so big and it's easier to find when not having to move to another file, but if you really think it's better I don't mind to move it.

PR #66499, opened 2026-06-04
tumblr.com: Saving .gifv image adds .html to end of filename

J0pan:
#### d5d7c2176bc747df9090acab2eac8193424b4273
<pre>
tumblr.com: Saving .gifv image adds .html to end of filename
<a href="https://bugs.webkit.org/show_bug.cgi?id=316350">https://bugs.webkit.org/show_bug.cgi?id=316350</a>
<a href="https://rdar.apple.com/70161124">rdar://70161124</a>

Reviewed by Sammy Gill.

The problem: when an &lt;img&gt;&apos;s url coincides with the main document url,
Save Image suggests a filename with the wrong extension
(e.g., tumblr_inline_*.gifv.html). WebFrame::suggestedFilenameForResourceWithURL
and mimeTypeForResourceWithURL short-circuit to the document loader&apos;s
response when the image url equals the document url, so an &lt;img&gt; on an
HTML interstitial gets the document&apos;s text/html MIME instead of the
image&apos;s image/* MIME. NSURLResponse.suggestedFilename then appends .html
to the url-derived filename, and MIMETypeRegistry::correctExtensionForMIMEType
can&apos;t recover when the url extension is unregistered.

The fix: add a WebFrame::ResourceType { Generic, Image } hint and thread it
through suggestedFilenameForResourceWithURL and mimeTypeForResourceWithURL.
When the caller passes ResourceType::Image, the main resource short circuit
skips itself if the document&apos;s MIME is non-image (the contradiction the bug
hinges on), and the methods fall through to CachedResourceLoader::cachedResource
to find the actual image&apos;s response. WebHitTestResultData no longer reaches
into the render tree from Source/WebKit/Shared/.

Test: Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm

* Source/WebKit/Shared/WebHitTestResultData.cpp:
(WebKit::imageSuggestedFilenameFromHitTestResult):
* Source/WebKit/WebProcess/WebPage/WebFrame.cpp:
(WebKit::WebFrame::suggestedFilenameForResourceWithURL const):
(WebKit::WebFrame::mimeTypeForResourceWithURL const):
* Source/WebKit/WebProcess/WebPage/WebFrame.h:
* Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm:
(TestWebKitAPI::imageSuggestedFilenameFromCollidingImageURL):
(TestWebKitAPI::TEST(ContextMenuTests, HitTestResultImageSuggestedFilenameWhenURLCollidesWithMainResource)):
(TestWebKitAPI::TEST(ContextMenuTests, HitTestResultImageSuggestedFilenameWhenURLCollidesAndNoContentDisposition)):

Canonical link: <a href="https://commits.webkit.org/315059@main">https://commits.webkit.org/315059@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/0dced01ed08b2e9128a61cb00bc190b2452d514e

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/167620 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/41117 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/34712 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/176677 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/125301 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/bfd81866-3e54-47d9-ab56-4d15652fa419) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/169486 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/41552 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/41129 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/130416 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/91678 "1 flakes 5 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/cb858161-79e9-456b-8bfa-6b933f985360) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/170579 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/32833 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/150611 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/110933 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/974b3a2e-9284-4a4f-bfd8-d039174aac03) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/31815 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/30611 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/25410 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/141802 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/28306 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/179217 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/32258 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/30024 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/138818 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/41189 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/34670 "Passed tests") | [❌ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/138703 "") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/37428 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/41877 "Built successfully") | | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/105037 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/33958 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/26977 "Passed tests") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/40381 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/113627 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/39778 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/5078 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/40029 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/39923 "Built successfully") | | | |

### DIFF
--- a/Source/WebKit/Shared/WebHitTestResultData.cpp
+++ b/Source/WebKit/Shared/WebHitTestResultData.cpp
@@ -34,6 +34,7 @@
 #include <WebCore/NavigationAction.h>
 #include <WebCore/Node.h>
 #include <WebCore/RenderImage.h>
+#include <WebCore/ResourceResponse.h>
 #include <WebCore/SharedBuffer.h>
 #include <WebCore/Text.h>
 #include <wtf/URL.h>
@@ -81,8 +82,8 @@ static String imageSuggestedFilenameFromHitTestResult(const HitTestResult& hitTe
         return nullString();
 
     auto url = hitTestResult.absoluteImageURL();
-    auto filename = webFrame->suggestedFilenameForResourceWithURL(url);
-    auto mimeType = webFrame->mimeTypeForResourceWithURL(url);
+    auto filename = webFrame->suggestedFilenameForResourceWithURL(url, WebFrame::ResourceType::Image);
+    auto mimeType = webFrame->mimeTypeForResourceWithURL(url, WebFrame::ResourceType::Image);
     return MIMETypeRegistry::correctExtensionForMIMEType(filename, mimeType);
 }
 
--- a/Source/WebKit/WebProcess/WebPage/WebFrame.cpp
+++ b/Source/WebKit/WebProcess/WebPage/WebFrame.cpp
@@ -1221,7 +1221,7 @@ String WebFrame::provisionalURL() const
     return provisionalDocumentLoader->url().string();
 }
 
-String WebFrame::suggestedFilenameForResourceWithURL(const URL& url) const
+String WebFrame::suggestedFilenameForResourceWithURL(const URL& url, ResourceType resourceType) const
 {
     RefPtr localFrame = dynamicDowncast<LocalFrame>(m_coreFrame.get());
     if (!localFrame)
@@ -1231,11 +1231,12 @@ String WebFrame::suggestedFilenameForResourceWithURL(const URL& url) const
     if (!loader)
         return String();
 
-    // First, try the main resource.
-    if (loader->url() == url)
-        return loader->response().suggestedFilename();
+    if (loader->url() == url) {
+        const auto& mimeType = loader->response().mimeType();
+        if (resourceType != ResourceType::Image || mimeType.startsWithIgnoringASCIICase("image/"_s))
+            return loader->response().suggestedFilename();
+    }
 
-    // Next, try subresources.
     RefPtr<ArchiveResource> resource = loader->subresource(url);
     if (resource)
         return resource->response().suggestedFilename();
@@ -1243,7 +1244,7 @@ String WebFrame::suggestedFilenameForResourceWithURL(const URL& url) const
     return String();
 }
 
-String WebFrame::mimeTypeForResourceWithURL(const URL& url) const
+String WebFrame::mimeTypeForResourceWithURL(const URL& url, ResourceType resourceType) const
 {
     RefPtr localFrame = dynamicDowncast<LocalFrame>(m_coreFrame.get());
     if (!localFrame)
@@ -1253,11 +1254,12 @@ String WebFrame::mimeTypeForResourceWithURL(const URL& url) const
     if (!loader)
         return String();
 
-    // First, try the main resource.
-    if (loader->url() == url)
-        return loader->response().mimeType();
+    if (loader->url() == url) {
+        const auto& mimeType = loader->response().mimeType();
+        if (resourceType != ResourceType::Image || mimeType.startsWithIgnoringASCIICase("image/"_s))
+            return mimeType;
+    }
 
-    // Next, try subresources.
     RefPtr<ArchiveResource> resource = loader->subresource(url);
     if (resource)
         return resource->mimeType();
--- a/Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
+++ b/Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
@@ -27,6 +27,7 @@
 
 #if PLATFORM(MAC)
 
+#import "Helpers/cocoa/HTTPServer.h"
 #import "Helpers/mac/AppKitSPI.h"
 #import "InstanceMethodSwizzler.h"
 #import "Helpers/cocoa/PasteboardUtilities.h"
@@ -756,6 +757,66 @@ TEST(ContextMenuTests, HitTestResultImageSuggestedFilename)
     [webView mouseDownAtPoint:NSMakePoint(200, 200) simulatePressure:NO withFlags:0 eventType:NSEventTypeRightMouseDown];
     [webView mouseUpAtPoint:NSMakePoint(200, 200) withFlags:0 eventType:NSEventTypeRightMouseUp];
     Util::run(&gotProposedMenu);
+}
+
+static RetainPtr<NSString> imageSuggestedFilenameFromCollidingImageURL(HashMap<String, String> imageResponseHeaders)
+{
+    using namespace TestWebKitAPI;
+
+    RetainPtr imageData = [NSData dataWithContentsOfFile:[NSBundle.test_resourcesBundle pathForResource:@"sunset-in-cupertino-200px" ofType:@"png"]];
+
+    HTTPServer server([imageData, imageResponseHeaders](Connection connection) {
+        connection.receiveHTTPRequest([connection, imageData, imageResponseHeaders](Vector<char>&&) {
+            connection.send(HTTPResponse({ { "Content-Type"_s, "text/html"_s } },
+                "<img id='collision' src='collision.gifv' style='width:300px;height:300px'>"_s).serialize(), [connection, imageData, imageResponseHeaders] {
+                connection.receiveHTTPRequest([connection, imageData, imageResponseHeaders](Vector<char>&&) {
+                    HashMap<String, String> headers = imageResponseHeaders;
+                    connection.send(HTTPResponse(WTF::move(headers), imageData).serialize());
+                });
+            });
+        });
+    });
+
+    RetainPtr delegate = adoptNS([[TestUIDelegate alloc] init]);
+    __block bool gotProposedMenu = false;
+    __block RetainPtr<NSString> capturedFilename;
+    [delegate setGetContextMenuFromProposedMenu:^(NSMenu *menu, _WKContextMenuElementInfo *elementInfo, id<NSSecureCoding>, void (^completion)(NSMenu *)) {
+        EXPECT_NOT_NULL(elementInfo.hitTestResult);
+        capturedFilename = elementInfo.hitTestResult.imageSuggestedFilename;
+        completion(nil);
+        gotProposedMenu = true;
+    }];
+
+    RetainPtr webView = adoptNS([[TestWKWebView alloc] initWithFrame:NSMakeRect(0, 0, 400, 400)]);
+    [webView setUIDelegate:delegate];
+    RetainPtr collisionURL = [NSURL URLWithString:[NSString stringWithFormat:@"http://127.0.0.1:%d/collision.gifv", server.port()]];
+    [webView synchronouslyLoadRequest:[NSURLRequest requestWithURL:collisionURL]];
+
+    auto midpoint = [webView getElementMidpoint:@"#collision"].value();
+    [webView mouseDownAtPoint:midpoint simulatePressure:NO withFlags:0 eventType:NSEventTypeRightMouseDown];
+    [webView mouseUpAtPoint:midpoint withFlags:0 eventType:NSEventTypeRightMouseUp];
+    Util::run(&gotProposedMenu);
+
+    return capturedFilename;
+}
+
+TEST(ContextMenuTests, HitTestResultImageSuggestedFilenameWhenURLCollidesWithMainResource)
+{
+    RetainPtr filename = imageSuggestedFilenameFromCollidingImageURL({
+        { "Content-Type"_s, "image/png"_s },
+        { "Content-Disposition"_s, "inline; filename=\"collision.png\""_s },
+    });
+    EXPECT_FALSE([filename hasSuffix:@".html"]);
+    EXPECT_TRUE([filename hasSuffix:@".png"]);
+}
+
+TEST(ContextMenuTests, HitTestResultImageSuggestedFilenameWhenURLCollidesAndNoContentDisposition)
+{
+    RetainPtr filename = imageSuggestedFilenameFromCollidingImageURL({
+        { "Content-Type"_s, "image/png"_s },
+    });
+    EXPECT_FALSE([filename hasSuffix:@".html"]);
+    EXPECT_GT([filename length], 0u);
 }
 
 TEST(ContextMenuTests, HitTestResultLinkWithInvalidURL)
--- a/Source/WebKit/WebProcess/WebPage/WebFrame.h
+++ b/Source/WebKit/WebProcess/WebPage/WebFrame.h
@@ -224,8 +224,9 @@ public:
     bool allowsFollowingLink(const URL&) const;
 
     String provisionalURL() const;
-    String suggestedFilenameForResourceWithURL(const URL&) const;
-    String mimeTypeForResourceWithURL(const URL&) const;
+    enum class ResourceType : uint8_t { Generic, Image };
+    String suggestedFilenameForResourceWithURL(const URL&, ResourceType = ResourceType::Generic) const;
+    String mimeTypeForResourceWithURL(const URL&, ResourceType = ResourceType::Generic) const;
 
     void setTextDirection(const String&);
     void updateFrameRectFromRemote(WebCore::IntRect);


aprotyas (commented):
typo in the commit message:

> The fix: the fix:

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,115 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
     Util::run(&gotProposedMenu);
 }
 
+// rdar://70161124
```
aprotyas: No need to annotate with the bug you're fixing. This is easily reachable through git log.

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,115 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
...
+        0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x01,
+        0x08, 0x00, 0x00, 0x00, 0x00, 0x3a, 0x7e, 0x9b,
+        0x55, 0x00, 0x00, 0x00, 0x0a, 0x49, 0x44, 0x41,
+        0x54, 0x78, 0x9c, 0x63, 0x00, 0x00, 0x00, 0x02,
+        0x00, 0x01, 0xe2, 0x21, 0xbc, 0x33, 0x00, 0x00,
+        0x00, 0x00, 0x49, 0x45, 0x4e, 0x44, 0xae, 0x42,
+        0x60, 0x82,
+    } };
+
+    HTTPServer server([] (Connection connection) {
```
aprotyas: style guide nit: no space between capture list and arg list
```suggestion
    HTTPServer server([](Connection connection) {
```

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,115 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
...
+        connection.receiveHTTPRequest([=](Vector<char>&&) {
```
aprotyas: just capture `connection`
```suggestion
        connection.receiveHTTPRequest([connection](Vector<char>&&) {
```

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,115 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
...
+            connection.send(HTTPResponse({ { "Content-Type"_s, "text/html"_s } },
+                "<img src='collision.gifv' style='width:300px;height:300px'>"_s).serialize(), [=] {
```
aprotyas: same here
```suggestion
                "<img src='collision.gifv' style='width:300px;height:300px'>"_s).serialize(), [connection] {
```

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,115 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
...
+                connection.receiveHTTPRequest([=](Vector<char>&&) {
```
aprotyas: and here
```suggestion
                connection.receiveHTTPRequest([connection](Vector<char>&&) {
```

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,115 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
     Util::run(&gotProposedMenu);
 }
 
+// rdar://70161124
+TEST(ContextMenuTests, HitTestResultImageSuggestedFilenameWhenURLCollidesWithMainResource)
```
rr-codes: optional, please consider authoring these tests (and/or future tests) using Swift Testing, which offers much better ergonomics and is easier to use

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,115 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
...
+    HTTPServer server([] (Connection connection) {
+        connection.receiveHTTPRequest([=](Vector<char>&&) {
+            connection.send(HTTPResponse({ { "Content-Type"_s, "text/html"_s } },
+                "<img src='collision.gifv' style='width:300px;height:300px'>"_s).serialize(), [=] {
+                connection.receiveHTTPRequest([=](Vector<char>&&) {
+                    RetainPtr imageData = [NSData dataWithBytes:pngBytes.data() length:pngBytes.size()];
+                    connection.send(HTTPResponse({
+                        { "Content-Type"_s, "image/png"_s },
+                        { "Content-Disposition"_s, "inline; filename=\"collision.png\""_s },
+                    }, imageData.get()).serialize());
```
aprotyas: you can drop the `.get()` here
```suggestion
                    }, imageData).serialize());
```

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,115 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
...
+    __block RetainPtr<NSString> capturedFilename;
+    [delegate setGetContextMenuFromProposedMenu:^(NSMenu *menu, _WKContextMenuElementInfo *elementInfo, id<NSSecureCoding>, void (^completion)(NSMenu *)) {
+        EXPECT_NOT_NULL(elementInfo.hitTestResult);
+        capturedFilename = elementInfo.hitTestResult.imageSuggestedFilename;
+        completion(nil);
+        gotProposedMenu = true;
+    }];
+
+    RetainPtr webView = adoptNS([[TestWKWebView alloc] initWithFrame:NSMakeRect(0, 0, 400, 400)]);
+    [webView setUIDelegate:delegate.get()];
```
aprotyas: nit: you can drop the `get()`
```suggestion
    [webView setUIDelegate:delegate];
```

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,115 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
...
+    NSURL *collisionURL = [NSURL URLWithString:[NSString stringWithFormat:@"http://127.0.0.1:%d/collision.gifv", server.port()]];
```
aprotyas: nit
```suggestion
    RetainPtr collisionURL = [NSURL URLWithString:[NSString stringWithFormat:@"http://127.0.0.1:%d/collision.gifv", server.port()]];
```

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,115 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
...
+    [webView synchronouslyLoadRequest:[NSURLRequest requestWithURL:collisionURL]];
+
+    [webView mouseDownAtPoint:NSMakePoint(20, 380) simulatePressure:NO withFlags:0 eventType:NSEventTypeRightMouseDown];
+    [webView mouseUpAtPoint:NSMakePoint(20, 380) withFlags:0 eventType:NSEventTypeRightMouseUp];
```
aprotyas: can you replace the hardcoded coordinates with whatever you get from `-getElementMidpoint:`?

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,115 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
...
+    Util::run(&gotProposedMenu);
+
+    EXPECT_FALSE([capturedFilename hasSuffix:@".html"]);
+    EXPECT_TRUE([capturedFilename hasSuffix:@".png"]);
+}
+
+// https://bugs.webkit.org/show_bug.cgi?id=316350
```
aprotyas: Same thing, no need to link to a bug tracker.

Source/WebKit/Shared/WebHitTestResultData.cpp
```diff
@@ -83,6 +85,25 @@ static String imageSuggestedFilenameFromHitTestResult(const HitTestResult& hitTe
     auto url = hitTestResult.absoluteImageURL();
     auto filename = webFrame->suggestedFilenameForResourceWithURL(url);
     auto mimeType = webFrame->mimeTypeForResourceWithURL(url);
+
+    // https://bugs.webkit.org/show_bug.cgi?id=316350
```
aprotyas: No need for a bug tracker link here.

aprotyas (commented):
I left a bunch of small comments. I think there's a bunch of code you can share between the two test cases? Since it looks like only the server implementations are slightly different.

Source/WebKit/Shared/WebHitTestResultData.cpp
```diff
@@ -83,6 +85,25 @@ static String imageSuggestedFilenameFromHitTestResult(const HitTestResult& hitTe
...
+    if (!mimeType.isEmpty() && !mimeType.startsWithIgnoringASCIICase("image/"_s)) {
+        if (RefPtr target = hitTestResult.innerNonSharedNode()) {
+            if (CheckedPtr renderer = dynamicDowncast<RenderImage>(target->renderer())) {
+                if (RefPtr cachedImage = renderer->cachedImage(); cachedImage && !cachedImage->errorOccurred()) {
```
sammygill: Is there a way we could approach this without having to reach through the DOM and render trees at this point? Since we already know that the resource is an image at this point can we somehow use that information when getting the suggested filename and MIME type?

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,77 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
...
+        EXPECT_NOT_NULL(elementInfo.hitTestResult);
+        capturedFilename = elementInfo.hitTestResult.imageSuggestedFilename;
+        completion(nil);
+        gotProposedMenu = true;
+    }];
+
+    RetainPtr webView = adoptNS([[TestWKWebView alloc] initWithFrame:NSMakeRect(0, 0, 400, 400)]);
+    [webView setUIDelegate:delegate.get()];
+    RetainPtr collisionURL = [NSURL URLWithString:[NSString stringWithFormat:@"http://127.0.0.1:%d/collision.gifv", server.port()]];
+    [webView synchronouslyLoadRequest:[NSURLRequest requestWithURL:collisionURL.get()]];
```
aprotyas: I think you can drop all of the `.get()` calls in here

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,77 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
...
+    [webView mouseDownAtPoint:midpoint simulatePressure:NO withFlags:0 eventType:NSEventTypeRightMouseDown];
+    [webView mouseUpAtPoint:midpoint withFlags:0 eventType:NSEventTypeRightMouseUp];
+    Util::run(&gotProposedMenu);
+
+    return capturedFilename;
+}
+
+TEST(ContextMenuTests, HitTestResultImageSuggestedFilenameWhenURLCollidesWithMainResource)
+{
+    auto filename = imageSuggestedFilenameFromCollidingImageURL({
```
aprotyas:
```suggestion
    RetainPtr filename = imageSuggestedFilenameFromCollidingImageURL({
```

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,77 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
...
+        { "Content-Type"_s, "image/png"_s },
+        { "Content-Disposition"_s, "inline; filename=\"collision.png\""_s },
+    });
+    EXPECT_FALSE([filename hasSuffix:@".html"]);
+    EXPECT_TRUE([filename hasSuffix:@".png"]);
+}
+
+TEST(ContextMenuTests, HitTestResultImageSuggestedFilenameWhenURLCollidesAndNoContentDisposition)
+{
+    auto filename = imageSuggestedFilenameFromCollidingImageURL({
```
aprotyas:
```suggestion
    RetainPtr filename = imageSuggestedFilenameFromCollidingImageURL({
```

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
J0pan: kept gtest here to stay consistent with surroundings, would you prefer if I opened a separate rdar to switch all to swift testing?

rr-codes: yes, that would be great, thanks!!

Source/WebKit/Shared/WebHitTestResultData.cpp
J0pan: that's a good point, agree that it shouldn't walk into this layer, for this pr I pulled the walk into a static `imageResponseFromHitTestResult` so the call site stops touching dom/render directly.

J0pan: would you prefer a stronger divide? I can add `WebFrame::imageResponseForURL(url)` / `HitTestResult::imageResourceResponse()` accessor that returns the image's `ResourceResponse` from the memory cache, so this file stops walking the render tree at all.

should I do that as a sibling change, or incorporate that change in this patch?

sammygill: Well, I guess I am wondering is whether we even need to touch the render/DOM trees at all? From what I can tell the hit test result exposes some functionality to determine information regarding the contents of the hit test, such as if it contains a whole image (`HitTestResult::hasEntireImage`), which is similar to what you are doing. 

So, if we get a result back from `suggestedFilenameForResourceWithURL`/`mimeTypeForResourceWithURL` that tells us it is a HTML document I think it might be safe to assume that it is incorrect since that contradicts the previous information we already have? So maybe we should pass in that additional information to those two functions so that they can also use it when figuring out what to return?

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
J0pan: rdar://179033385

Source/WebKit/WebProcess/WebPage/WebFrame.cpp
```diff
@@ -1228,19 +1232,28 @@ String WebFrame::suggestedFilenameForResourceWithURL(const URL& url) const
...
         return resource->response().suggestedFilename();
 
+    if (type == ResourceType::Image) {
+        if (RefPtr document = localFrame->document()) {
+            RefPtr cached = document->cachedResourceLoader().cachedResource(url.string());
+            if (RefPtr cachedImage = dynamicDowncast<CachedImage>(cached); cachedImage && !cachedImage->errorOccurred())
+                return cachedImage->response().suggestedFilename();
+        }
+    }
+
```
sammygill: I would have thought that the logic to checking the subresource would have fixed the bug. Is that not the case? If it does then I think it might be worth adding an additional testcase to demonstrate why we need this extra logic. It would also be worth to double check what other browsers are doing in this scenario.

Source/WebKit/WebProcess/WebPage/WebFrame.cpp
```diff
@@ -1250,15 +1263,24 @@ String WebFrame::mimeTypeForResourceWithURL(const URL& url) const
...
     if (resource)
         return resource->mimeType();
 
+    if (type == ResourceType::Image) {
+        if (RefPtr document = localFrame->document()) {
+            RefPtr cached = document->cachedResourceLoader().cachedResource(url.string());
+            if (RefPtr cachedImage = dynamicDowncast<CachedImage>(cached); cachedImage && !cachedImage->errorOccurred())
+                return cachedImage->response().mimeType();
+        }
+    }
```
sammygill: I would have thought that the logic to checking the subresource would have fixed the bug. Is that not the case? If it does then I think it might be worth adding an additional testcase to demonstrate why we need this extra logic. It would also be worth to double check what other browsers are doing in this scenario.

Source/WebKit/WebProcess/WebPage/WebFrame.cpp
```diff
@@ -1218,7 +1222,7 @@ String WebFrame::provisionalURL() const
     return provisionalDocumentLoader->url().string();
 }
 
-String WebFrame::suggestedFilenameForResourceWithURL(const URL& url) const
+String WebFrame::suggestedFilenameForResourceWithURL(const URL& url, ResourceType type) const
```
sammygill: Maybe `resourceType` for clarity?

Source/WebKit/WebProcess/WebPage/WebFrame.cpp
```diff
@@ -1228,19 +1232,28 @@ String WebFrame::suggestedFilenameForResourceWithURL(const URL& url) const
...
+            if (RefPtr cachedImage = dynamicDowncast<CachedImage>(cached); cachedImage && !cachedImage->errorOccurred())
+                return cachedImage->response().suggestedFilename();
+        }
+    }
+
     return String();
 }
 
-String WebFrame::mimeTypeForResourceWithURL(const URL& url) const
+String WebFrame::mimeTypeForResourceWithURL(const URL& url, ResourceType type) const
```
sammygill: Maybe `resourceType` for clarity?

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
```diff
@@ -758,6 +759,77 @@ static void swizzledOrderFrontColorPanel(id, SEL, id)
     Util::run(&gotProposedMenu);
 }
 
+static RetainPtr<NSString> imageSuggestedFilenameFromCollidingImageURL(HashMap<String, String> imageResponseHeaders)
+{
+    using namespace TestWebKitAPI;
+
+    static constexpr std::array<uint8_t, 67> pngBytes { {
+        0x89, 0x50, 0x4e, 0x47, 0x0d, 0x0a, 0x1a, 0x0a,
```
sammygill: Was a little surprised to see a random stream of bytes for the test! Is there a way we could reuse some existing resources or something along those lines? Maybe for the purposes of this test it isn't a big deal but might be worth taking a look into. Maybe @aprotyas @rr-codes have opinions with regards to this and perhaps other bits of the test?

sammygill (approved):

Tools/TestWebKitAPI/Tests/WebKit/WKWebView/mac/ContextMenuTests.mm
J0pan: switched to use existing png in code base used by another test

J0pan:
Test failures in `win-tests` and `api-gtk` are unrelated.

PR #29, opened 2021-11-01
[webkitscmpy] Reduce autoinstalls on import

JonWBedard:
#### 7c8293e6d4d1879271b238a89607492c4acc6125
<pre>
[webkitscmpy] Reduce autoinstalls on import
<a href="https://bugs.webkit.org/show_bug.cgi?id=232574">https://bugs.webkit.org/show_bug.cgi?id=232574</a>
&lt;rdar://problem/84894275 &gt;

Reviewed by NOBODY (OOPS!).

* Scripts/libraries/webkitscmpy/webkitscmpy/__init__.py: Remove whichcraft (webkitcorepy registers it).
* Scripts/libraries/webkitscmpy/webkitscmpy/local/git.py: Remove unused webkitscmpy imports.
* Scripts/libraries/webkitscmpy/webkitscmpy/local/scm.py: Remote unused webkitcorepy import.
* Scripts/libraries/webkitscmpy/webkitscmpy/program/command.py: Move whichcraft to function.
* Scripts/libraries/webkitscmpy/webkitscmpy/program/setup.py: Move jinja2 to function.
* Scripts/libraries/webkitscmpy/webkitscmpy/remote/svn.py: Move fasteners, xmltodict import to functions.
</pre>

PR #66573, opened 2026-06-05
Text fragment directives beginning with apostrophe can incorrectly match in middle of words.

richwaters:
#### b6bdbd59138434ab6a0263011111ed9e04b8c3ea
<pre>
Text fragment directives beginning with apostrophe can incorrectly match in middle of words.
<a href="https://bugs.webkit.org/show_bug.cgi?id=315937">https://bugs.webkit.org/show_bug.cgi?id=315937</a>

Reviewed by NOBODY (OOPS!).

The current code drops the AtWordStarts find option when creating ICUSearcher if the search target begins with a separator character. The separator check uses a fixed character set that isn&apos;t context aware. This can cause fragment directives beginning with those separators to match substrings within words. The change adds FindOption::IcuOnlyWordBoundaries that is set when resolving fragment directives. When set, the ICUSearcher constructor won&apos;t remove the AtWordStarts constraint. Additionally, isWordStartMatch and isWordEndMatch use ICU&apos;s ubrk_isBoundary directly to check for boundaries.

This change is broader than what is minimally necessary to fix the issue. For text directives, boundary detection always uses the UAX #29 compliant ubrk_isBoundary rather than findNextWordFromIndex, which has differing per-platform implementations. This consolidation of boundary detection has a side effect of removing unnecessary context in generated fragments on iOS.

Lastly, while I believe the ICUSearcher constructor isn&apos;t the correct place to remove AtWordStarts, removing that code altogether would have effects beyond just fragment directives. Specifically, it would change user-facing behavior where &quot;.org&quot; in &quot;webkit.org&quot; is considered to be AtWordStarts (LayoutTests/editing/text-iterator/findString.html). That behavior seems intentional.

Test: http/tests/scroll-to-text-fragment/start-word-bounded-leading-apostrophe.html
Test: http/tests/scroll-to-text-fragment/start-word-bounded-katakana.html

* LayoutTests/http/tests/scroll-to-text-fragment/start-word-bounded-leading-apostrophe-expected.html: Added.
* LayoutTests/http/tests/scroll-to-text-fragment/start-word-bounded-leading-apostrophe.html: Added.
* LayoutTests/http/tests/scroll-to-text-fragment/start-word-bounded-katakana-expected.html: Added.
* LayoutTests/http/tests/scroll-to-text-fragment/start-word-bounded-katakana.html: Added.
* LayoutTests/platform/ios/http/tests/scroll-to-text-fragment/generation-deal-with-newlines-in-directive-expected.txt: Removed.
* Source/WebCore/dom/FragmentDirectiveRangeFinder.cpp:
(WebCore::FragmentDirectiveRangeFinder::findRangeFromNodeList): Set FindOption::IcuOnlyWordBoundaries when matching directives.
* Source/WebCore/editing/CachedMatchFinder.cpp:
(WebCore::matchAffectingOptions): Add FindOption::IcuOnlyWordBoundaries.
(WebCore::CachedMatchFinder::performSearch): Assert IcuOnlyWordBoundaries isn&apos;t used here.
* Source/WebCore/editing/ICUSearcher.cpp:
(WebCore::ICUSearcher::ICUSearcher): Don&apos;t remove AtWordStarts if FindOption::IcuOnlyWordBoundaries is set.
(WebCore::isWordStartMatch): Use ubrk_isBoundary directly for boundary checking when FindOption::IcuOnlyWordBoundaries is set.
(WebCore::isWordEndMatch): Use ubrk_isBoundary directly for boundary checking when FindOption::IcuOnlyWordBoundaries is set.
(WebCore::parseFindOptions): Add FindOption::IcuOnlyWordBoundaries.
* Source/WebCore/editing/FindOptions.h:
</pre>
https://github.com/WebKit/WebKit/commit/b6bdbd59138434ab6a0263011111ed9e04b8c3ea

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/177891 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/51829 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/45623 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/187501 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/141138 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/179754 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/53122 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/51538 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/138664 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/96367 "2 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/180847 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/42193 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/159405 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/119127 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/40856 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/39213 "Passed tests") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/151261 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/38899 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/35962 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/44420 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/43449 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/189930 "Built successfully") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/51496 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/158935 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/147786 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/52178 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/35528 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/147568 "Passed tests") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/42278 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/124430 "Built successfully") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/118861 "The change is no longer eligible for processing.") | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/50686 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/13879 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/50082 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/50322 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/50144 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/editing/CachedMatchFinder.cpp
+++ b/Source/WebCore/editing/CachedMatchFinder.cpp
@@ -42,7 +42,7 @@ namespace WebCore {
 
 static inline FindOptions matchAffectingOptions(FindOptions options)
 {
-    static constexpr OptionSet matchAffectingFlags { FindOption::CaseInsensitive, FindOption::AtWordStarts, FindOption::TreatMedialCapitalAsWordStart, FindOption::AtWordEnds, FindOption::DoNotTraverseFlatTree };
+    static constexpr OptionSet matchAffectingFlags { FindOption::CaseInsensitive, FindOption::AtWordStarts, FindOption::TreatMedialCapitalAsWordStart, FindOption::AtWordEnds, FindOption::DoNotTraverseFlatTree, FindOption::IcuOnlyWordBoundaries };
     return options & matchAffectingFlags;
 }
 
@@ -66,6 +66,9 @@ CachedMatchFinder::CachedMatchFinder(Document& document)
 
 void CachedMatchFinder::performSearch(StringView buffer, unsigned startOffset, const String& target, FindOptions options, NOESCAPE const Function<SearchShouldContinue(size_t, size_t)>& callback)
 {
+    // IcuOnlyWordBoundaries is intended for use with TextIterator/SearchBuffer. It's untested with CachedMatchFinder.
+    ASSERT(!options.contains(FindOption::IcuOnlyWordBoundaries));
+
     if (buffer.isEmpty() || target.isEmpty())
         return;
 
--- a/Source/WebCore/editing/FindOptions.h
+++ b/Source/WebCore/editing/FindOptions.h
@@ -42,6 +42,7 @@ enum class FindOption : uint16_t {
     AtWordEnds = 1 << 7,
     DoNotTraverseFlatTree = 1 << 8,
     DoNotSetSelection = 1 << 9,
+    IcuOnlyWordBoundaries = 1 << 10
 };
 
 using FindOptions = OptionSet<FindOption>;
--- a/Source/WebCore/editing/ICUSearcher.cpp
+++ b/Source/WebCore/editing/ICUSearcher.cpp
@@ -64,11 +64,13 @@ ICUSearcher::ICUSearcher(const String& foldedTarget, FindOptions& options)
     lock();
     // Characters in the separator category never really occur at the beginning of a word,
     // so if the target begins with such a character, we just ignore the AtWordStart option.
-    if (options.contains(FindOption::AtWordStarts) && foldedTarget.length()) {
-        char32_t targetFirstCharacter;
-        U16_GET(foldedTarget, 0, 0u, foldedTarget.length(), targetFirstCharacter);
-        if (isSeparator(targetFirstCharacter))
-            options.remove(FindOption::AtWordStarts);
+    if (options.contains(FindOption::AtWordStarts) && !options.contains(FindOption::IcuOnlyWordBoundaries)) {
+        if (foldedTarget.length()) {
+            char32_t targetFirstCharacter;
+            U16_GET(foldedTarget, 0, 0u, foldedTarget.length(), targetFirstCharacter);
+            if (isSeparator(targetFirstCharacter))
+                options.remove(FindOption::AtWordStarts);
+        }
     }
 
     UCollationStrength strength = options.contains(FindOption::CaseInsensitive) ? UCOL_SECONDARY : UCOL_TERTIARY;
@@ -326,6 +328,13 @@ bool isWordStartMatch(std::span<const char16_t> buffer, size_t start, size_t len
     if (!start)
         return true;
 
+    if (options.contains(FindOption::IcuOnlyWordBoundaries)) {
+        // buffer is capped by SearchBuffer's sliding window or StringImpl::MaxLength
+        RELEASE_ASSERT(buffer.size() <= static_cast<size_t>(std::numeric_limits<int32_t>::max()));
+        auto* iterator = WTF::wordBreakIterator(buffer);
+        return iterator && ubrk_isBoundary(iterator, static_cast<int32_t>(start));
+    }
+
     int size = buffer.size();
     int offset = start;
     char32_t firstCharacter;
@@ -381,7 +390,16 @@ bool isWordEndMatch(std::span<const char16_t> buffer, size_t start, size_t lengt
 {
     ASSERT(length);
     ASSERT(options.contains(FindOption::AtWordEnds));
-    UNUSED_PARAM(options);
+
+    if (options.contains(FindOption::IcuOnlyWordBoundaries)) {
+        size_t end = start + length;
+
+        // buffer is capped by SearchBuffer's sliding window or StringImpl::MaxLength
+        // end == buffer.size is valid offset to ubrk_isBoundary
+        RELEASE_ASSERT(end <= buffer.size() && buffer.size() <= static_cast<size_t>(std::numeric_limits<int32_t>::max()));
+        auto* iterator = WTF::wordBreakIterator(buffer);
+        return iterator && ubrk_isBoundary(iterator, static_cast<int32_t>(end));
+    }
 
     int size = buffer.size();
     int offset = static_cast<int>(start);
--- a/LayoutTests/http/tests/scroll-to-text-fragment/start-word-bounded-katakana-expected.html
+++ b/LayoutTests/http/tests/scroll-to-text-fragment/start-word-bounded-katakana-expected.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+<meta charset="utf-8" />
+<title>Scroll to text fragment - target inside a katakana word must be word-bounded</title>
+<style>
+  span {
+    background-color: pink;
+  }
+</style>
+
+<p>Pass condition: only the katakana for bread on the last line is highlighted; the first line is not highlighted.</p>
+<!-- champagne -->
+<p>シャンパン</p>
+<!-- bread -->
+<p><span>パン</span></p>
--- a/LayoutTests/http/tests/scroll-to-text-fragment/start-word-bounded-katakana.html
+++ b/LayoutTests/http/tests/scroll-to-text-fragment/start-word-bounded-katakana.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<meta charset="utf-8" />
+<title>Scroll to text fragment - target inside a katakana word must be word-bounded</title>
+<link rel="help" href="https://wicg.github.io/scroll-to-text-fragment/">
+<meta name="assert" content="This test checks that a fragment directive target which appears as a substring inside a katakana word is rejected.">
+<meta name="fuzzy" content="maxDifference=0-2; totalPixels=0-1500" />
+<style>
+  ::target-text { background-color: pink; }
+</style>
+
+<p>Pass condition: only the katakana for bread on the last line is highlighted; the first line is not highlighted.</p>
+<!-- champagne -->
+<p>シャンパン</p>
+<!-- bread -->
+<p>パン</p>
+
+<script>
+  location.href = "#:~:text=%E3%83%91%E3%83%B3";
+</script>
+</html>
--- a/LayoutTests/http/tests/scroll-to-text-fragment/start-word-bounded-leading-apostrophe-expected.html
+++ b/LayoutTests/http/tests/scroll-to-text-fragment/start-word-bounded-leading-apostrophe-expected.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<meta charset="utf-8" />
+<title>Scroll to text fragment - target starting with apostrophe must be word-bounded on the left</title>
+<style>
+  span {
+    background-color: pink;
+  }
+</style>
+<p>Pass condition: only the target in the final sentence below is highlighted; nothing in the earlier sentence is highlighted.</p>
+<p>Anyway, it's our only chance.</p>
+<p>And this <span>'s</span> our last time together.</p>
--- a/LayoutTests/http/tests/scroll-to-text-fragment/start-word-bounded-leading-apostrophe.html
+++ b/LayoutTests/http/tests/scroll-to-text-fragment/start-word-bounded-leading-apostrophe.html
@@ -1,0 +1,16 @@
+<!DOCTYPE html>
+<meta charset="utf-8" />
+<title>Scroll to text fragment - target starting with apostrophe must be word-bounded on the left</title>
+<link rel="help" href="https://wicg.github.io/scroll-to-text-fragment/">
+<meta name="assert" content="A text directive target beginning with an apostrophe must only match when its start is at a word boundary. The directive should resolve to the standalone apostrophe-s in the final sentence, not the apostrophe-s inside the word it's earlier in the document.">
+<meta name="fuzzy" content="maxDifference=0-2; totalPixels=0-1500" />
+<style>
+  ::target-text { background-color: pink; }
+</style>
+<p>Pass condition: only the target in the final sentence below is highlighted; nothing in the earlier sentence is highlighted.</p>
+<p>Anyway, it's our only chance.</p>
+<p>And this 's our last time together.</p>
+<script>
+  location.href = "#:~:text=%27s,-our";
+</script>
+</html>
--- a/LayoutTests/platform/ios/http/tests/scroll-to-text-fragment/generation-deal-with-newlines-in-directive-expected.txt
+++ b/LayoutTests/platform/ios/http/tests/scroll-to-text-fragment/generation-deal-with-newlines-in-directive-expected.txt
@@ -1,1 +1,0 @@
-:~:text=it.-,This%20is%20the%20text%20without%20a%20space%20fakecode.h%20and%20the%20end%20of%20the%20selecton.,-Additional
--- a/Source/WebCore/dom/FragmentDirectiveRangeFinder.cpp
+++ b/Source/WebCore/dom/FragmentDirectiveRangeFinder.cpp
@@ -157,6 +157,7 @@ static std::optional<SimpleRange> findRangeFromNodeList(const String& query, con
         options.add(FindOption::AtWordStarts);
     if (wordEndBounded == WordBounded::Yes)
         options.add(FindOption::AtWordEnds);
+    options.add(FindOption::IcuOnlyWordBoundaries);
 
     auto foundText = findPlainText(searchRange, foldedQuery, options);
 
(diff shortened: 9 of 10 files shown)


richwaters:
@rniwa  @cdumez  Just checking whether either of you has had a chance to look at this. I took a quick look at the EWS failures and they appear unrelated to this change. Thanks!

richwaters:
@rniwa  @cdumez  I've rebased to main. Would @darinadler be an appropriate reviewer for this as well, since he reviewed my previous PR in this area?

richwaters:
@rniwa  @cdumez  @darinadler  Just checking back in on this PR that fixes text fragment word boundary handling. I rebased and cleaned up ICUSearcher::isWordStartMatch to just have a single branch for StrictWordBoundaries. I also added a null guard before calling ubrk_isBoundary. I had hoped to get all the tests to green, but it seems like the windows tests have been stuck at "Unable to build WebKit without PR, retrying build" for a few days now.

Source/WebCore/editing/CachedMatchFinder.cpp
darinadler: Putting just this one new flag on a separate line doesn’t seem elegant. Either stick with a single long line or move the entire thing to a vertical format, please.

Source/WebCore/editing/ICUSearcher.cpp
darinadler: The line break here doesn’t seem elegant. I suggest we put all the checks on a single long line or find another way to make this a more conventional format where the `if` condition doesn’t line up with the body of the `if` statement, maybe even using a named boolean local.

It also makes slightly more logical sense to me to check the flags in `options` before the `foldedTarget.length()` guard. That guard is more of a mechanical thing more directly related to range checking for `U16_GET`, and I think a way to deal with that would be to break it out into a separate nested `if` statement.

darinadler: I suggest we add a local variable called `adjustedEnd`, because `adjustedStart + length` appears three different places below now.

Source/WebCore/editing/FindOptions.h
darinadler: I am not sure that this name speaks for itself.

I accept your decision that we need word boundaries based on ICU for places where interoperability and consistency with other web browsers is more important than meeting user’s expectations for what can be found. On the other hand, I think that at least for now, in "find" user interface features we need word boundaries that may differ, say, based on the language the user speaks, their native locale, and cultural understanding of what words are.

But I am not sure that “strict” is the best name for “ICU’s version” vs. “a local version designed to cater to a particular user’s expectations”. I’m not sure why it’s OK to use a text break definition that is based on the user locale. It seems we will have interoperability problems if it is different for different users. So this may not go far enough.

I don’t think it’s right that using `ubrk_isBoundary` gives “straightforward spec compliance” when the ICU iterator was based on the end users’s locale. Where is the specification so I can look at that?

Source/WebCore/editing/ICUSearcher.cpp
darinadler: Due to the risks of re-entrantly using this while there is a word break iterator already in use for another string, the code says we should use CachedTextBreakIterator instead. What guarantees this won’t have that kind of problem?

I think we can consider using `auto*` here instead of `UBreakIterator*` and the word `iterator` instead of the abbreviation `it`.

darinadler: What obviates the need for `FontCascade::isCJKIdeographOrSymbol` here? Do we have test cases that cover that case?

I suggest we assert that `start - contextOffset` fits in an `int32_t` without overflow and also write a comment about why that’s guaranteed, which presumably has something to do with the maximum length of a `StringView`.

darinadler: Due to the risks of re-entrantly using this while there is a word break iterator already in use for another string, the code says we should use CachedTextBreakIterator instead. What guarantees this won’t have that kind of problem?

darinadler: I suggest we assert that `adjustedEnd` fits in an `int32_t` without overflow even though it’s a `size_t` also write a comment about why that’s guaranteed, which presumably has something to do with the maximum length of a `StringView`.

It’s not great that this function uses both `int32_t` and `int` for the same data type, in both cases as part of interfacing with ICU. Would be better to be consistent. See `endWord` below.

I think we can consider using `auto*` here instead of `UBreakIterator*` and the word `iterator` instead of the abbreviation `it`.

darinadler (approved):
Seems OK, but need improvement.

richwaters:
Thanks for the detailed review. It's much appreciated. You make a lot of good points, especially in regards to some bigger issues. "This may not go far enough" captures the locale issue nicely. 

For my part, I was mostly focused on getting the issues with text directives corrected, making use of the existing machinery and not touching the alternate path. I've narrowed the focus a bit and renamed the option to IcuOnlyWordBoundaries and added an assert to guard against use in the more general case. I've also taken a closer look at some of that existing machinery both to determine the feasibility of addressing some of your comments and to evaluate what is necessary for the IcuOnlyWordBoundaries branch. My inline replies address the specifics, but the underlying concept is that when the option is set, ICU alone makes the boundary decisions without shortcuts.

Also, I rebased onto main to pick up relevant ICUSearcher changes from https://github.com/WebKit/WebKit/pull/70054. Apologies for the extra compare noise.

Source/WebCore/editing/CachedMatchFinder.cpp
richwaters: Right. Sorry about that. I personally prefer single long lines. This is corrected.

Source/WebCore/editing/ICUSearcher.cpp
richwaters: I made the 'nested if' change and removed extra linebreaks.

richwaters: > I suggest we add a local variable called adjustedEnd, because adjustedStart + length appears three different places below now.

This is no longer applicable as the IcuOnlyWordBoundaries branch in both isWordStartMatch & isWordEndMatch sends buffer in full to ICU without adjustment by either ```extractSubspanIncludingContextNeededForDictionaryBasedWordBreak``` or the newer ```boundedWordContextForNonComplexText```. The reason for the latter is described in PR 70054:

```
The slowdown is caused by passing the entire buffer to findNextWordFromIndex
from TextBoundaries.h on every match. Note that this only happens when the
match does not require context for word boundaries. On mac,
findNextWordFromIndex copies the buffer to a new string.

To speed this up, we limit the size of the buffer passed to findNextWordFromIndex.
Instead of the whole buffer, we send a window that includes the match. This way,
the copy can be much faster.
```

The IcuOnlyWordBoundaries branch does not call findNextWordFromIndex, it does not copy the buffer to a new string, and the buffer is already windowed by the TextIterator/SearchBuffer path used for text directives. 

```extractSubspanIncludingContextNeededForDictionaryBasedWordBreak``` is a more interesting case. From  

https://github.com/WebKit/WebKit/pull/60036 there is this explanation of why it was introduced:

```
This difference has exposed an issue in WebCore::findNextWordFromIndex when
finding word boundaries for context-sensitive words. The macOS version of
findNextWordFromIndex uses a dictionary based approach, and for some reason,
many unrelated characters followed by a context-sensitive word results in
inconsistent word boundaries.

To mitigate this issue, I introduced a function:
extractSubspanIncludingContextNeededForDictionaryBasedWordBreak, which will
limit the context surrounding context-sensitive words, to ensure consistent
word boundaries.
```

Again, findNextWordFromIndex is not called on the IcuOnlyWordBoundaries path. Aside from that, ```extractSubspanIncludingContextNeededForDictionaryBasedWordBreak``` is built on top of ```requiresContextForWordBoundary``` which has this comment near the top:  

```// FIXME: This function won't be needed when https://bugs.webkit.org/show_bug.cgi?id=120656 is fixed.```

I've been unable to find an example or test case for the "many unrelated characters followed by a context-sensitive word" issue. If there is some more information about this in the radar filing or another source that I don't have access to, I'd be interested in seeing that. Without that information, I think it's best to remove the call and fall back to the concept of "icu only".

File: Source/WebCore/editing/FindOptions.h
Comment (richwaters) [reply Source/WebCore/editing/FindOptions.h]: > I am not sure that this name speaks for itself.

Yes. I was thinking in terms of text directives and it doesn't hold up at the lower level. I've renamed to IcuOnlyWordBoundaries. Another option is "Uax29WordBoundaries" or if we want to get specific, "TextDirectiveWordBoundaries".


> I don’t think it’s right that using ubrk_isBoundary gives “straightforward spec compliance” when the ICU iterator was based on the end users’s locale. Where is the specification so I can look at that?

Yes. I overstated that "straightforward spec compliance". I removed it from the commit message. You're right about the locale thing. It should come from the document and not the user preferences. 

This is the section of the "URL Fragment Text Directives" spec that speaks to word boundaries:

• https://wicg.github.io/scroll-to-text-fragment/#word-boundaries

Near the top of the section it talks of a startLocale and endLocale. Then in the "3.6.1 Finding Ranges in a Document Section" it says:

> If wordStartBounded is true and matchIndex is not at a word boundary in searchBuffer, given the language from start’s node as the locale; or wordEndBounded is true and matchIndex + queryString’s length is not at a word boundary in searchBuffer, given the language from end’s node as the locale:


So, for full compliance we'd need to obtain the locale from the specific portion of the document being checked. I'm not sure that's feasible within the current architecture (it might be better approached with CachedMatchedFinder), and I think that's beyond the scope of this PR. That said, it might be possible to use the top level document language or leave it empty. For now, I haven't addressed this. I believe the user-based locale is a pre-existing issue for the non USE(APPKIT) paths. For AppKit, it's not apparent how the NSAttributedString segmentation handles locale.

File: Source/WebCore/editing/ICUSearcher.cpp
Comment (richwaters) [reply Source/WebCore/editing/ICUSearcher.cpp]: Nothing. It could be safe if all callers use it in the way the IcuOnlyWordBoundaries path does in that it gets the iterator, uses it and is done with it before returning. Also, all callers would need to be on the main thread. I'm not at all sure that all the callers do that. 

Unfortunately, CachedTextBreakIterator does not have a Word or WordBoundary mode in TextBreakIteratorICU like it does in TextBreakIteratorCF. Using CachedTextBreakIterator would require adding that mode. It looks fairly straightforward (famous last words) to do that, but that might be beyond the scope of this PR. I believe the use of the wordBreakIterator in this fashion is a pre-existing issue that appears in various places, including the fallback in FragmentDirectiveRangeFinder::findRangeFromNodeList that calls indexIsAtWordBoundary. I don't think there is any new risk added with this patch.


> I think we can consider using auto* here instead of UBreakIterator* and the word iterator instead of the abbreviation it.

Done.

Comment (richwaters) [reply Source/WebCore/editing/ICUSearcher.cpp]: > What obviates the need for FontCascade::isCJKIdeographOrSymbol here? Do we have test cases that cover that case?

It looks like this was introduced with: https://bugs.webkit.org/show_bug.cgi?id=60730 and commit: https://github.com/WebKit/WebKit/commit/201c523eb0cda4a9961231497fc9d5d7e120e4ba with test cases in findString.html:

```
await testFindString("\u8d77\u52d5\u6226\u58eb", "\u52d5\u6226\u58eb", ["AtWordStarts"], [[1, 4], []]);
await testFindString("\u8d77\u52d5\u6226\u58eb", "\u6226\u58eb", ["AtWordStarts"], [[2, 4], []]);
await testFindString("\u8d77\u52d5\u6226\u58eb", "\u58eb", ["AtWordStarts"], [[3, 4], []]);
```

They all search 起動戦士 and expect to find word start at each position. As far as I can tell from llms and online translators there are only 2 words in that string: 起動 (startup) & 戦士 (warrior). With isCJKIdeographOrSymbol dropped, only offset 2 is considered a word start. I'm not qualified to assess whether that's correct, but I'll fallback on the underlying concept of "icu only" mentioned at the top. 

I've also added a reference test showing the downside of that check, which is a text directive matching one word パン ("bread") inside another word シャンパン ("champagne").

> I suggest we assert that start - contextOffset fits in an int32_t without overflow and also write a comment about why that’s guaranteed, which presumably has something to do with the maximum length of a StringView.

There is no longer the contextOffset subtraction for reasons described in this prior response: https://github.com/WebKit/WebKit/pull/66573#discussion_r3694405541. An assert has been added for buffer.size(), but it looks like I missed the one for start. I'll include that with the next round of changes.

Comment (richwaters) [reply Source/WebCore/editing/ICUSearcher.cpp]: This is the mirror of the isWordStartMatch issue addressed above: https://github.com/WebKit/WebKit/pull/66573#discussion_r3694460051. The same issues apply. It can be fixed by adding a word mode to CachedTextBreakIterator, but I don't believe the usage in this patch introduces a new risk.

Comment (richwaters) [reply Source/WebCore/editing/ICUSearcher.cpp]: > I suggest we assert that adjustedEnd fits in an int32_t without overflow even though it’s a size_t also write a comment about why that’s guaranteed, which presumably has something to do with the maximum length of a StringView.

The var is now just 'end' since there is no longer an adjustment applied and the entire buffer is now sent to ubrk_isBoundary. The rationale for that is described in a prior reply: https://github.com/WebKit/WebKit/pull/66573#discussion_r3694405541. I've added an  assert and comment.

> It’s not great that this function uses both int32_t and int for the same data type, in both cases as part of interfacing with ICU. Would be better to be consistent. See endWord below.

For this iteration, I'd like to push back on this a bit. The ubrk_isBoundary declaration specifies int32_t, which is the type used elsewhere in this file by functions that directly call ICU (setPattern, setText, etc.). The int parameter to findEndWordBoundary comes from its own declaration. findEndWordBoundary calls findWordBoundary and, at least for the APPKIT path, there is no visible interaction with ICU:

```
void findWordBoundary(StringView text, int position, int* start, int* end)
{
#if USE(APPKIT)
    auto attributedString = adoptNS([[NSAttributedString alloc] initWithString:text.createNSStringWithoutCopying().get()]);
    NSRange range = [attributedString doubleClickAtIndex:std::min<unsigned>(position, text.length() - 1)];
    *start = range.location;
    *end = range.location + range.length;
#else
```

> I think we can consider using auto* here instead of UBreakIterator* and the word iterator instead of the abbreviation it.

Done..

PR #66581, opened 2026-06-05
[CMake] Set SDK based on CMAKE_OSX_SYSROOT and pin compiler tools using the associated toolchain

emw-apple:
#### 70d5a8f6fd34f7d1282fc49e85b212d4daaddd33
<pre>
[CMake] Set SDK based on CMAKE_OSX_SYSROOT and pin compiler tools using the associated toolchain
<a href="https://bugs.webkit.org/show_bug.cgi?id=316442">https://bugs.webkit.org/show_bug.cgi?id=316442</a>
<a href="https://rdar.apple.com/178844557">rdar://178844557</a>

Reviewed by Geoffrey Garen and Keith Miller.

CMAKE_OSX_SYSROOT is the built-in variable that controls which SDK is
being used on Apple platforms (misleadingly named, as it works for *any*
Xcode SDK). Clean up some of our logic that resolves and sets
SDK-related things to use this variable. The end result is that one can
customize the active SDK at configure time like:

   cmake --preset mac-release -DCMAKE_OSX_SYSROOT=macosx.internal

That SDK will be used to configure the build, and pin compiler tools for
ninja to use.

Restructure xcrun code to run before the initial call to project(),
since that&apos;s when the build starts performing compiler checks. Also,
read from the SDK&apos;s SDKSettings.json for more discrete information. This
will be used next to get information about the deployment target and
default architectures.

Internal SDK support (for Mac) is not complete. For example, local
WebKitAdditions is not yet supported. But this is enough to get JSC
building on recent internal SDKs.

* CMakeLists.txt:
* Source/WTF/wtf/PlatformCocoa.cmake: Use the correct Mig binary and SDK
  root. This reveals an error on the iOS build. Canonically, we do not build
  MachExceptions.defs on public embedded SDKs. Add a condition to
  reflect this.
* Source/WTF/wtf/PlatformJSCOnly.cmake: Ditto.
* Source/cmake/OptionsIOS.cmake:
* Source/cmake/WebKitXcodeSDK.cmake: Rename of WebKitXcrun, now runs SDK
  initialization logic.
* Source/cmake/WebKitXcrun.cmake: Removed.

Canonical link: <a href="https://commits.webkit.org/314875@main">https://commits.webkit.org/314875@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/6522f419dc26bd82c8b10d0b7c313dd36b94eba0

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/167507 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/41002 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/34597 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/176560 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/125188 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/41438 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/41016 "Built successfully") | [  ~~🧪 wpe-wk2~~](https://ews-build.webkit.org/#/builders/34/builds/130312 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/125188 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/170466 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/162/builds/32725 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/150501 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/110829 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/31708 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/30405 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/25295 "Built successfully") | | 
| [  ~~🛠 🧪 jsc-x86-64~~](https://ews-build.webkit.org/#/builders/20/builds/159761 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/141695 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/28196 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/179101 "Built successfully") | | 
| [  ~~🛠 🧪 jsc-debug-arm64~~](https://ews-build.webkit.org/#/builders/171/builds/28484 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/31996 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/29914 "Passed tests") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/1/builds/138708 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/41076 "Built successfully") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/34562 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/138595 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/41764 "Built successfully") | | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/104878 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/25487 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/33852 "Passed tests") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/26867 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/200184 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/40268 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/113349 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/53803 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/39665 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/4966 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/39916 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/39810 "Built successfully") | | | |


CMakeLists.txt
```diff
@@ -66,32 +67,42 @@ if (PORT STREQUAL "IOS")
...
-        OUTPUT_VARIABLE _macos_sysroot
-        OUTPUT_STRIP_TRAILING_WHITESPACE
-        RESULT_VARIABLE _xcrun_result
-        ERROR_QUIET)
-    if (_xcrun_result EQUAL 0 AND _macos_sysroot)
-        set(CMAKE_OSX_SYSROOT "${_macos_sysroot}" CACHE PATH "macOS SDK path" FORCE)
+if (CMAKE_HOST_SYSTEM_NAME STREQUAL "Darwin")
+    include(WebKitXcrun)
+
+    if (CMAKE_OSX_SYSROOT)
```
geoffreygaren: Do we need to care about CMAKE_OSX_SYSROOT being empty by default in newer CMakes?

CMakeLists.txt
```diff
@@ -66,32 +67,42 @@ if (PORT STREQUAL "IOS")
...
+    else ()
+        message(FATAL_ERROR "Building for an Apple platform without an SDK "
+            "directory (CMAKE_OSX_SYSROOT) or a hard-coded PORT provided.")
     endif ()
-    unset(_macos_sysroot)
-    unset(_xcrun_result)
+
+    # Subsequent use of `xcrun` or any of the system Xcode and BSD tools will
+    # use the selected SDK and toolchain.
+    set($ENV{SDKROOT} ${CMAKE_OSX_SYSROOT})
```
kmiller68: I think this needs to be `ENV{SDKROOT}`? The `$` expands `$ENV{SDKROOT}` and assigns to the value IIUC?

CMakeLists.txt
```diff
@@ -7,6 +7,7 @@
 # and loading the cross-compilation settings from CMAKE_TOOLCHAIN_FILE.
 
 cmake_minimum_required(VERSION 3.20)
+set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/Source/cmake")
 
 if (PORT STREQUAL "IOS")
```
kmiller68, line 17: Can we kill all this below?

CMakeLists.txt
```diff
@@ -66,32 +67,42 @@ if (PORT STREQUAL "IOS")
...
+
+    if (CMAKE_OSX_SYSROOT)
+        WEBKIT_RESOLVE_SDK(${CMAKE_OSX_SYSROOT})
+    elseif (PORT STREQUAL "Mac")
+        # FIXME: Support macosx.internal
+        WEBKIT_RESOLVE_SDK(macosx)
+    elseif (PORT STREQUAL "IOS" and CMAKE_IOS_SIMULATOR)
+        WEBKIT_RESOLVE_SDK(iphoneos.internal iphoneos)
+    elseif (PORT STREQUAL "IOS" and NOT CMAKE_IOS_SIMULATOR)
+        WEBKIT_RESOLVE_SDK(iphonesimulator.internal iphonesimulator)
```
kmiller68: These seem backwards?

Source/cmake/WebKitXcrun.cmake
```diff
@@ -12,7 +12,7 @@ function(WEBKIT_RESOLVE_SDK)
             OUTPUT_STRIP_TRAILING_WHITESPACE
             ERROR_QUIET)
         if (_sdk_result EQUAL 0 AND _sdk_version)
-            set(WEBKIT_SDK "${_sdk}" PARENT_SCOPE)
+            set(CMAKE_OSX_SYSROOT "${_sdk}" PARENT_SCOPE)
```
kmiller68: This seems wrong since it seems like other places are expecting a path not a name? e.g.

```
            if (EXISTS "${CMAKE_OSX_SYSROOT}/usr/local/include/unicode_private.modulemap")
                target_compile_options(${_target} PRIVATE
                    "$<$<COMPILE_LANGUAGE:Swift>:SHELL:-Xcc -isystem${CMAKE_OSX_SYSROOT}/usr/local/include>"
                    "$<$<COMPILE_LANGUAGE:Swift>:SHELL:-Xcc -fmodule-map-file=${CMAKE_OSX_SYSROOT}/usr/local/include/unicode_private.modulemap>")
            endif ()
```

CMakeLists.txt
```diff
@@ -66,32 +67,42 @@ if (PORT STREQUAL "IOS")
...
+    set($ENV{SDKROOT} ${CMAKE_OSX_SYSROOT})
+
+    WEBKIT_RESOLVE_TOOL(CMAKE_C_COMPILER "clang")
+    WEBKIT_RESOLVE_TOOL(CMAKE_ASM_COMPILER "clang")
+    WEBKIT_RESOLVE_TOOL(CMAKE_CXX_COMPILER "clang++")
+    WEBKIT_RESOLVE_TOOL(CMAKE_OBJC_COMPILER "clang")
+    WEBKIT_RESOLVE_TOOL(CMAKE_OBJCXX_COMPILER "clang++")
+    WEBKIT_RESOLVE_TOOL(CMAKE_Swift_COMPILER "swiftc")
+    WEBKIT_RESOLVE_TOOL(CMAKE_INSTALL_NAME_TOOL "install_name_tool")
+    WEBKIT_RESOLVE_TOOL(GPERF_EXECUTABLE "gperf")
```
kmiller68: Isn't this missing a bunch of CMake tools? e.g. CMAKE_LINKER, CMAKE_LIBTOOL I believe we use those. So those will default to whatever the current `xcode-select` says?

emw-apple: Maybe. I looked at every variable starting with `/usr/bin` in the CMakeCache, but will figure out how it's invoking the linker tools and a few others.

CMakeLists.txt
emw-apple: Good catch, thanks!

CMakeLists.txt
emw-apple: I don't think so. As written, it's an error for _neither_ CMAKE_OSX_SYSROOT nor PORT to be defined. If you only provide a port name, we pick a sensible default SDK.

Source/cmake/WebKitXcrun.cmake
emw-apple: Ah, you're right. Latest revision resolves whatever SDK-like thing we pass in to the actual path first, via `xcrun --show-sdk-path`.

Source/cmake/WebKitXcodeSDK.cmake
```diff
@@ -0,0 +1,91 @@
...
+function(WEBKIT_RESOLVE_SDK)
+    foreach (_sdk IN LISTS ARGN)
+        execute_process(COMMAND xcrun --sdk ${_sdk} --show-sdk-path
+            OUTPUT_VARIABLE _sdk_path
+            RESULT_VARIABLE _sdk_result
+            OUTPUT_STRIP_TRAILING_WHITESPACE
+            ERROR_QUIET)
+        if (_sdk_result EQUAL 0 AND _sdk_path)
+            file(READ "${_sdk_path}/SDKSettings.json" _sdk_settings)
+            string(JSON _sdk_version GET ${_sdk_settings} Version)
```
geoffreygaren, line 16: Can we use `execute_process(COMMAND xcrun --sdk ${_sdk} --show-sdk-version` instead, to be a little more forwards-compatible?

Source/cmake/WebKitXcodeSDK.cmake
```diff
@@ -0,0 +1,91 @@
...
+            set(_sdk_version "${_sdk_version}" PARENT_SCOPE)
+            string(JSON _sdk_canonical_name GET ${_sdk_settings} CanonicalName)
+
+            message(STATUS "Xcode SDK: ${_sdk_canonical_name} at ${_sdk_path}")
+            set(CACHE{CMAKE_OSX_SYSROOT} TYPE PATH FORCE VALUE ${_sdk_path})
```
geoffreygaren, line 21: I think it's incorrect to set this as a cached value. When `ninja` notices a change to CMake files, it blows away the cache, and this value is lost. But our build config hard depends on this value, so we're risking a mystery torn build. I think we need to set a standard value instead.

Source/cmake/WebKitXcodeSDK.cmake
```diff
@@ -0,0 +1,91 @@
...
+    # FIXME: Support macosx.internal.
+    WEBKIT_RESOLVE_SDK(macosx)
+elseif (PORT STREQUAL "IOS" AND CMAKE_IOS_SIMULATOR)
+    WEBKIT_RESOLVE_SDK(iphonesimulator.internal iphonesimulator)
+elseif (PORT STREQUAL "IOS" AND NOT CMAKE_IOS_SIMULATOR)
+    WEBKIT_RESOLVE_SDK(iphoneos.internal iphoneos)
+else ()
+    message(FATAL_ERROR "Building for an Apple platform without an SDK "
+        "directory (CMAKE_OSX_SYSROOT) or a hard-coded PORT provided.")
+endif ()
```
geoffreygaren, line 77: What's the use case for inferring CMAKE_OSX_SYSROOT when it isn't set? My mental model is that 99% of the time a preset will set CMAKE_OSX_SYSROOT, and 1% of the time a niche use case will set this value explicitly on the command line. 

If my mental model is correct, then implicitly picking an SDK is either dead code or a potential surprise.

geoffreygaren (approved):
This looks like the right approach but I think it needs some cleanup before landing (or right after landing).

Source/cmake/WebKitXcodeSDK.cmake
```diff
@@ -0,0 +1,91 @@
...
+    set(${OUTPUT_VAR} "${_out}" PARENT_SCOPE)
+endfunction()
+
+function(WEBKIT_RESOLVE_TOOL OUTPUT_VAR _tool)
+    WEBKIT_XCRUN(_path -f ${_tool})
+    if (NOT EXISTS "${_path}")
+        message(SEND_ERROR "Cannot find ${_tool} in the active SDK and "
+            "toolchain (${CMAKE_OSX_SYSROOT})")
+    else ()
+        set(${OUTPUT_VAR} "${_path}" PARENT_SCOPE)
```
kmiller68: Do we want these to be `PARENT_SCOPE` or `CACHE`? Seems like once a tool is resolved it shouldn't need to be resolved again? Do we expect the build to work when changing SDKs without blowing away the build directory?

I expect `CACHE`ing makes subsequent reconfigures faster if we check for `if (EXISTS ${OUTPUT_VAR}) return()`. We can investigate in a follow up. Worth a FIXME though.

Source/cmake/WebKitXcodeSDK.cmake
```diff
@@ -0,0 +1,91 @@
...
+    message(FATAL_ERROR "xcrun could not locate any SDK in: ${ARGN}")
+endfunction()
+
+# Runs `xcrun --sdk ${CMAKE_OSX_SYSROOT} <args>` and capture stdout in OUTPUT_VAR
+# (in the caller's scope).
+function(WEBKIT_XCRUN OUTPUT_VAR)
+    if (NOT CMAKE_OSX_SYSROOT)
+        message(FATAL_ERROR "WEBKIT_XCRUN called before WEBKIT_RESOLVE_SDK")
+    endif ()
+    execute_process(COMMAND xcrun --sdk ${CMAKE_OSX_SYSROOT} ${ARGN}
```
kmiller68: If we're relying on `ENV{SDKROOT}` why do we need the `--sdk ${CMAKE_OSX_SYSROOT}` for calls here? Seems like either we should be using the CMAKE_OSX_SYSROOT or SDKROOT consistently.

kmiller68 (approved):
r=me with two comments

Source/cmake/WebKitXcodeSDK.cmake
emw-apple: We could, although there are some other things I'd like to use SDKSettings.json for. Most notably, discovering which architectures an SDK can build for, so we can avoid some hard-coding.

Clang itself parses SDKSettings.json, so while the file is not formally documented I think it's reasonably stable for us.

Source/cmake/WebKitXcodeSDK.cmake
emw-apple: Only to match existing behavior, where you can run e.g. `cmake -DPORT=Mac` and get a reasonable SDK. The current presets do this, but I support cleaning them up to specify the SDK up front.

(Longer term I'd like to get us to a point where there is a single PORT for Apple platform builds, and the SDK is the main differentiator.)

Source/cmake/WebKitXcodeSDK.cmake
emw-apple: Hm, not sure I understand. IIUC, Ninja does not delete cached values, it re-runs cmake with the existing cache (rules.ninja shows it literally running `cmake --regenerate-during-build -B ... -S ...`. That would cause CMake to enter this section of code again, read the previously configured CMAKE_OSX_SYSROOT, and re-resolve it. 

(Maybe the re-resolution is unnecessary, but it's quite fast, and it's needed to support detecting a nonsensical configuration such as the `Mac` port with an iOS SDK.)

emw-apple: Also, I set it as a cached variable because I noticed that CMake's own code was detecting an unset CMAKE_OSX_SYSROOT and overshadowing our normal variable. From [their docs](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html), it sounds like we need to either put it in the cache or adopt a policy:

> CMAKE_OSX_SYSROOT 
> …
> This variable should be set as a CACHE entry (or else CMake may remove it while initializing a cache entry of the same name) unless policy [CMP0126](https://cmake.org/cmake/help/latest/policy/CMP0126.html#policy:CMP0126) is set to NEW.

Source/cmake/WebKitXcodeSDK.cmake
emw-apple: Caching makes sense, I'll do that. It's probably correct to do this to prevent other bits of CMake logic from overwriting some of the well-known variables such as CMAKE_CXX_COMPILER that we set here.

PR #66617, opened 2026-06-06
[CMake] Enable the internal-SDK build for macOS ports

zakariaridouh:
#### 8638d4640f4a5143f71aa5137b8328ea08fb5a6f
<pre>
[CMake] Enable the internal-SDK build for macOS ports
<a href="https://bugs.webkit.org/show_bug.cgi?id=316479">https://bugs.webkit.org/show_bug.cgi?id=316479</a>
<a href="https://rdar.apple.com/178878233">rdar://178878233</a>

Reviewed by NOBODY (OOPS!).

Get the macOS CMake port building against the internal SDK. This is scoped
to that configuration -- iOS, the public SDK, the non-Cocoa ports, and the
Xcode build are unaffected.

- Resolve the SDK as macosx.internal with a macosx fallback, kept in
  lockstep between the pre-project block in CMakeLists.txt and OptionsMac.
- Enable OBJC/OBJCXX so .mm sources get the OBJCXX rules, flags, and PCH.
- Add the internal SDK&apos;s PrivateFrameworks and usr/local/include search
  paths, define USE_APPLE_INTERNAL_SDK for the Swift clang importer, and
  mask TextInput_Private&apos;s modulemap for explicit-module Swift builds.
- Generate the AppleFeatures.h stub only when no real header is reachable
  on the host or in the SDK, so the internal SDK&apos;s header wins.
- Order just-built frameworks ahead of the SDK&apos;s PrivateFrameworks for all
  Apple builds (previously iOS only).
- Inline WebKitAdditions into marker-bearing public headers via
  replace-webkit-additions-includes.py, matching Xcode.

The header replacement and a WebExtension unification fix live in shared
CMake files, so both are gated to CMAKE_SYSTEM_NAME STREQUAL &quot;Darwin&quot; to
leave iOS untouched. A few Cocoa sources don&apos;t survive the build&apos;s unified
bundles, so de-unify them (@no-unify) and route the WebExtension Cocoa
directory through the WebKitARC subtarget. Since this build doesn&apos;t compile
the Swift UIProcess interop, pin ENABLE_IPC_TESTING_SWIFT off and gate
WKTextSelectionController behind a new HAVE(WK_TEXT_SELECTION_CONTROLLER)
capability instead of a build-system #if in the source.

Misc correctness:
- WebPageProxyMac: pass NSApp to -stopSpeaking: (-Wnonnull).
- WebViewImpl.h: forward-declare InteractionInformationAtPosition for
  standalone compilation.
- SequesteredImmortalHeap: export installScavenger()/scavengeImpl().

* CMakeLists.txt:
* Source/WTF/wtf/PlatformHave.h:
* Source/WTF/wtf/SequesteredImmortalHeap.h:
* Source/WebCore/PlatformMac.cmake:
* Source/WebCore/SourcesCocoa.txt:
* Source/WebKit/PlatformCocoa.cmake:
* Source/WebKit/SourcesCocoa.txt:
* Source/WebKit/UIProcess/mac/WebPageProxyMac.mm:
(WebKit::WebPageProxy::stopSpeaking):
* Source/WebKit/UIProcess/mac/WebViewImpl.h:
* Source/WebKit/UIProcess/mac/WebViewImpl.mm:
(WebKit::m_flagsChangedEventMonitorTrackingArea):
* Source/cmake/OptionsCocoa.cmake:
* Tools/TestWebKitAPI/PlatformMac.cmake:
* Tools/TestWebKitAPI/SourcesCocoa.txt:
* Tools/TestWebKitAPI/SourcesMac.txt:
</pre>
https://github.com/WebKit/WebKit/commit/8638d4640f4a5143f71aa5137b8328ea08fb5a6f

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/166905 "Passed style check") | [❌ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/40395 "Hash 8638d464 for PR 66617 does not build (failure)") | [❌ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/33976 "Hash 8638d464 for PR 66617 does not build (failure)") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/175953 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/124163 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/168771 "Passed tests") | [❌ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/40828 "Hash 8638d464 for PR 66617 does not build (failure)") | [❌ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/40403 "Hash 8638d464 for PR 66617 does not build (failure)") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/129790 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/91396 "1 flakes 5 failures") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/169864 "Passed tests") | [❌ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/155/builds/40828 "Hash 8638d464 for PR 66617 does not build (failure)") | [❌ 🧪 api-mac](https://ews-build.webkit.org/#/builders/168/builds/33976 "Hash 8638d464 for PR 66617 does not build (failure)") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/110315 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [❌ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/155/builds/40828 "Hash 8638d464 for PR 66617 does not build (failure)") | [❌ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/156/builds/40403 "Hash 8638d464 for PR 66617 does not build (failure)") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/24689 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/20/builds/159062 "Built successfully and passed tests") | [❌ 🧪 api-ios](https://ews-build.webkit.org/#/builders/155/builds/40828 "Hash 8638d464 for PR 66617 does not build (failure)") | [❌ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/168/builds/33976 "Hash 8638d464 for PR 66617 does not build (failure)") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/178491 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/27799 "Built successfully and passed tests") | [❌ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/31389 "Failed to build and analyze WebKit") | [❌ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/168/builds/33976 "Hash 8638d464 for PR 66617 does not build (failure)") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/138159 "Passed tests") | | 
| | [❌ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/40465 "Hash 8638d464 for PR 66617 does not build (failure)") | [❌ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/156/builds/40403 "Hash 8638d464 for PR 66617 does not build (failure)") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/138070 "Passed tests") | | 
| | [❌ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/41153 "Hash 8638d464 for PR 66617 does not build (failure)") | [❌ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/168/builds/33976 "Hash 8638d464 for PR 66617 does not build (failure)") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/103979 "Built successfully") | | 
| | [❌ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/160/builds/41153 "Hash 8638d464 for PR 66617 does not build (failure)") | [❌ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/168/builds/33976 "Hash 8638d464 for PR 66617 does not build (failure)") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/199584 "Built successfully") | | 
| | [❌ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/39664 "Hash 8638d464 for PR 66617 does not build (failure)") | [❌ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/112738 "Failed to build and analyze WebKit") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/53663 "Passed tests") | | 
| | [❌ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/39060 "Hash 8638d464 for PR 66617 does not build (failure)") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/168/builds/33976 "Hash 8638d464 for PR 66617 does not build (failure)") | | | 
| | [❌ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/39305 "Hash 8638d464 for PR 66617 does not build (failure)") | | | | 
| | [❌ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/39204 "Hash 8638d464 for PR 66617 does not build (failure)") | | | |


aprotyas (commented):
> A few Cocoa sources don't survive the build's unified bundles, so de-unify them (@no-unify)

what does this mean? Couldn't we make unified build fixes instead? or is this a separate problem.

Source/WTF/wtf/PlatformHave.h
```diff
@@ -1847,3 +1847,9 @@
     || (PLATFORM(VISION) && __VISION_OS_VERSION_MIN_REQUIRED >= 260400))
 #define HAVE_CORE_TEXT_GLYPHHASCOMPLEXCOLOR_FUNCTION 1
 #endif
+
+// WKTextSelectionController is a Swift class; builds that don't compile the WebKit
+// Swift UIProcess API turn this off in their options file.
+#if !defined(HAVE_WK_TEXT_SELECTION_CONTROLLER) && PLATFORM(MAC)
+#define HAVE_WK_TEXT_SELECTION_CONTROLLER 1
```
aprotyas, line 1854: By default, you want this to mirror `HAVE_APPKIT_GESTURES_SUPPORT`, no? Or else there can be open source configurations where `HAVE_APPKIT_GESTURES_SUPPORT == 0` but `WKTextSelectionController` still compiles.

zakariaridouh:
@aprotyas sadly its a slightly different problem. Because of the recent switch to jumbo unified builds bundling 128 headers together I was running into a few issues with  AVFoundation softlink & real SPI usages in different files that were getting bundled together. 

Where a file that `SOFT_LINK`s an AVFoundation class gets bundled up with one that imports the real SPI for that class, the soft link macros and the real declaration redefined the same thing (and wouldn't compile). 

Other ones that I had to exclude from the unified bundles were some 3rd party libs (like libRTC and ANGLE) that defined the same macros names and these would collide when they ended up in the same TU.

rr-codes (changes requested):
I don't think we should be making source code changes to appease CMake; rather, the CMake build should be fixed so that it works the same as the Xcode build

Source/cmake/OptionsMac.cmake
```diff
@@ -136,15 +135,26 @@ WEBKIT_OPTION_END()
...
+SET_AND_EXPOSE_TO_BUILD(HAVE_WK_TEXT_SELECTION_CONTROLLER OFF)
+
 # -----------------------------------------------------------------------------
 # Toolchain / SDK resolution
 # -----------------------------------------------------------------------------
 include(WebKitXcrun)
-WEBKIT_RESOLVE_SDK(macosx)
+# Prefer the internal SDK with a public fallback, matching the top-level
+# CMakeLists.txt (the toolchain must match CMAKE_OSX_SYSROOT).
+WEBKIT_RESOLVE_SDK(macosx.internal macosx)
```
geoffreygaren, line 154: I think we need to discuss this. I would prefer internal vs public SDK to be a preset, so that you can have either / both, and it is explicit which one you have.

Source/WebCore/SourcesCocoa.txt
```diff
@@ -150,7 +150,7 @@ bindings/js/ScriptControllerMac.mm @nonARC
 bridge/objc/ObjCRuntimeObject.mm @nonARC @no-unify-when(bundle<=8) @cost:4
 bridge/objc/WebScriptObject.mm @nonARC @no-unify-when(bundle<=8) @cost:6
 bridge/objc/objc_class.mm @nonARC @no-unify-when(bundle<=8) @cost:5
-bridge/objc/objc_instance.mm @nonARC @no-unify-when(bundle<=8) @cost:4
+bridge/objc/objc_instance.mm @nonARC @no-unify @cost:4
```
geoffreygaren, line 153: @no-unify is bad for build speed. Let's figure out how to support the Internal SDK without regressing the unified build.

Source/WebKit/CMakeLists.txt
```diff
@@ -1051,6 +1121,13 @@ endmacro()
 GENERATE_MESSAGE_SOURCES(WebKit_DERIVED_SOURCES "${WebKit_MESSAGES_IN_FILES}")
 list(APPEND WebKit_UNIFIED_SOURCE_LIST_FILES SourcesMessageReceivers.txt)
 list(APPEND WebKit_UNIFIED_SOURCE_EXCLUDES "^[A-Za-z0-9_]+MessageReceiver\\.cpp")
+# Mac jumbo only: these ARC files redefine @interface WKWebView when unified (an in-tree
+# header vs. its staged forwarding copy), so de-unify the directory. Excluding (rather
+# than @no-unify) keeps them off the non-ARC list; they're re-added to WebKit_ARC_SOURCES
+# below. iOS doesn't hit the collision at its bundle size, so leave its unification alone.
+if (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    list(APPEND WebKit_UNIFIED_SOURCE_EXCLUDES "(^|/)UIProcess/Extensions/Cocoa/[^ ]*\\.mm([ ]|$)")
+endif ()
```
geoffreygaren, line 1130: Let's figure out how to support the internal SDK without regressing the unified build.

Source/cmake/OptionsMac.cmake
stwrt: I agree with @geoffreygaren here; being able to switch between public and internal SDKs with some like `cmake --preset mac-dev-debug-internal` would be nice.

emw-apple: +1

It might be less controversial to start with support for overriding it at configure time (`-DCMAKE_OSX_SYSROOT=macosx.internal`). That's what I am working towards in #66581 (I realize we are stepping on each other's toes here.)

Source/cmake/OptionsMac.cmake
```diff
@@ -205,14 +215,71 @@ set(WebKit_LIBRARY_TYPE SHARED)
...
+    unset(_textinput_private_real)
+    unset(_textinput_private_current)
+    unset(_ti_paths)
+    unset(_ti_objs)
+    unset(_ti_roots)
+    unset(_vfs_overlay)
+    unset(_empty_modulemap)
+endif ()
+unset(_textinput_private)
+unset(_textinput_fw)
```
emw-apple, line 272: Gross… we don't have to do anything like this in the canonical Xcode build. Why here?

Source/cmake/OptionsMac.cmake
```diff
@@ -205,14 +215,71 @@ set(WebKit_LIBRARY_TYPE SHARED)
...
+
+# The internal SDK ships additional WebKit SPI headers under usr/local/include.
+# Add it for clang and the Swift clang importer. Mirrors OptionsIOS.cmake.
+if (CMAKE_OSX_SYSROOT AND EXISTS "${CMAKE_OSX_SYSROOT}/usr/local/include")
```
emw-apple, line 276: nits:
- misleading comment, the internal SDK content is not about "WebKit SPI"
- the if statement should probably be a static condition, no need for the uncertainty of an existence check

```suggestion
if (USE_APPLE_INTERNAL_SDK)
```

Source/WebKit/CMakeLists.txt
```diff
@@ -965,12 +965,82 @@ WEBKIT_SETUP_SWIFT_AND_GENERATE_SWIFT_CPP_INTEROP_HEADER(WebKit WebKit
...
+                set(_pubh_src "${CMAKE_CURRENT_SOURCE_DIR}/${_pubh}")
+            endif ()
+            file(STRINGS "${_pubh_src}" _pubh_marker LIMIT_COUNT 1 REGEX "API_WEBKIT_ADDITIONS_REPLACEMENT")
+            if (_pubh_marker)
+                list(APPEND _wk_headers_to_replace "${_pubh}")
+            else ()
+                list(APPEND _wk_headers_to_symlink "${_pubh}")
+            endif ()
+        endforeach ()
+    endif ()
```
emw-apple, line 991: IMO this is too fine-grained for the build system to be figuring out at configure time. You should instead do what we do in Xcode, and have a script that copies public/private headers and optionally runs them through `replace-webkit-additions-includes.py`.

It's also a little weird to symlink some headers but not others. In Xcode, we just copy. There are also other transformations we do to headers (some `sed` processing and `unifdef`) that would make them impossible to symlink.

See `Source/WebKit/Scripts/postprocess-header-rule`. We might even be able to use the same script via a custom command.

Source/WebKit/CMakeLists.txt
```diff
@@ -1320,6 +1397,13 @@ endif ()
 list(APPEND WebKit_SOURCES ${WebKit_DERIVED_SOURCES})
 
 WEBKIT_COMPUTE_SOURCES(WebKit)
+# Re-add the WebExtension files excluded above (Mac only) to the ARC library so they
+# compile with -fobjc-arc.
+if (CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    file(GLOB_RECURSE _webext_arc_sources "${WEBKIT_DIR}/UIProcess/Extensions/Cocoa/*.mm")
```
emw-apple, line 1403: Glob patterns to find sources is almost always a bad idea; it means that adding/removing sources will break the incremental build.

emw-apple (changes requested):
1. I think there is too much going on in here to review in depth. Curious if you can think of ways to split it up. One idea would be to land the shared CMake logic first, then fix  each project's build configuration one PR at a time.

2. It seems like you should be able to get the build working without changes to product code, especially so many. Our codebase already builds this way in Xcode, so in general, changing the product to make CMake work feels wrong.

3. re: your CMake style. I see a lot of logic that is doing existence checks that don't really make sense. Things like AppleFeatures, WebKitAdditions, various scripts should be assumed to _always_ be present in the internal SDK. Making it all conditional reduces clarity. This line in your commit message is a good example of the lack of clarity:

   > - Generate the AppleFeatures.h stub only when no real header is reachable
  on the host or in the SDK, so the internal SDK's header wins.

   It's doing the right thing, but there's no need for so much indirection. Just use real AppleFeatures in the internal build, and don't generate it at all in the public build. We need to get out of the business of generating these stub headers altogether.

PR #66630, opened 2026-06-06
SVGURIReference href.baseVal setter writes to xlink:href instead of href

Ahmad-S792:
#### 01cbe1a3dd618b91db0fd296ebd4d742efe902ba
<pre>
SVGURIReference href.baseVal setter writes to xlink:href instead of href
<a href="https://bugs.webkit.org/show_bug.cgi?id=316487">https://bugs.webkit.org/show_bug.cgi?id=316487</a>
<a href="https://rdar.apple.com/178905586">rdar://178905586</a>

Reviewed by NOBODY (OOPS!).

This patch aligns WebKit with Gecko / Firefox with Blink / Chromium.

The &quot;href&quot; IDL property is backed by a single animated property registered
under both the &quot;href&quot; and the deprecated &quot;xlink:href&quot; attributes. When that
property was updated through the DOM (e.g. the href.baseVal setter), the
lazy synchronization serialized the value to whichever name was queried
first, so a hasAttributeNS(xlinkNs, &quot;href&quot;) check would invent an
&quot;xlink:href&quot; attribute.

SVG 2 §16.1.6 [1] makes &quot;href&quot; canonical and keeps &quot;xlink:href&quot; only for
backwards compatibility: when both are present the non-namespaced one wins.
Route href synchronization to a single backing attribute, preferring &quot;href&quot;
and falling back to &quot;xlink:href&quot; only when &quot;xlink:href&quot; exists and &quot;href&quot;
does not, so the sibling attribute is never fabricated. This mirrors the
read direction already in SVGURIReference::parseAttribute().

The resolution lives in a single attributeNameForSynchronization() helper
shared by both synchronizeAttribute() and synchronizeAllAttributes(), which
returns the attribute unchanged for non-href names. The empty-value check
now inspects the stored attribute directly via the hasStoredAttribute()
helper (elementData() &amp;&amp; findAttributeByName()) instead of hasAttribute(),
which avoids the re-entrant synchronizeAttribute() call that hasAttribute()
would otherwise trigger. hasStoredAttribute() is also reused by
attributeNameForSynchronization() to resolve the backing href attribute.

[1] <a href="https://w3c.github.io/svgwg/svg2-draft/linking.html#XLinkRefAttrs">https://w3c.github.io/svgwg/svg2-draft/linking.html#XLinkRefAttrs</a>

* LayoutTests/imported/w3c/web-platform-tests/svg/linking/scripted/xlink-href-compat-expected.txt: Progression
(WebCore::SVGElement::synchronizeAttribute):
(WebCore::SVGElement::synchronizeAllAttributes):
(WebCore::SVGElement::hasStoredAttribute const):
(WebCore::SVGElement::attributeNameForSynchronization const):
* Source/WebCore/svg/SVGElement.h:
</pre>
https://github.com/WebKit/WebKit/commit/01cbe1a3dd618b91db0fd296ebd4d742efe902ba

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/187984 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/61795 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/55526 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/198169 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/152756 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/atp/75f75c30-b3af-11f1-e3a8-2589a933be41) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/189846 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/63217 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/61519 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/146769 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/101377 "") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/atp/75f75c30-b3af-11f1-7629-a95509695c8f) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/190939 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/50509 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/167221 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/127290 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/atp/75f75c30-b3af-11f1-2771-5c83de4db9c2) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/49242 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/47036 "Passed tests") | | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/159490 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/46869 "Passed tests") | [✅ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/9280 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/54161 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/51643 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/200224 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/61392 "Built successfully") | | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/156479 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/62111 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/43332 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/156126 "Passed tests") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/50417 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/134213 "Built successfully") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/131942 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/60476 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/21855 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/59880 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/60071 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/60044 "Built successfully") | | | |


karlcow (commented):
nit:

[1] https://www.w3.org/TR/SVG2/linking.html#XLinkRefAttrs

This needs to be https://w3c.github.io/svgwg/svg2-draft/linking.html#XLinkRefAttrs

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -559,13 +560,16 @@ void SVGElement::synchronizeAttribute(const QualifiedName& name)
 {
     // If the value of the property has changed, serialize the new value to the attribute.
     if (auto value = propertyRegistry().synchronize(name)) {
+        // https://w3c.github.io/svgwg/svg2-draft/linking.html#XLinkRefAttrs
+        const QualifiedName& targetName = (name.matches(SVGNames::hrefAttr) || name.matches(XLinkNames::hrefAttr)) ? backingHrefAttributeName() : name;
```
nikolaszimmermann: I dislike the logic duplication and would welcome a helper function that returns a 'const QualifiedName&' and accepts a 'const QualifiedName&' as argument. That new helper function also avoids the need for backingHrefAttributeName().

Something like:

```
const QualifiedName& SVGElement::attributeNameForSynchronization(const QualifiedName& attributeName) const
{
    if (!attributeName.matches(SVGNames::hrefAttr) && !attributeName.matches(XLinkNames::hrefAttr))
        return attributeName;
    auto useAttributeName = attributeName;
    // Here the logic that lives in backingHrefAttributeName()
    return useAttributeName;
}
```

WDYT?

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -559,13 +560,16 @@ void SVGElement::synchronizeAttribute(const QualifiedName& name)
...
+
         // If the serialized value is empty and the attribute doesn't exist,
         // don't recreate it. This handles the case where the attribute was
         // explicitly removed after the list was emptied.
-        if (value->isEmpty() && !hasAttribute(name))
+        if (value->isEmpty() && !(elementData() && findAttributeByName(targetName)))
```
nikolaszimmermann, line 568: Ouch, so hasAttribute previously did a synchronization as well? Did we break the re-entrancy before?

```
bool Element::hasAttribute(const QualifiedName& name) const
{
    if (!elementData())
        return false;
    synchronizeAttribute(name);
    return elementData()->findAttributeByName(name);
}
```
?

nikolaszimmermann (changes requested):

Source/WebCore/svg/SVGElement.cpp
Ahmad-S792: Make sense, added helper function.

Source/WebCore/svg/SVGElement.cpp
Ahmad-S792: I think re-entry was just single depth, fixed it and also updated commit message.

Ahmad-S792: @nikolaszimmermann - If you can have another look, I just rebased. If anything I am happy to address.

karlcow (commented):
I'm giving another round of review with the eye of SVG2 spec and interoperability.

karlcow (commented):
The write rule is normative too

SVG 2, Interface SVGAnimatedString, 

> On setting [baseVal](https://w3c.github.io/svgwg/svg2-draft/types.html#__svg__SVGAnimatedString__baseVal), the following steps are run:
> 
> 1. If the reflected attribute’s element is an [SVGScriptElement](https://w3c.github.io/svgwg/svg2-draft/interact.html#InterfaceSVGScriptElement), let value be the result of executing the [Get Trusted Type compliant string](https://www.w3.org/TR/trusted-types/#get-trusted-type-compliant-string-algorithm) algorithm, with [TrustedScriptURL](https://www.w3.org/TR/trusted-types/#trustedscripturl), reflected attribute’s Document's relevant global object, 'SVGScriptElement href', and 'script'.
> 2. Otherwise, let value be the specified value.
> 3. If the reflected attribute is not present, the [SVGAnimatedString](https://w3c.github.io/svgwg/svg2-draft/types.html#InterfaceSVGAnimatedString) object is defined to additionally reflect a second, deprecated attribute, and that deprecated attribute is present, then set that deprecated attribute to value.
> 4. Otherwise, set the reflected attribute to value.

https://w3c.github.io/svgwg/svg2-draft/types.html#__svg__SVGAnimatedString__baseVal

The commit message only covers getting.

More than matching Gecko and Blink, it matches what the spec requires.

Measured on an element carrying only xlink:href, then assigning href.baseVal: 
1. Firefox 157 and Chrome 154 end with one attribute, the xlink one overwritten. 
2. Safari 26.4 and 27.0 ended with two, the xlink one stale at its old value. 

After this patch WebKit ends with one, and the prefix survives, because setSynchronizedLazyAttribute finds the attribute by index and sets
the value in place rather than replacing the name.


**All GOOD**

Non reviewer approval. Thanks for working on this @Ahmad-S792

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -600,8 +603,28 @@ void SVGElement::synchronizeAllAttributes()
     // SVGPropertyRegistry::synchronizeAllAttributes() returns the new values of
     // the properties which have changed but not committed yet.
     auto map = propertyRegistry().synchronizeAllAttributes();
-    for (const auto& entry : map)
-        setSynchronizedLazyAttribute(entry.key, AtomString { entry.value });
+    for (const auto& entry : map) {
+        const QualifiedName& targetName = attributeNameForSynchronization(entry.key);
```
nikolaszimmermann: const auto&

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -585,13 +586,15 @@ void SVGElement::synchronizeAttribute(const QualifiedName& name)
 {
     // If the value of the property has changed, serialize the new value to the attribute.
     if (auto value = propertyRegistry().synchronize(name)) {
+        const QualifiedName& targetName = attributeNameForSynchronization(name);
```
nikolaszimmermann: const auto&

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -600,8 +603,28 @@ void SVGElement::synchronizeAllAttributes()
...
+{
+    if (!attributeName.matches(SVGNames::hrefAttr) && !attributeName.matches(XLinkNames::hrefAttr))
+        return attributeName;
+
+    // The "href" property is shared between the "href" and deprecated "xlink:href"
+    // attributes. SVG 2 makes "href" canonical, but a pre-existing "xlink:href" (with
+    // no "href") keeps backing the property for backwards compatibility, so route
+    // synchronization there and never fabricate the sibling attribute.
+    // https://w3c.github.io/svgwg/svg2-draft/linking.html#XLinkRefAttrs
+    auto hasStoredAttribute = [&](const QualifiedName& name) {
```
nikolaszimmermann: This could be an inlined helper, so you can share the elementData && findAttributeByName logic with this function and synchronizeAllAttributes

nikolaszimmermann (commented):
Just have a few nits, than this looks ready.

This is potentially perf sensitive, adding the matches(SVGNames::hrefAttr / xlinkNames::hrefAttr) to each attr synchronization.

I would feel  more comfortable if @shallawa could have a look as well

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -600,8 +603,28 @@ void SVGElement::synchronizeAllAttributes()
...
-        setSynchronizedLazyAttribute(entry.key, AtomString { entry.value });
+    for (const auto& entry : map) {
+        const QualifiedName& targetName = attributeNameForSynchronization(entry.key);
+        setSynchronizedLazyAttribute(targetName, AtomString { entry.value });
+    }
+}
+
+const QualifiedName& SVGElement::attributeNameForSynchronization(const QualifiedName& attributeName) const
+{
+    if (!attributeName.matches(SVGNames::hrefAttr) && !attributeName.matches(XLinkNames::hrefAttr))
```
nikolaszimmermann: Maybe an [[unlikely]] annotation?

Source/WebCore/svg/SVGElement.cpp
Ahmad-S792: Went with preferred auto& in WebKit code.

Source/WebCore/svg/SVGElement.cpp
Ahmad-S792: Went with preferred auto& in WebKit code.

Source/WebCore/svg/SVGElement.cpp
Ahmad-S792: Done.

Source/WebCore/svg/SVGElement.cpp
Ahmad-S792: Done.

Source/WebCore/svg/SVGElement.cpp
```diff
@@ -600,8 +603,30 @@ void SVGElement::synchronizeAllAttributes()
...
+}
+
+bool SVGElement::hasStoredAttribute(const QualifiedName& name) const
+{
+    return elementData() && findAttributeByName(name);
+}
+
+const QualifiedName& SVGElement::attributeNameForSynchronization(const QualifiedName& attributeName) const
+{
+    if (attributeName.matches(SVGNames::hrefAttr) || attributeName.matches(XLinkNames::hrefAttr)) [[unlikely]] {
```
shallawa, line 619: Yes I agree with @nikolaszimmermann. This seems hacky approach.

If we already collected the attributes of the `SVGElement` by calling `synchronizeAllAttributes()`, why do we have to change the attribute's name here?

I think it is better to make `synchronizeAllAttributes()` return only one attribute for `href`. I think `SVGURIReference` can decide which href to be used from the `contextElement`.

karlcow:
seeAlso https://github.com/w3c/svgwg/issues/1043

PR #66703, opened 2026-06-08
[JSC][Temporal] Cache ICU UCalendar per CalendarID in CalendarICUBridge

hyjorc1:
#### 7636f61497084c5b86ae9a6e2d6157bca4f3f7ee
<pre>
[JSC][Temporal] Cache ICU UCalendar per CalendarID in CalendarICUBridge
<a href="https://bugs.webkit.org/show_bug.cgi?id=316569">https://bugs.webkit.org/show_bug.cgi?id=316569</a>
<a href="https://rdar.apple.com/179036919">rdar://179036919</a>

Reviewed by Yusuke Suzuki.

Previously every CalendarICUBridge call (calendarYear, calendarMonth,
calendarDay, etc.) called ucal_open/ucal_close, allocating and freeing
an ICU calendar object on each invocation.

This patch caches one UCalendar template per CalendarID in a static
per-process Vector. Access is controlled by a per-entry WTF::Lock and
a double-checked atomic initialized flag (acquire/release ordering).

The withCalendar(calendarId, lambda) helper acquires the entry lock,
runs the lambda with the cached UCalendar*, and releases immediately,
keeping lock scope minimal. For operations that need an independent
calendar state (calendarDateUntil month loop, calendarDateFromFields
countCal), the template is cloned via ucal_clone inside the lock.

Canonical link: <a href="https://commits.webkit.org/314904@main">https://commits.webkit.org/314904@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/a7e9a69131a91ee1889c7a6b3880a58c948142ca

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/167492 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/40987 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/34582 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/176545 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/125173 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 ios-apple 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/41423 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/41001 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/130299 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/125173 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [loading 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/50653ada-8eaa-48ea-8065-9e74ad17781e) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/170451 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/32710 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/150487 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/110816 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/31693 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/30391 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/25280 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/20/builds/159778 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/141680 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/28182 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/179085 "Built successfully") | | 
| [  ~~🛠 🧪 jsc-debug-arm64~~](https://ews-build.webkit.org/#/builders/171/builds/28600 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/31981 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/29900 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/138695 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/41061 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/34548 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/138582 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/41749 "Built successfully") | | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/104854 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/25499 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/33838 "Passed tests") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/26853 "Passed tests") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/200169 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/40253 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/113334 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/53800 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/39650 "Built successfully") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/4952 "") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/39901 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/39795 "Built successfully") | | | |

### DIFF
--- a/Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
+++ b/Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
@@ -32,6 +32,8 @@
 #include <unicode/ucal.h>
 #include <wtf/CheckedArithmetic.h>
 #include <wtf/DateMath.h>
+#include <wtf/Lock.h>
+#include <wtf/NeverDestroyed.h>
 #include <wtf/text/MakeString.h>
 #include <wtf/unicode/icu/ICUHelpers.h>
 
@@ -57,11 +59,9 @@ static CString buildICULocale(StringView calendarId)
     return makeString("und@calendar="_s, icuKeyword).utf8();
 }
 
-// openCalendar — internal: opens ICU UCalendar for the given CalendarID, set to UTC.
+// buildCalendarTemplate — internal: opens ICU UCalendar for the given CalendarID, set to UTC.
 // NOTE: For Gregory/ISO/Japanese/Buddhist/Roc: sets Gregorian change date to -infinity for proleptic Gregorian arithmetic.
-// FIXME: Cache one UCalendar per CalendarID in a Vector<std::unique_ptr<UCalendar, ICUDeleter<ucal_close>>, numberOfCalendarIDs>
-//        to avoid repeated ucal_open overhead on hot paths like calendarYear/calendarMonth/calendarDay.
-static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> openCalendar(CalendarID calendarId)
+static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> buildCalendarTemplate(CalendarID calendarId)
 {
     auto str = calendarIDToString(calendarId);
     auto locale = buildICULocale(str);
@@ -84,34 +84,77 @@ static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> openCalendar(CalendarI
     return cal;
 }
 
+struct CalendarCacheEntry {
+    WTF::Lock useLock;
+    std::atomic<bool> initialized { false };
+    UCalendar* cal { nullptr };
+};
+
+static CalendarCacheEntry& calendarCacheEntry(CalendarID calendarId)
+{
+    static LazyNeverDestroyed<Vector<CalendarCacheEntry>> cache;
+    static std::once_flag onceFlag;
+    std::call_once(onceFlag, [] {
+        cache.construct(intlAvailableCalendars().size());
+    });
+    return cache.get()[calendarId];
+}
+
+static void ensureCalendar(CalendarCacheEntry& entry, CalendarID calendarId)
+{
+    if (entry.initialized.load(std::memory_order_acquire))
+        return;
+    WTF::Locker locker { entry.useLock };
+    if (!entry.initialized.load(std::memory_order_relaxed)) {
+        auto owned = buildCalendarTemplate(calendarId);
+        entry.cal = owned.release();
+        entry.initialized.store(true, std::memory_order_release);
+    }
+}
+
+template<typename F>
+static auto withCalendar(CalendarID calendarId, F&& fn) -> decltype(fn(static_cast<UCalendar*>(nullptr)))
+{
+    CalendarCacheEntry& entry = calendarCacheEntry(calendarId);
+    ensureCalendar(entry, calendarId);
+    WTF::Locker locker { entry.useLock };
+    if (entry.cal)
+        ucal_clear(entry.cal);
+    return fn(entry.cal);
+}
+
 // lunarCalendarExtendedYearFor1972 — probes UCAL_EXTENDED_YEAR for the given lunisolar calendar
 // at ISO 1972-02-15. Returns 1972 on ISO-proleptic ICU; epoch-based year on older Apple ICU.
 // Chinese uses epoch 2637 BCE -> returns 4609 on ICU ≥76; Dangi uses epoch 2333 BCE -> returns 4305.
 // Result is cached per calendar: ICU version is fixed for the process lifetime.
 int32_t lunarCalendarExtendedYearFor1972(CalendarID calendarId)
 {
-    static int32_t chineseCached = INT32_MIN;
-    static int32_t dangiCached = INT32_MIN;
-    int32_t& cached = (calendarId == dangiCalendarID()) ? dangiCached : chineseCached;
-    if (cached != INT32_MIN)
-        return cached;
-    auto cal = openCalendar(calendarId);
-    if (!cal)
-        return cached = 1972; // fallback: assume ISO-proleptic
+    static std::atomic<int32_t> chineseCached { INT32_MIN };
+    static std::atomic<int32_t> dangiCached { INT32_MIN };
+    auto& cached = (calendarId == dangiCalendarID()) ? dangiCached : chineseCached;
+    int32_t value = cached.load(std::memory_order_relaxed);
+    if (value != INT32_MIN)
+        return value;
     // Use ucal_setMillis with a precomputed ISO epoch time — NOT ucal_setDateTime which
     // sets calendar-native fields (not ISO fields) on a non-Gregorian calendar.
     // ISO 1972-02-15 00:00:00 UTC = 66,960,000,000 ms from Unix epoch (1970-01-01).
     static constexpr double iso1972Feb15EpochMs = 66960000000.0;
-    UErrorCode status = U_ZERO_ERROR;
-    ucal_setMillis(cal.get(), iso1972Feb15EpochMs, &status);
-    if (U_FAILURE(status)) [[unlikely]]
-        return cached = 1972; // fallback: assume ISO-proleptic
-    int32_t extYear = ucal_get(cal.get(), UCAL_EXTENDED_YEAR, &status);
-    if (U_FAILURE(status)) [[unlikely]]
-        return cached = 1972; // fallback: assume ISO-proleptic
-    // Return the raw EXTENDED_YEAR: 1972 on ISO-proleptic ICU,
-    // epoch-based year (whatever the current ICU uses) on older Apple ICU.
-    return cached = extYear;
+    value = withCalendar(calendarId, [&](UCalendar* cal) -> int32_t {
+        if (!cal)
+            return 1972; // fallback: assume ISO-proleptic
+        UErrorCode status = U_ZERO_ERROR;
+        ucal_setMillis(cal, iso1972Feb15EpochMs, &status);
+        if (U_FAILURE(status)) [[unlikely]]
+            return 1972; // fallback: assume ISO-proleptic
+        int32_t extYear = ucal_get(cal, UCAL_EXTENDED_YEAR, &status);
+        if (U_FAILURE(status)) [[unlikely]]
+            return 1972; // fallback: assume ISO-proleptic
+        // Return the raw EXTENDED_YEAR: 1972 on ISO-proleptic ICU,
+        // epoch-based year (whatever the current ICU uses) on older Apple ICU.
+        return extYear;
+    });
+    cached.store(value, std::memory_order_relaxed);
+    return value;
 }
 
 // isoDateToEpochMs — internal: converts ISO PlainDate to epoch ms at noon UTC (avoids DST boundary issues)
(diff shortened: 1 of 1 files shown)


Constellation (commented):
Calendar is shared, but there is no ownership management between multiple threads. So it is possible that two threads use the same calendar at the same time, and setting configuration etc. interleaves each other (maybe, ICU data corruption can happen inside the class). So, we need a locking per calendar ID to manage the exclusive access.

hyjorc1:
> Calendar is shared, but there is no ownership management between multiple threads. So it is possible that two threads use the same calendar at the same time, and setting configuration etc. interleaves each other (maybe, ICU data corruption can happen inside the class). So, we need a locking per calendar ID to manage the exclusive access.

Thanks, fixed with per-thread cache.

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
```diff
@@ -83,6 +83,28 @@ static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> openCalendar(CalendarI
...
+static UCalendar* openCalendar(CalendarID calendarId)
+{
+    using CalendarPtr = std::unique_ptr<UCalendar, ICUDeleter<ucal_close>>;
+    struct PerThreadCache {
+        PerThreadCache()
+            : slots(intlAvailableCalendars().size())
+        {
+        }
+        Vector<CalendarPtr> slots;
+    };
```
Constellation: I think this is too costly in terms of memory while the usage is not so high. why not just having a lock per CalendarID?

hyjorc1: Ok, updated.

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
```diff
@@ -83,6 +83,79 @@ static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> openCalendar(CalendarI
...
+
+static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> cloneCalendar(CalendarID calendarId)
+{
+    CalendarCacheEntry& entry = calendarCacheEntry(calendarId);
+    ensureCalendar(entry, calendarId);
+    if (!entry.cal)
+        return nullptr;
+    UErrorCode status = U_ZERO_ERROR;
+    return std::unique_ptr<UCalendar, ICUDeleter<ucal_close>>(ucal_clone(entry.cal, &status));
+}
```
Constellation: Ah, there is an issue. If the already locked calendar is not the same to the calendarId, the calendar here is unlocked, and calling `ucal_clone` with it would cause the memory corruption issue inside ICU.

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
```diff
@@ -838,24 +914,26 @@ static bool compareSurpassesOrdinally(
 // resolveMonthCodeToOrdinal — internal: resolves a monthCode to its 1-based ordinal position in the given year
 static int32_t resolveMonthCodeToOrdinal(CalendarID calendarId, const String& monthCode, int32_t year)
 {
-    auto cal = openCalendar(calendarId);
+    // Clone: called mid-session inside diffCalendarDate while it holds the calendar lock.
+    auto calOwned = cloneCalendar(calendarId);
```
Constellation: The issue is that cloneCalendar's cloning target is not guaranteed to be locked. If it is not, then it causes a memory corruption issue.
So, if we would like to use cloning, we should pass an already opened calendar for calendarId as a parameter.

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
```diff
@@ -164,16 +237,16 @@ static constexpr auto japaneseEras = WTF::toArray<JapaneseEra>({
...
-    // YEAR=2: YEAR=1/January doesn't exist for eras starting mid-year (e.g. Showa on Dec 25).
-    ucal_set(cal.get(), UCAL_YEAR, 2);
-    ucal_set(cal.get(), UCAL_MONTH, UCAL_JANUARY);
-    ucal_set(cal.get(), UCAL_DAY_OF_MONTH, 1);
-    int32_t extYear = ucal_get(cal.get(), UCAL_EXTENDED_YEAR, &status);
+    ucal_set(cal, UCAL_ERA, icuEraCode);
+    ucal_set(cal, UCAL_YEAR, 2);
+    ucal_set(cal, UCAL_MONTH, UCAL_JANUARY);
+    ucal_set(cal, UCAL_DAY_OF_MONTH, 1);
+    int32_t extYear = ucal_get(cal, UCAL_EXTENDED_YEAR, &status);
```
Constellation: For japanese calendar ID case, it is having many issues (it can be unlocked and causing memory corruption). Maybe, we should have an independent cache for this japanese calendar.

Constellation (commented):
Also, once lock is taken, I think we should pass through `AbstractLocker&` to the callees from that, to ensure that the lock is taken since it is too hard to know whether the lock is taken or not.

Constellation (commented):
And/or, we should minimize the range of opening a calendar & taking a lock. We should compute all the related things without opening a calendar, opening it, bulk-setting & getting and then close as soon as possible. If we do so, potentially we do not need to clone some of calendars, which is significantly faster.

hyjorc1:
> And/or, we should minimize the range of opening a calendar & taking a lock. We should compute all the related things without opening a calendar, opening it, bulk-setting & getting and then close as soon as possible. If we do so, potentially we do not need to clone some of calendars, which is significantly faster.

Updated. The last update addresses all concerns.

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
```diff
@@ -84,6 +83,41 @@ static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> openCalendar(CalendarI
...
+}
+
+// withCalendar — acquires the per-calendarId lock, runs fn(UCalendar*), releases immediately.
+// Lock scope is minimal: only fn's execution. Each call acquires and releases independently.
+template<typename F>
+static auto withCalendar(CalendarID calendarId, F&& fn) -> decltype(fn(static_cast<UCalendar*>(nullptr)))
+{
+    CalendarCacheEntry& entry = calendarCacheEntry(calendarId);
+    ensureCalendar(entry, calendarId);
+    WTF::Locker locker { entry.useLock };
```
Constellation, line 120: Let's call `ucal_clear` to ensure that this calendar is no longer holding the previous state.

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
```diff
@@ -376,51 +413,67 @@ static std::optional<int32_t> mapTemporalEraToICUEra(CalendarID calendarId, Stri
...
-    if (calendarHasEras(calendarId)) {
-        int32_t icuEra = ucal_get(cal.get(), UCAL_ERA, &status);
+        raw.ucalYear = ucal_get(cal, UCAL_YEAR, &status);
         if (U_FAILURE(status)) [[unlikely]]
             return makeUnexpected(rangeError(icuReadCalendarFailed));
-        fields.era = mapICUEraToTemporalEra(calendarId, icuEra);
-        // NOTE: Japanese dates before 1873 use "ce"/"bce" per spec, not ICU era names.
-        fields.eraYear = ucal_get(cal.get(), UCAL_YEAR, &status);
+        raw.ordinalMonth = computeOrdinalMonth(cal, calendarId);
+        if (!raw.ordinalMonth) [[unlikely]]
```
Constellation: computeOrdinalMonth is internally calling ucal_add etc. so this modifies the state of `cal`. Is it fine to call this thing before obtaining the subsequent fields?

Constellation: I wonder if we need to open the calendar twice, since we modify the cal here?
Or at least, probably we should do this at the end of this function instead of allowing the rest of the code to get the field from the modified calendar.

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
```diff
@@ -839,47 +916,48 @@ static bool compareSurpassesOrdinally(
...
-            // Hebrew M05L (Adar I) constrains FORWARD to M06 (Adar) in non-leap years.
-            // icu4x components/calendar/src/cal/hebrew.rs ordinal_from_month: M05L -> ordinal 6 with Overflow::Constrain.
-            // All other calendars constrain backward to the previous existing month.
-            if (calendarId == hebrewCalendarID() && monthCode == "M05L"_s)
+        int32_t savedYear = ucal_get(cal, UCAL_EXTENDED_YEAR, &status);
+        if (U_FAILURE(status)) [[unlikely]]
+            return 1;
+        int32_t lastOrdinal = 1;
+        for (int i = 0; i < 14; i++) {
+            auto curCode = computeMonthCode(cal, calendarId);
```
Constellation: I would like to rename computeMonthCode to getMonthCode as it is not modifying cal.

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
```diff
@@ -1449,297 +1584,308 @@ int32_t ecmaReferenceYear(CalendarID calendarId, uint8_t monthNumber, bool isLea
...
-        if (monthCode->monthNumber > 13) [[unlikely]]
-            return makeUnexpected(rangeError("month is out of range"_s));
-        if (monthCode->monthNumber == 13 && calendarId != copticCalendarID() && calendarId != ethiopicCalendarID() && calendarId != ethioaaCalendarID()) [[unlikely]]
-            return makeUnexpected(rangeError("month is out of range"_s));
+    return withCalendar(calendarId, [&](UCalendar* cal) -> TemporalResult<ISO8601::PlainDate> {
+        if (!cal) [[unlikely]]
+            return makeUnexpected(rangeError(icuOpenCalendarFailed));
+
+        UErrorCode status = U_ZERO_ERROR;
+        ucal_clear(cal);
```
Constellation: Set it in `withCalendar` and remove this one.

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
Constellation: (If cal is not nullptr)

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
hyjorc1: Previously, we did restore millis at the end of `computeOrdinalMonth`. The last update moved all uses of `computeOrdinalMonth` to the end of function passed to `withCalendar`. So, the restore step is removed now.

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
```diff
@@ -84,6 +83,41 @@ static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> openCalendar(CalendarI
...
+struct CalendarCacheEntry {
+    WTF::Lock useLock;
+    std::atomic<bool> initialized { false };
+    UCalendar* cal { nullptr };
+};
+
+static CalendarCacheEntry& calendarCacheEntry(CalendarID calendarId)
+{
+    static NeverDestroyed<Vector<CalendarCacheEntry>> s_cache(intlAvailableCalendars().size());
+    return s_cache.get()[calendarId];
```
Constellation: Use `LazyNeverDestroyed` and call_once as NeverDestroyed<> initialization is racy.
```
static LazyNeverDestroyed<Vector<CalendarCacheEntry>> cache;
static std::once_flag onceKey;
std::call_once(onceKey, [&] {
    cache.construct(intlAvailableCalendars().size());
});
return cache.get()[calendarId];
```

Constellation (approved):

PR #66767, opened 2026-06-09
Introduce UsedStyle

anttijk:
#### 7fa01598b59ff410eb384131964531a6117f286c
<pre>
Introduce UsedStyle
<a href="https://bugs.webkit.org/show_bug.cgi?id=316654">https://bugs.webkit.org/show_bug.cgi?id=316654</a>
<a href="https://rdar.apple.com/179104082">rdar://179104082</a>

Reviewed by NOBODY (OOPS!).

UsedStyle is a stack type that maps ComputedStyle properties to the values used in layout, through
generated or hand-written getters chosen per-property via the `used-style` codegen property in
CSSProperties.json. (Close to, but not exactly, the spec&apos;s post-layout &quot;used style&quot;.)

This first patch covers float, clear, and the four margin longhands, routing the matching layout
reads through usedStyle().

* Source/WebCore/CMakeLists.txt:
* Source/WebCore/DerivedSources-output.xcfilelist:
* Source/WebCore/DerivedSources.make:
* Source/WebCore/Sources.txt:
* Source/WebCore/css/scripts/process-css-properties.py:
(StylePropertyCodeGenProperties):

Add the GenerateUsedStyle generator and the used-style codegen property. Only opted-in properties
are emitted, and an unrecognized used-style value is rejected.

* Source/WebCore/css/scripts/test/TestCSSProperties.json:
* Source/WebCore/css/scripts/test/TestCSSPropertiesResults/UsedStyleProperties.cpp: Added.
* Source/WebCore/css/scripts/test/TestCSSPropertiesResults/UsedStyleProperties.h: Added.
(WebCore::UsedStyleProperties::UsedStyleProperties):
(WebCore::UsedStyleProperties::computedStyle const):

* Source/WebCore/layout/formattingContexts/FormattingGeometry.cpp:
(WebCore::Layout::FormattingGeometry::computedHorizontalMargin const):
(WebCore::Layout::FormattingGeometry::computedVerticalMargin const):

Read margins through a renderer-free UsedStyle; the std::optional result feeds the auto-margin solver.

* Source/WebCore/layout/integration/LayoutIntegrationBoxGeometryUpdater.cpp:
(WebCore::LayoutIntegration::BoxGeometryUpdater::horizontalLogicalMargin):
(WebCore::LayoutIntegration::BoxGeometryUpdater::verticalLogicalMargin):
* Source/WebCore/layout/integration/LayoutIntegrationFormattingContextLayout.cpp:
(WebCore::LayoutIntegration::populateIFCWithNewlyPlacedFloats):
* Source/WebCore/layout/integration/inline/LayoutIntegrationLineLayout.cpp:
(WebCore::LayoutIntegration::LineLayout::preparePlacedFloats):
* Source/WebCore/rendering/FloatingObjects.cpp:
(WebCore::FloatingObject::FloatingObject):
(WebCore::RenderBlock::intrinsicLogicalMarginStartAndEnd const):
(WebCore::RenderBlock::computeBlockIntrinsicLogicalWidths const):
* Source/WebCore/rendering/RenderBlockFlow.cpp:
(WebCore::RenderBlockFlow::selfCollapsingMarginBeforeWithClear):
(WebCore::RenderBlockFlow::marginBeforeEstimateForChild const):
(WebCore::RenderBlockFlow::computeLogicalLocationForFloat):
(WebCore::RenderBlockFlow::positionNewFloats):
(WebCore::RenderBlockFlow::computedClearDeltaForChild):
(WebCore::RenderBlockFlow::computeInlineIntrinsicLogicalWidths const):
* Source/WebCore/rendering/RenderBox.cpp:
(WebCore::RenderBox::computeLogicalWidth const):
(WebCore::RenderBox::computeBlockDirectionMargins const):
(WebCore::RenderBox::blockAxisMarginForStretch const):
* Source/WebCore/rendering/RenderBoxInlines.h:
(WebCore::RenderBox::marginBoxRect const):
* Source/WebCore/rendering/RenderBoxModelObject.cpp:
(WebCore::RenderBoxModelObject::computeStickyPositionConstraints const):
* Source/WebCore/rendering/RenderElement.h:
* Source/WebCore/rendering/RenderElementInlines.h:
(WebCore::RenderElement::usedStyle const):
* Source/WebCore/rendering/RenderInline.cpp:
(WebCore::RenderInline::marginLeft const):
(WebCore::RenderInline::marginRight const):
(WebCore::RenderInline::marginTop const):
(WebCore::RenderInline::marginBottom const):
(WebCore::RenderInline::marginStart const):
(WebCore::RenderInline::marginEnd const):
(WebCore::RenderInline::marginBefore const):
(WebCore::RenderInline::marginAfter const):
(WebCore::computeMargin): Deleted.
* Source/WebCore/rendering/RenderReplaced.cpp:
(WebCore::RenderReplaced::computeConstrainedLogicalWidth const):
* Source/WebCore/rendering/RenderScrollbarPart.cpp:
(WebCore::RenderScrollbarPart::computeScrollbarWidth):
(WebCore::RenderScrollbarPart::computeScrollbarHeight):
* Source/WebCore/rendering/RenderTable.cpp:
(WebCore::RenderTable::updateLogicalWidth):
* Source/WebCore/rendering/style/UsedStyle.cpp: Added.
(WebCore::UsedStyle::UsedStyle):

Construct a renderer-free UsedStyle from a layout-tree box.

(WebCore::UsedStyleProperties::resolveReferenceSize const):

Resolves a ReferenceSize to a LayoutUnit; ContainingBlockWidth visits the containing block.

(WebCore::UsedStyle::floating const):
(WebCore::UsedStyle::clear const):

Map the computed value to physical left/right using the containing block&apos;s writing mode.

* Source/WebCore/rendering/style/UsedStyle.h: Added.
* Source/WebCore/rendering/style/UsedStyleReferenceSize.h: Added.

ReferenceSize names the renderer-derived percentage reference an evaluate-zoomed getter
resolves against, resolved lazily so the walk is not paid for fixed values. Callers that
already hold the reference pass a LayoutUnit instead.

(WebCore::Style::ComputedStyle::usedClear): Deleted.
(WebCore::Style::ComputedStyle::usedFloat): Deleted.

Moved to UsedStyle.

* Tools/Scripts/webkitpy/style/checkers/jsonchecker.py:
(JSONCSSPropertiesChecker.check_codegen_properties):
</pre>
https://github.com/WebKit/WebKit/commit/7fa01598b59ff410eb384131964531a6117f286c

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/181491 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/55438 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/49240 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/191714 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/145817 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/052f1753-8aff-4942-84ef-89a6fc99cf91) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/183353 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/56744 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/55159 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/141654 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/98305 "1 flakes 12 failures") | [loading 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/6dde872f-f408-45d0-bf83-3351408a9194) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/184446 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/45338 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/162403 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/122094 "Passed tests") | | [loading 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/31fed071-1ecd-433b-8e28-8d771db3db09) 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/180815 "Passed tests") | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/44016 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/42285 "Passed tests") | | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/154417 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/41930 "Passed tests") | [✅ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/2875 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/48064 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/46541 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/193642 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/55107 "Built successfully") | | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/151057 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/55794 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/38552 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/150764 "Passed tests") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/45342 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/128077 "Built successfully") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/123708 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/54310 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/16928 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/53692 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/53930 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/53757 "Built successfully") | | | |


alanbaradlay (approved):

fantasai:
> This does not exactly match the spec concept of "used style" which is post-layout but it is close enough.

Can we just call it `RenderStyle` then?

Source/WebCore/rendering/style/UsedStyle.cpp
```diff
@@ -0,0 +1,96 @@
...
+    switch (referenceSize) {
+    case ReferenceSize::Zero:
+        return 0_lu;
+    case ReferenceSize::ContainingBlockWidth: {
+        // The inline size of the containing block, as used to resolve percentage margins/padding.
+        // containingBlockLogicalWidthForContent() handles out-of-flow, grid items and table cells
+        // correctly (not just containingBlock()->contentBoxLogicalWidth()).
+        ASSERT(m_renderer);
+        CheckedRef renderBoxModelObject = downcast<RenderBoxModelObject>(*m_renderer);
+        return renderBoxModelObject->containingBlockLogicalWidthForContent();
```
fantasai, line 54: Hiding this inside a style getter seems dangerous.

Source/WebCore/rendering/style/UsedStyle.cpp
```diff
@@ -0,0 +1,96 @@
...
+    }
+    }
+
+    ASSERT_NOT_REACHED();
+    return 0_lu;
+}
+
+UsedFloat UsedStyle::floating() const
+{
+    auto writingMode = m_renderer->containingBlock()->writingMode();
```
fantasai, line 64: While I think this is kindof ok for float/clear because it's rarely used/checked, in general I don't think we should encapsulate this check. What writing system the layout code needs will vary, so `WritingMode` imho should be a parameter that gets passed in.

So I think we shouldn't do that here, either. It sets up a bad precedent.

fantasai: Also in loops over children where we're pulling this info on every child, it's more efficient to pull it once into a local variable, and then pass it in. So that's another reason not to do this.

Again, it might be fine for float/clear because they're rare. But we should set up patterns here that will work for everything.

fantasai (changes requested):

Source/WebCore/rendering/style/UsedStyle.cpp
fantasai: Basically, I think style should not be calling layout code (outside of interleaved layout stuff). Layout should consume style data. It's ok for style to become more "processed" so it's easier to consume, and I think in fact we want that. But calling into Render methods is something to avoid.

It's not just a layering issue. When we hide layout logic inside style, the layout user is less likely to realize what they're doing and the implications of it.

Source/WebCore/rendering/RenderElement.h
```diff
@@ -76,6 +77,7 @@ class RenderElement : public RenderObject {
     bool hasInitializedStyle() const { return m_hasInitializedStyle; }
 
     const Style::ComputedStyle& style() const LIFETIME_BOUND { return m_style; }
+    inline UsedStyle usedStyle() const LIFETIME_BOUND;
     const Style::ComputedStyle* parentStyle() const LIFETIME_BOUND { return !m_parent ? nullptr : &m_parent->style(); }
     const Style::ComputedStyle& firstLineStyle() const LIFETIME_BOUND;
```
fantasai, line 82: I think we should consider making `style()` return the "used style", and have layout code specifically ask for `computedStyle()` or `style().computed()` when it wants to get the computed value specifically. I expect it's going to be unusual for layout to want the computed style, so making the "used style" the norm for RenderElement makes more sense to me. But I'm not super clear on the intended distinction between UsedStyle and RenderStyle here.

Personally, I think we should have Style::ComputedStyle, which represents the actual computed style as needed for inheritance, and RenderStyle, which represents the style data more processed towards the "used style" for use in layout; and that's it. I don't think we should have three different "style" objects, it becomes unclear which one you're supposed to use.

Source/WebCore/rendering/RenderInline.cpp
```diff
@@ -267,51 +267,44 @@ LayoutPoint RenderInline::firstInlineBoxTopLeft() const
...
-{
-    return Style::evaluateMinimum<LayoutUnit>(margin, [&] ALWAYS_INLINE_LAMBDA {
-        return std::max<LayoutUnit>(0, renderer->containingBlock()->contentBoxLogicalWidth());
-    }, zoomFactor);
-}
-
 LayoutUnit RenderInline::marginLeft() const
 {
-    return computeMargin(this, style().marginLeft(), style().usedZoomForLength());
+    return usedStyle().marginLeft(ReferenceSize::ContainingBlockWidth);
```
fantasai: Following up on https://github.com/WebKit/WebKit/pull/66767/changes#r3390084622 - Here, for example, if we encapsulated all the zoom nonsense (which really does belong encapsulated within style) it would be a perfectly readable

> style().marginLeft(containingBlock()->contentBoxLogicalWidth());

This doesn't hide the layout call. It makes it really obvious what's happening. But also it isn't the convoluted snarl that the original code you're replacing is. This is, imho, the correct level of encapuslation here.

Source/WebCore/rendering/RenderElement.h
weinig: I agree about the renaming of `style()` to `computedStyle()` or `style().computed()`, but I assume @anttik has done it this way as a bridging mechanism, as changing it wholesale would change a lot of callsites, and so making the name change could be done a bit later.

Also, `RenderStyle` is no more, so it's only two types, one for computed style, one for "used style".

fantasai (commented):
So I guess overall my take is:
- Style should pre-process style data, and provide convenience methods, to make the style data easily usable in layout. This is why we want a layer on top of ComputedStyle: because style logic should stay inside style, and repetitive intervening calculations should be stuffed into methods that encapsulate that logic.
- But also layout should consume style, not the other way around. Layout calls should stay in layout.

Therefore:
- Stuff like zoom factors and display-based value conversions should be encapsulated by the style objects. They just add clutter to layout.
- Style objects can accept WritingMode and LayoutUnit arguments to resolve things, but shouldn't hold or use render objects directly.

Source/WebCore/rendering/RenderElement.h
fantasai: > changing it wholesale would change a lot of callsites

Can we avoid that by instead making UsedStyle inherit from ComputedStyle? That wouldn't work with the current approach, which is wrapping ComputedStyle and RenderElement into a single object -- but I don't think we should be doing that in the first place.

> Also, RenderStyle is no more, so it's only two types, one for computed style, one for "used style".

Ok, yeah, then let's call the style prepared for rendering "RenderStyle". :) Since it's not really the used style anyway, and it's keeping its core purpose (representing style data for layout/rendering).

fantasai:
> style should consume layout

Uh, wow, got that totally backwards. I meant "layout should consume style". Edited to match intent. ^_^;;

Source/WebCore/rendering/RenderElement.h
anttijk: Right, the idea is to do this incrementally. This is not the end state. We'd eventually rename

style() -> computedStyle()
usedStyle() -> style()

They shouldn't inherit each other, they are different level types (UsedStyle is layout) and expose different level interface (computed style vs fully resolved). Also they are very different kinds of types, UsedStyle being a non-persistent view.

Source/WebCore/rendering/style/UsedStyle.cpp
anttijk: The intention here is specifically that this is a rendering side type and can perform style transformations that can't be resolved by style alone. This allows us to consolidate a lot of repetitive and bug-prone code. Note also that it is not in Style namespace and sits in rendering subdirectory.

Source/WebCore/rendering/style/UsedStyle.cpp
anttijk: Figuring out what is correct WritingMode to pass for each client is very difficult. The idea is to encapsulate as much of that here as possible. Note also that the patch just translates the existing call sites as-is.

Reading writingMode() out of style you are already reading (which is most of them) is extremely efficient and manual local caching is unlikely to have any benefits.

anttijk:
> Can we just call it `RenderStyle` then?

I think it is too early for that. People have mental model what RenderStyle is and this is very different. We may rename later.

Source/WebCore/rendering/RenderInline.cpp
anttijk: The reason ContainingBlockWidth is nice (besides reducing burden to clients) is that most Lengths don't need a reference size to resolve. Doing it your way does an unnecessary tree walk in the common case. This will also work nicely with future properties like padding.

Note that your code is also subtly wrong which sort of reinforces the point (it should use `containingBlockLogicalWidthForContent()`).

Source/WebCore/rendering/RenderElement.h
anttijk: I don't think we want to recycle RenderStyle name this soon but I'm open to other suggestions. `ResolvedStyle` , `LayoutStyle`,...?

Source/WebCore/rendering/RenderInline.cpp
anttijk: I removed ReferenceSize::Zero which indeed was providing no value.

PR #66790, opened 2026-06-09
[JSC][WASM][Debugger] Replace polling-based STW participation with VMBlockingScope

hyjorc1:
#### fb30bac88bb45e0804e337122dc8bbf8f2fa0d00
<pre>
[JSC][WASM][Debugger] Replace polling-based STW participation with VMBlockingScope
<a href="https://bugs.webkit.org/show_bug.cgi?id=316687">https://bugs.webkit.org/show_bug.cgi?id=316687</a>
<a href="https://rdar.apple.com/179141754">rdar://179141754</a>

Reviewed by Mark Lam.

Worker threads blocked in Atomics.wait/memory.atomic.wait or WebCore synchronous
operations (SubtleCrypto, IndexedDB, WebSocket, FileSystem, Notification) previously
polled NeedStopTheWorld every 50ms to cooperate with the WASM debugger&apos;s
Stop-The-World protocol.

Replace with VMBlockingScope RAII: entering a blocking op increments
m_numberOfBlockedVMs, making the STW trigger (stopped + blocked == active)
fire without polling. On unblock, the VM atomically transitions to the
stopped counter and proceeds through normal STW participation.

Design assumption: the main thread VM is never OS-blocked while entered,
ensuring m_numberOfStoppedVMs &gt; 0 when the condition fires so a VM is
always available to service the callback.

Tests:
* JSTests/wasm/debugger/tests/tests.py:
(MemoryAtomicWaitTestCase.execute):
(MemoryAtomicWaitNoTimeoutTestCase.execute):

Canonical link: <a href="https://commits.webkit.org/316724@main">https://commits.webkit.org/316724@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/6b5202e2d5886da75405727089710620092fe6f4

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/173109 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/47040 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/40692 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/182682 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/130665 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/174974 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/48334 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/46723 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/134628 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/95037 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/51960bfd-a234-438b-8fcc-cf9cb1fe1db9) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/176067 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/38018 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/155345 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/115097 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/cf5c5d3c-0845-41d4-bcf7-c1598fbfc78d) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/36663 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/35007 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/31167 "Built successfully") | | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/3860 "Passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/147087 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/34843 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/185104 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/34371 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/37905 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/39209 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/143458 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/46709 "Built successfully") | | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/143330 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/47388 "Built successfully") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/31678 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/112461 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/26306 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/38285 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/119137 "Built successfully") | [⏳ 🛠 jsc-armv7 ](https://ews-build.webkit.org/#/builders/JSC-ARMv7-32bits-Build-EWS "Waiting in queue, processing has not started yet") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/45894 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/9789 "Passed tests") | [⏳ 🧪 jsc-armv7-tests ](https://ews-build.webkit.org/#/builders/JSC-ARMv7-32bits-Build-EWS "Waiting in queue, processing has not started yet") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/45296 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/45527 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/45353 "Built successfully") | | | |


Source/WebCore/workers/WorkerNotificationClient.cpp
```diff
@@ -32,7 +32,7 @@
 #include "NotificationResources.h"
 #include "WorkerGlobalScope.h"
 #include "WorkerLoaderProxy.h"
-#include "WorkerSTWParticipation.h"
+#include <JavaScriptCore/VMManager.h>
```
MenloDorian: Moved this below the `WorkerThread.h` include.  Check the style checker.

Source/WebCore/workers/WorkerGlobalScope.cpp
```diff
@@ -75,7 +75,7 @@
 #include "WorkerNavigator.h"
 #include "WorkerOrWorkletGlobalScope.h"
 #include "WorkerReportingProxy.h"
-#include "WorkerSTWParticipation.h"
+#include <JavaScriptCore/VMManager.h>
```
MenloDorian: Move this #include down below.

Source/WebCore/Modules/websockets/WorkerThreadableWebSocketChannel.cpp
```diff
@@ -45,7 +45,7 @@
 #include "WorkerGlobalScope.h"
 #include "WorkerLoaderProxy.h"
 #include "WorkerRunLoop.h"
-#include "WorkerSTWParticipation.h"
+#include <JavaScriptCore/VMManager.h>
```
MenloDorian: Ditto.  Move below.

Source/WebCore/Modules/filesystem/WorkerFileSystemStorageConnection.cpp
```diff
@@ -32,7 +32,7 @@
 #include "FileSystemSyncAccessHandle.h"
 #include "WorkerGlobalScope.h"
 #include "WorkerLoaderProxy.h"
-#include "WorkerSTWParticipation.h"
+#include <JavaScriptCore/VMManager.h>
```
MenloDorian: Move this down below.

Source/JavaScriptCore/runtime/WaiterListManager.cpp
```diff
@@ -122,6 +83,7 @@ WaiterListManager::WaitSyncResult WaiterListManager::waitSyncImpl(VM& vm, ValueT
     Ref<WaiterList> list = findOrCreateList(ptr);
     MonotonicTime time = MonotonicTime::timePointFromNow(timeout);
 
+    VMBlockingScope waitScope(vm, StopTheWorldEvent::WasmAtomicsWaitBlocked);
```
MenloDorian: Atomics is a JS API.  See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics.  Why `WasmAtomicsWaitBlocked`?  I suggest just naming this `AtomicsWaitBlocked`.

Also, why not put this `VMBlockingScope` inside the `{}` block below just before `Locker listLocker`?  I think that will more clearly document that this scope object is intended for that scope.

Source/JavaScriptCore/runtime/VMManager.h
```diff
@@ -314,6 +318,12 @@ class VMManager {
     void resumeTheWorld() WTF_REQUIRES_LOCK(m_worldLock);
     void NODELETE incrementActiveVMs(VM&) WTF_REQUIRES_LOCK(m_worldLock);
     void decrementActiveVMs(VM&) WTF_REQUIRES_LOCK(m_worldLock);
+    void handleVMExit(VM&) WTF_REQUIRES_LOCK(m_worldLock);
+    void enterStopTheWorldParticipation(VM&, StopTheWorldEvent);
+    bool allActiveVMsAtStoppingPoint() const WTF_REQUIRES_LOCK(m_worldLock)
```
MenloDorian: Let's call this `allActiveVMsHaveReachedStoppingPoint` instead.

Source/JavaScriptCore/runtime/VMManager.cpp
```diff
@@ -330,11 +316,113 @@ void VMManager::resumeTheWorld() WTF_REQUIRES_LOCK(m_worldLock)
...
+        // ability to just wake the desired one up. So, wake up all the threads and let
+        // them sort themselves out.
+        //
+        // But if the targetVM thread is this thread, then pass the control to another
+        // thread, any thread. That's because this thread is dying imminently.
+        if (m_targetVM == &vm) {
+            m_targetVM = nullptr;
+            m_useRunOneMode = false;
+            // Keep m_blockingScopeTarget since it is still being debugged if available.
+            // notifyAll() below will let any other VM pick up as proxy.
```
MenloDorian: This comment seem out of place.  Perhaps a different comment should accompany the `if (m_blockingScopeTarget == &vm)` block below.  I don't see anything here that `Keep m_blockingScopeTarget since it is still being debugged`.  What am I missing?

Source/JavaScriptCore/runtime/VMManager.cpp
```diff
@@ -330,11 +316,113 @@ void VMManager::resumeTheWorld() WTF_REQUIRES_LOCK(m_worldLock)
...
+VMBlockingScope::~VMBlockingScope()
+{
+    if (m_event)
+        VMManager::singleton().notifyVMUnblocking(m_vm, *m_event);
+}
+
+static void setVMStopped(VM& vm)
+{
+#if ENABLE(WEBASSEMBLY_DEBUGGER)
+    if (Options::enableWasmDebugger()) [[unlikely]]
```
MenloDorian: nit: I suggest just checking `vm.debugState()` instead of `Options::enableWasmDebugger()`.  It requires less loads and works better for CPU pointer chasing.

Source/JavaScriptCore/runtime/VMManager.cpp
```diff
@@ -330,11 +316,113 @@ void VMManager::resumeTheWorld() WTF_REQUIRES_LOCK(m_worldLock)
...
+        vm.debugState()->setStopped();
+#else
+    UNUSED_PARAM(vm);
+#endif
+}
+
+static void clearVMStopped(VM& vm)
+{
+#if ENABLE(WEBASSEMBLY_DEBUGGER)
+    if (Options::enableWasmDebugger()) [[unlikely]]
```
MenloDorian: nit: Just checking `vm.debugState()` instead of `Options::enableWasmDebugger()`.

Source/JavaScriptCore/runtime/VMManager.cpp
```diff
@@ -330,11 +316,113 @@ void VMManager::resumeTheWorld() WTF_REQUIRES_LOCK(m_worldLock)
...
+        VMManager::singleton().notifyVMBlocking(m_vm);
+}
+
+VMBlockingScope::~VMBlockingScope()
+{
+    if (m_event)
+        VMManager::singleton().notifyVMUnblocking(m_vm, *m_event);
+}
+
+static void setVMStopped(VM& vm)
```
MenloDorian: nit: I suggest naming this function `notifyDebuggerOfVMStopping` instead.  It makes it clearer in the client what this function is doing.

Source/JavaScriptCore/runtime/VMManager.cpp
```diff
@@ -330,11 +316,113 @@ void VMManager::resumeTheWorld() WTF_REQUIRES_LOCK(m_worldLock)
...
+{
+#if ENABLE(WEBASSEMBLY_DEBUGGER)
+    if (Options::enableWasmDebugger()) [[unlikely]]
+        vm.debugState()->setStopped();
+#else
+    UNUSED_PARAM(vm);
+#endif
+}
+
+static void clearVMStopped(VM& vm)
```
MenloDorian: nit: I suggest naming this function `notifyDebuggerOfVMResuming` instead.

Source/JavaScriptCore/runtime/VMManager.cpp
```diff
@@ -330,11 +316,113 @@ void VMManager::resumeTheWorld() WTF_REQUIRES_LOCK(m_worldLock)
...
+#endif
+}
+
+void VMManager::notifyVMBlocking(VM& vm)
+{
+    ASSERT(vm.isEntered());
+
+    Locker lock { m_worldLock };
+
+    RELEASE_ASSERT(!std::exchange(vm.traps().m_isInBlockingScope, true));
```
MenloDorian: nit: can you remove the empty lines here?  It'd be nice if the code is a little denser.  You can also move the ASSERT after the lock so that the assertions are grouped together, and also to emphasize the required lock at the top.  I don't think there's any added value in doing the ASSERT first.

Source/JavaScriptCore/runtime/VMManager.cpp
```diff
@@ -330,11 +316,113 @@ void VMManager::resumeTheWorld() WTF_REQUIRES_LOCK(m_worldLock)
...
+
+    ++m_numberOfBlockedVMs;
+
+    setVMStopped(vm);
+
+    if (m_worldMode != Mode::RunAll && allActiveVMsAtStoppingPoint()) {
```
MenloDorian: nit: Ditto: can you remove the empty lines here to increase code density a bit?

Source/JavaScriptCore/runtime/VMManager.cpp
```diff
@@ -330,11 +316,113 @@ void VMManager::resumeTheWorld() WTF_REQUIRES_LOCK(m_worldLock)
...
+}
+
+void VMManager::notifyVMUnblocking(VM& vm, StopTheWorldEvent exitEvent)
+{
+    ASSERT(vm.isEntered());
+
+    {
+        Locker lock { m_worldLock };
+
+        RELEASE_ASSERT(std::exchange(vm.traps().m_isInBlockingScope, false));
```
MenloDorian: nit: can you remove the empty lines.  Also, I suggest moving the ASSERT after the lock to group the assertions together.

Source/JavaScriptCore/runtime/VMManager.cpp
```diff
@@ -330,11 +316,113 @@ void VMManager::resumeTheWorld() WTF_REQUIRES_LOCK(m_worldLock)
...
+
+        --m_numberOfBlockedVMs;
+
```
MenloDorian: nit: remove this empty line.

Source/JavaScriptCore/runtime/VMManager.cpp
```diff
@@ -454,9 +544,12 @@ void VMManager::notifyVMStop(VM& vm, StopTheWorldEvent event)
         switch (m_currentStopReason) {
         case StopReason::GC:
             RELEASE_ASSERT_NOT_REACHED();
-        case StopReason::WasmDebugger:
-            status = g_jscConfig.wasmDebuggerOnStop(vm, event);
+        case StopReason::WasmDebugger: {
+            // Use blocking scope target as debuggee when set (proxy mode).
+            VM& debuggee = m_blockingScopeTarget ? *m_blockingScopeTarget : *m_targetVM;
```
MenloDorian: How can this be correct?  By the time we get here, shouldn't `m_targetVM` always be the current VM?  Also, why is it valid to call the WasmDebugger from this thread with a different VM than the current VM?

At minimum, we need a comment here explaining why this is appropriate.  It certainly feels unintuitive.

Source/JavaScriptCore/runtime/VMManager.cpp
```diff
@@ -482,11 +575,27 @@ void VMManager::notifyVMStop(VM& vm, StopTheWorldEvent event)
                 m_useRunOneMode = status.second != STW_RESUME_ALL_TOKEN;
         }
 
-        if (status.second && status.second != STW_RESUME_ALL_TOKEN && status.second != m_targetVM) {
+        bool switchContext = [&]() {
```
MenloDorian: Let's call this `needContextSwitch` instead.

Source/JavaScriptCore/runtime/VMManager.cpp
hyjorc1: The current `debugState()` ensures the debug state is not null. So, I will add `debugStateIfExists()` for that purpose.

Source/JavaScriptCore/runtime/VMManager.cpp
hyjorc1: Ditto

Source/JavaScriptCore/runtime/VMManager.cpp
hyjorc1: Discussed offline. Changed
```
m_targetVM -> m_servingVM // Serving the call back
m_blockingScopeTarget -> m_targetVM // Call back target
```

Source/JavaScriptCore/runtime/VMManager.cpp
hyjorc1: Removed. Since no need to update the original code at all with changes: https://github.com/WebKit/WebKit/pull/66790#discussion_r3516475964.

Source/JavaScriptCore/runtime/VMManager.cpp
```diff
@@ -330,12 +318,108 @@ void VMManager::resumeTheWorld() WTF_REQUIRES_LOCK(m_worldLock)
...
+        Locker lock { m_worldLock };
+
+        ASSERT(vm.isEntered());
+        RELEASE_ASSERT(std::exchange(vm.traps().m_isInBlockingScope, false));
+        --m_numberOfBlockedVMs;
+        if (m_worldMode == Mode::RunAll) {
+            notifyDebuggerOfVMResuming(vm);
+            return;
+        }
+        ++m_numberOfStoppedVMs;
```
MenloDorian, line 421: I think this increment of `m_numberOfStoppedVMs` deserves a comment.  If I understand correctly, `m_numberOfStoppedVMs` needs to increased because the VM is transitioning from a VM blocked to a VM stopped state.  It was not in the stopped state initially because it was blocked on a wait of sorts.  

With that said, it seems that we should only increment `m_numberOfStoppedVMs` if we are still trying to stop the world.  Should we be stopping (and therefore incrementing otherwise?  Your implementation here is behaviorally correct i.e. we assume we need to stop, and unconditionally increment `m_numberOfStoppedVMs` and let `enterStopTheWorldParticipation` sort it out.  However, this means every blocking / wait operation will pessimistically call `enterStopTheWorldParticipation` to determine if it needs to stop. Since we're already holding the `m_worldLock`, can we not check `VMManager::m_worldMode` and only choose to stop if and only if `m_worldMode` != `Mode::RunAll` or `Mode::RunOne`?

```
    bool needToStop = false;
    {
        Locker lock { m_worldLock };
        ...
        if (VMManager::isStoppingTheWorld()) {
            needToStop = true;
            ++m_numberOfStoppedVMs;
        }
    }

    if (needToStop)
        enterStopTheWorldParticipation(vm, exitEvent);
```

Would that work?

Source/JavaScriptCore/runtime/VMManager.h
```diff
@@ -358,10 +369,17 @@ class VMManager {
     // stop, atomically read-and-cleared by last VM exiting notifyVMStop().
     Atomic<bool> m_needsWasmDebuggerOnResume { false };
 
-    // Indicates the targetVM that will service the StopTheWorld request, or the targetVM that may
+    // Indicates the VM that will service the StopTheWorld request (i.e. drive the
+    // enterStopTheWorldParticipation loop and invoke the callback), or the VM that may
     // continue running in RunOne mode.
```
MenloDorian, line 374: This comment made me realize that RunOne mode should only be allowed if `m_servingVM` == `m_targetVM`.  It doesn't make sense to RunOne on a different VM.  Can you add some RELEASE_ASSERTs for this, or do so in a follow up patch?

Either that, or RunOne on a blocked `m_targetVM` simply does nothing and calls the client / debugger back immediately.

Or maybe VMManager needs to simulate running where nothing happens (because the thread is blocked), and the user's only recourse is to Ctrl-C again.  

You don't have to address this in this PR, but please give it some thought, and do a follow up if needed.

MenloDorian (approved):
r=me with some suggestions for improvements.

Source/JavaScriptCore/runtime/VMManager.cpp
hyjorc1: `VMManager::isStoppingTheWorld()` would help readability, but the existing check
```
if (m_worldMode == Mode::RunAll) {
    notifyDebuggerOfVMResuming(vm);
    return;
}
```
already guards for that and serves the same purpose. I'll add a comment here to make that clearer.

As for `RunOne` mode: as long as this VM isn't the serving VM, it still needs to be stopped in `enterStopTheWorldParticipation`. So `== Mode::RunAll` (equivalently, `!= Mode::RunAll` on the "needs to stop" path) is the correct check here. `RunOne` should still fall through to `enterStopTheWorldParticipation` rather than being excluded alongside `RunAll`.

Source/JavaScriptCore/runtime/VMManager.h
hyjorc1: I will add this
```
RELEASE_ASSERT(m_servingVM == &vm && m_servingVM == m_targetVM);
```
for all `Mode::RunOne` check and set sites.

PR #66824, opened 2026-06-10
[LinkedIn] Tapping a video in the feed on iPad does not surface media controls

Phinny01:
#### 740f093fdcb9330b1e4685167b2be2b057cb1cf5
<pre>
[LinkedIn] Tapping a video in the feed on iPad does not surface media controls
<a href="https://bugs.webkit.org/show_bug.cgi?id=316707">https://bugs.webkit.org/show_bug.cgi?id=316707</a>
<a href="https://rdar.apple.com/171231918">rdar://171231918</a>

Reviewed by Abrar Rahman Protyasha and Brent Fulgham.

LinkedIn calls preventDefault() on touchend for taps inside the feed, including taps targeting the
&lt;video&gt;. iOS treats that as &quot;page handled&quot; and cancels UITapGestureRecognizer, so no synthetic
click fires and the player&apos;s user-active toggle never flips.

Add a linkedin.com quirk that overrides the touchend &quot;handled&quot; result to false in
WebPage::dispatchTouchEvent when the target hit-tests to an HTMLMediaElement, so the tap commits
normally. Pair with a node-level gate so the resulting click lands on the &lt;video&gt;.
The page&apos;s touchend handler still runs; only the gesture-cancellation signal is suppressed.

* Source/WebCore/dom/Node.cpp:
(WebCore::Node::willRespondToMouseClickEventsWithEditability const):
* Source/WebCore/page/Quirks.h:
(WebKit::WebPage::dispatchTouchEvent):

Canonical link: <a href="https://commits.webkit.org/315049@main">https://commits.webkit.org/315049@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/bb3758520c225e7a9ad30040b8fb0ec2221ac496

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/167825 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/41322 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/34917 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/176883 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/125506 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/169691 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/41757 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/41335 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/130569 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/91772 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/170784 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/33023 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/150807 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/111086 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/32004 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/30708 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/25615 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/141993 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/28502 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/179425 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/32472 "Built successfully") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/30220 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/138987 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/41394 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/34865 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/138871 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/37422 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/42082 "Built successfully") | | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/105309 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/34146 "Passed tests") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/27261 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/40586 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/113838 "Built successfully") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/39983 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/5276 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/40234 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/40128 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/dom/Node.cpp
+++ b/Source/WebCore/dom/Node.cpp
@@ -51,6 +51,7 @@
 #include "HTMLDialogElement.h"
 #include "HTMLElement.h"
 #include "HTMLImageElement.h"
+#include "HTMLMediaElement.h"
 #include "HTMLSlotElement.h"
 #include "HTMLStyleElement.h"
 #include "InputEvent.h"
@@ -2861,6 +2862,11 @@ bool Node::willRespondToMouseClickEventsWithEditability(Editability editability)
 #endif
     if (editability != Editability::ReadOnly)
         return true;
+
+#if PLATFORM(IOS_FAMILY) && ENABLE(IOS_TOUCH_EVENTS)
+    if (document().quirks().shouldAllowNativeTapsOnMediaElements(this))
+        return true;
+#endif
 
     auto& eventNames = WebCore::eventNames();
     return eventTypes().containsIf([&](const auto& type) {
--- a/Source/WebCore/page/Quirks.cpp
+++ b/Source/WebCore/page/Quirks.cpp
@@ -678,6 +678,19 @@ bool Quirks::shouldComputeSimulatedMouseEventMovementDelta() const
 
     return m_quirksData.isTikTok || m_quirksData.isFacebook;
 }
+
+#if PLATFORM(IOS_FAMILY) && ENABLE(IOS_TOUCH_EVENTS)
+bool Quirks::shouldAllowNativeTapsOnMediaElements(const Node* node) const
+{
+    QUIRKS_EARLY_RETURN_IF_DISABLED_WITH_VALUE(false);
+
+    if (!m_quirksData.isLinkedIn)
+        return false;
+
+    return is<HTMLMediaElement>(node) && downcast<HTMLMediaElement>(*node).hasClassName("vjs-tech"_s);
+}
+#endif
+
 #endif
 
 // live.com rdar://52116170
@@ -3554,6 +3567,13 @@ static void handleIMDBQuirks(QuirksData& quirksData, const URL& /* quirksURL */,
 
 }
 
+static void handleLinkedInQuirks(QuirksData& quirksData, const URL& /* quirksURL */, const String& quirksDomainString, const URL&  /* documentURL */)
+{
+    QUIRKS_EARLY_RETURN_IF_NOT_DOMAIN("linkedin.com"_s);
+
+    quirksData.isLinkedIn = true;
+}
+
 static void handleLiveQuirks(QuirksData& quirksData, const URL& quirksURL, const String& quirksDomainString, const URL& /* documentURL */)
 {
     QUIRKS_EARLY_RETURN_IF_NOT_DOMAIN("live.com"_s);
@@ -4102,6 +4122,7 @@ void Quirks::determineRelevantQuirks()
         { "imdb"_s, &handleIMDBQuirks },
         { "instagram"_s, &handleInstagramQuirks },
         { "invideo"_s, &handleInVideoQuirks },
+        { "linkedin"_s, &handleLinkedInQuirks },
         { "live"_s, &handleLiveQuirks },
 #if PLATFORM(MAC)
         { "madisoncityk12"_s, &handleMadisonCityK12Quirks },
--- a/Source/WebKit/WebProcess/WebPage/WebPage.cpp
+++ b/Source/WebKit/WebProcess/WebPage/WebPage.cpp
@@ -4254,6 +4254,18 @@ Expected<bool, WebCore::RemoteFrameGeometryTransformer> WebPage::dispatchTouchEv
     CurrentEvent currentEvent(touchEvent);
     auto handleTouchEventResult = handleTouchEvent(frameID, touchEvent, m_page.get());
     updatePotentialTapSecurityOrigin(touchEvent, handleTouchEventResult.value_or(false));
+
+    if (touchEvent.type() == WebEventType::TouchEnd && handleTouchEventResult.value_or(false)) {
+        if (RefPtr localMainFrame = this->localMainFrame()) {
+            if (RefPtr document = localMainFrame->document()) {
+                FloatPoint adjustedPoint;
+                RefPtr responder = localMainFrame->nodeRespondingToClickEvents(FloatPoint(touchEvent.position()), adjustedPoint);
+                if (document->quirks().shouldAllowNativeTapsOnMediaElements(responder.get()))
+                    handleTouchEventResult = false;
+            }
+        }
+    }
+
     return handleTouchEventResult;
 }
 
--- a/Source/WebCore/page/Quirks.h
+++ b/Source/WebCore/page/Quirks.h
@@ -363,6 +363,10 @@ public:
 
     void determineRelevantQuirks();
 
+#if PLATFORM(IOS_FAMILY) && ENABLE(IOS_TOUCH_EVENTS)
+    WEBCORE_EXPORT bool shouldAllowNativeTapsOnMediaElements(const Node*) const;
+#endif
+
 #if PLATFORM(IOS_FAMILY)
     bool NODELETE shouldSendFakeTouchForceChangeEvent() const;
 #endif
--- a/Source/WebCore/page/QuirksData.h
+++ b/Source/WebCore/page/QuirksData.h
@@ -46,6 +46,7 @@ struct QuirksData {
     bool isGoogleMaps : 1 { false };
     bool isIHeart : 1 { false };
     bool isInVideo : 1 { false };
+    bool isLinkedIn : 1 { false };
     bool isNBA : 1 { false };
     bool isNetflix : 1 { false };
     bool isOutlook : 1 { false };


lilyspiniolas:
Does this happen universally for all <video> elements on LinkedIn? If not I wonder if we can make this more targeted.

Phinny01:
> Does this happen universally for all

Yes this happens for all videos on the LinkedIn home feed

brentfulgham (approved):
The Quirks part of this change looks perfect, and the changes to Node and WebPage look reasonable and in line with other quirks work we have done.

I think a media expert should ideally confirm that the Node and WebPage changes are as expected for media purposes.

Source/WebCore/page/Quirks.cpp
```diff
@@ -677,6 +677,23 @@ bool Quirks::shouldComputeSimulatedMouseEventMovementDelta() const
 
     return m_quirksData.isTikTok || m_quirksData.isFacebook;
 }
+
+bool Quirks::shouldTreatMediaElementsAsRespondingToClicks() const
+{
+    QUIRKS_EARLY_RETURN_IF_DISABLED_WITH_VALUE(false);
+
+    return m_quirksData.isLinkedIn;
+}
```
aprotyas: why is this separately defined? Can't you just use `shouldAllowNativeTapsOnMediaElements()`?

Source/WebCore/dom/Node.cpp
```diff
@@ -2862,6 +2863,11 @@ bool Node::willRespondToMouseClickEventsWithEditability(Editability editability)
     if (editability != Editability::ReadOnly)
         return true;
 
+#if PLATFORM(IOS_FAMILY) && ENABLE(TOUCH_EVENTS)
+    if (is<HTMLMediaElement>(*this) && document().quirks().shouldTreatMediaElementsAsRespondingToClicks())
```
aprotyas: I don't understand this part. Do media elements not respond to synthetic clicks otherwise? I know you're scoping this to LinkedIn, but is this the right thing to do in general? How do media components elsewhere on the web work?

Source/WebKit/WebProcess/WebPage/WebPage.cpp
```diff
@@ -4254,6 +4254,18 @@ Expected<bool, WebCore::RemoteFrameGeometryTransformer> WebPage::dispatchTouchEv
     CurrentEvent currentEvent(touchEvent);
     auto handleTouchEventResult = handleTouchEvent(frameID, touchEvent, m_page.get());
     updatePotentialTapSecurityOrigin(touchEvent, handleTouchEventResult.value_or(false));
+
+    if (touchEvent.type() == WebEventType::TouchEnd && handleTouchEventResult.value_or(false)) {
+        if (RefPtr localMainFrame = this->localMainFrame()) {
+            if (RefPtr document = localMainFrame->document(); document && document->quirks().shouldAllowNativeTapsOnMediaElements()) {
+                FloatPoint adjustedPoint;
+                RefPtr responder = localMainFrame->nodeRespondingToClickEvents(FloatPoint(touchEvent.position()), adjustedPoint);
+                if (responder && is<HTMLMediaElement>(*responder))
```
aprotyas: Since we know that LinkedIn.com uses Video.js, could we not limit the quirk scope further?

Source/WebCore/page/Quirks.cpp
Phinny01: you're right. I removed it

Source/WebCore/dom/Node.cpp
Phinny01: It is true that any site with the same pattern (`preventDefault` on `touchend` + click handler on a parent + no listener on` <video>`) would hit the same bug. preventDefault on `touchend` is how WebKit suppresses synthetic clicks on iOS, so undoing it everywhere changes the contract for pages that aren't broken. A per-site quirk lets us fix LinkedIn without touching that.

aprotyas: Ideally we also want to understand why the Video.js component which prevent defaults the touchdown doesn't do its own hover. Or maybe it's a limitation in our approach if this is working on Android mobile devices? (i.e. touchend.preventDefault() + `:hover` active both work at the same time)

Phinny01: Will test on android and dig further as discussed in slack

Source/WebKit/WebProcess/WebPage/WebPage.cpp
Phinny01: Going narrower would mean a class lookup at touch-time and binding to Video.js's class names. Going broader (any .video-js globally) would change behavior on every Video.js consumer. If we find another site with the same bug, it would be easier to add it to the quirk set then than to library-detect now.

aprotyas: I would say it's usually safe to bind to class names of established library components, and class name lookups are not _particularly_ expensive, especially if we only do them after determining (a) we're on LinkedIn.com (b) this is a media element

But yeah, we're probably bikeshedding at this point.

Phinny01: Agreed it's bikeshedding — but went ahead and updated to gate on the vjs-tech class.

Source/WebCore/page/Quirks.cpp
```diff
@@ -678,6 +677,16 @@ bool Quirks::shouldComputeSimulatedMouseEventMovementDelta() const
...
     return m_quirksData.isTikTok || m_quirksData.isFacebook;
 }
+
+#if PLATFORM(IOS_FAMILY) && ENABLE(IOS_TOUCH_EVENTS)
+bool Quirks::shouldAllowNativeTapsOnMediaElements() const
+{
+    QUIRKS_EARLY_RETURN_IF_DISABLED_WITH_VALUE(false);
+
+    return m_quirksData.isLinkedIn;
+}
```
aprotyas: shouldn't this subsume the entire logic?

i.e. we pass a node here, and it does the following logic:

```cpp
bool Quirks::shouldAllowNativeTapsOnMediaElements(auto node) const
{
    if (!m_quirksData.isLinkedIn)
        return false;

    return is<HTMLMediaElement>(node) && node->hasClassName("vjs-tech"_s)
}
```

aprotyas: This way you address the asymmetry in our gating logic between `willRespondToMouseClickEventsWithEditability` vs `dispatchTouchEvent`, too

aprotyas (approved):

PR #66893, opened 2026-06-10
[JSC][Temporal] Cache ICU UCalendar per TimeZoneID in TimeZoneICUBridge

hyjorc1:
#### 7830cc6341b519edd8ffa9726c0bebc1715d52da
<pre>
[JSC][Temporal] Cache ICU UCalendar per TimeZoneID in TimeZoneICUBridge
<a href="https://bugs.webkit.org/show_bug.cgi?id=316815">https://bugs.webkit.org/show_bug.cgi?id=316815</a>
<a href="https://rdar.apple.com/179262861">rdar://179262861</a>

Reviewed by Yusuke Suzuki.

TimeZoneICUBridge: Previously every getOffsetNanosecondsFor,
getPossibleEpochNanosecondsFor, and getTimeZoneTransition call opened
and closed a UCalendar via ucal_open/ucal_close per invocation.

This patch caches one UCalendar per IANA timezone in a static per-process
TinyLRUCache&lt;TimeZone, RefPtr&lt;TimeZoneCacheEntry&gt;, 8&gt;. Using TimeZone as
the key avoids cross-thread String ownership issues and uses the
already-resolved canonical ID for comparison. TimeZoneCacheEntry is
ThreadSafeRefCounted so callers hold a RefPtr that keeps the entry alive
even if the LRU evicts it. Each entry owns its UCalendar via
unique_ptr&lt;UCalendar, ICUDeleter&lt;ucal_close&gt;&gt; so it is properly closed
when evicted and destroyed. Initialization is folded into withTimeZone
under the entry lock, no separate atomic flag needed.

getTimeZoneTransition previously cloned the calendar to query before/after
offsets at each transition. Since only ucal_setMillis + ucal_get are used
(no ucal_set field overrides), cal can be reused directly for the queries
and ucal_clone is eliminated.

getNamedTimeZoneEpochNanoseconds previously called getOffsetNanosecondsFor
recursively while holding the entry lock; that call is now inlined using
getOffsetMsAtEpoch to avoid deadlock.

CalendarICUBridge: buildCalendarTemplate now takes const AbstractLocker&amp;
to document that the caller holds the entry lock. CalendarCacheEntry::cal
is now unique_ptr&lt;UCalendar, ICUDeleter&lt;ucal_close&gt;&gt; for proper cleanup.
Initialization is folded into withCalendar under the entry lock. Two
ucal_clone calls are eliminated:
- calendarDateUntil: trialCalOwned clone removed; workCal walks the month
  loop directly since ucal_setMillis(startMs) resets it afterward.
- calendarDateFromFields: countCal clone removed; cal counts months then
  resets via ucal_setMillis(calMs) before advancing to the target month.

Canonical link: <a href="https://commits.webkit.org/315075@main">https://commits.webkit.org/315075@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/f14313a473ca7316daa736df775928b3380c0f52

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/167945 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/41442 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/35038 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/177240 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/125638 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/fc4dde89-db81-40d7-8277-df71c174e62a) 
| | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/41877 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/41457 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/130656 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/91834 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/02debcda-1370-43bf-837a-66297a7f0ea4) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/170904 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/33129 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/150911 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/111173 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/02e1cd0d-4e2a-448a-8a68-210bdb1af85a) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/32110 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/30812 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/25943 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/20/builds/160563 "Built successfully and passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/142100 "Passed tests") | [✅ 🧪 mac-wk1](https://ews-build.webkit.org/#/builders/167/builds/28605 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/179840 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/29191 "Built successfully and passed tests") | [❌ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/32592 "") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/30323 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/139080 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/41514 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/34969 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/138962 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/42202 "Built successfully") | | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/105481 "Built successfully") | | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/25579 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/34250 "Passed tests") | [❌ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/27279 "Found 1 new test failure: imported/w3c/web-platform-tests/web-locks/acquire.https.any.sharedworker.html (failure)") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/200622 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/40706 "Built successfully") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/113958 "Built successfully") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/53895 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/40103 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/5379 "Passed tests") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/40354 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/40248 "Built successfully") | | | |

### DIFF
--- a/Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
+++ b/Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
@@ -61,7 +61,7 @@ static CString buildICULocale(StringView calendarId)
 
 // buildCalendarTemplate — internal: opens ICU UCalendar for the given CalendarID, set to UTC.
 // NOTE: For Gregory/ISO/Japanese/Buddhist/Roc: sets Gregorian change date to -infinity for proleptic Gregorian arithmetic.
-static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> buildCalendarTemplate(CalendarID calendarId)
+static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> buildCalendarTemplate(const AbstractLocker&, CalendarID calendarId)
 {
     auto str = calendarIDToString(calendarId);
     auto locale = buildICULocale(str);
@@ -85,9 +85,8 @@ static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> buildCalendarTemplate(
 }
 
 struct CalendarCacheEntry {
-    WTF::Lock useLock;
-    std::atomic<bool> initialized { false };
-    UCalendar* cal { nullptr };
+    Lock useLock;
+    std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> cal;
 };
 
 static CalendarCacheEntry& calendarCacheEntry(CalendarID calendarId)
@@ -100,27 +99,16 @@ static CalendarCacheEntry& calendarCacheEntry(CalendarID calendarId)
     return cache.get()[calendarId];
 }
 
-static void ensureCalendar(CalendarCacheEntry& entry, CalendarID calendarId)
-{
-    if (entry.initialized.load(std::memory_order_acquire))
-        return;
-    WTF::Locker locker { entry.useLock };
-    if (!entry.initialized.load(std::memory_order_relaxed)) {
-        auto owned = buildCalendarTemplate(calendarId);
-        entry.cal = owned.release();
-        entry.initialized.store(true, std::memory_order_release);
-    }
-}
-
 template<typename F>
 static auto withCalendar(CalendarID calendarId, F&& fn) -> decltype(fn(static_cast<UCalendar*>(nullptr)))
 {
     CalendarCacheEntry& entry = calendarCacheEntry(calendarId);
-    ensureCalendar(entry, calendarId);
-    WTF::Locker locker { entry.useLock };
+    Locker locker { entry.useLock };
+    if (!entry.cal)
+        entry.cal = buildCalendarTemplate(locker, calendarId);
     if (entry.cal)
-        ucal_clear(entry.cal);
-    return fn(entry.cal);
+        ucal_clear(entry.cal.get());
+    return fn(entry.cal.get());
 }
 
 // lunarCalendarExtendedYearFor1972 — probes UCAL_EXTENDED_YEAR for the given lunisolar calendar
@@ -1307,22 +1295,16 @@ TemporalResult<ISO8601::Duration> calendarDateUntil(CalendarID calendarId, const
         if (U_FAILURE(status)) [[unlikely]]
             return makeUnexpected(rangeError(icuCalendarArithmeticFailed));
 
-        // Clone for surpassesMonths iteration.
-        UErrorCode cloneStatus = U_ZERO_ERROR;
-        auto trialCalOwned = std::unique_ptr<UCalendar, ICUDeleter<ucal_close>>(ucal_clone(cal, &cloneStatus));
-        if (!trialCalOwned || U_FAILURE(cloneStatus)) [[unlikely]]
-            return makeUnexpected(rangeError(icuOpenCalendarFailed));
-
         for (;;) {
-            auto surpasses = surpassesMonths(trialCalOwned.get(), calendarId, sign, source.day, target.year, target.monthCode, target.ordinalMonth, target.day);
+            auto surpasses = surpassesMonths(cal, calendarId, sign, source.day, target.year, target.monthCode, target.ordinalMonth, target.day);
             if (!surpasses) [[unlikely]]
                 return makeUnexpected(rangeError(icuCalendarArithmeticFailed));
             if (*surpasses)
                 break;
             months = candidateMonths;
             candidateMonths += sign;
-            // trialCal already advanced by surpassesMonths — no reset needed.
-            // (Resetting to cal's fixed position would cause infinite loop since cal never moves.)
+            // cal already advanced by surpassesMonths — no reset needed here.
+            // ucal_setMillis(cal, startMs) below resets it before applying the final months count.
         }
 
         // Restore cal to (one + years), apply total months in one ucal_add (avoids undo asymmetry).
@@ -1762,33 +1744,30 @@ TemporalResult<ISO8601::PlainDate> calendarDateFromFields(CalendarID calendarId,
             ucal_set(cal, UCAL_DAY_OF_MONTH, 1);
             ucal_getMillis(cal, &status);
             if (U_FAILURE(status)) [[unlikely]]
-                return makeUnexpected(rangeError("Failed to resolve lunisolar calendar"_s)); // Use a separate calendar to count months in year.
-            // Clone: need an independent copy to advance months while cal stays at its current date.
-            UErrorCode cloneStatus = U_ZERO_ERROR;
-            auto countCal = std::unique_ptr<UCalendar, ICUDeleter<ucal_close>>(ucal_clone(cal, &cloneStatus));
-            if (U_FAILURE(cloneStatus)) [[unlikely]]
-                return makeUnexpected(rangeError("Failed to open counting calendar"_s));
+                return makeUnexpected(rangeError("Failed to resolve lunisolar calendar"_s));
+            // Count months in year by walking cal forward, then reset to year start.
             double calMs = ucal_getMillis(cal, &status);
             if (U_FAILURE(status)) [[unlikely]]
                 return makeUnexpected(rangeError(icuReadCalendarFailed));
-            ucal_setMillis(countCal.get(), calMs, &status);
-            if (U_FAILURE(status)) [[unlikely]]
-                return makeUnexpected(rangeError(icuSetCalendarFailed));
-            int32_t savedYear = ucal_get(countCal.get(), UCAL_EXTENDED_YEAR, &status);
+            int32_t savedYear = ucal_get(cal, UCAL_EXTENDED_YEAR, &status);
             if (U_FAILURE(status)) [[unlikely]]
                 return makeUnexpected(rangeError(icuReadCalendarFailed));
             int32_t monthsInYear = 1;
             for (int i = 0; i < 14; i++) {
-                ucal_add(countCal.get(), UCAL_MONTH, 1, &status);
+                ucal_add(cal, UCAL_MONTH, 1, &status);
                 if (U_FAILURE(status)) [[unlikely]]
                     return makeUnexpected(rangeError(icuCalendarArithmeticFailed));
-                int32_t curYear = ucal_get(countCal.get(), UCAL_EXTENDED_YEAR, &status);
+                int32_t curYear = ucal_get(cal, UCAL_EXTENDED_YEAR, &status);
                 if (U_FAILURE(status)) [[unlikely]]
                     return makeUnexpected(rangeError(icuReadCalendarFailed));
                 if (curYear != savedYear)
                     break;
                 monthsInYear++;
             }
+            // Reset to year start before advancing to target month.
+            ucal_setMillis(cal, calMs, &status);
+            if (U_FAILURE(status)) [[unlikely]]
+                return makeUnexpected(rangeError(icuSetCalendarFailed));
 
             // Clamp or reject month against monthsInYear.
             uint8_t resolvedMonth = month;
--- a/Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
+++ b/Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
@@ -30,26 +30,63 @@
 #include "CalendarICUBridge.h"
 #include "DateConstructor.h"
 #include "IntlObject.h"
+#include "JSCTimeZone.h"
 #include <unicode/ucal.h>
 #include <wtf/DateMath.h>
+#include <wtf/Lock.h>
+#include <wtf/NeverDestroyed.h>
+#include <wtf/TZoneMalloc.h>
+#include <wtf/ThreadSafeRefCounted.h>
+#include <wtf/TinyLRUCache.h>
 #include <wtf/unicode/icu/ICUHelpers.h>
 
 namespace JSC {
 namespace TemporalCore {
 
-// openCalendarForTimeZone — internal: opens ICU UCalendar for a named (IANA) timezone
-static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> openCalendarForTimeZone(const TimeZone& timeZone)
+class TimeZoneCacheEntry : public ThreadSafeRefCounted<TimeZoneCacheEntry> {
+    WTF_MAKE_TZONE_ALLOCATED(TimeZoneCacheEntry);
+public:
+    Lock useLock;
+    std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> cal;
+};
+WTF_MAKE_TZONE_ALLOCATED_IMPL(TimeZoneCacheEntry);
+
+struct TimeZoneLRUCachePolicy {
+    static bool isKeyNull(const TimeZone&) { return false; }
+    static RefPtr<TimeZoneCacheEntry> createValueForNullKey() { return nullptr; }
+    static RefPtr<TimeZoneCacheEntry> createValueForKey(const TimeZone&) { return adoptRef(*new TimeZoneCacheEntry); }
+    static TimeZone createKeyForStorage(const TimeZone& tz) { return tz; }
+};
+
+static RefPtr<TimeZoneCacheEntry> timeZoneCacheEntry(const TimeZone& timeZone)
+{
+    static Lock cacheLock;
+    static LazyNeverDestroyed<TinyLRUCache<TimeZone, RefPtr<TimeZoneCacheEntry>, 8, TimeZoneLRUCachePolicy>> cache;
+    static std::once_flag onceFlag;
+    std::call_once(onceFlag, [] {
+        cache.construct();
+    });
+    Locker locker { cacheLock };
+    return cache.get().get(timeZone);
+}
+
+template<typename F>
+static auto withTimeZone(const TimeZone& timeZone, F&& fn) -> decltype(fn(static_cast<UCalendar*>(nullptr)))
 {
     ASSERT(timeZone.isID());
-    String tzStr = timeZone.toICUString();
-    StringView view(tzStr);
-    auto upconverted = view.upconvertedCharacters();
-    UErrorCode status = U_ZERO_ERROR;
-    auto calendar = std::unique_ptr<UCalendar, ICUDeleter<ucal_close>>(
-        ucal_open(upconverted, view.length(), "", UCAL_DEFAULT, &status));
-    if (U_FAILURE(status)) [[unlikely]]
-        return nullptr;
-    return calendar;
+    auto entry = timeZoneCacheEntry(timeZone);
+    ASSERT(entry);
+    Locker locker { entry->useLock };
+    if (!entry->cal) {
+        String tzStr = timeZone.toICUString();
+        StringView view(tzStr);
+        auto upconverted = view.upconvertedCharacters();
+        UErrorCode status = U_ZERO_ERROR;
+        entry->cal = std::unique_ptr<UCalendar, ICUDeleter<ucal_close>>(ucal_open(upconverted, view.length(), "", UCAL_DEFAULT, &status));
+    }
+    if (entry->cal)
+        ucal_clear(entry->cal.get());
+    return fn(entry->cal.get());
 }
 
 // getOffsetMsAtEpoch — internal: queries ICU for UTC+DST offset in ms at a given epoch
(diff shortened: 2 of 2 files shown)


Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -32,24 +32,64 @@
...
+#include <wtf/TZoneMalloc.h>
 #include <wtf/unicode/icu/ICUHelpers.h>
 
 namespace JSC {
 namespace TemporalCore {
 
-// openCalendarForTimeZone — internal: opens ICU UCalendar for a named (IANA) timezone
-static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> openCalendarForTimeZone(const TimeZone& timeZone)
+struct TimeZoneCacheEntry {
+WTF_MAKE_STRUCT_TZONE_ALLOCATED(TimeZoneCacheEntry);
```
Constellation: Add space?

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -32,24 +32,64 @@
...
+    WTF::Lock useLock;
+    std::atomic<bool> initialized { false };
+    UCalendar* cal { nullptr };
+};
+WTF_MAKE_TZONE_ALLOCATED_IMPL(TimeZoneCacheEntry);
+
+static TimeZoneCacheEntry& timeZoneCacheEntry(const String& tzId)
```
Constellation: Let's use TimeZone as a key instead of String representation. Setting String key to the process-shared-singleton will move String ownership to the other thread and causing thread-safety issue.

The issue is that TimeZone can have so many keys due to random offset (`+10:00` and `+10:01` offsets for example), and there is no good way to keep things small right now, so it becomes infinitely growing.

I would like to suggest introducing `WTF::TinyLRUCache<TimeZone, Ref<Entry>>`, and getting an entry, taking a lock, and use it. So then if some concurrent use of TinyLRUCache will purge that entry, still we hold the ownership (since it is ref-counted via ThreadSafeRefCounted). Then we can bound the max of cached things while we can keep recently used one alive.

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -32,24 +32,64 @@
...
+    std::call_once(onceFlag, [] {
+        cache.construct();
+    });
+    WTF::Locker locker { cacheMapLock };
+    return *cache.get().ensure(tzId, [] {
+        return makeUnique<TimeZoneCacheEntry>();
+    }).iterator->value;
+}
+
+static void ensureTimeZone(TimeZoneCacheEntry& entry, const String& tzId)
```
Constellation: Let's take TimeZone.

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -32,24 +32,64 @@
...
-    StringView view(tzStr);
-    auto upconverted = view.upconvertedCharacters();
-    UErrorCode status = U_ZERO_ERROR;
-    auto calendar = std::unique_ptr<UCalendar, ICUDeleter<ucal_close>>(
-        ucal_open(upconverted, view.length(), "", UCAL_DEFAULT, &status));
-    if (U_FAILURE(status)) [[unlikely]]
-        return nullptr;
-    return calendar;
+    String tzId = timeZone.toICUString();
+    TimeZoneCacheEntry& entry = timeZoneCacheEntry(tzId);
```
Constellation: Let's pass TimeZone.

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -30,26 +30,71 @@
...
+    static bool isKeyNull(const TimeZone&) { return false; }
+    static RefPtr<TimeZoneCacheEntry> createValueForNullKey() { return nullptr; }
+    static RefPtr<TimeZoneCacheEntry> createValueForKey(const TimeZone&) { return adoptRef(*new TimeZoneCacheEntry); }
+    static TimeZone createKeyForStorage(const TimeZone& tz) { return tz; }
+};
+
+static RefPtr<TimeZoneCacheEntry> timeZoneCacheEntry(const TimeZone& timeZone)
+{
+    static WTF::Lock cacheLock;
+    static NeverDestroyed<TinyLRUCache<TimeZone, RefPtr<TimeZoneCacheEntry>, 8, TimeZoneLRUCachePolicy>> cache;
```
Constellation: Let's use LazyNeverDestroyed & std::call_once unless the data structure can have constexpr constructor which ensures that initialization is fully happening at compile time.

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -30,26 +30,71 @@
...
+struct TimeZoneLRUCachePolicy {
+    static bool isKeyNull(const TimeZone&) { return false; }
+    static RefPtr<TimeZoneCacheEntry> createValueForNullKey() { return nullptr; }
+    static RefPtr<TimeZoneCacheEntry> createValueForKey(const TimeZone&) { return adoptRef(*new TimeZoneCacheEntry); }
+    static TimeZone createKeyForStorage(const TimeZone& tz) { return tz; }
+};
+
+static RefPtr<TimeZoneCacheEntry> timeZoneCacheEntry(const TimeZone& timeZone)
+{
+    static WTF::Lock cacheLock;
```
Constellation: WTF:: is not necessary

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -30,26 +30,71 @@
...
+    static NeverDestroyed<TinyLRUCache<TimeZone, RefPtr<TimeZoneCacheEntry>, 8, TimeZoneLRUCachePolicy>> cache;
+    WTF::Locker locker { cacheLock };
```
Constellation: WTF:: is not necessary

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -30,26 +30,71 @@
...
+    return cache.get().get(timeZone);
+}
+
+static void ensureTimeZone(TimeZoneCacheEntry& entry, const TimeZone& timeZone)
+{
+    if (entry.initialized.load(std::memory_order_acquire))
+        return;
+    WTF::Locker locker { entry.useLock };
```
Constellation: Ditto

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -30,26 +30,71 @@
...
-    UErrorCode status = U_ZERO_ERROR;
-    auto calendar = std::unique_ptr<UCalendar, ICUDeleter<ucal_close>>(
-        ucal_open(upconverted, view.length(), "", UCAL_DEFAULT, &status));
-    if (U_FAILURE(status)) [[unlikely]]
-        return nullptr;
-    return calendar;
+    auto entry = timeZoneCacheEntry(timeZone);
+    ASSERT(entry);
+    ensureTimeZone(*entry, timeZone);
+    WTF::Locker locker { entry->useLock };
```
Constellation: Ditto

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -30,26 +30,71 @@
...
 
 namespace JSC {
 namespace TemporalCore {
 
-// openCalendarForTimeZone — internal: opens ICU UCalendar for a named (IANA) timezone
-static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> openCalendarForTimeZone(const TimeZone& timeZone)
+class TimeZoneCacheEntry : public ThreadSafeRefCounted<TimeZoneCacheEntry> {
+    WTF_MAKE_TZONE_ALLOCATED(TimeZoneCacheEntry);
+public:
+    WTF::Lock useLock;
```
Constellation: WTF is not necessary

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -30,26 +30,71 @@
...
+{
+    static WTF::Lock cacheLock;
+    static NeverDestroyed<TinyLRUCache<TimeZone, RefPtr<TimeZoneCacheEntry>, 8, TimeZoneLRUCachePolicy>> cache;
+    WTF::Locker locker { cacheLock };
+    return cache.get().get(timeZone);
+}
+
+static void ensureTimeZone(TimeZoneCacheEntry& entry, const TimeZone& timeZone)
+{
+    if (entry.initialized.load(std::memory_order_acquire))
```
Constellation: I'm actually not sure whether this atomic fast path is necessary since we will rake a lock immediately after this function. Why not taking a lock and initialize it & use it?

hyjorc1: Right, removed and updated for `withCalendar` as well.

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
```diff
@@ -100,24 +99,13 @@ static CalendarCacheEntry& calendarCacheEntry(CalendarID calendarId)
...
-
 template<typename F>
 static auto withCalendar(CalendarID calendarId, F&& fn) -> decltype(fn(static_cast<UCalendar*>(nullptr)))
 {
     CalendarCacheEntry& entry = calendarCacheEntry(calendarId);
-    ensureCalendar(entry, calendarId);
-    WTF::Locker locker { entry.useLock };
+    Locker locker { entry.useLock };
+    if (!entry.cal)
+        entry.cal = buildCalendarTemplate(calendarId).release();
```
Constellation: buildCalendarTemplate should take AbstractLocker& to make sure that it is requiring a lock

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -30,26 +30,63 @@
...
 namespace JSC {
 namespace TemporalCore {
 
-// openCalendarForTimeZone — internal: opens ICU UCalendar for a named (IANA) timezone
-static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> openCalendarForTimeZone(const TimeZone& timeZone)
+class TimeZoneCacheEntry : public ThreadSafeRefCounted<TimeZoneCacheEntry> {
+    WTF_MAKE_TZONE_ALLOCATED(TimeZoneCacheEntry);
+public:
+    Lock useLock;
+    UCalendar* cal { nullptr };
```
Constellation: Thus, we should use std::unique_ptr with appropriate deleter since this entry can be removed.

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -30,26 +30,63 @@
...
-    String tzStr = timeZone.toICUString();
-    StringView view(tzStr);
-    auto upconverted = view.upconvertedCharacters();
-    UErrorCode status = U_ZERO_ERROR;
-    auto calendar = std::unique_ptr<UCalendar, ICUDeleter<ucal_close>>(
-        ucal_open(upconverted, view.length(), "", UCAL_DEFAULT, &status));
-    if (U_FAILURE(status)) [[unlikely]]
-        return nullptr;
-    return calendar;
+    auto entry = timeZoneCacheEntry(timeZone);
```
Constellation, line 77: It is possible that the past entry gets removed from LRUCache. So entry should have an appropriate thing to destroy the content.

Source/JavaScriptCore/runtime/temporal/core/CalendarICUBridge.cpp
```diff
@@ -85,8 +85,7 @@ static std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> buildCalendarTemplate(
 }
 
 struct CalendarCacheEntry {
-    WTF::Lock useLock;
-    std::atomic<bool> initialized { false };
+    Lock useLock;
     UCalendar* cal { nullptr };
```
Constellation: I think, in general, it would be great to have std::unique_ptr with appropriate deleter even if it persists.

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
hyjorc1: Right, resolved with `std::unique_ptr<UCalendar, ICUDeleter<ucal_close>> cal`.

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -165,136 +199,138 @@ static TemporalResult<PossibleEpochNanoseconds> getNamedTimeZoneEpochNanoseconds
...
+        const double maxFoldWindowMs = 90'000'000.0; // 25 hours in ms
+        std::optional<ISO8601::ExactTime> secondCandidate;
 
-    // ICU quirk: retry from 1ms later to catch transition-boundary fold case.
-    if (!secondCandidate) {
-        auto tryFoldFromOffset = [&](double probeEpochMs) -> bool {
+        auto tryFoldFromTransition = [&](UTimeZoneTransitionType transType) -> bool {
             UErrorCode status = U_ZERO_ERROR;
-            ucal_setMillis(calendar.get(), probeEpochMs, &status);
+            ucal_setMillis(cal, icuEpochMs, &status);
```
Constellation, line 282: Do we need to do ucal_clear before as we are reusing already obtained calendar?

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -165,136 +199,138 @@ static TemporalResult<PossibleEpochNanoseconds> getNamedTimeZoneEpochNanoseconds
...
+        };
+
+        if (!tryFoldFromTransition(UCAL_TZ_TRANSITION_PREVIOUS))
+            tryFoldFromTransition(UCAL_TZ_TRANSITION_NEXT);
+
+        // ICU quirk: retry from 1ms later to catch transition-boundary fold case.
+        if (!secondCandidate) {
+            auto tryFoldFromOffset = [&](double probeEpochMs) -> bool {
+                UErrorCode status = U_ZERO_ERROR;
+                ucal_setMillis(cal, probeEpochMs, &status);
```
Constellation, line 316: Ditto

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
```diff
@@ -358,78 +395,64 @@ TemporalResult<std::optional<ISO8601::ExactTime>> getTimeZoneTransition(const Ti
...
+
+            auto transition = ISO8601::ExactTime::fromEpochMilliseconds(static_cast<int64_t>(transitionMs));
+            if (!transition.isValid())
+                return std::optional<ISO8601::ExactTime> { std::nullopt };
+
+            // Check if offset actually changed at this transition using a clone.
+            double beforeMs = transitionMs - 1.0;
+            double afterMs = transitionMs;
+            UErrorCode s2 = U_ZERO_ERROR;
+            auto cal2 = std::unique_ptr<UCalendar, ICUDeleter<ucal_close>>(ucal_clone(cal, &s2));
```
Constellation: Do we need this cloning? Why not just clearing the ucal first, set & get data, move and set & get data?

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
hyjorc1: My understanding is that we only need to do `ucal_clear` when building a date from individual fields using `ucal_set(cal, FIELD, value)`. In that case, stale field values from a previous use can
  interfere without clear. Here, for both `tryFoldFromTransition` and `tryFoldFromOffset` only need to read the data after `ucal_setMillis` (establishes the ground truth when no `ucal_set` is used after it). So, I think we don't need `ucal_clear` here.

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
hyjorc1: See https://github.com/WebKit/WebKit/pull/66893#discussion_r3398905461

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
hyjorc1: Thanks for bring it up I've double checked all the other related use sites and ensured no other `ucal_clear` missing.

Source/JavaScriptCore/runtime/temporal/core/TimeZoneICUBridge.cpp
hyjorc1: Good point. Fixed this one. And I also cleared other two `ucal_clone` uses in `CalendarICUBridge.cpp`.

Constellation (approved):

PR #66977, opened 2026-06-11
Navigation API: extract a MethodTrackerRegistry for the upcoming/ongoing tracker slots

basuke:
#### 26559c13662dfdf3815d1e73fef7e3f56314b967
<pre>
Navigation API: extract a MethodTrackerRegistry for the upcoming/ongoing tracker slots
<a href="https://bugs.webkit.org/show_bug.cgi?id=316879">https://bugs.webkit.org/show_bug.cgi?id=316879</a>
<a href="https://rdar.apple.com/179303305">rdar://179303305</a>

Reviewed by Chris Dumez and Rupin Mittal.

m_apiMethodTrackersLock guarded three Navigation member slots that hold
NavigationAPIMethodTrackers (the ongoing tracker, the upcoming
non-traverse slot, and the upcoming-traverse map). All mutation happens
on the main thread; the lock exists only because
visitAdditionalChildrenInGCThread visits the trackers&apos; info() values
from the GC thread.

The lock granularity was correct, but its use was dispersed: about a
dozen call sites hand-rolled &quot;lock, copy RefPtr, unlock&quot; blocks, nothing
prevented new code from touching a slot without the lock, and the
long-standing FIXMEs about ongoing/upcoming-slot invariants had nowhere
to live.

The promoteUpcomingAPIMethodTracker(const String&amp;) algorithm also
encoded the traverse / non-traverse distinction in
null-vs-empty-vs-non-empty String semantics, with three branches:

  if (!destinationKey.isEmpty())          // take from upcoming-traverse
  else if (destinationKey.isNull())       // take from upcoming-non-traverse
  else if (destinationKey.isEmpty() &amp;&amp; !upcoming-traverse.isEmpty())
                                          // take HashMap first entry

The third branch was unreachable: NavigationDestination::key() returns
either a UUID (non-null, non-empty) or null, never a non-null empty
string, and even if reachable its HashMap-iteration-order semantics
were unspecified-behavior. The second branch was also a latent spec
deviation: if a navigation.navigate() call had registered an upcoming
non-traverse tracker, and a browser-driven cross-document Traverse then
arrived whose target HistoryItem wasn&apos;t in m_entries (so destination
key is null), the second branch would steal the non-traverse tracker
to satisfy the traversal — wiring the navigate()&apos;s committed/finished
promises to an unrelated traversal&apos;s outcome. That has been removed
along with the rest of the old dispatch.

Move the slots, the lock, and the spec&apos;s slot algorithms (promote-an-
upcoming-api-method-tracker-to-ongoing, navigation-api-method-tracker-
clean-up, the upcoming-traverse lookup, the GC-thread visit) into a
private nested class, MethodTrackerRegistry. Navigation keeps tracker
*creation* (it needs the global object and the promises); the registry
only stores, promotes, looks up, and cleans up. Its API:

    setUpcomingNonTraverse(Ref&lt;NavigationAPIMethodTracker&gt;&amp;&amp;);
    addUpcomingTraverse(const String&amp; key, Ref&lt;NavigationAPIMethodTracker&gt;&amp;&amp;);
    upcomingTraverse(const String&amp; key) const -&gt; RefPtr&lt;...&gt;;
    ongoing() const -&gt; RefPtr&lt;...&gt;;
    takeUpcomingNonTraverseIfEquals(NavigationAPIMethodTracker&amp;) -&gt; RefPtr&lt;...&gt;;
    promoteUpcomingNonTraverseToOngoing() -&gt; RefPtr&lt;...&gt;;
    promoteUpcomingTraverseToOngoing(const String&amp; destinationKey) -&gt; RefPtr&lt;...&gt;;
    unregister(NavigationAPIMethodTracker&amp;);
    isEmpty() const -&gt; bool;
    visitInGCThread(JSC::AbstractSlotVisitor&amp;) const;

The traverse / non-traverse distinction is now encoded as two methods
the caller picks based on its existing navigationType discriminator,
instead of a runtime branch on a String&apos;s null/empty state. The two
promote methods return the newly-promoted tracker, so the call site
inside innerDispatchNavigateEvent gets the ongoing tracker without a
second lock acquisition. takeUpcomingNonTraverseIfEquals returns the
taken tracker (or null) instead of a bool, matching the rest of the
registry&apos;s take→RefPtr shape. The pure slot-op methods on Navigation
(promoteUpcomingAPIMethodTracker, cleanupAPIMethodTracker) are removed;
their call sites now invoke the registry directly. The two facades
that still mix tracker creation with policy
(maybeSetUpcomingNonTraversalTracker, addUpcomingTraverseAPIMethodTracker)
stay on Navigation and forward storage to the registry.

upcomingTraverseMethodTracker(const String&amp;) — the public accessor used
by NavigationScheduler — now returns RefPtr instead of a raw pointer.
The returned pointer was previously fetched under the lock and used
after the lock released; the existing caller already assigned to a
local RefPtr at the call site, so the change is type-compatible and
makes lifetime safe by construction.

All Locker use and WTF_GUARDED_BY_LOCK annotations now live in one
screenful inside the registry. The pre-existing FIXME assertions about
the ongoing-tracker and upcoming-traverse-key invariants also moved
into the registry&apos;s promote / unregister paths.

This is preparation for the upcoming NavigationInterceptOptions
precommitHandler work: with the registry in place, precommit support
reduces to one more tracker state (added in 316812@main) and one more
registry slot, instead of another round of flags threaded through
Navigation.

Behavior change limited to the previously-described latent fix in the
null-key Traverse path; existing imported/w3c/web-platform-tests/
navigation-api tests pass in both release and debug.

* Source/WebCore/page/Navigation.h:
(WebCore::Navigation::MethodTrackerRegistry):
* Source/WebCore/page/Navigation.cpp:
(WebCore::Navigation::maybeSetUpcomingNonTraversalTracker):
(WebCore::Navigation::addUpcomingTraverseAPIMethodTracker):
(WebCore::Navigation::navigate):
(WebCore::Navigation::performTraversal):
(WebCore::Navigation::resolveFinishedPromise):
(WebCore::Navigation::rejectFinishedPromise):
(WebCore::Navigation::notifyCommittedToEntry):
(WebCore::Navigation::updateForNavigation):
(WebCore::Navigation::abortOngoingNavigation):
(WebCore::Navigation::handleSameDocumentNavigation):
(WebCore::Navigation::innerDispatchNavigateEvent):
(WebCore::Navigation::visitAdditionalChildrenInGCThread):
(WebCore::Navigation::MethodTrackerRegistry::setUpcomingNonTraverse):
(WebCore::Navigation::MethodTrackerRegistry::addUpcomingTraverse):
(WebCore::Navigation::MethodTrackerRegistry::upcomingTraverse):
(WebCore::Navigation::MethodTrackerRegistry::ongoing):
(WebCore::Navigation::MethodTrackerRegistry::takeUpcomingNonTraverseIfEquals):
(WebCore::Navigation::MethodTrackerRegistry::promoteUpcomingNonTraverseToOngoing):
(WebCore::Navigation::MethodTrackerRegistry::promoteUpcomingTraverseToOngoing):
(WebCore::Navigation::MethodTrackerRegistry::unregister):
(WebCore::Navigation::MethodTrackerRegistry::isEmpty):
(WebCore::Navigation::MethodTrackerRegistry::visitInGCThread):
(WebCore::Navigation::upcomingTraverseMethodTracker):
(WebCore::Navigation::cleanupAPIMethodTracker): Deleted.
(WebCore::Navigation::promoteUpcomingAPIMethodTracker): Deleted.

Canonical link: <a href="https://commits.webkit.org/315301@main">https://commits.webkit.org/315301@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/bdef8fda1a7e0ab1070ae7b351a7a87722ffdecb

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/168478 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/42035 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/35624 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/177806 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/126179 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/2d63c81c-cd6f-407a-a796-38fe4586bef6) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/170344 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/42411 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/41997 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/131134 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/92229 "6 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/aff90345-3787-4cbc-a4c4-4754cffc3c76) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/171437 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/33593 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/151407 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/111645 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/7dc6f6f4-5e0d-4d34-83c7-7cf6fafec8f1) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/32579 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/31282 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/26502 "Built successfully") | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/142565 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/30811 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/180406 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/33130 "Built successfully") | [❌ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/35445 "Found 1 new test failure: fast/speechsynthesis/speech-synthesis-utterance-gc.html (failure)") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/139572 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/42047 "Built successfully") | | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/139433 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/37585 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/42735 "Built successfully") | [❌ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/27784 "Too many flaky failures: http/tests/contentextensions/video-element-resource-type.html, http/tests/navigation/redirect-preserves-fragment.html, http/tests/resourceLoadStatistics/downgraded-referrer-for-navigation-with-link-query-from-prevalent-resource.html, http/tests/security/cached-cross-origin-preloaded-css-stylesheet.html, http/tests/security/contentSecurityPolicy/1.1/stylehash-multiple-policies.html, http/tests/security/contentSecurityPolicy/report-only.py, http/tests/security/mixedContent/mainframe-onload-after-iframe-block.https.html, http/tests/site-isolation/frame-index.html, http/tests/storageAccess/deny-storage-access-under-opener-if-auto-dismiss-ephemeral.html, http/tests/webgpu/webgpu/web_platform/copyToTexture/image_file.html (failure)") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/106388 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/34717 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/114495 "Built successfully") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/41239 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/5867 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/40636 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/40887 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/40781 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/page/Navigation.cpp
+++ b/Source/WebCore/page/Navigation.cpp
@@ -420,13 +420,10 @@ ExceptionOr<RefPtr<SerializedScriptValue>> Navigation::serializeState(JSC::JSVal
 // https://html.spec.whatwg.org/multipage/nav-history-apis.html#maybe-set-the-upcoming-non-traverse-api-method-tracker
 RefPtr<NavigationAPIMethodTracker> Navigation::maybeSetUpcomingNonTraversalTracker(JSC::JSGlobalObject& globalObject, Ref<DeferredPromise>&& committed, Ref<DeferredPromise>&& finished, JSC::JSValue info, RefPtr<SerializedScriptValue>&& serializedState)
 {
-    RefPtr apiMethodTracker = NavigationAPIMethodTracker::create(globalObject, WTF::move(committed), WTF::move(finished), WTF::move(info), WTF::move(serializedState));
+    Ref apiMethodTracker = NavigationAPIMethodTracker::create(globalObject, WTF::move(committed), WTF::move(finished), WTF::move(info), WTF::move(serializedState));
 
-    // FIXME: We should be able to assert m_upcomingNonTraverseMethodTracker is empty.
-    if (!hasEntriesAndEventsDisabled()) {
-        Locker locker { m_apiMethodTrackersLock };
-        m_upcomingNonTraverseMethodTracker = apiMethodTracker;
-    }
+    if (!hasEntriesAndEventsDisabled())
+        m_methodTrackers.setUpcomingNonTraverse(apiMethodTracker.copyRef());
 
     return apiMethodTracker;
 }
@@ -434,13 +431,10 @@ RefPtr<NavigationAPIMethodTracker> Navigation::maybeSetUpcomingNonTraversalTrack
 // https://html.spec.whatwg.org/multipage/nav-history-apis.html#add-an-upcoming-traverse-api-method-tracker
 RefPtr<NavigationAPIMethodTracker> Navigation::addUpcomingTraverseAPIMethodTracker(JSC::JSGlobalObject& globalObject, Ref<DeferredPromise>&& committed, Ref<DeferredPromise>&& finished, const String& key, JSC::JSValue info)
 {
-    RefPtr apiMethodTracker = NavigationAPIMethodTracker::create(globalObject, WTF::move(committed), WTF::move(finished), WTF::move(info), nullptr);
+    Ref apiMethodTracker = NavigationAPIMethodTracker::create(globalObject, WTF::move(committed), WTF::move(finished), WTF::move(info), nullptr);
     apiMethodTracker->setKey(key);
 
-    {
-        Locker locker { m_apiMethodTrackersLock };
-        m_upcomingTraverseMethodTrackers.add(key, *apiMethodTracker);
-    }
+    m_methodTrackers.addUpcomingTraverse(key, apiMethodTracker.copyRef());
 
     return apiMethodTracker;
 }
@@ -516,16 +510,9 @@ Navigation::Result Navigation::navigate(JSC::JSGlobalObject& globalObject, const
     request.setIsFromNavigationAPI(true);
     frame()->loader().loadFrameRequest(WTF::move(request), nullptr, { });
 
-    // If the load() call never made it to the point that NavigateEvent was emitted, thus promoteUpcomingAPIMethodTracker() called, this will be true.
-    bool loadWasAborted = [&] {
-        Locker locker { m_apiMethodTrackersLock };
-        if (m_upcomingNonTraverseMethodTracker != apiMethodTracker)
-            return false;
-        m_upcomingNonTraverseMethodTracker = nullptr;
-        return true;
-    }();
-    if (loadWasAborted)
-        rejectFinishedPromise(apiMethodTracker.get());
+    // If the load() call never made it to the point that NavigateEvent was emitted, thus the upcoming-non-traverse slot was never promoted, this returns the tracker so we can reject it.
+    if (RefPtr aborted = m_methodTrackers.takeUpcomingNonTraverseIfEquals(*apiMethodTracker))
+        rejectFinishedPromise(aborted);
 
     return apiMethodTrackerDerivedResult(*apiMethodTracker);
 }
@@ -551,11 +538,8 @@ Navigation::Result Navigation::performTraversal(JSC::JSGlobalObject& globalObjec
         return { createDOMPromise(committed), createDOMPromise(finished) };
     }
 
-    {
-        Locker locker { m_apiMethodTrackersLock };
-        if (auto existingMethodTracker = m_upcomingTraverseMethodTrackers.getOptional(key))
-            return apiMethodTrackerDerivedResult(*existingMethodTracker);
-    }
+    if (RefPtr existingMethodTracker = m_methodTrackers.upcomingTraverse(key))
+        return apiMethodTrackerDerivedResult(*existingMethodTracker);
 
     RefPtr apiMethodTracker = addUpcomingTraverseAPIMethodTracker(globalObject, WTF::move(committed), WTF::move(finished), key, options.info);
 
@@ -564,7 +548,7 @@ Navigation::Result Navigation::performTraversal(JSC::JSGlobalObject& globalObjec
         if (result != ScheduleHistoryNavigationResult::Aborted)
             return;
 
-        // Going through rejectFinishedPromise also removes the tracker from m_upcomingTraverseMethodTrackers;
+        // Going through rejectFinishedPromise also removes the tracker from the upcoming-traverse map;
         // otherwise a later traversal to the same key would be deduplicated against the settled promises.
         if (RefPtr protectedThis = weakThis.get())
             protectedThis->rejectFinishedPromise(apiMethodTracker.get());
@@ -673,7 +657,7 @@ void Navigation::resolveFinishedPromise(NavigationAPIMethodTracker* apiMethodTra
 {
     apiMethodTracker->resolveFinished();
     if (apiMethodTracker->isSettled())
-        cleanupAPIMethodTracker(apiMethodTracker);
+        m_methodTrackers.unregister(*apiMethodTracker);
 }
 
 // https://html.spec.whatwg.org/multipage/nav-history-apis.html#reject-the-finished-promise
@@ -682,7 +666,7 @@ void Navigation::rejectFinishedPromise(NavigationAPIMethodTracker* apiMethodTrac
     RELEASE_LOG(Navigation, "rejectFinishedPromise: rejecting promises for tracker=%p with exception='%s'", apiMethodTracker, exception.message().utf8().data());
 
     apiMethodTracker->rejectFinished(exception, exceptionObject);
-    cleanupAPIMethodTracker(apiMethodTracker);
+    m_methodTrackers.unregister(*apiMethodTracker);
 }
 
 void Navigation::rejectFinishedPromise(NavigationAPIMethodTracker* apiMethodTracker)
@@ -712,7 +696,7 @@ void Navigation::notifyCommittedToEntry(NavigationAPIMethodTracker* apiMethodTra
 
     apiMethodTracker->commitTo(*entry, navigationType);
     if (apiMethodTracker->isSettled())
-        cleanupAPIMethodTracker(apiMethodTracker);
+        m_methodTrackers.unregister(*apiMethodTracker);
 }
 
 void Navigation::updateNavigationEntry(Ref<HistoryItem>&& item, ShouldCopyStateObjectFromCurrentEntry shouldCopyStateObjectFromCurrentEntry)
@@ -826,11 +810,7 @@ void Navigation::updateForNavigation(Ref<HistoryItem>&& item, NavigationNavigati
             Ref { m_entries[*m_currentEntryIndex] }->setState(oldCurrentEntry->state());
     }
 
-    RefPtr<NavigationAPIMethodTracker> ongoingAPIMethodTracker;
-    {
-        Locker locker { m_apiMethodTrackersLock };
-        ongoingAPIMethodTracker = m_ongoingAPIMethodTracker;
-    }
+    RefPtr ongoingAPIMethodTracker = m_methodTrackers.ongoing();
     if (ongoingAPIMethodTracker && shouldNotifyCommitted == ShouldNotifyCommitted::Yes)
         notifyCommittedToEntry(ongoingAPIMethodTracker.get(), protect(currentEntry()).get(), navigationType);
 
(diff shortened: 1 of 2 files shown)


Source/WebCore/page/Navigation.h
```diff
@@ -291,6 +289,33 @@ class Navigation final : public RefCounted<Navigation>, public EventTarget, publ
...
+        RefPtr<NavigationAPIMethodTracker> promoteUpcomingTraverseToOngoing(const String& destinationKey) WTF_EXCLUDES_LOCK(m_lock);
+
+        // https://html.spec.whatwg.org/multipage/nav-history-apis.html#navigation-api-method-tracker-clean-up
+        void unregister(NavigationAPIMethodTracker&) WTF_EXCLUDES_LOCK(m_lock);
+
+        bool isEmpty() const WTF_EXCLUDES_LOCK(m_lock);
+        void visitInGCThread(JSC::AbstractSlotVisitor&) const WTF_EXCLUDES_LOCK(m_lock);
+
+    private:
+        mutable Lock m_lock;
```
jelee53, line 313: Nit: the previous lock was named `m_apiMethodTrackersLock` so maybe we could have some more specific and expressive naming here? eg: `m_methodTrackerRegistryLock`?

Source/WebCore/page/Navigation.cpp
```diff
@@ -516,16 +510,9 @@ Navigation::Result Navigation::navigate(JSC::JSGlobalObject& globalObject, const
...
-        if (m_upcomingNonTraverseMethodTracker != apiMethodTracker)
-            return false;
-        m_upcomingNonTraverseMethodTracker = nullptr;
-        return true;
-    }();
-    if (loadWasAborted)
-        rejectFinishedPromise(apiMethodTracker.get());
+    // If the load() call never made it to the point that NavigateEvent was emitted, thus the upcoming-non-traverse slot was never promoted, this returns the tracker so we can reject it.
+    if (RefPtr aborted = m_methodTrackers.takeUpcomingNonTraverseIfEquals(*apiMethodTracker))
+        rejectFinishedPromise(aborted.get());
```
jelee53: Nit: do we need the `get()`? It might implicitly convert to a raw pointer.

jelee53:
Nit: 

In the commit description, `All Locker use and WTF_GUARDED_BY_LOCK annotations now live in one screenful inside the registry.`  I wonder if this is a typo? Not sure what a `screenful` is.  

But the commit description is very detailed and it was helpful to read.

Source/WebCore/page/Navigation.h
basuke: No, that's not good. This is intentional. The primary goal was to extract a dedicated class specifically for lock management, so there is no need to give it a redundant or overly verbose name. With ideal design, each class should have only one or two locks (for read and write). In reality, there're so many non-ideal cases in the wild :p

Source/WebCore/page/Navigation.cpp
```diff
@@ -915,46 +895,99 @@ static bool documentCanHaveURLRewritten(const Document& document, const URL& tar
...
+    m_ongoing = WTF::move(m_upcomingNonTraverse);
+    return m_ongoing;
+}
+
+RefPtr<NavigationAPIMethodTracker> Navigation::MethodTrackerRegistry::promoteUpcomingTraverseToOngoing(const String& destinationKey)
+{
+    assertIsMainThread();
+    Locker locker { m_lock };
+    // FIXME: We should be able to assert m_ongoing is unset.
+    m_ongoing = m_upcomingTraverse.take(destinationKey);
```
cdumez, line 952: I think we may have to check that `destinationKey` is non null before we call take() here, to avoid crashes. The caller passes `NavigationDestination::key()` which is `m_entry ? m_entry->key() : String()` so it looks a little fragile to not handle a null String.

cdumez (approved):
LGTM with suggestion.

Source/WebCore/page/Navigation.cpp
basuke: Thanks! Since I want to keep this PR strictly limited to mechanical moves, I'm planning to address those FIXMEs in a subsequent PR. Let's go ahead and land this one as it is.

cdumez: I don't think I am OK with this given that my finding seems to be a regression from your PR. `Navigation::promoteUpcomingAPIMethodTracker()` was checking `destinationKey.isEmpty()` as far as I can tell before your change. It looks to me like this check was lost in your code reshuffle, which is why I was pointing it out.

basuke: Sorry, I should have surfaced this in the code rather than only in the commit message. Both deletions were intentional — see paragraphs 3-4 of the commit message:

- **Empty non-null branch**: unreachable. `NavigationDestination::key()` returns either a UUID or null, never `""`.
- **Null branch**: a latent spec deviation. It would steal an unrelated `navigation.navigate()`'s upcoming non-traverse tracker to satisfy a browser-driven Traverse to an untracked HistoryItem, wiring that `navigate()`'s promises to the traversal's outcome.

On the crash worry: `HashMap<String, ...>::take(nullString)` is safe — null is the `StringHash` empty-sentinel, so the lookup short-circuits and returns a default `RefPtr`.

I've amended to pin the invariant in code:

```cpp
ASSERT(destinationKey.isNull() || !destinationKey.isEmpty());
m_ongoing = m_upcomingTraverse.take(destinationKey);
```

basuke: I apologize that some functional code changes slipped in while I was supposedly just doing a mechanical refactoring. The logic had become unnecessarily complex, so I went ahead and cleaned it up.

Source/WebCore/page/Navigation.h
```diff
@@ -188,7 +188,7 @@ class Navigation final : public RefCounted<Navigation>, public EventTarget, publ
     ScriptExecutionContext* scriptExecutionContext() const final;
 
     void rejectFinishedPromise(NavigationAPIMethodTracker*);
-    NavigationAPIMethodTracker* upcomingTraverseMethodTracker(const String& key) const;
+    RefPtr<NavigationAPIMethodTracker> upcomingTraverseMethodTracker(const String& key) const;
```
RupinMittal: Nit: Is this necessary? Looks like it would be fine to return a raw pointer here given that we don't usually return smart pointers from functions.

Source/WebCore/page/Navigation.cpp
RupinMittal: Agreed, I believe it should implicitly convert so we shouldn't need it.

Source/WebCore/page/Navigation.cpp
```diff
@@ -420,27 +420,21 @@ ExceptionOr<RefPtr<SerializedScriptValue>> Navigation::serializeState(JSC::JSVal
...
-    RefPtr apiMethodTracker = NavigationAPIMethodTracker::create(globalObject, WTF::move(committed), WTF::move(finished), WTF::move(info), WTF::move(serializedState));
+    Ref apiMethodTracker = NavigationAPIMethodTracker::create(globalObject, WTF::move(committed), WTF::move(finished), WTF::move(info), WTF::move(serializedState));
 
-    // FIXME: We should be able to assert m_upcomingNonTraverseMethodTracker is empty.
-    if (!hasEntriesAndEventsDisabled()) {
-        Locker locker { m_apiMethodTrackersLock };
-        m_upcomingNonTraverseMethodTracker = apiMethodTracker;
-    }
+    if (!hasEntriesAndEventsDisabled())
+        m_methodTrackers.setUpcomingNonTraverse(apiMethodTracker.copyRef());
```
RupinMittal, line 426: Nit: Use `WTF::move` instead of `copyRef` to avoid churn.

Source/WebCore/page/Navigation.cpp
```diff
@@ -420,27 +420,21 @@ ExceptionOr<RefPtr<SerializedScriptValue>> Navigation::serializeState(JSC::JSVal
...
 {
-    RefPtr apiMethodTracker = NavigationAPIMethodTracker::create(globalObject, WTF::move(committed), WTF::move(finished), WTF::move(info), nullptr);
+    Ref apiMethodTracker = NavigationAPIMethodTracker::create(globalObject, WTF::move(committed), WTF::move(finished), WTF::move(info), nullptr);
     apiMethodTracker->setKey(key);
 
-    {
-        Locker locker { m_apiMethodTrackersLock };
-        m_upcomingTraverseMethodTrackers.add(key, *apiMethodTracker);
-    }
+    m_methodTrackers.addUpcomingTraverse(key, apiMethodTracker.copyRef());
```
RupinMittal, line 437: Same here.

Source/WebCore/page/Navigation.h
```diff
@@ -291,6 +289,33 @@ class Navigation final : public RefCounted<Navigation>, public EventTarget, publ
...
+        void setUpcomingNonTraverse(Ref<NavigationAPIMethodTracker>&&) WTF_EXCLUDES_LOCK(m_lock);
+        void addUpcomingTraverse(const String& key, Ref<NavigationAPIMethodTracker>&&) WTF_EXCLUDES_LOCK(m_lock);
+        RefPtr<NavigationAPIMethodTracker> upcomingTraverse(const String& key) const WTF_EXCLUDES_LOCK(m_lock);
+        RefPtr<NavigationAPIMethodTracker> ongoing() const WTF_EXCLUDES_LOCK(m_lock);
+
+        RefPtr<NavigationAPIMethodTracker> takeUpcomingNonTraverseIfEquals(NavigationAPIMethodTracker&) WTF_EXCLUDES_LOCK(m_lock);
+
+        // https://html.spec.whatwg.org/multipage/nav-history-apis.html#promote-an-upcoming-api-method-tracker-to-ongoing
+        RefPtr<NavigationAPIMethodTracker> promoteUpcomingNonTraverseToOngoing() WTF_EXCLUDES_LOCK(m_lock);
+        RefPtr<NavigationAPIMethodTracker> promoteUpcomingTraverseToOngoing(const String& destinationKey) WTF_EXCLUDES_LOCK(m_lock);
```
RupinMittal: Nit: I believe we commonly return raw pointers and leave it up to the caller to protect them.

RupinMittal (approved):
LGTM with nits.

Source/WebCore/page/Navigation.h
basuke: Right, except the method that takes out the value. In this case, `takeUpcomingNonTraverseIfEquals` should return `RefPtr` to hold the reference until it is stored in the call site. But others are okay to be raw pointer. Thanks.

Source/WebCore/page/Navigation.cpp
basuke: But we return `apiMethodTracker` after that. We can store war value before but that also makes another churn. I think it okay to use `copyRef()` here.

Source/WebCore/page/Navigation.cpp
basuke: Ditto.

Source/WebCore/page/Navigation.cpp
RupinMittal: Oh good point! `copyRef()` is good then.

PR #67037, opened 2026-06-12
CSS object-view-box: implement rendering and natural sizes

tangalice:
#### 0d63417844a1f3efbec6c8cece36bf78d39c9263
<pre>
CSS object-view-box: implement rendering and natural sizes
<a href="https://bugs.webkit.org/show_bug.cgi?id=316882">https://bugs.webkit.org/show_bug.cgi?id=316882</a>
<a href="https://rdar.apple.com/179304860">rdar://179304860</a>

Reviewed by Simon Fraser.

Implements rendering and natural size computation for the CSS object-view-box
property across replaced elements (img, canvas, video) and view-transition
pseudo-elements.

For replaced elements (RenderImage, RenderHTMLCanvas, RenderVideo):
- resolveObjectViewBox() computes the view-box rect from inset/rect/xywh values
- replacedContentRect() uses the view-box dimensions as the effective intrinsic
  size for object-fit/object-position calculations
- paintReplaced() draws the full image scaled/translated so the view-box region
  maps onto the replaced content rect, clipped to the content box

For view-transition captures (RenderViewTransitionCapture):
- paintReplaced() uses an explicit srcRect to draw only the view-box sub-region
  of the snapshot, preventing bilinear bleed at the view-box boundary
- image() returns null when object-view-box is set, routing rendering through
  the software paint path (paintReplaced) rather than CA direct compositing
- layout() computes m_viewBoxLocalContentsRect for the CA contentsRect geometry
- canUseExistingLayers() accounts for object-view-box when deciding whether to
  reuse existing CA layers

Layout tests:
- 36/36 css-images object-view-box WPT tests pass, plus 2 new tests added by
  this patch (object-view-box-auto-size-relayout.html, object-view-box-zoom.html)
- 4/4 css-view-transitions object-view-box WPT tests pass (object-view-box-old/
  new-image use scale3d in expected refs and maxDifference=85/totalPixels=600
  fuzzy tolerance to account for snapshot-vs-live rendering quality difference;
  old/new-content-with-object-view-box.html are skipped on iOS due to a 1px
  device-pixel rounding difference)
- Adds 2 new fast/css repaint tests verifying the opaque-foreground optimization
  still behaves correctly with subset/superset object-view-box insets
- Removes 25 ImageOnlyFailure entries from TestExpectations

* LayoutTests/fast/css/object-view-box-subset-obscures-background-no-repaint-expected.txt: Added.
* LayoutTests/fast/css/object-view-box-subset-obscures-background-no-repaint.html: Added.
* LayoutTests/fast/css/object-view-box-superset-does-not-obscure-background-repaint-expected.txt: Added.
* LayoutTests/fast/css/object-view-box-superset-does-not-obscure-background-repaint.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-auto-size-relayout-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-auto-size-relayout-ref.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-auto-size-relayout.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-zoom-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-zoom-ref.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-zoom.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-view-transitions/object-view-box-new-image-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-view-transitions/object-view-box-new-image.html:
* LayoutTests/imported/w3c/web-platform-tests/css/css-view-transitions/object-view-box-old-image-expected.html: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-view-transitions/object-view-box-old-image.html:
* LayoutTests/platform/wk2/TestExpectations:
* Source/WebCore/platform/graphics/GeometryUtilities.cpp:
* Source/WebCore/platform/graphics/GeometryUtilities.h:
* Source/WebCore/rendering/RenderHTMLCanvas.cpp:
* Source/WebCore/rendering/RenderImage.cpp:
* Source/WebCore/rendering/RenderReplaced.cpp:
* Source/WebCore/rendering/RenderReplaced.h:
* Source/WebCore/rendering/RenderVideo.cpp:
* Source/WebCore/rendering/RenderViewTransitionCapture.cpp:
* Source/WebCore/rendering/RenderViewTransitionCapture.h:
* Source/WebCore/rendering/updating/RenderTreeUpdaterViewTransition.cpp:
* Source/WebCore/style/StyleDifference.cpp:

Canonical link: <a href="https://commits.webkit.org/317112@main">https://commits.webkit.org/317112@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/a3ae504276d4953fe627197ee8257669bd01a4eb

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/174345 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/48278 "Built successfully") | [  ~~🛠 mac~~](https://ews-build.webkit.org/#/builders/168/builds/42059 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/183921 "Built successfully") | [  ~~🛠 win~~](https://ews-build.webkit.org/#/builders/59/builds/132213 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/176210 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/49570 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/47960 "Built successfully") | [  ~~🧪 wpe-wk2~~](https://ews-build.webkit.org/#/builders/34/builds/135623 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/59/builds/132213 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/177303 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/162/builds/39050 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac~~](https://ews-build.webkit.org/#/builders/18/builds/156415 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-wpe~~](https://ews-build.webkit.org/#/builders/41/builds/116101 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/37691 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/36061 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/32403 "Built successfully") | 
| | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/148114 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-wk2~~](https://ews-build.webkit.org/#/builders/169/builds/35907 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/186347 "Built successfully") | 
| | [  ~~🛠 ios-safer-cpp~~](https://ews-build.webkit.org/#/builders/174/builds/39552 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 mac-AS-debug-wk2~~](https://ews-build.webkit.org/#/builders/161/builds/40258 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🧪 gtk-wk2~~](https://ews-build.webkit.org/#/builders/1/builds/144531 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/47945 "Built successfully") | [  ~~🧪 mac-wk2-stress~~](https://ews-build.webkit.org/#/builders/8/builds/155969 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/144393 "Passed tests") | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/48624 "Built successfully") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/32527 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/114166 "The change is no longer eligible for processing.") | 
| [✅ 🛠 🧪 unsafe-merge](https://ews-build.webkit.org/#/builders/22/builds/26500 "Built successfully and passed tests") | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/39291 "Passed tests") | [  ~~🛠 mac-safer-cpp~~](https://ews-build.webkit.org/#/builders/120/builds/120562 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/47130 "Built successfully") | [  ~~🧪 mac-site-isolation~~](https://ews-build.webkit.org/#/builders/185/builds/10871 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/46533 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/46764 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/46591 "Built successfully") | | |

### DIFF
--- a/LayoutTests/TestExpectations
+++ b/LayoutTests/TestExpectations
@@ -4498,33 +4498,6 @@ imported/w3c/web-platform-tests/css/css-images/object-position-svg-001e.html [ I
 imported/w3c/web-platform-tests/css/css-images/object-position-svg-001o.html [ ImageOnlyFailure ]
 imported/w3c/web-platform-tests/css/css-images/object-position-svg-002e.html [ ImageOnlyFailure ]
 imported/w3c/web-platform-tests/css/css-images/object-position-svg-002o.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-contain-intrinsic-size.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-contain-canvas.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-contain-img.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-contain-svg.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-contain-video.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-cover-canvas.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-cover-img.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-cover-svg.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-cover-video.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-fill-canvas.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-fill-img.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-fill-svg.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-fill-video.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-none-canvas.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-none-img.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-none-svg.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-fit-none-video.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-property-changed.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-rect-percentage.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-rect.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-size-containment.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-writing-mode-canvas.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-writing-mode-img.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-writing-mode-svg.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-writing-mode-video.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-xywh-percentage.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-images/object-view-box-xywh.html [ ImageOnlyFailure ]
 imported/w3c/web-platform-tests/css/css-images/tiled-gradients.html [ ImageOnlyFailure ]
 
 # These need fuzziness.
@@ -7163,7 +7136,6 @@ imported/w3c/web-platform-tests/navigation-api/scroll-behavior/after-transition-
 imported/w3c/web-platform-tests/navigation-api/scroll-behavior/manual-scroll-reload.html [ Skip ]
 
 # -- View Transitions -- #
-
 imported/w3c/web-platform-tests/css/css-view-transitions/pseudo-element-overflow-clip-with-border-radius.html [ ImageOnlyFailure ]
 
 # Timeouts
@@ -7200,12 +7172,6 @@ imported/w3c/web-platform-tests/css/css-view-transitions/scoped/ [ Skip ]
 
 # waitUntil() method.
 imported/w3c/web-platform-tests/css/css-view-transitions/view-transition-waituntil-animation-manipulation.html [ ImageOnlyFailure ]
-
-# Depends on object-view-box (unimplemented).
-imported/w3c/web-platform-tests/css/css-view-transitions/old-content-with-object-view-box.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-view-transitions/object-view-box-new-image.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-view-transitions/object-view-box-old-image.html [ ImageOnlyFailure ]
-imported/w3c/web-platform-tests/css/css-view-transitions/new-content-with-object-view-box.html [ ImageOnlyFailure ]
 
 # -- END: View transitions -- #
 
--- a/LayoutTests/fast/css/object-view-box-subset-obscures-background-no-repaint.html
+++ b/LayoutTests/fast/css/object-view-box-subset-obscures-background-no-repaint.html
@@ -1,0 +1,52 @@
+<!DOCTYPE html> <!-- webkit-test-runner [ CSSObjectViewBoxEnabled=true ] -->
+<html>
+<head>
+<script>jsTestIsAsync = true;</script>
+<script src="../../resources/js-test.js"></script>
+<style>
+    img {
+        display: block;
+        width: 100px;
+        height: 100px;
+        object-fit: fill;
+        background-image: url(../repaint/resources/animated.gif);
+        /* Positive insets: the view box is a subset of the natural image, so
+           the opaque foreground still fully covers the background in every
+           frame. RenderImage::foregroundIsKnownToBeOpaqueInRect() should
+           still recognize this and suppress the animated background's
+           repeated repaint, just like it does with no object-view-box at all. */
+        object-view-box: inset(10px 10px 10px 10px);
+    }
+</style>
+</head>
+<body>
+<img src="../repaint/resources/apple.jpg">
+<script>
+    description("A subset object-view-box (all-positive insets) should not prevent the opaque-foreground optimization from suppressing an obscured animated background's repaint. This test requires DRT.");
+
+    async function start()
+    {
+        if (!window.testRunner || !window.internals) {
+            finishJSTest();
+            return;
+        }
+
+        document.body.offsetTop;
+        await testRunner.displayAndTrackRepaints();
+        setTimeout(logRepaints, 200);
+    }
+
+    function logRepaints()
+    {
+        repaintRects = window.internals.repaintRectsAsText();
+        window.internals.stopTrackingRepaints();
+
+        shouldBeEqualToString("repaintRects", "");
+
+        finishJSTest();
+    }
+
+    window.addEventListener("load", start);
+</script>
+</body>
+</html>
--- a/LayoutTests/fast/css/object-view-box-superset-does-not-obscure-background-repaint.html
+++ b/LayoutTests/fast/css/object-view-box-superset-does-not-obscure-background-repaint.html
@@ -1,0 +1,52 @@
+<!DOCTYPE html> <!-- webkit-test-runner [ CSSObjectViewBoxEnabled=true ] -->
+<html>
+<head>
+<script>jsTestIsAsync = true;</script>
+<script src="../../resources/js-test.js"></script>
+<style>
+    img {
+        display: block;
+        width: 100px;
+        height: 100px;
+        object-fit: fill;
+        background-image: url(../repaint/resources/animated.gif);
+        /* Negative left inset: the view box is a superset of the natural
+           image, so the painted content is scaled down and leaves a real
+           gap where the background shows through. The opaque-foreground
+           optimization must not suppress the animated background's repaint
+           in this case, or the gap would go stale. */
+        object-view-box: inset(0px 0px 0px -50px);
+    }
+</style>
+</head>
+<body>
+<img src="../repaint/resources/apple.jpg">
+<script>
+    description("A superset object-view-box (negative inset) must not let the opaque-foreground optimization suppress an obscured animated background's repaint, since the painted content leaves real gaps. This test requires DRT.");
+
+    async function start()
+    {
+        if (!window.testRunner || !window.internals) {
+            finishJSTest();
+            return;
+        }
+
+        document.body.offsetTop;
+        await testRunner.displayAndTrackRepaints();
+        setTimeout(logRepaints, 200);
+    }
+
+    function logRepaints()
+    {
+        repaintRects = window.internals.repaintRectsAsText();
+        window.internals.stopTrackingRepaints();
+
+        shouldNotBeEqualToString("repaintRects", "");
+
+        finishJSTest();
+    }
+
+    window.addEventListener("load", start);
+</script>
+</body>
+</html>
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-auto-size-relayout-ref.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-auto-size-relayout-ref.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+<style>
+  div {
+    display: block;
+    width: 100px;
+    height: 50px;
+    background-color: red;
+  }
+</style>
+<div></div>
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-auto-size-relayout.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-auto-size-relayout.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<html class="reftest-wait">
+<title>Changing object-view-box on an auto-sized element triggers relayout</title>
+<link rel="match" href="object-view-box-auto-size-relayout-ref.html">
+<link rel="help" href="https://drafts.csswg.org/css-images-4/#the-object-view-box">
+
+<style>
+  canvas {
+    display: block;
+    width: 100px;
+    height: auto;
+  }
+</style>
+<body>
+<canvas id="canvas" width="100" height="100"></canvas>
+<script>
+  var ctx = document.getElementById("canvas").getContext("2d");
+  ctx.fillStyle = "red";
+  ctx.fillRect(0, 0, 100, 100);
+
+  function afterFrame(callback)
+  {
+    requestAnimationFrame(() => requestAnimationFrame(callback));
+  }
+
+  // Let the canvas lay out and paint at its natural 100x100 (1:1) size first,
+  // then crop away the bottom half. Since height is auto and there's no
+  // explicit aspect-ratio, the used height is derived from the cropped
+  // (2:1) intrinsic ratio, so this must trigger a relayout, not just a repaint.
+  window.onload = () => {
+    afterFrame(() => {
+      canvas.style.objectViewBox = "inset(0px 0px 50px 0px)";
+      afterFrame(() => {
+        document.documentElement.classList.remove("reftest-wait");
+      });
+    });
+  };
+</script>
+</body>
+</html>
--- a/LayoutTests/imported/w3c/web-platform-tests/css/css-view-transitions/object-view-box-new-image-expected.html
+++ b/LayoutTests/imported/w3c/web-platform-tests/css/css-view-transitions/object-view-box-new-image-expected.html
@@ -7,7 +7,7 @@
   contain: paint;
   width: 100px;
   height: 100px;
-  transform: scale(2.0, 3.0);
+  transform: scale3d(2.0, 3.0, 1.0);
   position: relative;
   top: 200px;
   left: 200px;
--- a/Source/WebCore/platform/graphics/GeometryUtilities.h
+++ b/Source/WebCore/platform/graphics/GeometryUtilities.h
@@ -57,6 +57,10 @@ WEBCORE_EXPORT FloatRect NODELETE mapRect(const FloatRect&, const FloatRect& src
 WEBCORE_EXPORT FloatRect NODELETE largestRectWithAspectRatioInsideRect(float aspectRatio, const FloatRect&);
 WEBCORE_EXPORT FloatRect NODELETE smallestRectWithAspectRatioAroundRect(float aspectRatio, const FloatRect&);
 
+// Given a natural size and a subrect of it, compute the rect that natural-sized content
+// would occupy such that subrect maps exactly onto destRect. Used by object-view-box rendering.
+FloatRect NODELETE fullRectFromSubrectAndSize(const FloatSize& naturalSize, const FloatRect& subrect, const FloatRect& destRect);
+
 FloatSize NODELETE sizeWithAreaAndAspectRatio(float area, float aspectRatio);
 
 // Compute a rect that encloses all points covered by the given rect if it were rotated a full turn around (0,0).
--- a/Source/WebCore/rendering/RenderHTMLCanvas.cpp
+++ b/Source/WebCore/rendering/RenderHTMLCanvas.cpp
@@ -90,11 +90,16 @@ void RenderHTMLCanvas::paintReplaced(PaintInfo& paintInfo, const LayoutPoint& pa
     LayoutRect replacedContentRect = this->replacedContentRect();
     replacedContentRect.moveBy(paintOffset);
 
-    // Not allowed to overflow the content box.
-    bool clip = !contentBoxRect.contains(replacedContentRect);
-    GraphicsContextStateSaver stateSaver(paintInfo.context(), clip);
+    LayoutRect paintRect = computePaintRectForObjectViewBox(replacedContentRect);
+
+    // Not allowed to overflow the content box. Snap before testing containment so that
+    // pixel-snapping can't push paintRect outside contentBoxRect without a clip being applied
+    // (or vice versa).
+    IntRect snappedContentBoxRect = snappedIntRect(contentBoxRect);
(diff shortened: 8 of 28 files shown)


Source/WTF/Scripts/Preferences/UnifiedWebPreferences.yaml
```diff
@@ -1251,17 +1251,17 @@ CSSMathDepthEnabled:
       
 CSSObjectViewBoxEnabled:
   type: bool
-  status: testable
+  status: stable
```
smfr: I don't think we should do this until we're confident that object-view-box works with video, including transitions to video fullscreen which are special.

LayoutTests/TestExpectations
```diff
@@ -7174,7 +7174,7 @@ imported/w3c/web-platform-tests/css/css-view-transitions/scoped/ [ Skip ]
 # waitUntil() method.
 imported/w3c/web-platform-tests/css/css-view-transitions/view-transition-waituntil-animation-manipulation.html [ ImageOnlyFailure ]
 
-# Depends on object-view-box (unimplemented).
+# object-view-box view transitions integration not yet implemented.
```
nt1m: You'll probably want similar logic in RenderViewTransitionCapture::paintReplaced to progress these tests

LayoutTests/TestExpectations
```diff
@@ -4511,7 +4512,6 @@ imported/w3c/web-platform-tests/css/css-images/object-view-box-writing-mode-svg.
 imported/w3c/web-platform-tests/css/css-images/object-view-box-writing-mode-video.html [ ImageOnlyFailure ]
 imported/w3c/web-platform-tests/css/css-images/object-view-box-xywh-percentage.html [ ImageOnlyFailure ]
 imported/w3c/web-platform-tests/css/css-images/object-view-box-xywh.html [ ImageOnlyFailure ]
```
nt1m: If all these tests pass, they should be dropped from the list

Source/WebCore/rendering/RenderReplaced.cpp
```diff
@@ -580,25 +593,53 @@ void RenderReplaced::computeIntrinsicSizesConstrainedByTransferredMinMaxSizes(Fl
...
+    auto left   = Style::evaluate<float>(viewBox->parameters.insets.left(),   physicalIntrinsicSize.width(),  zoom);
+    auto top    = Style::evaluate<float>(viewBox->parameters.insets.top(),    physicalIntrinsicSize.height(), zoom);
+    auto right  = Style::evaluate<float>(viewBox->parameters.insets.right(),  physicalIntrinsicSize.width(),  zoom);
+    auto bottom = Style::evaluate<float>(viewBox->parameters.insets.bottom(), physicalIntrinsicSize.height(), zoom);
+
+    auto width  = physicalIntrinsicSize.width()  - left - right;
+    auto height = physicalIntrinsicSize.height() - top  - bottom;
+
+    if (width <= 0 || height <= 0)
+        return std::nullopt;
```
nt1m, line 626: Is it correct to treat 0 with/height this way? We should have testcases covering this if we don't and see if we match Chrome here.

Source/WebCore/rendering/RenderViewTransitionCapture.cpp
nt1m, line 1: What's the simplest approach we can use to get the view transition tests passing with blurriness? It looks like this has gone through multiple rounds of iterations without solving the blurriness problem, so if we can go back to a simpler version of this, I think it would be better.

Source/WebCore/rendering/RenderReplaced.cpp
tangalice: This behavior matches Chrome, they use IsEmpty() on the compute view box rect (covers 0 and negative dimensions). There are already two WPT tests covering this (object-view-box-empty-bounds, object-view-box-negative-bounds), both are passing right now. Should I still add a local webkit test to explicitly cover the 0 case?

LayoutTests/imported/w3c/web-platform-tests/css/css-view-transitions/object-view-box-new-image-expected.html
```diff
@@ -7,7 +7,7 @@
   contain: paint;
   width: 100px;
   height: 100px;
-  transform: scale(2.0, 3.0);
+  transform: scale3d(2.0, 3.0, 1.0);
```
nt1m, line 10: Did you mean to change this test?

nt1m (commented):
Would be great if @smfr could take a look.

LayoutTests/imported/w3c/web-platform-tests/css/css-view-transitions/object-view-box-new-image-expected.html
tangalice: yes, without scale3d the pixel diff is maxDifference=223/totalPixels=5527 instead of 85/600. I changed it so it disables CA's tiling resolution boost so the reference matches the blur quality better. I noticed some of the other tests in the directory also do this, is there a different approach I should do instead?

nt1m: Hmm, seems ok then

Source/WebCore/rendering/RenderHTMLCanvas.cpp
```diff
@@ -90,21 +90,40 @@ void RenderHTMLCanvas::paintReplaced(PaintInfo& paintInfo, const LayoutPoint& pa
     LayoutRect replacedContentRect = this->replacedContentRect();
     replacedContentRect.moveBy(paintOffset);
 
+    if (auto viewBox = resolveObjectViewBox(FloatSize(intrinsicSize()))) {
```
smfr: This should not replicate code shared with the path below. It would be better to refactor to create a lambda to compute `fullCanvasRect`, and to conditionally clip.

Source/WebCore/rendering/RenderImage.cpp
```diff
@@ -686,6 +686,46 @@ void RenderImage::paintReplaced(PaintInfo& paintInfo, const LayoutPoint& paintOf
         replacedContentRect.moveBy(paintOffset);
     }
 
+    // When object-view-box is set, replacedContentRect is already sized to the view box
+    // dimensions via object-fit/object-position. We need to draw the full image scaled so
+    // the view box sub-region maps exactly onto replacedContentRect, clipped appropriately.
+    if (auto viewBox = resolveObjectViewBox(FloatSize(intrinsicSize())); viewBox && !isDimensionlessSVG()) {
```
smfr: Same comment about not duplicating code.

Source/WebCore/rendering/RenderReplaced.cpp
```diff
@@ -580,25 +593,57 @@ void RenderReplaced::computeIntrinsicSizesConstrainedByTransferredMinMaxSizes(Fl
     }
 }
 
+std::optional<FloatRect> RenderReplaced::resolveObjectViewBox(const FloatSize& physicalIntrinsicSize) const
```
smfr: I would call this `resolvedObjectViewBox` since it returns the value.

Source/WebCore/rendering/RenderReplaced.cpp
```diff
@@ -580,25 +593,57 @@ void RenderReplaced::computeIntrinsicSizesConstrainedByTransferredMinMaxSizes(Fl
...
+{
+    auto viewBox = style().objectViewBox().tryRect();
+    if (!viewBox)
+        return std::nullopt;
+
+    auto zoom = style().usedZoomForLength();
+    auto left   = Style::evaluate<float>(viewBox->parameters.insets.left(),   physicalIntrinsicSize.width(),  zoom);
```
smfr: We don't generally align things like this.

Consider using `RectEdges<>` here, and call them "insets", like `leftInset`, not just `left`.

Source/WebCore/rendering/RenderVideo.cpp
```diff
@@ -256,17 +257,61 @@ void RenderVideo::paintReplaced(PaintInfo& paintInfo, const LayoutPoint& paintOf
 
     auto rect = videoBoxRect;
     rect.moveBy(paintOffset);
-    GraphicsContext& context = paintInfo.context();
+
+    LayoutRect contentRect = contentBoxRect();
+    contentRect.moveBy(paintOffset);
+
+    if (auto viewBox = resolveObjectViewBox(FloatSize(intrinsicSize()))) {
```
smfr: Same comment about not duplicating code.

Source/WebCore/rendering/RenderViewTransitionCapture.cpp
```diff
@@ -86,6 +100,30 @@ void RenderViewTransitionCapture::paintReplaced(PaintInfo& paintInfo, const Layo
 
     FloatRect paintRect = m_localOverflowRect;
 
+    if (auto viewBox = resolveObjectViewBox(FloatSize(intrinsicSize()))) {
```
smfr: Here too. The way to think about is that the "important" function call here (the `context.drawImageBuffer()`) should only occur once.

Source/WebCore/rendering/RenderViewTransitionCapture.cpp
```diff
@@ -102,7 +140,27 @@ void RenderViewTransitionCapture::layout()
     m_localOverflowRect.scale(m_scale.width(), m_scale.height());
     m_localOverflowRect.moveBy(replacedContentRect().location());
 
-    addVisualOverflow(m_localOverflowRect);
+    if (auto viewBox = resolveObjectViewBox(FloatSize(intrinsicSize()))) {
```
smfr: Here too. There should be just one call to `addVisualOverflow`.

smfr (changes requested):

Source/WebCore/rendering/RenderImage.cpp
```diff
@@ -686,18 +686,35 @@ void RenderImage::paintReplaced(PaintInfo& paintInfo, const LayoutPoint& paintOf
...
+                replacedContentRect.x() - viewBox->x() * scaleX,
+                replacedContentRect.y() - viewBox->y() * scaleY,
+                intrinsicSize().width() * scaleX,
+                intrinsicSize().height() * scaleY
+            };
+        }
+        return std::nullopt;
+    }();
+
+    bool clip = fullImageRect || !contentBoxRect.contains(replacedContentRect);
```
smfr: I think it would be better to assign to `paintRect`, and then keep the contains check:

```
bool clip = !contentBoxRect.contains(paintRect);
```

Source/WebCore/rendering/RenderVideo.cpp
```diff
@@ -256,29 +257,44 @@ void RenderVideo::paintReplaced(PaintInfo& paintInfo, const LayoutPoint& paintOf
...
+                rect.x() - viewBox->x() * scaleX,
+                rect.y() - viewBox->y() * scaleY,
+                intrinsicSize().width() * scaleX,
+                intrinsicSize().height() * scaleY
+            };
+        }
+        return std::nullopt;
+    }();
+
+    bool clip = fullVideoRect || !contentRect.contains(rect);
```
smfr: Here also, just initialize a `paintRect` with either `contentBoxRect()` or the result of the lambda, then 
```
bool clip = !contentRect.contains(paintRect);
```

It would be good for the variable naming here to match RenderReplaced. Maybe they should actually share code.

Source/WebCore/rendering/RenderVideo.cpp
```diff
@@ -256,29 +257,44 @@ void RenderVideo::paintReplaced(PaintInfo& paintInfo, const LayoutPoint& paintOf
...
     }
 
-    bool clip = !contentRect.contains(rect);
-    GraphicsContextStateSaver stateSaver(context, clip);
-    if (clip)
-        context.clip(contentRect);
-
     if (displayingPoster) {
-        paintIntoRect(paintInfo, rect);
+        paintIntoRect(paintInfo, fullVideoRect ? FloatRect(*fullVideoRect) : rect);
```
smfr: Can we just use the `paintRect` here without the ternary?

Source/WebCore/rendering/RenderVideo.cpp
```diff
@@ -299,7 +315,7 @@ void RenderVideo::paintReplaced(PaintInfo& paintInfo, const LayoutPoint& paintOf
         && !paintInfo.paintBehavior.contains(PaintBehavior::Snapshotting))
         return;
 
-    videoElement->paint(context, rect);
+    videoElement->paint(context, fullVideoRect ? FloatRect(*fullVideoRect) : FloatRect(rect));
```
smfr: Ditto

Source/WebCore/rendering/RenderViewTransitionCapture.cpp
```diff
@@ -81,14 +95,40 @@ void RenderViewTransitionCapture::paintReplaced(PaintInfo& paintInfo, const Layo
...
+        if (hasNonVisibleOverflow()) {
+            // Clipped (e.g. contain: paint): map only the view-box region onto replacedContentRect.
+            dstRect = FloatRect(replacedContentRect);
+            srcRect = FloatRect(*viewBox);
+        } else {
+            // Overflow-visible: scale the full image so the view-box region maps onto replacedContentRect.
+            float scaleX = replacedContentRect.width() / viewBox->width();
+            float scaleY = replacedContentRect.height() / viewBox->height();
+            dstRect = snapRectToDevicePixels(LayoutRect {
+                replacedContentRect.x() - LayoutUnit(viewBox->x() * scaleX),
```
smfr: I see this code in 4 places in this PR; please move it to `GeometryUtilities.h/cpp`; note that we have things like `largestRectWithAspectRatioInsideRect` there already.

smfr (changes requested):

Source/WebCore/platform/graphics/GeometryUtilities.h
```diff
@@ -55,6 +55,10 @@ WEBCORE_EXPORT FloatRect NODELETE mapRect(const FloatRect&, const FloatRect& src
 WEBCORE_EXPORT FloatRect NODELETE largestRectWithAspectRatioInsideRect(float aspectRatio, const FloatRect&);
 WEBCORE_EXPORT FloatRect NODELETE smallestRectWithAspectRatioAroundRect(float aspectRatio, const FloatRect&);
 
+// Given a natural content size and a view-box sub-rect, compute the full content rect scaled so
+// that viewBox maps exactly onto destRect. Used by object-view-box rendering.
+FloatRect NODELETE naturalRectForObjectViewBox(const FloatSize& naturalSize, const FloatRect& viewBox, const FloatRect& destRect);
```
smfr: Here in this generic file, we should use a name that doesn't have "object view box" in it; something like
```
FloatRect NODELETE fullRectFromSubrectAndSize(const FloatSize& naturalSize, const FloatRect& subrect, const FloatRect& destRect);
```

Source/WebCore/rendering/RenderHTMLCanvas.cpp
```diff
@@ -90,11 +90,13 @@ void RenderHTMLCanvas::paintReplaced(PaintInfo& paintInfo, const LayoutPoint& pa
     LayoutRect replacedContentRect = this->replacedContentRect();
     replacedContentRect.moveBy(paintOffset);
 
+    LayoutRect paintRect = computePaintRectForObjectViewBox(replacedContentRect);
+
     // Not allowed to overflow the content box.
-    bool clip = !contentBoxRect.contains(replacedContentRect);
-    GraphicsContextStateSaver stateSaver(paintInfo.context(), clip);
+    bool clip = !contentBoxRect.contains(paintRect);
```
smfr: We should really do the clipping test after snapping (but that's wrong before your change).

Source/WebCore/rendering/RenderImage.cpp
```diff
@@ -843,6 +847,10 @@ bool RenderImage::foregroundIsKnownToBeOpaqueInRect(const LayoutRect& localRect,
     if (auto objectFit = style().objectFit(); objectFit != ObjectFit::Fill && objectFit != ObjectFit::Cover)
         return false;
 
+    // object-view-box crops the image so it may not cover the full content box.
```
smfr: This could be refined in a followup to compute if the image fully covers the content box.

Source/WebCore/rendering/RenderReplaced.cpp
```diff
@@ -568,37 +582,79 @@ void RenderReplaced::computeIntrinsicSizesConstrainedByTransferredMinMaxSizes(Fl
...
     }
 }
 
+std::optional<FloatRect> RenderReplaced::resolvedObjectViewBox(const FloatSize& physicalIntrinsicSize) const
+{
+    auto viewBox = style().objectViewBox().tryRect();
+    if (!viewBox)
+        return std::nullopt;
+
+    auto zoom = style().usedZoomForLength();
```
smfr, line 614: Are there any tests which exercise zoom != 1? If no, consider adding one.

Source/WebCore/style/StyleDifference.cpp
```diff
@@ -175,7 +175,8 @@ class DifferenceFunctions final {
         ASSERT(&a != &b);
 
         if (a.usedAppearance != b.usedAppearance
-            || a.textOverflow != b.textOverflow)
+            || a.textOverflow != b.textOverflow
+            || a.objectViewBox != b.objectViewBox)
```
smfr: Do we have any tests that ensure that repainting happens when `object-view-box` changes?

Does an `object-view-box` change require layout? Generally I would expect it to only require repaint, unless it can require layout overflow to be recomputed, so this may not be the right place to add this check.

tangalice: object-view-box-property-changed.html covers the fixed size case updating visually on a view box change. I'll add a test to cover the auto size relayout case.

LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-auto-size-relayout.html
nt1m, line 1: All of the tests should be converted to WPT and put in imported/w3c/web-platform-tests/css/css-images/ . This way, other browser vendors can benefit from these tests as well.We can also get tests results from Chrome when we export.

tangalice: I can convert and move object-view-box-auto-size-relayout.html. 
Simon commented earlier asking for a test with Zoom != 1 so object-view-box-zoom.html uses non-standard zoom, should I rewrite it without zoom and move it, or leave it local? 
The two repaint tests (object-view-box-subset-obscures-background-no-repaint.html, -superset-does-not-obscure-background-repaint.html) are checking internal webkit behavior on opaque-foreground repaint which isn't observable web behavior. I also modeled them on an existing webkit local test fast/repaint/obscured-background-no-repaint.html. I think these should stay local

nt1m: You're right that the repaint ones can't be moved to WPT, but the zoom one exercises CSS Zoom which is standardized.

LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-auto-size-relayout-ref.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html>
+<title>Changing object-view-box on an auto-sized element triggers relayout : ref</title>
+<link rel="help" href="https://drafts.csswg.org/css-images-4/#the-object-view-box">
```
smfr: Minor: "ref" files don't need a title or link.

smfr (approved):

LayoutTests/imported/w3c/web-platform-tests/css/css-images/object-view-box-auto-size-relayout-ref.html
nt1m: expected.html also

LayoutTests/fast/css/object-view-box-subset-obscures-background-no-repaint.html
```diff
@@ -0,0 +1,53 @@
+<!DOCTYPE html> <!-- webkit-test-runner [ CSSObjectViewBoxEnabled=true ] -->
+<html>
+<head>
+<script>jsTestIsAsync = true;</script>
+<script src="../../resources/js-test-pre.js"></script>
```
nt1m: please use js-test.js and remove js-test-pre/js-test-post references

LayoutTests/fast/css/object-view-box-superset-does-not-obscure-background-repaint.html
```diff
@@ -0,0 +1,53 @@
+<!DOCTYPE html> <!-- webkit-test-runner [ CSSObjectViewBoxEnabled=true ] -->
+<html>
+<head>
+<script>jsTestIsAsync = true;</script>
+<script src="../../resources/js-test-pre.js"></script>
```
nt1m: ditto here

PR #67161, opened 2026-06-15
Refactor Style::ComputedStyleProperties generation

weinig:
#### 17ba481400a9fdbb8e35184f2fcbffc58ce51cad
<pre>
Refactor Style::ComputedStyleProperties generation
<a href="https://bugs.webkit.org/show_bug.cgi?id=317100">https://bugs.webkit.org/show_bug.cgi?id=317100</a>

Reviewed by Antti Koivisto.

Refactors the generation of Style::ComputedStyleProperties to make it easier to move
around the various different functions each property or logical property group might
need. Specifically, each function prototype now has its own class which knows all the
defaults about it including:
  - its signature
  - its dependencies (what files need to be included for the definition to compile)
  - if it should be inline or not (so that we know if its definition should be
    inline in a header, or if it can be in the cpp file)

This will let us move these functions around more easily going forward, allowing
experimentation with out-of-lining more of them easier.

The change includes support for fine-grained includes for each function type, but
in this refactor, they are not used. This means we need now have a hardcoded
mapping of the computed style container path components to filenames. This is
necessary because CSSProperties.json only includes the variable names like:

  &quot;computed-style-storage-path&quot;: [&quot;m_nonInheritedData&quot;, &quot;miscData&quot;]

so a mapping from something like &quot;miscData&quot; to &quot;StyleNonInheritedMiscData.h&quot; is
needed. In the future, we should consider making the filenames match a strict
pattern or find a way to encode this mapping directly in CSSProperties.json.

Additional mappings were needed for getting the filenames for some values of
the &quot;computed-style-type&quot; key. The majority of types were mappable by removing
the in &quot;::&quot; string and adding a .h (e.g. Style::Color maps to StyleColor.h)
but a few special cases were needed. These should be resolved in the future
by adding files for all the special cases.

The one small change to the test output is due to a more consistent ordering
of logical property group getters.

* Source/WebCore/css/scripts/process-css-properties.py:
* Source/WebCore/css/scripts/test/TestCSSPropertiesResults/StyleComputedStyleProperties.h:

Canonical link: <a href="https://commits.webkit.org/315378@main">https://commits.webkit.org/315378@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/fb4b627697937af613f2a649d433cf2696881b06

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/168793 "5 style errors") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/42352 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/35947 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/178124 "Built successfully") | [❌ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/126500 "Failed to checkout and rebase branch from PR 67161") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/170659 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/42729 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/42312 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/131430 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/59/builds/126500 "Failed to checkout and rebase branch from PR 67161") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/171752 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/33884 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/151704 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/111934 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/32871 "Passed tests") | [  ~~🧪 api-mac-debug~~](https://ews-build.webkit.org/#/builders/165/builds/31585 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/26819 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/142862 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/31104 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/180725 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/33455 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/35753 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/139877 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/42364 "Built successfully") | | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/139721 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/37631 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/43053 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/28076 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/106804 "Built successfully") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/35003 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/114820 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/41556 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/6164 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/40953 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/41207 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/41098 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/css/scripts/process-css-properties.py
+++ b/Source/WebCore/css/scripts/process-css-properties.py
@@ -1268,10 +1268,64 @@ class StyleProperty:
             function_specifiers += ['inline']
         return function_specifiers
 
+    @property
+    def getter_inlines_filename(self):
+        return f"StyleComputedStyleProperties+GettersInlines.h"
+
+    @property
+    def setter_inlines_filename(self):
+        return f"StyleComputedStyleProperties+SettersInlines.h"
+
+    @property
+    def initial_inlines_filename(self):
+        return f"StyleComputedStyleProperties+InitialInlines.h"
+
+    @property
+    def computed_style_type_filename(self):
+        SPECIAL_CASES = {
+            'WebCore::Color': 'Color.h',
+            'Style::TextShadows': 'StyleTextShadow.h',
+            'Style::BoxShadows': 'StyleBoxShadow.h',
+            'Style::AnchorNames': 'StyleAnchorName.h',
+            'Style::ContainerNames': 'StyleContainerName.h',
+            'Style::BorderRadiusValue': 'StyleBorderRadius.h',
+            'Style::InsetEdge': 'StyleInset.h',
+            'Style::MarginEdge': 'StyleMargin.h',
+            'Style::PaddingEdge': 'StylePadding.h',
+            'Style::ScrollPaddingEdge': 'StyleScrollPadding.h',
+            'Style::ScrollMarginEdge': 'StyleScrollMargin.h',
+            'Style::PerspectiveOriginX': 'StylePerspectiveOrigin.h',
+            'Style::PerspectiveOriginY': 'StylePerspectiveOrigin.h',
+            'Style::TransformOriginX': 'StyleTransformOrigin.h',
+            'Style::TransformOriginY': 'StyleTransformOrigin.h',
+            'Style::TransformOriginZ': 'StyleTransformOrigin.h',
+            'Style::PositionTryOrder': 'PositionTryOrder.h',
+            'Style::ViewTransitionClasses': 'StyleViewTransitionClass.h',
+            'StyleAppearance': 'StyleAppearance.h',
+            'ApplePayButtonStyle': 'ApplePayButtonPart.h',
+            'ApplePayButtonType': 'ApplePayButtonPart.h',
+            'AppleVisualEffect': 'AppleVisualEffect.h',
+            'Kerning': 'TextFlags.h',
+            'FontOpticalSizing': 'TextFlags.h',
+            'UnicodeBidi': 'UnicodeBidi.h',
+        }
+
+        computed_style_type = self.codegen_properties.computed_style_type
+
+        if special_cased_result := SPECIAL_CASES.get(computed_style_type):
+            return special_cased_result
+
+        if computed_style_type.startswith('Style::Webkit'):
+            return f"StyleWebKit{computed_style_type.removeprefix('Style::Webkit')}.h"
+        if computed_style_type.startswith('Style::'):
+            return f"Style{computed_style_type.removeprefix('Style::')}.h"
+        return f"RenderStyleConstants.h"
+
 
 class StyleProperties:
     def __init__(self, properties):
         self.properties = properties
+
         self.properties_by_name = {property.name: property for property in properties}
         self.logical_property_groups = {}
         self._all = None
@@ -3233,19 +3287,20 @@ class GenerationContext:
         to.newline()
 
     def generate_includes(self, *, to, headers=[], system_headers=[]):
-        for header in headers:
+        for header in sorted(headers):
             to.write(f"#include \"{header}\"")
-        for header in system_headers:
+        for header in sorted(system_headers):
             to.write(f"#include {header}")
         to.newline()
 
     def generate_cpp_required_includes(self, *, to, header):
-        self.generate_includes(to=to, headers=["config.h", header])
+        to.write(f"#include \"config.h\"")
+        self.generate_includes(to=to, headers=[header])
 
     def generate_forward_declarations(self, *, to, structs=[], classes=[]):
-        for struct in structs:
+        for struct in sorted(structs):
             to.write(f"struct {struct};")
-        for class_ in classes:
+        for class_ in sorted(classes):
             to.write(f"class {class_};")
         to.newline()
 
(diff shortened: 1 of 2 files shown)


anttijk (commented):
Rubber-stamped with this Claude review:

Verdict: Solid, well-structured refactor. The class-per-function-prototype model is a clear improvement over the previous loose collection of _generate_*
  methods, and it's functionally equivalent for the current property set (consistent with EWS: mac/ios/wpe/win all built, tests pass). I found one real latent 
  bug, plus a couple of minor inconsistencies worth fixing before landing.

  🐞 Bug — undefined return_type in _generate_setter_function_definition_shared

  def _generate_setter_function_definition_shared(*, to, function_signature, storage_access, did_set_expression=None):
      ...
      if function_signature.function_return_type == 'void':
          ...
      elif return_type == 'bool':        # ← NameError: 'return_type' is not defined

  When the function was inlined into the new shared helper, the void check was updated to function_signature.function_return_type but the bool branch still
  references the old parameter name return_type, which no longer exists in scope.

  This is currently dormant, not reachable: the only properties with setter_return_type == 'bool' are the four with computed-style-setter-returns-if-changed: 
  true (direction, text-orientation, writing-mode, zoom), and all four also set computed-style-setter-custom: true — so generate_function_definition is never
  called for them. That's why the generator doesn't crash and EWS builds.
  
  But it's a trap: the moment anyone adds a non-custom setter with returns-if-changed, the generator dies with a NameError. Should be
  function_signature.function_return_type == 'bool'. Cheap to fix now.

  🔸 Minor — missing to.newline() in the two SideStartEnd getter classes

  ComputedStyleLogicalPropertyGroupSideStartEndGetterForSpecifiedWritingMode and ...ForCurrentWritingMode both end their generate_function_definition with
  to.write("}") and no trailing to.newline(). Every sibling — BeforeAfter, LogicalLeftRight, and the StartEnd setter — ends with to.newline(), and the old
  combined code emitted a blank line after each definition.
  
  Net effect: in real generated +GettersInlines.h, the start/end getter definitions for side groups (margin, padding, inset, scroll-margin, …) get crammed
  together without the blank-line separation every other getter has. Whitespace-only, compiles fine, and the test fixture doesn't catch it (its only logical
  group is axis), but it's an unintended inconsistency.
  
  🔸 Minor — commit message undersells the output change

  The description says "The one small change to the test output is due to a more consistent ordering of logical property group getters." That's true of the
  test fixture, but the real StyleComputedStyleProperties.h declaration order also changes for side groups (the new all_functions lists are grouped per-edge:
  start(WM), start(), setStart, end(WM)…, whereas the old code grouped all WM-getters, then all current-getters, then all setters). Harmless — declaration
  order only — but worth a sentence so the real diff isn't a surprise. Easy to confirm by regenerating.

  📝 Note — ~half the new code is dead in this PR

  ComputedStylePropertyStorage.storage_path_filenames, computed_style_type_filename (with its SPECIAL_CASES table), and all the includes_needed_for_* /
  includes_needed_for_custom properties are defined but never called by any generate() path — the inline headers still pull in the umbrella custom headers
  exactly as before. This is intentional scaffolding for the future out-of-lining work (the PR says so), but it means none of those filename mappings are
  exercised or tested. Any typo in the SPECIAL_CASES map or the storage-path tree will sit latent until a follow-up wires it up. Fine to land as-is, but
  flagging that this portion is unverified — a reviewer can't easily tell if 'Style::PositionTryOrder': 'PositionTryOrder.h' etc. are correct.

anttijk (approved):

PR #67226, opened 2026-06-16
Add --agent option to git-webkit review

AndresGonzalezApple:
#### 1f81e72bcbe2c876823c3f87376ada0d22949c34
<pre>
Add --agent option to git-webkit review
<a href="https://bugs.webkit.org/show_bug.cgi?id=317154">https://bugs.webkit.org/show_bug.cgi?id=317154</a>

Reviewed by NOBODY (OOPS!).

Add an optional --agent flag to `git-webkit review`. When set, after fetching
the PR diff and existing comments, the command runs a local AI coding agent
headless and read-only against the checked-out (pre-change) source tree, then
folds the agent&apos;s findings into the review file as plain `agent:`-prefixed
lines anchored to the relevant diff lines. The Claude Code CLI is the only
backend wired up today; everything specific to it is confined to a single
helper so the rest of the module is harness-neutral.

The agent runs read-only: it is granted Read, Grep and Glob plus an explicit
allowlist of non-mutating git subcommands (log, show, diff, blame, grep,
cat-file, ls-files, ls-tree, rev-parse), and the permission mode denies
anything not on that list rather than prompting for it, so the agent can
explore the checkout but cannot commit, push, branch or otherwise alter it.

The review instructions are passed as a system prompt and the pull request as
the user prompt, so nothing a PR author writes is interpolated into the
instructions. Findings come back through a JSON schema the backend validates,
rather than by asking the model to format its reply correctly.

The comments behave like the reviewer&apos;s own in-editor comments rather than
rendered PR comments: no &gt;&gt;&gt;&gt; / &lt;&lt;&lt;&lt; markers, and they are kept out of the diff
baseline that invoke_wizard diffs the edited file against, so on save the
reviewer keeps the worthwhile ones and deletes the rest. Kept comments post
like any hand-written inline comment, with the `agent:` / `[severity]` prefix
stripped. The agent never posts to the pull request directly. Comments already
on the pull request are shown to the agent as context so it does not re-raise
points that have been made, but cannot be anchored to.

The checkout is not necessarily the state the diff was taken against, so
--agent warns when it is dirty or not on the pull-request base, and the agent
is told to verify each finding against the file as it reads it.

Degrades gracefully if the agent CLI is missing, errors, times out, or returns
output that does not match the schema, in which case the normal un-annotated
review file is used.

The diff annotation, anchoring and finding-folding work on the materialized
review diff every remote produces, so they are tied to neither the review
program nor the remote hosting the pull request and live in a standalone
webkitscmpy.agent_review module with unit-test coverage; review.py just calls
into it.

* Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py: Added.
(strip_line):
(parse_json):
(_header_path):
(diff_maps):
(annotate_diff):
(_debug_dump):
(_result_message):
(_run_claude_cli):
(_user_prompt):
(run):
* Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py:
(Review.parser):
(Review.invoke_wizard):
(Review.warn_if_not_pre_change_tree):
(Review.main):
* Tools/Scripts/libraries/webkitscmpy/webkitscmpy/test/agent_review_unittest.py: Added.
(MockPullRequest):
(TestStripLine):
(TestParseJSON):
(TestDiffMaps):
(TestAnnotateDiff):
(TestBackend):
(TestUserPrompt):
</pre>
https://github.com/WebKit/WebKit/commit/1f81e72bcbe2c876823c3f87376ada0d22949c34

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/181090 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/55035 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/48833 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/191307 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/145413 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/0ccf2a83-0782-4103-be50-eb1237ccc2a3) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/182952 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/56333 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/54751 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/141307 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/98002 "") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/e9e61bbd-f71b-47f4-9396-d58158f3d20e) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/184045 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/44970 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/162057 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/121758 "Passed tests") | | ⏳ 🛠 vision-apple 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/180414 "Passed tests") | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/43643 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/41924 "Passed tests") | | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/154047 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/41585 "Passed tests") | [✅ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/2464 "Built successfully") | | 
| | | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/46179 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/193239 "Built successfully") | | 
| [✅ 🧪 services](https://ews-build.webkit.org/#/builders/28/builds/180491 "Passed tests") | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/54701 "Built successfully") | | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/150693 "Passed tests") | | 
| | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/55389 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/38205 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/150409 "Passed tests") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/45001 "Passed tests") | | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/123199 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/53906 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/16581 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/53288 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/53525 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/53353 "Built successfully") | | | |

### DIFF
--- a/Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
+++ b/Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
@@ -1,0 +1,474 @@
+# Copyright (C) 2026 Apple Inc. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+# 1.  Redistributions of source code must retain the above copyright
+#     notice, this list of conditions and the following disclaimer.
+# 2.  Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in the
+#     documentation and/or other materials provided with the distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY APPLE INC. AND ITS CONTRIBUTORS "AS IS" AND
+# ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+# DISCLAIMED. IN NO EVENT SHALL APPLE INC. OR ITS CONTRIBUTORS BE LIABLE FOR
+# ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+"""Run a local AI coding agent as a code reviewer and fold its findings into a
+review diff.
+
+The diff annotation, anchoring and finding-folding here work on the materialized
+review diff that pull_request.diff(comments=True) yields, which is the one
+representation every remote converges on, so this module is independent of both
+the review program and the remote hosting the pull request. The only backend
+currently wired up is the Claude Code CLI; everything specific to it is confined
+to _run_claude_cli() so the rest of the module stays backend-neutral."""
+
+import json
+import os
+import re
+import shutil
+import subprocess
+import sys
+import tempfile
+
+from collections import defaultdict
+
+from webkitcorepy import run as run_command, string_utils
+
+# The marker prefixed to every folded finding in the review file. It (and any
+# '[severity]' tag that follows) is stripped from comments the reviewer keeps so
+# they post as clean prose.
+COMMENT_PREFIX = 'agent: '
+
+# The line prefix used to show the agent review comments already on the pull
+# request, so it does not re-raise points that have been made.
+PRIOR_COMMENT_PREFIX = '# prior comment: '
+
+# The system prompt: what the agent should do, independent of which pull request
+# it is looking at. The pull request itself is passed as the user prompt.
+REVIEW_INSTRUCTIONS = f"""\
+You are performing a rigorous, comprehensive code review of a pull request, run \
+locally inside the repository the PR modifies. The working tree is the \
+PRE-CHANGE state of the codebase, so you can read any file, grep, and follow \
+references to understand context the diff alone does not show.
+
+The checkout is not guaranteed to be at the exact commit the diff was taken \
+against, so treat the diff as authoritative for what is changing and verify \
+every finding against the file as you actually read it. Say so rather than \
+reporting a finding if the two disagree.
+
+Cover, with equal seriousness, everything that a thorough correctness review AND \
+a thorough simplification/cleanup review would surface:
+  - Correctness & logic bugs, off-by-one, null handling, error paths, \
+concurrency, resource leaks, security issues.
+  - Reuse, simplification, and efficiency: duplicated logic, dead code, needless \
+complexity, redundant work, simpler equivalent constructs.
+  - Conventions & clarity: adherence to nearby code style, naming, confusing \
+structure, missing or misleading comments.
+  - Test gaps and edge cases the change introduces or fails to handle.
+
+Only report issues you have verified by reading the surrounding code. Do not \
+nitpick things a linter/compiler would catch (imports, formatting, type errors). \
+Skip pre-existing issues on lines the PR did not touch. Prefer fewer, \
+high-signal findings over volume, but be thorough across the categories above. \
+Each finding must be anchored to a specific changed line.
+
+The diff tags every line with its file and absolute line number: 'N<n>' is a \
+line number in the new (post-change) file, 'O<n>' is a line number in the old \
+(pre-change) file. Anchor each comment to one of those tagged lines: use side \
+"new" with the N<n> number for added and context lines, side "old" with the \
+O<n> number for removed lines.
+
+Lines prefixed '{PRIOR_COMMENT_PREFIX.strip()}' are review comments already on \
+the pull request. They are context only: do not repeat a point that has already \
+been made, and do not anchor a finding to them.\
+"""
+
+# The findings contract, enforced by the backend rather than by asking the model
+# to format its reply correctly.
+FINDINGS_SCHEMA = {
+    'type': 'object',
+    'properties': {
+        'comments': {
+            'type': 'array',
+            'items': {
+                'type': 'object',
+                'properties': {
+                    'file': {'type': 'string'},
+                    'line': {'type': 'integer'},
+                    'side': {'type': 'string', 'enum': ['new', 'old']},
+                    'severity': {'type': 'string', 'enum': ['bug', 'cleanup', 'nit', 'question']},
+                    'text': {'type': 'string'},
+                },
+                'required': ['file', 'line', 'side', 'severity', 'text'],
+                'additionalProperties': False,
+            },
+        },
+    },
+    'required': ['comments'],
+    'additionalProperties': False,
+}
+
+
+def strip_line(line):
+    """Strip the 'agent: ' marker (and a leading '[severity] ' tag) from a single
+    line, so that kept agent findings post as clean prose."""
+    if line.startswith(COMMENT_PREFIX):
+        line = line[len(COMMENT_PREFIX):]
+        line = re.sub(r'^\[[^\]]+\]\s*', '', line)
+    return line
+
+
+def parse_json(text):
+    """Parse text as JSON, returning (value, error): value is the parsed value, or
+    None with a short human-readable reason in error."""
+    if not text or not text.strip():
+        return None, 'empty text'
+    try:
+        return json.loads(text), None
+    except Exception as e:
+        return None, str(e)
+
+
+def _header_path(old_line, new_line):
+    """Return the path a '--- <old>' / '+++ <new>' file-header pair refers to, or
+    None if the pair does not look like one. The new side is preferred so that a
+    rename anchors to its new path, but a deleted file's new side is /dev/null,
+    in which case the path comes from the old side."""
+    for line, marker in ((new_line, '+++ b/'), (old_line, '--- a/')):
+        if line.startswith(marker):
+            return line[len(marker):]
+    return None
+
+
+def diff_maps(diff_lines):
+    """Walk the materialized review diff (which already contains any >>>>/<<<<
+    PR-comment blocks) and return:
+      annotated: a line-numbered rendering of the diff for the model, each
+                 content line tagged 'N<dest>' or 'O<src>'.
+      anchor: {(side, file, abs_line): index} -> the diff_lines index whose line
+              a comment should be inserted after.
+      file_header: {file: index} of each file's new-side header line, for
+                   file-level fallbacks when an anchor line is not visible in
+                   the diff."""
+    annotated = []
+    anchor = {}
+    file_header = {}
+    file = None
+    in_comment = False
+    new_no = old_no = None
+    for idx, line in enumerate(diff_lines):
+        if line == '>>>>':
+            in_comment = True
+            continue
+        if line == '<<<<':
+            in_comment = False
+            continue
+        if in_comment:
+            # Existing PR comments are shown to the model as context, but they are
+            # not part of the file, so they neither advance the line counters nor
+            # become anchors.
+            annotated.append(f'{PRIOR_COMMENT_PREFIX}{line}')
+            continue
+        # A file-header pair is a '--- <old>' line immediately followed by a
+        # '+++ <new>' line. Keying header detection on that adjacency avoids
+        # mistaking an added content line whose text happens to start with
+        # '++ b/' (rendered '+++ b/...') for a new file header, and likewise a
+        # removed line starting with '-- '.
+        if line.startswith('+++ ') and idx and diff_lines[idx - 1].startswith('--- '):
+            path = _header_path(diff_lines[idx - 1], line)
+            if path:
+                file = path
+                file_header[file] = idx
+                new_no = old_no = None
+                annotated.append('')
+                annotated.append(f'### File: {file}')
+                continue
+        if line.startswith('diff ') or line.startswith('index '):
+            continue
+        # The old side of a header pair carries no information the new side does
+        # not, so it is dropped; the same adjacency and path checks apply, or a
+        # removed line rendering as '--- ...' would be dropped as a header.
+        if line.startswith('--- ') and idx + 1 < len(diff_lines) and diff_lines[idx + 1].startswith('+++ '):
+            if _header_path(line, diff_lines[idx + 1]):
+                continue
+        if line.startswith('@@'):
+            match = re.match(r'@@ -(\d+)(?:,\d+)? \+(\d+)(?:,\d+)? @@', line)
+            if match:
+                old_no = int(match.group(1))
+                new_no = int(match.group(2))
+            annotated.append(line)
+            continue
+        if file is None or new_no is None:
+            continue
+        leader = line[:1]
+        body = line[1:] if leader in (' ', '+', '-') else line
+        if leader == '+':
+            anchor[('destination', file, new_no)] = idx
+            tag = f'N{new_no}'
+            annotated.append(f'{tag:>8} + {body}')
+            new_no += 1
+        elif leader == '-':
+            anchor[('source', file, old_no)] = idx
+            tag = f'O{old_no}'
+            annotated.append(f'{tag:>8} - {body}')
+            old_no += 1
+        else:
+            anchor[('destination', file, new_no)] = idx
+            anchor[('source', file, old_no)] = idx
+            tag = f'N{new_no}'
+            annotated.append(f'{tag:>8}   {body}')
+            new_no += 1
+            old_no += 1
+    return '\n'.join(annotated), anchor, file_header
+
+
+def annotate_diff(diff_lines, findings):
+    """Fold a list of finding dicts into a copy of diff_lines as plain 'agent: '
+    lines anchored after the relevant diff line. Returns (rebuilt, placed,
+    skipped): rebuilt is the annotated diff (None if nothing could be placed),
+    placed is the number of findings folded in, and skipped counts findings that
+    fell back to a file-level anchor because their line was not in the diff."""
+    _, anchor, file_header = diff_maps(diff_lines)
+
+    insertions = defaultdict(list)
+    placed = 0
+    skipped = 0
+    for c in findings:
+        if not isinstance(c, dict):
+            continue
+        path = c.get('file')
+        line = c.get('line')
+        side = 'source' if c.get('side') == 'old' else 'destination'
+        text = (c.get('text') or '').strip()
+        if not path or not text:
+            continue
+        try:
+            line = int(line)
+        except (TypeError, ValueError):
+            line = None
+        # Source and destination line numbers are counted independently, so a
+        # line number that misses on the side the finding names cannot be looked
+        # up on the other side: a hit there would be a coincidence. Fall back to
+        # the file instead, which is at least honest about being approximate.
+        idx = anchor.get((side, path, line)) if line is not None else None
+        if idx is None:
+            idx = file_header.get(path)
+            if idx is None:
+                continue
+            skipped += 1
+        severity = c.get('severity')
+        body = f'[{severity}] {text}' if severity else text
+        body_lines = body.splitlines() or ['']
+        insertions[idx].extend([COMMENT_PREFIX + body_lines[0]] + body_lines[1:])
+        placed += 1
+
+    if not insertions:
+        return None, 0, skipped
+
+    # Rebuild the diff, dropping each insertion AFTER the target line and after
+    # any existing >>>>/<<<< comment block already attached to it.
+    rebuilt = []
+    i = 0
+    n = len(diff_lines)
+    while i < n:
+        rebuilt.append(diff_lines[i])
+        j = i + 1
+        if j < n and diff_lines[j] == '>>>>':
+            while j < n and diff_lines[j] != '<<<<':
+                rebuilt.append(diff_lines[j])
+                j += 1
+            if j < n:  # the closing '<<<<'
+                rebuilt.append(diff_lines[j])
+                j += 1
+        rebuilt.extend(insertions.get(i, []))
+        i = j
+
+    return rebuilt, placed, skipped
+
+
+def _debug_dump(content):
+    """Write content to a per-process temp file for post-mortem inspection and
+    return the path (returned even if the write failed, so callers can still
+    name it in a warning)."""
+    path = os.path.join(tempfile.gettempdir(), f'git-webkit-agent-review-{os.getpid()}.txt')
+    try:
+        with open(path, 'w') as f:
+            f.write(content)
+    except Exception:
+        pass
+    return path
+
+
+# Tools the agent is allowed to use: Read/Grep/Glob plus a set of non-mutating
+# git subcommands, expressed in the Claude Code CLI --allowedTools syntax.
+# Anything that can mutate the repository (commit, push, update-ref, branch,
+# reset, ...) is deliberately excluded so the review cannot alter the checkout it
+# is inspecting, and combined with '--permission-mode dontAsk' anything not
+# listed here is denied rather than prompted for.
+_CLAUDE_ALLOWED_TOOLS = [
+    'Read', 'Grep', 'Glob',
+    'Bash(git log:*)',
+    'Bash(git show:*)',
+    'Bash(git diff:*)',
+    'Bash(git blame:*)',
+    'Bash(git grep:*)',
+    'Bash(git cat-file:*)',
+    'Bash(git ls-files:*)',
+    'Bash(git ls-tree:*)',
+    'Bash(git rev-parse:*)',
+]
+
+# How many lines of the CLI's stderr to relay when it exits non-zero.
+_STDERR_LINES = 20
+
+
+def _result_message(envelope):
+    """Pick the terminal 'result' message out of the CLI's --output-format json
+    output. That is a single result object, unless the reviewer has verbose
+    output enabled, in which case the CLI emits the whole message array
+    instead."""
+    if isinstance(envelope, dict):
+        return envelope
+    if isinstance(envelope, list):
+        for message in reversed(envelope):
+            if isinstance(message, dict) and message.get('type') == 'result':
+                return message
+    return None
+
+
+def _run_claude_cli(system_prompt, user_prompt, repo_root):
+    """Run the Claude Code CLI headless and read-only against repo_root, returning
+    the review findings it produced (as a JSON value matching FINDINGS_SCHEMA) or
+    None on any failure. Everything specific to the Claude Code CLI, its
+    executable, argv and output envelope, is confined here."""
+    claude_bin = shutil.which('claude')
+    if not claude_bin:
+        sys.stderr.write("Warning: 'claude' CLI not found on PATH; skipping --agent review.\n")
+        return None
+
+    # --allowedTools takes a variable number of values, so it goes last.
+    cmd = [claude_bin, '-p',
+           '--output-format', 'json',
+           '--json-schema', json.dumps(FINDINGS_SCHEMA),
+           '--append-system-prompt', system_prompt,
+           '--permission-mode', 'dontAsk',
+           '--allowedTools'] + _CLAUDE_ALLOWED_TOOLS
+
+    print('Running agent review (read-only, this can take a few minutes)...')
+    try:
+        proc = run_command(cmd, input=user_prompt, capture_output=True,
+                           encoding='utf-8', cwd=repo_root, timeout=900)
+    except subprocess.TimeoutExpired:
+        sys.stderr.write('Warning: Agent review timed out; continuing without annotations.\n')
+        return None
+    except Exception as e:
+        sys.stderr.write(f'Warning: Agent review failed to run ({e}); continuing.\n')
+        return None
+
+    if proc.returncode != 0:
+        sys.stderr.write(f'Warning: Agent review exited with status {proc.returncode}.\n')
+        err = proc.stderr.strip().splitlines()
+        if len(err) > _STDERR_LINES:
+            sys.stderr.write(f'  (last {_STDERR_LINES} of {len(err)} stderr lines)\n')
+            err = err[-_STDERR_LINES:]
+        for line in err:
+            sys.stderr.write(f'  {line}\n')
+        return None
+
+    raw = proc.stdout or ''
+    envelope, error = parse_json(raw)
+    result = _result_message(envelope)
+    if result is None:
+        debug_path = _debug_dump(raw)
+        reason = error or 'no terminal result message in the output'
+        sys.stderr.write(f'Warning: Could not parse the agent CLI output envelope ({reason}).\n')
+        sys.stderr.write(f'  Captured {len(raw)} bytes of stdout; saved to {debug_path}\n')
+        if raw.strip():
+            preview = raw.strip()[:300].replace('\n', ' ')
+            sys.stderr.write(f'  Output begins: {preview}\n')
+        return None
+
+    subtype = result.get('subtype')
+    if result.get('is_error') or (subtype and subtype != 'success'):
+        is_error = result.get('is_error')
+        sys.stderr.write(f'Warning: Agent review reported a problem (subtype={subtype}, is_error={is_error}).\n')
+
+    # The schema is validated by the CLI, which hands back the parsed findings.
+    findings = result.get('structured_output')
+    if findings is not None:
+        return findings
+
+    # No structured output: the CLI rejected the schema and fell back to plain
+    # text, so parse the reply as JSON ourselves.
+    text = result.get('result')
+    if isinstance(text, str) and text.strip():
+        findings, error = parse_json(text)
+        if findings is not None:
+            return findings
+        debug_path = _debug_dump(text)
+        sys.stderr.write(f"Warning: Could not parse the agent's review findings ({error}).\n")
+        sys.stderr.write(f"  Agent's reply ({len(text)} chars) saved to {debug_path}\n")
+        return None
+
+    debug_path = _debug_dump(json.dumps(result, indent=2, sort_keys=True))
+    sys.stderr.write(f'Warning: Agent returned no findings to parse (subtype={subtype}).\n')
+    sys.stderr.write(f'  Full envelope saved to {debug_path}\n')
+    return None
+
+
+def _user_prompt(pull_request, annotated_diff):
+    """The pull request under review, as the user half of the prompt. The content
+    is interpolated once and never re-scanned, so nothing a PR author writes can
+    be mistaken for part of the review instructions."""
+    return '\n'.join([
+        '=== PULL REQUEST ===',
+        f'Title: {pull_request.title or ""}',
+        '',
+        'Description:',
+        pull_request.body or '(none)',
+        '',
+        '=== ANNOTATED DIFF ===',
+        annotated_diff,
+    ])
+
+
+def run(pull_request, diff_lines, repo_root=None):
+    """Run a headless, read-only AI agent code review of the PR and return a copy
+    of diff_lines with the agent's findings folded in as plain 'agent:' lines
+    anchored to the relevant diff lines. Returns None on any failure, so the
+    caller falls back to the normal un-annotated review file. repo_root is the
+    directory the agent runs in, defaulting to the current working directory."""
+    annotated, _, _ = diff_maps(diff_lines)
+
+    findings = _run_claude_cli(
+        REVIEW_INSTRUCTIONS, _user_prompt(pull_request, annotated),
+        repo_root or os.getcwd(),
+    )
+    if findings is None:
+        return None
+
+    found = findings.get('comments') if isinstance(findings, dict) else findings
+    if not isinstance(found, list):
+        debug_path = _debug_dump(json.dumps(findings, indent=2, sort_keys=True))
+        sys.stderr.write(
+            f"Warning: The agent's findings had no 'comments' list "
+            f'(top-level type was {type(findings).__name__}).\n')
+        sys.stderr.write(f'  Findings saved to {debug_path}\n')
+        return None
+
+    rebuilt, placed, skipped = annotate_diff(diff_lines, found)
+    if rebuilt is None:
+        print('Agent review: no comments to add.')
+        return None
+
+    counted = string_utils.pluralize(placed, 'comment')
+    print(f'Agent review: {counted} added ({skipped} not anchored to a visible diff line).')
+    return rebuilt
--- a/Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py
+++ b/Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py
@@ -32,7 +32,7 @@ from .command import Command
 from .pull_request import PullRequest
 
 from webkitcorepy import arguments, run, string_utils, Terminal
-from webkitscmpy import local, log, remote
+from webkitscmpy import agent_review, local, log, remote
 
 
 class Review(Command):
@@ -64,6 +64,12 @@ class Review(Command):
             dest='dry_run', default=False,
             help='View but do not edit a pull-request',
             action=arguments.NoAction,
+        )
+        parser.add_argument(
+            '--agent',
+            dest='agent', default=False,
+            help='Run a local AI agent code review and fold its comments into the review file',
+            action='store_true',
         )
 
     @classmethod
@@ -127,6 +133,7 @@ class Review(Command):
         messages=None,
         comments=None,
         diff=None,
+        diff_with_agent=None,
     ):
         # Textual wizards are split into 4 sections, seperated by a line of '=' characters
         output = [[], [], [], []]
@@ -167,12 +174,21 @@ class Review(Command):
         for line in diff:
             output[3].append(line.rstrip())
 
+        # The file written for editing may carry the agent's annotations folded
+        # into the diff section, but output[3] stays clean: it is the baseline the
+        # diff of the edited file is taken against, so the agent's lines register
+        # as new reviewer comments (postable on save) rather than pre-existing
+        # ones.
+        file_sections = list(output)
+        if diff_with_agent is not None:
+            file_sections[3] = [line.rstrip() for line in diff_with_agent]
+
         with open(edited_path, 'w') as ofile:
             cnt = 0
-            while cnt < len(output):
-                for line in output[cnt]:
+            while cnt < len(file_sections):
+                for line in file_sections[cnt]:
                     ofile.write(line + '\n')
-                if output[cnt] and cnt + 1 < len(output):
+                if file_sections[cnt] and cnt + 1 < len(file_sections):
                     ofile.write('=' * cls.SEPERATOR_SIZE + '\n')
                 cnt += 1
 
@@ -326,6 +342,41 @@ class Review(Command):
         return response
 
     @classmethod
+    def warn_if_not_pre_change_tree(cls, pull_request, local_repository):
+        """The agent reads the checkout to understand the code the diff changes, so
+        warn when the checkout is not plausibly the state that diff was taken
+        against: findings from an unrelated tree read as confidently as real
+        ones."""
+        if not local_repository:
+            sys.stderr.write('Warning: No local checkout to review against; the agent has no source to read.\n')
+            return
+        modified = local_repository.modified()
+        if modified:
+            counted = string_utils.pluralize(len(modified), 'file')
+            verb = 'has' if len(modified) == 1 else 'have'
+            sys.stderr.write(f'Warning: {counted} in the checkout {verb} uncommitted changes the agent will read as-is.\n')
+        if not pull_request.base:
+            return
+        # The base branch of a pull-request need not exist locally, in which case
+        # there is nothing to compare the checkout against.
+        try:
+            merge_base = local_repository.merge_base(
+                pull_request.base, 'HEAD',
+                include_log=False, include_identifier=False,
+            )
+            head = local_repository.commit(include_log=False, include_identifier=False)
+        except (ValueError, local_repository.Exception):
+            merge_base = head = None
+        if not merge_base or not head:
+            sys.stderr.write(
+                f"Warning: Cannot resolve '{pull_request.base}' locally to check the checkout against the diff.\n")
+            return
+        if merge_base.hash != head.hash:
+            sys.stderr.write(
+                f"Warning: This checkout is not on '{pull_request.base}' (the pull-request base), so the agent is\n")
+            sys.stderr.write('    reading source the diff was not taken against and may report accordingly.\n')
+
+    @classmethod
     def main(cls, args, repository, **kwargs):
         editor = cls.editor(repository)
 
@@ -342,6 +393,8 @@ class Review(Command):
         if not repository:
             sys.stderr.write('No repository provided\n')
             return 1
+
+        local_repository = repository if isinstance(repository, local.Git) else None
 
         if isinstance(repository, local.Git):
             original = repository
@@ -383,6 +436,14 @@ class Review(Command):
             if pr_issue.labels:
                 header['Labels'] = string_utils.join(pr_issue.labels)
 
+        diff_lines = list(pull_request.diff(comments=True))
+        diff_with_agent = None
+        if args.agent:
+            cls.warn_if_not_pre_change_tree(pull_request, local_repository)
+            diff_with_agent = agent_review.run(
+                pull_request, diff_lines,
+                repo_root=local_repository.root_path if local_repository else None,
+            )
         print('Waiting for user to modify textual pull-request...')
         response = cls.invoke_wizard(
             editor=editor,
@@ -392,8 +453,22 @@ class Review(Command):
             comments=[
                 '{}: {}'.format(comment.author, re.sub(cls.DETAILS_RE, '', comment.content, flags=re.S))
                 for comment in pull_request.comments
-            ], diff=pull_request.diff(comments=True),
+            ], diff=diff_lines,
+            diff_with_agent=diff_with_agent,
         )
+
+        # The agent's annotations carry an 'agent: ' marker in the review file;
+        # strip it (and any '[severity]' tag) from comments the user kept so they
+        # post as clean prose. truncate_strs has already split every comment body
+        # into a list of lines, so several findings stacked at one anchor each get
+        # stripped.
+        if args.agent:
+            diff_section = response.get('diff') or {}
+            for line_comments in (diff_section.get('diff_comments') or {}).values():
+                for key, body in line_comments.items():
+                    line_comments[key] = [agent_review.strip_line(line) for line in body]
+            if diff_section.get('comment'):
+                diff_section['comment'] = [agent_review.strip_line(line) for line in diff_section['comment']]
 
         if args.dry_run:
             if response:
--- a/Tools/Scripts/libraries/webkitscmpy/webkitscmpy/test/agent_review_unittest.py
+++ b/Tools/Scripts/libraries/webkitscmpy/webkitscmpy/test/agent_review_unittest.py
@@ -1,0 +1,359 @@
+# Copyright (C) 2026 Apple Inc. All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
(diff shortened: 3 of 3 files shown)


Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py
```diff
@@ -325,6 +344,326 @@ def invoke_wizard(
...
+        annotated, anchor, file_header = cls._claude_diff_maps(diff_lines)
+        prompt = (cls.CLAUDE_REVIEW_INSTRUCTIONS
+                  .replace('{title}', pull_request.title or '')
+                  .replace('{description}', pull_request.body or '(none)')
+                  .replace('{diff}', annotated))
+
+        cmd = [claude_bin, '-p',
+               '--output-format', 'json',
+               '--permission-mode', 'default',
+               '--allowedTools', 'Read', 'Grep', 'Glob', 'Task', 'Bash(git:*)']
```
gsnedders: Allowing `git:*` is definitely wrong — it shouldn't be able to `git commit`, `git push`, `git update-ref`, `git branch`, etc.!

gsnedders (changes requested):
IMO, we should try and make as much of this as possible generic, avoid using "claude" in variable names everywhere. There's relatively little of this which is inherently tied to Claude Code, and we may well want to use other harnesses and models in the future.

There's also just enough complexity here that it would probably be good to split out of `program.review`, because a lot of things like parsing diffs and rewriting them aren't really inherently tied to the review program.

And, (probably unsurprisingly!), unit tests for all this complex parsing etc. really are needed.

(Also: I haven't had a super close look at the it, but that's what stands out!)

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py
AndresGonzalezApple: Fixed. Replaced `Bash(git:*)` with an explicit allowlist of read-only subcommands only — `git log`/`show`/`diff`/`blame`/`grep`/`cat-file`/`ls-files`/`ls-tree`/`rev-parse` — so the agent can explore history but can't `commit`, `push`, `branch`, `update-ref`, etc. It's now the `_CLAUDE_ALLOWED_TOOLS` list in the new `agent_review` module.

AndresGonzalezApple:
Thanks for the review — addressed all three points:

- **Generic naming:** the flag is now `--agent` and the identifiers/messages are harness-neutral. The Claude Code CLI is still the only backend, but everything specific to it (executable lookup, argv, the stream-json envelope, and the allowed-tools list) is confined to a single `_run_claude_cli()` helper, so wiring up another harness/model later is localized rather than spread across the module.
- **Split out of `program.review`:** the diff annotation/anchoring, lenient JSON extraction, and finding-folding now live in a standalone `webkitscmpy/agent_review.py`; `review.py` just calls `agent_review.run(...)`.
- **Unit tests:** added `agent_review_unittest.py` (26 tests) covering the diff parsing/anchoring, JSON extraction, finding-folding, and prefix stripping.

Also narrowed the git tool allowlist to read-only subcommands per your inline comment.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,420 @@
...
+    human-readable reason in error."""
+    if not text or not text.strip():
+        return None, 'empty text'
+    text = text.strip()
+
+    candidates = [text]
+    fence = re.search(r'```(?:json)?\s*(.*?)\s*```', text, re.DOTALL)
+    if fence:
+        candidates.append(fence.group(1).strip())
+    for open_ch, close_ch in (('{', '}'), ('[', ']')):
```
AndresGonzalezApple, line 127: The '{'..'}' substring candidate is appended before the '['..']' one, so a top-level JSON array embedded in prose that contains exactly one object (e.g. 'here: [{"file":...}] done') gets mis-extracted as just the inner object. That object then fails the findings.get('comments') check in run() and the whole review is silently dropped. Trying the bracket candidate first (or preferring whichever matches the outermost delimiter) would be more robust; harmless given the model is told to return an object, but worth noting.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/test/agent_review_unittest.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def test_added_line_resembling_file_header_is_content(self):
+        # An added line whose text starts with '++ b/' renders as '+++ b/...';
+        # it must be treated as content, not mistaken for a new file header.
+        diff = [
+            '--- a/ChangeLog',
+            '+++ b/ChangeLog',
+            '@@ -1,0 +1,2 @@',
+            '+regular added line',
+            '++ b/looks-like-a-header',
```
AndresGonzalezApple: This test claims to exercise the adjacency guard for a content line 'rendered +++ b/...', but the diff entry '++ b/looks-like-a-header' has only two leading '+', so it never satisfies startswith('+++ b/') in diff_maps and would be treated as content even without the diff_lines[idx-1].startswith('--- ') guard. To actually validate the guard the entry should be '+++ b/looks-like-a-header' (content whose text starts with '++ b/', preceded by a non-'--- ' line).

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Fixed. `extract_json` now tries the JSON slice whose opening delimiter appears earliest (outermost) first, so a prose-wrapped top-level array like `here: [{...}] done` extracts as the array rather than the inner object. Added a regression test (`test_prose_wrapped_array_of_one_object`).

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/test/agent_review_unittest.py
AndresGonzalezApple: Good catch — the `++ b/` entry had only two `+` and never hit `startswith('+++ b/')`, so the test passed regardless of the guard. Changed it to `+++ b/looks-like-a-header` (an added line whose content is `++ b/...`, preceded by a non-`--- ` line), which now actually exercises the adjacency guard.

AndresGonzalezApple:
@gsnedders when you get a chance, could you take another look? I believe the earlier feedback is addressed:

- **Generic naming** — flag is now `--agent`; the Claude-specific bits (executable, argv, stream-json envelope, allowed-tools) are confined to a single `_run_claude_cli()` helper so the rest reads harness-neutral.
- **Split out of `program.review`** — diff annotation/anchoring, JSON extraction, and finding-folding now live in a standalone `webkitscmpy/agent_review.py`.
- **Unit tests** — added `agent_review_unittest.py` (27 tests) over the parsing/annotation/folding/prefix-stripping.
- **`git:*`** — narrowed to an explicit read-only subcommand allowlist.

EWS is green except `webkitpy`, which is failing on `webkitpy.style.checker_unittest.GlobalVariablesTest.test_webkit_base_filter_rules` — a pre-existing/unrelated flake (the previous commit passed webkitpy, this change only touches `webkitscmpy`, and that test exercises the webkitpy style-checker config, not anything here).

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+# Helpers for running a local AI coding agent as a code reviewer and folding its
+# findings into a review diff. The diff parsing, annotation and finding-folding
+# here are independent of the review program, so they can be reused and unit
+# tested on their own. The only backend currently wired up is the Claude Code
+# CLI; everything specific to it is confined to _run_claude_cli() so the rest of
+# the module stays backend-neutral.
```
gsnedders, line 28: This should just be the module docstring

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+    if isinstance(value, list):
+        return [strip_line(line) for line in value]
+    return '\n'.join(strip_line(line) for line in value.split('\n'))
+
+
+def extract_json(text):
+    """Best-effort parse of a JSON value out of model text, tolerating markdown
+    fences, surrounding prose, a top-level array, and trailing commas. Returns
+    (value, error): value is the parsed object/list, or None with a short
+    human-readable reason in error."""
```
gsnedders, line 114: Most vaguely recent guidance very much suggests _not_ being lax with JSON parsing — and actually requiring proper JSON output. With `--output-format json`, are you actually seeing invalid JSON being output frequently?

Claude's docs all say that it _will_ give valid JSON with `--output-format json`.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+OUTPUT: Respond with raw JSON ONLY, no prose and no markdown fences. Schema:
+{"comments": [
+  {"file": "<path exactly as shown after '### File:'>",
+   "line": <integer line number you are anchoring to>,
+   "side": "new" | "old",
+   "severity": "bug" | "cleanup" | "nit" | "question",
+   "text": "<concise comment, one or a few sentences>"}
+]}
+Use side "new" with the N<n> number for added/context lines, side "old" with the \
+O<n> number for removed lines. If you have no findings, return {"comments": []}.
```
gsnedders, line 82: Most LLM tools nowadays have ways to enforce JSON output, and to provide a JSON schema the output must conform to. I'd expect this goes better than putting instructions in the user prompt?

In Claude Code's case, see https://code.claude.com/docs/en/headless#get-structured-output

You're already using `--output-format json` below, so why not go all the way and provide an actual schema too?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+from collections import defaultdict
+
+from webkitcorepy import run as run_command, string_utils
+
+# The marker prefixed to every folded finding in the review file. It (and any
+# '[severity]' tag that follows) is stripped from comments the reviewer keeps so
+# they post as clean prose.
+COMMENT_PREFIX = 'agent: '
+
+REVIEW_INSTRUCTIONS = """\
```
gsnedders, line 47: Don't we want to split this up into a system prompt (containing the review instructions) and a user prompt (containing the PR content)? The two are clearly different, with different levels of risk, etc.

Indeed, https://code.claude.com/docs/en/headless#customize-the-system-prompt uses automated PR review as an example of when to use the system prompt.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+    Everything specific to the Claude Code CLI, its executable, argv and
+    stream-json envelope, is confined here."""
+    claude_bin = shutil.which('claude')
+    if not claude_bin:
+        sys.stderr.write("Warning: 'claude' CLI not found on PATH; skipping --agent review.\n")
+        return None
+
+    cmd = [claude_bin, '-p',
+           '--output-format', 'json',
+           '--permission-mode', 'default',
```
gsnedders, line 328: Don't we want `dontAsk`, so only the tools listed in `--allowedTools` are allowed?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+
+Only report issues you have verified by reading the surrounding code. Do not \
+nitpick things a linter/compiler would catch (imports, formatting, type errors). \
+Skip pre-existing issues on lines the PR did not touch. Prefer fewer, \
+high-signal findings over volume, but be thorough across the categories above. \
+Each finding must be anchored to a specific changed line.
+
+The diff below tags every line with its file and absolute line number: 'N<n>' is \
+a line number in the new (post-change) file, 'O<n>' is a line number in the old \
+(pre-change) file. Anchor each comment to one of those tagged lines.
```
gsnedders, line 71: How well does this actually work? Unified diffs are something that exist plenty within training data, and I wouldn't be surprised if correct line number attribution is increasingly something in post-training, versus adding metadata into the stream.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+
+def diff_maps(diff_lines):
+    """Walk the materialized review diff (which already contains any >>>>/<<<<
+    PR-comment blocks) and return:
+      annotated: a line-numbered rendering of the diff for the model, each
+                 content line tagged 'N<dest>' or 'O<src>'.
+      anchor: {(side, file, abs_line): index} -> the diff_lines index whose line
+              a comment should be inserted after.
+      file_header: {file: index} of each '+++ b/<file>' line, for file-level
+                   fallbacks when an anchor line is not visible in the diff."""
```
gsnedders, line 158: We have so many diff parsers already, including three within `webkitscmpy`, which makes me feel like we shouldn't need another!

We already have:

 1. `Tools/Scripts/libraries/webkitscmpy/webkitscmpy/remote/git_hub.py`
 2. `Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py`
 3. `Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/diff/html_diff.py`
 4. `Tools/Scripts/webkitpy/common/checkout/diff_parser.py`

Can we really not just use the one that already exists in `Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py`? Are we really doing something so different here?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+            old_no += 1
+    return '\n'.join(annotated), anchor, file_header
+
+
+def annotate_diff(diff_lines, findings):
+    """Fold a list of finding dicts into a copy of diff_lines as plain 'agent: '
+    lines anchored after the relevant diff line. Returns (rebuilt, placed,
+    skipped): rebuilt is the annotated diff (None if nothing could be placed),
+    placed is the number of findings folded in, and skipped counts findings that
+    fell back to a file-level anchor because their line was not in the diff."""
```
gsnedders, line 223: Again, can we not re-use the logic which already exists in `Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py`?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+    cmd = [claude_bin, '-p',
+           '--output-format', 'json',
+           '--permission-mode', 'default',
+           '--allowedTools'] + _CLAUDE_ALLOWED_TOOLS
+
+    print('Running agent review (read-only, this can take a few minutes)...')
+    try:
+        proc = run_command(cmd, input=prompt, capture_output=True,
+                           encoding='utf-8', errors='replace',
+                           cwd=repo_root, timeout=900)
```
gsnedders, line 335: Do we actually want to pass `encoding='utf-8', errors='replace'`, or do we just want bytes as output and let the JSON parser parse whatever output it gets?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+    stream-json envelope, is confined here."""
+    claude_bin = shutil.which('claude')
+    if not claude_bin:
+        sys.stderr.write("Warning: 'claude' CLI not found on PATH; skipping --agent review.\n")
+        return None
+
+    cmd = [claude_bin, '-p',
+           '--output-format', 'json',
+           '--permission-mode', 'default',
+           '--allowedTools'] + _CLAUDE_ALLOWED_TOOLS
```
gsnedders, line 329: This uses the default model specified in `~/.claude.json`, whatever the user has configured there! Is that intentional?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+                           cwd=repo_root, timeout=900)
+    except subprocess.TimeoutExpired:
+        sys.stderr.write('Warning: Agent review timed out; continuing without annotations.\n')
+        return None
+    except Exception as e:
+        sys.stderr.write('Warning: Agent review failed to run ({}); continuing.\n'.format(e))
+        return None
+
+    if proc.returncode != 0:
+        err = (proc.stderr or '').strip()
```
gsnedders, line 344:
```suggestion
        err = proc.stderr.strip()
```

With `capture_output=True`, `stderr` will not be `None`.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+        sys.stderr.write('Warning: Agent review exited with status {}.\n'.format(proc.returncode))
+        if err:
+            sys.stderr.write('  {}\n'.format(err.splitlines()[-1]))
```
gsnedders, line 347: Giving only the last line of output might not be that helpful?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+    print('Running agent review (read-only, this can take a few minutes)...')
+    try:
+        proc = run_command(cmd, input=prompt, capture_output=True,
+                           encoding='utf-8', errors='replace',
+                           cwd=repo_root, timeout=900)
+    except subprocess.TimeoutExpired:
+        sys.stderr.write('Warning: Agent review timed out; continuing without annotations.\n')
+        return None
+    except Exception as e:
+        sys.stderr.write('Warning: Agent review failed to run ({}); continuing.\n'.format(e))
```
gsnedders, line 340: This really applies throughout this file, but we probably want to be using f-strings in general, not `str.format`.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+    if wrapper.get('is_error') or (subtype and subtype != 'success'):
+        sys.stderr.write('Warning: Agent review reported a problem (subtype={}, is_error={}).\n'.format(
+            subtype, wrapper.get('is_error')))
+
+    result_text = wrapper.get('result')
+    if not isinstance(result_text, str) or not result_text.strip():
+        debug_path = _debug_dump(json.dumps(wrapper, indent=2, sort_keys=True))
+        sys.stderr.write("Warning: Agent returned no 'result' text to parse (subtype={}).\n".format(subtype))
+        sys.stderr.write('  Full envelope saved to {}\n'.format(debug_path))
+        return None
```
gsnedders, line 370: This doesn't match the actual format that Claude Code gives; trying to review this PR with itself gives:

```
DEBUG:sitecustomize:Popen executable='/Users/gsnedders/.local/bin/claude' args=['/Users/gsnedders/.local/bin/claude', '-p', '--output-format', 'json', '--permission-mode', 'default', '--allowedTools', 'Read', 'Grep', 'Glob', 'Task', 'Bash(git log:*)', 'Bash(git show:*)', 'Bash(git diff:*)', 'Bash(git blame:*)', 'Bash(git grep:*)', 'Bash(git cat-file:*)', 'Bash(git ls-files:*)', 'Bash(git ls-tree:*)', 'Bash(git rev-parse:*)'] cwd='/Volumes/gsnedders/projects/Safari/OS2'
xWarning: Could not parse the agent CLI output envelope (None).
  Captured 237819 bytes of stdout; saved to /var/folders/s_/xqf0y_kx0g5bflff0vsd3xvc0000gn/T/git-webkit-agent-review-24230.txt
  Output begins: [{"type":"system","subtype":"init","cwd":"/Volumes/gsnedders/projects/Safari/OS2","session_id":"61c6ae81-0c9f-4a00-85cc-9027632f3385","tools":["Task","Bash","CronCreate","CronDelete","CronList","Edit","EnterWorktree","ExitWorktree","Glob","Grep","LSP","NotebookEdit","Read","ReportFindings","Schedule
```

It seems like `--output-format json` gives a sequence of messages, corresponding to `SDKMessage` in [the SDK](https://code.claude.com/docs/en/agent-sdk/typescript#sdkmessage), at least for me with Claude Code 2.1.219. I do recognise that this contradicts [Claude Code's docs](https://code.claude.com/docs/en/headless#get-structured-output), but this is consistently the behaviour I see locally.

This seems to work consistently to parse its output:

```
jq '.[] | select(.type == "result" and .subtype == "success") | .result | ltrimstr("```json\n") | rtrimstr("\n```") | fromjson' /var/folders/s_/xqf0y_kx0g5bflff0vsd3xvc0000gn/T/git-webkit-agent-review-24230.txt
```

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+            in_comment = False
+            continue
+        if in_comment:
+            continue
+        # A file-header pair is a '--- <old>' line immediately followed by a
+        # '+++ <new>' line. Keying header detection on that adjacency avoids
+        # mistaking an added content line whose text happens to start with
+        # '++ b/' (rendered '+++ b/...') for a new file header, and likewise a
+        # removed line starting with '-- '.
+        if line.startswith('+++ b/') and idx and diff_lines[idx - 1].startswith('--- '):
```
gsnedders, line 179: *From a self-review run of this `--agent` feature against its own PR:*

> File-header detection only matches `+++ b/<path>`, but a deleted file's new-side header is `+++ /dev/null`, which never matches. If the deleted file is first in the diff, `file` stays `None` and its content is dropped entirely — no anchor, no file-level fallback, so any finding for it is silently discarded. If it isn't first, its lines get misattributed to whatever the previous file was, under the wrong line numbers. Added files aren't affected, since `--- /dev/null` still satisfies the old-side check.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+        try:
+            line = int(line)
+        except (TypeError, ValueError):
+            line = None
+        idx = None
+        if line is not None:
+            idx = anchor.get((side, path, line))
+            if idx is None:
+                other = 'destination' if side == 'source' else 'source'
+                idx = anchor.get((other, path, line))
```
gsnedders, line 247: *From a self-review run of this `--agent` feature against its own PR:*

> This fallback tries the same line number on the opposite side when the direct anchor lookup misses — but source and destination line counters are independent, so a same-number hit here can be pure coincidence. If the model ever mislabels `side`, this silently anchors the comment to unrelated code with no signal it was a guess rather than an exact match, and it skips the safer file-level fallback just below (`file_header.get(path)`).

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+    new_no = old_no = None
+    for idx, line in enumerate(diff_lines):
+        if line == '>>>>':
+            in_comment = True
+            continue
+        if line == '<<<<':
+            in_comment = False
+            continue
+        if in_comment:
+            continue
```
gsnedders, line 173: *From a self-review run of this `--agent` feature against its own PR:*

> This unconditionally strips content inside existing `>>>>`/`<<<<` review-comment blocks before building the annotated diff sent to the model, so the agent never sees prior human discussion. Intentional (keep the prompt focused on just the diff), or should existing comments be surfaced so the agent doesn't re-flag things already addressed?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+    'Bash(git log:*)',
+    'Bash(git show:*)',
+    'Bash(git diff:*)',
+    'Bash(git blame:*)',
+    'Bash(git grep:*)',
+    'Bash(git cat-file:*)',
+    'Bash(git ls-files:*)',
+    'Bash(git ls-tree:*)',
+    'Bash(git rev-parse:*)',
+]
```
gsnedders, line 313: *From a self-review run of this `--agent` feature against its own PR, plus a caveat after checking the docs:*

> `Read`/`Grep`/`Glob` aren't scoped to `repo_root` here, and the prompt embeds attacker-controlled PR title/body/diff text. That said, Claude Code scopes these tools to cwd plus any `--add-dir` grants by default, so the realistic exposure is probably just this checkout's tree rather than the whole filesystem — worth confirming there's no broader `--add-dir` grant anywhere else in the launch path before treating this as a real exfiltration path. Separately, `Task` is left unrestricted alongside an otherwise mutation-free allowlist; it's unclear whether a `Task`-spawned subagent is bound by the parent's `--allowedTools` list specifically (docs only confirm stricter *permission modes* are inherited), so worth confirming directly rather than assuming either way.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py
```diff
@@ -384,6 +402,13 @@ def main(cls, args, repository, **kwargs):
                 header['Labels'] = string_utils.join(pr_issue.labels)
 
         print('Waiting for user to modify textual pull-request...')
+        diff_lines = list(pull_request.diff(comments=True))
+        diff_with_agent = None
+        if args.agent:
+            diff_with_agent = agent_review.run(
+                pull_request, diff_lines,
+                repo_root=local_repository.root_path if local_repository else None,
```
gsnedders, line 410: *From a self-review run of this `--agent` feature against its own PR:*

> Nothing here verifies the local checkout is actually at the PR's base commit before handing it to the agent. `REVIEW_INSTRUCTIONS` tells the agent it's looking at the pre-change state, but `pull_request` is fetched over the API, not derived from the checkout — running this from an unrelated branch would produce confidently wrong findings with no warning.

…and even running this from the correct branch might not have the actual base commit we get the diff for.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py
```diff
@@ -384,6 +402,13 @@ def main(cls, args, repository, **kwargs):
                 header['Labels'] = string_utils.join(pr_issue.labels)
 
         print('Waiting for user to modify textual pull-request...')
+        diff_lines = list(pull_request.diff(comments=True))
+        diff_with_agent = None
+        if args.agent:
```
gsnedders, line 407: *From a self-review run of this `--agent` feature against its own PR:*

> This prints before the (potentially 900s) `agent_review.run()` call, so with `--agent` the user sees this message well before `invoke_wizard()` below — which is what actually opens the editor — even runs. Move the print to just before the `invoke_wizard()` call.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py
```diff
@@ -392,9 +417,21 @@ def main(cls, args, repository, **kwargs):
...
+        # The agent's annotations carry an 'agent: ' marker in the review file;
+        # strip it (and any '[severity]' tag) from comments the user kept so they
+        # post as clean prose.
+        if args.agent and isinstance(response, dict):
+            diff_section = response.get('diff') or {}
+            for line_comments in (diff_section.get('diff_comments') or {}).values():
+                for key, body in list(line_comments.items()):
+                    line_comments[key] = agent_review.strip_prefix(body)
+            if diff_section.get('comment'):
+                diff_section['comment'] = agent_review.strip_prefix(diff_section['comment'])
```
gsnedders, line 433: Why is this `.get` everywhere? `cls.invoke_wizard` is our own function, we should be able to guarantee its return type.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
```diff
@@ -0,0 +1,426 @@
...
+    """Strip the 'agent: ' marker (and a leading '[severity] ' tag) so that kept
+    agent findings post as clean prose. Accepts a string or a list of lines
+    (truncate_strs turns comment bodies into lists) and strips per line so
+    several stacked findings at one anchor all post cleanly."""
+    def strip_line(line):
+        if line.startswith(COMMENT_PREFIX):
+            line = line[len(COMMENT_PREFIX):]
+            line = re.sub(r'^\[[^\]]+\]\s*', '', line)
+        return line
+    if isinstance(value, list):
```
gsnedders, line 105: We shouldn't bother handling both list and non-list cases in one function — we should just expose `strip_line` and the callers can deal with whatever their type is.

gsnedders (changes requested):
`Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py` both parses and inserts annotations into diffs already, doesn't it? Can we not refactor that to just re-use the logic there?

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Done, it's the module docstring now.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Converted `agent_review.py` throughout, and used f-strings for the new code in `review.py` as well.

I left `review.py`'s pre-existing `.format` calls alone rather than fold a style sweep of the whole file into this change. Happy to do that as a follow-up if you'd rather it were consistent.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Applied. Confirmed in `webkitcorepy/subprocess_utils.py`: `capture_output=True` sets both `stdout` and `stderr` to `PIPE` unconditionally before calling `subprocess.run`, so `proc.stderr` can't be `None` here.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Agreed. It now writes up to the last 20 stderr lines, each indented, and says how many there were when it has to truncate.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Done, the module exposes `strip_line()` only.

Both call sites in `review.py` run after `truncate_strs()`, which has already split every comment body into a list of lines, so they just map over it. Simpler on both sides.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py
AndresGonzalezApple: Moved. The print now happens immediately before `invoke_wizard()`, after the agent run has finished.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py
AndresGonzalezApple: Dropped the `isinstance(response, dict)` guard, which was pure paranoia.

I kept `response.get('diff')` though: `invoke_wizard` deliberately deletes the key when there is nothing in it (`del response['diff']`, and the same for `metadata` and `comments`), and the rest of `main` already relies on that. Tightening the contract so the keys are always present looks worthwhile, but it would touch every existing consumer, so I would rather not fold it into this change.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Fixed.

Header detection still keys on the `--- ` / `+++ ` adjacency, but the path now comes from whichever side isn't `/dev/null`, via a small `_header_path()` helper. A deleted file gets a header, anchors and a file-level fallback like any other. `test_deleted_file_header` covers it with the deleted file first in the diff, which is the case that dropped its content outright.

Tightening it that way had a second benefit: `_header_path()` returns `None` when neither side looks like `a/<path>` or `b/<path>`, so a removed line rendering as `--- ...` followed by an added line rendering as `+++ ...` is now treated as content instead of being swallowed as a header pair. That's `test_header_pair_without_a_or_b_prefix_is_content`.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Agreed, removed.

A finding whose line misses on the side it names now falls through to the file-level anchor and is counted in the "not anchored to a visible diff line" total, so a mislabelled `side` surfaces as an admittedly approximate anchor instead of a confident wrong one. `test_side_mismatch_falls_back_to_file_level` replaces the test that asserted the old behaviour.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: It was intentional, but you're right that it's the wrong call: the agent re-raising something already discussed costs more than a slightly longer prompt.

Existing comments are now rendered into the annotated diff as `# prior comment: <line>`, and the instructions tell the agent they are context only. They still don't advance the line counters and can't be anchored to, since they aren't part of the file. `test_comment_block_surfaced_but_not_anchored` pins both halves of that.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Dropped `Task`. Whether a subagent inherits the parent's `--allowedTools` isn't something the docs commit to, and nothing about reviewing a diff needs it, so guessing either way was the wrong trade.

On the filesystem tools: `Read`/`Grep`/`Glob` are scoped to cwd plus any `--add-dir` grants, we pass `cwd=repo_root` and never pass `--add-dir`, so the reachable tree is the checkout. The more substantive change is `--permission-mode dontAsk` (see the other thread), which denies anything not pre-approved rather than prompting, so the allowlist is now the actual boundary instead of a property of the default path.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/program/review.py
AndresGonzalezApple: Both halves are fair, and neither is fully fixable from here: the diff comes from the API, against a merge base the checkout needn't even have fetched. What I've done:

- `Review.warn_if_not_pre_change_tree()` warns when the checkout has uncommitted changes, and when `merge_base(pull_request.base, HEAD) != HEAD`, i.e. HEAD isn't in the base branch's history. That catches running it from an unrelated branch, which is the case that produces confidently wrong findings. It is advisory only, and it degrades quietly when the base branch doesn't exist locally, since `find()` raises `ValueError` for an unresolvable ref and that's caught.
- For your second point, the instructions now tell the agent the checkout isn't guaranteed to be at the commit the diff was taken against, to treat the diff as authoritative, and to say so rather than report a finding when the file it reads disagrees.

Happy to make the mismatch fatal instead if you'd rather `--agent` refuse than warn.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Good catch, and I think the contradiction with the docs is explained by your `verbose` setting.

In the bundled CLI (2.1.219, same version you're on) the headless JSON path is effectively:

```js
case "json":
  ...
  if (verbose) { print(JSON.stringify(allMessages)); break }
  print(JSON.stringify(resultMessage)); break
```

So `--output-format json` emits the whole `SDKMessage` array when verbose is enabled, via `--verbose` or `"verbose": true` in settings, and just the single result object otherwise. The docs describe the non-verbose shape; you were getting the other one.

Rather than depend on the reviewer's config either way, `_result_message()` now accepts both: a dict is used as-is, a list is scanned in reverse for the last `type == "result"` message. `test_result_message_from_object` and `test_result_message_from_verbose_array` cover it.

Two related things from your paste:

- The fences around `result` are moot now. With `--json-schema` the findings come back as an already-parsed `structured_output` object rather than text to unwrap. See the schema thread.
- The unhelpful `(None)` in that warning is fixed. It was reporting the JSON parse error even when parsing had succeeded and it was the *result message* that was missing, which is exactly the case you hit.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Agreed, done.

`--json-schema` is present in 2.1.219 ("JSON Schema for structured output validation"), so the contract is now a module-level `FINDINGS_SCHEMA` passed on the command line, and the prompt no longer carries a hand-written schema or a "raw JSON ONLY, no markdown fences" plea.

Worth knowing how it's implemented, because it's stronger than I expected: the schema becomes a `StructuredOutput` tool the model must call, the CLI validates the call against the schema and retries on mismatch, and the validated object arrives on the result message as `structured_output`. On the happy path there is no text to parse at all. If the schema is rejected at init the CLI disables structured output and falls back to text, so that path is still handled, just strictly.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Agreed, and the laxness is gone. `extract_json` is now `parse_json`, a plain `json.loads` with a friendly error string: no fence stripping, no prose slicing, no trailing-comma repair. `test_markdown_fence_rejected`, `test_surrounding_prose_rejected` and `test_trailing_comma_rejected` pin that down.

On "are you actually seeing invalid JSON": yes, and your own paste in the output-envelope thread is an example, with the reply wrapped in a markdown fence (```` ```json ````). So the tolerance was empirically earned rather than superstition. What makes it unnecessary is `--json-schema`, not stricter faith in the model: on the happy path the findings now arrive as an already-validated object, so there is nothing to salvage.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Done. `REVIEW_INSTRUCTIONS` goes in via `--append-system-prompt`, and the pull request (title, description, annotated diff) is the user prompt on stdin. I used `--append-system-prompt` rather than `--system-prompt` to keep the CLI's own tool guidance in place.

It paid for itself beyond the risk split: the instructions are now a static string with nothing to substitute into them, which let me delete the single-pass `re.sub` templating that existed only so a PR title containing the literal text `{diff}` couldn't be interpolated. `test_review_instructions_are_not_a_template` asserts there's no `{` left in them.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Yes, switched.

For the record, the bundled CLI describes the two as: `default` is "Standard behavior, prompts for dangerous operations", `dontAsk` is "Don't prompt for permissions, deny if not pre-approved". So `dontAsk` is exactly the intent and `--allowedTools` is now the real boundary. `default` was still accepted (it's aliased to `manual`), so this wasn't broken before, it was just relying on headless mode having nowhere to prompt rather than on an explicit deny.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Intentional only in the sense that it wasn't considered, and I would like to keep it that way for now: the flag stays simple and picks up whatever the reviewer has configured.

If it turns out reviews need a pinned model, or an override so cost can be traded against depth, I would rather add the model parameters as their own change than grow this one.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Kept the text decoding: the output gets sliced for the stderr previews and written to a temp file for post-mortems, so bytes would only move the decode rather than remove it.

I did drop `errors='replace'`. On invalid UTF-8 I would rather raise into the existing handler, and degrade to the un-annotated review file, than hand `json.loads` a string that has been quietly corrupted with U+FFFD.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Fair challenge, and I went and read all four before answering. My conclusion is that none of them is a reusable "unified diff to (file, old line, new line)" walker; each is welded to the comment-addressing scheme of one particular host.

1. `remote/git_hub.py` -- `_line_to_position_map` is the closest: it walks a text diff keyed on `+++ b/` and `@@`. But it only produces **new-side** numbers, and its values are GitHub *positions*, which aren't usable for placing a comment into the review file.
2. `program/review.py` -- the fourth-section walker consumes a `difflib.ndiff` stream of the edited file against the baseline, and also produces GitHub-style positions. Different input, different output.
3. `program/diff/html_diff.py` -- renders to HTML.
4. `webkitpy/common/checkout/diff_parser.py` -- different library, not importable from here.

The fourth remote is the interesting one: BitBucket/Stash never parses a text diff at all. `bitbucket.py`'s `_position_converters` walks Stash's **JSON** diff (`segment['lines']`, `line['source']` / `line['destination']`), and the text diff is synthesized afterwards by `json_to_diff()`. So reusing GitHub's helper would tie `--agent` to GitHub and break it for Stash.

What `diff_maps` needs that none of the four provides: old-side numbers as well as new, an index back into `diff_lines` so a finding can be inserted in the right place, and handling of the `>>>>`/`<<<<` blocks (which it now surfaces to the agent rather than dropping). Its input is `pull_request.diff(comments=True)`, the one representation both remotes converge on, which is what keeps this portable across remotes -- and, for that matter, across repositories and hosts generally, which is something I would like to preserve.

I would still take a shared walker if you want one, but it would mean reworking `_line_to_position_map` to yield both sides plus indices and having GitHub derive its positions from that -- a change to tested GitHub comment placement, which felt like the wrong thing to fold into this PR. Say the word if you'd prefer it here.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Same answer as [the `diff_maps` thread](https://github.com/WebKit/WebKit/pull/67226#discussion_r3724995607).

Short version: `review.py`'s walker works on a `difflib.ndiff` stream and produces GitHub positions, whereas `annotate_diff` needs indices into `diff_lines` to insert at, and Stash's comment addressing doesn't go through a text diff at all. Details in that thread.

Tools/Scripts/libraries/webkitscmpy/webkitscmpy/agent_review.py
AndresGonzalezApple: Honest answer: I haven't measured it against the alternative, so I can't yet claim the tags earn their place.

What I can say is why they're there rather than trusting attribution from a plain unified diff. The tag is the thing the model has to quote back (`line` plus `side`), so `annotate_diff` can look it up exactly and count what it couldn't place: that's the "N not anchored to a visible diff line" number printed at the end of a run. Misattribution surfaces as a file-level fallback rather than silently landing on the wrong line. Two of the findings in this PR were about exactly that failure mode, and both were findable because the anchoring is checkable.

I'll get numbers, though: placed vs. not-anchored across a handful of PRs, with the tags and with a plain diff, and I'll drop the tagging if it makes no difference. Will report back on this thread.

AndresGonzalezApple:
@gsnedders thanks, this was a genuinely useful round. New revision pushed; every inline thread has a reply, but the ones worth calling out:

- **Output envelope** — root-caused. `--output-format json` emits the full `SDKMessage` array when `verbose` is enabled and just the result object otherwise, which is why your run contradicted the docs. It now handles both shapes rather than depending on anyone's config.
- **Structured output** — `--json-schema` is wired up, so findings come back as a CLI-validated `structured_output` object. That let me delete all the lenient JSON parsing, which was the right call for the reason you gave, just via the schema rather than via stricter faith in the model.
- **Prompt split** — instructions via `--append-system-prompt`, the PR as the user prompt. Side benefit: the instructions are now a static string, so the `re.sub` templating that existed only to stop a PR title interpolating itself is gone.
- **Permissions** — `--permission-mode dontAsk`, and `Task` dropped from the allowlist, so the allowlist is the actual boundary rather than a property of the default path.
- **Two anchoring bugs fixed**: deleted files (`+++ /dev/null`) now get a header and anchors, and the cross-side line-number fallback is gone in favour of an honest file-level anchor.
- **Prior PR comments** are now shown to the agent as context so it stops re-raising settled points.
- **Base-tree mismatch** now warns, and the agent is told the checkout may not match the diff.

Two I didn't simply do as asked:

- **A fifth diff parser**: I read all four and don't think any is reusable here — GitHub's walker yields new-side numbers and GitHub *positions*, `review.py`'s works on an `ndiff` stream, and Stash doesn't parse a text diff at all (it walks the JSON diff and synthesizes the text afterwards). Reusing either would tie `--agent` to one host. Reasoning and the offer to do a shared walker anyway are in that thread.
- **Model selection**: still whatever the reviewer has configured. Keeping the flag simple; model parameters can be their own change if they turn out to be needed.

Still outstanding on my side: actual measurements for the line-number tagging question, which I'll post on that thread rather than hand-wave.

Unrelated, and I'll file it separately: `git_hub.py`'s `reviewers()` iterates `request('pulls/N/reviews')` without a `None` check, so a transient API failure tracebacks `git-webkit review` rather than degrading. I hit it while testing this (the endpoint was 503/404ing for a while earlier today, it has since recovered).

AndresGonzalezApple:
@gsnedders — gentle nudge on this one. A new revision (`1f81e72`) went up yesterday and all 25 open threads now have replies, so it's waiting on another look to clear the `CHANGES_REQUESTED`.

Most of your points are simply done: the JSON schema, the system/user prompt split, `dontAsk`, dropping `Task`, the docstring, f-strings, the stderr handling, `strip_line`, and both anchoring bugs (deleted files, and the cross-side line-number guess). I also root-caused the output-envelope mismatch you hit — it's the `verbose` setting, details in [that thread](https://github.com/WebKit/WebKit/pull/67226#discussion_r3732319441).

Three I answered rather than acted on, and those are the ones actually worth your time:

- [The fifth diff parser](https://github.com/WebKit/WebKit/pull/67226#discussion_r3724995607) — I read all four before answering. None is a reusable "unified diff to (file, old line, new line)" walker; the clincher is that Stash never parses a text diff at all, so reusing GitHub's helper would break BitBucket. I offered to build a shared walker anyway if you'd still prefer it.
- [Model selection](https://github.com/WebKit/WebKit/pull/67226#discussion_r3731409291) — left as whatever the reviewer has configured, on keep-it-simple grounds, with model parameters as a possible follow-up.
- [The N/O line tagging](https://github.com/WebKit/WebKit/pull/67226#discussion_r3724815120) — I haven't measured it against a plain unified diff, so I said so instead of claiming it works. Numbers to follow on that thread.

EWS on this revision is 37/38 green. The one red is `win-tests`, which failed with "Failed to download built product from S3" rather than on anything in the change, and the `webkitpy` failure from the previous revision is gone.

@JonWBedard @facetothefate @briannafan — you're on the reviewer request list here as well, if any of you has cycles for a `webkitscmpy` change.

AndresGonzalezApple:
Another gentle ping — this is still blocked on review, and nothing has moved since my last comment: all 25 threads answered, EWS 37/38 with the only red being `win-tests` failing to download its built product from S3.

@gsnedders if you're heads-down on other things, genuinely no problem — I'd rather this go to whoever has cycles than sit. The only threads that still need a decision rather than a code change are the three where I pushed back instead of complying: [the diff parser](https://github.com/WebKit/WebKit/pull/67226#discussion_r3724995607), model selection, and the line-number tagging. Everything else you raised is done.

@JonWBedard @facetothefate @briannafan — you're all on the reviewer request list here, if one of you would rather pick this up.

AndresGonzalezApple:
Third ping, and I'll make this one a concrete ask rather than just a bump.

It's been about eleven days since the last nudge and two weeks since the revision went up. Every one of the 25 threads has a reply, and EWS is 37/38 with the only red being `win-tests` failing to download its built product from S3.

@gsnedders your `CHANGES_REQUESTED` is what's blocking this, so either of these unblocks me and I'm equally happy with both:

1. Re-review it. Only three threads need a decision rather than a code change — [the diff parser](https://github.com/WebKit/WebKit/pull/67226#discussion_r3724995607), [model selection](https://github.com/WebKit/WebKit/pull/67226#discussion_r3731409291), and [the line-number tagging](https://github.com/WebKit/WebKit/pull/67226#discussion_r3724815120). Everything else you raised is done.
2. Dismiss the review and hand it off, if you'd rather not carry this one. No hard feelings at all — I'd just like it to move.

@JonWBedard @facetothefate @briannafan — you're all on the request list. This is a `webkitscmpy`-only change, so if one of you can take it instead, that works too.

If there's a better venue for finding a `webkitscmpy` reviewer than pinging here, I'm glad to take it there instead.

PR #67301, opened 2026-06-16
Short circuit same-realm MessagePort message delivery

francovs:
#### f7a7c2332942260c44489ff33acecbd87a1152e4
<pre>
Short circuit same-realm MessagePort message delivery
<a href="https://bugs.webkit.org/show_bug.cgi?id=317242">https://bugs.webkit.org/show_bug.cgi?id=317242</a>
<a href="https://rdar.apple.com/168247413">rdar://168247413</a>

Reviewed by Chris Dumez.

A path that does not involve IPC nor thread-hopping was added to
MessagePort::postMessage(). It is only used when both the sending and
receiving ends have never been shipped. This use case is not uncommon
because using a port pair simply for scheduling reasons is a known
workaround for nested setTimeout() timing restrictions.

When any of the ends is disentangled, all unhandled messages are
drained to the MessagePortChannelProvider, and the local path is
never used again.

One new test verifying the correctness of port transfer, for an
already started local port:
* LayoutTests/fast/events/message-port-transfer-preserves-pending-message.html

Three microbenchmarks were added:
* PerformanceTests/DOM/MessagePort/throughput-same-context-batch.html
* PerformanceTests/DOM/MessagePort/throughput-same-context-new-ports.html
* PerformanceTests/DOM/MessagePort/throughput-same-context.html

throughput-same-context-new-ports.html is of extra relevance because
it is analogous to a real requestIdleCallback() polyfill.

Canonical link: <a href="https://commits.webkit.org/315709@main">https://commits.webkit.org/315709@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/315a9aa60617b8d56538b5e15b67cdbca494b1af

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/169680 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/43602 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/36965 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/179039 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/127392 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/d76e32c6-c4b5-4400-8b23-3f2fc137730a) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/171546 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/44394 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/43231 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/132226 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/93141 "9 failures") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/6c17c524-6e3b-4f74-bbad-95018079a056) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/172639 "Passed tests") | [  ~~🧪 ios-wk2~~](https://ews-build.webkit.org/#/builders/162/builds/35268 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/152605 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/112729 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/a3bd55c3-e77d-4739-9965-16971d68d4fe) 
| | [  ~~🧪 ios-wk2-wpt~~](https://ews-build.webkit.org/#/builders/154/builds/33982 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/32365 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/27736 "Built successfully") | | 
| [⏳ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/JSC-Tests-x86-64-EWS "Waiting to run tests") | [  ~~🧪 api-ios~~](https://ews-build.webkit.org/#/builders/13/builds/144343 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/32004 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/181638 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/34346 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/36531 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/140673 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/43258 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/152075 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/140508 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/37868 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/43947 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/28984 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/108193 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/35776 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/115712 "Built successfully") | | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/42457 "Built successfully") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/7077 "") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/41850 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/42105 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/41993 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/dom/MessagePort.cpp
+++ b/Source/WebCore/dom/MessagePort.cpp
@@ -205,6 +205,17 @@ ExceptionOr<void> MessagePort::postMessage(JSC::JSGlobalObject& globalObject, JS
 
     LOG(MessagePorts, "Actually posting message to port %s (to be received by port %s)", m_identifier.logString().utf8().data(), m_remoteIdentifier.logString().utf8().data());
 
+    if (RefPtr partner = m_localPartner) {
+        partner->m_localQueue.append(WTF::move(message));
+        ++partner->m_newLocalMessages;
+        if (partner->started()) {
+            queueTaskKeepingObjectAlive(*partner, TaskSource::PostedMessageQueue, [](auto& port) mutable {
+                port.dispatchMessages();
+            });
+        }
+        return { };
+    }
+
     protect(MessagePortChannelProvider::fromContext(*protect(scriptExecutionContext())))->postMessageToRemote(WTF::move(message), m_remoteIdentifier);
     return { };
 }
@@ -220,6 +231,17 @@ TransferredMessagePort MessagePort::disentangle()
     m_entangled = false;
 
     Ref context = *scriptExecutionContext();
+    if (RefPtr localSibling = m_localPartner) {
+        localSibling->m_localPartner = nullptr;
+        m_localPartner = nullptr;
+    }
+
+    // Drain messages before disentanglement:
+    auto localMessagesToQueue = std::exchange(m_localQueue, { });
+    m_newLocalMessages = 0;
+    for (auto& message : localMessagesToQueue)
+        protect(MessagePortChannelProvider::fromContext(context))->postMessageToRemote(WTF::move(message), m_identifier);
+
     protect(MessagePortChannelProvider::fromContext(context))->messagePortDisentangled(m_identifier);
 
     // We can't receive any messages or generate any events after this, so remove ourselves from the list of active ports.
@@ -256,6 +278,10 @@ void MessagePort::start()
         return;
 
     m_started = true;
+
+    if (m_localPartner.get() && m_localQueue.isEmpty())
+        return;
+
     protect(scriptExecutionContext())->processMessageForPortSoon(m_identifier, [pendingActivity = makePendingActivity(*this)] { });
 }
 
@@ -264,6 +290,12 @@ void MessagePort::close()
     if (m_isDetached)
         return;
     m_isDetached = true;
+
+    m_localQueue.clear();
+    m_newLocalMessages = 0;
+    if (RefPtr partner = m_localPartner)
+        partner->m_localPartner = nullptr;
+    m_localPartner = nullptr;
 
     ensureOnMainThread([identifier = m_identifier] {
         MessagePortChannelProvider::singleton().messagePortClosed(identifier);
@@ -291,6 +323,20 @@ void MessagePort::dispatchMessages()
     if (!context || context->activeDOMObjectsAreSuspended() || !isEntangled())
         return;
 
+    LOG(MessagePorts, "Dispatching messages on MessagePort %s (%p)", m_identifier.logString().utf8().data(), this);
+    while (m_newLocalMessages) {
+        --m_newLocalMessages;
+        queueTaskKeepingObjectAlive(*this, TaskSource::PostedMessageQueue, [](auto& port) {
+            LOG(MessagePorts, "Draining one local message on MessagePort %s (%p)", port.m_identifier.logString().utf8().data(), &port);
+            port.drainOneLocalMessage();
+        });
+    }
+
+    // Scheduling message handling on a locally entangled port can cause races if
+    // the port is later disentangled:
+    if (m_localPartner.get())
+        return;
+
     auto messagesTakenHandler = [pendingActivity = makePendingActivity(*this)](Vector<MessageWithMessagePorts>&& messages, CompletionHandler<void()>&& completionCallback) mutable {
         auto scopeExit = makeScopeExit(WTF::move(completionCallback));
 
@@ -313,7 +359,7 @@ void MessagePort::dispatchMessages()
 
             if (pendingActivity->object().m_messageHandler) {
                 ASSERT(message.transferredPorts.isEmpty());
-                pendingActivity->object().m_messageHandler(*downcast<JSDOMGlobalObject>(globalObject), message.message.releaseNonNull().get());
+                pendingActivity->object().m_messageHandler(*downcast<JSDOMGlobalObject>(globalObject), message.message.releaseNonNull());
                 continue;
             }
 
@@ -335,9 +381,43 @@ void MessagePort::dispatchMessages()
     protect(MessagePortChannelProvider::fromContext(*context))->takeAllMessagesForPort(m_identifier, WTF::move(messagesTakenHandler));
 }
 
+void MessagePort::drainOneLocalMessage()
+{
+    RefPtr context = scriptExecutionContext();
+    if (!context || !context->globalObject() || context->activeDOMObjectsAreSuspended() || !isEntangled())
+        return;
+
+    ASSERT(context->isContextThread());
+    if (RefPtr workerGlobalScope = dynamicDowncast<WorkerGlobalScope>(*context)) {
+        if (workerGlobalScope->isClosing())
+            return;
+    }
+
+    auto* globalObject = context->globalObject();
+    Ref vm = globalObject->vm();
+    auto scope = DECLARE_TOP_EXCEPTION_SCOPE(vm);
+
+    auto message = m_localQueue.takeFirst();
+
+    if (m_messageHandler) {
+        ASSERT(message.transferredPorts.isEmpty());
+        m_messageHandler(*downcast<JSDOMGlobalObject>(globalObject), message.message.releaseNonNull());
+        return;
+    }
+
+    auto ports = MessagePort::entanglePorts(*context, WTF::move(message.transferredPorts));
+    auto event = MessageEvent::create(*globalObject, message.message.releaseNonNull(), { }, { }, { }, WTF::move(ports));
+    if (scope.exception()) [[unlikely]] {
+        // Currently, we assume that the only way we can get here is if we have a termination.
+        RELEASE_ASSERT(vm->hasPendingTerminationException());
+        return;
+    }
+    dispatchEvent(event.event);
+}
+
 void MessagePort::dispatchEvent(Event& event)
 {
-    if (m_isDetached)
+    if (!isEntangled())
         return;
 
     if (RefPtr globalScope = dynamicDowncast<WorkerGlobalScope>(scriptExecutionContext())) {
@@ -404,6 +484,19 @@ Ref<MessagePort> MessagePort::entangle(ScriptExecutionContext& context, Transfer
     Ref port = MessagePort::create(context, transferredPort.first, transferredPort.second);
     port->entangle();
     return port;
+}
+
+// FIXME: avoid SUPRESS_NODELETE with lazyInitialize() - ThreadSafeWeakPtr still unsupported
+SUPPRESS_NODELETE void MessagePort::entangleLocally(MessagePort& port1, MessagePort& port2)
+{
+    ASSERT(port1.scriptExecutionContext() == port2.scriptExecutionContext());
+    ASSERT(port2.identifier() == port1.m_remoteIdentifier);
+    ASSERT(port1.identifier() == port2.m_remoteIdentifier);
+    // Assertions ensure the assignment doesn't violate NODELETE.
+    ASSERT(!port1.m_localPartner.get());
+    ASSERT(!port2.m_localPartner.get());
+    port1.m_localPartner = port2;
+    port2.m_localPartner = port1;
 }
 
 bool MessagePort::addEventListener(const AtomString& eventType, Ref<EventListener>&& listener, const AddEventListenerOptions& options)
--- a/Source/WebCore/dom/MessagePort.h
+++ b/Source/WebCore/dom/MessagePort.h
@@ -32,6 +32,7 @@
 #include <WebCore/MessagePortChannel.h>
 #include <WebCore/MessagePortIdentifier.h>
 #include <WebCore/MessageWithMessagePorts.h>
+#include <wtf/Deque.h>
 #include <wtf/ThreadSafeWeakPtr.h>
 #include <wtf/WeakPtr.h>
 
@@ -108,12 +109,17 @@ public:
     TransferredMessagePort disentangle();
     static Ref<MessagePort> entangle(ScriptExecutionContext&, TransferredMessagePort&&);
 
+    // Short-circuits message delivery for same-context ports. Should only be used for
+    // ports that have never been shippped, to avoid potential message reordering.
+    static void NODELETE entangleLocally(MessagePort&, MessagePort&);
+
 private:
     MessagePort(ScriptExecutionContext&, const MessagePortIdentifier& local, const MessagePortIdentifier& remote);
 
     bool addEventListener(const AtomString& eventType, Ref<EventListener>&&, const AddEventListenerOptions&) final;
     using EventTarget::addEventListener;
     bool removeEventListener(const AtomString& eventType, EventListener&, const EventListenerOptions&) final;
+    void drainOneLocalMessage();
 
     // ActiveDOMObject.
     void contextDestroyed() final;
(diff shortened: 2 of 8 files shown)


Source/WebCore/dom/MessagePort.h
```diff
@@ -108,12 +108,17 @@ class MessagePort final : public ActiveDOMObject, public EventTarget, public Thr
     TransferredMessagePort disentangle();
     static Ref<MessagePort> entangle(ScriptExecutionContext&, TransferredMessagePort&&);
 
+    // Short-circuits message delivery for same-context ports. Should only be used for
+    // ports that have never been shippped, to avoid potential message reordering.
+    static void entangleLocally(ThreadSafeWeakPtr<MessagePort>, ThreadSafeWeakPtr<MessagePort>);
```
cdumez: We shouldn't be passes ThreadSafeWeakPtr as parameter, just pass in C++ references since this is what the call site has.

Source/WebCore/dom/MessagePort.cpp
```diff
@@ -406,6 +436,19 @@ Ref<MessagePort> MessagePort::entangle(ScriptExecutionContext& context, Transfer
     return port;
 }
 
+void MessagePort::entangleLocally(ThreadSafeWeakPtr<MessagePort> port1, ThreadSafeWeakPtr<MessagePort> port2)
+{
+    auto port1Ptr = port1.get();
```
cdumez: These 2 locals can go away.

Source/WebCore/dom/MessagePort.cpp
```diff
@@ -406,6 +436,19 @@ Ref<MessagePort> MessagePort::entangle(ScriptExecutionContext& context, Transfer
...
+    auto port2Ptr = port2.get();
+    ASSERT(port1Ptr);
```
cdumez: These 2 asserts can go away.

Source/WebCore/dom/MessagePort.cpp
```diff
@@ -205,6 +205,17 @@ ExceptionOr<void> MessagePort::postMessage(JSC::JSGlobalObject& globalObject, JS
 
     LOG(MessagePorts, "Actually posting message to port %s (to be received by port %s)", m_identifier.logString().utf8().data(), m_remoteIdentifier.logString().utf8().data());
 
+    if (auto sibling = m_localPartner.get()) {
```
cdumez: In new code, never hide smart pointers behind `auto`, as per the safer cpp guidelines. This should just be:
`if (RefPtr sibling = m_localPartner) {`

Source/WebCore/dom/MessagePort.cpp
```diff
@@ -220,6 +231,16 @@ TransferredMessagePort MessagePort::disentangle()
     m_entangled = false;
 
     Ref context = *scriptExecutionContext();
+    if (auto localSibling = m_localPartner.get()) {
```
cdumez: ditto.

Source/WebCore/dom/MessagePort.cpp
```diff
@@ -291,53 +312,62 @@ void MessagePort::dispatchMessages()
...
-            auto ports = MessagePort::entanglePorts(*context, WTF::move(message.transferredPorts));
-            auto event = MessageEvent::create(*globalObject, message.message.releaseNonNull(), { }, { }, { }, WTF::move(ports));
-            if (scope.exception()) [[unlikely]] {
-                // Currently, we assume that the only way we can get here is if we have a termination.
-                RELEASE_ASSERT(vm->hasPendingTerminationException());
-                return;
-            }
+        if (m_messageHandler) {
+            ASSERT(message.transferredPorts.isEmpty());
+            m_messageHandler(*downcast<JSDOMGlobalObject>(globalObject), message.message.releaseNonNull().get());
```
cdumez: If you don't need a `Ref&&&` (and it seems you don't since you're calling `.get()`, then `*message.message` would be much simpler.

Source/WebCore/dom/MessagePort.cpp
```diff
@@ -291,53 +312,62 @@ void MessagePort::dispatchMessages()
...
+            continue;
+        }
 
-            // Per specification, each MessagePort object has a task source called the port message queue.
-            queueTaskKeepingObjectAlive(pendingActivity->object(), TaskSource::PostedMessageQueue, [event = WTF::move(event)](auto& port) {
-                port.dispatchEvent(event.event);
-            });
+        auto ports = MessagePort::entanglePorts(*context, WTF::move(message.transferredPorts));
+        auto event = MessageEvent::create(*globalObject, message.message.releaseNonNull(), { }, { }, { }, WTF::move(ports));
```
cdumez: Never hide smart pointers behind `auto`.

Source/WebCore/dom/MessagePort.h
```diff
@@ -132,6 +137,8 @@ class MessagePort final : public ActiveDOMObject, public EventTarget, public Thr
     MessagePortIdentifier m_remoteIdentifier;
 
     MessageHandler m_messageHandler;
+    Vector<WebCore::MessageWithMessagePorts> m_localQueue;
```
cdumez: Unnecessary `WebCore::`.

Source/WebCore/dom/MessagePort.cpp
```diff
@@ -205,6 +205,17 @@ ExceptionOr<void> MessagePort::postMessage(JSC::JSGlobalObject& globalObject, JS
 
     LOG(MessagePorts, "Actually posting message to port %s (to be received by port %s)", m_identifier.logString().utf8().data(), m_remoteIdentifier.logString().utf8().data());
 
+    if (auto sibling = m_localPartner.get()) {
+        sibling->m_localQueue.append(WTF::move(message));
+        if (!sibling->m_started)
+            return { };
+
+        queueTaskKeepingObjectAlive(*sibling, TaskSource::PostedMessageQueue, [](auto& port) mutable {
```
cdumez: The batch benchmark sends 10000 messages synchronously → 10000 appends + 10000 queued tasks, of which the first drains the whole queue and the other 9999 are no-ops *I think* (m_localQueue empty, partner non-null → nothing). It is correct, but it undercuts the throughput the benchmark is meant to show. Since `dispatchMessages` always drains the entire queue, I believe you can coalesce by only scheduling when the queue transitions empty→non-empty:
```
  bool needsDispatch = sibling->m_localQueue.isEmpty();
  sibling->m_localQueue.append(WTF::move(message));
  if (sibling->m_started && needsDispatch)
      queueTaskKeepingObjectAlive(*sibling, ...);
```

cdumez (commented):
`MessagePort::close()` sets `m_isDetached` to true but leaves m_localPartner intact on both ends and leaves m_localQueue populated. A postMessage to a closed sibling still takes the fast path, appends to the dead sibling's m_localQueue, and schedules a dispatchMessages that early-returns on !isEntangled(). Net effect (message dropped) matches the channel path, so it's not a correctness bug, but the message lingers in the closed port's m_localQueue until GC. We should consider clearing m_localQueue and breaking the partnership in close() for hygiene and to release the serialized payloads promptly.

Source/WebCore/dom/MessagePort.cpp
francovs: This one is `MessageEvent::MessageEventWithStrongData`, not `Ref<MessageEvent>`. Still removed the `auto`.

cdumez: Oh, `auto` was fine in that case. I did not realize this did not return a smart pointer.

cdumez:
Would you mind adding the following test case to your PR?
```
<!DOCTYPE html>
<html>
<body>
<p>Test that a message already posted to a same-context MessagePort is not lost when
that port is transferred from within another port's message handler. The message
should be delivered either to the port in this context (before it is neutered) or
to the re-entangled port in the destination, but it must not be silently dropped.
See https://bugs.webkit.org/show_bug.cgi?id=317242</p>
<pre id="log"></pre>
<script>
"use strict";

if (window.testRunner) {
    testRunner.dumpAsText();
    testRunner.waitUntilDone();
}

function log(message)
{
    document.getElementById("log").textContent += message + "\n";
}

let toBDelivered = false;

function finish()
{
    if (toBDelivered)
        log("PASS: 'toB' was delivered.");
    else
        log("FAIL: 'toB' was silently lost.");

    if (window.testRunner)
        testRunner.notifyDone();
}

// Channel whose port we will transfer mid-flight.
const channel1 = new MessageChannel();
const a = channel1.port1;
const b = channel1.port2;

// Channel used purely to transfer "b" to a new entangled port in this same context.
const channel2 = new MessageChannel();
const sender = channel2.port1;
const receiver = channel2.port2;

// If "toB" is delivered to "b" in this context (before "b" is neutered), record it.
b.onmessage = (event) => {
    if (event.data === "toB")
        toBDelivered = true;
};

// "a" receives "toA" first. Its handler runs *after* the message-dispatch tasks for
// both ports have already drained their queues, which is exactly the window in which
// the bug drops the in-flight "toB".
a.onmessage = () => {
    sender.postMessage("transfer", [b]);
};

// If "toB" instead travels with the transferred port, the re-entangled port receives it.
receiver.onmessage = (event) => {
    const transferredPort = event.ports[0];
    transferredPort.onmessage = (ev) => {
        if (ev.data === "toB")
            toBDelivered = true;
    };
    transferredPort.start();
};

// Post both messages synchronously so their dispatch tasks land in the same batch:
// "toA" -> a (triggers the transfer), "toB" -> b (the in-flight message under test).
b.postMessage("toA");
a.postMessage("toB");

// Give every queued task ample time to run, then report the outcome deterministically
// (rather than relying on a delivery that may never happen).
setTimeout(finish, 100);
</script>
</body>
</html>
```

I spotted a potential regression in your PR and asked Claude to write a regression test for it. I have verified that the test is passing *without* your PR (as well as in Chrome and Firefox). Please make sure this test passes *with* your PR (I did not verify but I suspect it will fail).

Source/WebCore/dom/MessagePort.cpp
```diff
@@ -265,6 +289,11 @@ void MessagePort::close()
         return;
     m_isDetached = true;
 
+    m_localQueue.clear();
+    if (RefPtr partner = m_localPartner.get())
```
cdumez: I doubt we need `.get()` here.

Source/WebCore/dom/MessagePort.cpp
francovs: The temporary [was actually necessary due SaferCpp](https://ews-build.s3-us-west-2.amazonaws.com/macOS-Safer-CPP-Checks-EWS/5fa81b37-115323/scan-build-output/StaticAnalyzerRegressions/WebKitLegacy/StaticAnalyzerReports/report-MessagePort.cpp-dispatchMessages-21-dd4856.html#EndPath).

~To simplify, changing `MessageHandler` to take `Ref<SerializedScriptValue>` instead of a raw reference, avoiding the `.get()`.~

Source/WebCore/dom/MessagePort.h
```diff
@@ -72,7 +72,7 @@ class MessagePort final : public ActiveDOMObject, public EventTarget, public Thr
     void close();
     void entangle();
 
-    using MessageHandler = Function<void(JSDOMGlobalObject&, SerializedScriptValue&)>;
+    using MessageHandler = Function<void(JSDOMGlobalObject&, Ref<SerializedScriptValue>)>;
```
cdumez: Why this change. The previous code was better, we don't want to pass smart pointers around like this as parameter, not unless they are r-value reference. The previous code was better as it avoided ref-counting churn.

cdumez (approved):
LGTM

francovs:
`LayoutTests/fast/events/message-port-transfer-preserves-pending-message.html` is only passing because port b consistently handles the message before being shipped, causing the test to not exercise the behavior it was supposed to. This only happens due the `processMessageForPortSoon()` call on `start()`. Adding a delay before doing the two `postMessage()` calls turns the test into a consistent failure.

Failure doesn't reproduce on main. Will have to rework the dispatch logic to ensure messages only leave the local queue at the last possible moment.

Source/WebCore/dom/MessagePort.h
cdumez: I think we should use an unsigned type here.

Source/WebCore/dom/MessagePort.cpp
cdumez: It is a little sad to construct `messagesTakenHandler` above when `m_localPartner.get()` is non-null. I think we should do a:
```
if (m_localPartner.get())
    return;
```
earlier, before we even construct the messagesTakenHandler.

cdumez (approved):
LGTM with nits.

PR #67336, opened 2026-06-17
Extend CSS::CustomIdent to support an unresolved ident() function alternative

sakupi01:
#### 7f655a4e9f5df4f6278acbe61596f8e641b06d83
<pre>
Extend CSS::CustomIdent to support an unresolved ident() function alternative
<a href="https://bugs.webkit.org/show_bug.cgi?id=284895">https://bugs.webkit.org/show_bug.cgi?id=284895</a>

Reviewed by Tim Nguyen.

Change CSS::CustomIdent&apos;s representation from a plain AtomString to
Variant&lt;AtomString, IdentFunction&gt;, where IdentFunction models
&lt;ident()&gt; = ident( &lt;ident-arg&gt;+ ), &lt;ident-arg&gt; = &lt;string&gt; | &lt;integer&gt; | &lt;ident&gt;
(<a href="https://drafts.csswg.org/css-values-5/#ident).">https://drafts.csswg.org/css-values-5/#ident).</a> Arguments are stored
unmerged so that the specified value serializes back in function form,
as required by WPT. Resolution happens in ToStyle&lt;CSS::CustomIdent&gt;,
which evaluates &lt;integer&gt; arguments (including math functions) with
BuilderState context and concatenates the argument codepoints.

Remove CustomIdent::tryResolved(), as no call site needs eager
resolution. The @counter-style/@view-transition descriptor converters
use only the plain identifier alternative, since ident() is invalid in
descriptors per the CSSWG resolution &quot;ident() is only valid within an
element context&quot; (<a href="https://github.com/w3c/csswg-drafts/issues/12219)">https://github.com/w3c/csswg-drafts/issues/12219)</a>; a
FIXME in consumeUnresolvedCustomIdent() records the parser-side
rejection needed once ident() parsing lands.
ShorthandSerializer::valueIDIncludingCustomIdent() likewise checks only
the plain identifier alternative, because the animation-name
disambiguation is about lexical ambiguity of the serialized text, which
the ident() function form can never cause. The deprecated CSSOM string
accessors return the plain identifier, falling back to the
function-form serialization.

No behavior change: nothing constructs the new alternative yet. All
code that read CustomIdent&apos;s string directly was updated to handle the
unresolved alternative explicitly.
Also add a CSSCustomIdentValue case to
CSSValue::collectComputedStyleDependencies so that a future ident()
with computed-value-time dependencies is correctly rejected in
computationally independent contexts (@property initial values).

* Source/WebCore/css/CSSCounterStyleDescriptors.cpp:
(WebCore::symbolFromCSSValue):
(WebCore::fallbackNameFromCSSValue):
(WebCore::extendsSystemAndFirstSymbolValueFromStyleProperties):
* Source/WebCore/css/CSSCounterValue.cpp:
(WebCore::CSSCounterValue::customCSSText):
* Source/WebCore/css/CSSCustomIdentValue.cpp:
(WebCore::CSSCustomIdentValue::stringValue):
* Source/WebCore/css/CSSValue.cpp:
(WebCore::CSSValue::collectComputedStyleDependencies):
* Source/WebCore/css/CSSValueKeywords.in:
* Source/WebCore/css/CSSViewTransitionRule.cpp:
(WebCore::StyleRuleViewTransition::StyleRuleViewTransition):
* Source/WebCore/css/DeprecatedCSSOMPrimitiveValue.cpp:
(WebCore::DeprecatedCSSOMPrimitiveValue::getStringValue):
(WebCore::DeprecatedCSSOMPrimitiveValue::getCounterValue):
* Source/WebCore/css/ShorthandSerializer.cpp:
(WebCore::ShorthandSerializer::valueIDIncludingCustomIdent):
* Source/WebCore/css/calc/CSSCalcTree+Serialization.cpp:
(WebCore::CSSCalc::serializeMathFunctionArguments):
(WebCore::CSSCalc::serializeCalculationTree):
* Source/WebCore/css/parser/CSSPropertyParserConsumer+Ident.cpp:
(WebCore::CSSPropertyParserHelpers::consumeUnresolvedCustomIdent):
(WebCore::CSSPropertyParserHelpers::consumeUnresolvedCustomIdentExcluding):
(WebCore::CSSPropertyParserHelpers::consumeUnresolvedDashedIdent):
* Source/WebCore/css/parser/CSSPropertyParserConsumer+ViewTransition.cpp:
(WebCore::CSSPropertyParserHelpers::consumeViewTransitionTypes):
* Source/WebCore/css/typedom/CSSStyleValueFactory.cpp:
(WebCore::CSSStyleValueFactory::reifyValue):
* Source/WebCore/css/values/primitives/CSSCustomIdent.cpp:
(WebCore::CSS::CustomIdent::isNull):
(WebCore::CSS::CustomIdent::startsWith):
(WebCore::CSS::CustomIdent::operator==):
(WebCore::CSS::Serialize&lt;IdentFunctionIdent&gt;::operator()):
(WebCore::CSS::Serialize&lt;CustomIdent&gt;::operator()):
(WebCore::CSS::ComputedStyleDependenciesCollector&lt;CustomIdent&gt;::operator()):
(WebCore::CSS::CSSValueChildrenVisitor&lt;CustomIdent&gt;::operator()):
(WebCore::CSS::operator&lt;&lt;):
(WebCore::CSS::add):
* Source/WebCore/css/values/primitives/CSSCustomIdent.h:
* Source/WebCore/style/StyleBuilder.cpp:
(WebCore::Style::Builder::applyProperty):
* Source/WebCore/style/values/primitives/StyleCustomIdent.cpp:
(WebCore::Style::ToStyle&lt;CSS::CustomIdent&gt;::operator()):

Canonical link: <a href="https://commits.webkit.org/316524@main">https://commits.webkit.org/316524@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/1122c6b3ba47834ba7e86b0a5a3995625730beae

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/172216 "4 style errors") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/46133 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/39553 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/181662 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/129770 "Built successfully") 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/174081 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/47422 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/45773 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/133951 "Passed tests") | [❌ 🧪 win-tests](https://ews-build.webkit.org/#/builders/60/builds/94493 "2 flakes 23 failures") 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/175174 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/37382 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/154495 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/114424 "Passed tests") | 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/36013 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/34286 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/30272 "Built successfully") | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/146451 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/34001 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/184200 "Built successfully") | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/36804 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/38488 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/142704 "Passed tests") | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/45800 "Built successfully") | | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/142586 "Passed tests") | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/38597 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/46480 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/30847 "Passed tests") | [  ~~🛠 playstation~~](https://ews-build.webkit.org/#/builders/134/builds/111183 "The change is no longer eligible for processing.") | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/37672 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/118235 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/44998 "Built successfully") | [❌ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/8941 "") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/44398 "Built successfully") | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/44630 "Built successfully") | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/44457 "Built successfully") | | |

### DIFF
--- a/Source/WebCore/css/parser/CSSPropertyParserConsumer+ViewTransition.cpp
+++ b/Source/WebCore/css/parser/CSSPropertyParserConsumer+ViewTransition.cpp
@@ -48,8 +48,8 @@ RefPtr<CSSValue> consumeViewTransitionTypes(CSSParserTokenRange& range, CSS::Pro
     do {
         auto type = consumeUnresolvedCustomIdentExcluding(range, state, { CSSValueNone });
         if (!type)
-             return nullptr;
-        if (type->value.startsWith("-ua-"_s))
+            return nullptr;
+        if (type->startsWith("-ua-"_s))
             return nullptr;
         list.append(CSSCustomIdentValue::create(WTF::move(*type)));
     } while (!range.atEnd());
--- a/Source/WebCore/css/typedom/CSSStyleValueFactory.cpp
+++ b/Source/WebCore/css/typedom/CSSStyleValueFactory.cpp
@@ -277,8 +277,10 @@ static ExceptionOr<Ref<CSSStyleValue>> reifyValue(CSS::SpecificKeyword auto cons
     return WebCore::reifyValue(keyword.value);
 }
 
-static ExceptionOr<Ref<CSSStyleValue>> reifyValue(const CSS::CustomIdent& customIdent)
+static ExceptionOr<Ref<CSSStyleValue>> reifyValue(const CSS::CustomIdent& customIdent, const CSSValue& cssValue, AssociatedProperty&& associatedProperty)
 {
+    if (!customIdent.isResolved())
+        return CSSStyleValue::create(Ref(const_cast<CSSValue&>(cssValue)), WTF::move(associatedProperty));
     return upcast<CSSStyleValue>(CSSOMKeywordValue::rectifyKeywordish(CSS::serializationForCSS(CSS::defaultSerializationContext(), customIdent)));
 }
 
@@ -289,7 +291,7 @@ ExceptionOr<Ref<CSSStyleValue>> CSSStyleValueFactory::reifyValue(Document& docum
     else if (auto* keywordValue = dynamicDowncast<CSSKeywordValue>(cssValue))
         return WebCore::reifyValue(keywordValue->keyword());
     else if (auto* customIdentValue = dynamicDowncast<CSSCustomIdentValue>(cssValue))
-        return WebCore::reifyValue(customIdentValue->customIdent());
+        return WebCore::reifyValue(customIdentValue->customIdent(), cssValue, WTF::move(associatedProperty));
     else if (auto* imageValue = dynamicDowncast<CSSImageValue>(cssValue))
         return Ref<CSSStyleValue> { CSSStyleImageValue::create(const_cast<CSSImageValue&>(*imageValue), document) };
     else if (auto* referenceValue = dynamicDowncast<CSSSubstitutionValue>(cssValue))
@@ -383,7 +385,7 @@ ExceptionOr<Ref<CSSStyleValue>> CSSStyleValueFactory::reifyValue(Document& docum
                 return WebCore::reifyValue(keyword);
             },
             [&](const CSS::CustomIdent& customIdent) -> ExceptionOr<Ref<CSSStyleValue>> {
-                return WebCore::reifyValue(customIdent);
+                return WebCore::reifyValue(customIdent, cssValue, WTF::move(associatedProperty));
             },
             [&](const auto&) -> ExceptionOr<Ref<CSSStyleValue>> {
                 return CSSStyleValue::create(Ref(const_cast<CSSValue&>(cssValue)), WTF::move(associatedProperty));
--- a/Source/WebCore/css/values/primitives/CSSCustomIdent.cpp
+++ b/Source/WebCore/css/values/primitives/CSSCustomIdent.cpp
@@ -29,6 +29,10 @@
 
 #include "CSSCustomIdentValue.h"
 #include "CSSMarkup.h"
+#include "CSSPrimitiveNumericTypes+CSSValueVisitation.h"
+#include "CSSPrimitiveNumericTypes+ComputedStyleDependencies.h"
+#include "CSSPrimitiveNumericTypes+Serialization.h"
+#include "CSSSerializationContext.h"
 #include "DeprecatedCSSOMPrimitiveValue.h"
 #include <wtf/Hasher.h>
 #include <wtf/text/TextStream.h>
@@ -36,9 +40,78 @@
 namespace WebCore {
 namespace CSS {
 
-void Serialize<CustomIdent>::operator()(StringBuilder& builder, const SerializationContext&, const CustomIdent& value)
+bool CustomIdent::isNull() const
+{
+    auto* resolved = std::get_if<AtomString>(&value);
+    return resolved && resolved->isNull();
+}
+
+bool CustomIdent::startsWith(StringView prefix) const
+{
+    return WTF::switchOn(value,
+        [&](const AtomString& resolved) {
+            return resolved.startsWith(prefix);
+        },
+        [&](const IdentFunction& function) {
+            // Build the identifier with every <integer> argument treated as "0", then test the prefix.
+            // https://github.com/w3c/csswg-drafts/issues/12206#issuecomment-3998743769
+            StringBuilder builder;
+            for (auto& argument : function.parameters) {
+                WTF::switchOn(argument,
+                    [&](const IdentFunctionIdent& ident) { builder.append(ident.value); },
+                    [&](const String& string) { builder.append(string.value); },
+                    [&](const Integer<>&) { builder.append('0'); }
+                );
+            }
+            return StringView { builder }.startsWith(prefix);
+        }
+    );
+}
+
+// MARK: - Equality
+
+SUPPRESS_NODELETE bool CustomIdent::operator==(const CustomIdent& other) const
+{
+    return value == other.value;
+}
+
+void Serialize<IdentFunctionIdent>::operator()(StringBuilder& builder, const SerializationContext&, const IdentFunctionIdent& value)
 {
     WebCore::serializeIdentifier(builder, value.value);
+}
+
+void Serialize<CustomIdent>::operator()(StringBuilder& builder, const SerializationContext& context, const CustomIdent& value)
+{
+    WTF::switchOn(value.value,
+        [&](const AtomString& resolved) {
+            WebCore::serializeIdentifier(builder, resolved);
+        },
+        [&](const IdentFunction& function) {
+            serializationForCSS(builder, context, function);
+        }
+    );
+}
+
+void ComputedStyleDependenciesCollector<CustomIdent>::operator()(ComputedStyleDependencies& dependencies, const CustomIdent& value)
+{
+    WTF::switchOn(value.value,
+        [&](const AtomString&) { },
+        [&](const IdentFunction& function) {
+            collectComputedStyleDependencies(dependencies, function);
+        }
+    );
+}
+
+IterationStatus CSSValueChildrenVisitor<CustomIdent>::operator()(NOESCAPE const Function<IterationStatus(CSSValue&)>& func, const CustomIdent& value)
+{
+    return WTF::switchOn(value.value,
+        [&](const AtomString&) {
+            return IterationStatus::Continue;
+        },
+        [&](const IdentFunction& function) {
+            return visitCSSValueChildren(func, function);
+        }
+    );
 }
 
 Ref<CSSValue> CSSValueCreation<CustomIdent>::operator()(CSSValuePool&, const CustomIdent& value)
@@ -55,14 +128,32 @@ Ref<DeprecatedCSSOMValue> DeprecatedCSSOMValueCreation<CustomIdent>::operator()(
 
 WTF::TextStream& operator<<(WTF::TextStream& ts, const CustomIdent& value)
 {
-    return ts << value.value;
+    WTF::switchOn(value.value,
+        [&](const AtomString& resolved) {
+            ts << resolved;
+        },
+        [&](const IdentFunction&) {
+            ts << serializationForCSS(defaultSerializationContext(), value);
+        }
+    );
+    return ts;
 }
 
 // MARK: - Hashing
 
-void add(Hasher& hasher, const CustomIdent& value)
+SUPPRESS_NODELETE void add(Hasher& hasher, const CustomIdent& value)
 {
-    add(hasher, value.value);
+    add(hasher, value.value.index());
+
+    WTF::switchOn(value.value,
+        [&](const AtomString& resolved) {
+            add(hasher, resolved);
+        },
+        [&](const IdentFunction& function) {
+            // Hash the serialization; an <integer> argument may be an unevaluated calc with no raw value.
+            add(hasher, serializationForCSS(defaultSerializationContext(), function));
+        }
+    );
 }
 
 } // namespace CSS
--- a/Source/WebCore/css/values/primitives/CSSCustomIdent.h
+++ b/Source/WebCore/css/values/primitives/CSSCustomIdent.h
@@ -26,28 +26,49 @@
 
 #pragma once
 
+#include <WebCore/CSSPrimitiveNumericTypes.h>
+#include <WebCore/CSSString.h>
 #include <WebCore/CSSValueTypes.h>
 #include <wtf/text/AtomString.h>
 
 namespace WebCore {
 namespace CSS {
 
-// https://drafts.csswg.org/css-values-4/#identifier-value
-struct CustomIdent {
+// Non-keyword <ident> argument to the ident() function, contributing its codepoints to the resulting identifier.
+struct IdentFunctionIdent {
     AtomString value;
 
-    std::optional<AtomString> tryResolved() const
-    {
-        // FIXME: When ident() function is added, only successfully resolve when there are no <integer> components.
-        return value;
-    }
+    bool operator==(const IdentFunctionIdent&) const = default;
+};
 
-    bool operator==(const CustomIdent&) const = default;
+template<> struct Serialize<IdentFunctionIdent> { void operator()(StringBuilder&, const SerializationContext&, const IdentFunctionIdent&); };
+template<> struct ComputedStyleDependenciesCollector<IdentFunctionIdent> { constexpr void operator()(ComputedStyleDependencies&, const IdentFunctionIdent&) { } };
+template<> struct CSSValueChildrenVisitor<IdentFunctionIdent> { constexpr IterationStatus operator()(NOESCAPE const Function<IterationStatus(CSSValue&)>&, const IdentFunctionIdent&) { return IterationStatus::Continue; } };
+
+// <ident-arg> = <string> | <integer> | <ident>
+using IdentFunctionArg = Variant<IdentFunctionIdent, String, Integer<>>;
+
+// <ident()> = ident( <ident-arg>+ )
+// https://drafts.csswg.org/css-values-5/#ident
+using IdentFunction = FunctionNotation<CSSValueIdent, SpaceSeparatedVector<IdentFunctionArg>>;
+
+// https://drafts.csswg.org/css-values-4/#identifier-value
+struct CustomIdent {
+    Variant<AtomString, IdentFunction> value;
+
+    bool isResolved() const { return WTF::holdsAlternative<AtomString>(value); }
+    bool isNull() const;
+
+    // Prefix test for parse-time validity checks, treating every <integer> component as "0".
+    // See https://github.com/w3c/csswg-drafts/issues/12206#issuecomment-3998743769.
+    bool startsWith(StringView prefix) const;
+
+    bool NODELETE operator==(const CustomIdent&) const;
 };
 
 template<> struct Serialize<CustomIdent> { void operator()(StringBuilder&, const SerializationContext&, const CustomIdent&); };
-template<> struct ComputedStyleDependenciesCollector<CustomIdent> { constexpr void operator()(ComputedStyleDependencies&, const CustomIdent&) { } };
-template<> struct CSSValueChildrenVisitor<CustomIdent> { constexpr IterationStatus operator()(NOESCAPE const Function<IterationStatus(CSSValue&)>&, const CustomIdent&) { return IterationStatus::Continue; } };
+template<> struct ComputedStyleDependenciesCollector<CustomIdent> { void operator()(ComputedStyleDependencies&, const CustomIdent&); };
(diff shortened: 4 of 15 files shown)


Source/WebCore/css/parser/CSSPropertyParserConsumer+ViewTransition.cpp
```diff
@@ -48,8 +48,10 @@ RefPtr<CSSValue> consumeViewTransitionTypes(CSSParserTokenRange& range, CSS::Pro
     do {
         auto type = consumeUnresolvedCustomIdentExcluding(range, state, { CSSValueNone });
         if (!type)
-             return nullptr;
-        if (type->value.startsWith("-ua-"_s))
+            return nullptr;
+        // The types descriptor has no style building context, so only eagerly resolvable idents apply.
+        auto resolved = type->tryResolved();
+        if (!resolved || resolved->startsWith("-ua-"_s))
```
weinig: Instead of using tryResolved() here, I think what we want is to add a CSS::CustomIdent::startsWith() function that uses the logic described in https://github.com/w3c/csswg-drafts/issues/12206#issuecomment-3998743769 to perform that check. 

The algorithm is "pretend all `<integer>` arguments are 0".

This will be needed for the `consume{foo}IdentExcluding` functions as well.

Source/WebCore/css/typedom/CSSStyleValueFactory.cpp
```diff
@@ -288,9 +288,11 @@ ExceptionOr<Ref<CSSStyleValue>> CSSStyleValueFactory::reifyValue(Document& docum
         return WebCore::reifyValue(*primitiveValue);
     else if (auto* keywordValue = dynamicDowncast<CSSKeywordValue>(cssValue))
         return WebCore::reifyValue(keywordValue->keyword());
-    else if (auto* customIdentValue = dynamicDowncast<CSSCustomIdentValue>(cssValue))
-        return WebCore::reifyValue(customIdentValue->customIdent());
-    else if (auto* imageValue = dynamicDowncast<CSSImageValue>(cssValue))
+    else if (auto* customIdentValue = dynamicDowncast<CSSCustomIdentValue>(cssValue)) {
+        if (customIdentValue->customIdent().isResolved())
+            return WebCore::reifyValue(customIdentValue->customIdent());
```
weinig: Can the logic move to reifyValue(const CSS::CustomIdent& customIdent) above so it can be shared with the code below that repeats it?

Source/WebCore/css/values/primitives/CSSCustomIdent.h
```diff
@@ -26,28 +26,57 @@
...
     bool operator==(const CustomIdent&) const = default;
 };
 
 template<> struct Serialize<CustomIdent> { void operator()(StringBuilder&, const SerializationContext&, const CustomIdent&); };
-template<> struct ComputedStyleDependenciesCollector<CustomIdent> { constexpr void operator()(ComputedStyleDependencies&, const CustomIdent&) { } };
-template<> struct CSSValueChildrenVisitor<CustomIdent> { constexpr IterationStatus operator()(NOESCAPE const Function<IterationStatus(CSSValue&)>&, const CustomIdent&) { return IterationStatus::Continue; } };
+template<> struct ComputedStyleDependenciesCollector<CustomIdent> {
+    void operator()(ComputedStyleDependencies&, const CustomIdent&);
+};
+template<> struct CSSValueChildrenVisitor<CustomIdent> {
```
weinig: Let's keep these on a single line for consistency.

Source/WebCore/css/parser/CSSPropertyParserConsumer+ViewTransition.cpp
sakupi01: Thanks for the review, @weinig!

Regarding the parse-time validation, I should apply a check logic using the "mock-evaluation algorithm".

Related to this, I was looking into https://github.com/w3c/csswg-drafts/issues/11663 , where they discussed whether `ident()` arguments should be more strictly validated at parse time for other obviously invalid cases. ( Though, I think `ident(custom)` example in https://github.com/w3c/csswg-drafts/issues/11663 is going to be considered as IACVT per https://github.com/w3c/csswg-drafts/issues/12206#issuecomment-3998743769 .)

Do you think we should also chat with them about these eager validation checks, or should we take the issue 11663 as a future parser improvements?

sakupi01: @weinig 

Also, https://github.com/w3c/csswg-drafts/issues/12206 seems to be needing spec edits and WPT addition.
Would it be better to create a spec PR? For WPT, I have to add it in later PR, though.

Source/WebCore/css/values/primitives/CSSCustomIdent.h
sakupi01: There're couple of style error, but I kept the lines single as single line here follows the convention.

```
Source/WebCore/css/values/primitives/CSSCustomIdent.h:44: error: [whitespace/newline] More than one command on the same line
Source/WebCore/css/values/primitives/CSSCustomIdent.h:46: error: [whitespace/newline] More than one command on the same line
Source/WebCore/css/values/primitives/CSSCustomIdent.h:77: error: [whitespace/newline] More than one command on the same line
Source/WebCore/css/values/primitives/CSSCustomIdent.h:78: error: [whitespace/newline] More than one command on the same line
```

weinig: Yeah, the style bot is often wrong.

Source/WebCore/css/parser/CSSPropertyParserConsumer+ViewTransition.cpp
weinig: > Do you think we should also chat with them about these eager validation checks, or should we take the issue 11663 as a future parser improvements?

I left a comment in the followup bug, https://github.com/w3c/csswg-drafts/issues/12668.

> Also, https://github.com/w3c/csswg-drafts/issues/12206 seems to be needing spec edits and WPT addition.
Would it be better to create a spec PR? For WPT, I have to add it in later PR, though.

I think in this case, it's ok to use the working group resolution as an indicator of what it is supposed to be.

sakupi01: @weinig 

Do you think I should wait for the comment on this? I think this opinion makes sense. 

https://github.com/w3c/csswg-drafts/issues/12206#issuecomment-4750414359

nt1m: I'm happy with the current resolution I think, -1 would be turned into 0 and it would not be considered as a dashed ident. We could always revisit if you (or someone else) find the behavior unintuitive as a web developer.

sakupi01: I commented my take on the issue ( sorry if my understanding is not right ).

sakupi01: @nt1m 

Thank you for the follow up! I do agree that for `ident("-" -1)`, the current "mock zero" happens to yield `"-0"` .

However, for example, an invalid syntax like `view-transition-type: ident("-ua" -2 "-name");` evaluates to `"-ua0-name"` under the mock-zero algorithm. Because `"-ua0-name"` is a valid identifier that doesn't start with the restricted `"-ua-"` prefix, it mistakenly bypasses the parse-time validation check?

Even though, I think this could be a further concern, and we would be fine with mock-zero for now?

sakupi01:
@weinig cc: @nt1m 

I've fixed the Safer CPP error and would appreciate a re-review when you have a chance.
(Regarding the style error, I left it as-is based on this discussion:　https://github.com/WebKit/WebKit/pull/67336#discussion_r3430465462 . Also, I kept using  "pretend all <integer> arguments are 0" algorithm reflecting current working group resolution. )

Source/WebCore/css/values/primitives/CSSCustomIdent.cpp
```diff
@@ -55,14 +152,44 @@ Ref<DeprecatedCSSOMValue> DeprecatedCSSOMValueCreation<CustomIdent>::operator()(
...
+        [&](const IdentFunction&) {
+            ts << serializationForCSS(defaultSerializationContext(), value);
+        }
+    );
+    return ts;
 }
 
 // MARK: - Hashing
 
+// Dispatched with std::get_if rather than WTF::switchOn: switchOn builds a by-value
```
cdumez: I think we should prioritize readable code. I'd keep using `WTF::switchOn()`, omit this comment and mark the function as `SUPRESS_NODELETE`.

Source/WebCore/css/CSSCanvasValue.h
```diff
@@ -41,7 +41,8 @@ class CSSCanvasValue final : public CSSValue {
     ~CSSCanvasValue();
 
     String customCSSText(const CSS::SerializationContext&) const;
-    bool NODELETE equals(const CSSCanvasValue&) const;
+    // Not NODELETE: comparing CSS::CustomIdent can descend into a ref-counted calc() tree.
```
cdumez: Please remove all these comments, they are not necessary.

I also don't quite understand the comment. Could you explain (on GitHub, not in a comment) why `equals()` can end up destroying objects now when it couldn't before? It is a little surprising for a `const` function and takes a `const` parameter. Odds are we just need to mark a function somewhere as `SUPPRESS_NODELETE` but I'm not sure which function because I wasn't able to trace from `CustomIdent`'s `operator==()` to a ref-counted `calc()` yet.

sakupi01: @cdumez 

The reason I commented out `NODELETE` in 2d97785 was to address a Safer CPP error triggered in https://github.com/WebKit/WebKit/pull/67336#issuecomment-4731106210 after extending `CustomIdent::value` to `Variant<AtomString, IdentFunction>` in this PR.

Since `IdentFunction` now carrys `CSS::Integer<>` arguments, comparing `CustomIdent` instances involves traversing the internal ref-counted `calc()` tree. The static analyzer caught the destruction possibility during this process.

In reality, however, both objects being compared co-own the reference to the `calc()` tree, meaning no objects will actually be destroyed during this const operation. So, the Safer CPP error would be a false positive, I guess.

Since these `Ref`s are generated inside inline functions like `PrimitiveData::operator==`, annotating deeper functions wouldn't prevent the checker from flagging the caller. Therefore, restoring `equals()`'s `NODELETE`, and moving `CustomIdent::operator==` out-of-line and marking it as `SUPPRESS_NODELETE` would be a good boundary to stop the analyzer from descending further, keeping the higher-level `equals()` implementations clean. How does that sound for you?

cdumez: I think I would mark `bool UnevaluatedCalcBase::operator==(const UnevaluatedCalcBase& other) const` as NODELETE and SUPPRESS_NODELETE. My understanding is that it would resolve the issue closest to the root.

Source/WebCore/css/values/primitives/CSSCustomIdent.cpp
```diff
@@ -29,18 +29,123 @@
...
+                    [&](const IdentFunctionIdent& ident) {
+                        builder.append(ident.value);
+                        return true;
+                    },
+                    [&](const String& string) {
+                        builder.append(string.value);
+                        return true;
+                    },
+                    [&](const Integer<>&) {
+                        return false;
```
nt1m: Should there be a FIXME here to make this work? If not, can we add a comment explaining why we don't do anything here?

Source/WebCore/css/values/primitives/CSSCustomIdent.cpp
```diff
@@ -29,18 +29,123 @@
...
+                    }
+                );
+                if (!resolvedArgument)
+                    return std::nullopt;
+            }
+            if (builder.isEmpty())
+                return std::nullopt;
```
nt1m: What testcases would fall into this check? Would returning an empty string be more correct than failing in some cases?

nt1m (approved):
Would be good if @cdumez and @weinig could take a look as well, but this looks good to me!

Source/WebCore/css/values/primitives/CSSCustomIdent.cpp
sakupi01: Empty string would fall into here (e.g. `ident("" "")`).
I think per spec, an identifier token must be a non-empty sequence of valid identifier code points. 
https://drafts.csswg.org/css-syntax-3/#check-if-three-code-points-would-start-an-ident-sequence

An empty string cannot trigger the tokenization of an identifier, so `ident("")` or `ident("" "")` resulting in a zero-length string fails to meet the grammar of a `<custom-ident>`. Therefore, I think returning `std::nullopt` is correct per spec.

Source/WebCore/css/values/primitives/CSSCustomIdent.cpp
sakupi01: A comment added, thank you!
https://github.com/WebKit/WebKit/commit/6a48a59b835541d68abf4607a27ba8fcf3a5639b#diff-6732bbda82a3cbc411b023129619f5bd261c71fe8cc8ae771c98a562d1ca9860

The same description was in header file and kept, but commenting in here is more precise.

Source/WebCore/css/CSSCanvasValue.h
sakupi01: Thank you! I have added `NODELETE` & `SUPPRESS_NODELETE` for `UnevaluatedCalcBase` and having a look if Safer CPP analyzer is happy with this.
https://github.com/WebKit/WebKit/pull/67336/changes/6a48a59b835541d68abf4607a27ba8fcf3a5639b#diff-ff14bd4c174ef310601e10f11d01d2a7ed11ee6d47592668ee623f36c8f1623f

sakupi01: @cdumez 

I tried `UnevaluatedCalcBase::operator==` with `NODELETE` + `SUPPRESS_NODELETE`, but Safer CPP still flags all three `equals()`. https://github.com/WebKit/WebKit/pull/67336#issuecomment-4809266966

So I reverted `UnevaluatedCalcBase::operator==` -level NODELETE plan and tried `CustomIdent::operator==` with `NODELETE` + `SUPPRESS_NODELETE` again.  I hope fed34fa is fine.

Source/WebCore/css/values/primitives/CSSCustomIdent.cpp
```diff
@@ -29,18 +29,125 @@
...
 namespace CSS {
 
-void Serialize<CustomIdent>::operator()(StringBuilder& builder, const SerializationContext&, const CustomIdent& value)
+bool CustomIdent::isNull() const
+{
+    auto* resolved = std::get_if<AtomString>(&value);
+    return resolved && resolved->isNull();
+}
+
+std::optional<AtomString> CustomIdent::tryResolved() const
```
weinig: My feeling is that all places where tryResolved() is called are probably wrong and either mock-evaluation or real resolving should be used instead. Why do we still have this?

Source/WebCore/css/values/primitives/CSSCustomIdent.cpp
nt1m: Ah makes sense, you indeed can't have an empty identifier!

Source/WebCore/css/values/primitives/CSSCustomIdent.cpp
sakupi01: @weinig Sorry for the late response.

I considered carefully about this and I think you're right. 

- ShorthandSerializer: because the check is about the serialized text, which for `ident()` is always the function form and can never collide with a keyword.
- The descriptor converters: because per w3c/csswg-drafts#12219 ("ident() is only valid within an element context"), an ident() in a descriptor should be rejected by the parser, not resolved later (FIXME added in `consumeUnresolvedCustomIdent()`).
- The deprecated CSSOM string accessors: they return the plain identifier, or fall back to the function-form serialization.

As a result, there's no place uses tryResolved(), so I could delete it. Thank you!

PR #67339, opened 2026-06-17
[Site Isolation] Aggregate cross-process resources and loaderId in Page.getResourceTree

stwrt:
#### df1a3bdc6d50329c1d9bb29b4a3d3480853ae547
<pre>
[Site Isolation] Aggregate cross-process resources and loaderId in Page.getResourceTree
<a href="https://bugs.webkit.org/show_bug.cgi?id=316667">https://bugs.webkit.org/show_bug.cgi?id=316667</a>
<a href="https://rdar.apple.com/179117459">rdar://179117459</a>

Reviewed by BJ Burg.

Under Site Isolation, Page.getResourceTree returned the cross-origin frame tree with
correct structure and url/origin (prior changes) but every frame had an empty loaderId
and no subresources -- it was a frame tree, not a resource tree. The subresources live in
each frame&apos;s hosting WebContent process (CachedResourceLoader), which the UIProcess
ProxyingPageAgent cannot read directly.

Gather each frame&apos;s resources from its hosting process as typed IPC data and build the
aggregated tree in the proxy:

- Make Page.getResourceTree asynchronous (the proxy must round-trip to the WebContent
  processes before it can answer). Both implementations adopt the callback form: the
  in-process InspectorPageAgent answers synchronously; the proxy answers once every
  process has replied.

- Carry each frame&apos;s resources as typed IPC rather than a JSON string. Add
  Inspector::FrameResource / Inspector::FrameResourceData (in a new
  Shared/InspectorPageTypes.serialization.in), plain serializable mirrors of
  Page.FrameResource, so secure decoding is generated rather than hand-rolled. The
  WebContent handler (WebInspectorBackend::GetFrameResourceData) returns this typed data
  and the Page.FrameResource protocol objects are constructed in the UIProcess. The
  CachedResource -&gt; data extraction is shared by both paths via
  ResourceUtilities::buildResourceDataForFrame, and buildResourceObject builds one protocol
  object from the data; buildResourceObjectsForFrame (used by the in-process agent) is
  rebuilt on top of these so every path produces identical output.

- Carry loaderId as a WebCore::ScriptExecutionContextIdentifier rather than a
  pre-formatted string. PageAgentProxy sends the committed document&apos;s identifier in the
  live Page.frameNavigated event; the ProxyingPageAgent caches it and derives the
  deterministic, hosting-process-qualified protocol loaderId
  (IdentifierRegistry::protocolLoaderId) only at the protocol boundary. buildFrameTree
  prefers the cached identifier and falls back to the one gathered from the hosting process
  for frames that committed before the inspector connected (e.g. the main frame). Full
  correlation with live network-event loaderIds remains webkit.org/b/308895.

- ProxyingPageAgent::getResourceTree walks the authoritative WebFrameProxy tree to group
  frame ids by their hosting WebContent process, then asks each process
  (GetFrameResourceData) only for the frames it hosts. Walking the frame tree -- rather
  than WebPageProxy::forEachWebContentProcess() -- guarantees coverage matches exactly the
  processes hosting frames, and scoping each request to a process&apos;s own frames avoids
  fetching every frame&apos;s resources from every process. A CallbackAggregator builds the
  stitched tree once every reply has landed.

* Source/JavaScriptCore/inspector/protocol/Page.json:
Mark getResourceTree async.

* Source/WebCore/inspector/InspectorResourceUtilities.h:
* Source/WebCore/inspector/InspectorResourceUtilities.cpp:
(Inspector::ResourceUtilities::buildResourceDataForFrame):
(Inspector::ResourceUtilities::buildResourceObject):
(Inspector::ResourceUtilities::buildResourceObjectsForFrame):
Add Inspector::FrameResource / FrameResourceData, the shared CachedResource -&gt; data
extraction (buildResourceDataForFrame), and the protocol builder (buildResourceObject).
Rebuild buildResourceObjectsForFrame on top of them so the in-process and cross-process
paths produce identical output.

(WebCore::InspectorPageAgent::getResourceTree):
(WebCore::InspectorPageAgent::buildObjectForFrameTree):
Adopt the async callback signature (answer synchronously); use buildResourceObjectsForFrame.

* Source/WebKit/Shared/InspectorPageTypes.serialization.in: Added.
* Source/WebKit/DerivedSources.make:
* Source/WebKit/DerivedSources-input.xcfilelist:
Generate secure coders for Inspector::FrameResource / FrameResourceData.

Map the new types to their header for message generation.

* Source/WebKit/WebProcess/Inspector/WebInspectorBackend.h:
* Source/WebKit/WebProcess/Inspector/WebInspectorBackend.cpp:
(WebKit::WebInspectorBackend::getFrameResourceData):
* Source/WebKit/WebProcess/Inspector/WebInspectorBackend.messages.in:
Return each requested local frame&apos;s { loaderId, resources } as typed data for the proxy
to aggregate.

* Source/WebKit/WebProcess/Inspector/PageAgentProxy.cpp:
(WebKit::PageAgentProxy::frameNavigated):
Send the committed document&apos;s ScriptExecutionContextIdentifier as the loaderId.

* Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.h:
* Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.cpp:
(Inspector::ProxyingPageAgent::frameNavigated):
(Inspector::ProxyingPageAgent::buildFrameTree):
(Inspector::ProxyingPageAgent::getResourceTree):
Cache loaderId as a ScriptExecutionContextIdentifier; walk the frame tree to gather typed
resources per hosting process and build the aggregated tree asynchronously.

* Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.messages.in:
Carry loaderId as a ScriptExecutionContextIdentifier in FrameNavigated.

* LayoutTests/http/tests/site-isolation/inspector/page/resource-tree-resources-cross-origin-iframe.html: Added.
* LayoutTests/http/tests/site-isolation/inspector/page/resource-tree-resources-cross-origin-iframe-expected.txt: Added.
* LayoutTests/http/tests/site-isolation/inspector/page/resources/resource-tree-frame-with-subresource.html: Added.
* LayoutTests/http/tests/site-isolation/inspector/page/resources/resource-tree-subframe-style.css: Added.
Assert getResourceTree reports loaderId for the main and cross-origin child frames and
aggregates the child&apos;s cached stylesheet subresource cross-process.

Canonical link: <a href="https://commits.webkit.org/315428@main">https://commits.webkit.org/315428@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/12cef39f67e35e4e2550aab7bdfbefc100319a22

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/168955 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/42526 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/36116 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/178308 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/126666 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/6dbfaae4-2127-479d-9420-bb8daf06133c) 
| [loading 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/170821 "Build is in progress. Recent messages:OS: Tahoe (26.5), Xcode: 26.5; Skipping applying patch since patch_id isn't provided; Checked out pull request; Passed bindings tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/42900 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/42501 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/131565 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/92560 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/b0d8a3ef-b63b-4073-bd5d-58a93c7aceba) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/171914 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/34021 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/151835 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/112068 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/9e3e4f99-fedf-4da3-8a91-4b42e965e839/e0fba71f-9acb-458b-bc48-32f9a6611305) 
| [✅ 🧪 webkitpy](https://ews-build.webkit.org/#/builders/6/builds/168276 "Passed tests") | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/33008 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/31717 "Passed tests") | [✅ 🛠 gtk3-libwebrtc](https://ews-build.webkit.org/#/builders/173/builds/27011 "Built successfully") | | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/253 "Passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/142999 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/31235 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/180910 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/30210 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/33621 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/35886 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/140013 "Passed tests") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/42533 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/151305 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/139856 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/37656 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/43222 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/28207 "Passed tests") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/107048 "Built successfully") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/35143 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/114987 "Built successfully") | [✅ 🛠 jsc-armv7](https://ews-build.webkit.org/#/builders/35/builds/201634 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/41731 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/6295 "Passed tests") | [✅ 🧪 jsc-armv7-tests](https://ews-build.webkit.org/#/builders/25/builds/54122 "Passed tests") | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/41125 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/41380 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/41268 "Built successfully") | | | |


Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.cpp
```diff
@@ -90,15 +93,16 @@ static String protocolFrameIdForFrameID(FrameIdentifier frameID)
...
-    m_cachedFrameDocumentInfo.set(frameID, CachedFrameDocumentInfo { url, mimeType, securityOrigin });
+    // Cache the committing frame's real document info so getResourceTree()/buildFrameTree()
+    // can report it for cross-origin children, whose commit the inspectedPage's WebFrameProxy
+    // never observes (its url/origin stay stale). See webkit.org/b/308896.
+    m_cachedFrameDocumentInfo.set(frameID, CachedFrameDocumentInfo { url, mimeType, securityOrigin, loaderId });
 
     auto frameObject = Protocol::Page::Frame::create()
         .setId(protocolFrameIdForFrameID(frameID))
-        .setLoaderId(String()) // FIXME: <https://webkit.org/b/308895> get loaderId from document identifier
+        .setLoaderId(loaderId)
```
burg: I'm a little concerned we are passing around loader ID (ScriptExecutionContextIdentifier) as a string prior to this point. Can we make this have its own type?

Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.cpp
```diff
@@ -268,17 +272,39 @@ Ref<Protocol::Page::FrameResourceTree> ProxyingPageAgent::buildFrameTree(const W
...
         url = it->value.url;
         securityOrigin = it->value.securityOrigin;
         if (!it->value.mimeType.isEmpty())
             mimeType = it->value.mimeType;
+        loaderId = it->value.loaderId;
     }
     String name = frame.frameName();
 
+    // Parse the per-frame data gathered from this frame's hosting WebContent process: a JSON
+    // object { loaderId, resources }. Resources are the frame's cached subresources (a JSON array
```
burg: It'd be better if we could use typed IPC internally rather than JSON string. Also, it would handle all the secure decoding business that we'd rather not have to reinvent. So it might be easier. But, this means the protocol JSON construction would have to happen in UIProcess rather than passing along the resources array.

Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.cpp
```diff
@@ -291,30 +317,53 @@ Ref<Protocol::Page::FrameResourceTree> ProxyingPageAgent::buildFrameTree(const W
...
+        RefPtr mainFrame = inspectedPage->mainFrame();
+        if (!mainFrame) {
+            callback->sendFailure("Missing main frame"_s);
+            return;
+        }
+        callback->sendSuccess(protectedThis->buildFrameTree(*mainFrame, nullptr, *resourcesByFrame));
+    });
 
-    return buildFrameTree(*mainFrame, nullptr);
+    inspectedPage->forEachWebContentProcess([&](auto& webProcess, auto pageID) {
```
burg: Let's be careful that this will actually iterate all of the processes involved in the frame tree. From our Page/Network standup, we had to move away from `forEachWebContentProcess`.

Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.cpp
```diff
@@ -291,30 +317,53 @@ Ref<Protocol::Page::FrameResourceTree> ProxyingPageAgent::buildFrameTree(const W
...
+        webProcess.sendWithAsyncReply(Messages::WebInspectorBackend::GetFrameResourceData { }, [resourcesByFrame, aggregator](Vector<std::pair<FrameIdentifier, String>>&& items) {
```
burg: Is it a good or bad idea to make this a optionally frame-based API? like `Messages::WebInspectorBackend::GetFrameResourceData { frameID }`

It seems like for the single frame case, there is no difference. For the case with many subframes it would fracture the view of the page (read over N updates) so it might not make sense for the initial payload. But surely there's a lot of wasted IPC traffic here because we usually don't need all resources for all frames.

Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.cpp
stwrt: The FrameNavigated event and the GetFrameResourceData reply now carry the WebCore::ScriptExecutionContextIdentifier directly, and ProxyingPageAgent caches it as that type. We no longer pass a pre-formatted string across IPC or hold one in the cache.

Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.cpp
stwrt: Switched to typed IPC.

Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.cpp
stwrt: It now walks the authoritative WebFrameProxy tree and derives the process set from the frames themselves.

Source/WebCore/inspector/agents/InspectorPageAgent.cpp
```diff
@@ -871,31 +873,11 @@ Ref<Inspector::Protocol::Page::FrameResourceTree> InspectorPageAgent::buildObjec
 
     auto* localFrame = downcast<LocalFrame>(frame);
     auto frameObject = buildObjectForFrame(localFrame);
-    auto subresources = JSON::ArrayOf<Inspector::Protocol::Page::FrameResource>::create();
     auto result = Inspector::Protocol::Page::FrameResourceTree::create()
         .setFrame(WTF::move(frameObject))
-        .setResources(subresources.copyRef())
+        .setResources(ResourceUtilities::buildResourceObjectsForFrame(*localFrame))
```
burg, line 893: We don't check if localFrame is null here.

Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.cpp
```diff
@@ -291,30 +310,84 @@ Ref<Protocol::Page::FrameResourceTree> ProxyingPageAgent::buildFrameTree(const W
...
+void ProxyingPageAgent::getResourceTree(Ref<GetResourceTreeCallback>&& callback)
 {
-    if (!m_enabled)
-        return makeUnexpected("Not supported without Site Isolation"_s);
+    if (!m_enabled) {
+        callback->sendFailure("Not supported without Site Isolation"_s);
+        return;
+    }
 
     Ref inspectedPage = m_inspectedPage.get();
```
burg, line 333: We should take this opportunity to opt into 'modern' async codegen. This means using `CommandCallback<T>` / `CommandCallbackOf<T...>` as the argument type. It should also make it easier to join the results with lambdas. Feel free to do this in a followup as it's a bit disruptive to stacked PRs.

Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.cpp
```diff
@@ -291,30 +310,84 @@ Ref<Protocol::Page::FrameResourceTree> ProxyingPageAgent::buildFrameTree(const W
...
+        if (!mainFrame) {
+            callback->sendFailure("Missing main frame"_s);
+            return;
+        }
+        callback->sendSuccess(protectedThis->buildFrameTree(*mainFrame, nullptr, *resourcesByFrame));
+    });
+
+    for (auto& entry : framesByProcess.values()) {
+        Ref process = *entry.process;
+        process->sendWithAsyncReply(Messages::WebInspectorBackend::GetFrameResourceData { WTF::move(entry.frameIDs) }, [resourcesByFrame, aggregator](Vector<std::pair<FrameIdentifier, FrameResourceData>>&& items) {
```
burg, line 386: What's the defined behavior if some or all frameID are invalid? do we get an error or result count back?

burg (approved):
Commit nit: weird double line:
> Under Site Isolation, Page.getResourceTree returned the cross-origin frame tree with
correct structure and url/origin (prior changes) but every frame had an empty loaderId
correct structure and url/origin (prior changes) but every frame had an empty loaderId

Source/WebCore/inspector/agents/InspectorPageAgent.cpp
stwrt: buildObjectForFrameTree now does a checked dynamicDowncast<LocalFrame> and returns a resource-less stub if it's null, instead of dereferencing.

so it degrades gracefully rather than crashing — consistent with how the function already handles the RemoteFrame case.

Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.cpp
stwrt: I'll do it as a follow-up as you suggested — converting the codegen is disruptive to the stacked PRs, and it'll be cleaner as its own change

Source/WebKit/UIProcess/Inspector/Agents/ProxyingPageAgent.cpp
stwrt: No error, and the result count can be less than requested. The WebContent handler resolves each requested id via WebProcess::webFrame(frameID) and skips any that don't resolve to a local frame in that process, so the reply simply omits them getFrameResourceData never fails on a bad id.

PR #67454, opened 2026-06-18
[css-ui] outline-offset computed value should be snapped as a line width

Ahmad-S792:
#### 3adec8e64df0536e4d34c37902714552fbc0057d
<pre>
[css-ui] outline-offset computed value should be snapped as a line width
<a href="https://bugs.webkit.org/show_bug.cgi?id=317427">https://bugs.webkit.org/show_bug.cgi?id=317427</a>
<a href="https://rdar.apple.com/180054053">rdar://180054053</a>

Reviewed by Sam Weinig and Darin Adler.

This patch aligns WebKit with Gecko / Firefox and Blink / Chromium.

The CSSWG resolved that `outline-offset` should be snapped to an integer number of
device pixels the same way `outline-width` and `border-width` are
(w3c/csswg-drafts#12906). WebKit used the raw, unsnapped length, so
getComputedStyle().outlineOffset returned values like &quot;0.1px&quot; instead of &quot;1px&quot;.

Move &quot;snap a length as a border width&quot;
(<a href="https://drafts.csswg.org/css-values-4/#snap-a-length-as-a-border-width)">https://drafts.csswg.org/css-values-4/#snap-a-length-as-a-border-width)</a> out of
`LineWidth` into its own file, exposing the `float` overload directly and adding
signed overloads for lengths that may be negative, which snap the magnitude and
restore the sign: zero is unchanged, a magnitude below 1 device pixel rounds away
from zero, a larger magnitude rounds toward zero.

`outline-offset` now snaps in the same three places `outline-width` does:

 - `CSSValueConversion&lt;OutlineOffset&gt;` snaps the computed value.
 - `Blending&lt;OutlineOffset&gt;` re-snaps blended values, keeping animations on device
   pixels. It also has to supply `canBlend()`, since a full specialization replaces
   the generic `ValueOrKeyword` one, which is what makes `inset` interpolate
   discretely against a length rather than blending as an empty value.
 - `Evaluation&lt;UsedOutlineOffset&gt;` re-snaps after zoom is applied, keeping the used
   value on device pixels at any page zoom. `ComputedStyle::usedOutlineOffset()`
   returns `Style::UsedOutlineOffset` so that evaluating it requires a device scale
   factor, and its callers pass one.

The manual `floorToDevicePixel()` in RenderThemeCocoa is subsumed by that used-value
snapping. It is not quite equivalent: for a magnitude below one device pixel, `floor`
gave 0 while snapping rounds away from zero to 1 device pixel, so an iOS focus ring
at a zoom below 1 now scales by the offset instead of ignoring it. That is the
behavior `outline-width` already has.

`inset` resolves to the negated used `outline-width`, and snapping both after zoom
makes them cancel exactly, which they previously did not at fractional zoom.

* LayoutTests/fast/borders/hidpi-outline-hairline-painting-expected.html: Rebaselined.
This reference was authored against unsnapped offsets, and at the 2x device scale factor
this test runs at, snapping the offset changes the painted outline geometry.
* LayoutTests/fast/css/outline-offset-used-value-zoom-rounding-expected.txt: Added.
* LayoutTests/fast/css/outline-offset-used-value-zoom-rounding.html: Added. Covers the
used value after zoom, which needs `internals` since it is not web-exposed.
* LayoutTests/imported/w3c/web-platform-tests/css/css-borders/outline-offset-rounding.tentative.html:
The test as imported hard-codes expectations that assume 1 device pixel is 1 CSS pixel, so its
sub-pixel sub-tests reported FAIL at any other device scale factor even though the engine was
behaving as specified. Derive the expectations from window.devicePixelRatio instead, and add
sub-tests that cross-check outline-offset against outline-width for the same input, which is a
direct statement of the CSSWG resolution and holds at any device scale factor. Verified all 30
sub-tests pass at 1x, 2x and 3x. To be upstreamed.
* LayoutTests/imported/w3c/web-platform-tests/css/css-borders/outline-offset-rounding.tentative-expected.txt: Progression
* LayoutTests/imported/w3c/web-platform-tests/css/css-ui/animation/outline-offset-interpolation-rounding.tentative-expected.txt: Added.
* LayoutTests/imported/w3c/web-platform-tests/css/css-ui/animation/outline-offset-interpolation-rounding.tentative.html: Added.
Covers snapping of blended values, deriving expectations from window.devicePixelRatio so they hold
at any device scale factor. The `inset` sub-test is gated on CSS.supports(), since Chromium and
Gecko do not implement it yet and drop the keyframe. To be upstreamed.
* Source/WebCore/Sources.txt:
(WebCore::AccessibilityRenderObject::elementPath const):
* Source/WebCore/rendering/OutlinePainter.cpp:
(WebCore::OutlinePainter::paintOutline const):
(WebCore::OutlinePainter::paintOutlineWithLineRects const):
(WebCore::OutlinePainter::paintFocusRing const):
(WebCore::RenderElement::repaintAfterLayoutIfNeeded):
* Source/WebCore/rendering/cocoa/RenderThemeCocoa.mm:
(WebCore::drawFocusRingForPathForVectorBasedControls):
(WebCore::Style::ComputedStyle::usedOutlineOffset const):
(WebCore::Style::ComputedStyle::usedOutlineSize const):
* Source/WebCore/style/values/backgrounds/StyleLineWidth.cpp:
(WebCore::Style::snapLengthAsBorderWidth): Deleted.
(WebCore::Style::LineWidth::snapLengthAsBorderWidth): Deleted.
* Source/WebCore/style/values/backgrounds/StyleLineWidth.h:
* Source/WebCore/style/values/outline/StyleOutlineOffset.cpp:
(WebCore::Style::CSSValueConversion&lt;OutlineOffset&gt;::operator()):
(WebCore::Style::Blending&lt;OutlineOffset&gt;::canBlend):
(WebCore::Style::Blending&lt;OutlineOffset&gt;::blend):
(WebCore::Style::Evaluation&lt;UsedOutlineOffset, float&gt;::operator()):
(WebCore::Style::Evaluation&lt;UsedOutlineOffset, LayoutUnit&gt;::operator()):
* Source/WebCore/style/values/outline/StyleOutlineOffset.h:
* Source/WebCore/style/values/primitives/StyleSnapLengthAsBorderWidth.cpp: Added.
(WebCore::Style::snapLengthAsBorderWidth):
(WebCore::Style::snapSignedLengthAsBorderWidth):
* Source/WebCore/style/values/primitives/StyleSnapLengthAsBorderWidth.h: Added.
(WebCore::Internals::usedOutlineOffset):

Canonical link: <a href="https://commits.webkit.org/319236@main">https://commits.webkit.org/319236@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/0c9090c747f1c8ce38edf3053618b5710ce3ab3c

| [❌ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/180761 "1 style error") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/54706 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/48504 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/190974 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/145084 "Built successfully") | [✅ 🛠 ios-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/ae28d4e3-28f6-4e5d-9efa-beb3be2fc940) 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/182623 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/56004 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/54422 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/141010 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/97716 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | [✅ 🛠 mac-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/10528cbb-a6ea-4635-8ba6-e826ca6b4c99) 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/183716 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/44671 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/161766 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/121462 "Passed tests") | | [✅ 🛠 vision-apple](https://ews-bridge.webkit.apple.com/builds/sw/T-276/92699e7b-e2c4-41dd-8b79-0fc7c00d5b52/b4107158-ab15-4d88-9c1e-41b668d24bf0) 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/43344 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/41622 "Passed tests") | | | 
| | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/153748 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/41295 "Passed tests") | [✅ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/2135 "Built successfully") | | 
| | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/47318 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/45877 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/192909 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/54372 "Built successfully") | [✅ 🧪 mac-wk2-stress](https://ews-build.webkit.org/#/builders/8/builds/161284 "Passed tests") | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/150389 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/40274 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/55060 "Built successfully") | [✅ 🧪 mac-intel-wk2](https://ews-build.webkit.org/#/builders/170/builds/37913 "Passed tests") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/150106 "Passed tests") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/44700 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/127326 "Built successfully") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/122612 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/53577 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/16288 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/52959 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/53196 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/53024 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/Sources.txt
+++ b/Source/WebCore/Sources.txt
@@ -3516,6 +3516,7 @@ style/values/non-standard/StyleWebKitLocale.cpp
 style/values/non-standard/StyleWebKitOverflowScrolling.cpp
 style/values/non-standard/StyleWebKitTextStrokeWidth.cpp
 style/values/non-standard/StyleWebKitTouchCallout.cpp
+style/values/outline/StyleOutlineOffset.cpp @header:RenderStyleGetters
 style/values/overflow/StyleBlockEllipsis.cpp
 style/values/overflow/StyleScrollbarGutter.cpp
 style/values/overflow/StyleOverflowClipMargin.cpp
@@ -3526,6 +3527,7 @@ style/values/primitives/StyleLengthResolution.cpp @header:RenderStyleGetters @co
 style/values/primitives/StylePosition.cpp
 style/values/primitives/StylePrimitiveData.cpp
 style/values/primitives/StyleRatio.cpp
+style/values/primitives/StyleSnapLengthAsBorderWidth.cpp
 style/values/primitives/StyleString.cpp
 style/values/primitives/StyleURL.cpp
 style/values/primitives/StyleUnevaluatedCalculation.cpp
--- a/Source/WebCore/style/values/outline/StyleOutlineOffset.cpp
+++ b/Source/WebCore/style/values/outline/StyleOutlineOffset.cpp
@@ -1,0 +1,90 @@
+/*
+ * Copyright (C) 2026 Apple Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ * 1. Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ * 2. Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * THIS SOFTWARE IS PROVIDED BY APPLE INC. ``AS IS'' AND ANY
+ * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR
+ * PURPOSE ARE DISCLAIMED.  IN NO EVENT SHALL APPLE INC. OR
+ * CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+ * EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+ * PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+ * OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "config.h"
+#include "StyleOutlineOffset.h"
+
+#include "CSSKeywordValue.h"
+#include "Document.h"
+#include "StyleBuilderState.h"
+#include "StyleComputedStyle+GettersInlines.h"
+#include "StyleInterpolationClient.h"
+#include "StyleInterpolationContext.h"
+#include "StylePrimitiveNumericTypes+Blending.h"
+#include "StylePrimitiveNumericTypes+CSSValueConversion.h"
+#include "StylePrimitiveNumericTypes+Evaluation.h"
+#include "StyleSnapLengthAsBorderWidth.h"
+
+namespace WebCore {
+namespace Style {
+
+// MARK: - Conversion
+
+auto CSSValueConversion<OutlineOffset>::operator()(BuilderState& state, const CSSValue& value) -> OutlineOffset
+{
+    if (RefPtr keywordValue = dynamicDowncast<CSSKeywordValue>(value)) {
+        if (keywordValue->valueID() == CSSValueInset)
+            return CSS::Keyword::Inset { };
+    }
+
+    return snapSignedLengthAsBorderWidth(toStyleFromCSSValue<Length<>>(state, value), state.style().deviceScaleFactor());
+}
+
+// MARK: - Blending
+
+auto Blending<OutlineOffset>::canBlend(const OutlineOffset& a, const OutlineOffset& b) -> bool
+{
+    if (a.isInset() || b.isInset())
+        return false;
+    return Style::canBlend(*a.tryLength(), *b.tryLength());
+}
+
+auto Blending<OutlineOffset>::blend(const OutlineOffset& a, const OutlineOffset& b, const Style::ComputedStyle& aStyle, const Style::ComputedStyle& bStyle, const Interpolation::Context& context) -> OutlineOffset
+{
+    if (context.isDiscrete) {
+        ASSERT(!context.progress || context.progress == 1);
+        return context.progress ? b : a;
+    }
+
+    auto blendedValue = Style::blend(*a.tryLength(), *b.tryLength(), aStyle, bStyle, context);
+    if (RefPtr document = context.client.document())
+        return snapSignedLengthAsBorderWidth(blendedValue, document->deviceScaleFactor());
+    return blendedValue;
+}
+
+// MARK: - Evaluation
+
+auto Evaluation<UsedOutlineOffset, float>::operator()(const UsedOutlineOffset& value, ZoomFactor zoom, float deviceScaleFactor) -> float
+{
+    return snapSignedLengthAsBorderWidth(evaluate<float>(value.value, zoom), deviceScaleFactor);
+}
+
+auto Evaluation<UsedOutlineOffset, LayoutUnit>::operator()(const UsedOutlineOffset& value, ZoomFactor zoom, float deviceScaleFactor) -> LayoutUnit
+{
+    // NOTE: Using `evaluate<float>`, not `evaluate<LayoutUnit>`, as snapSignedLengthAsBorderWidth takes a `float`.
+    return LayoutUnit { snapSignedLengthAsBorderWidth(evaluate<float>(value.value, zoom), deviceScaleFactor) };
+}
+
+} // namespace Style
+} // namespace WebCore
--- a/Source/WebCore/style/values/outline/StyleOutlineOffset.h
+++ b/Source/WebCore/style/values/outline/StyleOutlineOffset.h
@@ -24,6 +24,7 @@
 
 #pragma once
 
+#include <WebCore/LayoutUnit.h>
 #include <WebCore/StylePrimitiveNumeric.h>
 #include <WebCore/StyleValueTypes.h>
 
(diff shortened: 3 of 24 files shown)


LayoutTests/platform/ios/imported/w3c/web-platform-tests/css/css-borders/outline-offset-rounding.tentative-expected.txt
```diff
@@ -0,0 +1,22 @@
+
+PASS 0px
+PASS -0px
+FAIL 0.1px assert_equals: expected "1px" but got "0.5px"
```
darinadler: Why does this tentative test have to be written in a way that will fail if the test system has a 2x device scale factor? Who wrote it? Can we write a better one instead? Really don’t want `FAIL` when the browser engine is behaving correctly as designed!

Source/WebCore/style/values/outline/StyleOutlineOffset.h
```diff
@@ -37,8 +37,17 @@ struct OutlineOffset : ValueOrKeyword<Length<>, CSS::Keyword::Inset> {
 
     bool isInset() const { return isKeyword(); }
     std::optional<Length<>> tryLength() const { return tryValue(); }
+
+    // Snaps `outline-offset` to an integer number of device pixels, like a line width. Unlike a
+    // line width, the offset may be negative, so the magnitude is snapped and the sign preserved.
+    // https://drafts.csswg.org/css-values-4/#snap-as-a-line-width
+    static Style::Length<> snapLengthAsBorderWidth(Style::Length<>, float deviceScaleFactor);
```
darinadler: Why does this need to be a static member of `Style::OutlineOffset` and in the header? Since the function is only used inside `StyleOutlineOffset.cpp`, can it just be a local function and not mentioned in the header at all?

Source/WebCore/Sources.txt
```diff
@@ -3459,6 +3459,7 @@ style/values/non-standard/StyleWebKitLocale.cpp
 style/values/non-standard/StyleWebKitOverflowScrolling.cpp
 style/values/non-standard/StyleWebKitTextStrokeWidth.cpp
 style/values/non-standard/StyleWebKitTouchCallout.cpp
+style/values/outline/StyleOutlineOffset.cpp @header:RenderStyleGetters @cost:5
```
darinadler, line 3462: Why the `@header` and `@cost` annotation here?

darinadler (approved):

Source/WebCore/Sources.txt
Ahmad-S792: This was added because of CSSValueConversion calls state.style().deviceScaleFactor(), and that getter is an inline defined in StyleComputedStyleBase+GettersInlines.h. I think adding header as is lead to complain in style checker. I just matched other files what they were doing e.g., style/values/backgrounds/StyleLineWidth.cpp.

Source/WebCore/style/values/outline/StyleOutlineOffset.h
Ahmad-S792: Fixed.

LayoutTests/platform/ios/imported/w3c/web-platform-tests/css/css-borders/outline-offset-rounding.tentative-expected.txt
Ahmad-S792: Fixed test - will upstream later.

Source/WebCore/Sources.txt
darinadler: I still don't understand this, including whether adding `@header` is correct or unnecessary.

darinadler (approved):

Source/WebCore/Sources.txt
Ahmad-S792: @darinadler - I looked more and it seems that @header is needed by `generate-unified-source-bundles.py` script, which bundles headers which are heavy like StyleComputedStyleBase+GettersInlines.h and it is used by Xcode port only (post 315023@main - removed for CMake build). It is to pack the header better rather than per directory .cpp pool and we need it because we use this header for `deviceScaleFactor()` in `StyleComputedStyleBase.h` (as inline) and then defined at `StyleComputedStyleBase+GettersInlines.h`. While I compiled with and without @cost bit and it seems to make no difference, so I dropped it.

Source/WebCore/style/values/outline/StyleOutlineOffset.cpp
```diff
@@ -0,0 +1,63 @@
...
+// line width, the offset may be negative, so the magnitude is snapped and the sign preserved.
+// https://drafts.csswg.org/css-values-4/#snap-as-a-line-width
+static Length<> snapOutlineOffsetAsLineWidth(Length<> length, float deviceScaleFactor)
+{
+    auto value = length.unresolvedValue();
+    if (!value)
+        return length;
+
+    auto sign = value < 0.0f ? -1.0f : 1.0f;
+    return Length<> { sign * LineWidth::snapLengthAsBorderWidth(std::abs(value), deviceScaleFactor).unresolvedValue() };
```
weinig: I would expose the snapLengthAsBorderWidth() that takes a float directly. Maybe even move it to it and its friends to their own StyleSnapLengthAsBorderWidth file.

Source/WebCore/style/values/outline/StyleOutlineOffset.cpp
```diff
@@ -0,0 +1,63 @@
...
+auto CSSValueConversion<OutlineOffset>::operator()(BuilderState& state, const CSSValue& value) -> OutlineOffset
+{
+    if (RefPtr keywordValue = dynamicDowncast<CSSKeywordValue>(value)) {
+        if (keywordValue->valueID() == CSSValueInset)
+            return CSS::Keyword::Inset { };
+    }
+
+    return snapOutlineOffsetAsLineWidth(toStyleFromCSSValue<Length<>>(state, value), state.style().deviceScaleFactor());
+}
+
```
weinig, line 88: We maintain the device pixel snapping for animations by using a custom Blending interface for LineWidth. Why don't we need to do that here?

Source/WebCore/style/values/outline/StyleOutlineOffset.cpp
```diff
@@ -0,0 +1,63 @@
...
+auto CSSValueConversion<OutlineOffset>::operator()(BuilderState& state, const CSSValue& value) -> OutlineOffset
+{
+    if (RefPtr keywordValue = dynamicDowncast<CSSKeywordValue>(value)) {
+        if (keywordValue->valueID() == CSSValueInset)
+            return CSS::Keyword::Inset { };
+    }
+
+    return snapOutlineOffsetAsLineWidth(toStyleFromCSSValue<Length<>>(state, value), state.style().deviceScaleFactor());
+}
```
weinig, line 87: We maintain the device pixel after zoom application by using a custom Evaluation interface that takes the deviceScaleFactor in addition to the zoom for LineWidth. Why don't we need to do that here?

weinig (approved):

Ahmad-S792:
@weinig - I tried to address your feedback and added few test cases, can you have another look? Happy to iterate as needed.

weinig:
> @weinig - I tried to address your feedback and added few test cases, can you have another look? Happy to iterate as needed.

Looks great.

PR #67465, opened 2026-06-18
Support writing ISOBoxes as well as reading/parsing

jernoble:
#### 6446fdc8e0be049f61e48d8beb8556ee8e2e1009
<pre>
Support writing ISOBoxes as well as reading/parsing
<a href="https://rdar.apple.com/180057740">rdar://180057740</a>
<a href="https://bugs.webkit.org/show_bug.cgi?id=317430">https://bugs.webkit.org/show_bug.cgi?id=317430</a>

Reviewed by Jean-Yves Avenard.

ISOBoxes and their subclasses are parsed representations of ISO-BMFF media data. In addition
to reading these boxes from files and memory, we may also want to serialize these boxes to
memory, for features like mp4-file authoring and editing.

Add the ability to write a box (and all its children) to memory. Each box must be able to
calculate its own size, so there is a new pair of methods: `partialSize` and `requiredSize`.
Base ISOBoxes boxes need their subclasses&apos; sizes to be fully calculated, hence the split
between the two functions. Also add a `serialize` and virtual `write` method, and a convenience
template `checkedReadSequence` and `checkedWriteSequence` to support reading and writing
data arrays.

Test: Tools/TestWebKitAPI/Tests/WebCore/ISOBox.cpp

* Source/WebCore/platform/graphics/iso/ISOBox.cpp:
(WebCore::ISOBox::ISOBox):
(WebCore::ISOBox::peekBox):
(WebCore::ISOBox::read):
(WebCore::ISOBox::updateSize):
(WebCore::ISOBox::partialSize const):
(WebCore::ISOBox::requiredSize const):
(WebCore::ISOBox::parse):
(WebCore::ISOBox::write const):
(WebCore::ISOBox::serialize const):
(WebCore::ISOFullBox::ISOFullBox):
(WebCore::ISOFullBox::parse):
(WebCore::ISOFullBox::write const):
(WebCore::ISOFullBox::parseVersionAndFlags):
* Source/WebCore/platform/graphics/iso/ISOBox.h:
(WebCore::ISOBox::checkedRead):
(WebCore::ISOBox::checkedReadSequence):
(WebCore::ISOBox::checkedWrite):
(WebCore::ISOBox::checkedWriteSequence):
(WebCore::ISOFullBox::setVersion):
(WebCore::ISOFullBox::setFlags):
* Source/WebCore/platform/graphics/iso/ISOOriginalFormatBox.cpp:
(WebCore::ISOOriginalFormatBox::ISOOriginalFormatBox):
(WebCore::ISOOriginalFormatBox::parse):
* Source/WebCore/platform/graphics/iso/ISOOriginalFormatBox.h:
* Source/WebCore/platform/graphics/iso/ISOProtectionSchemeInfoBox.cpp:
(WebCore::ISOProtectionSchemeInfoBox::ISOProtectionSchemeInfoBox):
(WebCore::ISOProtectionSchemeInfoBox::parse):
(WebCore::ISOProtectionSchemeInfoBox::updateSize):
(WebCore::ISOProtectionSchemeInfoBox::partialSize const):
* Source/WebCore/platform/graphics/iso/ISOProtectionSchemeInfoBox.h:
* Source/WebCore/platform/graphics/iso/ISOProtectionSystemSpecificHeaderBox.cpp:
(WebCore::ISOProtectionSystemSpecificHeaderBox::commonSystemID const):
(WebCore::ISOProtectionSystemSpecificHeaderBox::ISOProtectionSystemSpecificHeaderBox):
(WebCore::ISOProtectionSystemSpecificHeaderBox::peekSystemID):
(WebCore::ISOProtectionSystemSpecificHeaderBox::parse):
(WebCore::ISOProtectionSystemSpecificHeaderBox::parseData):
(WebCore::ISOProtectionSystemSpecificHeaderBox::write const):
(WebCore::ISOProtectionSystemSpecificHeaderBox::writeData const):
(WebCore::ISOProtectionSystemSpecificHeaderBox::partialSize const):
* Source/WebCore/platform/graphics/iso/ISOProtectionSystemSpecificHeaderBox.h:
* Source/WebCore/platform/graphics/iso/ISOSchemeInformationBox.cpp:
(WebCore::ISOSchemeInformationBox::ISOSchemeInformationBox):
(WebCore::ISOSchemeInformationBox::parse):
(WebCore::ISOSchemeInformationBox::updateSize):
(WebCore::ISOSchemeInformationBox::partialSize const):
* Source/WebCore/platform/graphics/iso/ISOSchemeInformationBox.h:
* Source/WebCore/platform/graphics/iso/ISOSchemeTypeBox.cpp:
(WebCore::ISOSchemeTypeBox::ISOSchemeTypeBox):
(WebCore::ISOSchemeTypeBox::parse):
* Source/WebCore/platform/graphics/iso/ISOSchemeTypeBox.h:
* Source/WebCore/platform/graphics/iso/ISOTrackEncryptionBox.cpp:
(WebCore::ISOTrackEncryptionBox::ISOTrackEncryptionBox):
(WebCore::ISOTrackEncryptionBox::parse):
(WebCore::ISOTrackEncryptionBox::parsePayload):
* Source/WebCore/platform/graphics/iso/ISOTrackEncryptionBox.h:
* Source/WebCore/platform/graphics/iso/ISOVTTCue.cpp:
(WebCore::ISOWebVTTCue::ISOWebVTTCue):
* Source/WebCore/platform/graphics/iso/ISOVTTCue.h:
* Tools/TestWebKitAPI/Tests/WebCore/ISOBox.cpp:
(TestWebKitAPI::TEST(ISOBox, ISOFairPlayStreamingPsshBox)):
(TestWebKitAPI::serializeBox):
(TestWebKitAPI::TEST(ISOBox, FairPlay_KeyRequestInfoBox_Write)):
(TestWebKitAPI::TEST(ISOBox, FairPlay_KeyRequestInfoBox_Roundtrip)):
(TestWebKitAPI::TEST(ISOBox, FairPlay_InfoBox_Write)):
(TestWebKitAPI::TEST(ISOBox, FairPlay_InfoBox_Roundtrip)):
(TestWebKitAPI::TEST(ISOBox, FairPlay_KeyAssetIdBox_ParseAndRewrite)):
(TestWebKitAPI::TEST(ISOBox, FairPlay_KeyRequestBox_WriteWithFkriOnly)):
(TestWebKitAPI::TEST(ISOBox, FairPlay_KeyRequestBox_RoundtripWithAssetId)):
(TestWebKitAPI::TEST(ISOBox, FairPlay_InitDataBox_WriteAndRoundtrip)):
(TestWebKitAPI::TEST(ISOBox, FairPlay_PsshBox_ParseWriteParseRoundtrip)):
(TestWebKitAPI::TEST(ISOBox, PsshBox_V1_PartialSizeIncludesSystemIDAndDataSize)):

Canonical link: <a href="https://commits.webkit.org/319410@main">https://commits.webkit.org/319410@main</a>
</pre>

https://github.com/WebKit/WebKit/commit/774c34b5f6427b4834307054f535e912a8922626

| [✅ 🧪 style](https://ews-build.webkit.org/#/builders/38/builds/181351 "Passed style check") | [✅ 🛠 ios](https://ews-build.webkit.org/#/builders/159/builds/55298 "Built successfully") | [✅ 🛠 mac](https://ews-build.webkit.org/#/builders/168/builds/49100 "Built successfully") | [✅ 🛠 wpe](https://ews-build.webkit.org/#/builders/5/builds/191574 "Built successfully") | [✅ 🛠 win](https://ews-build.webkit.org/#/builders/59/builds/145677 "Built successfully") | ⏳ 🛠 ios-apple 
| [✅ 🧪 bindings](https://ews-build.webkit.org/#/builders/9/builds/183213 "Passed tests") | [✅ 🛠 ios-sim](https://ews-build.webkit.org/#/builders/155/builds/56602 "Built successfully") | [✅ 🛠 mac-AS-debug](https://ews-build.webkit.org/#/builders/156/builds/55019 "Built successfully") | [✅ 🧪 wpe-wk2](https://ews-build.webkit.org/#/builders/34/builds/141540 "Passed tests") | [  ~~🧪 win-tests~~](https://ews-build.webkit.org/#/builders/60/builds/98197 "The change is no longer eligible for processing. Pull Request was already closed when EWS attempted to process it.") | ⏳ 🛠 mac-apple 
| [✅ 🧪 webkitperl](https://ews-build.webkit.org/#/builders/11/builds/184306 "Passed tests") | [✅ 🧪 ios-wk2](https://ews-build.webkit.org/#/builders/162/builds/45216 "Passed tests") | [✅ 🧪 api-mac](https://ews-build.webkit.org/#/builders/18/builds/162291 "Passed tests") | [✅ 🧪 api-wpe](https://ews-build.webkit.org/#/builders/41/builds/121980 "Passed tests") | | ⏳ 🛠 vision-apple 
| | [✅ 🧪 ios-wk2-wpt](https://ews-build.webkit.org/#/builders/154/builds/43894 "Passed tests") | [✅ 🧪 api-mac-debug](https://ews-build.webkit.org/#/builders/165/builds/42165 "Passed tests") | [✅ 🧪 jsc-wpe](https://ews-build.webkit.org/#/builders/251/builds/3946 "Passed tests") | | 
| [✅ 🧪 jsc-x86-64](https://ews-build.webkit.org/#/builders/218/builds/10763 "Passed tests") | [✅ 🧪 api-ios](https://ews-build.webkit.org/#/builders/13/builds/154296 "Passed tests") | [✅ 🧪 mac-wk2](https://ews-build.webkit.org/#/builders/169/builds/41817 "Passed tests") | [✅ 🛠 gtk3-gcc](https://ews-build.webkit.org/#/builders/284/builds/2730 "Built successfully") | | 
| [✅ 🛠 🧪 jsc-debug-arm64](https://ews-build.webkit.org/#/builders/171/builds/42627 "Built successfully and passed tests") | [✅ 🛠 ios-safer-cpp](https://ews-build.webkit.org/#/builders/174/builds/47924 "Built successfully") | [✅ 🧪 mac-AS-debug-wk2](https://ews-build.webkit.org/#/builders/161/builds/46421 "Passed tests") | [✅ 🛠 gtk](https://ews-build.webkit.org/#/builders/2/builds/193502 "Built successfully") | | 
| | [✅ 🛠 vision](https://ews-build.webkit.org/#/builders/153/builds/54967 "Built successfully") | | [✅ 🧪 gtk-wk2](https://ews-build.webkit.org/#/builders/1/builds/150935 "Passed tests") | | 
| [✅ 🛠 🧪 merge](https://ews-build.webkit.org/#/builders/19/builds/40435 "Built successfully and passed tests") | [✅ 🛠 vision-sim](https://ews-build.webkit.org/#/builders/160/builds/55654 "Built successfully") | [  ~~🧪 mac-intel-wk2~~](https://ews-build.webkit.org/#/builders/170/builds/38440 "The change is no longer eligible for processing. Commit was outdated when EWS attempted to process it.") | [✅ 🧪 api-gtk](https://ews-build.webkit.org/#/builders/21/builds/150641 "Passed tests") | | 
| | [✅ 🧪 vision-wk2](https://ews-build.webkit.org/#/builders/164/builds/45224 "Passed tests") | [✅ 🛠 mac-safer-cpp](https://ews-build.webkit.org/#/builders/120/builds/127935 "Built successfully") | [✅ 🛠 playstation](https://ews-build.webkit.org/#/builders/134/builds/123535 "Built successfully") | | 
| | [✅ 🛠 tv](https://ews-build.webkit.org/#/builders/158/builds/54170 "Built successfully") | [✅ 🧪 mac-site-isolation](https://ews-build.webkit.org/#/builders/185/builds/16816 "Passed tests") | | | 
| | [✅ 🛠 tv-sim](https://ews-build.webkit.org/#/builders/157/builds/53552 "Built successfully") | | | | 
| | [✅ 🛠 watch](https://ews-build.webkit.org/#/builders/163/builds/53790 "Built successfully") | | | | 
| | [✅ 🛠 watch-sim](https://ews-build.webkit.org/#/builders/152/builds/53617 "Built successfully") | | | |

### DIFF
--- a/Source/WebCore/Modules/encryptedmedia/InitDataRegistry.cpp
+++ b/Source/WebCore/Modules/encryptedmedia/InitDataRegistry.cpp
@@ -32,7 +32,6 @@
 #include "ISOProtectionSystemSpecificHeaderBox.h"
 #include "NotImplemented.h"
 #include "SharedBuffer.h"
-#include <JavaScriptCore/DataView.h>
 #include <wtf/JSONValues.h>
 #include <wtf/NeverDestroyed.h>
 #include <wtf/text/Base64.h>
@@ -130,7 +129,7 @@ std::optional<Vector<std::unique_ptr<ISOProtectionSystemSpecificHeaderBox>>> Ini
     unsigned offset = 0;
     Vector<std::unique_ptr<ISOProtectionSystemSpecificHeaderBox>> psshBoxes;
 
-    auto view = JSC::DataView::create(buffer.tryCreateArrayBuffer(), offset, buffer.size());
+    auto view = buffer.span();
     while (auto optionalBoxType = ISOBox::peekBox(view, offset)) {
         auto& boxTypeName = optionalBoxType.value().first;
         auto& boxSize = optionalBoxType.value().second;
@@ -166,11 +165,7 @@ RefPtr<SharedBuffer> InitDataRegistry::extractFairPlayPsshFromCenc(const SharedB
     if (buffer.size() >= kCencMaxBoxSize)
         return nullptr;
 
-    auto arrayBuffer = buffer.tryCreateArrayBuffer();
-    if (!arrayBuffer)
-        return nullptr;
-
-    auto view = JSC::DataView::create(WTF::move(arrayBuffer), 0, buffer.size());
+    auto view = buffer.span();
     unsigned offset = 0;
     while (auto optionalBoxType = ISOBox::peekBox(view, offset)) {
         auto& [boxTypeName, boxSize] = optionalBoxType.value();
--- a/Source/WebCore/platform/graphics/avfoundation/ISOFairPlayStreamingPsshBox.h
+++ b/Source/WebCore/platform/graphics/avfoundation/ISOFairPlayStreamingPsshBox.h
@@ -38,27 +38,37 @@ public:
 
     static FourCC boxTypeName() { return std::span { "fpsi" }; }
 
+    void setScheme(FourCC scheme) { m_scheme = scheme; }
     FourCC scheme() const { return m_scheme; }
 
-    WEBCORE_EXPORT bool parse(JSC::DataView&, unsigned& offset) final;
+    WEBCORE_EXPORT bool parse(const ByteView&, unsigned& offset) final;
+    WEBCORE_EXPORT bool pack(MutableByteView&, unsigned& offset) const final;
 
 private:
+    uint64_t partialSize() const final { return ISOFullBox::partialSize() + sizeof(uint32_t); }
+
     FourCC m_scheme;
 };
 
 class ISOFairPlayStreamingKeyRequestInfoBox final : public ISOFullBox {
 public:
     WEBCORE_EXPORT ISOFairPlayStreamingKeyRequestInfoBox();
+    WEBCORE_EXPORT ISOFairPlayStreamingKeyRequestInfoBox(const ISOFairPlayStreamingKeyRequestInfoBox&);
+    ISOFairPlayStreamingKeyRequestInfoBox(ISOFairPlayStreamingKeyRequestInfoBox&&);
     WEBCORE_EXPORT ~ISOFairPlayStreamingKeyRequestInfoBox();
 
     static FourCC boxTypeName() { return std::span { "fkri" }; }
 
     using KeyID = Vector<uint8_t, 16>;
+    void setKeyID(KeyID keyID) { m_keyID = keyID; }
     const KeyID& keyID() const LIFETIME_BOUND { return m_keyID; }
 
-    WEBCORE_EXPORT bool parse(JSC::DataView&, unsigned& offset) final;
+    WEBCORE_EXPORT bool parse(const ByteView&, unsigned& offset) final;
+    WEBCORE_EXPORT bool pack(MutableByteView&, unsigned& offset) const final;
 
 private:
+    uint64_t partialSize() const final { return ISOFullBox::partialSize() + m_keyID.size(); }
+
     KeyID m_keyID;
 };
 
@@ -66,7 +76,7 @@ class ISOFairPlayStreamingKeyAssetIdBox final : public ISOBox {
 public:
     WEBCORE_EXPORT ISOFairPlayStreamingKeyAssetIdBox();
     WEBCORE_EXPORT ISOFairPlayStreamingKeyAssetIdBox(const ISOFairPlayStreamingKeyAssetIdBox&);
-    ISOFairPlayStreamingKeyAssetIdBox(ISOFairPlayStreamingKeyAssetIdBox&&) = default;
+    WEBCORE_EXPORT ISOFairPlayStreamingKeyAssetIdBox(ISOFairPlayStreamingKeyAssetIdBox&&);
     WEBCORE_EXPORT ~ISOFairPlayStreamingKeyAssetIdBox();
 
     ISOFairPlayStreamingKeyAssetIdBox& operator=(const ISOFairPlayStreamingKeyAssetIdBox&) = default;
@@ -74,10 +84,14 @@ public:
 
     static FourCC boxTypeName() { return std::span { "fkai" }; }
     const Vector<uint8_t>& data() const LIFETIME_BOUND { return m_data; }
+    Vector<uint8_t>& data() LIFETIME_BOUND { return m_data; }
 
-    WEBCORE_EXPORT bool parse(JSC::DataView&, unsigned& offset) final;
+    WEBCORE_EXPORT bool parse(const ByteView&, unsigned& offset) final;
+    WEBCORE_EXPORT bool pack(MutableByteView&, unsigned& offset) const final;
 
 private:
+    uint64_t partialSize() const final { return ISOBox::partialSize() + m_data.size(); }
+
     Vector<uint8_t> m_data;
 };
 
@@ -85,7 +99,7 @@ class ISOFairPlayStreamingKeyContextBox final : public ISOBox {
 public:
     WEBCORE_EXPORT ISOFairPlayStreamingKeyContextBox();
     WEBCORE_EXPORT ISOFairPlayStreamingKeyContextBox(const ISOFairPlayStreamingKeyContextBox&);
-    ISOFairPlayStreamingKeyContextBox(ISOFairPlayStreamingKeyContextBox&&) = default;
+    WEBCORE_EXPORT ISOFairPlayStreamingKeyContextBox(ISOFairPlayStreamingKeyContextBox&&);
     WEBCORE_EXPORT ~ISOFairPlayStreamingKeyContextBox();
 
     ISOFairPlayStreamingKeyContextBox& operator=(const ISOFairPlayStreamingKeyContextBox&) = default;
@@ -93,10 +107,14 @@ public:
 
     static FourCC boxTypeName() { return std::span { "fkcx" }; }
     const Vector<uint8_t>& data() const LIFETIME_BOUND { return m_data; }
+    Vector<uint8_t>& data() LIFETIME_BOUND { return m_data; }
 
-    WEBCORE_EXPORT bool parse(JSC::DataView&, unsigned& offset) final;
+    WEBCORE_EXPORT bool parse(const ByteView&, unsigned& offset) final;
+    WEBCORE_EXPORT bool pack(MutableByteView&, unsigned& offset) const final;
 
 private:
+    uint64_t partialSize() const final { return ISOBox::partialSize() + m_data.size(); }
+
     Vector<uint8_t> m_data;
 };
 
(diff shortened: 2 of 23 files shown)


Source/WebCore/Modules/encryptedmedia/InitDataRegistry.cpp
jyavenard: that's not necessary, it's the equivalent behaviour for std::optional<T>::operator== ; explicit form is unneeded.

jyavenard: why this change? `keyID.span()` seems cleaner to me .

Source/WebCore/platform/graphics/iso/ISOBox.cpp
jyavenard, line 172: could be made a two line
```suggestion
    if (firstSize == 1 && !checkedWrite<uint64_t>(m_size, view, offset, BigEndian))
        return false;
```

jyavenard: The mix of ArrayBuffer and and non gigacage memory (CFData or SharedBuffer) is dangerous and caused bug https://commits.webkit.org/309465@main due to the introduction of the Gigacage

We should avoid the use of ArrayBuffer whenever possible for non JSC code.

More problematic in this instance is the signature of write(JSC::DataView&) is that you can never directly write into a SharedBuffer or a Vector, we always need the double allocation (first gigacage allocation for the ArrayBuffer and then a SharedBuffer) and then read the data twice,

Write should take a span<uint8_t> and have no JSC in its path. 

This code is unused for now, so I don't think it's out of scope to stop using JSC::DataView and co. for new code. A big-endian helper to write is trivial.

Source/WebCore/platform/graphics/avfoundation/ISOFairPlayStreamingPsshBox.h
jyavenard, line 41: that's unused

jyavenard, line 63: that's unused

jyavenard, line 154: is this non-const version used?

jyavenard: so the writer never touch m_initDataBox ; shouldn't it?

jyavenard, line 139: This seems wrong; we override partialSize() for this box but not write(). write() doesn't consult partialSize() at all — the base one just emits size+type from m_size and stops, so the buffer ends up sized for the versions while nothing ever writes them. They come out as zeros and serialize() still reports success, so a round trip silently replaces the real versions with 0.

partialSize() / write() are a must have combination. Maybe we should make those two pure virtual so that you can't define one without defining the other ; that's more change though and would have to touch unrelated box for this commit. 
Maybe a comment that if you override partialSize() then write() should be overridden.

Source/WebCore/platform/graphics/iso/ISOBox.cpp
jyavenard, line 193: It seems to me that as m_size is in use but may be calculated differently (parse() or updateSize()), we should compare on output offset and m_size as a mismatch indicate an error.

something like:
```suggestion
    if (!write(view, offset) || offset != m_size)
```

would detect the error at least, but not which box did it; could even ASSERT

jyavenard (commented):
I think this needs some changes.

In the commit it mentions some tests in the list, I can't see them were they missed ?

Source/WebCore/platform/graphics/avfoundation/ISOFairPlayStreamingPsshBox.h
jernoble: This will be used by a subsequent patch.

jernoble: This will be used by a subsequent patch.

jernoble: Yes.

Source/WebCore/platform/graphics/iso/ISOBox.cpp
```diff
@@ -108,30 +146,87 @@ bool ISOBox::parse(DataView& view, unsigned& offset)
...
+        m_extendedType = WTF::move(extendedType);
     }
 
     return true;
 }
 
+bool ISOBox::write(MutableByteView& view, unsigned& offset) const
+{
+    uint32_t firstSize = 1u;
+    if (m_size <= 1lu << 31)
```
jyavenard, line 162:
```suggestion
    if (m_size <= 1lu << 32)
```

8 bytes size are to be used if element is > UINT32_MAX

and that's what is used for checking just above in `ISOBox::requiredSize()`

so here we have a conflict between the two: if size is > 1<<31 and < 1<<32
that requiredSize is 4 bytes short and we write 4 bytes too many

Source/WebCore/platform/graphics/iso/ISOBox.cpp
jyavenard: I still think my earlier comment apply. 

What about returning a RefPtr<SharedBuffer> to differentiate if there's an error (nullptr) or nothing to serialise (empty SharedBuffer) ?

Source/WebCore/platform/graphics/iso/ISOTrackEncryptionBox.cpp
```diff
@@ -101,6 +98,63 @@ bool ISOTrackEncryptionBox::parsePayload(DataView& view, unsigned& offset)
...
+    // reserved(1) + reserved-or-crypt/skip(1) + isProtected(1) + perSampleIVSize(1) + KID(16)
+    uint64_t size = ISOFullBox::partialSize() + 20;
+
+    if (m_defaultIsProtected == 1 && !m_defaultPerSampleIVSize)
+        size += 1 + m_defaultConstantIV.size();
+
+    return size;
+}
+
 bool ISOTrackEncryptionBox::parseWithoutTypeAndSize(std::span<const uint8_t> data)
```
jyavenard, line 158: this method is now redundant; we only added this to avoid the JSC' ArrayBuffer gigacage issue and that problem was removed entirely.

could be dropped.

Source/WebCore/Modules/encryptedmedia/InitDataRegistry.cpp
jyavenard, line 35: that's a dead header now, can remove it

Source/WebCore/platform/graphics/iso/ISOBox.cpp
```diff
@@ -108,30 +146,87 @@ bool ISOBox::parse(DataView& view, unsigned& offset)
...
+        if (!checkedWrite<uint64_t>(m_size, view, offset, BigEndian))
+            return false;
+    }
+
+    if (m_boxType == std::span { "uuid" }) {
+        ASSERT(m_extendedType);
+        if (!m_extendedType)
+            return false;
+
+        if (!checkedWrite<ExtendedType>(*m_extendedType, view, offset, BigEndian))
```
jyavenard, line 181: that too should be `checkedWriteSequence` 

that's pre-existing and read->write or write->read keeps it correct, but internally it's wrong to have it in the wrong order. and isn't consistent with the other writer that are writing similar array of bytes 

so:
```suggestion
        if (!checkedWriteSequence<uint8_t>(*m_extendedType, view, offset, BigEndian))
```

Source/WebCore/platform/graphics/iso/ISOBox.h
```diff
@@ -26,76 +26,131 @@
...
-    virtual bool parse(JSC::DataView&, unsigned& offset);
+    virtual bool parse(const ByteView&, unsigned& offset);
 
     enum Endianness { BigEndian, LittleEndian };
 
-    template<typename T, typename R, typename V>
-    static bool checkedRead(R& returnValue, V& view, unsigned& offset, Endianness endianness)
+    template<typename T, typename R>
+    static bool checkedRead(R& returnValue, const ByteView& view, unsigned& offset, Endianness endianness)
     {
```
jyavenard, line 81:
```suggestion
    static_assert(std::is_integral_v<T>, "non-scalar T would be byte-reversed; use checkedReadSequence/checkedWriteSequence for byte arrays")
```
or something like that.

same in checkedWrite
so we can't call checkedRead/checkedWrite that would otherwise silently do byte-reverse. Like those 16 bytes uuid, endianess is irrelevant for those and so we shouldn't use that method for those objects.

jyavenard (approved):
r=me with nits (though the size mismatch could be problematic with a fuzzer)

There's a few `#include <JavaScriptCore/DataView.h>` and ArrayBuffer.h in the affected files that are no longer necessary, could be removed.

Source/WebCore/platform/graphics/iso/ISOBox.h
jernoble: This method can be used for non-integral types. If you want to read or write an array of `uint64_t`, for example.

Source/WebCore/platform/graphics/iso/ISOBox.cpp
jernoble: I'll use `std::numeric_limits<uint32_t>::max()`.

Source/WebCore/platform/graphics/iso/ISOTrackEncryptionBox.cpp
jernoble: It's still used by `getKeyIDs(CMFormatDescriptionRef)`.

Source/WebCore/platform/graphics/iso/ISOBox.cpp
jernoble: Returning an empty `SharedBuffer` seems sufficient to signal an error.

Source/WebCore/platform/graphics/avfoundation/ISOFairPlayStreamingPsshBox.h
jernoble: It seems like this needs some refactoring. There should probably be a `ISOBox` `partialWrite()` method that is overridden by subclasses, and a non-virtual `ISOBox` `write()` that calls it. (Like the split between `read()` and `parse()`.)

This would allow the `write()` method to validate that the offset and size make sense.

